PR #3580, opened 2015-07-14
ISPN-5602 Create timeout scheduler executor

pruivo:
https://issues.jboss.org/browse/ISPN-5602

Notes for reviewer:

* I didn't add any configuration for the timeout since the only thing configurable is the thread name.
* In UnitTestTestNGListener, I changed to print the threads names only when trace is not enabled. (so no stack trace)
* In JGroupsTransport, findbugs reported ```Collection<URL> confs``` as a possible source of NPE so I changed it.


tristantarrant:
Pushed, thanks

PR #3581, opened 2015-07-14
ISPN-5603 Make BlockingTaskAwareExecutorService.checkForReadyTasks() …

pruivo:
…asynchronous

https://issues.jboss.org/browse/ISPN-5603

### DIFF
--- a/core/src/main/java/org/infinispan/factories/NamedExecutorsFactory.java
+++ b/core/src/main/java/org/infinispan/factories/NamedExecutorsFactory.java
@@ -201,8 +201,11 @@ public class NamedExecutorsFactory extends NamedComponentFactory implements Auto
          case SCHEDULED:
             return (T) new LazyInitializingScheduledExecutorService(executorFactory, threadFactory);
          case BLOCKING:
-            return (T) new LazyInitializingBlockingTaskAwareExecutorService(
-                  executorFactory, threadFactory, globalComponentRegistry.getTimeService());
+            final String controllerName = "Controller-" + shortened(componentName) + "-" +
+                  globalConfiguration.transport().nodeName();
+            return (T) new LazyInitializingBlockingTaskAwareExecutorService(executorFactory, threadFactory,
+                                                                            globalComponentRegistry.getTimeService(),
+                                                                            controllerName);
          default:
             return (T) new LazyInitializingExecutorService(executorFactory, threadFactory);
       }
--- a/core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
+++ b/core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
@@ -4,15 +4,16 @@ import org.infinispan.util.TimeService;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
-import java.util.ArrayList;
+import java.util.ArrayDeque;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Queue;
 import java.util.concurrent.AbstractExecutorService;
-import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -26,16 +27,19 @@ import java.util.concurrent.TimeUnit;
 public class BlockingTaskAwareExecutorServiceImpl extends AbstractExecutorService implements BlockingTaskAwareExecutorService {
 
    private static final Log log = LogFactory.getLog(BlockingTaskAwareExecutorServiceImpl.class);
-   private final BlockingQueue<BlockingRunnable> blockedTasks;
+   private final Queue<BlockingRunnable> blockedTasks;
    private final ExecutorService executorService;
    private final TimeService timeService;
+   private final ControllerThread controllerThread;
    private volatile boolean shutdown;
 
-   public BlockingTaskAwareExecutorServiceImpl(ExecutorService executorService, TimeService timeService) {
-      this.blockedTasks = new LinkedBlockingQueue<BlockingRunnable>();
+   public BlockingTaskAwareExecutorServiceImpl(String controllerThreadName, ExecutorService executorService, TimeService timeService) {
+      this.blockedTasks = new ConcurrentLinkedQueue<>();
       this.executorService = executorService;
       this.timeService = timeService;
       this.shutdown = false;
+      this.controllerThread = new ControllerThread(controllerThreadName);
+      controllerThread.start();
    }
 
    @Override
@@ -45,26 +49,16 @@ public class BlockingTaskAwareExecutorServiceImpl extends AbstractExecutorServic
       }
       if (runnable.isReady()) {
          doExecute(runnable);
+         if (log.isTraceEnabled()) {
+            log.tracef("Added a new task directly: %s task(s) are waiting", blockedTasks.size());
+         }
       } else {
+         //we no longer submit directly to the executor service.
          blockedTasks.offer(runnable);
-         //case: T1 is adding a task and T2 is releasing one. problem to solve:
-         //T1: is adding a new task. runnable.isReady() returns false
-         //T2: meanwhile, T2 releases a resources that is blocking the T1's runnable
-         //T2: also, T2 invokes checkForReadyTasks(), that does nothing because the queue is empty
-         //T1: continues and add the runnable to the queue
-         //problem: if no more interaction with this object is done, the runnable will be kept in the queue forever.
-
-         boolean checkPendingTasks;
-         synchronized (blockedTasks) {
-            //to synchronize with the thread invoking checkForReadyTasks();
-            checkPendingTasks = runnable.isReady();
+         controllerThread.checkForReadyTask();
+         if (log.isTraceEnabled()) {
+            log.tracef("Added a new task to the queue: %s task(s) are waiting", blockedTasks.size());
          }
-         if (checkPendingTasks) {
-            checkForReadyTasks();
-         }
-      }
-      if (log.isTraceEnabled()) {
-         log.tracef("Added a new task: %s task(s) are waiting", blockedTasks.size());
       }
    }
 
@@ -76,7 +70,8 @@ public class BlockingTaskAwareExecutorServiceImpl extends AbstractExecutorServic
    @Override
    public List<Runnable> shutdownNow() {
       shutdown = true;
-      List<Runnable> runnableList = new LinkedList<Runnable>();
+      controllerThread.interrupt();
+      List<Runnable> runnableList = new LinkedList<>();
       runnableList.addAll(executorService.shutdownNow());
       runnableList.addAll(blockedTasks);
       return runnableList;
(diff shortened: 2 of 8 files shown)


core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
```diff
@@ -142,4 +115,75 @@ private void doExecute(BlockingRunnable runnable) {
          blockedTasks.offer(runnable);
       }
    }
+
+   private class ControllerThread extends Thread {
+
+      private final LinkedList<BlockingRunnable> readyList;
```
wburns: Seems readyList can be just a locally scoped variable in the run method.

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
```diff
@@ -142,4 +115,75 @@ private void doExecute(BlockingRunnable runnable) {
...
+      private final Semaphore semaphore;
+      private volatile boolean interrupted;
+
+      public ControllerThread(String controllerThreadName) {
+         super(controllerThreadName);
+         readyList = new LinkedList<>();
```
wburns: I normally prefer ArrayDeque over LinkedList since we aren't adding/removing elements from the middle

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
```diff
@@ -142,4 +115,75 @@ private void doExecute(BlockingRunnable runnable) {
...
+                  iterator.remove();
+                  readyList.add(runnable);
+               }
+            }
+
+            if (log.isTraceEnabled()) {
+               log.tracef("Tasks to be executed=%s, still pending=%s", readyList.size(), blockedTasks.size());
+            }
+
+            while (!readyList.isEmpty()) {
```
wburns: I would recommend to use either an iterator calling remove after each or calling pollFirst in a loop would be much better (latter seems better to me).  Although it won't make that much of a difference if you leave it as is either.

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
```diff
@@ -29,13 +29,16 @@
    private final BlockingQueue<BlockingRunnable> blockedTasks;
    private final ExecutorService executorService;
    private final TimeService timeService;
+   private final ControllerThread controllerThread;
    private volatile boolean shutdown;
 
-   public BlockingTaskAwareExecutorServiceImpl(ExecutorService executorService, TimeService timeService) {
-      this.blockedTasks = new LinkedBlockingQueue<BlockingRunnable>();
+   public BlockingTaskAwareExecutorServiceImpl(String controllerThreadName, ExecutorService executorService, TimeService timeService) {
+      this.blockedTasks = new LinkedBlockingQueue<>();
```
wburns: I wonder is there a reason this is specifically a BlockingQueue?  We don't block waiting for elements to be added, we instead use the semaphore and we don't have a bounds (besides Integer.MAX_VALUE).  Can this not just be a standard ConcurrentLinkedQueue ?

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
```diff
@@ -29,13 +29,16 @@
...
       this.executorService = executorService;
       this.timeService = timeService;
       this.shutdown = false;
+      this.controllerThread = new ControllerThread(controllerThreadName);
+      controllerThread.start();
```
wburns: I don't think we can start any arbitrary thread when we are used in a container like jboss.  I believe we may have to make another thread execturo?  @tristantarrant can you confirm that?

pruivo: I'm open to suggestions :)

tristantarrant: Well, WildFly 9 has deprecated the thread subsystem and all subsystems requiring their own threads/executors need to handle them themselves, so I think that is not a hurdle.

core/src/main/java/org/infinispan/factories/NamedExecutorsFactory.java
```diff
@@ -201,8 +201,10 @@ public void stop() {
          case SCHEDULED:
             return (T) new LazyInitializingScheduledExecutorService(executorFactory, threadFactory);
          case BLOCKING:
-            return (T) new LazyInitializingBlockingTaskAwareExecutorService(
-                  executorFactory, threadFactory, globalComponentRegistry.getTimeService());
+            return (T) new LazyInitializingBlockingTaskAwareExecutorService(executorFactory, threadFactory,
+                                                                            globalComponentRegistry.getTimeService(),
+                                                                            "Controller-" + shortened(componentName) +
+                                                                                  "-" + globalConfiguration.transport().nodeName());
```
danberindei: I think a local variable for the thread name would look better.

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
```diff
@@ -43,26 +47,9 @@ public final void execute(BlockingRunnable runnable) {
...
-         synchronized (blockedTasks) {
-            //to synchronize with the thread invoking checkForReadyTasks();
-            checkPendingTasks = runnable.isReady();
-         }
-         if (checkPendingTasks) {
-            checkForReadyTasks();
-         }
-      }
+      //we no longer submit directly to the executor service.
+      blockedTasks.offer(runnable);
```
danberindei: Does this mean we no longer "overflow" to the OOB thread pool?

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
```diff
@@ -76,9 +63,10 @@ public void shutdown() {
    @Override
    public List<Runnable> shutdownNow() {
       shutdown = true;
-      List<Runnable> runnableList = new LinkedList<Runnable>();
+      List<Runnable> runnableList = new LinkedList<>();
       runnableList.addAll(executorService.shutdownNow());
       runnableList.addAll(blockedTasks);
+      controllerThread.interrupt();
```
danberindei: Shouldn't you interrupt the controller first?

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
```diff
@@ -142,4 +116,79 @@ private void doExecute(BlockingRunnable runnable) {
...
+               }
+            }
+
+            if (log.isTraceEnabled()) {
+               log.tracef("Tasks to be executed=%s, still pending=~%s", readyList.size(), size);
+            }
+
+            BlockingRunnable runnable;
+            while ((runnable = readyList.pollFirst()) != null) {
+               doExecute(runnable);
```
danberindei: Could this block the controller thread if the remote thread pool's is configured with `CallerRunsPolicy`?

core/src/test/java/org/infinispan/executors/BlockingTaskAwareExecutorServiceTest.java
```diff
@@ -53,33 +47,24 @@ public boolean isSatisfied() throws Exception {
...
 
          for (DoSomething doSomething : tasks) {
             assert !doSomething.isReady();
             assert !doSomething.isExecuted();
          }
 
-         for (DoSomething doSomething : tasks) {
-            doSomething.markReady();
-         }
+         tasks.forEach(BlockingTaskAwareExecutorServiceTest.DoSomething::markReady);
```
danberindei: I think `t -> t.markReady()` would have been easier on the eyes here.

core/src/test/java/org/infinispan/executors/BlockingTaskAwareExecutorServiceTest.java
```diff
@@ -88,9 +73,9 @@ public boolean isSatisfied() throws Exception {
    }
 
    private BlockingTaskAwareExecutorServiceImpl createExecutorService() {
-      return new BlockingTaskAwareExecutorServiceImpl(new ThreadPoolExecutor(1, 2, 60, TimeUnit.SECONDS,
-                                                                             new LinkedBlockingQueue<Runnable>(1000),
-                                                                             new DummyThreadFactory()),
+      return new BlockingTaskAwareExecutorServiceImpl("BlockingAwareTest-Controller", new ThreadPoolExecutor(1, 2, 60, TimeUnit.SECONDS,
```
danberindei: Please use the full test class name for the controller thread, it's better for log filtering. Also, some local variables would help keep the code to the left.

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
pruivo: I hope so :)

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
pruivo: yes. if you configure it to use WithinThreadExecutor, we hit the same problem. 
What I can do is to change the default to AbortPolicy (the task will be enqueued again). Other suggestions are welcome :D

core/src/test/java/org/infinispan/executors/BlockingTaskAwareExecutorServiceTest.java
pruivo: the IDEA puts a yellow line with a warning with it :) and it is more annoying lol

pruivo:
@danberindei updated!

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
danberindei: Please change the default to AbortPolicy then, CallerRunsPolicy is ok if we overflow to the OOB thread pool, but not if it blocks all other commands.

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
danberindei: Thinking about this some more, do we really need to go through the controller thread here? I fear adding an extra context switch per command is going to increase latency, especially in low contention scenarios (or for commands that only block to wait for topology changes, like commit).

core/src/test/java/org/infinispan/executors/BlockingTaskAwareExecutorServiceTest.java
danberindei: Fair enough :)

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
pruivo: with the default configuration no, we don't. but a user could inject their own executor service and we can end up with OOB threads executing the commands. 

IMO, most of the cases our executor service is used, so it may be good to go directly through the executor service if the task is ready.

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
pruivo: well, we have a little problem. the ThreadPoolFactory is the same for all executors service in Infinispan and the AbortPolicy causes a lot of problem in the others. I have half of the test suite failing with RejectedExecutionException (most of them in the state transfer executor)

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
danberindei: I wasn't actually thinking of the overflow to the OOB pool as a bad thing, because I think many users only upped the number of OOB threads and kept the default size for the remote thread pool (also because the remote thread pool configuration was broken in the server for a really long time). Of course, with the new locking we should require less remote threads, so it shouldn't matter much either way.

Would you like to try going back to direct execution here, or leave it for another PR?

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
danberindei: Oops... then perhaps it makes sense to create another issue to deal with creating a separate thread factory and possibly direct execution (without going through the controller thread).

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
pruivo: the direct execution does not break anything (at leas in my pc). If you want, I can put it back.

danberindei: Please do, and I'll integrate the PR.

pruivo:
@danberindei updated!

danberindei:
Integrated, thanks Pedro!

PR #3582, opened 2015-07-14
ISPN-5597 Verify if clustered Arquillian based tests are really clust…

mgencur:
…ered

https://issues.jboss.org/browse/ISPN-5597

I've tried all tests in integrationtests/ folder that use arquillian and changing the protocol does not change the test results. Anyway, I've changed the protocol for future tests that might be affected.


danberindei:
Integrated, thanks Martin!

PR #3586, opened 2015-07-15
ISPN-5609 Create CommandUUID

pruivo:
https://issues.jboss.org/browse/ISPN-5609


core/src/main/java/org/infinispan/commands/CommandUUID.java
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Represents an unique identified for non-transaction write commands.
+ *
+ * It is used to lock the key for a specific command.
+ *
+ * @author Pedro Ruivo
+ * @since 8.0
+ */
+public class CommandUUID {
```
danberindei: The `UUID` name usually implies a particular representation: http://tools.ietf.org/html/rfc4122.html

`CommandId` would also be too close to our existing command ids, so I would suggest something like `CommandInvocationId` or `NonTxInvocationId`.

core/src/main/java/org/infinispan/commands/CommandUUID.java
```diff
@@ -0,0 +1,94 @@
...
+
+   private final Address address;
+   private final long id;
+
+   private CommandUUID(Address address, long id) {
+      this.address = address;
+      this.id = id;
+   }
+
+   public static CommandUUID generateUUID(Address address) {
```
danberindei: I would also move the id generation to a (global) component, and avoid using static methods.

danberindei:
@pruivo Last CI build also had 2 weird failures in the spring module, please have a look.

core/src/main/java/org/infinispan/commands/CommandUUID.java
pruivo: in my mind, UUID is just an universal id. 
I'll change it to CommandInvocationId

core/src/main/java/org/infinispan/commands/CommandUUID.java
pruivo: Is it worth it? I mean, create a global component will take memory that does not need to be wasted.

pruivo:
@danberindei updated. I'm not able to reproduce the failures but it does not seem related. Mainly because the ids are not used anywhere yet. 

I left a question to you: https://github.com/infinispan/infinispan/pull/3586#discussion_r34992226

pruivo:
@danberindei no problem about the commit message :)

danberindei:
@pruivo Oops, I missed something: We should probably not generate an id for write commands in a transactional cache (maybe with the exception of PFER?)

PR #3587, opened 2015-07-16
ISPN-3076 Upgrade to JBoss Logging 3.2.1

tristantarrant:
https://issues.jboss.org/browse/ISPN-3076


wburns:
Looks like there are some bugs related to osgi from this.

tristantarrant:
Updated the PR with much more stuff:
- ported all custom appenders/filters to the log4j2 api
- actually run the tests with log4j2
- upgrade Karaf and Pax Exam to work with log4j2
- upgrade Hibernate ORM

There is still an issue with the OSGi JPA tests, which I'm trying to resolve with help.

tristantarrant:
Thanks to @isavin  for providing the needed fixes. This is now ready

danberindei:
@tristantarrant there's still this warning message I was seeing when I first tried switching to log4j2. Not sure how much of a problem it is:

    2015-07-21 19:58:45,981 WARN The Logger org.infinispan.remoting.transport.jgroups.CommandAwareRpcDispatcher was created with the message factory org.apache.logging.log4j.message.ParameterizedMessageFactory@18fb808d and is now requested with the message factory org.apache.logging.log4j.message.StringFormatterMessageFactory@6f84895e, which may create log events with unexpected formatting.

There's also a bad format specifier in `AbstractInfinispanTest.ThreadCleaner.close()` that was ignored by log4j 1.2, but writes a lot of stack traces to the console with log4j2.

tristantarrant:
I have fixed AbstractInfinispanTest.ThreadCleaner.close() issue. The other is caused by JGroups attempting to retrieve a logger for MessageDispatcher using LogFactory.getLog(getClass()) which, when using inheritance, means it will request another logger for CommandAwareRpcDispatcher. I'll issue a fix for JGroups.

danberindei:
@tristantarrant This hack seems to get rid of the warning, you you think it's worth including it in the PR?

    private static final Log log = LogFactory.getLog(CommandAwareRpcDispatcher.Hack.class);

    /**
     * Make sure the logger we retrieve with our LogFactory is not the same as the one used by JGroups,
     * because they use JGroups uses a different formatter.
     */
    private static class Hack {
    }

tristantarrant:
Nah, let's wait for 3.6.5 which will include https://github.com/belaban/JGroups/pull/211

danberindei:
@tristantarrant I ran the test suite again and found a couple more problems:

1. Using `+` instead of `,`:

 	at org.infinispan.distribution.topologyaware.TopologyAwareConsistentHashFactoryTest.testConsistencyWhenNodeLeaves(TopologyAwareConsistentHashFactoryTest.java:427)

2. Trying to log a mock `Throwable` from `LoggingCallable` (used by `fork()`) throws a NPE:

	at org.infinispan.interceptors.distribution.L1WriteSynchronizerTest.lambda$testSpawnedThreadBlockingException$150(L1WriteSynchronizerTest.java:181)

I don't see any reason why the test couldn't create a proper `Exception`.

If you fix these 2 and integrate JGroups 3.6.5, the build output should look clean again even with TRACE enabled.

tristantarrant:
@danberindei fixed both 1. and 2. above, also fixed another issue in AbstractInfinispanTest where a %i was used in place of %d. As for 3.6.5, we'll just have to wait for it :)

tristantarrant:
Ignore the CI checks, they were fine before and failed because the agents were reverted to an older JDK

danberindei:
Integrated, thanks Tristan!

PR #3592, opened 2015-07-20
ISPN-5611 Make the remote iterator to respect the RequestBalancingStr…

gustavonalle:
…ategy in the Hot Rod client

https://issues.jboss.org/browse/ISPN-5611


gustavonalle:
I don't think it'd work. There's no guarantee that a Hot Rod server will exist at localhost, and even assuming that, in the event of failover, a new server needs to be picked to carry on with the interation

gustavonalle:
I see where you are getting to. And it can be achieved with an external balancing strategy, so this PR is really about removing the logic from the client itself and make it pluggable

pruivo:
ok @gustavonalle . Testing and pushing...

pruivo:
integrated! thanks @gustavonalle !

PR #3594, opened 2015-07-20
ISPN-5620 Create PendingLockManager

pruivo:
https://issues.jboss.org/browse/ISPN-5620

PendingLockManager will be responsible to check for older topology id transaction that may have locks acquired. Newer transactions would have to wait for them to complete before proceeding in lock acquisition.

### DIFF
--- a/core/src/main/java/org/infinispan/commands/tx/PrepareCommand.java
+++ b/core/src/main/java/org/infinispan/commands/tx/PrepareCommand.java
@@ -11,6 +11,7 @@ import org.infinispan.commands.write.WriteCommand;
 import org.infinispan.commons.hash.Hash;
 import org.infinispan.commons.hash.MurmurHash3;
 import org.infinispan.commons.util.InfinispanCollections;
+import org.infinispan.context.Flag;
 import org.infinispan.context.InvocationContext;
 import org.infinispan.context.impl.RemoteTxInvocationContext;
 import org.infinispan.context.impl.TxInvocationContext;
@@ -230,6 +231,9 @@ public class PrepareCommand extends AbstractTransactionBoundaryCommand {
       }
       Set<Object> set = new HashSet<>(modifications.length);
       for (WriteCommand wc : modifications) {
+         if (wc.hasFlag(Flag.SKIP_LOCKING)) {
+            continue;
+         }
          switch (wc.getCommandId()) {
             case PutKeyValueCommand.COMMAND_ID:
             case RemoveCommand.COMMAND_ID:
--- a/core/src/main/java/org/infinispan/factories/LockManagerFactory.java
+++ b/core/src/main/java/org/infinispan/factories/LockManagerFactory.java
@@ -4,6 +4,9 @@ import org.infinispan.factories.annotations.DefaultFactoryFor;
 import org.infinispan.util.concurrent.locks.DeadlockDetectingLockManager;
 import org.infinispan.util.concurrent.locks.LockManager;
 import org.infinispan.util.concurrent.locks.LockManagerImpl;
+import org.infinispan.util.concurrent.locks.PendingLockManager;
+import org.infinispan.util.concurrent.locks.impl.DefaultPendingLockManager;
+import org.infinispan.util.concurrent.locks.impl.NoOpPendingLockManager;
 
 /**
  * Factory class that creates instances of {@link LockManager}.
@@ -11,14 +14,20 @@ import org.infinispan.util.concurrent.locks.LockManagerImpl;
  * @author Manik Surtani (<a href="mailto:manik@jboss.org">manik@jboss.org</a>)
  * @since 4.0
  */
-@DefaultFactoryFor(classes = LockManager.class)
+@DefaultFactoryFor(classes = {LockManager.class, PendingLockManager.class} )
 public class LockManagerFactory extends AbstractNamedCacheComponentFactory implements AutoInstantiableFactory {
+
    @Override
    public <T> T construct(Class<T> componentType) {
-      if (configuration.deadlockDetection().enabled()) {
-         return (T) new DeadlockDetectingLockManager();
-      } else {
-         return (T) new LockManagerImpl();
+      if (PendingLockManager.class.equals(componentType)) {
+         return componentType.cast(configuration.clustering().cacheMode().isClustered() ?
+                                         new DefaultPendingLockManager() :
+                                         NoOpPendingLockManager.getInstance());
+      } else if (LockManager.class.equals(componentType)) {
+         return componentType.cast(configuration.deadlockDetection().enabled() ?
+                                         new DeadlockDetectingLockManager() :
+                                         new LockManagerImpl());
       }
+      throw new IllegalArgumentException("Unexpected component type " + componentType + ".");
    }
 }
--- a/core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
@@ -1,10 +1,7 @@
 package org.infinispan.interceptors.locking;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-
 import org.infinispan.commands.DataCommand;
-import org.infinispan.commands.FlagAffectedCommand;
+import org.infinispan.commands.LocalFlagAffectedCommand;
 import org.infinispan.commands.read.GetCacheEntryCommand;
 import org.infinispan.commands.read.GetKeyValueCommand;
 import org.infinispan.commands.write.ClearCommand;
@@ -14,14 +11,20 @@ import org.infinispan.commands.write.InvalidateL1Command;
 import org.infinispan.commands.write.PutKeyValueCommand;
 import org.infinispan.commands.write.RemoveCommand;
 import org.infinispan.commands.write.ReplaceCommand;
-import org.infinispan.configuration.cache.CacheMode;
 import org.infinispan.container.DataContainer;
-import org.infinispan.container.EntryFactory;
+import org.infinispan.context.Flag;
 import org.infinispan.context.InvocationContext;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.interceptors.base.CommandInterceptor;
 import org.infinispan.util.concurrent.TimeoutException;
 import org.infinispan.util.concurrent.locks.LockManager;
+import org.infinispan.util.concurrent.locks.LockUtil;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.stream.Stream;
 
 /**
  * Base class for various locking interceptors in this package.
@@ -33,15 +36,13 @@ public abstract class AbstractLockingInterceptor extends CommandInterceptor {
 
    protected LockManager lockManager;
    protected DataContainer<Object, Object> dataContainer;
-   protected EntryFactory entryFactory;
    protected ClusteringDependentLogic cdl;
 
    @Inject
-   public void setDependencies(LockManager lockManager, DataContainer<Object, Object> dataContainer, EntryFactory entryFactory,
+   public void setDependencies(LockManager lockManager, DataContainer<Object, Object> dataContainer,
                                ClusteringDependentLogic cdl) {
       this.lockManager = lockManager;
       this.dataContainer = dataContainer;
-      this.entryFactory = entryFactory;
       this.cdl = cdl;
    }
 
@@ -80,14 +81,13 @@ public abstract class AbstractLockingInterceptor extends CommandInterceptor {
    protected abstract Object visitDataWriteCommand(InvocationContext ctx, DataWriteCommand command) throws Throwable;
 
    // We need this method in here because of putForExternalRead
-   protected Object visitNonTxDataWriteCommand(InvocationContext ctx, DataWriteCommand command) throws Throwable {
+   protected final Object visitNonTxDataWriteCommand(InvocationContext ctx, DataWriteCommand command) throws Throwable {
       try {
-         if (!shouldLock(command.getKey(), command))
+         if (hasSkipLocking(command) || !shouldLockKey(command.getKey())) {
             return invokeNextInterceptor(ctx, command);
-         lockKey(ctx, command);
+         }
+         lockAndRecord(ctx, command.getKey(), getLockTimeoutMillis(command));
          return invokeNextInterceptor(ctx, command);
-      } catch (Throwable te) {
-         throw cleanLocksAndRethrow(ctx, te);
       } finally {
          lockManager.unlockAll(ctx);
       }
@@ -96,14 +96,15 @@ public abstract class AbstractLockingInterceptor extends CommandInterceptor {
    @Override
    public final Object visitInvalidateCommand(InvocationContext ctx, InvalidateCommand command) throws Throwable {
       try {
-         boolean skipLocking = hasSkipLocking(command);
-         long lockTimeout = getLockAcquisitionTimeout(command, skipLocking);
-         for (Object key : command.getKeys()) {
-            lockKey(ctx, key, lockTimeout, skipLocking);
+         if (hasSkipLocking(command)) {
+            return invokeNextInterceptor(ctx, command);
          }
+         lockAllAndRecord(ctx, Arrays.asList(command.getKeys()), getLockTimeoutMillis(command));
          return invokeNextInterceptor(ctx, command);
       } finally {
-         if (!ctx.isInTxScope()) lockManager.unlockAll(ctx);
+         if (!ctx.isInTxScope()) {
+            lockManager.unlockAll(ctx);
+         }
       }
    }
 
@@ -114,28 +115,29 @@ public abstract class AbstractLockingInterceptor extends CommandInterceptor {
          return null;
       }
 
-      Object[] keys = command.getKeys();
+      if (hasSkipLocking(command)) {
+         return invokeNextInterceptor(ctx, command);
+      }
+
+      final Object[] keys = command.getKeys();
       try {
          if (keys != null && keys.length >= 1) {
-            ArrayList<Object> keysCopy = new ArrayList<>(Arrays.asList(keys));
-            boolean skipLocking = hasSkipLocking(command);
-            for (Object key : command.getKeys()) {
+            ArrayList<Object> keysToInvalidate = new ArrayList<>(keys.length);
+            for (Object key : keys) {
                try {
-                  lockKey(ctx, key, 0, skipLocking);
+                  lockAndRecord(ctx, key, 0);
+                  keysToInvalidate.add(key);
                } catch (TimeoutException te) {
                   getLog().unableToLockToInvalidate(key, cdl.getAddress());
-                  keysCopy.remove(key);
-                  if (keysCopy.isEmpty())
-                     return null;
                }
             }
-            command.setKeys(keysCopy.toArray());
+            if (keysToInvalidate.isEmpty()) {
+               return null;
+            }
+            command.setKeys(keysToInvalidate.toArray());
          }
          return invokeNextInterceptor(ctx, command);
-      } catch (Throwable te) {
-         throw cleanLocksAndRethrow(ctx, te);
-      }
-      finally {
+      } finally {
          command.setKeys(keys);
          if (!ctx.isInTxScope()) lockManager.unlockAll(ctx);
       }
@@ -146,23 +148,33 @@ public abstract class AbstractLockingInterceptor extends CommandInterceptor {
       return te;
    }
 
-   protected final boolean shouldLock(Object key, FlagAffectedCommand command) {
-      if (hasSkipLocking(command))
-         return false;
-      if (cacheConfiguration.clustering().cacheMode() == CacheMode.LOCAL)
-         return true;
-      boolean shouldLock = cdl.localNodeIsPrimaryOwner(key);
+   protected final long getLockTimeoutMillis(LocalFlagAffectedCommand command) {
+      return command.hasFlag(Flag.ZERO_LOCK_ACQUISITION_TIMEOUT) ? 0 : cacheConfiguration.locking().lockAcquisitionTimeout();
+   }
+
+   protected final boolean shouldLockKey(Object key) {
+      //only the primary owner acquires the lock.
+      boolean shouldLock = LockUtil.getLockOwnership(key, cdl) == LockUtil.LockOwnership.PRIMARY;
       getLog().tracef("Are (%s) we the lock owners for key '%s'? %s", cdl.getAddress(), key, shouldLock);
       return shouldLock;
    }
 
-   protected final void lockKey(InvocationContext ctx, DataWriteCommand command) throws InterruptedException {
-      boolean skipLocking = hasSkipLocking(command);
-      long lockTimeout = getLockAcquisitionTimeout(command, skipLocking);
-      lockKey(ctx, command.getKey(), lockTimeout, skipLocking);
+   protected final void lockAndRecord(InvocationContext context, Object key, long timeout) throws InterruptedException {
+      lockManager.acquireLock(context, key, timeout, false);
    }
 
-   protected final void lockKey(InvocationContext ctx, Object key, long timeoutMillis, boolean skipLocking) throws InterruptedException {
-      lockManager.acquireLockNoCheck(ctx, key, timeoutMillis, skipLocking);
+   protected final void lockAllAndRecord(InvocationContext context, Stream<?> keys, long timeout) throws InterruptedException {
+      for (Iterator<?> iterator = keys.iterator(); iterator.hasNext();) {
+         lockManager.acquireLock(context, iterator.next(), timeout, false);
+      }
+   }
+
+   protected final void lockAllAndRecord(InvocationContext context, Collection<?> keys, long timeout) throws InterruptedException {
+      if (keys.isEmpty()) {
+         return;
+      }
+      for (Object key : keys) {
+         lockManager.acquireLock(context, key, timeout, false);
+      }
    }
 }
--- a/core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
@@ -1,7 +1,5 @@
 package org.infinispan.interceptors.locking;
 
-import org.infinispan.atomic.DeltaCompositeKey;
-import org.infinispan.commands.CommandsFactory;
 import org.infinispan.commands.read.GetAllCommand;
 import org.infinispan.commands.tx.CommitCommand;
 import org.infinispan.commands.tx.PrepareCommand;
@@ -15,18 +13,14 @@ import org.infinispan.factories.annotations.Inject;
 import org.infinispan.partitionhandling.impl.PartitionHandlingManager;
 import org.infinispan.remoting.rpc.RpcManager;
 import org.infinispan.statetransfer.OutdatedTopologyException;
-import org.infinispan.transaction.impl.LocalTransaction;
-import org.infinispan.transaction.impl.TransactionTable;
-import org.infinispan.transaction.xa.CacheTransaction;
-import org.infinispan.transaction.xa.GlobalTransaction;
-import org.infinispan.util.TimeService;
-import org.infinispan.util.concurrent.TimeoutException;
+import org.infinispan.util.concurrent.locks.LockUtil;
+import org.infinispan.util.concurrent.locks.PendingLockManager;
 import org.infinispan.util.logging.Log;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.concurrent.TimeUnit;
-
-import static org.infinispan.commons.util.Util.toStr;
 
 /**
  * Base class for transaction based locking interceptors.
@@ -36,20 +30,17 @@ import static org.infinispan.commons.util.Util.toStr;
  */
 public abstract class AbstractTxLockingInterceptor extends AbstractLockingInterceptor {
 
-   protected TransactionTable txTable;
    protected RpcManager rpcManager;
-   private boolean clustered;
-   private TimeService timeService;
    private PartitionHandlingManager partitionHandlingManager;
+   private PendingLockManager pendingLockManager;
 
    @Inject
-   public void setDependencies(TransactionTable txTable, RpcManager rpcManager, TimeService timeService,
-                               PartitionHandlingManager partitionHandlingManager, CommandsFactory commandsFactory) {
-      this.txTable = txTable;
+   public void setDependencies(RpcManager rpcManager,
+                               PartitionHandlingManager partitionHandlingManager,
+                               PendingLockManager pendingLockManager) {
       this.rpcManager = rpcManager;
-      clustered = rpcManager != null;
-      this.timeService = timeService;
       this.partitionHandlingManager = partitionHandlingManager;
+      this.pendingLockManager = pendingLockManager;
    }
 
    @Override
@@ -108,17 +99,71 @@ public abstract class AbstractTxLockingInterceptor extends AbstractLockingInterc
     * becomes a primary owner a new transaction trying to obtain the "real" lock will have to wait for all backup
     * locks to be released. The backup lock will be released either by a commit/rollback/unlock command or by
     * the originator leaving the cluster (if recovery is disabled).
+    *
+    * @return {@code true} if the key was really locked.
     */
-   protected final void lockAndRegisterBackupLock(TxInvocationContext ctx, Object key, long lockTimeout, boolean skipLocking) throws InterruptedException {
-      //with DeltaCompositeKey, the locks should be acquired in the owner of the delta aware key.
-      Object keyToCheck = key instanceof DeltaCompositeKey ?
-            ((DeltaCompositeKey) key).getDeltaAwareValueKey() :
-            key;
-      if (cdl.localNodeIsPrimaryOwner(keyToCheck)) {
-         lockKeyAndCheckOwnership(ctx, key, lockTimeout, skipLocking);
-      } else if (cdl.localNodeIsOwner(keyToCheck)) {
-         ctx.getCacheTransaction().addBackupLockForKey(key);
+   protected final boolean lockOrRegisterBackupLock(TxInvocationContext<?> ctx, Object key, long lockTimeout)
+         throws InterruptedException {
+      final Log log = getLog();
+      final boolean trace = log.isTraceEnabled();
+
+      switch (LockUtil.getLockOwnership(key, cdl)) {
+         case PRIMARY:
+            if (trace) {
+               log.tracef("Acquiring locks on %s.", key);
+            }
+            checkPendingAndLockKey(ctx, key, lockTimeout);
+            return true;
+         case BACKUP:
+            if (trace) {
+               log.tracef("Acquiring backup locks on %s.", key);
+            }
+            ctx.getCacheTransaction().addBackupLockForKey(key);
+         default:
+            return false;
       }
+   }
+
+   /**
+    * Same as {@link #lockOrRegisterBackupLock(TxInvocationContext, Object, long)}
+    *
+    * @return a collection with the keys locked.
+    */
+   protected final Collection<Object> lockAllOrRegisterBackupLock(TxInvocationContext<?> ctx, Collection<Object> keys,
+                                                                  long lockTimeout) throws InterruptedException {
+      if (keys.isEmpty()) {
+         return Collections.emptyList();
+      }
+
+      final Log log = getLog();
+      final boolean trace = log.isTraceEnabled();
+
+      Collection<Object> keysToLock = new ArrayList<>(keys.size());
+
+      for (Object key : keys) {
+         switch (LockUtil.getLockOwnership(key, cdl)) {
+            case PRIMARY:
+               if (trace) {
+                  log.tracef("Acquiring locks on %s.", key);
+               }
+               keysToLock.add(key);
+               break;
+            case BACKUP:
+               if (trace) {
+                  log.tracef("Acquiring backup locks on %s.", key);
+               }
+               ctx.getCacheTransaction().addBackupLockForKey(key);
+            default:
+               break;
+         }
+      }
+
+      if (keysToLock.isEmpty()) {
+         return Collections.emptyList();
+      }
+
+      checkPendingAndLockAllKeys(ctx, keysToLock, lockTimeout);
+      return keysToLock;
    }
 
    /**
@@ -141,80 +186,17 @@ public abstract class AbstractTxLockingInterceptor extends AbstractLockingInterc
     * Note: The algorithm described below only when nodes leave the cluster, so it doesn't add a performance burden
     * when the cluster is stable.
     */
-   protected final void lockKeyAndCheckOwnership(InvocationContext ctx, Object key, long lockTimeout, boolean skipLocking) throws InterruptedException {
-      TxInvocationContext txContext = (TxInvocationContext) ctx;
-      int transactionTopologyId = -1;
-      boolean checkForPendingLocks = false;
-      if (clustered) {
-         CacheTransaction tx = txContext.getCacheTransaction();
-         boolean isFromStateTransfer = txContext.isOriginLocal() && ((LocalTransaction)tx).isFromStateTransfer();
-         // if the transaction is from state transfer it should not wait for the backup locks of other transactions
-         if (!isFromStateTransfer) {
-            transactionTopologyId = tx.getTopologyId();
-            if (transactionTopologyId != TransactionTable.CACHE_STOPPED_TOPOLOGY_ID) {
-               checkForPendingLocks = txTable.getMinTopologyId() < transactionTopologyId;
-            }
-         }
-      }
-
-      Log log = getLog();
-      boolean trace = log.isTraceEnabled();
-      if (checkForPendingLocks) {
-         if (trace)
-            log.tracef("Checking for pending locks and then locking key %s", toStr(key));
-
-         final long expectedEndTime = timeService.expectedEndTime(cacheConfiguration.locking().lockAcquisitionTimeout(),
-                                                                  TimeUnit.MILLISECONDS);
-
-         // Check local transactions first
-         waitForTransactionsToComplete(txContext, txTable.getLocalTransactions(), key, transactionTopologyId, expectedEndTime);
-
-         // ... then remote ones
-         waitForTransactionsToComplete(txContext, txTable.getRemoteTransactions(), key, transactionTopologyId, expectedEndTime);
-
-         // Then try to acquire a lock
-         if (trace)
-            log.tracef("Finished waiting for other potential lockers, trying to acquire the lock on %s", toStr(key));
-
-         final long remaining = timeService.remainingTime(expectedEndTime, TimeUnit.MILLISECONDS);
-         lockManager.acquireLock(ctx, key, remaining, skipLocking);
-      } else {
-         if (trace)
-            log.tracef("Locking key %s, no need to check for pending locks.", toStr(key));
-
-         lockManager.acquireLock(ctx, key, lockTimeout, skipLocking);
-      }
+   private void checkPendingAndLockKey(InvocationContext ctx, Object key, long lockTimeout) throws InterruptedException {
+      final long remaining = pendingLockManager.awaitPendingTransactionsForKey((TxInvocationContext<?>) ctx, key,
+                                                                               lockTimeout, TimeUnit.MILLISECONDS);
+      lockAndRecord(ctx, key, remaining);
    }
 
-   private void waitForTransactionsToComplete(TxInvocationContext txContext, Collection<? extends CacheTransaction> transactions,
-                                              Object key, int transactionTopologyId, long expectedEndTime) throws InterruptedException {
-      GlobalTransaction thisTransaction = txContext.getGlobalTransaction();
-      for (CacheTransaction tx : transactions) {
-         if (tx.getTopologyId() < transactionTopologyId) {
-            // don't wait for the current transaction
-            if (tx.getGlobalTransaction().equals(thisTransaction))
-               continue;
-
-            boolean txCompleted = false;
-
-            long remaining;
-            while ((remaining = timeService.remainingTime(expectedEndTime, TimeUnit.MILLISECONDS)) > 0) {
-               if (tx.waitForLockRelease(key, remaining)) {
-                  txCompleted = true;
-                  break;
-               }
-            }
-
-            if (!txCompleted) {
-               throw newTimeoutException(key, tx, txContext);
-            }
-         }
-      }
-   }
-
-   private TimeoutException newTimeoutException(Object key, CacheTransaction tx, TxInvocationContext txContext) {
-      return new TimeoutException("Could not acquire lock on " + key + " on behalf of transaction " +
-                                       txContext.getGlobalTransaction() + ". Waiting to complete tx: " + tx + ".");
+   private void checkPendingAndLockAllKeys(InvocationContext ctx, Collection<Object> keys, long lockTimeout)
+         throws InterruptedException {
+      final long remaining = pendingLockManager.awaitPendingTransactionsForAllKeys((TxInvocationContext<?>) ctx, keys,
+                                                                                   lockTimeout, TimeUnit.MILLISECONDS);
+      lockAllAndRecord(ctx, keys, remaining);
    }
 
    private boolean releaseLockOnTxCompletion(TxInvocationContext ctx) {
--- a/core/src/main/java/org/infinispan/interceptors/locking/NonTransactionalLockingInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/locking/NonTransactionalLockingInterceptor.java
@@ -53,19 +53,11 @@ public class NonTransactionalLockingInterceptor extends AbstractLockingIntercept
    public Object visitPutMapCommand(InvocationContext ctx, PutMapCommand command) throws Throwable {
       assertNonTransactional(ctx);
       try {
-         if (!command.isForwarded()) {
-            boolean skipLocking = hasSkipLocking(command);
-            long lockTimeout = getLockAcquisitionTimeout(command, skipLocking);
-            for (Object key : command.getMap().keySet()) {
-               if (shouldLock(key, command))
-                  lockKey(ctx, key, lockTimeout, skipLocking);
-            }
+         if (!command.isForwarded() && !hasSkipLocking(command)) {
+            lockAllAndRecord(ctx, command.getMap().keySet().stream().filter(this::shouldLockKey), getLockTimeoutMillis(command));
          }
          return invokeNextInterceptor(ctx, command);
-      } catch (Throwable te) {
-         throw cleanLocksAndRethrow(ctx, te);
-      }
-      finally {
+      } finally {
          lockManager.unlockAll(ctx);
       }
    }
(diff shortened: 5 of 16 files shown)


commons/src/main/java/org/infinispan/commons/util/InfinispanCollections.java
```diff
@@ -301,15 +301,28 @@ public int size() {
...
-   public static final <T> boolean containsAny(Collection<T> haystack, Collection<T> needle) {
+   public static <T> boolean containsAny(Collection<T> haystack, Collection<T> needle) {
       for (T element : needle) {
          if (haystack.contains(element))
             return true;
       }
       return false;
    }
+
+   public static <T> T findAnyCommon(Collection<T> c1, Collection<T> c2) {
```
wburns: Seems this might be nice to return Optional<T> maybe?

core/src/main/java/org/infinispan/factories/LockManagerFactory.java
```diff
@@ -11,10 +14,15 @@
...
  */
-@DefaultFactoryFor(classes = LockManager.class)
+@DefaultFactoryFor(classes = {LockManager.class, PendingLockManager.class} )
 public class LockManagerFactory extends AbstractNamedCacheComponentFactory implements AutoInstantiableFactory {
+   @SuppressWarnings("unchecked")
    @Override
    public <T> T construct(Class<T> componentType) {
+      if (PendingLockManager.class.equals(componentType)) {
+         final boolean clustered = configuration.clustering().cacheMode().isClustered();
+         return clustered ? (T) new DefaultPendingLockManager() : (T) NoOpPendingLockManager.getInstance();
```
wburns: I know I may be late to the game with this (the unchecked cast is everywhere), but is there any reason we don't use 
```
return componentType.cast(clustered ? new DefaultPendingLockManager() : NoOpPendingLockManager.getInstance());
```

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
```diff
@@ -146,23 +149,33 @@ protected final Throwable cleanLocksAndRethrow(InvocationContext ctx, Throwable
...
 
-   protected final void lockKey(InvocationContext ctx, Object key, long timeoutMillis, boolean skipLocking) throws InterruptedException {
-      lockManager.acquireLockNoCheck(ctx, key, timeoutMillis, skipLocking);
+   protected final void lockAllAndRecord(InvocationContext context, Collection<?> keys, long timeout) throws InterruptedException {
+      for (Object key : keys) {
+         lockManager.acquireLock(context, key, timeout, false);
+      }
+   }
+
+   protected final Collection<Object> filterKeysToLock(Collection<Object> keys) {
```
wburns: I think this would be better if we could tweak it to return a Stream<Object> instead so we don't need to create a collection that is just thrown away.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
```diff
@@ -109,15 +100,55 @@ protected final Object invokeNextAndCommitIf1Pc(TxInvocationContext ctx, Prepare
...
-   protected final void lockAndRegisterBackupLock(TxInvocationContext ctx, Object key, long lockTimeout, boolean skipLocking) throws InterruptedException {
-      //with DeltaCompositeKey, the locks should be acquired in the owner of the delta aware key.
-      Object keyToCheck = key instanceof DeltaCompositeKey ?
-            ((DeltaCompositeKey) key).getDeltaAwareValueKey() :
-            key;
-      if (cdl.localNodeIsPrimaryOwner(keyToCheck)) {
-         lockKeyAndCheckOwnership(ctx, key, lockTimeout, skipLocking);
-      } else if (cdl.localNodeIsOwner(keyToCheck)) {
-         ctx.getCacheTransaction().addBackupLockForKey(key);
+   protected final void lockAndRegisterBackupLock(TxInvocationContext<?> ctx, Object key, long lockTimeout, Action action)
```
wburns: I know this method was already named this, but should it not be an or instead of and?

commons/src/main/java/org/infinispan/commons/util/InfinispanCollections.java
pruivo: Didn't get what you mean. ```findAny()``` returns a random element and I want a random element common to both collections :)

wburns: Sorry I wrote my example wrong:
```
return c1.stream().filter(c2:contains).findAny();
```

core/src/main/java/org/infinispan/factories/LockManagerFactory.java
pruivo: it is never too late to change :) I'll change this class

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
```diff
@@ -80,14 +82,13 @@ public Object visitGetCacheEntryCommand(InvocationContext ctx, GetCacheEntryComm
...
       try {
-         if (!shouldLock(command.getKey(), command))
+         if (hasSkipLocking(command) || !shouldLockKey(command.getKey())) {
             return invokeNextInterceptor(ctx, command);
-         lockKey(ctx, command);
+         }
+         lockAndRecord(ctx, command.getKey(), getLockTimeoutMillis(command));
          return invokeNextInterceptor(ctx, command);
-      } catch (Throwable te) {
-         throw cleanLocksAndRethrow(ctx, te);
```
pruivo: nop, it is used in Optimistic and Pessimistic locking interceptor.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
```diff
@@ -109,15 +100,55 @@ protected final Object invokeNextAndCommitIf1Pc(TxInvocationContext ctx, Prepare
...
+               action.execute(ctx, key);
+            }
+            break;
+         case BACKUP:
+            ctx.getCacheTransaction().addBackupLockForKeys(Collections.singletonList(key));
+            break;
+         default:
+            break; //no-op
+      }
+      ctx.addAllAffectedKeys(Collections.singleton(key));
```
wburns: Do we need to do this even when not an owner?

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
wburns: Yeah but they don't override the method. It seems they should be doing a try finally block like we have here instead to me.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -0,0 +1,373 @@
...
+      return timeService.remainingTime(expectedEndTime, TimeUnit.MILLISECONDS);
+   }
+
+   private long checkForPendingAnyLocks(Collection<Object> keys, GlobalTransaction globalTransaction, int transactionTopologyId, long lockTimeout) throws InterruptedException {
+      if (trace)
+         log.tracef("Checking for pending locks and then locking key %s", toStr(keys));
+
+      final long expectedEndTime = timeService.expectedEndTime(lockTimeout, TimeUnit.MILLISECONDS);
+
+      final CacheTransaction lockOwner = waitForTransactionsToComplete(getTransactionWithAnyLockedKey(transactionTopologyId, keys, globalTransaction),
```
wburns: So I am trying to follow this, but correct me if I am wrong. What happens if you get in here with say tx2 that wants to write to k1 & k2.  However tx1 already has k1 locked.  Then while waiting tx3 comes in and locks k2 since no one is holding that lock.  Will tx2 mistakingly think it will be able to obtain the k2 lock when tx1 unlocks k1?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -0,0 +1,373 @@
...
+   public void inject(TransactionTable transactionTable, TimeService timeService,
+                      @ComponentName(TIMEOUT_SCHEDULE_EXECUTOR) ScheduledExecutorService timeoutExecutor) {
+      this.transactionTable = transactionTable;
+      this.timeService = timeService;
+      this.timeoutExecutor = timeoutExecutor;
+   }
+
+   @Override
+   public PendingLockPromise checkPendingForKey(TxInvocationContext<?> ctx, Object key, long time, TimeUnit unit) {
+      final GlobalTransaction globalTransaction = ctx.getGlobalTransaction();
```
wburns: A general question for this class, but it seems the purpose is just to wait for locks to be released not to actually acquire them.  Normally when I unblock like this I would expect the lock to be acquired as well.  But it seems this is still up to the LockManager to perform this duty?

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
wburns: You are saying because this is only ever invoked from as an owner.  Seems we should throw an exception in the default case then I would think?

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
pruivo: Well, I don't mind to use a Stream, but I would have to change everything to use a Stream instead of a Collection

wburns: There is only 1 reference that I saw that then invokes lockAllAndRecord and tbh lockAllAndRecord could be removed and references replaced with keys.forEach(k -> lockAndRecord(context, k, timeout)); if you wanted :)  But it is up to you, I am fine either way.

core/src/main/java/org/infinispan/commands/tx/PrepareCommand.java
```diff
@@ -230,6 +231,9 @@ public boolean isReturnValueExpected() {
       }
       Set<Object> set = new HashSet<>(modifications.length);
       for (WriteCommand wc : modifications) {
+         if (wc.hasFlag(Flag.SKIP_LOCKING)) {
```
wburns: I am sorry to piggyback this change on yours.  But it seems it would be more efficient to use a TreeSet if sort is true instead of sorting at the very end using the array.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
pruivo: actually there is one case where the node is not the owner. If the transaction is executed in a non-owner, no locks are acquired but we still have to populate the affected keys. The TxDistributionInterceptor will use the affected key to send the prepare.

wburns: Sounds good, just wanted to make sure :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
pruivo: this class is only to handle the case in which the topology changes and a backup owner becomes a primary owner. Since the backups does not acquire any locks, if the topology changes, the newer transaction needs to wait for the previous topology id transaction to complete before trying to acquire the lock in the LockManager. 
If it is not performing, the newer transaction would acquire the lock, since the lock manager would be empty for that key(s).
I hope this replies to this and the comment below.

core/src/main/java/org/infinispan/commands/tx/PrepareCommand.java
pruivo: I don't think so. If we have 2 keys with the same hash code, one of them would not be inserted in the TreeSet because the comparator would return 0.

wburns: Ah, yes forgot TreeSet doesn't use equals :(

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
wburns: Thanks!  Makes sense, I am not as familiar with the locking deep down nitty gritty details.

wburns:
Looks good in general to me, however @danberindei would want to look this over, I am not as familiar with the locking code.

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
pruivo: I have doubts because of the following:
* I don't like to have ```lock*``` and ```lockAll*``` variants. But in tx, you usually get multiple keys and in non-tx, you get a single key. If I could have a method that works fast enough for both scenarios, it would be great :)
* I don't wanna add a ```lockStream*``` variant. It would be good for PutMap but in tx I need to filter the keys by primary and backup owners (so I still need the ```lockAll*```). I'm doing the filtering here to avoid iterate over keys multiple times: https://github.com/infinispan/infinispan/pull/3594/files#diff-399714d17ac7a50a5a4e6b9f74620076R34
* If I use only ```lockStream*```, I could wrap the single key in a ```Stream``` and I could create two ```Stream```s for the scenario above, but it looks expensive.

I'm open to suggestions and yours expertise :D

wburns: Something like https://github.com/wburns/infinispan/commit/a3b997178d208a932853a80de79d191e1d9799d3 could be done, which I threw together quickly.  I didn't go into all of the methods, some of them are a bit harder.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
```diff
@@ -141,84 +172,25 @@ protected final void lockAndRegisterBackupLock(TxInvocationContext ctx, Object k
...
-               throw newTimeoutException(key, tx, txContext);
-            }
-         }
-      }
-   }
-
-   private TimeoutException newTimeoutException(Object key, CacheTransaction tx, TxInvocationContext txContext) {
-      return new TimeoutException("Could not acquire lock on " + key + " on behalf of transaction " +
-                                       txContext.getGlobalTransaction() + ". Waiting to complete tx: " + tx + ".");
+   private void lockAllAndCheckPending(InvocationContext ctx, Collection<Object> keys, long lockTimeout)
```
wburns: Seems it should be checkPendingAndLockAll, maybe?

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
wburns: Also for the one you linked you could do something like https://github.com/infinispan/infinispan/commit/38506cc65aa226efd9e2b38db3dce1969e906822

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
```diff
@@ -96,14 +95,15 @@ protected Object visitNonTxDataWriteCommand(InvocationContext ctx, DataWriteComm
...
    public final Object visitInvalidateCommand(InvocationContext ctx, InvalidateCommand command) throws Throwable {
       try {
-         boolean skipLocking = hasSkipLocking(command);
-         long lockTimeout = getLockAcquisitionTimeout(command, skipLocking);
-         for (Object key : command.getKeys()) {
-            lockKey(ctx, key, lockTimeout, skipLocking);
+         if (hasSkipLocking(command)) {
+            return invokeNextInterceptor(ctx, command);
          }
+         lockAllAndRecord(ctx, Arrays.stream(command.getKeys()), getLockTimeoutMillis(command));
```
danberindei: Wouldn't `Arrays.asList()` be enough here? `Arrays.stream` seems to have a lot of memory overhead.

pruivo: @wburns recommended to use a Stream to avoid creating unnecessary collections to throw away after the locking. 
https://github.com/infinispan/infinispan/pull/3594#discussion_r35098510

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
```diff
@@ -146,23 +147,28 @@ protected final Throwable cleanLocksAndRethrow(InvocationContext ctx, Throwable
...
+
+   protected final void lockAllAndRecord(InvocationContext context, Stream<?> keys, long timeout) throws InterruptedException {
+      for (Iterator<?> iterator = keys.iterator(); iterator.hasNext();) {
+         lockManager.acquireLock(context, iterator.next(), timeout, false);
+      }
    }
 
-   protected final void lockKey(InvocationContext ctx, Object key, long timeoutMillis, boolean skipLocking) throws InterruptedException {
-      lockManager.acquireLockNoCheck(ctx, key, timeoutMillis, skipLocking);
+   protected final Stream<Object> filterKeysToLock(Stream<Object> keys) {
```
danberindei: Unused?

pruivo: it is used in non-tx put map.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
```diff
@@ -109,16 +101,60 @@ protected final Object invokeNextAndCommitIf1Pc(TxInvocationContext ctx, Prepare
...
+               return true;
+            case BACKUP:
+               if (trace) {
+                  log.tracef("Acquiring backup locks on %s.", key);
+               }
+               ctx.getCacheTransaction().addBackupLockForKey(key);
+            default:
+               return false;
+         }
+      }).collect(Collectors.toList());
```
danberindei: I think the iterative version would be simpler here.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
```diff
@@ -109,16 +101,60 @@ protected final Object invokeNextAndCommitIf1Pc(TxInvocationContext ctx, Prepare
...
+
+      checkPendingAndLockAllKeys(ctx, keysToLock, lockTimeout);
+      if (action != null) {
+         keysToLock.forEach(key -> action.execute(ctx, key));
```
danberindei: Here too, returning the collection of locked keys and performing the action outside would be simpler.

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
danberindei: I think you can avoid `filterKeysToLock` without creating any new Stream or Collection instances :)

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
danberindei: My bad, my IntelliJ hadn't noticed all the file changes for some reason and it wasn't finding any reference.

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
pruivo: I don't see how. You have to filter each keys you should lock in the PutMapCommand

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
pruivo: I don't see what you mean here. ```keysToLock``` is a collection. Doing ```for (Object key : keysToLock)...``` or ```keysToLock.forEach(...)``` looks the same to me.

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
danberindei: `lockAllAndRecord` could have a filter parameter and apply the filter itself.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
danberindei: Actually, I think all keys need to be added to the affected keys collection on the originator of the tx, regardless of ownership.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
danberindei: I meant the caller can do the same thing itself, there's no need to support it in this method (with the additional null check).

pruivo: then the caller has to filter the keys again. otherwise, I had to copy the filtering code everywhere before this method is invoked.

core/src/main/java/org/infinispan/interceptors/locking/OptimisticLockingInterceptor.java
```diff
@@ -39,68 +30,27 @@
...
-   @Override
-   protected Log getLog() {
-      return log;
-   }
+   private boolean needToMarkReads;
 
    @Start
    public void start() {
-      if (cacheConfiguration.clustering().cacheMode() == CacheMode.LOCAL &&
+      needToMarkReads = cacheConfiguration.clustering().cacheMode() == CacheMode.LOCAL &&
```
danberindei: @pruivo Off-topic, do you have any idea why we only mark reads in LOCAL mode? The `unreliableReturnValue()` check also seems wrong, because it shouldn't affect `cache.get()`.

core/src/main/java/org/infinispan/interceptors/locking/OptimisticLockingInterceptor.java
```diff
@@ -139,6 +89,28 @@ public Object visitPutMapCommand(InvocationContext ctx, PutMapCommand command) t
...
 
    @Override
+   public Object visitLockControlCommand(TxInvocationContext ctx, LockControlCommand command) throws Throwable {
+      throw new InvalidCacheUsageException("Explicit locking is not allowed with optimistic caches!");
+   }
+
+   @Override
+   protected Log getLog() {
+      return log;
+   }
```
danberindei: I think it was better having `getLog()` at the beginning than between visit methods.

pruivo: my beautiful IDEA orders the methods by visibility: public -> protected -> private :D 

I'll put the order back :P

core/src/main/java/org/infinispan/interceptors/locking/OptimisticLockingInterceptor.java
pruivo: because in clustered mode we use the versions to know when the key is read. So, you don't need to mark the reads.

In local mode, we don't use any versioning, so we need to keep track of the keys read to perform the write skew check.

About the unsafe, it is before my time. But I think we need it because it affects the writes (remember that the put() should read the key before writing it (except if we have flags))

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
pruivo: I found a solution for it

core/src/main/java/org/infinispan/interceptors/locking/OptimisticLockingInterceptor.java
danberindei: I only noticed that it was moved, I didn't realize the other visit methods aren't from the `Visitor` interface :)

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
```diff
@@ -183,15 +184,9 @@ public void notifyOnTransactionFinished() {
...
-      if (txComplete) return true; //using an unsafe optimisation: if it's true, we for sure have the latest read of the value without needing memory barriers
-      final boolean potentiallyLocked = hasLockOrIsLockBackup(key);
-      if (trace) log.tracef("Transaction gtx=%s potentially locks key %s? %s", tx, key, potentiallyLocked);
-      if (potentiallyLocked) {
-         notifier.await(lockAcquisitionTimeout, TimeUnit.MILLISECONDS);
-         return txComplete;
-      }
-      return true;
+   public final boolean waitForLockRelease(long lockAcquisitionTimeout) throws InterruptedException {
+      notifier.await(lockAcquisitionTimeout, TimeUnit.MILLISECONDS);
```
danberindei: I missed that PR, but isn't `Notifier` a lot like `CompletedFuture`?

wburns: That was something I was wondering as well, it is very similar even to CountdownLatch, but a bit more advanced.

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
wburns: Actually I forgot ``Arrays.asList(..)`` just uses the backing array.  In that case it isn't bad - the old one created multiple collections.  But to be honest there isn't much difference between ``Arrays.asList(..).toStream()`` vs ``Arrays.asStream(..)`` except the latter is a bit more condense.

Also I think the reason @pruivo separated filtering and the locking is because not all uses require filtering, which is simple to do with streams but less efficient (collections) or not possible without multiple iterations (regular for loop).  Granted we could have a different method for each or pass in a always true filter for a single version, but I personally like the separation of ideas.  But like I said in my other comments this is all personal preference so I am fine either way :)

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
pruivo: as I said in the javadoc, the semantic is the same as ```new CountDownLatch(1)``` but with listeners. 
CompletedFuture has a lot of garbage not needed (I don't need exception, values and that stuff)

danberindei: Does the exception/return value support add a lot of overhead? Otherwise I'd rather standardize on CF everywhere instead of a custom synchronization primitive.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
```diff
@@ -108,17 +99,71 @@ protected final Object invokeNextAndCommitIf1Pc(TxInvocationContext ctx, Prepare
...
+      if (keys.isEmpty()) {
+         return Collections.emptyList();
       }
+
+      final Log log = getLog();
+      final boolean trace = log.isTraceEnabled();
+
+      Collection<Object> keysToLock = new ArrayList<>(keys.size());
+
+      for (Object key : keys) {
```
wburns: This used to add to the context before, is that still needed?

pruivo:
@danberindei @wburns updated!

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
pruivo: It is not needed. I find out that the visit* methods were added the key to the affected set. So, it is redundant to add it here :)

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
```diff
@@ -230,13 +225,37 @@ public void clearLockedKeys() {
...
+   public boolean containsAnyLockOrBackupLock(Stream<Object> keys) {
+      return findFirstLockedOrBackupLocked(keys) != null;
+   }
+
+   @Override
+   public Object findFirstLockedOrBackupLocked(Stream<Object> keys) {
+      Set<Object> lockedKeysCopy = lockedKeys;
+      Set<Object> backupKeyLocksCopy = backupKeyLocks;
+      if (lockedKeysCopy != null && backupKeyLocksCopy != null) {
+         return keys.filter(key -> lockedKeysCopy.contains(key) || backupKeyLocksCopy.contains(key)).findFirst().orElse(null);
```
wburns: This before used to return the primary lock first, is that okay that this behavior is changed?

wburns: Also I would suggest using findAny instead of findFirst (this case should be the same), unless you specify the collections must be ordered in some fashion.

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
pruivo: I don't think so. but using a CF as a latch sounds a little weird.

danberindei: Unlike a regular `Future`, CF never does any computation itself and can only be completed by someone else calling `complete()` or `completeExceptionally()`. Sounds a lot like a latch to me...

pruivo:
@danberindei @wburns updated! removed the notifier

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
pruivo: updated

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
```diff
@@ -345,11 +372,6 @@ public long getCreationTime() {
 
    @Override
    public final void addListener(TransactionCompletedListener listener) {
-      notifier.add(listener);
-   }
-
-   @Override
-   public void invoke(TransactionCompletedListener listener) {
-      listener.onCompletion();
+      notifier.thenRunAsync(listener::onCompletion);
```
danberindei: We shouldn't use the async methods without an explicit thread pool parameter, or they'll use the common pool.

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
```diff
@@ -230,13 +233,37 @@ public void clearLockedKeys() {
...
-      //stopgap fix for ISPN-2728. The real fix would be to synchronize this with the intrinsic lock.
+   @Override
+   public boolean containsLockOrBackupLock(Object key) {
       Set<Object> lockedKeysCopy = lockedKeys;
       Set<Object> backupKeyLocksCopy = backupKeyLocks;
       return (lockedKeysCopy != null && lockedKeysCopy.contains(key)) || (backupKeyLocksCopy != null && backupKeyLocksCopy.contains(key));
    }
 
+   @Override
+   public boolean containsAnyLockOrBackupLock(Stream<Object> keys) {
```
danberindei: This one really seems unused :)

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
```diff
@@ -230,13 +233,37 @@ public void clearLockedKeys() {
...
+      Set<Object> backupKeyLocksCopy = backupKeyLocks;
+      if (lockedKeysCopy != null && backupKeyLocksCopy != null) {
+         return keys.filter(key -> lockedKeysCopy.contains(key) || backupKeyLocksCopy.contains(key)).findAny().orElse(null);
+      } else if (lockedKeysCopy != null) {
+         return keys.filter(lockedKeysCopy::contains).findAny().orElse(null);
+      } else if (backupKeyLocksCopy != null) {
+         return keys.filter(backupKeyLocksCopy::contains).findAny().orElse(null);
+      }
+      return null;
+   }
```
danberindei: Doesn't this functional approach create a new stream instance for each intermediate step? Seems like a lot of garbage to just iterate over a collection of keys and call `contains()` twice in each iteration...

core/src/main/java/org/infinispan/util/concurrent/locks/LockUtil.java
```diff
@@ -0,0 +1,54 @@
...
+         return LockOwnership.PRIMARY;
+      } else if (clusteringDependentLogic.localNodeIsOwner(keyToCheck)) {
+         return LockOwnership.BACKUP;
+      } else {
+         return LockOwnership.NO_OWNER;
+      }
+   }
+
+
+   public enum LockOwnership {
```
danberindei: I don't think this really belongs in the `o.i.util.concurrent` package, it would fit more in `o.i.distribution`.

OTOH I was going to introduce a `KeyOwnership` myself, so I can move it later.

core/src/main/java/org/infinispan/util/concurrent/locks/PendingLockManager.java
```diff
@@ -0,0 +1,65 @@
...
+    * Same as {@link #awaitPendingTransactionsForAllKeys(TxInvocationContext, Collection, long, TimeUnit)} but
+    * non-blocking.
+    *
+    * @param ctx  the {@link TxInvocationContext}.
+    * @param keys the keys to check.
+    * @param time timeout.
+    * @param unit {@link TimeUnit} of {@code time}.
+    * @return a {@link PendingLockPromise}.
+    */
+   PendingLockPromise checkPendingForAllKeys(TxInvocationContext<?> ctx, Collection<Object> keys, long time, TimeUnit unit);
```
danberindei: Nitpicking: you could remove the `All` qualifier to make the method names a tiny bit shorter.

core/src/main/java/org/infinispan/util/concurrent/locks/PendingLockPromise.java
```diff
@@ -0,0 +1,64 @@
+package org.infinispan.util.concurrent.locks;
+
+/**
+ * A promise returned by {@link PendingLockManager}.
+ * <p>
+ * When a transaction need to waits for older topology transaction, this class allows it check the state. If the
+ * transaction does not need to wait, or all older transaction has finished or has timed out, the {@link #isReady()}
+ * method returns {@code true}. Also, it allows to add listeners to be notified when it is ready.
```
danberindei: Some grammar issues here: needs to wait, allows it to check, all older transactions have finished or have timed out, it allows adding (or it allows the caller to add).

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
pruivo: are you saying that we must never use ```stream()```?

core/src/main/java/org/infinispan/util/concurrent/locks/PendingLockManager.java
```diff
@@ -0,0 +1,65 @@
...
+    * acquired.
+    *
+    * @param ctx  the {@link TxInvocationContext}.
+    * @param keys the keys to check.
+    * @param time timeout.
+    * @param unit {@link TimeUnit} of {@code time}.
+    * @return the remaining timeout.
+    * @throws InterruptedException if the thread is interrupted while waiting.
+    */
+   long awaitPendingTransactionsForAllKeys(TxInvocationContext<?> ctx, Collection<Object> keys, long time, TimeUnit unit)
```
danberindei: Couldn't you replace usages of `awaitPendingTransactionsForAllKeys(ctx, keys, time, unit)` with `checkPendingForAllKeys(ctx, keys).get(time, unit)`? If yes, I suggest removing these methods.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -0,0 +1,387 @@
...
+      }
+      final int txTopologyId = getTopologyId(ctx);
+      if (txTopologyId != NO_PENDING_CHECK) {
+         if (trace) {
+            log.tracef("Checking for pending locks and then locking key %s", toStr(key));
+         }
+         return createAndStore(getTransactionWithLockedKey(txTopologyId, key, globalTransaction),
+                               globalTransaction, time, unit);
+      }
+      return PendingLockPromise.NO_OP;
```
danberindei: Doesn't seem correct, because it will override the timeout parameter with `Integer.MAX_VALUE`.

core/src/main/java/org/infinispan/util/concurrent/locks/PendingLockManager.java
pruivo: I could but I don't want to:
* I would have to create a PendingLockPromise everytime (for local operations, it is not created)
* The pendingLockPromiseMap cleanup would be messy. the remote commands create a promise and when they are executed, the await() will remove it from the map and check it state (throwing timeout or or not)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -0,0 +1,387 @@
...
+      }
+
+      if (lockOwner != null) {
+         timeout(lockOwner, globalTransaction);
+      }
+
+      return timeService.remainingTime(expectedEndTime, TimeUnit.MILLISECONDS);
+   }
+
+   private long checkForPendingAnyLocks(Collection<Object> keys, GlobalTransaction globalTransaction, int transactionTopologyId, long lockTimeout) throws InterruptedException {
```
danberindei: `checkForAnyPendingLocks()` would sound better.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
pruivo: it is not a problem because I don't check the remaining timeout when it is NO_OP. and NO_OP means it hasn't wait anything :)

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
danberindei: I guess so :) I don't see a need for it at such a low level, it's way too hard to figure out what the performance implications are.

core/src/main/java/org/infinispan/util/concurrent/locks/PendingLockManager.java
danberindei: Ok, but then the javadoc for `checkPendingTransactionsForKey()` is wrong. It should clearly mention that a call to `check` must be followed by a call to `await`.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
danberindei: I guess I'll have to wait and see how `checkPendingTransactionsForKey()` is used :)

core/src/main/java/org/infinispan/util/concurrent/locks/PendingLockManager.java
pruivo: ok

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
pruivo: are you  going to remove all the streaming stuff? great....

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -0,0 +1,387 @@
...
+   }
+
+   private Collection<PendingTransaction> getTransactionWithAnyLockedKey(int transactionTopologyId,
+                                                                         Collection<Object> keys,
+                                                                         GlobalTransaction globalTransaction) {
+
+      Function<CacheTransaction, PendingTransaction> filter = transaction -> {
+         if (transaction.getTopologyId() < transactionTopologyId &&
+               !transaction.getGlobalTransaction().equals(globalTransaction)) {
+            Object key = transaction.findAnyLockedOrBackupLocked(keys.stream());
```
danberindei: Create the stream only once outside the filter? Or even better, use the collection directly ;)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -0,0 +1,387 @@
...
+      final int totalSize = localTransactions.size() + remoteTransactions.size();
+      if (totalSize == 0) {
+         return Collections.emptyList();
+      }
+      final List<PendingTransaction> allTransactions = new ArrayList<>(totalSize);
+      final Collector<PendingTransaction, ?, Collection<PendingTransaction>> collector = Collectors.toCollection(() -> allTransactions);
+
+
+      if (!localTransactions.isEmpty()) {
+         localTransactions.stream().map(mapper).filter(tx -> tx != null).collect(collector);
```
danberindei: Mixing mapping and filtering looks kind of fishy to me, better use `forEach` I think.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
pruivo: you can't reuse a stream as you can't reuse iterators... I'll use the collection here but the cost is the same since I have to iterate over it...

pruivo:
@danberindei updated! (I think)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
danberindei: Interesting, I wasn't aware streams can't be reused.

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
danberindei: What bothers me with streams is that I don't really know what the performance is. Does Hotspot eliminate the allocations with escape analysis? Is it able to inline the lambdas? I think it's just not worth it to fret over all the performance aspects when the equivalent iterative code is just as easy to read.

I think streams are really useful when you're doing something with a lot of data (like the entire cache), and a couple of object allocations won't make a difference there. But if all the user wants to do is put a key in a map, every allocation becomes significant.

pruivo:
hey @danberindei , is there something missing?

core/src/main/java/org/infinispan/interceptors/locking/NonTransactionalLockingInterceptor.java
```diff
@@ -53,19 +53,11 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
       try {
-         if (!command.isForwarded()) {
-            boolean skipLocking = hasSkipLocking(command);
-            long lockTimeout = getLockAcquisitionTimeout(command, skipLocking);
-            for (Object key : command.getMap().keySet()) {
-               if (shouldLock(key, command))
-                  lockKey(ctx, key, lockTimeout, skipLocking);
-            }
+         if (!command.isForwarded() && !hasSkipLocking(command)) {
+            lockAllAndRecord(ctx, command.getMap().keySet().stream().filter(this::shouldLockKey), getLockTimeoutMillis(command));
```
danberindei: I see I forgot to complain about this instance of `stream()` ;)

pruivo: that is one place where it makes sense. otherwise, you pay the cost of creating an iterator, iterate and store in another collection.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -0,0 +1,392 @@
...
+
+   private Collection<PendingTransaction> filterAndCollectTransactions(Filter filter) {
+      final Collection<? extends CacheTransaction> localTransactions = transactionTable.getLocalTransactions();
+      final Collection<? extends CacheTransaction> remoteTransactions = transactionTable.getRemoteTransactions();
+      final int totalSize = localTransactions.size() + remoteTransactions.size();
+      if (totalSize == 0) {
+         return Collections.emptyList();
+      }
+      final List<PendingTransaction> allTransactions = new ArrayList<>(totalSize);
+      final Consumer<CacheTransaction> consumer = transaction -> {
```
danberindei: I meant something like this: http://paste.fedoraproject.org/247332/65314214/

I think putting the filtering and collection in a single place actually makes it easier to read, you don't have to learn what a null return value means in the filter. Plus it allows us to remove the 'And' from the method name, which is always a bonus for me :)

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
wburns: This stack trace talks about it a bit more [1].  The problem is that in some cases it is faster and others it may not be.  But in my findings most cases the performance difference is negligible both ways.  The reason I even mentioned Streams in the first place to Pedro was as a way to prevent additional entry iterations which imo is a much bigger performance issue.

[1] http://stackoverflow.com/questions/22658322/java-8-performance-of-streams-vs-collections#comment36791431_22670380

core/src/main/java/org/infinispan/interceptors/locking/NonTransactionalLockingInterceptor.java
danberindei: This is what I had in mind, forgot to link to it yesterday: https://github.com/danberindei/infinispan/commit/f64b5be248c263954cbea0b88d52dd623da4b48f

It's not such a big difference though, so you can keep the stream version if you prefer it.

pruivo: I prefer to keep the stream here. It looks better than check the filter for all the multiple keys commands when you only need them for PutMapCommand.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
pruivo: agree, I'll change it.

pruivo:
@danberindei updated!

core/src/main/java/org/infinispan/interceptors/locking/NonTransactionalLockingInterceptor.java
danberindei: Ok, although I'm not entirely convinced about your argument - you also have to call `.stream()` for each command when you only really need it for `PutMapCommand`.

pruivo: I don't understand what you mean. the PutMapCommand is the only one which uses stream(). the rest of the commands, I'm passing a collection as argument.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -0,0 +1,381 @@
...
+                                                                      GlobalTransaction globalTransaction) {
+      final Collection<PendingTransaction> pendingTransactions = new ArrayList<>();
+      forEachTransaction(transaction -> {
+         if (transaction.getTopologyId() < transactionTopologyId &&
+               !transaction.getGlobalTransaction().equals(globalTransaction) &&
+               transaction.containsLockOrBackupLock(key)) {
+            pendingTransactions.add(new PendingTransaction(transaction, key));
+         }
+      });
+      return pendingTransactions.isEmpty() ? Collections.emptyList() : pendingTransactions;
```
danberindei: I don't think the conditional is needed, the `pendingTransactions` list is already instantiated, so using `emptyList()` doesn't save us the allocation.

pruivo: but allows it to be garbage collected immediately.

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
danberindei: For future reference, I ran the JMH benchmark in the most voted response on
that page and the difference between streams and foreach was quite
significant:

N=10_000
Benchmark Mode Cnt Score Error Units
StreamVsVanilla.stream avgt 200 10.421 ± 0.179 ns/op
StreamVsVanilla.vanilla avgt 200 7.813 ± 0.081 ns/op

N=5
Benchmark Mode Cnt Score Error Units
StreamVsVanilla.stream avgt 200 29.490 ± 0.190 ns/op
StreamVsVanilla.vanilla avgt 200 12.885 ± 0.119 ns/op

N=1
Benchmark Mode Cnt Score Error Units
StreamVsVanilla.stream avgt 50 115.964 ± 3.237 ns/op
StreamVsVanilla.vanilla avgt 50 37.412 ± 0.712 ns/op

(N is the number of elements in the list, and the timing is per element)
Considering that Radim's SimpleCacheImpl does the entire put in 70ns
(caveat, his machine is probably a little faster than mine), I'd say it's
probably best to stay away from streams in our internals.

danberindei:
Integrated, thanks for your patience Pedro!

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
wburns: I can understand that yes.  Also when using iteration, if using an array list we should use for int loops instead as they are also 16% faster.

Although we need to be cognizant of what the alternatives are more than just saying 1 thing is bad imo.  For example iterating over the collection multiple times or creating a new collection could be worse then just using a stream.

core/src/main/java/org/infinispan/interceptors/locking/NonTransactionalLockingInterceptor.java
danberindei: Ah, back when I was trying those changes you didn't have the `lockAndRecord()` overload that took a collection.

core/src/main/java/org/infinispan/transaction/impl/AbstractCacheTransaction.java
danberindei: >
> I can understand that yes. Also when using iteration, if using an array
> list we should use for int loops instead as they are also 16% faster.
>

I also did some more tests with forEach() and for int, and on my i7
forEach() seems to handle lists with a 5 elements a little better:

N=5
Benchmark Mode Cnt Score Error Units
StreamVsVanilla.forEach avgt 50 11.059 ± 0.144 ns/op
StreamVsVanilla.forInt avgt 50 13.175 ± 0.104 ns/op
StreamVsVanilla.vanilla avgt 50 12.583 ± 0.053 ns/op
N=1
Benchmark Mode Cnt Score Error Units
StreamVsVanilla.forEach avgt 50 32.876 ± 0.165 ns/op
StreamVsVanilla.forInt avgt 50 32.924 ± 0.521 ns/op
StreamVsVanilla.vanilla avgt 50 35.940 ± 0.207 ns/op

 Although we need to be cognizant of what the alternatives are more than
> just saying 1 thing is bad imo. For example iterating over the collection
> multiple times or creating a new collection could be worse then just using
> a stream.
>

Indeed, the alternative is not always cheaper. But for our needs, I think
it's easy enough to do the same stuff streams do without allocating new
objects. Like streams collect a bunch of operations in a pipeline
(allocating new objects in the process) and execute them all with a single
collection iteration, we can also collect operations in method parameters
(using only the stack) and execute them with a single collection iteration.

PR #3596, opened 2015-07-21
Ispn 5619 Deprecate EntryRetriever with distributed streams

wburns:
This PR handles moving over all internal references of EntryRetriever to use new distributed streams instead.  
* Added segment completion listener support.  
* Added partition handling support for distributed streams.  
* Added support for injection of limited stream intermediate functions required for indexing.  
* Also fixed some bugs in distributed streams that were not covered in tests which now are due to conversion (specifically around using streams with compatibility or remote caches).

https://issues.jboss.org/browse/ISPN-5619


wburns:
This issue is not quite completed, I just wanted to get some test coverage going as it should be good now.

core/src/test/java/org/infinispan/notifications/cachelistener/CacheNotifierImplInitialTransferDistTest.java
```diff
@@ -738,57 +734,76 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
...
-      return retriever;
-   }
+         return mockSet;
+      }).when(mockChain).invoke(Mockito.any(InvocationContext.class),
+              (VisitableCommand) argThat(new IsInstanceOf(EntrySetCommand.class)));
+      TestingUtil.replaceComponent(cache, InterceptorChain.class, mockChain, true);
+      return chain;
+   }
+
+//   protected EntryRetriever waitUntilClosingIterator(final Cache<?, ?> cache, final CheckPoint checkPoint) {
```
wburns: Need to remove this still

core/src/main/java/org/infinispan/stream/impl/termop/object/MapIteratorOperation.java
```diff
@@ -2,6 +2,7 @@
 
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.container.entries.ImmortalCacheEntry;
+import org.infinispan.factories.ComponentRegistry;
```
wburns: Remove

core/src/main/java/org/infinispan/stream/impl/termop/primitive/ForEachDoubleOperation.java
```diff
@@ -2,6 +2,7 @@
 
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.container.entries.ImmortalCacheEntry;
+import org.infinispan.factories.ComponentRegistry;
```
wburns: Remove

core/src/main/java/org/infinispan/filter/CacheFilters.java
```diff
@@ -0,0 +1,202 @@
...
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+/**
+ * TODO:
```
wburns: Add comments

wburns:
Updated and rebased and ready for review.

query/src/main/java/org/infinispan/query/impl/massindex/IndexWorker.java
```diff
@@ -75,9 +74,11 @@ private Object extractValue(Object wrappedValue) {
    public Void call() throws Exception {
       preIndex();
       KeyValueFilter filter = getFilter();
-      try (CloseableIterator<CacheEntry<Object, String>> iterator = entryRetriever.retrieveEntries(filter, null, Util.asSet(Flag.CACHE_MODE_LOCAL), null)) {
+      try (Stream<CacheEntry<Object, Object>> stream = cache.getAdvancedCache().withFlags(Flag.CACHE_MODE_LOCAL)
+              .cacheEntrySet().stream()) {
```
gustavonalle: Hmm, what about ```parallelstream``` here?  :smile:

core/src/main/java/org/infinispan/iteration/impl/TrackingEntryIterableFromStream.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Stream;
+
+/**
+ * CloseableIterable that also tracks the iterators it spawns so it can properly close them when the close method is
+ * invoked.
+ *
+ * @author wburns
+ * @since 7.0
+ */
```
gustavonalle: 8.0

query/src/main/java/org/infinispan/query/impl/massindex/IndexWorker.java
wburns: We sure could.  I just didn't want to go changing the performance of this right away without being able to test if it was better or not.

wburns:
Also @gustavonalle just to let you know this doesn't include changes for REPL to notify segment completion.  I am still not sure what benefit this would provide without having lock striping.  Is there something I am missing there?

query/src/main/java/org/infinispan/query/impl/massindex/IndexWorker.java
gustavonalle: agreed, I will handle it separately 
https://issues.jboss.org/browse/ISPN-5622

gustavonalle:
The benefit, at least from the remote iterator POV, is to avoid keeping track of all the keys

wburns:
@gustavonalle Yes, sorry data container striping is what I meant.
But the problem is in REPL mode I can't complete any segments until I have fully iterated upon the entire data container.  At which point I will end up completing all of the segments at once, so there would be no time where you aren't holding onto all of the keys.  If a remote node dies it is still the same issue it doesn't know what keys has seen so it would have to start over again.  Unless there is some case you are thinking of?

gustavonalle:
@wburns I see, it's the same issue that the EntryRetriever has. I think I might have understood from our previous conversations that with Distributed Streams it'd be different. In this case, you're right, there's no point in sending notification for repl

query/src/main/java/org/infinispan/query/impl/massindex/IndexWorker.java
wburns: I wonder how fast or slow is the updateIndex method below?  Adding a parallel stream wouldn't benefit us much if this is not an extremely faster operation.  The other thought is you could refactor this code to use a forEach instead and then a parallel stream could have big benefit, but then again I hope the updateIndex wouldn't block or use IO else this isn't a good fit.

gustavonalle: ```updateIndex``` is a non-blocking operation, so it could benefit from more parallelism.

gustavonalle: https://developer.jboss.org/message/934848#934848

core/src/main/java/org/infinispan/iteration/impl/TrackingEntryIterableFromStream.java
wburns: Yeah I basically copied this from an existing 7.0 class as it does almost the same thing, heh.  Fixing and the javadoc comment.

gustavonalle:
pulling...

gustavonalle:
Integrated, thanks!

PR #3588, opened 2015-07-16
[ISPN-5568] - KeyAffinityService race condition on view change

baranowb:
https://issues.jboss.org/browse/WFLY-4942
https://issues.jboss.org/browse/ISPN-5568
https://bugzilla.redhat.com/show_bug.cgi?id=1233968

master: https://github.com/infinispan/infinispan/pull/3590
7.2.x: https://github.com/infinispan/infinispan/pull/3589
5.2.x: https://github.com/infinispan/infinispan/pull/3588

PR #3598, opened 2015-07-21
Updated ISPN-5568 KeyAffinityService race condition on view change for 5.2.x

danberindei:
https://issues.jboss.org/browse/ISPN-5568

Replaces https://github.com/infinispan/infinispan/pull/3588


tristantarrant:
Pushed

PR #3606, opened 2015-07-24
ISPN-2849 Don't keep threads blocked when waiting for locks to be rel…

pruivo:
…eased

https://issues.jboss.org/browse/ISPN-2849

another huge PR unfortunately :(


commons/src/main/java/org/infinispan/commons/util/InfinispanCollections.java
```diff
@@ -301,15 +302,29 @@ public int size() {
...
+   public static <T> void forEach(T[] array, Consumer<T> consumer) {
+      Objects.requireNonNull(array, "Array must be non-null.");
+      final int size = array.length;
+      if (size == 0) {
+         return;
+      } else if (size == 1) {
+         consumer.accept(array[0]);
+         return;
+      }
+      for (int i = 0; i < size; ++i) {
```
danberindei: I think only the for loop should be good enough - there's no allocation, and it will do the same `0 < size`, `1 < size` comparisons.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
```diff
@@ -27,15 +57,44 @@ public void handle(CacheRpcCommand command, Reply reply, DeliverOrder order) {
...
          final int commandTopologyId = extractCommandTopologyId(command);
-         final boolean waitTransactionalData = command.getCommandId() != StateRequestCommand.COMMAND_ID;
          final boolean onExecutorService = executeOnExecutorService(order, command);
-         handleRunnable(createDefaultRunnable(command, reply, commandTopologyId, waitTransactionalData, onExecutorService), onExecutorService);
+         final BlockingRunnable runnable;
+
+         switch (command.getCommandId()) {
+            case SingleRpcCommand.COMMAND_ID:
+               //lock ownership needs the first topology. we need to check it before invoking the getLockPromise()
+               if (isCommandSentBeforeFirstTopology(commandTopologyId)) {
```
danberindei: I think here you have the same problem of checking the topology of the command against the first topology potentially before the first topology was set (so it's `Integer.MAX_VALUE`).

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
```diff
@@ -45,4 +104,128 @@ protected boolean isTraceEnabled() {
...
+      final List<Object> keysToLock = new ArrayList<>(keys.size());
+      filterByPrimaryOwner(keys, keysToLock);
+      if (keysToLock.isEmpty()) {
+         return null;
+      }
+
+      final long timeoutMillis = command.hasZeroLockAcquisition() ? 0 : configuration.locking().lockAcquisitionTimeout();
+
+      if (keysToLock.size() == 1) {
+         return lockManager.lock(keysToLock.get(0), command.getLockOwner(), timeoutMillis, TimeUnit.MICROSECONDS);
```
danberindei: s/MICROSECONDS/MILLISECONDS/

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
```diff
@@ -27,15 +57,44 @@ public void handle(CacheRpcCommand command, Reply reply, DeliverOrder order) {
...
+                                                  getLockPromise((SingleRpcCommand) command));
+               break;
+            case MultipleRpcCommand.COMMAND_ID:
+               //lock ownership needs the first topology. we need to check it before invoking the getLockPromise()
+               if (isCommandSentBeforeFirstTopology(commandTopologyId)) {
+                  reply.reply(CacheNotFoundResponse.INSTANCE);
+                  return;
+               }
+               runnable = createLockAwareRunnable(command, reply, commandTopologyId, true, onExecutorService,
+                                                  getLockPromise((MultipleRpcCommand) command));
```
danberindei: I'm not sure about this, but I think it may be better to create a separate runnable for each command or it would introduce unnecessary delays. OTOH, since MultipleRpcCommands are only used by the replication queue, which has huge latencies anyway, it probably doesn't matter.

commons/src/main/java/org/infinispan/commons/util/InfinispanCollections.java
pruivo: you're right.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
pruivo: there is a problem but only for the state transfer/topology change commands and they are never wrap in ```SingleRpcCommand```. So it is safe :)

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
pruivo: lool ops :D

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
pruivo: IMO, I don't see any advantages of using a replication queue. Everything does some kind of batching (JGroups, TCP, OS) and it only complicates the code a lot.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
danberindei: I think it may be a problem for regular write commands as well. At least in theory, the coordinator can install the new topology that includes the joiner, send a write to the joiner (which is ignored, because the joiner hasn't received the join response yet), see the `CacheNotFoundResponse` as a sign that the joiner is already shutting down, and block waiting for a new topology that never comes.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
danberindei: Well, nobody argued against your proposal to remove the replication queue on the dev list, so if you issue a PR to remove it and mark the configuration deprecated I'm going to integrate it!

wburns: +1000

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
pruivo: I can do something similar to https://github.com/infinispan/infinispan/pull/3606/files#diff-376c88e74f4d66c0dff57dc91b234d89R356

when isReady() is invoked, I check if topology is received, then the first topology and in the end, acquire the check LockPromise. do you have another idea?

danberindei: TBH I don't really like the approach of keeping the knowledge of all the stages in a single class , I was thinking more of a chain approach, like with `CompletableFuture`. I realize `CompletableFuture` is not a perfect fit here, because you also need a way to propagate cancellation, but still I think it would be good to have more uniformity between the pending lock promises, the real lock promises, and the eventual topology promises.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
```diff
@@ -27,15 +57,44 @@ public void handle(CacheRpcCommand command, Reply reply, DeliverOrder order) {
...
+         }
+         handleRunnable(runnable, onExecutorService);
       } catch (Throwable throwable) {
          reply.reply(exceptionHandlingCommand(command, throwable));
       }
    }
 
    @Override
+   public void onEvent(LockState state) {
+      remoteCommandsExecutor.checkForReadyTasks();
```
danberindei: I'd also like it if we didn't have to call `checkForReadyTasks()` as often, because it goes through all the commands waiting for a lock. If the lock promise had a `runAsync` method, you could run the command directly on the remote thread pool, without having to go through the controller thread.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
```diff
@@ -27,15 +57,44 @@ public void handle(CacheRpcCommand command, Reply reply, DeliverOrder order) {
...
+
+         switch (command.getCommandId()) {
+            case SingleRpcCommand.COMMAND_ID:
+               //lock ownership needs the first topology. we need to check it before invoking the getLockPromise()
+               if (isCommandSentBeforeFirstTopology(commandTopologyId)) {
+                  reply.reply(CacheNotFoundResponse.INSTANCE);
+                  return;
+               }
+               runnable = createLockAwareRunnable(command, reply, commandTopologyId, true, onExecutorService,
+                                                  getLockPromise((SingleRpcCommand) command));
```
danberindei: Should we create a lock promise for commands that are not going to be executed on the remote thread pool (`ClusteredGetCommand`, async commands)?

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderTxPerCacheInboundInvocationHandler.java
```diff
@@ -42,4 +123,288 @@ protected boolean isTraceEnabled() {
...
+      final long timeoutMillis = command.hasZeroLockAcquisition() ? 0 : configuration.locking().lockAcquisitionTimeout();
+      final RemoteTxInvocationContext context = command.createContext();
+      if (context == null) {
+         return null;
+      }
+      keysToLock.forEach(context::addLockedKey);
+      if (keysToLock.size() == 1) {
+         return new SingleKeyReadyAction(context, timeoutMillis, keysToLock.get(0));
+      }
+      return new MultipleKeysReadyAction(context, timeoutMillis, keysToLock);
```
danberindei: `createReadyActionForLockControlCommand()` and `createReadyActionForPrepareCommand() look very similar.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
pruivo: IMO, I prefer to have it in a single class. this is an internal detail and we don't need to allocate N classes to do the check. then, as you said, we have to propagate the cancellation.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
pruivo: ```checkForReadyTasks()``` only changes a volatile integer. I think it is fine.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
pruivo: the lock promise is only for commands that acquire locks.

the cluster get does not acquire. in the locking interceptor, if get is invoked with force locking, it creates a ```LockControlCommand```. 

I don't get what you mean by async commands...

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderTxPerCacheInboundInvocationHandler.java
pruivo: probably with all the refactor I have missed that. I'll double check and if it is the case I'll remove one them :)

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
danberindei: You still need a promise object for each phase either way, so I don't think the memory usage would be much different. And (I hope) we wouldn't need two separate `ComposedLockPromise` classes, and yet one more `CompositeReadyAction`.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
danberindei: By async commands I meant the write commands and the prepare command in DIST_ASYNC/REPL_ASYNC, which are executed on JGroups' regular thread pool.
You're right that the remote get doesn't acquire the locks itself, but it does block the OOB thread if `acquireRemoteLock == true`, so maybe it should be covered by this as well?

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
danberindei: I was more worried about the controller thread having to wake up every time and go through all the blocked commands.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderPerCacheInboundInvocationHandler.java
pruivo: IMO, ```acquireRemoteLock``` can be removed. it is only set when the remote get is done for a write command but the write command should have acquired the lock before (it invoke the ```acquireRemoteIfNeeded()``` before ```invokeNext()```)

pruivo: about the async commands, I can't move them to the executor service... creating or not the lock promise is the same.

core/src/main/java/org/infinispan/util/StripedHashFunction.java
```diff
@@ -0,0 +1,66 @@
...
+         tmpNumSegments <<= 1;
+      }
+      lockSegmentShift = 32 - tempLockSegShift;
+      lockSegmentMask = tmpNumSegments - 1;
+      numSegments = tmpNumSegments;
+   }
+
+   /**
+    * Returns a hash code for non-null Object x. Uses the same hash code spreader as most other java.util hash tables,
+    * except that this uses the string representation of the object passed in.
```
danberindei: I think this comment needs an update...

core/src/main/java/org/infinispan/util/concurrent/locks/DeadlockDetectingLockManager.java
```diff
@@ -1,42 +1,46 @@
...
+ * </p>
+ * A thread calling {@link #lock(Object, Object, long, TimeUnit)} or {@link #lockAll(Collection, Object, long, TimeUnit)}
  * would run the deadlock detection algorithm only if all of the following take place:
  * - the call is made in the scope of a transaction (either locally originated or remotely originated)
  * - it cannot acquire lock on the given key and the lock owner is another transaction
  * - when comparing coin toss, this thread would loose against the other one - so it's always the potential loser that runs DLD.
- * If deadlock is detected then {@link LockManager#lockAndRecord(Object, org.infinispan.context.InvocationContext, long)} would throw an
- * {@link org.infinispan.util.concurrent.locks.DeadlockDetectedException}. This is subsequently handled in
- * in the interceptor chain - locks owned by this tx are released.
+ * If deadlock is detected then a{@link org.infinispan.util.concurrent.locks.DeadlockDetectedException}.
```
danberindei: Sentence seems incomplete.

core/src/main/java/org/infinispan/util/concurrent/locks/DeadlockDetectingLockManager.java
```diff
@@ -46,77 +50,62 @@
...
+            log.tracef("We (%s) win against the other (%s) transaction, so no point running rest of DLD", thisTx, lockOwnerTx);
          return false;
       }
       //do we have lock on what other tx intends to acquire?
-      return ownsLocalIntention(thisTx, lockOwnerTx.getLockIntention()) || ownsRemoteIntention(lockOwnerTx, thisTx, key) || isSameKeyDeadlock(key, thisTx, lockOwnerTx);
+      return ownsAnyLocalIntention(thisTx, lockOwnerTx) || ownsRemoteIntention(lockOwnerTx, thisTx) || isSameKeyDeadlock(thisTx, lockOwnerTx);
    }
 
-   private boolean isSameKeyDeadlock(Object key, DldGlobalTransaction thisTx, DldGlobalTransaction lockOwnerTx) {
+   private boolean isSameKeyDeadlock(DldGlobalTransaction thisTx, DldGlobalTransaction lockOwnerTx) {
```
danberindei: I think this method can be removed, it was only necessary when we were acquiring the lock on the originator/backup owners.

danberindei: I've created https://issues.jboss.org/browse/ISPN-5654

core/src/main/java/org/infinispan/util/concurrent/locks/KeyAwareLockListener.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * A listener for {@link KeyAwareLockPromise}.
+ * <p>
+ * This event contains the key that the lock owner is try to acquire.
+ *
+ * @author Pedro Ruivo
+ * @since 8.0
+ */
+public interface KeyAwareLockListener {
```
danberindei: Is this really used?

core/src/main/java/org/infinispan/util/concurrent/locks/LockContainer.java
```diff
@@ -0,0 +1,51 @@
...
+    * @param key the key to lock.
+    * @return the lock for a specific object to be acquired. If the lock does not exists, it is created.
+    */
+   ExtendedLockPromise acquire(Object key, Object lockOwner, long time, TimeUnit timeUnit);
+
+   /**
+    * @param key the key to lock.
+    * @return the lock for a specific object. If the lock does not exists, it return {@code null}.
+    */
+   InfinispanLock getLock(Object key);
```
danberindei: I don't think it's a good idea to return an impl class from a non-impl interface method. Maybe we can remove this method, and replace it with `getOwner(key)`?

core/src/main/java/org/infinispan/util/concurrent/locks/LockPromise.java
```diff
@@ -0,0 +1,54 @@
...
+
+   /**
+    * It locks the key (or keys) associated to this promise.
+    * <p>
+    * This method will block until the lock is available or the timeout is expired.
+    *
+    * @throws InterruptedException if the current thread is interrupted while acquiring the lock
+    * @throws TimeoutException     if we are unable to acquire the lock after a specified timeout.
+    */
+   void lock() throws InterruptedException, TimeoutException;
```
danberindei: I think `get()` would be a better name - the locking process already started when the promise was created, and the key may already be locked before the first call to `lock()`.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
```diff
@@ -0,0 +1,378 @@
...
+         //although will have the cost of creating an iterator, at least, we don't need to enter the synchronized section.
+         return lock(uniqueKeys.iterator().next(), lockOwner, time, unit);
+      }
+
+      if (trace) {
+         log.tracef("Lock all keys=%s for owner=%s. timeout=%s (%s)", uniqueKeys, lockOwner, time, unit);
+      }
+
+      final CompositeLockPromise compositeLockPromise = new CompositeLockPromise(uniqueKeys.size());
+      synchronized (this) {
```
danberindei: A comment on why we need the synchronization would be nice.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
```diff
@@ -0,0 +1,378 @@
...
+   public int getNumberOfLocksAvailable() {
+      return lockContainer.size() - lockContainer.getNumLocksHeld();
+   }
+
+   @Override
+   public InfinispanLock getLock(Object key) {
+      return lockContainer.getLock(key);
+   }
+
+   private void scheduleLockPromise(LockPromise promise, long time, TimeUnit unit) {
```
danberindei: `scheduleLockTimeoutTask()`?

core/src/main/java/org/infinispan/util/concurrent/locks/KeyAwareLockListener.java
pruivo: yes. extended-stats modules needs to have some association between keys and locks.

core/src/main/java/org/infinispan/util/concurrent/locks/LockContainer.java
pruivo: yes I think it can be removed. it is only needed for LockContainerHashingTest

core/src/main/java/org/infinispan/util/concurrent/locks/LockPromise.java
pruivo: not really. the lock is only acquired when this method is invoked, otherwise it is in available state the is cancelable. also, get() looks like you are returning something, that is not the case.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
pruivo: ack. Basically, it is to avoid internal deadlocks.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
pruivo: as you wish :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
```diff
@@ -0,0 +1,378 @@
...
+      if (!promise.isAvailable() && time > 0 && scheduler != null) {
+         final ScheduledFuture<?> future = scheduler.schedule(promise::isAvailable, time, unit);
```
danberindei: I'd feel safer if the task would time out the lock regardless of whether the scheduler executes the task right after or right before the timeout actually expires.

core/src/main/java/org/infinispan/util/concurrent/locks/LockContainer.java
pruivo: actually, it can. I need it in extended stats. I'll move the interface to impl package.

core/src/main/java/org/infinispan/util/concurrent/locks/DeadlockDetectingLockManager.java
pruivo: noted :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
danberindei: Also, see how I've avoided allocating new lambdas for the timeout task and cancellation here: https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/remoting/transport/jgroups/SingleResponseFuture.java

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,555 @@
...
+/**
+ * A special lock for Infinispan cache.
+ * <p>
+ * The main different with the traditional {@link java.util.concurrent.locks.Lock} is allowing to use any object as lock
+ * owner. It is possible to use a {@link Thread} as lock owner that makes similar to {@link
+ * java.util.concurrent.locks.Lock}.
+ * <p>
+ * In addition, it has an asynchronous interface. {@link #acquire(Object, long, TimeUnit)}  will not acquire the lock
+ * immediately (except if it is free) but will return a {@link ExtendedLockPromise}. This promise allow to test if the
+ * lock is acquired asynchronously and cancel the lock acquisition, without any blocking.
```
danberindei: I'd say the asynchronous acquisition is the main feature now :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
```diff
@@ -0,0 +1,378 @@
...
+         }
+         return KeyAwareLockPromise.NO_OP;
+      } else if (keys.size() == 1) {
+         //although will have the cost of creating an iterator, at least, we don't need to enter the synchronized section.
+         return lock(keys.iterator().next(), lockOwner, time, unit);
+      }
+
+      final Set<Object> uniqueKeys = filterDistinctKeys(keys);
+
+      if (uniqueKeys.isEmpty()) {
```
danberindei: There's another `isEmpty()` check above.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,555 @@
...
+ * @since 8.0
+ */
+public class InfinispanLock {
+
+   private static final Log log = LogFactory.getLog(InfinispanLock.class);
+   private static final boolean trace = log.isTraceEnabled();
+
+   private final Queue<LockPlaceHolder> pendingRequest;
+   private final ConcurrentMap<Object, LockPlaceHolder> lockOwners;
+   private final AtomicReferenceFieldUpdater<InfinispanLock, LockPlaceHolder> fieldUpdater;
```
danberindei: Why not static?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,555 @@
...
+   private final Runnable releaseRunnable;
+   private volatile TimeService timeService;
```
danberindei: I don't think it's worth making this volatile just for testing, tests should change the lock manager's time service before they start any operation.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: because I didn't notice it can be static. I'll change all of them :)

wburns: :) yeah that is the main benefit of it, since then you don't have to have an AtomicReference instance wrapping every single one.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,555 @@
...
+    * @throws NullPointerException if {@code lockOwner} is {@code null}.
+    */
+   public void release(Object lockOwner) {
+      Objects.requireNonNull(lockOwner, "Lock Owner should be non-null");
+
+      if (trace) {
+         log.tracef("Release lock for %s.", lockOwner);
+      }
+
+      LockPlaceHolder wantToRelease = lockOwners.get(lockOwner);
```
danberindei: I think you could just check the `lockOwner` param against `current.lockOwner` and ignore the release/throw an exception if they don't match.
Also, it's a bid odd that InfinispanLock is semi-reentrant: you can `acquire` a lock multiple times for the same owner, but you can only release it once. I think if you eliminate the reentrancy completely, you can remove the `lockOwners` map altogether.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,555 @@
...
+
+   /**
+    * It tests if the lock has the lock owner.
+    * <p>
+    * It return {@code true} if the lock owner is the current lock owner or it in the queue.
+    *
+    * @param lockOwner the lock owner to test.
+    * @return {@code true} if it contains the lock owner.
+    */
+   public boolean containsLockOwner(Object lockOwner) {
```
danberindei: Unused.

pruivo: use here: https://github.com/infinispan/infinispan/pull/3606/files#diff-9ff4035e4fd778413c872329ba47b67dR179

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: but I need this kind of semi-reentrant. I need to try acquire in inbound handler to check when the lock is available and then in the interceptors to really acquire the lock. and after that, I only want to release it once.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,555 @@
...
+
+   private static final Log log = LogFactory.getLog(InfinispanLock.class);
+   private static final boolean trace = log.isTraceEnabled();
+
+   private final Queue<LockPlaceHolder> pendingRequest;
+   private final ConcurrentMap<Object, LockPlaceHolder> lockOwners;
+   private final AtomicReferenceFieldUpdater<InfinispanLock, LockPlaceHolder> fieldUpdater;
+   private final Runnable releaseRunnable;
+   private volatile TimeService timeService;
+   private volatile LockPlaceHolder current;
```
danberindei: I'm not sure having separate fences for `pendingRequest`, `lockOwners`, and `current` helps, maybe `synchronized (this)` would be simpler. OTOH, I think you could replace all 3 fields with a single list for the requestors, like `AbstractQueuedSynchronizer` does, with the head replacing `current`, and with a linear search when you need to search for a particular owner in the list.

danberindei: Oops, `AbstractQueuedSynchronizer` inherits an `exclusiveOwnerThread` field from `AbstractOwnableSynchronizer`... perhaps in their case it's not worth creating the queue unless there is actual contention, but you create the queue all the time anyway.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,555 @@
...
+      } while (true);
+   }
+
+   private LockPlaceHolder createLockInfo(Object lockOwner, long time, TimeUnit timeUnit) {
+      return new LockPlaceHolder(lockOwner, timeService.expectedEndTime(time, timeUnit));
+   }
+
+   private class LockPlaceHolder implements ExtendedLockPromise {
+
+      private final AtomicReferenceFieldUpdater<LockPlaceHolder, LockState> stateUpdater;
```
danberindei: static?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,555 @@
...
+
+      private final Object lockOwner;
+      private final long timeout;
+      private final AtomicBoolean cleanup;
```
danberindei: Replace with `volatile int` and use an `AtomicIntegerFieldUpdater`?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,555 @@
...
+
+   private LockPlaceHolder createLockInfo(Object lockOwner, long time, TimeUnit timeUnit) {
+      return new LockPlaceHolder(lockOwner, timeService.expectedEndTime(time, timeUnit));
+   }
+
+   private class LockPlaceHolder implements ExtendedLockPromise {
+
+      private final AtomicReferenceFieldUpdater<LockPlaceHolder, LockState> stateUpdater;
+
+      private final Object lockOwner;
```
danberindei: I'm not sure I'd call it `lockOwner`, because it doesn't own the lock yet, it only owns this placeholder...

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: the lock owner are not threads anymore. I can implement the AbstractQueuedSynchronizer with any queues if you want. But in this case, you can have multiple threads invoking lock() with the same lock owner. And you have to make sure that only one will create the LockPlaceHolder. 

the function of the map is to avoid a linear search in the queue. Also, the queue does not have a addIfAbsent. Also, synchronize and iterate all the time would be much more expensive.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: hmmm. what is the advantage of using the volatile int over an AtomicBoolean?

wburns: I am not a big fan of that for something that is really just a boolean.  However I wonder if this can be done differently.  It seems this boolean is only to protect running the release runnable.  Is it instead possible to have the remove return the previous value or a boolean and then optionally run the runnable only if it removed something?  Then you can remove this boolean completely.  I haven't had a chance to read over all of the classes to see if that makes sense or not, let me know if it doesn't

Something like:
```
 private void cleanup() {
   if (remove(lockOwner) != null) {
      triggerReleased();
   }
}
```

pruivo:
Updated!
Main change: in *PerCacheInboundInvocationHandler, isReady() is made by invoked each action (should it be called step?). when the action is successfully completed, the next one is invoked. if it is canceled, it stops where it was.

core/src/main/java/org/infinispan/remoting/inboundhandler/action/PendingTxAction.java
```diff
@@ -0,0 +1,106 @@
...
+      if (context == null) {
+         //cancel everything. nobody was able to update the state from checking, so no need to check the CAS
+         cas(InternalState.CHECKING, InternalState.CANCELED);
+         return ActionStatus.CANCELED;
+      }
+      final long timeout = state.getTimeout();
+      final List<Object> keysToLock = getAndUpdateFilteredKeys(state);
+
+      if (keysToLock.isEmpty()) {
+         //nothing to do. nobody was able to update the state from checking, so no need to check the CAS
```
danberindei: Nitpicking: nobody else

core/src/main/java/org/infinispan/util/concurrent/locks/LockPromise.java
danberindei: Hmmm, so when a command becomes ready it doesn't have the locks yet, and it's possible that if will block if another command snoops in and acquires the lock? I think it would be better if a command only becomes ready after it has acquired all the locks it needs.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
```diff
@@ -0,0 +1,395 @@
...
+      Objects.requireNonNull(key, "Key must be non null");
+      Objects.requireNonNull(lockOwner, "Lock owner must be non null");
+      Objects.requireNonNull(unit, "Time unit must be non null");
+
+      if (trace) {
+         log.tracef("Lock key=%s for owner=%s. timeout=%s (%s)", key, lockOwner, time, unit);
+      }
+
+      ExtendedLockPromise promise = lockContainer.acquire(key, lockOwner, time, unit);
+      return new KeyAwareExtendedLockPromise(promise, key, unit.toMillis(time)).scheduleLockTimeoutTask(scheduler);
```
danberindei: Couldn't `LockPlaceholder` be key-aware itself?

pruivo: I think about it but there is one use case where I couldn't fit it. 
In extended-stats module, if you have a tx acquiring k1, k2 and striped locking configured and if k1 and k2 will hash on the same InfinispanLock, I would need a way to trigger 2 notifications with 2 different keys.

core/src/main/java/org/infinispan/util/concurrent/locks/LockPromise.java
pruivo: no, it is not possible. 
when in available state, the lock is only available for that particular lock owner. and from that state it can cancel() or lock() via LockPromise interface. 
so, lock stealing is not possible :)

danberindei: Ok, but then why no go directly to the `ACQUIRED` state, and only allow unlocking afterwards?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
```diff
@@ -0,0 +1,395 @@
...
+      @Override
+      public void setDeadlockChecker(DeadlockChecker deadlockChecker) {
+         for (LockPromise lockPromise : lockPromiseList) {
+            lockPromise.setDeadlockChecker(deadlockChecker);
+         }
+      }
+
+      @Override
+      public void onEvent(LockState state) {
+         if (state != LockState.AVAILABLE && UPDATER.compareAndSet(this, LockState.AVAILABLE, state)) {
```
danberindei: Won't this change the state of the composite promise to `AVAILABLE` before all the sub-promises are `AVAILABLE`?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
danberindei: Maybe striped locking could use a `LockPlaceHolder` subclass with a list of keys instead of a single key?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,537 @@
...
+    * Tests purpose only!
+    */
+   public void setTimeService(TimeService timeService) {
+      if (timeService != null) {
+         this.timeService = timeService;
+      }
+   }
+
+   /**
+    * It tries to acquit this lock.
```
danberindei: acquire :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
```diff
@@ -0,0 +1,395 @@
...
+      if (trace) {
+         log.tracef("Lock all keys=%s for owner=%s. timeout=%s (%s)", uniqueKeys, lockOwner, time, unit);
+      }
+
+      final CompositeLockPromise compositeLockPromise = new CompositeLockPromise(uniqueKeys.size());
+      //needed to avoid internal deadlock when 2 or more lock owner invokes this method with the same keys.
+      //ordering will not solve the problem since acquire() is non-blocking and each lock owner can iterate faster/slower than the other.
+      synchronized (this) {
+         for (Object key : uniqueKeys) {
+            compositeLockPromise.addLock(new KeyAwareExtendedLockPromise(lockContainer.acquire(key, lockOwner, time, unit), key, unit.toMillis(time)));
```
danberindei: Not sure if it's worth worrying about, but an OOME here could leak the lock.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
```diff
@@ -0,0 +1,395 @@
...
+
+      @Override
+      public void addListener(KeyAwareLockListener listener) {
+         for (KeyAwareExtendedLockPromise lockPromise : lockPromiseList) {
+            lockPromise.addListener(listener);
+         }
+      }
+
+      @Override
+      public void run() {
```
danberindei: `ScheduledExecutorFactory` will create a `Callable` wrapper for `Runnable` tasks, so in this case it's better to implement `Callable`.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
```diff
@@ -0,0 +1,395 @@
...
+            } catch (RuntimeException e) {
+               if (exception == null) {
+                  exception = e;
+                  exceptionType = ExceptionType.RUNTIME;
+               }
+            }
+         }
+         if (exception != null) {
+            switch (exceptionType) {
+               case INTERRUPTED:
```
danberindei: I think `exception instanceof InterruptedException` should also work... But I think maybe we want some kind of exception hierarchy, e.g. `InterruptedException` trumps everything else, in which case a separate variable for each exception type would be better.

Also, I think we should try to preserve all the exceptions when there are more of them, maybe with `addSuppressed()`.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
danberindei: I was thinking you could save the lock promise in the command/context, and then you wouldn't need to lock again. But I guess I'd better shut up and work on the sequential interceptors instead, because that would help :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
danberindei: Looks like I had the extended stats module disabled in IntelliJ, that's why it kept coloring all these methods gray...

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,537 @@
...
+    */
+   public boolean containsLockOwner(Object lockOwner) {
+      return lockOwners.containsKey(lockOwner);
+   }
+
+   private void onCanceled(LockPlaceHolder canceled) {
+      if (trace) {
+         log.tracef("Release lock for %s. It was canceled.", canceled.owner);
+      }
+      LockPlaceHolder wantToRelease = lockOwners.get(canceled.owner);
```
danberindei: Could `wantToRelease` be different from `canceled`? And it's not really clear who removes the owner from the `lockOwners` map in this case...

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,537 @@
...
+            log.tracef("%s not found!", canceled.owner);
+         }
+         //nothing to release
+         return;
+      }
+      LockPlaceHolder currentLocked = current;
+      if (currentLocked == wantToRelease) {
+         if (!casRelease(currentLocked)) {
+            if (trace) {
+               log.tracef("Releasing lock for %s. It is the current lock owner but another thread changed it.", canceled.owner);
```
danberindei: This looks a bit scary... if another thread released our lock, then we might have updated the data container/stores without holding the key lock?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,537 @@
...
+         }
+         if (casAcquire(nextPending)) {
+            //we set the current lock owner, so we must remove it from the queue
+            pendingRequest.remove(nextPending);
+            if (nextPending.setAvailable()) {
+               if (trace) {
+                  log.tracef("%s successfully acquired the lock.", nextPending);
+               }
+               //successfully acquired
+               deadlockCheck();
```
danberindei: Could we skip this? If we only do the deadlock checks from the timer task, then we can remove the `deadlockChecker` field and pass it as a parameter to `checkDeadlock`/`deadlockCheck` instead.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,537 @@
...
+      }
+
+      private void invoke(LockListener invoker) {
+         LockState state = lockState;
+         switch (state) {
+            case WAITING:
+               throw new IllegalStateException("WAITING is not a valid state to invoke the listener");
+            case ACQUIRED:
+            case RELEASED:
+               invoker.onEvent(LockState.AVAILABLE);
```
danberindei: This looks weird to me, why shouldn't the listener receive the actual state?

core/src/main/java/org/infinispan/util/concurrent/locks/LockPromise.java
pruivo: because I need 2 states: 
the ```AVAILABLE``` is cancelable (the lock is for that lock owner)
the ```ACQUIRED``` is not (means the lock owner has already entered in the critical section)

danberindei: If another owner cannot enter the `AVAILABLE` state at the same time, I think that makes it part of the critical section as well. What's the difference between cancelling a lock and releasing it?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: it is something to think about it in the future... of course, this will add another field into the context

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
pruivo: no. the ```lockState``` is only used to have information to trigger the listeners. 
and this method is invoked by all the locks in the list when they change state. so, if the composite receives a timeout or deadlock event, it will cancel all the locks.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/PerKeyLockContainer.java
```diff
@@ -0,0 +1,101 @@
...
+   public void inject(TimeService timeService) {
+      this.timeService = timeService;
+      for (InfinispanLock lock : lockMap.values()) {
+         lock.setTimeService(timeService);
+      }
+   }
+
+   @Override
+   public ExtendedLockPromise acquire(Object key, Object lockOwner, long time, TimeUnit timeUnit) {
+      AtomicReference<ExtendedLockPromise> reference = new AtomicReference<>();
```
danberindei: `ByRef` would also work here.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
pruivo: updated

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
pruivo: true. well, can I assume if OOME is thrown, the system is screwed? :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
danberindei: `AtomicBoolean` allocates a new object, that's the only difference :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
pruivo: my idea was only to throw the first failure.
I still need to invoke lock() in all the list.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: no they are always the same. 
the clear from the map occurs when lock() is invoked (and it throws timeout or deadlock). When the exception is thrown, I'm not sure if release() may never be invoked so I perform the cleanup. 
the second scenario is when the release is invoked (normal scenario, lock is acquired and then released)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: I'm 99.9% sure that it never happens (well, we can have a rollback be processed at the same time in different threads by different sources, e.g. the originator sends the rollback and a topology change creates another)
Also, nothing prevents you from writing a test that create N threads to invoke release :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: yes. I'll remove it.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: Because I don't want to expose the acquired/released state.
The goal for the listener is to notify when the lock is available. I.E. the lock() method is non-blocking and we only 3 cases for it. 1) when it is available, 2) when it timeouts, 3) when it deadlocks. All the other state are irrelevant.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
pruivo: the LockPlaceHolder is internal to InfinispanLock.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
danberindei: Then I guess you don't need the `lockOwners.get()` call here...

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
danberindei: The originator won't retry until it receives a response from all the owners, so a topology change can't cause this. A timeout during prepare/commit may end up with the prepare/commit and the rollback running at the same time, but `TransactionSynchronizerInterceptor` will block the rollback until the prepare finishes. Unless you can think of another way this could happen, I would like to throw an exception here.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
danberindei: I'm ok with merging AVAILABLE and ACQUIRED completely, not just for the listener, but I'm not so sure about RELEASED. It seems to me like that one should be an error, like WAITING.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,537 @@
...
+      }
+
+      private boolean setReleased() {
+         do {
+            LockState state = lockState;
+            switch (state) {
+               case WAITING:
+               case AVAILABLE:
+               case ACQUIRED:
+                  if (casState(state, LockState.RELEASED)) {
```
danberindei: Is it ok to go from WAITING straight to RELEASED?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -0,0 +1,537 @@
...
+    * @return {@code true} if this lock has no lock owner pending.
+    */
+   public boolean isEmpty() {
+      return isFree() && lockOwners.isEmpty();
+   }
+
+   /**
+    * @return {@code true} if the lock is not acquired and no lock owners are queued.
+    */
+   public boolean isFree() {
```
danberindei: I think `isFree()` and `isEmpty()` could both be replaced with `!isLocked()`, because it shouldn't be possible to have queued requests without a current owner.

core/src/test/java/org/infinispan/lock/LockContainerHashingTest.java
```diff
@@ -11,25 +11,26 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.locks.Lock;
 
 @Test(groups = "unit", testName = "lock.LockContainerHashingTest")
 public class LockContainerHashingTest extends AbstractInfinispanTest {
-   private AbstractStripedLockContainer<?> stripedLock;
+   private StripedLockContainer stripedLock;
```
danberindei: Since you've extracted the hashing functionality in the `StripedHashFunction` class, I think you could test it here directly.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: I don't see why not.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/PerKeyLockContainer.java
pruivo: updated.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: not really. the listener is triggered sequentially. if you have slow listener, it is possible the lock state would evolve from waiting => acquired => released.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: removed.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: yes you can. there is a little time interval in which the current owner is null between the casRelease and casAcquire.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: I don't like the idea of throwing an exception here. this is invoked in loops and it will destroy the iteration completely.

core/src/main/java/org/infinispan/util/concurrent/locks/LockPromise.java
pruivo: the source. the release is invoked by the owner. the cancel is invoked when some condition are detected.

core/src/test/java/org/infinispan/util/DeadlockDetectingLockManagerTest.java
```diff
@@ -31,77 +30,72 @@
...
+      try {
+         lockManager.lock("k", new DldGlobalTransaction(), 1000, TimeUnit.MILLISECONDS).lock();
+         fail("TimeoutException expected!");
+      } catch (TimeoutException e) {
+         //expected
+      }
 
-      assert lockManager.lockAndRecord("k", localTxContext, config.locking().lockAcquisitionTimeout());
-      assert lockManager.getOverlapWithNotDeadlockAwareLockOwners() == 1;
+      AssertJUnit.assertTrue(lockManager.getOverlapWithNotDeadlockAwareLockOwners() > 0); //it's not deterministic. can be 1 or more.
```
danberindei: Hmmm, I think this JMX attribute became kind of useless the moment we stopped allowing the mix of tx and non-tx operations in the same cache, and now even more so because the number of deadlock checks isn't deterministic. I'd deprecate the attribute and remove this check.

core/src/test/java/org/infinispan/util/DeadlockDetectingLockManagerTest.java
```diff
@@ -31,77 +30,72 @@
...
          //expected
       }
-      assertEquals(1l,lockManager.getDetectedLocalDeadlocks());
+      AssertJUnit.assertEquals(1, lockManager.getDetectedLocalDeadlocks());
+   }
+
+   private Configuration createConfiguration() {
+      Configuration configuration = mock(Configuration.class, RETURNS_DEEP_STUBS);
+      when(configuration.deadlockDetection().spinDuration()).thenReturn(SPIN_DURATION);
+      when(configuration.jmxStatistics().enabled()).thenReturn(true);
```
danberindei: Why not just create a real `Configuration`?

core/src/test/java/org/infinispan/util/DeadlockDetectingLockManagerTest.java
```diff
@@ -31,77 +30,72 @@
...
-            return ddgt;
-         }
-      };
+   private LockContainer createLockContainer() {
+      PerKeyLockContainer lockContainer = new PerKeyLockContainer(32, AnyEquivalence.getInstance());
+      lockContainer.inject(TIME_SERVICE);
+      return lockContainer;
    }
 
    public static class DeadlockDetectingLockManagerMock extends DeadlockDetectingLockManager {
```
danberindei: Using a mock for the class that's being tested looks odd, I'm wondering if we could remove it.

I tried to remove it myself, but then the real DeadlockDetectingLockManager gave me a `TimeoutException` instead of a `DeadlockDetectedException` in `testLocalDeadlock()`.

core/src/main/java/org/infinispan/util/concurrent/locks/LockPromise.java
danberindei: I don't think the source matters that much if they do the same thing (i.e. assume the owner is the `current` owner, replace it with the first owner from the `pendingRequest` queue).

pruivo: imagine the following:
```
LockPromise = infinispanLock.acquire() //state == available
LockPromise::cancel() //here the lock promise will be canceled
LockPromise.lock() //state == acquired
LockPromise::cancel()  //here the cancel should be ignored
```
How can you distinguish both cancel with a single state=

core/src/test/java/org/infinispan/util/DeadlockDetectingLockManagerTest.java
pruivo: well, the non deterministic in the test is caused because the deadlock check is done asynchronously in the schedule executor. 

if you really want, I can remove the attribute and the test.

core/src/test/java/org/infinispan/util/DeadlockDetectingLockManagerTest.java
pruivo: I dont see the need to create a full configuration because of 2 parameters.

core/src/test/java/org/infinispan/util/DeadlockDetectingLockManagerTest.java
pruivo: maybe we can improve the test when the new deadlock algorithm is implemented :)

core/src/test/java/org/infinispan/lock/LockContainerHashingTest.java
pruivo: agree.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
danberindei: Yeah, I forgot about that, adding the hooks to create the right subclass would probably complicate things too much. 

I'm not ready to give up on this yet, but after looking a little more at the extended-statistics LockManagers, I think it may be best to just forget about individual keys and record the same waiting time for all of them. It's not like waiting time was very accurate before either, if each tx locked the same keys it would average the waiting time for the first key with 0 for the others. WDYT?

core/src/main/java/org/infinispan/util/concurrent/locks/LockPromise.java
danberindei: For me, the question is, do you really need to distinguish between the two?

I tried merging the two states and I only got one test failure in core: https://github.com/danberindei/infinispan/commit/f68f387a83a75376155b8f920ab6dd031f5225da

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
danberindei: In theory it could recover, but yeah, that's not likely.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
danberindei: I'd say if we get an `InterruptedException` we need to throw that, and/or re-set the thread's interruption status, regardless of any other exceptions. I had in mind something like this:

         InterruptedException interruptedException = null;
         TimeoutException timeoutException = null;
         DeadlockDetectedException deadlockException = null;
         RuntimeException runtimeException = null;
         for (ExtendedLockPromise lockPromise : lockPromiseList) {
            try {
               //we still need to invoke lock in all the locks.
               lockPromise.lock();
               interruptedException = e;
            } catch (TimeoutException e) {
               timeoutException = e;
            } catch (DeadlockDetectedException e) {
               deadlockException = e;
            } catch (RuntimeException e) {
               runtimeException = e;
            }
         }
         if (interruptedException != null) {
            throw interruptedException;
         } else if (timeoutException != null) {
            throw timeoutException;
         } else if (deadlockException != null) {
            throw deadlockException;
         } else if (runtimeException != null) {
            throw runtimeException;
         }

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
danberindei: Is that really necessary? Couldn't it CAS the next owner directly, instead of having a `null` intermediate state?

danberindei:
Looks really good @pruivo  although in some places I think there is still room for simplification.

Have you tried running benchmarks? The test suite duration seems to be fine now (it was 1 minute longer than usual when I first tried it, probably because of the non-static field updaters), but I'm curious if you have some results with a more focused benchmark.

core/src/main/java/org/infinispan/util/concurrent/locks/LockPromise.java
pruivo: if the example above does not convince you, I'll merge the states... although, my sixth sense is telling me that I'm shooting myself...

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
pruivo: I know that the extended-stats are not very precise, but I tried to make the most reliable I can. Recording the same wait time for all of them does not seems right to me. Imagine that you have tx that invokes lockAll(K1,K2). K2 is locked by another tx. to be precise, K1 waiting time should be zero and K2 don't. If you average it, the results are no longer reliable.

I didn't get what you mean by
```
if each tx locked the same keys it would average the waiting time for the first key with 0 for the others.
```

core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
pruivo: I'll see what I can do

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
danberindei: I was talking about how master/Beta2 works: locks are acquired sequentially, so if all your txs acquire locks {`k1`, `k2`, `k3`}, in the same order, the reported waiting time reflects only the time spent waiting for the `k1` lock. 

The waiting times for `k2` and `k3` are reported as 0 on master, and with your approach they will be reported the same as the waiting time for `k1`. The average waiting time will be `x/3` on master, and `x` with your approach, and I'm starting to think that maybe the master result is more correct (and you could achieve the same by reporting only the total waiting time and the number of locked keys, without key-awareness).

core/src/main/java/org/infinispan/util/concurrent/locks/LockPromise.java
danberindei: It didn't convince me, sorry :)

Unless you have some future plans to use this difference in the behaviour of `cancel()` before and after `lock()`, I think it's better to treat the `LockPromise` as a regular `Future` that does its job regardless of whether someone calls `get()` or not.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
pruivo: it looks correct to me. if the keys are hold by the same transaction and we are no acquire the keys in parallel, then the waiting time must be ```x``` for all the keys. if the keys are hold by different transaction, then we would have ```x1, x2, x3```.

pruivo:
I haven't benchmark but I'm not expecting any throughput improvement or loss.  I'm just targeting a better  thread handling :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
danberindei: Ok, let's leave it like this. But if you get an idea to implement this without any overhead when extended statistics are not used, feel free to make the change :)

pruivo:
@danberindei rebased. is anything missing?

danberindei:
Integrated, thanks Pedro!

PR #3612, opened 2015-07-28
ISPN-5643 Simplify handling of previous values

danberindei:
https://issues.jboss.org/browse/ISPN-5643
https://issues.jboss.org/browse/ISPN-5644

* Add readExistingValues method to VisitableCommand
* Use DefaultResponseGenerator everywhere
* Set the IGNORE_RETURN_VALUES flag when unreliableReturnValues is enabled
* Fix marshallable check for LockControlCommand
* SKIP_CACHE_LOAD should skip the loader even for DELTA_WRITE
* Streamline SKIP_REMOTE_LOOKUP handling


core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -129,6 +129,8 @@
 @MBean(objectName = CacheImpl.OBJECT_NAME, description = "Component that represents an individual cache instance.")
 public class CacheImpl<K, V> implements AdvancedCache<K, V> {
    public static final String OBJECT_NAME = "Cache";
+   public static final EnumSet<Flag> IGNORE_RETURN_VALUES_FLAG_SET = EnumSet.of(IGNORE_RETURN_VALUES);
```
wburns: private?

core/src/main/java/org/infinispan/interceptors/ActivationInterceptor.java
```diff
@@ -2,6 +2,7 @@
 
 import org.infinispan.commands.FlagAffectedCommand;
 import org.infinispan.commands.write.PutMapCommand;
+import org.infinispan.commands.write.WriteCommand;
```
wburns: Seems it can be removed.

core/src/main/java/org/infinispan/interceptors/IsMarshallableInterceptor.java
```diff
@@ -95,15 +95,20 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
 
    @Override
    public Object visitLockControlCommand(TxInvocationContext ctx, LockControlCommand command) throws Throwable {
-      if (isStoreAsBinary() || isClusterInvocation(ctx, command))
-         checkMarshallable(command.getKeys());
+      if (isStoreAsBinary() || isClusterInvocation(ctx, command)) {
```
wburns: Did you mean to include this change in here?  This doesn't seem to relate to this PR.  Maybe another commit for reduce IsMarshallableInterceptor allocations?

core/src/main/java/org/infinispan/interceptors/VersionedEntryWrappingInterceptor.java
```diff
@@ -41,33 +41,47 @@ public void initialize(VersionGenerator versionGenerator) {
 
    @Override
    public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand command) throws Throwable {
+      VersionedPrepareCommand versionedPrepareCommand = (VersionedPrepareCommand) command;
       if (ctx.isOriginLocal()) {
-         ((VersionedPrepareCommand) command).setVersionsSeen(ctx.getCacheTransaction().getVersionsRead());
+         (versionedPrepareCommand).setVersionsSeen(ctx.getCacheTransaction().getVersionsRead());
```
wburns: parens can be removed still

core/src/main/java/org/infinispan/interceptors/VersionedEntryWrappingInterceptor.java
```diff
@@ -41,33 +41,47 @@ public void initialize(VersionGenerator versionGenerator) {
...
+      if (ctx.isOriginLocal() && !ctx.getCacheTransaction().isFromStateTransfer()) {
+         newVersionData =
+               cdl.createNewVersionsAndCheckForWriteSkews(versionGenerator, ctx, versionedPrepareCommand);
+      }
 
       Object retval = invokeNextInterceptor(ctx, command);
 
-      if (!ctx.isOriginLocal()) newVersionData = cdl.createNewVersionsAndCheckForWriteSkews(versionGenerator, ctx, (VersionedPrepareCommand) command);
-      if (command.isOnePhaseCommit()) ctx.getCacheTransaction().setUpdatedEntryVersions(((VersionedPrepareCommand) command).getVersionsSeen());
+      if (!ctx.isOriginLocal()) newVersionData =
```
wburns: Spacing looks pretty ugly now since there aren't braces.

core/src/main/java/org/infinispan/interceptors/VersionedEntryWrappingInterceptor.java
```diff
@@ -41,33 +41,47 @@ public void initialize(VersionGenerator versionGenerator) {
...
+            cdl.createNewVersionsAndCheckForWriteSkews(versionGenerator, ctx, versionedPrepareCommand);
+      if (command.isOnePhaseCommit()) {
+         ctx.getCacheTransaction().setUpdatedEntryVersions((versionedPrepareCommand).getVersionsSeen());
```
wburns: extra parens still

core/src/main/java/org/infinispan/interceptors/VersionedEntryWrappingInterceptor.java
```diff
@@ -41,33 +41,47 @@ public void initialize(VersionGenerator versionGenerator) {
 
    @Override
    public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand command) throws Throwable {
+      VersionedPrepareCommand versionedPrepareCommand = (VersionedPrepareCommand) command;
```
wburns: To be honest it looks like the changes here are just cleaning up the class and nothing with this PR, right?

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -504,15 +511,63 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
+      }
+      return false;
+   }
+
+   protected InternalCacheEntry fetchValueLocallyIfAvailable(ConsistentHash consistentHash, Object key) {
+      if (consistentHash.isKeyLocalToNode(rpcManager.getAddress(), key)) {
+         return dataContainer.get(key);
+      } else if (isL1Enabled) {
+         InternalCacheEntry entry = dataContainer.get(key);
+         return entry != null && entry.isL1Entry() ? entry : null;
```
wburns: The latter case shouldn't happen I would think, but maybe without partition support.  Should we remove the value then?

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -504,15 +511,63 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
+   protected abstract boolean previousValueIsRemote(Object key);
 
    protected abstract void remoteGetBeforeWrite(InvocationContext ctx, WriteCommand command, Object key) throws Throwable;
+
+   protected boolean readNeedsRemoteGet(InvocationContext ctx, AbstractDataCommand command) {
+      if (command.hasFlag(Flag.CACHE_MODE_LOCAL) || command.hasFlag(Flag.SKIP_REMOTE_LOOKUP)) {
+         return false;
+      }
+      Object key = command.getKey();
+      ConsistentHash ch = stateTransferManager.getCacheTopology().getReadConsistentHash();
```
wburns: This is not always used.  Maybe we should do ``!isValueAvailableLocally(stateTransferManager.getCacheTopology().getReadConsistentHash(), key);`` and assign it in the if block below?

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -290,110 +271,112 @@ private boolean checkCacheNotFoundResponseInPartitionHandling(TransactionBoundar
...
          // The return value only matters on the originator.
          // Conditional commands also check the previous value only on the originator.
-         if (isNeedReliableReturnValues(command) || command.isConditional())
-            return true;
+         if (!command.readsExistingValues()) {
+            return false;
+         }
+         // TODO Could make DELTA_WRITE/ApplyDeltaCommand override SKIP_REMOTE_LOOKUP by changing next line to
+         // return !command.hasFlag(Flag.SKIP_REMOTE_LOOKUP) || command.alwaysReadsExistingValues();
+         if (command.hasFlag(Flag.SKIP_REMOTE_LOOKUP)) {
```
wburns: It seems this could be changed to ``return !command.hasFlag(Flag.SKIP_REMOTE_LOOKUP))``

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -290,110 +271,112 @@ private boolean checkCacheNotFoundResponseInPartitionHandling(TransactionBoundar
...
+            return false;
+         }
+      } else {
+         // Ignore SKIP_REMOTE_LOOKUP on remote nodes
+         // TODO Can we ignore the CACHE_MODE_LOCAL flag as well?
+         if (!command.alwaysReadsExistingValues()) {
```
wburns: It seems this could be changed to ``return command.alwaysReadsExistingValues()``  But these 2 are personal preference, so however you wanted it :)

core/src/main/java/org/infinispan/remoting/responses/DefaultResponseGenerator.java
```diff
@@ -2,22 +2,30 @@
...
 public class DefaultResponseGenerator implements ResponseGenerator {
-   @Override
+   private static final Log log = LogFactory.getLog(DefaultResponseGenerator.class);
+   private static final boolean trace = log.isTraceEnabled();
+
    public Response getResponse(CacheRpcCommand command, Object returnValue) {
       if (returnValue instanceof Response)
          return (Response) returnValue;
 
-      if (returnValue instanceof EntryVersionsMap || command.isReturnValueExpected()) {
```
wburns: I am not familiar with the version stuff, but we don't need this anymore?

wburns:
It seems fine to me if tests pass, but I am not as familiar with the details with how this stuff normally operates.  Guessing @pruivo would be best to look this over for final confirmation.

core/src/main/java/org/infinispan/interceptors/IsMarshallableInterceptor.java
danberindei: Yeah, I should have kept it in a separate commit... it's not about allocations, the problem was that `checkMarshallable` was checking if the collection was serializable, instead of checking each key.

core/src/main/java/org/infinispan/interceptors/VersionedEntryWrappingInterceptor.java
danberindei: Yeah, I have another PR coming up that's more focused on the entry wrapping, so I could move it there.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: It can happen, if the consistent hash changed but we haven't yet removed the values that we no longer own: https://issues.jboss.org/browse/ISPN-4444

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: I'm moving the `ctx.isOriginLocal()` check at the beginning of the method.

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
danberindei: Right, I think I had it a little more complicated at some point and I thought there may be something else after the big if.

core/src/main/java/org/infinispan/remoting/responses/DefaultResponseGenerator.java
danberindei: No, because `VersionedPrepareCommand.isReturnValueExpected()` returns `true`.

danberindei:
Updated. I had to add another commit, to always create a new EnumSet when adding a flag to a command, otherwise the `IGNORE_RETURN_VALUES_FLAG_SET` constant in CacheImpl wouldn't really stay constant.

core/src/main/java/org/infinispan/interceptors/IsMarshallableInterceptor.java
wburns: Ah yes 2 for 1 deal :)

core/src/main/java/org/infinispan/context/Flag.java
```diff
@@ -284,6 +295,44 @@
       }
    }
 
+   public static Set<Flag> addFlag(Set<Flag> flags, Flag newFlag) {
```
wburns: :+1: this is a lot nicer than before

core/src/main/java/org/infinispan/commands/control/LockControlCommand.java
```diff
@@ -149,7 +148,7 @@ public void setParameters(int commandId, Object[] args) {
       globalTx = (GlobalTransaction) args[i++];
       unlock = (Boolean) args[i++];
       keys = (List<Object>) args[i++];
-      flags = (Set<Flag>) args[i];
+      flags = (EnumSet<Flag>) args[i];
```
wburns: This change seems unneeded, did you want to change the declared type of flags as well?  Otherwise I could see someone possibly passing in the wrong set impl.

danberindei: Ah yes, I wanted to convert everything to an `EnumSet` in `setFlags()`, but then I backed down on it. Reverting.

wburns:
It looks like we have a bunch of new test failures.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -129,6 +129,8 @@
 @MBean(objectName = CacheImpl.OBJECT_NAME, description = "Component that represents an individual cache instance.")
 public class CacheImpl<K, V> implements AdvancedCache<K, V> {
    public static final String OBJECT_NAME = "Cache";
+   private static final EnumSet<Flag> IGNORE_RETURN_VALUES_FLAG_SET = EnumSet.of(IGNORE_RETURN_VALUES);
```
pruivo: shouldn't this be immutable? just worried if in the future we add flags directly instead of copying it.

pruivo: btw, is it worth it? I mean, we safe one allocation but every time you add or remove flags you have to make a copy...

core/src/main/java/org/infinispan/commands/write/PutKeyValueCommand.java
```diff
@@ -73,6 +74,16 @@ public Object acceptVisitor(InvocationContext ctx, Visitor visitor) throws Throw
    }
 
    @Override
+   public boolean readsExistingValues() {
+      return putIfAbsent || flags == null || !flags.contains(Flag.IGNORE_RETURN_VALUES);
```
pruivo: you can use ```hasFlag()``` to be consistent with the other commands.

core/src/main/java/org/infinispan/commands/write/PutKeyValueCommand.java
```diff
@@ -73,6 +74,16 @@ public Object acceptVisitor(InvocationContext ctx, Visitor visitor) throws Throw
...
+   }
+
+   @Override
+   public boolean alwaysReadsExistingValues() {
+      return flags != null && flags.contains(Flag.DELTA_WRITE);
```
pruivo: same as above

core/src/main/java/org/infinispan/context/Flag.java
```diff
@@ -162,19 +171,21 @@
...
-    * with the conditional remove ({@link java.util.concurrent.ConcurrentMap#remove(Object, Object)})
-    * and conditional replace ({@link java.util.concurrent.ConcurrentMap#replace(Object, Object, Object)})
-    * operations which return a boolean result: this has a minimal serialization payload and is
-    * generally relevant in order to determine whether the operation was successful or not.
+    * cache but might not be interested in the return value.
+    *
+    * This flag is ignored for operations that need the existing value to execute
+    * correctly, e.g. {@link Cache#get(Object)},
+    * conditional remove ({@link Cache#remove(Object, Object)})
+    * and conditional replace ({@link Cache#replace(Object, Object, Object)}).
```
pruivo: replace is always conditional

core/src/main/java/org/infinispan/context/Flag.java
```diff
@@ -284,6 +295,44 @@
...
+      }
+      Collections.addAll(copy, newFlags);
+      return copy;
+   }
+
+   public static Set<Flag> addFlags(Set<Flag> flags, Set<Flag> newFlags) {
+      if (newFlags == null || newFlags.isEmpty()) {
+         return flags;
+      }
+      if (flags == null) {
```
pruivo: or flags.isEmpty?

core/src/main/java/org/infinispan/context/Flag.java
```diff
@@ -284,6 +295,44 @@
...
+      copy.add(newFlag);
+      return copy;
+   }
+
+   public static Set<Flag> addFlags(Set<Flag> flags, Flag... newFlags) {
+      if (newFlags == null || newFlags.length == 0) {
+         return flags;
+      }
+      EnumSet<Flag> copy;
+      if (flags == null) {
```
pruivo: or flags.isEmpty?

core/src/main/java/org/infinispan/context/Flag.java
```diff
@@ -284,6 +295,44 @@
       }
    }
 
+   public static Set<Flag> addFlag(Set<Flag> flags, Flag newFlag) {
+      if (flags == null) {
```
pruivo: or flags.isEmpty?

core/src/main/java/org/infinispan/context/Flag.java
```diff
@@ -3,6 +3,8 @@
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
+import java.util.Arrays;
```
pruivo: unused?

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -504,15 +511,64 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
+
+   protected boolean valueIsMissing(CacheEntry entry) {
+      return entry == null || (entry.isNull() && !entry.isRemoved() && !entry.skipLookup());
+   }
+
+   protected abstract boolean previousValueIsRemote(Object key);
 
    protected abstract void remoteGetBeforeWrite(InvocationContext ctx, WriteCommand command, Object key) throws Throwable;
+
+   protected boolean readNeedsRemoteGet(InvocationContext ctx, AbstractDataCommand command) {
```
pruivo: ```readNeedsRemoteValue``` sounds better IMO

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -504,15 +511,64 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
     * @return Whether a remote get is needed to obtain the previous values of the affected entries.
     */
-   protected abstract boolean needValuesFromPreviousOwners(InvocationContext ctx, WriteCommand command);
+   protected abstract boolean writeNeedsRemoteValue(InvocationContext ctx, WriteCommand command, Object key);
+
+   protected boolean valueIsMissing(CacheEntry entry) {
+      return entry == null || (entry.isNull() && !entry.isRemoved() && !entry.skipLookup());
+   }
+
+   protected abstract boolean previousValueIsRemote(Object key);
```
pruivo: can't you use ```isValueAvailableLocally```? 
if not, the implementation is the same in both subclasses.

pruivo:
In ```BaseDistributionInterceptor``` the method ```visitGetAllCommand```, it should check for ```IGNORE_RETURN_VALUES```.

core/src/main/java/org/infinispan/commands/control/LockControlCommand.java
```diff
@@ -15,7 +15,6 @@
 import org.infinispan.util.logging.LogFactory;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.EnumSet;
```
pruivo: unused import

pruivo:
@danberindei the query API module is failing. it needs the previous values from PutMapCommand.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
danberindei: If I made it immutable then it wouldn't be an `EnumSet` any more, and the bulk `EnumSet` methods work better when both sets are `EnumSets`.

I'm not sure it's worth it TBH, at first I thought I could remove all the flag additions in the interceptors but it didn't quite work out. I expect the number of allocations would be about the same either way, and since I already had this implemented I didn't try making a copy at the beginning.

core/src/main/java/org/infinispan/context/Flag.java
danberindei: `copyOf` isn't really more expensive than `noneOf`, so I'd rather avoid the `isEmpty` check.

galderz:
Needs rebasing

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: Yep, `isValueAvailableLocally` works. I was thinking the L1 lookup wouldn't help because the node executing the command is an owner in the write CH, but I guess that's not true in transactional caches.

danberindei:
Updated, the query tests are now fixed (but the values are still not accurate when using a store).

danberindei:
Updated

danberindei:
Test failure in IntegrationTest is unrelated.

pruivo:
@danberindei, in the functional commands, ```alwaysReadsExistingValues``` does not need an implementation

pruivo:
@danberindei it looks good. just the small comment above

wburns:
Pulling...

wburns:

PR #3613, opened 2015-07-28
[ISPN-5516] SingleFileStore should skip loading from disk if value an…

ryanemerson:
…d metadata are not requested

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
+++ b/core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
@@ -473,10 +473,22 @@ public class SingleFileStore<K, V> implements AdvancedLoadWriteStore<K, V> {
          resizeLock.readLock().unlock();
       }
 
+      org.infinispan.commons.io.ByteBuffer valueBb = null;
+      org.infinispan.commons.io.ByteBuffer metadataBb = null;
+
+      // If we only require the key, then no need to read disk
+      if (!loadValue && !loadMetadata) {
+         try {
+            return ctx.getMarshalledEntryFactory().newMarshalledEntry(key, valueBb, metadataBb);
+         } finally {
+            fe.unlock();
+         }
+      }
+
       final byte[] data;
       try {
          // load serialized data from disk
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -473,10 +473,26 @@ public boolean delete(Object key) {
...
 
+      org.infinispan.commons.io.ByteBuffer keyBb;
+      org.infinispan.commons.io.ByteBuffer valueBb = null;
+      org.infinispan.commons.io.ByteBuffer metadataBb = null;
+
+      // If we only require the key, then no need to read disk
+      if (!loadValue && !loadMetadata) {
+         try {
+            keyBb = ctx.getMarshaller().objectToBuffer(key);
+            return ctx.getMarshalledEntryFactory().newMarshalledEntry(keyBb, valueBb, metadataBb);
```
wburns: We should use https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/marshall/core/MarshalledEntryFactory.java#L18 so we don't have to marshall the key unless the caller wants the marshalled form.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -473,10 +473,26 @@ public boolean delete(Object key) {
          resizeLock.readLock().unlock();
       }
 
+      org.infinispan.commons.io.ByteBuffer keyBb;
+      org.infinispan.commons.io.ByteBuffer valueBb = null;
```
wburns: I believe we should not declare this here and leave it below and just pass null on line 484 for both values.  Is a bit more clear imo.

ryanemerson: If null is passed for both values, then we get an "ambiguous method call" error.  I passed valueBb and metadataBb to avoid casting for both values.

wburns: Ah yes I see.  Interestingly, the invocation you had below didn't need the casts :)  But with my other rework comment it will.

wburns:
Looks good, pulling...

wburns:

wburns:
https://issues.jboss.org/browse/ISPN-5516

PR #3614, opened 2015-07-29
ISPN-5625 Distributed stream iterator should try to stay local if pos…

gustavonalle:
…sible

https://issues.jboss.org/browse/ISPN-5625

Measured improvement of 2~3x in the Spark job execution (which relies heavily on iteration)

### DIFF
--- a/core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
+++ b/core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
@@ -22,7 +22,6 @@ import org.infinispan.util.concurrent.TimeoutException;
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
-import java.io.Serializable;
 import java.util.*;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -431,104 +431,37 @@ public long count() {
...
+      CloseableIterator<R> closeableIterator = new CloseableSuppliedIterator<>(supplier);
+      onClose(() -> supplier.close());
+      return closeableIterator;
+   }
 
-         Thread thread = Thread.currentThread();
-         executor.execute(() -> {
-            try {
-               log.tracef("Thread %s submitted iterator request for stream", thread);
+   private void ignoreRehashIteration(Consumer<R> consumer, IteratorSupplier<R> supplier, boolean iteratorParallelDistribute) {
```
wburns: Thanks, I was going to do this in another PR even :+1:

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -537,23 +470,130 @@ public long count() {
...
+         }
+      });
+   }
 
-      CloseableIterator<R> closeableIterator = new CloseableSuppliedIterator<>(supplier);
-      onClose(() -> supplier.close());
-      return closeableIterator;
+   private Set<Integer> segmentsToProcess(IteratorSupplier<R> supplier, KeyTrackingConsumer<Object, R> results,
+                                          Set<Integer> segmentsToProcess, UUID id) {
+      String strId = id == null ? "local" : "identifier " + id;
```
wburns: I would like if possible to avoid the string concat, maybe just leave it as id or some way of using ``if (log.traceEnabled())`` ?

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -537,23 +470,130 @@ public long count() {
...
+                     } catch (InterruptedException e) {
+                        throw new CacheException(e);
+                     }
+                     segmentsToProcess = segmentsToProcess(supplier, results, segmentsToProcess, id);
+                  } finally {
+                     csm.forgetOperation(id);
+                  }
+               } else {
+                  performLocalRehashAwareOperation(results, segmentsToProcess, ch, segments, op,
+                          () -> ch.getSegmentsForOwner(localAddress), null);
```
wburns: We can just return the segmentsForOwner from https://github.com/infinispan/infinispan/pull/3614/files#diff-be6f435a8edfd997f62d9a90ad59582bR514 if you push the variable out of the if brackets.  It can be defined as null and then only set if runLocal was set since we know stayLocal can only be true if it went into runLocal.  This will avoid an additional segmentsForOwner lookup.

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -537,23 +470,130 @@ public long count() {
...
+                       distributedBatchSize);
+               if (!stayLocal) {
+                  UUID id = csm.remoteStreamOperationRehashAware(iteratorParallelDistribute, parallel, ch,
+                          segmentsToProcess, keysToFilter, new AtomicReferenceArrayToMap<>(results.referenceArray),
+                          includeLoader, op, results);
+                  supplier.pending = id;
+                  try {
+                     if (runLocal) {
+                        performLocalRehashAwareOperation(results, segmentsToProcess, ch, segments, op,
+                                () -> ch.getPrimarySegmentsForOwner(localAddress), id);
```
wburns: Same here with getPrimarySegmentsForOwner, you can just pass segments as is since the way the logic works it will always be the primary owners.

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
wburns: Actually ignore what I said :)

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
wburns: Actually ignore what I said :)

core/src/test/java/org/infinispan/stream/DistributedStreamIteratorTest.java
```diff
@@ -293,6 +296,44 @@ public void testLocallyForcedStream() {
...
+
+      ClusterStreamManager clusterStreamManager = replaceWithSpy(cache0);
+
+      IntStream.rangeClosed(0, 499).boxed().forEach(i -> cache0.put(i, i.toString()));
+
+      ConsistentHash ch = cache0.getAdvancedCache().getDistributionManager().getConsistentHash();
+      Set<Integer> segmentsCache0 = ch.getSegmentsForOwner(cache0.getCacheManager().getAddress());
+
+      CacheStream<Map.Entry<Object, String>> stream = cache0.entrySet().stream();
+      if (!rehashAware) stream.disableRehashAware();
```
wburns: We should reassign stream to the return value.

wburns:
@gustavonalle Looks good, just a couple very minor comments.

Also a tip (not sure if you know or not), but when you change up the distributed cache stream code related to rehash, I would highly highly suggest running the DiistributedStreamRehashStressTest to verify there was no regression there.  Unfortunately it is a stress test so it doesn't get run with all of the other tests.

gustavonalle:
@wburns PR updated, and ```DistributedStreamRehashStressTest``` is passing

wburns:
Pulling...

wburns:
Integrated into master, Thanks @gustavonalle !

PR #3615, opened 2015-07-29
ISPN-694 Create expiration notification for in-memory cache entries

wburns:
* Removed CacheEntryEvicted event
* Added CacheEntryExpired event
* Expiration now occurs consistent across memory and stores

https://issues.jboss.org/browse/ISPN-694
https://issues.jboss.org/browse/ISPN-5646


wburns:
Preview atm until build confirms tests.

wburns:
Note this is only to officially add in expiration events with guarantees of 1 event per node.  This doesn't include clustered based expiration, that is for another PR.

I created https://issues.jboss.org/browse/ISPN-5646 to remove the old evicted event, however now it is hard to pull that change out as I essentially replaced it with the new expired event.  If you really want I can try to extract those.

wburns:
Also in case if someone is curious local expiration is not guaranteed for entries in a shared cache loader.  Due to possible events the first node that sees the shared entry has expired will remove it.  If you have any suggestions let me know.

wburns:
Tests seem to be passing fine now.  I am torn whether we should have an extra entry count excluding expired entry count exposed via JMX or not.

core/src/main/java/org/infinispan/container/DataContainer.java
```diff
@@ -160,9 +171,23 @@
...
+    * @return iterator that doesn't produce expired entries
+    */
+   @Override
+   Iterator<InternalCacheEntry<K, V>> iterator();
+
+   /**
+    * Same as {@link DataContainer#iterator()} except that is also returns expired entries.
+    * @return iterator that returns all entries including expired ones
+    */
+   Iterator<InternalCacheEntry<K, V>> expiredIterator();
```
danberindei: `iteratorIncludingExpired()`, to match `sizeIncludingExpired()`?

core/src/main/java/org/infinispan/expiration/ExpirationManager.java
```diff
@@ -29,7 +30,17 @@
    void processExpiration();
 
    /**
-    * @return true if expiration is enabled, false otherwise
+    * @return true if expiration reaper is enabled, false otherwise
```
danberindei: Reaper thread would be a bit clearer.

core/src/main/java/org/infinispan/expiration/ExpirationManager.java
```diff
@@ -29,7 +30,17 @@
...
     */
    boolean isEnabled();
+
+   /**
+    * This should be invoked passing in an entry that is now expired.  Note this method is assumed to not have the
```
danberindei: s/is assumed/assumes/

Also, I'm not sure this will work well with Pedro's locking PR https://github.com/infinispan/infinispan/pull/3606, which changes the lock owner from the calling thread to the command's invocation id.

core/src/main/java/org/infinispan/expiration/impl/ExpirationManagerImpl.java
```diff
@@ -107,6 +118,45 @@ public boolean isEnabled() {
...
+            return null;
+         }
+         return oldEntry;
+      }));
+   }
+
+   @Override
+   public void handleInStoreExpiration(K key) {
+      dataContainer.compute(key, ((k, oldEntry, factory) -> {
+         if (oldEntry == null || (oldEntry.canExpire() && oldEntry.isExpired(timeService.time()))) {
```
danberindei: Couldn't the entry be `null` because of another reason, e.g. it was removed explicitly?

core/src/main/java/org/infinispan/interceptors/CacheMgmtInterceptor.java
```diff
@@ -368,12 +368,12 @@ public long getAverageRemoveTime() {
    }
 
    @ManagedAttribute(
-         description = "Number of entries currently in the cache",
+         description = "Number of entries currently in the cache including expired entries",
```
danberindei: I'd also add "and excluding loaders/stores"

core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifier.java
```diff
@@ -55,13 +55,9 @@ void notifyCacheEntriesEvicted(Collection<InternalCacheEntry<? extends K, ? exte
          InvocationContext ctx, FlagAffectedCommand command);
 
    /**
-    * Syntactic sugar
-    * @param key key evicted
-    * @param value value evicted
-    * @param ctx context
+    * Notifies all registered listeners of a CacheEntryRemoved event.
```
danberindei: CacheEntryExpired

core/src/main/java/org/infinispan/notifications/cachelistener/annotation/CacheEntryExpired.java
```diff
@@ -6,24 +6,22 @@
...
+ * This annotation should be used on methods that need to be notified when a cache entry is expired
  * <p/>
  * Methods annotated with this annotation should be public and take in a single parameter, a {@link
- * org.infinispan.notifications.cachelistener.event.CacheEntryEvictedEvent} otherwise an {@link
+ * org.infinispan.notifications.cachelistener.event.CacheEntryExpiredEvent} otherwise an {@link
  * org.infinispan.notifications.IncorrectListenerException} will be thrown when registering your cache listener.
  * <p/>
+ *  Locking: notification is performed WITH locks on the given key.
+ * <p/>
  * Any exceptions thrown by the listener will abort the call. Any other listeners not yet called will not be called,
```
danberindei: I'd rather ignore expiration listener exceptions, since they're post-only.

wburns: Yes true, gotta love copy paste :)

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
```diff
@@ -78,7 +80,7 @@ public void testExpiredData() throws InterruptedException {
 
       dc.put("k", "v", new EmbeddedMetadata.Builder().lifespan(0, TimeUnit.MINUTES).build());
       Thread.sleep(100);
-      assert dc.size() == 1;
+      assert dc.size() == 0;
       dc.purgeExpired();
```
danberindei: Do you still need this call?

core/src/main/java/org/infinispan/container/DataContainer.java
wburns: Sure I was debating which way I should do it, was hoping someone else would chime in :+1:

core/src/main/java/org/infinispan/expiration/ExpirationManager.java
wburns: Makes sense

core/src/main/java/org/infinispan/expiration/impl/ExpirationManagerImpl.java
wburns: Yes now that I think about it there is a small gap between when the reaper thread found an expired entry before it acquire the data container lock that if another thread removes the value during that time it would inadvertently raise the event.  I am not sure how I can fix this other than rereading the store again which is pretty ugly, or have the store acquire the lock during the purging (even worse).

core/src/main/java/org/infinispan/interceptors/CacheMgmtInterceptor.java
wburns: Sure, maybe instead we change change to be in memory entries ?  Something like

> Number of entries currently in memory including expired entries

core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifier.java
wburns: Thanks for catching another copy paste :)

core/src/main/java/org/infinispan/expiration/ExpirationManager.java
wburns: The lock here is not the lock we keep but rather the lock in the underlying data container.  I will go back and triple check this is sufficient for the store case.

danberindei: I think then you need to remove the part about possibly going remote to acquire the lock.

wburns: I apologize that was for the next iteration that snuck in here :)

core/src/main/java/org/infinispan/expiration/impl/ExpirationManagerImpl.java
danberindei: I'm fine with leaving it as is, but adding a comment here and a disclaimer in the expiration listener javadoc.

core/src/main/java/org/infinispan/interceptors/CacheMgmtInterceptor.java
danberindei: Sounds good to me

core/src/main/java/org/infinispan/container/DataContainer.java
danberindei: `expiredIterator()` sounds to me like you're only iterating on the expired entries.

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
wburns: I left it just for good measure, but yeah it can be safely removed.  WDYT?

wburns:
Updated per comments.  Also was curious if you had an opinion on the possible extra JMX value for in memory not counting expired, or maybe we don't need it?

wburns:
Actually I forgot to add the config check to verify the reaper thread is enabled when a cache store is present, let me add that.

danberindei:
@wburns I'd say no need for another JMX attribute excluding expired, at least until someone requests it explicitly.

wburns:
@danberindei sounds good, that is what I was thinking as well

I added in a debug message for the reaper thread.

wburns:
Do not integrate yet, I have to manually test this doesn't break hibernate infinispan integration as well.

wburns:
Rebased and verified this still works with hibernate orm.

galderz:
Integrated, this will make a lot of people happy :), great work once again! :)

PR #3617, opened 2015-07-30
ISPN-5621 Remove add and addAll support from entrySet methods.

wburns:
https://issues.jboss.org/browse/ISPN-5621


core/src/test/java/org/infinispan/api/APINonTxTest.java
```diff
@@ -8,6 +8,9 @@
 import org.infinispan.test.SingleCacheManagerTest;
 import org.infinispan.test.fwk.TestCacheManagerFactory;
 import org.testng.annotations.Test;
+
+import javax.resource.NotSupportedException;
```
pruivo: I think this import is not used.

wburns: Yep :) auto imported the wrong exception and missed removing, doh.

wburns:
Refactored imports.

pruivo:
integrated! thanks @wburns :)

PR #3618, opened 2015-07-31
ISPN-5653 keySet().iterator() does not propagate flags to remove()

wburns:
* Made sure command passes a cache with flags set properly

https://issues.jboss.org/browse/ISPN-5653


wburns:
Note this is for 7.2.x branch, the master one will have to be different in a separate PR.

galderz:
Integrated

PR #3619, opened 2015-07-31
ISPN-5653 keySet().iterator() does not propagate flags to remove()

wburns:
* Made sure to support iterator remove and various flags
* Added additional documentation describing this interaction
* Added tests for keySet, values and entrySet versions

https://issues.jboss.org/browse/ISPN-5653


wburns:
Some remote test failures I have to check out.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/iteration/IterationManager.scala
```diff
@@ -80,7 +80,7 @@ class DefaultIterationManager(val cacheManager: EmbeddedCacheManager) extends It
 
    override def start(cacheName: String, segments: Option[JavaBitSet], filterConverterFactory: Option[String], batch: Integer): IterationId = {
       val iterationId = UUID.randomUUID().toString
-      val stream = cacheManager.getCache(cacheName).getAdvancedCache.cacheEntrySet.stream().asInstanceOf[CacheStream[_]]
+      val stream = cacheManager.getCache(cacheName).getAdvancedCache.cacheEntrySet.stream().asInstanceOf[CacheStream[CacheEntry[AnyRef, AnyRef]]]
```
wburns: @galderz can you take a look at the typing here?  Both @gustavonalle and I had issues with scala working with Java covariance here as it always wants to return a Stream instead of CacheStream as the CacheCollection defines.

wburns:
Tests should be fixed now.

core/src/main/java/org/infinispan/stream/impl/local/AbstractLocalCacheStream.java
```diff
@@ -294,9 +295,16 @@ public boolean noneMatch(Predicate<? super R> predicate) {
 
    @Override
    public CloseableIterator<R> iterator() {
-      return Closeables.iterator(getOrCreateStream());
+      // If the stream is null we can assume no intermediate operations and thus we can support remove
+      if (stream == null) {
+         return removableIterator(Closeables.iterator(getOrCreateStream()));
```
pruivo: nitpick: if the stream is null, isn't worth it to invoke a ```createStream``` to avoid double checking? 
the same logic could be applied to the else branch by passing the stream directly.

core/src/main/java/org/infinispan/interceptors/distribution/DistributionBulkInterceptor.java
```diff
@@ -77,8 +80,8 @@ private BackingEntrySet(Cache cache, CacheSet<CacheEntry<K, V>> entrySet, LocalF
 
       @Override
       public CloseableIterator<CacheEntry<K, V>> iterator() {
-         return new CloseableIteratorMapper<>(Closeables.iterator(stream()),
-                 e -> new EntryWrapper<>(cache, e));
+         return new CloseableIteratorMapper<>(new RemovableCloseableIterator<>(Closeables.iterator(stream()), cache,
+                 e -> e.getKey()), e -> new EntryWrapper<>(cache, e));
```
pruivo: nitpick: ```e -> e.getKey()``` could be replaced by ```CacheEntry::getKey```. the same for the change below.
ps. not sure which one is more efficient.

core/src/main/java/org/infinispan/stream/impl/RemovableCloseableIterator.java
```diff
@@ -0,0 +1,34 @@
+package org.infinispan.stream.impl;
+
+import org.infinispan.Cache;
+import org.infinispan.commons.util.CloseableIterator;
+import org.infinispan.container.entries.CacheEntry;
```
pruivo: unused imports :)

core/src/main/java/org/infinispan/stream/impl/RemovableIterator.java
```diff
@@ -0,0 +1,65 @@
+package org.infinispan.stream.impl;
+
+import org.infinispan.Cache;
+import org.infinispan.commons.util.CloseableIterator;
```
pruivo: unused imports :)

core/src/test/java/org/infinispan/iteration/BaseEntryRetrieverTest.java
```diff
@@ -205,5 +212,26 @@ public void testFilterAndConverterCombinedConverterNotPassed() {
...
+      }, 0);
+
+      for (Iterator<CacheEntry<Object, Void>> it = cache(0, CACHE_NAME).getAdvancedCache().withFlags(
+              Flag.SKIP_CACHE_STORE).filterEntries(AcceptAllKeyValueFilter.getInstance()).converter(
+              NullValueConverter.getInstance()).iterator(); it.hasNext();) {
+         assertTrue(values.containsKey(it.next().getKey()));
+         it.remove();
+      }
+   }
+}
```
pruivo: shouldn't check if the cache is empty in the end?

core/src/main/java/org/infinispan/interceptors/distribution/DistributionBulkInterceptor.java
```diff
@@ -24,10 +24,13 @@
 import org.infinispan.stream.StreamMarshalling;
 import org.infinispan.stream.impl.ClusterStreamManager;
 import org.infinispan.stream.impl.DistributedCacheStream;
+import org.infinispan.stream.impl.RemovableCloseableIterator;
+import org.infinispan.stream.impl.RemovableIterator;
 import org.infinispan.stream.impl.tx.TxClusterStreamManager;
 import org.infinispan.stream.impl.tx.TxDistributedCacheStream;
 
 import java.io.Serializable;
```
pruivo: unused import

core/src/main/java/org/infinispan/stream/impl/local/AbstractLocalCacheStream.java
```diff
@@ -7,6 +7,7 @@
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.distribution.ch.ConsistentHash;
 import org.infinispan.factories.ComponentRegistry;
+import org.infinispan.stream.impl.RemovableCloseableIterator;
```
pruivo: unused import and there is another one below.

pruivo:
@wburns some minor comments. I left the scala for the experts :)

core/src/main/java/org/infinispan/stream/impl/local/AbstractLocalCacheStream.java
wburns: But the problem is I can only apply the removable iterator if there was no intermediate operations.  After I invoke getOrCreateStream I can't tell if there was any intermediate operations.  Or am I missing your point?

core/src/main/java/org/infinispan/interceptors/distribution/DistributionBulkInterceptor.java
wburns: This would have 1 less class to load I believe.  Changing.

core/src/test/java/org/infinispan/iteration/BaseEntryRetrieverTest.java
wburns: I can add it sure.

core/src/main/java/org/infinispan/stream/impl/local/AbstractLocalCacheStream.java
pruivo: well, something like
```
if (stream == null) {
   return removableIterator(Closeables.iterator(createStream()));
 } else {
   return Closeables.iterator(stream);
}
```

wburns: So you are proposing a new createStream method.  That should work, not as pretty but then we wont' have another null check.

pruivo: or use the getAndCreate on the if branch :) is up to you

wburns:
Updated.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/iteration/IterationManager.scala
galderz: I had a look at this and dunno how to avoid the cast. The issue might be related to the fact that stream() method is marked as default but can't tell for sure. As indicated on IRC, if a similar scenario can be replicated in a small scale, it's easier to submit an issue at Scala.

gustavonalle: > I had a look at this and dunno how to avoid the cast

Me neither. After some debugging, I saw that the error occurs very early in the ```typer``` compiler phase

Anyway, created https://issues.scala-lang.org/browse/SI-9432

galderz:
Integrated

PR #3623, opened 2015-08-03
ISPN-5652 Change isEmpty to use anyMatch to prevent serializing entries

wburns:
https://issues.jboss.org/browse/ISPN-5652

### DIFF
--- a/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
+++ b/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
@@ -28,7 +28,6 @@ import org.infinispan.commons.CacheConfigurationException;
 import org.infinispan.commons.CacheException;
 import org.infinispan.commons.api.BasicCacheContainer;
 import org.infinispan.commons.marshall.StreamingMarshaller;
-import org.infinispan.commons.util.CloseableIterable;
 import org.infinispan.commons.util.Util;
 import org.infinispan.commons.util.concurrent.AbstractInProcessNotifyingFuture;
 import org.infinispan.commons.util.concurrent.NotifyingFuture;
@@ -50,10 +49,8 @@ import org.infinispan.factories.ComponentRegistry;
 import org.infinispan.factories.annotations.ComponentName;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.factories.annotations.SurvivesRestarts;
-import org.infinispan.filter.AcceptAllKeyValueFilter;
 import org.infinispan.filter.KeyFilter;
 import org.infinispan.filter.KeyValueFilter;
-import org.infinispan.filter.NullValueConverter;
 import org.infinispan.interceptors.InterceptorChain;
 import org.infinispan.interceptors.base.CommandInterceptor;
 import org.infinispan.iteration.EntryIterable;
@@ -381,11 +378,7 @@ public class CacheImpl<K, V> implements AdvancedCache<K, V> {
    }
 
    final boolean isEmpty(EnumSet<Flag> explicitFlags, ClassLoader explicitClassLoader) {
-      // TODO: replace with stream findAny isPresent instead
-      try (CloseableIterable<CacheEntry<K, Void>> iterable = filterEntries(AcceptAllKeyValueFilter.getInstance(),
-            explicitFlags, explicitClassLoader).converter(NullValueConverter.getInstance())) {
-         return !iterable.iterator().hasNext();
-      }
+      return !entrySet(explicitFlags, explicitClassLoader).stream().anyMatch(StreamMarshalling.alwaysTruePredicate());
    }
 
    @Override
@@ -606,7 +599,11 @@ public class CacheImpl<K, V> implements AdvancedCache<K, V> {
 
    @Override
    public CacheCollection<V> values() {
-      return new ValueCacheCollection<>(this, cacheEntrySet());
+      return values(null, null);
+   }
+
+   CacheCollection<V> values(EnumSet<Flag> explicitFlags, ClassLoader explicitClassLoader) {
+      return new ValueCacheCollection<>(this, cacheEntrySet(explicitFlags, explicitClassLoader));
    }
 
    @Override
--- a/core/src/main/java/org/infinispan/stream/StreamMarshalling.java
+++ b/core/src/main/java/org/infinispan/stream/StreamMarshalling.java
@@ -8,6 +8,7 @@ import org.jboss.marshalling.util.IdentityIntMap;
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/infinispan/stream/StreamMarshalling.java
```diff
@@ -113,6 +137,7 @@ public V apply(Map.Entry<K, V> kvEntry) {
       private static final int ENTRY_KEY_FUNCTION = 1;
       private static final int ENTRY_VALUE_FUNCTION = 2;
       private static final int NON_NULL_PREDICATE = 3;
+      private static final int ALWAYS_TRUE_PREDICATE = 4;
```
tristantarrant: Why not replace this with an enum and use the ordinal()  and values()[i] ?

wburns: Sure, I can change this over if you think it is a bit more clear (I agree it would be less error prone).

tristantarrant: even if you need to manually assign the id, I'd still prefer an Enum with that id passed in the constructor

wburns:
Updated to use enums instead.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -606,7 +599,11 @@ final void clear(EnumSet<Flag> explicitFlags, ClassLoader explicitClassLoader) {
 
    @Override
    public CacheCollection<V> values() {
-      return new ValueCacheCollection<>(this, cacheEntrySet());
+      return values(null, null);
+   }
+
+   public CacheCollection<V> values(EnumSet<Flag> explicitFlags, ClassLoader explicitClassLoader) {
```
danberindei: This one shouldn't be public.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -381,11 +378,7 @@ public final boolean isEmpty() {
    }
 
    final boolean isEmpty(EnumSet<Flag> explicitFlags, ClassLoader explicitClassLoader) {
-      // TODO: replace with stream findAny isPresent instead
-      try (CloseableIterable<CacheEntry<K, Void>> iterable = filterEntries(AcceptAllKeyValueFilter.getInstance(),
-            explicitFlags, explicitClassLoader).converter(NullValueConverter.getInstance())) {
-         return !iterable.iterator().hasNext();
-      }
+      return !entrySet(explicitFlags, explicitClassLoader).stream().anyMatch(StreamMarshalling.alwaysTruePredicate());
```
danberindei: I would have expected to see `entrySet(explicitFlags, explicitClassLoader).isEmpty()` here, but I see it's the other way around - ` `cache.isEmpty()`, perhaps ?

core/src/main/java/org/infinispan/stream/StreamMarshalling.java
```diff
@@ -108,25 +132,24 @@ public V apply(Map.Entry<K, V> kvEntry) {
...
 
    public static final class StreamMarshallingExternalizer implements AdvancedExternalizer<Object> {
-
-      private static final int EQUALITY_PREDICATE = 0;
-      private static final int ENTRY_KEY_FUNCTION = 1;
-      private static final int ENTRY_VALUE_FUNCTION = 2;
-      private static final int NON_NULL_PREDICATE = 3;
+      enum ExternalizerId {
+         EQUALITY_PREDICATE, ENTRY_KEY_FUNCTION, ENTRY_VALUE_FUNCTION, NON_NULL_PREDICATE, ALWAYS_TRUE_PREDICATE;
+      }
```
danberindei: I actually liked the `public static final int` version better - we don't use enums in any other place we need stable ids, like for command ids.

tristantarrant: I was actually hoping for using the enum to also populate the "objects" IdentityIntMap, e.g.
ExternalizerId(Predicate p) {
objects.put(p, this.ordinal());
}

tristantarrant: Additionally, command ids use manual assignments because they can be extended by third party modules

danberindei: @tristantarrant `ListExternalizer` also uses int ids... 

In the end, I really don't think `switch (ExternalizerId.values()[number])` is any better than `switch (number)`. And it also makes the `default` case useless, because an invalid number will cause an `ArrayIndexOutOfBoundsException`.

tristantarrant: Switch supports enum directly :)

danberindei: That doesn't counter anything I've said :)

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
wburns: Thanks :)

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
wburns: Unfortunately doing so would require putting the implementation in 4 different places.  I am hoping to redo this at some point as it would be more efficient then having to go back into the cache to again.  I was thinking I would refactor that when I changed the iterator/spliterator stuff as I have to rewrite a good chunk of related code anyways.

core/src/main/java/org/infinispan/stream/StreamMarshalling.java
wburns: I have updated with what I see as the cleanest approach.  Let me know, thanks!

wburns:
Removed trivial as I don't think that applies with the externalizer rewrite :)

core/src/main/java/org/infinispan/stream/StreamMarshalling.java
```diff
@@ -138,7 +169,7 @@ public Integer getId() {
       public void writeObject(ObjectOutput output, Object object) throws IOException {
          int number = objects.get(object.getClass(), -1);
          output.writeByte(number);
-         switch (number) {
+         switch (ExternalizerId.values()[number]) {
```
danberindei: If you put the enums in the map instead of their ordinals, I think you can skip the `values()` access. But you should probably add a null check here as well.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
danberindei: Yeah, I thought that might be the case...

core/src/main/java/org/infinispan/stream/StreamMarshalling.java
wburns: Hrmm I can tweak the map to contain the enum instead.  Although it should be impossible to get a null here, note the should, haha.

wburns: Sorry misinterpreted your null check comment :)

wburns:
Updated by changing over to a regular HashMap and storing the enum instead of the ordinal value.

danberindei:
Integrated, thanks Will!

PR #3624, opened 2015-08-03
Remove deprecated and confusing thread configuration methods

rhusar:
IIRC this originally came to existence as a workaround for "you can't just remove methods from public API" but this is even worse, since upgrading infinispan in a project does compile but fails with NPE at runtime when trying to use these configuration methods...

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/global/AbstractGlobalConfigurationBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/global/AbstractGlobalConfigurationBuilder.java
@@ -28,66 +28,6 @@ abstract class AbstractGlobalConfigurationBuilder implements GlobalConfiguration
       return globalConfig.serialization();
    }
 
-   /**
-    * @deprecated This method always returns null now.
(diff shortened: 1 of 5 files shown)


wburns:
I always enjoy removing deprecated methods. Pulling...

wburns:
Actually I will add a JIRA to track this as well.

wburns:
https://issues.jboss.org/browse/ISPN-5660

wburns:
Integrated into master, thanks @rhusar !

rhusar:
Thanks @wburns !

(I didn't open a Jira originally thinking that since it never worked nobody would need to know that its gone but you are right, lets be concise so thanks for opening one.)

alanfx:
I'm just missing the word "deceitful" in the JIRA! :-)

rhusar:
Browsing the thesaurus at night.. couldn't resist! :-)

alanfx:
Haha, writing code and browsing the thesaurus? Take it easy, Rado! :-)

PR #3625, opened 2015-08-04
ISPN-5379 Persistence throwing ConcurrentModificationException

danberindei:
https://issues.jboss.org/browse/ISPN-5379

Make the modifications list immutable after prepare starts

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/TxInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/TxInterceptor.java
@@ -129,6 +129,9 @@ public class TxInterceptor<K, V> extends CommandInterceptor implements JmxStatis
 
    @Override
    public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand command) throws Throwable {
+      // Debugging for ISPN-5379
+      ctx.getCacheTransaction().freezeModifications();
+
       //if it is remote and 2PC then first log the tx only after replying mods
       if (this.statisticsEnabled) prepares.incrementAndGet();
       if (!ctx.isOriginLocal()) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/interceptors/TxInterceptor.java
```diff
@@ -129,6 +129,9 @@ public void init(TransactionTable txTable, Configuration configuration, RpcManag
 
    @Override
    public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand command) throws Throwable {
+      // Debugging for ISPN-5379
```
tristantarrant: I guess this comment is useless now ?

wburns:
@danberindei do we know why this exception occurs?  It seems we shouldn't be able to modify this transaction in another thread as the transaction manager scopes each to their own thread.  It looks like to me the problem is that we are using a foreach loop to go over the elements and one is removed by an underlying interceptor.  Can we just not change the foreach loop at https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/interceptors/CacheWriterInterceptor.java#L214 to instead use an iterator directly?

core/src/main/java/org/infinispan/interceptors/TxInterceptor.java
danberindei: The whole commit was intended as a commit aid for ISPN-5379, hoping to make the problem more visible and enabling a proper fix.

danberindei:
@wburns The foreach loop doesn't do any comodification checks by itself, only the iterator does...

wburns:
@danberindei yes I was thinking in regards to using the iterator remove method

wburns:
I didn't realize this commit was only for the purpose of trying to diagnose an issue.  In that case I can see adding this, however I would not close the JIRA.

wburns:
Integrating without closing JIRA.. Hopefully we can roll this back later ;)

wburns:

danberindei:
@wburns Thinking about this some more, if we ever get a report of a failure in this area, it's not going to be about a `ConcurrentModificationException`, but about an `UnsupportedOperationException` when someone tries to mutate the modifications list, so maybe it would be ok to close the JIRA.

wburns:
Completed the JIRA.

denAbramoff:
i got a weird exception after this change, can you check it? i described it here https://stackoverflow.com/questions/75890762/infinispan-9-4-fails-by-putoperation-in-twophasecoordinator-beforecompletion

PR #3626, opened 2015-08-04
ISPN-5658 Soft Index File Store does not update sequence id in start()

rvansa:
https://issues.jboss.org/browse/ISPN-5658


tristantarrant:
Pushed, thanks

PR #3628, opened 2015-08-04
[7.2.x] ISPN-5650 NPE when trying to execute script from Hot Rod client

gustavonalle:
A side effect here:  ```javax.script.ScriptEngineManager``` logs an error to console (!) when it can't find any engine declared via META-INF/services, so when starting the server a red line is printed. For example, java 8 does not ship with rhino anymore, so:

```12:38:57,101 ERROR [stderr] (MSC service thread 1-4) ScriptEngineManager providers.next(): javax.script.ScriptEngineFactory: Provider com.sun.script.javascript.RhinoScriptEngineFactory not found```

This error is harmless though...


gustavonalle:
@tristantarrant Updated

tristantarrant:
Thanks @gustavonalle

PR #3630, opened 2015-08-04
ISPN-5632 Address rework comments for ISPN-5293

wburns:
https://issues.jboss.org/browse/ISPN-5632

### DIFF
--- a/core/src/main/java/org/infinispan/CacheStream.java
+++ b/core/src/main/java/org/infinispan/CacheStream.java
@@ -7,6 +7,7 @@ import java.util.Spliterator;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.stream.Collector;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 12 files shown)


wburns:
Covered the majority of comments.  The 2 main outstanding ones already have issues logged (primarily replication timeout and cache loader spliterator rewrite).

core/src/main/java/org/infinispan/CacheStream.java
```diff
@@ -190,10 +193,12 @@
 
    /**
     * {@inheritDoc}
-    * <p>This method has special usage when used with a distributed cache backing this set.  This operation will act
-    * as an intermediate iterator operation requiring data be brought locally for proper behavior.  This is
-    * described in more detail in the {@link CacheStream} documentation</p>
-    * <p>This intermediate iterator operation will be performed locally only requiring all elements to be in memory</p>
+    * <p>This method is never invoked on a remote node when used with a distributed cache backing this stream.  This
```
danberindei: The method itself can only be invoked by the user on the local node :)

I'd suggest "This operation is performed entirely on the local node", but then you already say the same thing below.

core/src/main/java/org/infinispan/CacheStream.java
```diff
@@ -244,8 +254,24 @@
...
     * @return the new stream
     */
    @Override
    Stream<R> distinct();
+
+   /**
+    * {@inheritDoc}
+    * <p>Note when using a distributed backing cache for this stream the collector must be marshallable.  This
+    * prevents the usage of {@link java.util.stream.Collectors} class.  However you can use the
+    * {@link org.infinispan.stream.CacheCollectors} static factory methods to create a collector that is created
```
danberindei: The duplicated "created" makes this a bit confusing.

core/src/main/java/org/infinispan/stream/impl/ClusterStreamManager.java
```diff
@@ -23,7 +25,9 @@
     */
    interface ResultsCallback<R> {
       /**
-       * Called back for intermediate data returned from an operation.  This is useful
+       * Called back for intermediate data returned from an operation.  This is useful for operations that utilized
+       * batch fetching such as {@link CacheStream#iterator()}, {@link CacheStream#spliterator()},
+       * {@link CacheStream#forEach(Consumer)} & {@link CacheStream#toArray()}.
```
danberindei: This is HTML, better stick with "and" :)

core/src/main/java/org/infinispan/stream/impl/termop/AbstractForEachOperation.java
```diff
@@ -62,10 +63,10 @@ public boolean lostSegment(boolean stopIfLost) {
       List<CacheEntry<K, K>> collectedValues = new ArrayList(batchSize);
 
       List<V> currentList = new ArrayList<>();
-      Object[] currentKey = new Object[1];
+      ByRef<K> ref = new ByRef<>(null);
```
danberindei: No need to change the name of the variable, I think.

core/src/main/java/org/infinispan/stream/impl/termop/object/FlatMapIteratorOperation.java
```diff
@@ -57,11 +58,10 @@ public boolean lostSegment(boolean stopIfLost) {
 
       List<CacheEntry<K, Collection<V>>> collectedValues = new ArrayList(batchSize);
 
-      List<V>[] currentList = new List[1];
-      currentList[0] = new ArrayList<>();
+      ByRef<List<V>> ref = new ByRef<>(new ArrayList<>());
       Object[] currentKey = new Object[1];
```
danberindei: Could use `ByRef` here, too.

core/src/main/java/org/infinispan/stream/impl/ClusterStreamManager.java
wburns: :) forgot, haha

core/src/main/java/org/infinispan/CacheStream.java
wburns: Yeah unfortunately the collector provided from CacheCollectors contains a serializable supplier that creates a non serializable collector.  Any suggestion for the wording? :)

danberindei: How about "static factory methods to create a serializable wrapper, which then creates the actual collector lazily after being deserialized"?

wburns: Alright udpating, also I will add a see tag to point to CacheCollectors.

core/src/main/java/org/infinispan/CacheStream.java
wburns: Sure let me try to update these to be a bit nicer.

wburns:
Updated and rebased.

core/src/main/java/org/infinispan/CacheStream.java
```diff
@@ -241,13 +248,30 @@
...
     * @return the new stream
     */
    @Override
    Stream<R> distinct();
+
+   /**
+    * {@inheritDoc}
+    * <p>Note when using a distributed backing cache for this stream the collector must be marshallable.  This
+    * prevents the usage of {@link java.util.stream.Collectors} class.  However you can use the
+    * {@link org.infinispan.stream.CacheCollectors} static factory methods to create a serializable wrapper, which then
```
galderz: This is a great enhancement. With this in place, it's easier than ever to do distributed collects, and so I've added this method to Traversable for ease of use.

galderz:
Integrated, thanks @wburns

PR #3631, opened 2015-08-05
[ISPN-5569] NullPointerException on JCache.putAll(ConcurrentHashMap)

ryanemerson:


jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCache.java
```diff
@@ -443,6 +444,30 @@ public Void call() {
       }
    }
 
+   private boolean containsNullKey(Map<? extends K, ? extends V> map) {
+      if (map == null) {
```
wburns: This would never be true due to line 422.

jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCache.java
```diff
@@ -443,6 +444,30 @@ public Void call() {
...
+      try {
+         return map.containsKey(null);
+      } catch (NullPointerException ex) {
+         /* Map doesn't support null keys. */
+         return false;
+      }
+   }
+
+   private boolean containsNullValue(Map<? extends K, ? extends V> map) {
+      if (map == null) {
```
wburns: This would never be true due to line 422.

jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCache.java
```diff
@@ -12,6 +12,7 @@
 import java.util.Set;
 import java.util.concurrent.BrokenBarrierException;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
```
wburns: Unused import

ryanemerson:
@wburns When making the changes that you suggested, I noticed that there were several unused imports.  So I have removed those as well as my erroneous ConcurrentHashMap import.

wburns:
@ryanemerson even better :+1:

pruivo:
Currently, we have 3 places (at least) where we check for ```null``` in maps.

* https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java#L1126
* https://github.com/infinispan/infinispan/blob/master/jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCache.java#L421
* https://github.com/infinispan/infinispan/blob/master/jcache/remote/src/main/java/org/infinispan/jcache/remote/JCache.java#L178

and we probably need to check for the remote cache as well. 

What about to put the assert method in ```InfinispanCollections``` and update everything to invoke it? something like this:
```
   public static void assertNotNullEntries(Map<?, ?> map, String name) {
      Objects.requireNonNull(map, String.format("Map '%s' must be non null.", name));
      boolean containsNull = false;
      try {
         containsNull = map.containsKey(null);
      } catch (NullPointerException ignored) {
         //null is not allowed.
      }
      if (containsNull) {
         throw new NullPointerException(String.format("Map '%s' contains null key.", name));
      }
      try {
         containsNull = map.containsValue(null);
      } catch (NullPointerException ignored) {
         //null is not allowed.
      }
      if (containsNull) {
         throw new NullPointerException(String.format("Map '%s' contains null value.", name));
      }
   }
```

WDYT? @wburns

wburns:
@pruivo To be honest the reason I didn't integrate this earlier is I am not a big fan of using a NPE exception for valid code base flow control.  Generating a stack trace can be quite costly.  However the inverse is to iterate over the entire contents to check for the existence of null.  Both are kinda crappy when used for keys (the former will be faster for larger collections, where as the latter will be faster for smaller).  Values the latter will always be better.  Therefore I am thinking what if we iterate through the entry set and check for a null value for either a key or value.  This should give ~ the same efficiency as a single containsValue call.

``
TO PASTE
``

pruivo:
@wburns makes sense :) In the most common case (i.e. the user never uses null), we have to iterate in containsValues. +1 to your approach

wburns:
@ryanemerson you alright with something like that?

ryanemerson:
@wburns Good idea.  I have made the changes, and included an additional method in InfinispanCollection to test for null values in sets.  I think I managed to remove all of the methods that have become redundant.

commons/src/main/java/org/infinispan/commons/util/InfinispanCollections.java
```diff
@@ -312,4 +324,20 @@ public int size() {
...
+      Objects.requireNonNull(map, () -> "Map '" + name + "' must be non null.");
+      Supplier<String> keySupplier = () -> "Map '" + name + "' contains null key.";
+      Supplier<String> valueSupplier = () -> "Map '" + name + "' contains null value.";
+      map.forEach((k, v) -> {
+         Objects.requireNonNull(k, keySupplier);
+         Objects.requireNonNull(v, valueSupplier);
+      });
+   }
+
+   public static void assertNotNullEntries(Set<?> set, String name) {
```
pruivo: don't wanna be nitpick, but you could replace the ```Set``` by ```Collection``` and make the method generic for everything :)

ryanemerson: Good point, I'll change that now

pruivo:
@ryanemerson @wburns looks good to me.

commons/src/main/java/org/infinispan/commons/util/InfinispanCollections.java
```diff
@@ -312,4 +324,20 @@ public int size() {
...
+      map.forEach((k, v) -> {
+         Objects.requireNonNull(k, keySupplier);
+         Objects.requireNonNull(v, valueSupplier);
+      });
+   }
+
+   public static void assertNotNullEntries(Collection<?> collection, String name) {
+      Objects.requireNonNull(collection, () -> "Collection '" + name + "' must be non null.");
+      Supplier<String> entrySupplier = () -> "Collection '" + name + "' contains null entry.";
+      collection.forEach((k) -> Objects.requireNonNull(k, entrySupplier));
```
wburns: Nitpick here, but parens aren't required around the k.

jcache/remote/src/main/java/org/infinispan/jcache/remote/JCache.java
```diff
@@ -428,61 +428,6 @@ protected Object getCacheStatisticsMXBean() {
       return this;
    }
 
-   private JCache<K, V> checkNotNullOrNullKV(Map<? extends K, ? extends V> map, String name) {
-      checkNotNull((Object) map, name);
-      if (containsNullKey(map)) {
-         throw log.parameterMustNotContainNullKeys(name);
```
wburns: I wonder if we need internationalization for this still, cause unfortunately we lose that with the new approach.  @galderz can you comment on that?

jcache/commons/src/test/java/org/infinispan/jcache/AbstractTwoCachesBasicOpsTest.java
```diff
@@ -107,6 +108,21 @@ public void testPutAll(Method m) {
    }
 
    @Test
+   public void testPutAllMapNullValuesNotAllowed(Method m) {
```
wburns: Also I think it would be great to add 2 simple tests that test to verify NPE is thrown if a null key or value is provided in the map.

wburns:
@ryanemerson looks great!  Just have a couple very minor things and wanted to get clarification on something first.

jcache/remote/src/main/java/org/infinispan/jcache/remote/JCache.java
galderz: Internationalization helps with the messages, and the messages can help distinguish which field is null. Since `Objects.requireNonNull` is called in the same statement, I don't know if a stacktrace would be enough to figure out which field is null. If stacktrace is not enough, I think we need an internationalized solution here.

galderz:
Needs rebasing too.

ryanemerson:
@wburns @galderz Rebased and suggested changes made. I wasn't sure about the localisation of the logs, so I haven't changed anything in that regard.  Let me know if further changes are required.

jcache/remote/src/main/java/org/infinispan/jcache/remote/JCache.java
wburns: Unfortunately the previous way never explicitly said which was null either.  It only distinguished between key or value which we still do.  I wonder is this enough still?

galderz:
Integrated, thanks @ryanemerson :)

PR #3633, opened 2015-08-05
[ISPN-5517] JDBCBinaryStore size method should query count directly i…

ryanemerson:
…nstead of iterating


wburns:
One problem with this is the fact that the other count would properly exclude expired entries.  This is one reason I have found to really hate the BasicCache API that allows for expiration as it means everywhere has to **always** pay attention to expired entries.

We could tweak this to do a count of all buckets that aren't expired and then manually count the buckets that could contain entries that expire.  But it would require a bit more work.

wburns:
@ryanemerson if you want I can close this out until someone can revisit it later?

ryanemerson:
@wburns It might be for the best.  I'll have another look at it if I get a chance, but no guarantees.

wburns:
Closing until we can do some more work here.

wburns:
Thanks @ryanemerson for the initial PR though :+1:

PR #3637, opened 2015-08-06
revert ISPN-5601

rvansa:
To keep binary compatibility between 7.x and 8.x


wburns:
Updated commit message and integrated, thanks @rvansa !

PR #3638, opened 2015-08-06
ISPN-5541 Add provided scope to javax api dependencies

dmstahl:
https://issues.jboss.org/browse/ISPN-5541

### DIFF
--- a/all/embedded/pom.xml
+++ b/all/embedded/pom.xml
@@ -95,6 +95,7 @@
       <dependency>
          <groupId>org.jboss.spec.javax.transaction</groupId>
          <artifactId>jboss-transaction-api_1.1_spec</artifactId>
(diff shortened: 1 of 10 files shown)


galderz:
If you're making javax.cache-api provided in ```jcache/remote```, shouldn't you make the same change in ```jcache/embedded``` ?

dmstahl:
The build fails if the change is made in jcache/embedded.  It looks like the problem is related to the       <module.skipComponentMetaDataProcessing>false</module.skipComponentMetaDataProcessing> property being set.  I'm not real sure what the best way to proceed is - either remove that property or leave javax.cache-api as a compile dependency?

dmstahl:
Updated

galderz:
Integrated, thanks @dmstahl

PR #3639, opened 2015-08-10
ISPN-5343 Resizing of BoundedEquivalentConcurrentHashMapV8

slaskawi:
Hey!

This is the first part of https://issues.jboss.org/browse/ISPN-5340 implementation. Please do not close this ticket after integrating this.

It allows `BoundedEquivalentConcurrentHashMapV8` dynamic resizing. Later on, this functionality will be invoked from JMX bean.

Related Design doc: https://github.com/infinispan/infinispan/wiki/Dynamic-JMX-exposer-for-Configuration

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8.java
+++ b/commons/src/main/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8.java
@@ -13,9 +13,6 @@
 
 package org.infinispan.commons.util.concurrent.jdk8backported;
 
-import static java.util.Collections.singletonMap;
-import static java.util.Collections.unmodifiableMap;
-
 import org.infinispan.commons.equivalence.AnyEquivalence;
 import org.infinispan.commons.equivalence.Equivalence;
 import org.infinispan.commons.logging.Log;
@@ -34,12 +31,15 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CountedCompleter;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.LockSupport;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.*;
+
+import static java.util.Collections.singletonMap;
+import static java.util.Collections.unmodifiableMap;
 
 /**
  * A hash table supporting full concurrency of retrievals and
@@ -440,6 +440,13 @@ public class BoundedEquivalentConcurrentHashMapV8<K,V> extends AbstractMap<K,V>
       Collection<Node<K, V>> findIfEntriesNeedEvicting();
 
       void onResize(long oldSize, long newSize);
+
+      /**
+       * Invoked when resizing the container.
+       *
+       * @param newSize New Size applied to the container.
+       */
+      void resize(long newSize);
    }
 
    static class NullEvictionPolicy<K, V> implements EvictionPolicy<K, V> {
@@ -491,6 +498,11 @@ public class BoundedEquivalentConcurrentHashMapV8<K,V> extends AbstractMap<K,V>
       public void onResize(long oldSize, long newSize) {
          // Do nothing.
       }
+
+      @Override
+      public void resize(long newSize) {
+         // Do nothing.
+      }
    }
 
    static class LRUNode<K, V> implements EvictionEntry<K, V> {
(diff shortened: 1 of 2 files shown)


commons/src/main/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8.java
```diff
@@ -513,7 +525,7 @@ public K getKey() {
       final BoundedEquivalentConcurrentHashMapV8<K, V> map;
       final StrippedConcurrentLinkedDeque<Node<K, V>> deque = 
             new StrippedConcurrentLinkedDeque<Node<K,V>>();
-      final long maxSize;
+      transient long maxSize;
```
wburns: Seems you wanted volatile instead.

core/src/test/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8BaseTest.java
```diff
@@ -32,6 +9,13 @@
 import org.infinispan.util.logging.LogFactory;
 import org.testng.annotations.Test;
 
+import java.util.Arrays;
+import java.util.Map;
```
slaskawi: I reimported IDE settings just to make sure if this rearangement of imports is legal. According to my IntelliJ this is how it should look like...

commons/src/main/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8.java
slaskawi: Of course.... ok, I get it... going for coffee... :)

core/src/test/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8BaseTest.java
```diff
@@ -64,6 +48,22 @@ public void testByteArrayPutIfAbsentFail() {
       byteArrayPutIfAbsentFail(createComparingConcurrentMap(), true);
    }
 
+   public void testSettingMaxSize() {
+      //given
+      BoundedEquivalentConcurrentHashMapV8<String, String> map = createMap(3, Eviction.LIRS);
```
wburns: You should use ```createMap(3, evictionPolicy());``` so this way it has coverage for all of various eviction policies.

core/src/test/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8BaseTest.java
wburns: Yeah I believe I wrote this class using Eclipse? Either way Eclipse and IntelliJ like to put them in different order.

core/src/test/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8BaseTest.java
slaskawi: Good point! Let me update it.

slaskawi:
Included @wburns's commit (sorry I had to amend it and include JIRA ID) and rebased.

commons/src/main/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8.java
```diff
@@ -1413,7 +1434,7 @@ public void onEntryRemove(Node<K, V> e) {
                      SizeAndEvicting sizeEvict = currentSize.get();
                      // If the size was changed behind our back by a remove we
                      // need to detect that
-                     if (sizeEvict.size - sizeEvict.evicting < maximumSize) {
+                     if (sizeEvict.size - sizeEvict.evicting <= maximumSize) {
```
wburns: Hrmm I am not sure we want this change.  This would make it so we would not evict a value when we have an exact size match.

Scenario:

maximumSize = 100

sizeEvict.size = 110
sizeEvict.evicting = 10

In this case they would equal as they should because we want to evict 10 entries.  In this case though we would only evict 9 since we decrement that value.

slaskawi: Ok, I'll fix it and squash the commits.

slaskawi: Haha I missed it because the tests were passing :D

slaskawi:
@wburns Ok, done. Thanks for finding this!

wburns:
Pulling...

wburns:
Integrated into master, thanks @slaskawi !

PR #3641, opened 2015-08-12
ISPN-5607 Preemptively invalidate near cache after writes

galderz:
https://issues.jboss.org/browse/ISPN-5607 

[Master ONLY!]

* Near cache stale reads can't be fully guaranteed since there's always
  the possibility of a read to come in when the server has already
  executed a write operation but it's in process of sending back the
  response.
* However, we can't at least guarantee that within a single
  thread, a read after a write will read the written value. The change
  proposed in this PR addresses this, by preemptively invalidating data
  after a successful write.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/InvalidatedNearRemoteCache.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/InvalidatedNearRemoteCache.java
@@ -1,0 +1,111 @@
+package org.infinispan.client.hotrod.impl;
+
+import org.infinispan.client.hotrod.Flag;
+import org.infinispan.client.hotrod.RemoteCacheManager;
+import org.infinispan.client.hotrod.VersionedValue;
+import org.infinispan.client.hotrod.near.NearCacheService;
+
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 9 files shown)


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/InvalidatedNearRemoteCache.java
```diff
@@ -0,0 +1,109 @@
...
+
+   @Override
+   public void clear() {
+      super.clear();
+      nearcache.clear(); // Clear near cache too
+   }
+
+   @SuppressWarnings("unchecked")
+   void invalidateNearCacheIfNeeded(Object key, Object prev) {
+      if (!operationsFactory.hasFlag(Flag.FORCE_RETURN_VALUE) || prev != null)
```
wburns: Looks like this one needs updating like 7.2.x

galderz: Yes it does, I'll do that right away.

galderz:
@wburns This should be fixed now

gustavonalle:
pulling

gustavonalle:
merged, thanks @galderz

PR #3642, opened 2015-08-12
ISPN-5607 Preemptively invalidate near cache after writes [7.2.x]

galderz:
https://issues.jboss.org/browse/ISPN-5607

[7.2.x ONLY]

* Near cache stale reads can't be fully guaranteed since there's always
  the possibility of a read to come in when the server has already
  executed a write operation but it's in process of sending back the
  response.
* However, we can't at least guarantee that within a single
  thread, a read after a write will read the written value. The change
  proposed in this PR addresses this, by preemptively invalidating data
  after a successful write.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/EagerNearRemoteCache.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/EagerNearRemoteCache.java
@@ -11,11 +11,12 @@ import org.infinispan.client.hotrod.near.NearCacheService;
  * @param <K>
  * @param <V>
  */
-public class NearRemoteCache<K, V> extends RemoteCacheImpl<K, V> {
+@Deprecated
+public class EagerNearRemoteCache<K, V> extends RemoteCacheImpl<K, V> {
 
    private final NearCacheService<K, V> nearcache;
 
-   public NearRemoteCache(RemoteCacheManager rcm, String name, NearCacheService<K, V> nearcache) {
+   public EagerNearRemoteCache(RemoteCacheManager rcm, String name, NearCacheService<K, V> nearcache) {
       super(rcm, name);
       this.nearcache = nearcache;
    }
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/InvalidatedNearRemoteCache.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/InvalidatedNearRemoteCache.java
@@ -1,0 +1,111 @@
+package org.infinispan.client.hotrod.impl;
+
+import org.infinispan.client.hotrod.Flag;
+import org.infinispan.client.hotrod.RemoteCacheManager;
+import org.infinispan.client.hotrod.VersionedValue;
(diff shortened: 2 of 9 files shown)


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/EagerNearRemoteCache.java
```diff
@@ -11,11 +11,12 @@
  * @param <K>
  * @param <V>
  */
-public class NearRemoteCache<K, V> extends RemoteCacheImpl<K, V> {
+@Deprecated
+public class EagerNearRemoteCache<K, V> extends RemoteCacheImpl<K, V> {
```
wburns: Do we have to worry about people referencing this class?  I am guessing not, but if so we can't really change the name.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/InvalidatedNearRemoteCache.java
```diff
@@ -0,0 +1,109 @@
...
+
+   @Override
+   public void clear() {
+      super.clear();
+      nearcache.clear(); // Clear near cache too
+   }
+
+   @SuppressWarnings("unchecked")
+   void invalidateNearCacheIfNeeded(Object key, Object prev) {
+      if (!operationsFactory.hasFlag(Flag.FORCE_RETURN_VALUE) || prev != null)
```
wburns: I am thinking this can just be ```if (prev != null)``` since not having force return value implies it will be null right?  This way we don't have the thread local lookup.  Also will the flag even be set at this point?  Isn't is removed in the call to super.remove etc.?

wburns: Actually that isn't true for a regular put.  However the issue with the flag being set may still be a problem.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/EagerNearRemoteCache.java
tristantarrant: This is in the impl package, so no worries imho

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/InvalidatedNearRemoteCache.java
galderz: Yeah, that's true. In RemoteCache once you retrieve the flag is gone. I'll remove that.

galderz: It can't just be `prev != null` because a remove might return null even though it's removed things, but the only way you really know if the removed worked is based on parameter return. So, you either check the flag and then check the previous value, or you do it regardless of the return value...

wburns:
Seems good, pulling..

wburns:
Integrated into 7.2.x, thanks @galderz !

PR #3648, opened 2015-08-13
[ISPN-5517] JDBCBinaryStore size method should query count directly i…

ryanemerson:
…nstead of iterating

### DIFF
--- a/persistence/jdbc/src/test/java/org/infinispan/persistence/jdbc/binary/JdbcBinaryStoreTest.java
+++ b/persistence/jdbc/src/test/java/org/infinispan/persistence/jdbc/binary/JdbcBinaryStoreTest.java
@@ -17,6 +17,7 @@ import java.io.Serializable;
 
 import static org.infinispan.test.TestingUtil.metadata;
 import static org.mockito.Mockito.mock;
+import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertNull;
 
 /**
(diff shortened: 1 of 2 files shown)


ryanemerson:
@wburns I think this addresses the problems with my original pull request.  The size() method now only counts non-expired entries.  

A crude performance comparison between the old and new size() method can be found at:
https://github.com/ryanemerson/infinispan/commit/0e41b28fdaea8fd6b51c74f80682f655592b029f

galderz:
@ryanemerson What are the performance comparison results in your environment?

Looks good to me, @wburns any more thoughts?

ryanemerson:
@galderz Results of a single execution.  
When a cache contains 240 entries, none of which have expired, the old method takes 7950521ns whereas the new takes 765363ns. 

Once we have moved time forward and half of the entries in the cache have expired, the old method takes 579729ns and the new method 184852ns. 

Throughout testing I never encountered an instance where the old implementation outperformed the new one, however my test case is hardly exhaustive.

gustavonalle:
pulling

gustavonalle:
merged, thanks @ryanemerson !

persistence/jdbc/src/test/java/org/infinispan/persistence/jdbc/binary/JdbcBinaryStoreTest.java
```diff
@@ -98,6 +99,18 @@ public void testPurgeExpiredAllCodepaths() throws Exception {
       UnitTestDatabaseManager.verifyConnectionLeaks(((JdbcBinaryStore) cl).getConnectionFactory());
    }
 
+   public void testCacheSize() {
+      long lifespan = 3000;
+      for (int i = 0; i < 120; i++) {
+         cl.write(marshalledEntry(internalCacheEntry("mortal" + i, "v1", lifespan)));
+         cl.write(marshalledEntry(internalCacheEntry("immortal" + i, "v1", -1)));
+      }
```
danberindei: @ryanemerson The test fails if I populate the entries like this:

      for (int i = 0; i < 240; i++) {
         if (i % 2 == 0) {
            cl.write(marshalledEntry(internalCacheEntry("key" + i, "v1", lifespan)));
         } else {
            cl.write(marshalledEntry(internalCacheEntry("key" + i, "v1", -1)));
         }
      }

Your test works because the mortal and immortal keys are hashed to different buckets, but if you mix mortal and immortal entries in the same bucket your count will be wrong. The same happens with different lifespans in the same bucket.

PR #3650, opened 2015-08-13
ISPN-5677 RemoteCache async methods use flags

galderz:
* They were previously being ignored as a result of not sending from the
  main thread to the thread executing the async operation.

https://issues.jboss.org/browse/ISPN-5677


wburns:
It seems that we don't clear out the flags after we submit the async operation.  I am thinking we should clear those out just like we do for non async to keep it consistent.

galderz:
@wburns Right, we should call `flags()` instead of `getFlags` so that they're cleared.

galderz:
@wburns Fixed

gustavonalle:
Pulling....

gustavonalle:
Merged, thanks @galderz!

PR #3654, opened 2015-08-17
ISPN-5680 Fix NPE in tests and remove unused imports

vjuranek:
https://issues.jboss.org/browse/ISPN-5680

### DIFF
--- a/core/src/main/java/org/infinispan/metadata/EmbeddedMetadata.java
+++ b/core/src/main/java/org/infinispan/metadata/EmbeddedMetadata.java
@@ -160,8 +160,8 @@ public class EmbeddedMetadata implements Metadata {
             long lifespan, TimeUnit lifespanUnit,
             long maxIdle, TimeUnit maxIdleUnit, EntryVersion version) {
          super(version);
-         this.lifespan = lifespanUnit.toMillis(lifespan);
-         this.maxIdle = maxIdleUnit.toMillis(maxIdle);
+         this.lifespan = lifespan < 0 ? -1 : lifespanUnit.toMillis(lifespan);
+         this.maxIdle = maxIdle < 0 ? -1 : maxIdleUnit.toMillis(maxIdle);
       }
 
       @Override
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
```diff
@@ -49,7 +49,7 @@ public void testSimpleExpirationLifespan() throws Exception {
    public void testSimpleExpirationMaxIdle() throws Exception {
 
       for (int i = 0; i < SIZE; i++) {
-         cache.put("key-" + i, "value-" + i,-1, null, 1, TimeUnit.MILLISECONDS);
+         cache.put("key-" + i, "value-" + i,-1, TimeUnit.MILLISECONDS, 1, TimeUnit.MILLISECONDS);
```
anistor: I believe the problem should be addressed in the Metadata implementation. Negative time values should not be converted, so the given TimeUnit could be null and not result in an exception.

vjuranek:
@anistor ah, right. PR updated

core/src/main/java/org/infinispan/metadata/EmbeddedMetadata.java
```diff
@@ -160,7 +160,7 @@ private EmbeddedExpirableMetadata(
             long lifespan, TimeUnit lifespanUnit,
             long maxIdle, TimeUnit maxIdleUnit, EntryVersion version) {
          super(version);
-         this.lifespan = lifespanUnit.toMillis(lifespan);
+         this.lifespan = lifespan < 0 ? lifespan : lifespanUnit.toMillis(lifespan);
          this.maxIdle = maxIdleUnit.toMillis(maxIdle);
```
wburns: We should handle maxIdle as well

anistor: Also, I believe negative values should be replaced with -1. This 'normalization' would allow equals () to work properly.

wburns: @anistor Makes sense :+1:

anistor: @vjuranek , please note there is another potential NPE in the toMillis() conversion in AbstractEmbeddedTimeoutMetadata.

vjuranek:
@anistor , @wburns : thanks, PR updated

anistor:
Looks good to me

anistor:
Pulling ...

anistor:
Integrated in master. Thanks @vjuranek !

PR #3656, opened 2015-08-20
ISPN-5553 Addressing feedback from Functional API PR for Beta3

galderz:
* Traversable simplified to have two types of operations only, intermediate and terminal. Feedback from @danberindei in last PR.
* ReadWrite operations now fire WriteListeners too. Feedback from @danberindei in last PR.
* Wrap entry views returned from lambdas in such way that they do not force existing MVCCEntry instances to be serialized, and make sure entry views returned cannot be used to mutate things. Feedback from @danberindei in last PR.
* Address lambda retry situations. Added tests and verified that they work though they incur in extra cost of shipping around previous metadata and value in the command. Currently valueMatcher defined in `@SerializeWith` annotation but work is in progress to fix this but will probably not be ready for CR1.
* MetaParam.Lookup.getMetaParam is gone since we can't have 100% guarantee that a meta param will be present (e.g. when running in compatibility mode. Based on feedback from @danberindei on last PR.
* Functional listeners are now post-event. Based on feedback from @wburns on last PR.
* Get rid of WaitMode param and have a dedicated param for how the CompletableFutures are constructed called FutureMode. More Params to be added in forthcoming PR but not ready for CR1.
* WriteOnlyMap multi-key operations now return CompletableFuture<Void> instead of CloseableIterator<Void>, since the iteration over empty results adds nothing for the user, and complicates things when it comes to retries.


commons/src/main/java/org/infinispan/commons/api/functional/FunctionalMap.java
```diff
@@ -345,20 +340,21 @@
        * @param f operation that consumes a value associated with a key in the
        *          entries collection and the {@link WriteEntryView} associated
        *          with that key in the cache
-       * @return a {@link CloseableIterator} that can be navigated to follow progress
-       *         as operations get invoked
+       * @return a {@link CompletableFuture} which will be completed the
```
tristantarrant: s/the/when/

commons/src/main/java/org/infinispan/commons/api/functional/Listeners.java
```diff
@@ -30,31 +30,21 @@ private Listeners() {
    }
 
    /**
-    * Read-write listeners enables user to register listeners for events
-    * happening in the read-write functional map.
+    * Read-write listeners enable user to register listeners for cache
```
tristantarrant: s/user/users/

commons/src/main/java/org/infinispan/commons/api/functional/MetaParam.java
```diff
@@ -44,31 +39,47 @@
...
+    * <pre>{@code
+    *     Optional<MetaLifespan<Long>> metaLifespan =
+    *          metaParamLookup.findMetaParam(MetaLifespan.class);
+    * }</pre>
+    *
+    * @apiNote A user that queries meta parameters can never assume that the
+    * meta parameter will always exist because there are scenarios, such as
+    * when compatibility mode is enabled, when meta parameters that are
+    * assumed to be present are not due to the API multiplexing that occurs.
+    * For example, when compatibility mode is enabled, REST server can't assume
```
tristantarrant: the REST server

commons/src/main/java/org/infinispan/commons/api/functional/Traversable.java
```diff
@@ -113,26 +64,27 @@
...
     */
    Optional<T> reduce(BinaryOperator<T> folder);
 
    /**
-    * Applies a binary folding operation to a start value and the result
-    * of each element having a mapping function applied.
+    * A terminal operation that applies a binary folding operation to a start
+    * value and the result of each element having a mapping function applied.
     *
     * @apiNote This is a map and reduce in one go, which could potentially be
```
tristantarrant: "This is a single-pass map/reduce"

galderz:
@tristantarrant All fixed

tristantarrant:
Pulling

tristantarrant:
Merged, thanks!

rvansa:
@galderz Hi Galder, I've noticed that in 6e6747dca76f3ea9d6a303cccbb2c11be6a3a923 you've added locking only for the single-key commands; was omitting *Many* commands intentional?

galderz:
@rvansa No, probably an oversight.

PR #3657, opened 2015-08-20
ISPN-5687 Server thread pools

tristantarrant:
https://issues.jboss.org/browse/ISPN-5687

Inline the thread pool configuration into the infinispan subsystem since WildFly 9 has deprecated the threads subsytem.

I apologize for the large number of changes, but these are necessary to minimize alignment with the equivalent subsystem in WF.
Also bump the WildFly used for integration tests to 10.0.0.Beta1


tristantarrant:
None of the failures are related to this, yay !

gustavonalle:
merged, tks @tristantarrant !

PR #3660, opened 2015-08-21
[backport] ISPN-4046 Support indexing and searching of null non-string properties

anistor:
jira: https://issues.jboss.org/browse/ISPN-4046

This is for 7.2.x!


anistor:
This is 'preview' only for now because a new HQL parser and hibernate-search need to be released first.

danberindei:
Integrated, thanks Adrian!

PR #3661, opened 2015-08-21
ISPN-5677 RemoteCache async methods use flags [7.2.x]

galderz:
* They were previously being ignored as a result of not sending from the
  main thread to the thread executing the async operation.


gustavonalle:
testing...

gustavonalle:
integrated, thanks @galderz !

PR #3665, opened 2015-08-21
ISPN-5692 Registering a remote event listener fails if cache is index…

anistor:
…ed, uses protobuf marshalling and includeInitialState=true

Jira: https://issues.jboss.org/browse/ISPN-5692


gustavonalle:
integrated, thanks @anistor

PR #3666, opened 2015-08-21
ISPN-5691 Automatically enable write-skew checks when using REPEATABLE_READ isolation and optimistic locking

tristantarrant:
https://issues.jboss.org/browse/ISPN-5691


galderz:

galderz:
Integrated in master but not 7.2.x since it does not apply cleanly. A PR is needed for 7.2.x.

PR #3669, opened 2015-08-24
ISPN-5688 Add Functional Map simple tutorial as a test

galderz:
https://issues.jboss.org/browse/ISPN-5688

### DIFF
--- a/core/src/test/java/org/infinispan/functional/FunctionalSimpleTutorialTest.java
+++ b/core/src/test/java/org/infinispan/functional/FunctionalSimpleTutorialTest.java
@@ -1,0 +1,87 @@
+package org.infinispan.functional;
+
+import org.infinispan.AdvancedCache;
+import org.infinispan.commons.api.functional.EntryView.ReadEntryView;
+import org.infinispan.commons.api.functional.FunctionalMap.ReadOnlyMap;
+import org.infinispan.commons.api.functional.FunctionalMap.ReadWriteMap;
+import org.infinispan.commons.api.functional.FunctionalMap.WriteOnlyMap;
+import org.infinispan.commons.api.functional.MetaParam.MetaLifespan;
+import org.infinispan.commons.api.functional.Traversable;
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/infinispan/functional/FunctionalSimpleTutorialTest.java
```diff
@@ -0,0 +1,83 @@
...
+         (v, writeView) -> writeView.set(v));
+
+      // When each write-only operation completes, execute a read-only operation to retrieve the value
+      CompletableFuture<String> readFuture1 =
+         writeFuture1.thenCompose(r -> readOnlyMap.eval("key1", ReadEntryView::get));
+      CompletableFuture<String> readFuture2 =
+         writeFuture2.thenCompose(r -> readOnlyMap.eval("key2", ReadEntryView::get));
+
+      // When the read-only operation completes, print it out
+      System.out.printf("Created entries: %n");
```
danberindei: We should use logging in the tests.

galderz: True, but I've tried to keep it as close as possible to the simple tutorial source code [here](https://github.com/infinispan/infinispan-simple-tutorials/blob/master/functional/src/main/java/org/infinispan/tutorial/simple/functional/InfinispanFunctional.java), hence why I left it printing out stuff to system out. The amount logged is minimal so would not really cause many issues.

danberindei: I have a script that collapses consecutive "Test suite progress:" lines in the console, so I guess it's more visible for me... but I don't think having the logging different in the test would cause many issues either :)

galderz:
@danberindei You win, updated :)

danberindei:
Thanks Galder, pulling...

danberindei:
Integrated!

PR #3672, opened 2015-08-25
Ispn 5679

wburns:
https://issues.jboss.org/browse/ISPN-5679
https://issues.jboss.org/browse/ISPN-5507

Note ISPN-5507 is not to be completed.  This is just to workaround an issue where the tests could hang forever since a tx is completed asynchronously from the main test thread.

### DIFF
--- a/core/src/main/java/org/infinispan/AdvancedCache.java
+++ b/core/src/main/java/org/infinispan/AdvancedCache.java
@@ -551,4 +551,20 @@ public interface AdvancedCache<K, V> extends Cache<K, V> {
     * @return the entry set containing all of the CacheEntries
     */
    CacheSet<CacheEntry<K, V>> cacheEntrySet();
+
+   /**
+    * Attempts to remove the entry if it is expired.  Due to expired entries not being consistent across nodes, this
+    * will still attempt to remove the value if it is not present.  Note that this will raise an expired event
+    * even if the entry is not present.  Normally this method should never be invoked except by the
+    * {@link ExpirationManager}.
+    * <p>
+    * This command will only remove the value if the value and lifespan also match if provided.
+    * <p>
+    * NOTE: This method may be removed at any point including in a minor release and is not supported for external
+    * usage.
+    * @param key the key that is expiring
+    * @param value the value that mapped to the given.  Null means it will match any value
+    * @param lifespan the lifespan that should match.  If null is provided it will match any lifespan value
+    */
+   void removeExpired(K key, V value, Long lifespan);
 }
--- a/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
+++ b/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
@@ -22,6 +22,7 @@ import org.infinispan.commands.write.EvictCommand;
 import org.infinispan.commands.write.PutKeyValueCommand;
 import org.infinispan.commands.write.PutMapCommand;
 import org.infinispan.commands.write.RemoveCommand;
+import org.infinispan.commands.write.RemoveExpiredCommand;
 import org.infinispan.commands.write.ReplaceCommand;
 import org.infinispan.commands.write.ValueMatcher;
 import org.infinispan.commons.CacheConfigurationException;
@@ -557,6 +558,14 @@ public class CacheImpl<K, V> implements AdvancedCache<K, V> {
       return (V) executeCommandAndCommitIfNeeded(ctx, command);
    }
 
+   @Override
+   public void removeExpired(K key, V value, Long lifespan) {
+      InvocationContext ctx = getInvocationContextWithImplicitTransaction(false, null, 1);
+      RemoveExpiredCommand command = commandsFactory.buildRemoveExpiredCommand(key, value, lifespan);
+      ctx.setLockOwner(command.getKeyLockOwner());
+      // Send an expired remove command to everyone
+      executeCommandAndCommitIfNeeded(ctx, command);
+   }
 
    @ManagedOperation(
          description = "Clears the cache",
--- a/core/src/main/java/org/infinispan/commands/CommandsFactory.java
+++ b/core/src/main/java/org/infinispan/commands/CommandsFactory.java
@@ -125,6 +125,15 @@ public interface CommandsFactory {
    InvalidateCommand buildInvalidateFromL1Command(Address origin, Set<Flag> flags, Collection<Object> keys);
 
    /**
+    * Builds an expired remove command that is used to remove only a specific expired entry
+    * @param key the key of the expired entry
+    * @param value the value of the entry when it was expired
+    * @param lifespan the lifespan that expired from the command
+    * @return a RemovedExpiredCommand
+    */
+   RemoveExpiredCommand buildRemoveExpiredCommand(Object key, Object value, Long lifespan);
+
+   /**
     * Builds a ReplaceCommand
     * @param key key to replace
     * @param oldValue existing value to check for if conditional, null if unconditional.
--- a/core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
+++ b/core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
@@ -1,0 +1,181 @@
+package org.infinispan.commands.write;
+
+import org.infinispan.commands.CommandInvocationId;
+import org.infinispan.commands.Visitor;
+import org.infinispan.commons.equivalence.Equivalence;
+import org.infinispan.commons.util.CollectionFactory;
+import org.infinispan.commons.util.Util;
+import org.infinispan.configuration.cache.Configuration;
+import org.infinispan.container.entries.CacheEntry;
+import org.infinispan.container.entries.ExpiryHelper;
+import org.infinispan.container.entries.MVCCEntry;
+import org.infinispan.context.Flag;
+import org.infinispan.context.InvocationContext;
+import org.infinispan.lifecycle.ComponentStatus;
+import org.infinispan.metadata.Metadata;
+import org.infinispan.notifications.cachelistener.CacheNotifier;
+import org.infinispan.util.TimeService;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.EnumSet;
+import java.util.Objects;
+import java.util.Set;
+
+
+/**
+ * Removes an entry that is expired from memory
+ *
+ * @author William Burns
+ * @since 8.0
+ */
+public class RemoveExpiredCommand extends RemoveCommand {
+   public static final int COMMAND_ID = 58;
+   private static final Log log = LogFactory.getLog(RemoveExpiredCommand.class);
+   private static final boolean trace = log.isTraceEnabled();
+
+   protected Long lifespan;
+   protected TimeService timeService;
+
+   public RemoveExpiredCommand() {
+      // The value matcher will always be the same, so we don't need to serialize it like we do for the other commands
+      this.valueMatcher = ValueMatcher.MATCH_EXPECTED_OR_NULL;
+   }
+
+   public RemoveExpiredCommand(Object key, Object value, Long lifespan, CacheNotifier notifier,
+           Equivalence valueEquivalence, TimeService timeService, CommandInvocationId commandInvocationId) {
+      //valueEquivalence can be null because this command never compares values.
+      super(key, value, notifier, null, valueEquivalence, commandInvocationId);
+      this.lifespan = lifespan;
+      this.timeService = timeService;
+      this.valueMatcher = ValueMatcher.MATCH_EXPECTED_OR_NULL;
+   }
+
+   public void init(CacheNotifier notifier, Configuration configuration, TimeService timeService) {
+      super.init(notifier, configuration);
+      this.timeService = timeService;
+   }
+
+   /**
+    * Performs an expiration on a specified entry
+    *
+    * @param ctx invocation context
+    * @return null
+    */
+   @Override
+   public Object perform(InvocationContext ctx) throws Throwable {
+      MVCCEntry e = (MVCCEntry) ctx.lookupEntry(key);
+      if (e != null && !e.isRemoved()) {
+         Object value = e.getValue();
+         // If the provided lifespan is null, that means it is a store removal command, so we can't compare lifespan
+         if (lifespan == null) {
+            if (valueMatcher.matches(e, value, e.getValue(), valueEquivalence)) {
+               e.setExpired(true);
+               return performRemove(e, ctx);
+            }
+         } else if (e.getMetadata() == null) {
+            // If there is no metadata and no value that means it is gone currently or not shown due to expired
+            // If we have a value though we should verify it matches the value as well
+            if (value == null || valueMatcher.matches(e, value, e.getValue(), valueEquivalence)) {
+               e.setExpired(true);
+               return performRemove(e, ctx);
+            }
+         } else if (e.getLifespan() > 0 && e.getLifespan() == lifespan) {
+            // If the entries lifespan is not positive that means it can't expire so don't even try to remove it
+            // Lastly if there is metadata we have to verify it equals our lifespan and the value match.
+            // TODO: add a threshold to verify this wasn't just created with the same value/lifespan just before expiring
+            if (valueMatcher.matches(e, value, e.getValue(), valueEquivalence)) {
+               e.setExpired(true);
+               return performRemove(e, ctx);
+            }
+         } else {
+            log.trace("Cannot remove entry as its lifespan or value do not match");
+         }
+      } else {
+         log.trace("Nothing to remove since the entry doesn't exist in the context or it is already removed");
+      }
+      successful = false;
+      return false;
+   }
+
+   @Override
+   public boolean isConditional() {
+      return true;
+   }
+
+   @Override
+   public void notify(InvocationContext ctx, Object removedValue, Metadata removedMetadata,
+         boolean isPre) {
+      if (!isPre) {
+         notifier.notifyCacheEntryExpired(key, value, removedMetadata, ctx);
+      }
+   }
+
+   @Override
+   public byte getCommandId() {
+      return COMMAND_ID;
+   }
+
+   @Override
+   public String toString() {
+      return "RemoveExpiredCommand{" +
+              "key=" + key +
+              "value=" + value +
+              "lifespan=" + lifespan +
+              '}';
+   }
+
+   @Override
+   public Object[] getParameters() {
+      return new Object[]{commandInvocationId, key, value, lifespan};
+   }
+
+   @Override
+   public void setParameters(int commandId, Object[] args) {
+      if (commandId != COMMAND_ID) throw new IllegalStateException("Invalid method id");
+      int i = 0;
+      commandInvocationId = (CommandInvocationId) args[i++];
+      key = args[i++];
+      value = args[i++];
+      lifespan = (long) args[i++];
+   }
+
+   @Override
+   public boolean ignoreCommandOnStatus(ComponentStatus status) {
+      switch (status) {
+         case FAILED:
+         case STOPPING:
+         case TERMINATED:
+            return true;
+         default:
+            return false;
+         }
+   }
+
+   @Override
+   public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+      if (!super.equals(o)) return false;
+      RemoveExpiredCommand that = (RemoveExpiredCommand) o;
+      return Objects.equals(lifespan, that.lifespan);
+   }
+
+   @Override
+   public int hashCode() {
+      return Objects.hash(super.hashCode(), lifespan);
+   }
+
+   @Override
+   public Set<Flag> getFlags() {
+      return EnumSet.of(Flag.SKIP_CACHE_LOAD);
+   }
+
+   @Override
+   public boolean hasFlag(Flag flag) {
+      // We skip cache load, since if the entry is not in memory then it wasn't updated since it last expired
+      return flag == Flag.SKIP_CACHE_LOAD;
+   }
+}
--- a/core/src/main/java/org/infinispan/container/entries/ReadCommittedEntry.java
+++ b/core/src/main/java/org/infinispan/container/entries/ReadCommittedEntry.java
@@ -48,13 +48,13 @@ public class ReadCommittedEntry implements MVCCEntry {
    // if this or any MVCC entry implementation ever needs to store a boolean, always use a flag instead.  This is far
    // more space-efficient.  Note that this value will be stored in a byte, which means up to 8 flags can be stored in
    // a single byte.  Always start shifting with 0, the last shift cannot be greater than 7.
-   protected static enum Flags {
+   protected enum Flags {
       CHANGED(1), // same as 1 << 0
       CREATED(1 << 1),
       REMOVED(1 << 2),
       VALID(1 << 3),
       EVICTED(1 << 4),
-      LOADED(1 << 5),
+      EXPIRED(1 << 5),
       SKIP_LOOKUP(1 << 6),
       COPIED(1 << 7);
 
@@ -259,6 +259,11 @@ public class ReadCommittedEntry implements MVCCEntry {
    }
 
    @Override
+   public boolean isExpired() {
+      return isFlagSet(EXPIRED);
+   }
+
+   @Override
    public final void setRemoved(boolean removed) {
       setFlag(removed, REMOVED);
    }
@@ -269,13 +274,17 @@ public class ReadCommittedEntry implements MVCCEntry {
    }
 
    @Override
+   public void setExpired(boolean expired) {
+      setFlag(expired, EXPIRED);
+   }
+
+   @Override
    public boolean isLoaded() {
-      return isFlagSet(LOADED);
+      return false;
    }
 
    @Override
    public void setLoaded(boolean loaded) {
-      setFlag(loaded, LOADED);
    }
 
    protected final void setFlag(boolean enable, Flags flag) {
@@ -304,6 +313,7 @@ public class ReadCommittedEntry implements MVCCEntry {
             ", isChanged=" + isChanged() +
             ", isRemoved=" + isRemoved() +
             ", isValid=" + isValid() +
+            ", isExpired=" + isExpired() +
             ", skipRemoteGet=" + skipLookup() +
             ", metadata=" + metadata +
             '}';
@@ -313,6 +323,7 @@ public class ReadCommittedEntry implements MVCCEntry {
    public boolean undelete(boolean doUndelete) {
       if (isRemoved() && doUndelete) {
          if (trace) log.trace("Entry is deleted in current scope.  Un-deleting.");
+         setExpired(false);
          setRemoved(false);
          setValid(true);
          setValue(null);
--- a/core/src/main/java/org/infinispan/expiration/ExpirationManager.java
+++ b/core/src/main/java/org/infinispan/expiration/ExpirationManager.java
@@ -35,11 +35,32 @@ public interface ExpirationManager<K, V> {
    boolean isEnabled();
 
    /**
-    * This should be invoked passing in an entry that is now expired.  Note this method assumes to not have the
-    * lock for this key and will attempt to acquire it.
+    * This should be invoked passing in an entry that is now expired.  This method may attempt to lock this key to
+    * preserve atomicity.
     * @param entry entry that is now expired
+    * @param currentTime the current time in milliseconds
     */
-   void handleInMemoryExpiration(InternalCacheEntry<K, V> entry);
+   void handleInMemoryExpiration(InternalCacheEntry<K, V> entry, long currentTime);
 
+   /**
+    * This is to be invoked when a store entry expires.  This method may attempt to lock this key to preserve atomicity.
+    * <p>
+    * Note this method doesn't currently take a {@link InternalCacheEntry} and this is due to a limitation in the
+    * cache store API.  This may cause some values to be removed if they were updated at the same time.
+    * @param key the key of the expired entry
+    */
    void handleInStoreExpiration(K key);
+
+   /**
+    * This is to be invoked with a when a write is known to occur to prevent expiration from happening.  This way we
+    * won't have a swarm of remote calls required.
+    * @param key the key to use
+    */
+   void registerWriteIncoming(K key);
+
+   /**
+    * This should always be invoked after registering write but after performing any operations required.
+    * @param key the key to use
+    */
+   void unregisterWrite(K key);
 }
--- a/core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
+++ b/core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
@@ -1,0 +1,132 @@
+package org.infinispan.expiration.impl;
+
+import net.jcip.annotations.ThreadSafe;
+import org.infinispan.AdvancedCache;
+import org.infinispan.commands.CommandsFactory;
+import org.infinispan.commands.write.RemoveExpiredCommand;
+import org.infinispan.commons.util.Util;
+import org.infinispan.container.entries.ExpiryHelper;
+import org.infinispan.container.entries.InternalCacheEntry;
+import org.infinispan.context.InvocationContext;
+import org.infinispan.context.InvocationContextContainer;
+import org.infinispan.context.InvocationContextFactory;
+import org.infinispan.factories.KnownComponentNames;
+import org.infinispan.factories.annotations.ComponentName;
+import org.infinispan.factories.annotations.Inject;
+import org.infinispan.interceptors.InterceptorChain;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+
+import java.util.Iterator;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Allows for cluster based expirations to occur.  This provides guarantees that when an entry is expired that it will
+ * expire that entry across the entire cluster at once.  This requires obtaining the lock for said entry before
+ * expiration is performed.  Since expiration can occur without holding onto the lock it is possible for an expiration
+ * to occur immediately after a value has been updated.  This can cause a premature expiration to occur.  Attempts
+ * are made to prevent this by using the expired entry's value and lifespan to limit this expiration so it only happens
+ * in a smaller amount of cases.
+ * <p>
+ * Cache stores however do not supply the value or metadata information which means if an entry is purged from the cache
+ * store that it will forcibly remove the value even if a concurrent write updated it just before.  This will be
+ * addressed by future SPI changes to the cache store.
+ * @param <K>
+ * @param <V>
+ */
+@ThreadSafe
+public class ClusterExpirationManager<K, V> extends ExpirationManagerImpl<K, V> {
+   protected static final Log log = LogFactory.getLog(ClusterExpirationManager.class);
+   protected static final boolean trace = log.isTraceEnabled();
+
+   private ExecutorService asyncExecutor;
+   private AdvancedCache<K, V> cache;
+
+   @Inject
+   public void inject(AdvancedCache<K, V> cache,
+           @ComponentName(KnownComponentNames.ASYNC_OPERATIONS_EXECUTOR) ExecutorService asyncExecutor) {
+      this.cache = cache;
+      this.asyncExecutor = asyncExecutor;
+   }
+
+   @Override
+   public void processExpiration() {
+      long start = 0;
+      if (!Thread.currentThread().isInterrupted()) {
+         try {
+            if (trace) {
+               log.trace("Purging data container of expired entries");
+               start = timeService.time();
+            }
+            long currentTimeMillis = timeService.wallClockTime();
+            for (Iterator<InternalCacheEntry<K, V>> purgeCandidates = dataContainer.iteratorIncludingExpired();
+                 purgeCandidates.hasNext();) {
+               InternalCacheEntry<K, V> e = purgeCandidates.next();
+               if (e.canExpire()) {
+                  if (ExpiryHelper.isExpiredMortal(e.getLifespan(), e.getCreated(), currentTimeMillis)) {
+                     handleLifespanExpireEntry(e);
+                  } else if (ExpiryHelper.isExpiredTransient(e.getMaxIdle(), e.getLastUsed(), currentTimeMillis)) {
+                     super.handleInMemoryExpiration(e, currentTimeMillis);
+                  }
+               }
+            }
+            if (trace) {
+               log.tracef("Purging data container completed in %s",
+                       Util.prettyPrintTime(timeService.timeDuration(start, TimeUnit.MILLISECONDS)));
+            }
+         } catch (Exception e) {
+            log.exceptionPurgingDataContainer(e);
+         }
+      }
+
+      if (!Thread.currentThread().isInterrupted()) {
+         persistenceManager.purgeExpired();
+      }
+   }
+
+   void handleLifespanExpireEntry(InternalCacheEntry<K, V> entry) {
+      K key = entry.getKey();
+      // The most used case will be a miss so no extra read before
+      if (expiring.putIfAbsent(key, key) == null) {
+         long lifespan = entry.getLifespan();
+         if (trace) {
+            log.tracef("Submitting expiration removal for key %s which had lifespan of %s", key, lifespan);
+         }
+         asyncExecutor.submit(() -> {
+            try {
+               removeExpired(key, entry.getValue(), lifespan);
+            } finally {
+               expiring.remove(key);
+            }
+         });
+      }
+   }
+
+   private void removeExpired(K key, V value, Long lifespan) {
+      cache.removeExpired(key, value, lifespan);
+   }
+
+   @Override
+   public void handleInMemoryExpiration(InternalCacheEntry<K, V> entry, long currentTime) {
+      if (ExpiryHelper.isExpiredMortal(entry.getLifespan(), entry.getCreated(), currentTime)) {
+         handleLifespanExpireEntry(entry);
+      } else {
+         super.handleInMemoryExpiration(entry, currentTime);
+      }
+   }
+
+   @Override
+   public void handleInStoreExpiration(K key) {
+      expiring.put(key, key);
+      // Unfortunately stores don't pull the entry so we can't tell exactly why it expired and thus we have to remove
+      // the entire value.  Unfortunately this could cause a concurrent write to be undone
+      try {
+         removeExpired(key, null, null);
+      } finally {
+         expiring.remove(key);
+      }
+   }
+}
--- a/core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
+++ b/core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
@@ -1,0 +1,66 @@
+package org.infinispan.expiration.impl;
+
+import org.infinispan.commands.write.ClearCommand;
+import org.infinispan.commands.write.DataWriteCommand;
+import org.infinispan.commands.write.PutKeyValueCommand;
+import org.infinispan.commands.write.PutMapCommand;
+import org.infinispan.commands.write.RemoveCommand;
+import org.infinispan.context.InvocationContext;
+import org.infinispan.expiration.ExpirationManager;
+import org.infinispan.factories.annotations.Inject;
+import org.infinispan.interceptors.base.CommandInterceptor;
+
+import java.util.Set;
+
+/**
+ * Interceptor used to ensure that when a writie occurs we don't expire the entry at the same time.  This can be both
+ * costly by going remote an extra time or even removing from the cache store when not required as it will be
+ * overwritten.
+ */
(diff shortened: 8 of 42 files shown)


wburns:
Also this is in preview as the testAllExpire method can fail intermittently when used in a tx cache.  I am looking into it closer.

core/src/test/java/org/infinispan/notifications/cachelistener/cluster/AbstractClusterListenerTest.java
```diff
@@ -522,4 +523,147 @@ public void testCacheEventFilterConverter() {
...
+      eventually(() -> clusterListener.events.size() >= expectCount);
+      assertEquals(expectCount, clusterListener.events.size());
+      CacheEntryEvent event = clusterListener.events.get(clusterListener.hasIncludeState() ? 1 : 0);
+
+      assertEquals(Event.Type.CACHE_ENTRY_EXPIRED, event.getType());
+      assertEquals(key, event.getKey());
+      assertEquals(expectedValue, event.getValue());
+   }
+
+   @Test(invocationCount = 10)
```
wburns: This will be removed when I fix test failures.  Sorry it snuck in here :)

wburns:
Should be all fixed now

core/src/main/java/org/infinispan/AdvancedCache.java
```diff
@@ -551,4 +551,17 @@
     * @return the entry set containing all of the CacheEntries
     */
    CacheSet<CacheEntry<K, V>> cacheEntrySet();
+
+   /**
+    * Attempts to remove the entry if it is expired.  Due to expired entries not being consistent across nodes, this
+    * will still attempt to remove the value if it is not present.  Note that this will raise an expired event
+    * even if the entry is not present.  Normally this method should never be invoked except by the
+    * {@link ExpirationManager}.
```
galderz: If this method is only to be executed by ExpirationManager, why expose it in AdvancedCache? Users have easy access to AdvancedCache. Couldn't ExpirationManager call this directly without the need of AdvancedCache? IOW, if it has all the components available, it could just call this code below without the need to add another method to a public interface?

```
+      InvocationContext ctx = getInvocationContextWithImplicitTransaction(false, null, 1);
+      RemoveExpiredCommand command = commandsFactory.buildRemoveExpiredCommand(key, value, lifespan);
+      ctx.setLockOwner(command.getKeyLockOwner());
+      // Send an expired remove command to everyone
+      executeCommandAndCommitIfNeeded(ctx, command);
```

galderz: As an example, `ReadWriteMapImpl` and similar implementations do this very same thing, with all the components available from AdvancedCache, without the need to add any extra methods to AdvancedCache.

core/src/main/java/org/infinispan/commands/CommandsFactory.java
```diff
@@ -125,6 +125,15 @@
    InvalidateCommand buildInvalidateFromL1Command(Address origin, Set<Flag> flags, Collection<Object> keys);
 
    /**
+    * Builds an expired remove command that is used to remove only a specific expired entry
+    * @param key the key of the expired entry
+    * @param value the value of the entry when it was expired
+    * @param lifespan the lifespan that expired from the command
+    * @return a RemovedExpiredCommand
+    */
+   RemoveExpiredCommand buildRemoveExpiredCommand(Object key, Object value, Long lifespan);
```
galderz: Also, do you really need to add a new command here? Assuming I understood correctly what `RemoveExpiredCommand` does, couldn't you have a built a `ReadWriteCommand` with a function that removes if both the value and expired entry are the same? It should work with minimal work without the need a brand new command :|

galderz: To be more specific, `ReadWriteKeyCommand` should do the job.

galderz: Or even `ReadWriteKeyValueCommand`

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
```diff
@@ -0,0 +1,114 @@
...
+      K key = entry.getKey();
+      // We remove the entry if it is the same from the datacontainer while also submitting an async remove expired
+      // to other members of the cluster
+      // The most used case will be a miss so no extra read before
+      if (expiring.putIfAbsent(key, key) == null) {
+         long lifespan = entry.getLifespan();
+         if (trace) {
+            log.tracef("Submitting expiration removal for key %s which had lifespan of %s", key, lifespan);
+         }
+         asyncExecutor.submit(() -> {
```
galderz: Ok, here's how I think this could have been simplified:

In the inject method, construct a ReadWriteMap out of a AdvancedCache (via a FunctionalMapImpl), and here, call `eval(K, V, BiFunction<V, ReadWriteEntryView<K, V>, R> f)` passing in the key, value, and a BiFunction implementation that takes lifespan in constructor. In the BiFunction apply implementation, extract find lifespan metaparam (this part I'm not sure if it's in place, in terms of interoperability with data stored via normal cache impl, but should be easy to put in place), then compare both lifespan and the value, and remove if needed.

On top of that, since eval() is async already, no need to depend on the async executor.

I agree that all pieces might not be in place to achieve this (the only one I have doubt about is whether you can retrieve lifespan from metadata for data stored via Cache/AdvancedCache), but it's really were we need to move towards, because as I said in the blog, we have tons of redundancy internally.

galderz: The `expiring.remove(key);` could have been executed as a continuation of the CompletableFuture, via thenAccept(), thenConsume or similar.

galderz:
Something else missing about ReadWriteMap is transactional support, but I don't think this would be relevant for this use case.

core/src/main/java/org/infinispan/AdvancedCache.java
wburns: I actually did manually invoke the command through the interceptor chain before, however this didn't support transactions.

I must admit I didn't look into using the FunctionalMap, but to be honest it wouldn't be very helpful timing wise for me as I have to backport this stuff as well, which doesn't have the FunctionalMap :(

wburns:
@galderz I don't think I can change this over in time for the release or also to backport.  I thought about adding a more explicit message to the AdvancedCache method to say this method could go away at any time.  Do you think that is okay and we can revisit changing this over?

galderz:
@wburns Sure that's fine. The changes are mostly about internals except the API in AdvancedCache. Happy with a note saying that it could go away.

wburns:
Added comment, thanks @galderz

core/src/main/java/org/infinispan/container/entries/ReadCommittedEntry.java
```diff
@@ -56,12 +56,13 @@ public void copyStateFlagsFrom(StateChangingEntry other) {
       EVICTED(1 << 4),
       LOADED(1 << 5),
       SKIP_LOOKUP(1 << 6),
-      COPIED(1 << 7);
+      COPIED(1 << 7),
+      EXPIRED(1 << 8);
```
danberindei: The `LOADED` flag isn't used any more, so byte would have been enough.

wburns: Sure I can swap those out

wburns: Removed LOADED and replaced with EXPIRED

danberindei:
@wburns Got some failures on my machine:

    testAllExpire(org.infinispan.notifications.cachelistener.cluster.ClusterListenerReplTxInitialStateTest)  Time elapsed: 0.108 sec  <<< FAILURE!
    java.lang.AssertionError: expected:<MagicKey#null{361595cc@ClusterListenerReplTxInitialStateTest-NodeA-57974/55}-expiring> but was:<null>
	at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:88)
	at org.infinispan.notifications.cachelistener.cluster.AbstractClusterListenerTest.testAllExpire(AbstractClusterListenerTest.java:665)

    testAllExpire(org.infinispan.notifications.cachelistener.cluster.ClusterListenerDistTxTest)  Time elapsed: 0.051 sec  <<< FAILURE!
    java.lang.AssertionError: expected:<MagicKey#null{f769efe6@ClusterListenerDistTxTest-NodeA-4275/16}-expiring> but was:<null>
	at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:88)
	at org.infinispan.notifications.cachelistener.cluster.AbstractClusterListenerTest.testAllExpire(AbstractClusterListenerTest.java:665)

danberindei:
Actually there's a compilation failure in `DeltaAwareCacheEntry` now.

wburns:
Whoops,  I will get that in a bit.

wburns:
Yeah that is the intermittent one i had before.  Thought I got it with
other changes but guess not.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -557,6 +558,14 @@ private V removeInternal(Object key, EnumSet<Flag> explicitFlags, InvocationCont
       return (V) executeCommandAndCommitIfNeeded(ctx, command);
    }
 
+   @Override
+   public void removeExpired(K key, V value, Long lifespan) {
+      InvocationContext ctx = getInvocationContextWithImplicitTransaction(false, null, 1);
+      RemoveExpiredCommand command = commandsFactory.buildRemoveExpiredCommand(key, value, lifespan);
+      ctx.setLockOwner(command.getKeyLockOwner());
```
danberindei: I'm not sure about this... The key locks are no longer reentrant, so unlocking the key for the `RemoveExpiredCommand` will unlock it for the transaction/operation that triggered it as well.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
```diff
@@ -0,0 +1,179 @@
...
+         Object value = e.getValue();
+         // If the provided lifespan is null, that means it is a store removal command, so we can't compare lifespan
+         if (lifespan == null) {
+            if (valueMatcher.matches(e, value, e.getValue(), valueEquivalence)) {
+               e.setExpired(true);
+               return performRemove(e, ctx);
+            }
+            // If there is no metadata and no value that means it is gone currently or not shown due to expired
+         } else if (e.getMetadata() == null) {
+            if (value == null) {
```
danberindei: I think the comment would look better inside the `else`. The way it is now, it's not clear what's supposed to happen when the metadata is null but the value is not.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
```diff
@@ -0,0 +1,179 @@
...
+    * @return null
+    */
+   @Override
+   public Object perform(InvocationContext ctx) throws Throwable {
+      MVCCEntry e = (MVCCEntry) ctx.lookupEntry(key);
+      if (e != null && !e.isRemoved()) {
+         Object value = e.getValue();
+         // If the provided lifespan is null, that means it is a store removal command, so we can't compare lifespan
+         if (lifespan == null) {
+            if (valueMatcher.matches(e, value, e.getValue(), valueEquivalence)) {
```
danberindei: Could this check fail and expire a newer value?

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
```diff
@@ -0,0 +1,179 @@
...
+            }
+            // Lastly if there is metadata we have to verify it equals our lifespan and the value match
+         } else if ( e.getLifespan() == lifespan) {
+            // TODO: add a threshold to verify this wasn't just created with the same value/lifespan just before expiring
+            if (valueMatcher.matches(e, value, e.getValue(), valueEquivalence)) {
+               e.setExpired(true);
+               return performRemove(e, ctx);
+            }
+         } else {
+            log.trace("Cannot remove entry as it's lifespan or value do not match");
```
danberindei: s/it's/its/

core/src/main/java/org/infinispan/expiration/ExpirationManager.java
```diff
@@ -35,11 +35,32 @@
    boolean isEnabled();
 
    /**
-    * This should be invoked passing in an entry that is now expired.  Note this method assumes to not have the
-    * lock for this key and will attempt to acquire it.
+    * This should be invoked passing in an entry that is now expired.  This method assumes the lock is not acquire
+    * and may attempt to lock this key to preserve atomicity.
```
danberindei: I think it would be better to say it doesn't make any assumptions about the key being locked, because it should work if the current tx has the lock.

core/src/main/java/org/infinispan/expiration/ExpirationManager.java
```diff
@@ -35,11 +35,32 @@
...
     * @param entry entry that is now expired
     */
    void handleInMemoryExpiration(InternalCacheEntry<K, V> entry);
 
+   /**
+    * This is to be invoked when a store entry expires.  This method assumes the lock is not acquire and may attempt
+    * to lock this key to preserve atomicity.
+    * <p>
+    * Note this method doesn't currently take a {@link InternalCacheEntry} and this is due to a limitation in the
+    * cache store API.  This may cause some values to be removed if they were updated at the same time.
```
danberindei: I wonder if we could do something to minimize the chance of this happening without an API change... at least in `SingleFileStore.purge()`, it can take a long time between the entry being added to the `entriesToPurge` collection and the listener being invoked.

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
```diff
@@ -0,0 +1,114 @@
...
+      }
+
+      if (!Thread.currentThread().isInterrupted()) {
+         persistenceManager.purgeExpired();
+      }
+   }
+
+   void handleLifespanExpireEntry(InternalCacheEntry<K, V> entry) {
+      K key = entry.getKey();
+      // We remove the entry if it is the same from the datacontainer while also submitting an async remove expired
```
danberindei: "if it is the same" sounds a little vague... I guess for the local data container you could do an `==` check?

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
```diff
@@ -0,0 +1,114 @@
...
+      }
+   }
+
+   private void removeExpired(K key, V value, Long lifespan) {
+      cache.removeExpired(key, value, lifespan);
+   }
+
+   @Override
+   public void handleInMemoryExpiration(InternalCacheEntry<K, V> entry) {
+      if (ExpiryHelper.isExpiredMortal(entry.getLifespan(), entry.getCreated(), timeService.wallClockTime())) {
```
danberindei: Maybe you could pass the current time as a parameter here?

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
```diff
@@ -0,0 +1,66 @@
...
+   }
+
+   @Override
+   public Object visitPutKeyValueCommand(InvocationContext ctx, PutKeyValueCommand command) throws Throwable {
+      return handleWriteCommand(ctx, command);
+   }
+
+   @Override
+   public Object visitClearCommand(InvocationContext ctx, ClearCommand command) throws Throwable {
+      // TODO: do we want to stop all expirations?
```
danberindei: `clear` should be rare enough that we can ignore it - as long as it's just a matter of performance.

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
```diff
@@ -0,0 +1,66 @@
...
+
+   @Override
+   public Object visitPutMapCommand(InvocationContext ctx, PutMapCommand command) throws Throwable {
+      Set<K> keys = (Set<K>) command.getAffectedKeys();
+      keys.forEach(expirationManager::registerWriteIncoming);
+      try {
+         return super.visitPutMapCommand(ctx, command);
+      } finally {
+         // TODO: what about for transactions?
+         keys.forEach(expirationManager::unregisterWrite);
```
danberindei: For transactions you'd have to unregister the write in `visitCommitCommand` (or `visitPrepareCommand`, for 1PC transactions). But I think you should also register the write only during the prepare, because the transaction could take a long time.

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
```diff
@@ -0,0 +1,66 @@
...
+      }
+   }
+
+   private Object handleWriteCommand(InvocationContext ctx, DataWriteCommand command) throws Throwable {
+      K key = (K) command.getKey();
+      expirationManager.registerWriteIncoming(key);
```
danberindei: Does this work if the write is unsuccessful (i.e. `replace(k, v1, v2)`, but the current value is `v3`)?

core/src/main/java/org/infinispan/expiration/impl/ExpirationManagerImpl.java
```diff
@@ -23,25 +23,29 @@
...
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
 @ThreadSafe
 public class ExpirationManagerImpl<K, V> implements ExpirationManager<K, V> {
-   private static final Log log = LogFactory.getLog(ExpirationManagerImpl.class);
-   private static final boolean trace = log.isTraceEnabled();
-   ScheduledFuture <?> expirationTask;
+   protected final Log log = LogFactory.getLog(this.getClass());
+   protected final boolean trace = log.isTraceEnabled();
```
danberindei: I'd rather have a separate log in `ClusteredExpirationManager`, with its own category.

core/src/main/java/org/infinispan/factories/EmptyConstructorNamedCacheFactory.java
```diff
@@ -116,7 +117,12 @@
          } else if (componentType.equals(EvictionManager.class)) {
             return (T) new EvictionManagerImpl();
          } else if (componentType.equals(ExpirationManager.class)) {
-            return (T) new ExpirationManagerImpl();
+            CacheMode cacheMode = configuration.clustering().cacheMode();
+            if (cacheMode.isDistributed() || cacheMode.isReplicated()) {
```
danberindei: I prefer using a dedicated factory when the actual component type depends on the configuration.

core/src/main/java/org/infinispan/factories/InterceptorChainFactory.java
```diff
@@ -226,7 +228,11 @@ public InterceptorChain buildInterceptorChain() {
          interceptorChain.appendInterceptor(createInterceptor(new GroupingInterceptor(), GroupingInterceptor.class), false);
       }
 
-      if (needsVersionAwareComponents && configuration.clustering().cacheMode().isClustered()) {
+      // The expiration interceptor must come before the entry wrapping interceptor as it can stop data container
+      // access from expiring things.
```
danberindei: If the write is unregistered after the data container write, what would happen with an entry with a lifespan of 1ms that expires before the write that inserted it is unregistered?

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
```diff
@@ -128,15 +129,19 @@ protected abstract void commitSingleEntry(CacheEntry entry, Metadata metadata, F
...
+              InvocationContext ctx, FlagAffectedCommand command, Object previousValue, Metadata previousMetadata) {
          boolean isWriteOnly = (command instanceof WriteCommand) && ((WriteCommand) command).isWriteOnly();
          if (removed) {
             if (command instanceof RemoveCommand) {
                ((RemoveCommand)command).notify(ctx, previousValue, previousMetadata, false);
             } else {
-               notifier.notifyCacheEntryRemoved(
-                  entry.getKey(), previousValue, previousMetadata, false, ctx, command);
+               if (expired) {
+                  notifier.notifyCacheEntryExpired(entry.getKey(), previousValue, previousMetadata, ctx);
```
danberindei: When does this happen? (As opposed to going through the `if` branch, because `RemoveExpiredCommand` is a `RemoveCommand`.)

core/src/main/java/org/infinispan/persistence/PersistenceUtil.java
```diff
@@ -89,7 +89,7 @@ public static InternalMetadata internalMetadata(InternalCacheValue icv) {
                                                          final AtomicReference<Boolean> isLoaded) {
       return dataContainer.compute(key, (k, oldEntry, factory) -> {
          //under the lock, check if the entry exists in the DataContainer
-         if (oldEntry != null) {
+         if (oldEntry != null && (!oldEntry.canExpire() || !oldEntry.isExpired(timeService.wallClockTime()))) {
```
danberindei: Does this mean we can remove the expiration check from `CacheLoader.load()` implementations?

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
```diff
@@ -0,0 +1,114 @@
...
+@ThreadSafe
+public class ClusterExpirationManager<K, V> extends ExpirationManagerImpl<K, V> {
+
+   private ExecutorService asyncExecutor;
+   private AdvancedCache<K, V> cache;
+
+   @Inject
+   public void inject(AdvancedCache<K, V> cache,
+           @ComponentName(KnownComponentNames.ASYNC_OPERATIONS_EXECUTOR) ExecutorService asyncExecutor) {
+      this.cache = cache;
```
danberindei: I still think you could have cast to `CacheImpl` here to avoid exposing `removeExpired` in `AdvancedCache`...

core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
```diff
@@ -88,7 +88,7 @@ protected void eventually(String message, Condition ec, long timeoutMillis, int
    }
 
    protected void eventually(Condition ec, long timeout, TimeUnit unit) {
-      eventually(null, ec, timeout, 500, unit);
+      eventually(null, ec, unit.toMillis(timeout), 500, TimeUnit.MILLISECONDS);
```
danberindei: Nice catch :)

core/src/test/java/org/infinispan/expiration/impl/ClusterExpirationFunctionalTest.java
```diff
@@ -0,0 +1,114 @@
...
+   private void testExpiredEntryRetrieval(Cache<Object, String> primaryOwner, Cache<Object, String> backupOwner,
+           ControlledTimeService timeService, boolean expireOnPrimary) throws Exception {
+      MagicKey key = new MagicKey(primaryOwner, backupOwner);
+      primaryOwner.put(key, key.toString(), 10, TimeUnit.MINUTES);
+
+      assertEquals(key.toString(), primaryOwner.get(key));
+      assertEquals(key.toString(), backupOwner.get(key));
+
+      // Now we expire on cache0, it should still exist on cache1
+      timeService.advance(TimeUnit.MINUTES.toMillis(10) + 1);
```
danberindei: `timeService` isn't necessarily from `cache0`. TBH it looks like you could remove all parameters except `expireOnPrimary`.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
wburns: The RemoveExpiredCommand is never done in an ongoing transaction or from a user operation.  The removal is done asynchronously when a get finds an expired value and the reaper thread is isolated.  Or is there something else here?

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: Sure

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: It can match a newer value since value will usually be null in this case, which matches anything.  Unfortunately CacheStore doesn't currently expose the value or the lifespan when expiring them :(

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: Fixed.

core/src/main/java/org/infinispan/expiration/ExpirationManager.java
wburns: Alright.

core/src/main/java/org/infinispan/expiration/ExpirationManager.java
wburns: The problem is the CacheStores are provided a PurgeListener which only takes a key.  I had thought about removing usage of the PurgeListener and instead inject the ExpirationManager into the TaskContext, but that only applies to process method as well.  I couldn't figure out anything nice, unless you have any suggestions?

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
wburns: Whoops I can remove that actually, it is an outdated comment.

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
wburns: Hrmm we could do that, sure

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
wburns: It actually is beneficial to keep it registered for as long as possible.  I have updated the ExpirationManagerImpl to have more comments on the map.

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
wburns: Yeah it is only performance.  I put it more there as maybe we can do something later.

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
wburns: So all the registered write is to basically block any expiration events from being generated on this node (either from a read from the write or another concurrent read).  In the case of an unsuccessful conditional command, it would just delay the entry from being expired.  Everything should operate correctly still.

core/src/main/java/org/infinispan/expiration/impl/ExpirationManagerImpl.java
wburns: This has its own category.  Do you mean you want it static?  This just makes it so you don't have to define the log in both classes.

core/src/main/java/org/infinispan/factories/EmptyConstructorNamedCacheFactory.java
wburns: Yeah I actually had one written before but I noticed other guys in this factory were already doing this.  I can add a new one though.

core/src/main/java/org/infinispan/factories/InterceptorChainFactory.java
wburns: Nothing it would just delay the expiration from occuring until after the write was unregistered.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
wburns: This happens with transactions since there is no provided command.  This was the entire reason I had to add EXPIRED to the ReadCommittedEntry :(

core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
wburns: Yeah my test kept hanging for what seemed forever, since it wanted to wait for 500 SECONDS :P

core/src/main/java/org/infinispan/persistence/PersistenceUtil.java
wburns: Unfortunately a lot of places call load without invoking this method :(  The big problem is this wouldn't call to load if it found an expired entry.

core/src/test/java/org/infinispan/expiration/impl/ClusterExpirationFunctionalTest.java
wburns: Yeah I was planning on reusing this with a different combination at some point.

wburns:
Everything should be fixed up except for the intermittent test failure.  I will be looking closer tomorrow.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
danberindei: Sorry, my mistake. I thought you were copying the lock owner from the get command that triggered the expiration :)

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
danberindei: Got it. I thought you maybe wanted on keep the lock for key on the primary owner while doing the expiration on all the backups, then the value check maybe wouldn't be necessary.
OTOH, doesn't this introduce inconsistencies in the cache, with a value present on one owner but expired on another?

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
```diff
@@ -0,0 +1,179 @@
...
+            }
+         } else if (e.getMetadata() == null) {
+            // If there is no metadata and no value that means it is gone currently or not shown due to expired
+            if (value == null) {
+               e.setExpired(true);
+               return performRemove(e, ctx);
+            }
+         } else if ( e.getLifespan() == lifespan) {
+            // Lastly if there is metadata we have to verify it equals our lifespan and the value match
+            // TODO: add a threshold to verify this wasn't just created with the same value/lifespan just before expiring
```
danberindei: Maybe check the entire metadata for equality? If versioning is enabled, checking the version should be better than checking the value.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
```diff
@@ -0,0 +1,179 @@
...
+         Object value = e.getValue();
+         // If the provided lifespan is null, that means it is a store removal command, so we can't compare lifespan
+         if (lifespan == null) {
+            if (valueMatcher.matches(e, value, e.getValue(), valueEquivalence)) {
+               e.setExpired(true);
+               return performRemove(e, ctx);
+            }
+         } else if (e.getMetadata() == null) {
+            // If there is no metadata and no value that means it is gone currently or not shown due to expired
+            if (value == null) {
```
danberindei: I wish the comment also said something about the `value != null` case.

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
danberindei: Hmmm, are you sure? At least `SingleFileStore` is going to remove the expired entry regardless of what the `PurgeListener` does, so there isn't going to be a second time.

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
danberindei: Ah right, concurrent reads are not going to see the entry anyway because they still check `isExpired()`. I still have 2 concerns:
1. Is it ok to skip the expiration listener if the tx only commits 1 minute after the entry expired (I'm using 1 minute as an example because that's what I think Narayana's default timeout is).
2. Is the listener is going to be fired for entries that expired in the store while the write is registered and the write is unsuccessful or the tx is rolled back (see below, https://github.com/infinispan/infinispan/pull/3672/files#diff-43946d5696eeade2af866ca53d0f9762R58).

core/src/main/java/org/infinispan/expiration/impl/ExpirationManagerImpl.java
danberindei: Yes, I meant I want it static - with an instance field, `if (trace)` might be harder to optimize away.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
danberindei: Ok, got it. Could we remove the `RemoveCommand.notify()` call instead, then, or is it needed for eviction/invalidation and those don't need tx support?

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
```diff
@@ -349,6 +360,12 @@ protected void commitSingleEntry(CacheEntry entry, Metadata metadata, FlagAffect
          // TODO: Can the reset be done after notification instead?
          boolean created = entry.isCreated();
          boolean removed = entry.isRemoved();
+         boolean expired;
+         if (entry instanceof MVCCEntry) {
```
danberindei: Not sure this has any measurable performance impact, but could we do the `instanceof` check only if `!created && !removed`?

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -658,7 +658,7 @@ private void mergeFreeEntries(List<FileEntry> entries) {
    }
    
    @Override
-   public void purge(Executor threadPool, final PurgeListener task) {
+   public void    purge(Executor threadPool, final PurgeListener task) {
```
danberindei: extra spaces

core/src/test/java/org/infinispan/notifications/cachelistener/cluster/AbstractClusterListenerTest.java
```diff
@@ -522,4 +523,145 @@ public void testCacheEventFilterConverter() {
...
+      eventually(() -> clusterListener.events.size() >= expectCount);
+      // We can't assert size here, thiis is because expiration is done asynchronously.  As such you could have more
+      // than 1 expire command come at the same time, although from different nodes.  Currently we assume a null is
+      // okay to say it was expired, so you can get multiple expirations
+      CacheEntryEvent event = clusterListener.events.get(clusterListener.hasIncludeState() ? 1 : 0);
+
+      assertEquals(Event.Type.CACHE_ENTRY_EXPIRED, event.getType());
+      assertEquals(key, event.getKey());
+      if (event.getValue() == null) {
+         System.currentTimeMillis();
```
danberindei: Did you want to log the time here?

danberindei:
I think it would be good to have some high-level description of how cluster-wide expiration events are supposed to work, in the `ExpirationManagerImpl` javadoc perhaps, and maybe what the guarantees/limitations are in the `CacheEntryExpired` javadoc?

core/src/test/java/org/infinispan/notifications/cachelistener/cluster/AbstractClusterListenerTest.java
wburns: That was my testing code, to try to reproduce the intermittent issues.  I didn't have much luck last night.  This will be removed before I fix it.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: It does do the expiration holding onto the locks.  The problem is this is invoked without holding the lock first, which means the value could change between when it saw it expired and before it finally obtains the lock.  The value and lifespan check is a last ditch check to make sure we don't remove a value that has since been added if possible.

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
wburns: Yes, if you had a purge and a concurrent conditional command that fails, that would be problematic.  A bit more complicated :)

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
wburns: 1. Not sure I understand the question.  In this case we skip the listener but will instead replace it with a create event or similar.
2. Ah good catch, yes.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
wburns: Yeah this was something I found out is that eviction & invalidation notifications don't work properly on a tx cache, heh.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: Sure, normally it isn't a valid call - but this assumes the user knows how to call the method.  I could even add that case I guess.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: While I like it, this requires rewriting quite a bit of code :( - Maybe add this later for 8.0.1 or something?

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
wburns: Unfortunately not blocking it though has its own implications.  Since cache store purging only grants access to the key, there is no way to guarantee the cache store purge doesn't remove the updated value if say the concurrent write updated the value.

wburns: But at least that case we have consistent notifications.  I am thinking for handleInStoreExpiration that we always fire the command then.  WDYT?

core/src/main/java/org/infinispan/expiration/impl/ExpirationManagerImpl.java
wburns: This is a tad confusing, but probably still understandable since when a ClusterExpirationManager is in use that you will see ExpirationManagerImpl trace messages for some things.  But that should be okay.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
wburns: hrmm okay

wburns: Also I changed it to just ``if (removed && entry instanceof MVCCEntry)``  Since remove expiried sets both removed and expired.

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
wburns: We can revisit this later, maybe even use the functional map stuff Galder mentioned.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: I added this and also added a check below for lifespan being greater than 0.  So at least in the case of a concurrent write that doesn't have expiration will not be overwritten :)

wburns:
Updated with reworked comments.  However I haven't squashed the last commit until tests all pass.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
danberindei: My idea was that `RemoveExpiredCommand` could do the expiration check again after obtaining the lock on the primary owner. Obviously, that wouldn't work for expiration triggered from the cache stores...

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
danberindei: 1. was about entries that appear expired (`cache.get(key) == null`) for a relatively long time before they are finally updated and/or the expiration listener is fired. Even if you changed `cache.get(key)` to return the stale value, like Sanne suggested, stores will once again break things by removing the entry directly.

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
danberindei: Maybe you could add a `boolean successful` parameter to `unregisterWrite()`, and if `false`, fire the expiration during `unregisterWrite()` (if someone tried to expire the entry). Something to consider for 8.1?

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
danberindei: Could you create a JIRA for that?

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
danberindei: Cool, I wasn't sure removed was set.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: That is the point of the RemoveExpiredCommand.  Part of the issue is the context doesn't expose entries as expirable ones :(  To be honest a lot of this screams for the JIRA to refactor the cache entry hierarchy.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
wburns: Yeah I thought I did, but I guess I didn't get around to it.

wburns: https://issues.jboss.org/browse/ISPN-5707

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
wburns: Sorry I still don't understand what problem this would cause.  In the case of get it will constantly return null.  It just won't raise an expiration event and then later the create event will be raised no longer requiring the expiration event.

wburns:
Tests all passed, squashing commits

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
danberindei: But you don't check if the entry is really expired in `RemoveExpiredCommand`, right? You rely on the expiration check that was done before, without holding the lock.

core/src/main/java/org/infinispan/expiration/impl/ExpirationInterceptor.java
danberindei: Ok, maybe it's not a problem, but it's still weird if I register an expiration listener to make sure I re-fetch the value before it expires, and it takes 1 minute between the time I start getting `null` from the cache and the time I can see the new value.

core/src/test/java/org/infinispan/expiration/impl/ClusterExpirationFunctionalTest.java
```diff
@@ -0,0 +1,114 @@
...
+   }
+
+   public void testExpiredOnBoth() {
+      MagicKey key = new MagicKey(cache0, cache1);
+      cache0.put(key, key.toString(), 10, TimeUnit.MINUTES);
+
+      assertEquals(key.toString(), cache0.get(key));
+      assertEquals(key.toString(), cache1.get(key));
+
+      // Now we expire on cache0, it should still exist on cache1
```
danberindei: Comment doesn't seem correct

core/src/test/java/org/infinispan/notifications/cachelistener/cluster/AbstractClusterListenerTest.java
```diff
@@ -522,4 +523,142 @@ public void testCacheEventFilterConverter() {
...
+
+      assertNull(cache0.get(key));
+      assertNull(cache1.get(key));
+      assertNull(cache2.get(key));
+
+      int expectCount = clusterListener.hasIncludeState() ? 2 : 1;
+      eventually(() -> clusterListener.events.size() >= expectCount);
+      // We can't assert size here, thiis is because expiration is done asynchronously.  As such you could have more
+      // than 1 expire command come at the same time, although from different nodes.  Currently we assume a null is
+      // okay to say it was expired, so you can get multiple expirations
```
danberindei: Is the comment about the `eventually` line above?

danberindei:
Integrated, thanks Will!

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: For a cache store command that is correct.  The RemoveExpired can look for expiration by it being null (means it was expired from data container or was actually null) or by verifying the value and lifespan only.  But yes if we had access to the ``InternalCacheEntry`` in the context instead it would allow for verifying expiration better.

danberindei: Definitely worth investigating, then. Maybe even use a new entry type, like Vladimir did with `DeltaAwareCacheEntry`. I admit I didn't really like it at first, but it started to make sense after a while :)

PR #3679, opened 2015-08-29
ISPN-5696 Make the javax.transaction API jar a compile dependency instead of provided

tristantarrant:
https://issues.jboss.org/browse/ISPN-5696


danberindei:
Integrated, thanks Tristan!

PR #3680, opened 2015-08-31
Revert "[ISPN-5517] JDBCBinaryStore size method should query count directly instead of iterating"

danberindei:
https://issues.jboss.org/browse/ISPN-5517

Need to revert the fix because it ignores the timestamp of the entries in the bucket, and so it fails when a bucket contains a mix of expired and live entries.


anistor:
Pulling ...

anistor:
Integrated. Thanks @danberindei !

PR #3681, opened 2015-08-31
ISPN-5710 Unknown entity name when indexing entities defined in inner…

anistor:
… classes

Additional fix.

jira: https://issues.jboss.org/browse/ISPN-5710


danberindei:
Integrated, thanks Adrian!

PR #3684, opened 2015-09-01
ISPN-4250 Re-add the DONT_LOOPBACK flag

danberindei:
The performance problem was fixed in JGroups with JGRP-1835

https://issues.jboss.org/browse/ISPN-4250


galderz:

PR #3686, opened 2015-09-01
No need to set query parameters again after construction

anistor:
This was a leftover from an earlier version which did not take the params in the constructor.
Add test for hybrid query with parameter.


anistor:
CI failures are not related.

PR #3688, opened 2015-09-02
Fix incorrect format specifiers/missing parameters in log messages

danberindei:


danberindei:
I added a commit to use log4j2 in Spring and Netty.

galderz:
Integrated

PR #3689, opened 2015-09-02
ISPN-5719 Use log4j 2.0 everywhere

danberindei:
https://issues.jboss.org/browse/ISPN-5719


danberindei:
@tristantarrant I think I may need some additional changes for the server, can you have a look?

tristantarrant:
Looking

tristantarrant:
Server doesn't use log4j, so it's good. Pushed

spring/spring4/pom.xml
```diff
@@ -126,12 +126,27 @@
...
+                    <groupId>log4j</groupId>
+                    <artifactId>log4j</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-core</artifactId>
             <scope>runtime</scope>
-            <optional>true</optional>
```
jekh: @danberindei - Did you mean to remove the optional tag for log4j2? Since it's no longer optional, projects that declare a dependency on infinispan-spring4 end up transitively pulling in log4j2 as a runtime dependency, even though they may not want to use log4j2 as their logging framework.

PR #3690, opened 2015-09-02
ISPN-5716 Update last-minute changes in Functional docs

galderz:
* To avoid confusion, replace lambda word mentions with functions.
* SerializeLambdaWith and MarshallableLambdas were renamed just before
  final into SerializeFunctionWith and MarshallableFunctions, so update
  documentation.
* Finally, added apidocs links for MarshallableFunctions.


danberindei:
Integrated, thanks Galder!

PR #3691, opened 2015-09-02
ISPN-5718 Random failures in

wburns:
ClusterListenerReplTxInitialStateTest.testAllExpired

* Fixed issue where PrepareCommand was causing additional expirations
* Cause of failure was due to not locking key with optimistic prepare

https://issues.jboss.org/browse/ISPN-5718


galderz:
Integrated

PR #3692, opened 2015-09-03
ISPN-5724 Include expiration and contiunous query packages into javadoc

vjuranek:
https://issues.jboss.org/browse/ISPN-5724


anistor:
Integrated in master. Thanks @vjuranek !

PR #3696, opened 2015-09-04
ISPN-5727 Stop client listener executor on shutdown

galderz:
* Make sure also that when a client listener is removed, the task is
  cancelled.
* Added a test helper method to detect thread leaks given a regular
  expression.

### DIFF
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerLeakTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerLeakTest.java
@@ -1,0 +1,43 @@
+package org.infinispan.client.hotrod.event;
+
+import org.infinispan.client.hotrod.RemoteCache;
+import org.infinispan.client.hotrod.test.RemoteCacheManagerCallable;
+import org.infinispan.client.hotrod.test.SingleHotRodServerTest;
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.testng.annotations.Test;
+
+import static org.infinispan.client.hotrod.test.HotRodClientTestingUtil.withRemoteCacheManager;
+import static org.infinispan.server.hotrod.test.HotRodTestingUtil.hotRodCacheConfiguration;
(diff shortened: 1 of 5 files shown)


client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerLeakTest.java
```diff
@@ -0,0 +1,29 @@
...
+         public void call() {
+            RemoteCache<Integer, String> remote = rcm.getCache();
+            EventLogListener<Integer> eventListener = new EventLogListener<>();
+            remote.addClientListener(eventListener);
+            eventListener.expectNoEvents();
+            remote.put(1, "one");
+            eventListener.expectOnlyCreatedEvent(1, cache());
+         }
+      });
+      detectThreadLeaks(".*Client-Listener.*");
```
danberindei: Won't this fail if there's another test running in parallel and using client listeners?

danberindei: Failed in CI (and on my machine):
http://ci.infinispan.org/viewLog.html?buildId=30019&buildTypeId=bt9&guest=1

galderz:
@danberindei Fixed the issue by adding cache name to thread name and only detecting that thread.

danberindei:
This failure in HotRodOperationsAuthzIT.testSupervisorAddListener seems relevant:
http://ci.infinispan.org/viewLog.html?buildId=30078&buildTypeId=bt9&guest=1

danberindei:
The `HotRodOperationsAuthzIT.testSupervisorAddListener` NPE is still present in the latest build.

danberindei:
Personally I'm not sure the thread pool is needed at all - I would expect most listeners are installed for long periods of time, so there should be no problem with every listener creating its own thread.

galderz:
@danberindei Should be fixed now. Removed the cancelling of individual threads since that's not really needed (fixes the failing unit test too). The important thing is making sure the thread pool is shutdown. I added a little await for termination in the end, for more predictive unit testing.

danberindei:
Integrated, thanks Galder!

PR #3702, opened 2015-09-09
ISPN-5736 Hot Rod protocol 1.4 does not exist

galderz:


danberindei:
Integrated, thanks Galder!

PR #3704, opened 2015-09-14
ISPN-5742 & ISPN-3926 Global State Persistence configuration 

tristantarrant:
First step for graceful shutdown / restart
https://issues.jboss.org/browse/ISPN-5742
- This adds a new global configuration section called <state-persistence /> which allows specifying a path to where state will be persisted
- Bumps the core schema to 8.1

https://issues.jboss.org/browse/ISPN-3926
- Adds a PERSISTENT flag to the Internal Cache Registry
- Makes the protobuf metadata cache and scripting cache persistent


tristantarrant:
This should also fix the breakage caused by the bump to 8.1

danberindei:
Integrated, thanks Tristan!

PR #3708, opened 2015-09-15
ISPN-5740 HotRod client write buffer is too large

danberindei:
https://issues.jboss.org/browse/ISPN-5740

The read/write buffer in the HotRod client should not depend on
the socket's native buffers size, because it's not supposed to
duplicate the functionality of the native send/receive buffers.


tristantarrant:
Pushing

tristantarrant:
Thanks @danberindei

tristantarrant:
Cherry-picked to 8.0.x as well

PR #3712, opened 2015-09-16
ISPN-5756 Improve EmbeddedCacheManager.removeCache() method

pruivo:
https://issues.jboss.org/browse/ISPN-5756

### DIFF
--- a/core/src/main/java/org/infinispan/commands/RemoveCacheCommand.java
+++ b/core/src/main/java/org/infinispan/commands/RemoveCacheCommand.java
@@ -3,13 +3,17 @@ package org.infinispan.commands;
 import org.infinispan.Cache;
 import org.infinispan.commands.remote.BaseRpcCommand;
 import org.infinispan.commons.util.Util;
+import org.infinispan.configuration.ConfigurationManager;
 import org.infinispan.context.InvocationContext;
+import org.infinispan.eviction.PassivationManager;
+import org.infinispan.factories.ComponentRegistry;
 import org.infinispan.factories.GlobalComponentRegistry;
 import org.infinispan.jmx.CacheJmxRegistration;
-import org.infinispan.persistence.manager.PersistenceManager;
 import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.persistence.manager.PersistenceManager;
+import org.infinispan.util.DependencyGraph;
 
-import static org.infinispan.persistence.manager.PersistenceManager.AccessMode.BOTH;
+import static org.infinispan.factories.KnownComponentNames.CACHE_DEPENDENCY_GRAPH;
 
 /**
  * Command to stop a cache and remove all its contents from both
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
```diff
@@ -264,6 +264,15 @@
    void removeCache(String cacheName);
 
    /**
+    * Removes the cache with the given name from the local node.
+    * <p/>
+    * This operation removes the cache and all its contents from memory and cache stores.
+    *
+    * @param cacheName Name of the cache to remove.
+    */
+   void removeCacheLocally(String cacheName);
```
galderz: Why expose this method publicly? I accept you might want to consolidate logic for locally removing a cache, but I don't see why we should expose this to the user. The user should just call `removeCache` and let the internals decide how it does it.

pruivo: I don't wanna to expose it, but it was the only way I found to remove the cache from ```configurationOverrides``` and ```cacheDependencyGraph```. Any suggestion is welcome.

danberindei: I guess we could expose the method only in `DefaultCacheManager` and cast in `RemoveCacheCommand.perform()`?

pruivo: that could work. let me check if nothing is broken with it.

pruivo:
@danberindei @galderz updated!

core/src/main/java/org/infinispan/commands/RemoveCacheCommand.java
```diff
@@ -23,31 +18,19 @@
...
    }
 
    @Override
    public Object perform(InvocationContext ctx) throws Throwable {
-      persistenceManager.setClearOnStop(true);
-      cacheJmxRegistration.setUnregisterCacheMBean(true);
-      Cache<Object, Object> cache = cacheManager.getCache(cacheName);
-      cache.stop();
-      registry.removeCache(cacheName);
+      ((DefaultCacheManager) cacheManager).removeCacheLocally(cacheName);
```
galderz: This will most likely blow up when running within Wildfly since Wildfly does not use DefaulCacheManager implementation directly AFAIK. They implement a delegating cache manager AFAIK.

I really don't see the need for a `removeCacheLocally` method at all. Can't you create a private method in this command, get the component registry from the cache manager, which can be injected, and in the private method do everything you do in removeCacheLocally? If the method needs to be accessible from multiple places, add it to a util class as static method passing in all the parameters needed?

pruivo: I need to access the private field in ```DefaultCacheManager``` (```configurationOverrides``` and ```cacheDependencyGraph```)

galderz: > I need to access the private field in DefaultCacheManager (configurationOverrides and cacheDependencyGraph)

Ok. I'd suggest you do the following:

1. Check with Paul Ferraro if your changes would be problematic for him. I suspect that if you did the cast to `DefaultCacheManager`, then they would if they used this operation. If it's not a problem, then go ahead with the current fix.

2. If Paul says it could be problematic, I still don't think we should expose this `removeCacheLocally` via EmbeddedCacheManager, so what I'd try to do is to break the dependency between DefaultCacheManager and the `configurationOverrides` and `cacheDependencyGraph`. These two could be moved to a global component that deals with those.

How does that sound?

pruivo: sound good. going for option number 2 :)

pruivo:
@galderz @danberindei updated.
Added the ```ConfigurationManager``` and I'm registering in the ```GlobalComponentRegistry``` with the ```CacheDependencyGraph```.

danberindei:
I like the `ConfigurationManager`, integrating.

danberindei:
Integrated, thanks Pedro!

gustavonalle:
This PR is causing the query suite to throw lots of TimeoutExceptions and eventually hang
http://ci.infinispan.org/viewLog.html?buildId=30959&buildTypeId=bt9&guest=1
Any idea @danberindei  @pruivo ?

pruivo:
@gustavonalle @danberindei You're right. Some deadlock is happening when creating the metadata/locking caches by hibernate. I did some digging and I believe that is caused by the ConfigurationManager and probably I didn't implement the behavior correctly. 

I'll take a deeper look

pruivo:
@gustavonalle @danberindei fixed!

danberindei:
Integrated, thanks again Pedro!

PR #3713, opened 2015-09-17
ISPN-4956 Task Management

tristantarrant:
https://issues.jboss.org/browse/ISPN-4956 Implement a TaskManager
https://issues.jboss.org/browse/ISPN-5778 Support java.time.Instant in queries
https://issues.jboss.org/browse/ISPN-5779 Support java.time.Instant in TimeService

See https://github.com/infinispan/infinispan/wiki/Task-Execution-Design

### DIFF
--- a/object-filter/src/main/java/org/infinispan/objectfilter/impl/hql/ReflectionPropertyHelper.java
+++ b/object-filter/src/main/java/org/infinispan/objectfilter/impl/hql/ReflectionPropertyHelper.java
@@ -22,6 +22,7 @@ public final class ReflectionPropertyHelper extends ObjectPropertyHelper<Class<?
 
    static {
       primitives.put(java.util.Date.class, java.util.Date.class);
+      primitives.put(java.time.Instant.class, java.time.Instant.class);
       primitives.put(String.class, String.class);
       primitives.put(Character.class, Character.class);
       primitives.put(char.class, Character.class);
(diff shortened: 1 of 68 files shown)


tristantarrant:
Some bits are incomplete (attaching user info to the events, cancelling tasks, clearing task history, ...)
A few notes:
- I'm not happy with the methods for querying the task repo: I have alternate methods, but I'd prefer to have just one, maybe with Optional arguments ... suggestions ?
- we should probably have a maximum retention on the task history cache.
- while I've used CompletableFutures, I am actually "joining". We want tasks to run within a thread pool: dedicated or use the async pool ?
- we should allow marking some tasks as trivial, and thus skip event generation
- maybe add a maximum execution time to tasks

object-filter/src/main/java/org/infinispan/objectfilter/impl/hql/ReflectionPropertyHelper.java
```diff
@@ -22,6 +22,7 @@
 
    static {
       primitives.put(java.util.Date.class, java.util.Date.class);
+      primitives.put(java.time.Instant.class, java.time.Instant.class);
```
anistor: I think this is also needed in RowPropertyHelper.

tristantarrant:
Closing this and issuing separate PRs.

PR #3718, opened 2015-09-21
ISPN-5783 Cache.put(null, null) leaks an implicit transaction

danberindei:
https://issues.jboss.org/browse/ISPN-5783


tristantarrant:
Pushed, thanks

PR #3721, opened 2015-09-22
ISPN-5785 Make the GMS join timeout overridable in the server

gustavonalle:
https://issues.jboss.org/browse/ISPN-5785


tristantarrant:
Pushed, thanks

PR #3723, opened 2015-09-23
ISPN-5520 Hot Rod client XSite failover support

galderz:
https://issues.jboss.org/browse/ISPN-5520

* This initial implementation supports dealing with entire clusters
  failing, and falling back on cluster information defined in Hot Rod
  configuration.
* It's necessary for at least a single server's IP address to be
  provided in the cluster definition, and from that the rest of the
  topology can be deduced.
* Once a node has connected to a different site cluster, it remains
  connected to that site until the cluster becomes unavailable, in which
  case it failovers to next cluster.
* If all cluster sites become unavailable, as a last resort the initial
  server configuration is tried.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/RetryOnFailureOperation.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/RetryOnFailureOperation.java
@@ -107,7 +107,10 @@ public abstract class RetryOnFailureOperation<T> extends HotRodOperation {
    protected int logTransportErrorAndThrowExceptionIfNeeded(int i, HotRodClientException e) {
       String message = "Exception encountered. Retry %d out of %d";
       if (i >= transportFactory.getMaxRetries() || transportFactory.getMaxRetries() < 0) {
-         if (!triedCompleteRestart) {
+         if (transportFactory.trySwitchCluster(cacheName)){
+            triedCompleteRestart = true;
+            return -1; // reset retry count
+         }  else if (!triedCompleteRestart) {
             log.debug("Cluster might have completely shut down, try resetting transport layer and topology id", e);
             transportFactory.reset(cacheName);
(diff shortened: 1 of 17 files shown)


tristantarrant:
@galderz getting a new failure in SiteDownFailoverTest.testFailoverAfterSiteShutdown

galderz:
@tristantarrant Should be fixed now :)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/RetryOnFailureOperation.java
```diff
@@ -107,7 +107,11 @@ protected boolean shouldRetry(int retryCount) {
    protected int logTransportErrorAndThrowExceptionIfNeeded(int i, HotRodClientException e) {
       String message = "Exception encountered. Retry %d out of %d";
       if (i >= transportFactory.getMaxRetries() || transportFactory.getMaxRetries() < 0) {
-         if (!triedCompleteRestart) {
+         if (transportFactory.trySwitchCluster(cacheName)){
+            log.debug("Switched to a different cluster");
```
tristantarrant: Since switching to a different cluster is quite a big thing, I suggest turning this into an info and promoting it to Log.java complete with the name of the cluster

tristantarrant:
Galder, this is now passing locally, so :+1: . One additional comment above

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/RetryOnFailureOperation.java
galderz: Makes sense, I'll get that sorted.

galderz:
@tristantarrant Should be updated now. I removed that log line and added the info message within TcpTransportFactory

tristantarrant:
Thanks man !

PR #3726, opened 2015-09-24
ISPN-5790 AuthorizationManager rework

tristantarrant:
https://issues.jboss.org/browse/ISPN-5790

- use an internal cache instead of the Cluster Registry
- add a GlobalSecurityManager with the ability to flush acl caches
- don't cache per-cache acl masks
- add a role check in addition to the existing permission check

### DIFF
--- a/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
+++ b/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
@@ -1821,4 +1821,5 @@ public class CacheImpl<K, V> implements AdvancedCache<K, V> {
    public Properties getConfigurationAsProperties() {
       return new PropertyFormatter().format(config);
    }
+
 }
--- a/core/src/main/java/org/infinispan/configuration/global/GlobalSecurityConfigurationChildBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/global/GlobalSecurityConfigurationChildBuilder.java
@@ -1,5 +1,7 @@
 package org.infinispan.configuration.global;
 
+import java.util.concurrent.TimeUnit;
+
 /**
  * GlobalSecurityConfigurationChildBuilder.
  *
@@ -12,9 +14,21 @@ public interface GlobalSecurityConfigurationChildBuilder extends GlobalConfigura
     */
    GlobalAuthorizationConfigurationBuilder authorization();
    /**
-    * Defines the timeout in milliseconds for which to cache user access roles
+    * Defines the timeout in milliseconds for which to cache user access roles. A value of -1 means the entries
+    * never expire. A value of 0 will disable the cache.
     *
     * @param securityCacheTimeout
+    * @deprecated Use {@link #securityCacheTimeout(long, TimeUnit)} instead
     */
+   @Deprecated
    GlobalSecurityConfigurationBuilder securityCacheTimeout(long securityCacheTimeout);
+
+   /**
+    * Defines the timeout for which to cache user access roles. A value of -1 means the entries
+    * never expire. A value of 0 will disable the cache.
+    *
+    * @param securityCacheTimeout the timeout
+    * @param unit the {@link TimeUnit}
+    */
+   GlobalSecurityConfigurationBuilder securityCacheTimeout(long securityCacheTimeout, TimeUnit unit);
 }
--- a/core/src/main/java/org/infinispan/factories/EmptyConstructorFactory.java
+++ b/core/src/main/java/org/infinispan/factories/EmptyConstructorFactory.java
@@ -10,10 +10,6 @@ import org.infinispan.factories.scopes.Scopes;
 import org.infinispan.marshall.core.ExternalizerTable;
 import org.infinispan.remoting.inboundhandler.GlobalInboundInvocationHandler;
 import org.infinispan.remoting.inboundhandler.InboundInvocationHandler;
-import org.infinispan.topology.ClusterTopologyManager;
-import org.infinispan.topology.ClusterTopologyManagerImpl;
-import org.infinispan.topology.LocalTopologyManager;
-import org.infinispan.topology.LocalTopologyManagerImpl;
 import org.infinispan.util.DefaultTimeService;
 import org.infinispan.util.TimeService;
 import org.infinispan.xsite.BackupReceiverRepository;
@@ -27,31 +23,26 @@ import org.infinispan.xsite.BackupReceiverRepositoryImpl;
  * @since 4.0
  */
 
-@DefaultFactoryFor(classes = {InboundInvocationHandler.class, RemoteCommandsFactory.class, ExternalizerTable.class,
-                              BackupReceiverRepository.class, CancellationService.class, TimeService.class})
+@DefaultFactoryFor(classes = {BackupReceiverRepository.class, CancellationService.class, ExternalizerTable.class,
+                              InboundInvocationHandler.class, RemoteCommandsFactory.class, TimeService.class})
 @Scope(Scopes.GLOBAL)
 public class EmptyConstructorFactory extends AbstractComponentFactory implements AutoInstantiableFactory {
 
    @Override
    @SuppressWarnings("unchecked")
    public <T> T construct(Class<T> componentType) {
-      if (componentType.equals(RemoteCommandsFactory.class))
-         return (T) new RemoteCommandsFactory();
-      else if (componentType.equals(ExternalizerTable.class))
-         return (T) new ExternalizerTable();
-      else if (componentType.equals(LocalTopologyManager.class))
-         return (T) new LocalTopologyManagerImpl();
-      else if (componentType.equals(ClusterTopologyManager.class))
-         return (T) new ClusterTopologyManagerImpl();
-      else if (componentType.equals(BackupReceiverRepository.class))
+      if (componentType.equals(BackupReceiverRepository.class))
          return (T) new BackupReceiverRepositoryImpl();
       else if (componentType.equals(CancellationService.class))
          return (T) new CancellationServiceImpl();
-      else if (componentType.equals(TimeService.class)) {
-         return (T) new DefaultTimeService();
-      } else if (componentType.equals(InboundInvocationHandler.class)) {
+      else if (componentType.equals(ExternalizerTable.class))
+         return (T) new ExternalizerTable();
+      else if (componentType.equals(InboundInvocationHandler.class))
          return (T) new GlobalInboundInvocationHandler();
-      }
+      else if (componentType.equals(RemoteCommandsFactory.class))
+         return (T) new RemoteCommandsFactory();
+      else if (componentType.equals(TimeService.class))
+         return (T) new DefaultTimeService();
 
       throw new CacheConfigurationException("Don't know how to create a " + componentType.getName());
    }
--- a/core/src/main/java/org/infinispan/registry/InternalCacheRegistry.java
+++ b/core/src/main/java/org/infinispan/registry/InternalCacheRegistry.java
@@ -18,17 +18,45 @@ import org.infinispan.factories.scopes.Scopes;
 public interface InternalCacheRegistry {
    enum Flag {
       EXCLUSIVE, // means that the cache must be declared only once
-      USER,  // means that this cache is visible to users
+      USER, // means that this cache is visible to users
       PERSISTENT, // means the cache should be made persistent across restarts if global state persistence is enabled
    }
 
+   /**
+    * Registers an internal cache. The cache will be marked as private and volatile
+    *
+    * @param name
+    *           The name of the cache
+    * @param configuration
+    *           The configuration for the cache
+    */
    void registerInternalCache(String name, Configuration configuration);
 
+   /**
+    * Registers an internal cache with the specified flags.
+    *
+    * @param name
+    *           The name of the cache
+    * @param configuration
+    *           The configuration for the cache
+    * @param flags
+    *           The flags which determine the behaviour of the cache. See {@link Flag}
+    */
    void registerInternalCache(String name, Configuration configuration, EnumSet<Flag> flags);
 
+   /**
+    * Returns whether the cache is internal, i.e. it has been registered using the
+    * {@link #registerInternalCache(String, Configuration)} method
+    */
    boolean isInternalCache(String name);
 
+   /**
+    * Retrieves the names of all the internal caches
+    */
    Set<String> getInternalCacheNames();
 
-   void filterInternalCaches(Set<String> names);
+   /**
+    * Removes the private caches from the specified set of cache names
+    */
+   void filterPrivateCaches(Set<String> names);
 }
--- a/core/src/main/java/org/infinispan/security/AuthorizationManager.java
+++ b/core/src/main/java/org/infinispan/security/AuthorizationManager.java
@@ -16,5 +16,15 @@ import org.infinispan.factories.scopes.Scopes;
  */
 @Scope(Scopes.NAMED_CACHE)
 public interface AuthorizationManager {
-   void checkPermission(AuthorizationPermission permissions);
+   /**
+    * Verifies that the {@link Subject} associated with the current {@link AccessControlContext}
+    * has the requested permission. A {@link SecurityException} is thrown otherwise.
+    */
+   void checkPermission(AuthorizationPermission permission);
+
+   /**
+    * Verifies that the {@link Subject} associated with the current {@link AccessControlContext}
+    * has the requested permission and role. A {@link SecurityException} is thrown otherwise.
+    */
+   void checkPermission(AuthorizationPermission permission, String role);
 }
--- a/core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
+++ b/core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
@@ -2,19 +2,21 @@ package org.infinispan.security.impl;
 
 import java.security.AccessControlException;
 import java.security.Principal;
+import java.security.acl.Group;
+import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.ConcurrentMap;
 
 import javax.security.auth.Subject;
 
+import org.infinispan.commons.util.CollectionFactory;
 import org.infinispan.commons.util.Util;
 import org.infinispan.configuration.cache.AuthorizationConfiguration;
 import org.infinispan.configuration.global.GlobalSecurityConfiguration;
-import org.infinispan.registry.ClusterRegistry;
 import org.infinispan.security.AuditContext;
 import org.infinispan.security.AuditLogger;
 import org.infinispan.security.AuditResponse;
-import org.infinispan.security.AuthorizationManager;
 import org.infinispan.security.AuthorizationPermission;
 import org.infinispan.security.PrincipalRoleMapper;
 import org.infinispan.security.Role;
@@ -35,27 +37,31 @@ public class AuthorizationHelper {
    private final AuditLogger audit;
    private final AuditContext context;
    private final String name;
-   private final ClusterRegistry<String, Subject, Integer> maskCache;
-   private final String maskCacheScope;
+   private final ConcurrentMap<CachePrincipalPair, SubjectACL> aclCache;
 
-   public AuthorizationHelper(GlobalSecurityConfiguration globalConfiguration, AuditContext context, String name, ClusterRegistry<String, Subject, Integer> clusterRegistry) {
+   public AuthorizationHelper(GlobalSecurityConfiguration globalConfiguration, AuditContext context, String name,
+         ConcurrentMap<CachePrincipalPair, SubjectACL> aclCache) {
       this.globalConfiguration = globalConfiguration;
       this.audit = globalConfiguration.authorization().auditLogger();
       this.context = context;
       this.name = name;
-      this.maskCache = clusterRegistry;
-      this.maskCacheScope = AuthorizationManager.class.getSimpleName() + "_" + name;
+      this.aclCache = aclCache;
    }
 
-   public AuthorizationHelper(GlobalSecurityConfiguration globalConfiguration, AuditContext context, String name) {
-      this(globalConfiguration, context, name, null);
+   public AuthorizationHelper(GlobalSecurityConfiguration security, AuditContext context, String name) {
+      this(security, context, name, CollectionFactory.makeBoundedConcurrentMap(10));
    }
 
    public void checkPermission(AuthorizationPermission perm) {
-      checkPermission(null, perm);
+      checkPermission(null, perm, null);
    }
 
    public void checkPermission(AuthorizationConfiguration configuration, AuthorizationPermission perm) {
+      checkPermission(configuration, perm, null);
+   }
+
+   public void checkPermission(AuthorizationConfiguration configuration, AuthorizationPermission perm,
+         String role) {
       if (globalConfiguration.authorization().enabled()) {
          if (Security.isPrivileged()) {
             Security.checkPermission(perm.getSecurityPermission());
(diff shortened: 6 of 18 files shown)


core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
```diff
@@ -88,44 +91,33 @@ private void checkSecurityManagerPermission(AuthorizationPermission perm) {
...
+         for(String role : subjectACL.roles) {
+            if (configuration == null || configuration.roles().contains(role)) {
+               Role globalRole = globalRoles.get(role);
+               if (globalRole != null)
+                  subjectMask |= globalRole.getMask();
             }
-            return mask;
          }
+         int permissionMask = requiredPermission.getMask();
+         return (subjectMask & permissionMask) == permissionMask && requestedRole.map(r -> subjectACL.roles.contains(r)).orElse(true);
```
pruivo: nitpick: lambda could be replaced by ```subjectACL.roles::contains```

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
```diff
@@ -88,44 +91,33 @@ private void checkSecurityManagerPermission(AuthorizationPermission perm) {
...
-                  maskCache.put(maskCacheScope, subject, mask, globalConfiguration.securityCacheTimeout(), TimeUnit.MILLISECONDS);
-               }
-            } catch (IllegalStateException e) {
-               // Ignore
+            return new SubjectACL(allRoles);
+         });
+         int subjectMask = 0;
+         Map<String, Role> globalRoles = globalConfiguration.authorization().roles();
+         for(String role : subjectACL.roles) {
+            if (configuration == null || configuration.roles().contains(role)) {
```
pruivo: you could check ```configuration == null``` before iterating

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
```diff
@@ -88,44 +91,33 @@ private void checkSecurityManagerPermission(AuthorizationPermission perm) {
...
+               Role globalRole = globalRoles.get(role);
+               if (globalRole != null)
```
pruivo: is it possible the ```configuration``` to have a role that is not defined in the ```global configuration```?

core/src/main/java/org/infinispan/security/impl/AuthorizationManagerImpl.java
```diff
@@ -32,15 +28,20 @@ public AuthorizationManagerImpl() {
    }
 
    @Inject
-   public void init(Cache<?, ?> cache, GlobalConfiguration globalConfiguration, Configuration configuration,
-         ClusterRegistry<String, Subject, Integer> clusterRegistry) {
+   public void init(Cache<?, ?> cache, GlobalConfiguration globalConfiguration, Configuration configuration, GlobalSecurityManager globalSecurityManager) {
       this.globalConfiguration = globalConfiguration.security();
```
pruivo: can be converted to a local variable.

core/src/test/java/org/infinispan/security/RolePermissionTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.infinispan.configuration.global.GlobalConfigurationBuilder;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.security.impl.IdentityRoleMapper;
+import org.infinispan.test.SingleCacheManagerTest;
+import org.infinispan.test.TestingUtil;
+import org.infinispan.test.fwk.TestCacheManagerFactory;
+import org.testng.annotations.Test;
+
+@Test(groups="functional", testName="security.RolePermissionTest")
+public class RolePermissionTest extends SingleCacheManagerTest {
```
pruivo: this test can be improved. If I understand correctly, this test checks if the subject has the permission and the role.  So, it makes sense to add tests for:
* check correct permission and no role
* check wrong permission and no role => throw exception
* check correct permission and correct role
* check correct permission and wrong role => throw exception
* check wrong permission and correct role => throw exception
* check wrong permission and wrong role => throw exception

core/src/main/java/org/infinispan/security/AuthorizationManager.java
```diff
@@ -16,5 +17,20 @@
...
+    * @param permission
+    */
+   void checkPermission(AuthorizationPermission permission);
+
+   /**
+    * Verifies that the {@link Subject} associated with the current {@link AccessControlContext}
+    * has the requested permission and role. A {@link SecurityException} is thrown otherwise.
+    *
+    * @param permission
+    * @param requiredRole
```
pruivo: wrong param name

core/src/main/java/org/infinispan/security/impl/SubjectACL.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.util.Set;
+
+/**
+ * SubjectRoles.
+ *
+ * @author Tristan Tarrant
+ * @since 8.1
+ */
+public class SubjectACL {
```
pruivo: Some comments:
* it looks like it is stored in cache. An externalizer would be fine
* the ACL mask could be calculated in the class and cached (it does not changes right?)
* the roles should not be exposed and a containsRole method should be added
* the roles could be stored as a sorted array and use binarySearch to find it (assuming that a subject as small role set, it would be memory efficient)

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
tristantarrant: No.

pruivo: so, theoretically, this condition is always true :)

core/src/main/java/org/infinispan/security/AuthorizationManager.java
danberindei: IntelliJ also warns when there's a `@param` without any description, I'd rather remove them altogether.

core/src/main/java/org/infinispan/security/impl/AuthorizationManagerImpl.java
tristantarrant: what can be converted ?

core/src/main/java/org/infinispan/security/impl/SubjectACL.java
tristantarrant: The ACL cache is local, in-memory only, so not sure an Externalizer makes sense

core/src/main/java/org/infinispan/security/impl/AuthorizationManagerImpl.java
pruivo: the ```globalConfiguration```. it is only used in the ```init()``` method

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
tristantarrant: That's an || not an &&

core/src/main/java/org/infinispan/security/impl/SubjectACL.java
tristantarrant: The ACL mask could be different per-cache...

tristantarrant: Probably the best way here is to change the AclCache key to be the combination of Cache/Subject. At this point we can cache the mask

core/src/test/java/org/infinispan/security/RolePermissionTest.java
tristantarrant: Added all permutations. Also added a subclassed test for running the same tests with the ACL cache disabled

tristantarrant:
I have made some further changes:
- the ACL cache is disabled if the global securityCacheTimeout is set to 0 (as per discussion with @Sanne yesterday on IRC) for cases where users need real-time enforcement of user permissions changes. The global security mbean has an MBean op to flush the ACL cache manually.
- the key in the ACL cache is now a combination of cache and the Subject's user principal
- the ACL mask for a subject is precomputed and cached
- The ACL cache is a "simple cache", since it is local-only

pruivo:
some nitpick about ```GlobalSecurityManagerImpl```:
* you could use a ConcurrentHashMap when expiration is not needed. 
* Instead of storing the CacheManager and a boolean, you can store the reference to the map.

core/src/main/java/org/infinispan/security/impl/CachePrincipalPair.java
```diff
@@ -0,0 +1,56 @@
...
+   CachePrincipalPair(Principal userPrincipal, String cacheName) {
+      this.principalName = userPrincipal.getName();
+      this.cacheName = cacheName;
+      this.hashCode = computeHashCode();
+   }
+
+   private int computeHashCode() {
+      final int prime = 31;
+      int result = 1;
+      result = prime * result + ((cacheName == null) ? 0 : cacheName.hashCode());
```
pruivo: can the ```cacheName``` or the ```principalName``` be ```null```?

core/src/test/java/org/infinispan/security/RolePermissionNoACLCacheTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.infinispan.configuration.global.GlobalConfigurationBuilder;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.security.impl.IdentityRoleMapper;
+import org.infinispan.test.TestingUtil;
+import org.infinispan.test.fwk.TestCacheManagerFactory;
+import org.testng.annotations.Test;
+
+@Test(groups="functional", testName="security.RolePermissionNoACLCacheTest")
+public class RolePermissionNoACLCacheTest extends RolePermissionTest {
+   static final Subject ADMIN = TestingUtil.makeSubject("admin");
```
pruivo: none of this static and ```authzManager``` are used.

core/src/test/java/org/infinispan/security/RolePermissionTest.java
```diff
@@ -0,0 +1,147 @@
...
+   static final Subject ADMIN = TestingUtil.makeSubject("admin");
+   static final Subject SUBJECT_A = TestingUtil.makeSubject("A", "role1");
+   static final Subject SUBJECT_B = TestingUtil.makeSubject("B", "role2");
```
pruivo: this poor subject is never used :)

Sanne:
A cache based on the Principals's name is only effective if you assume there is a limited number of principals accessing, otherwise you're risking to store a lot of data for no good reason - and probably still have a low hit rate.
I think that's another good reason to keep the ACL cache disabled by default.

tristantarrant:
@Sanne not really if we assume that a principal will perform operations on the cache in bursts: you want to cache the ACL on the first op so that subsequent ops can reuse it. Set a low expiration for this case.
Another case where the ACL cache is an advantage is when principals are used to coarsely identify applications and not nominal users. So few principals, lots of repeated access.

tristantarrant:
@Sanne I see the use-case of lots of individual principals less practical / useful in the real world.

tristantarrant:
@pruivo Regarding the GlobalSecurityManagerImpl nitpicks:
Unfortunately I cannot retrieve the cache in the @inject, since it may not be available yet at that stage. That is why I have that "cacheEnabled" boolean (which, I agree, is not very nice).

danberindei:
I'm getting a NPE in `CacheAuthorizationTest`:

    createBeforeClass(org.infinispan.security.CacheAuthorizationTest)  Time elapsed: 0.091 sec  <<< FAILURE!
    java.lang.NullPointerException
	at org.infinispan.security.impl.AuthorizationHelper.computeSubjectACL(AuthorizationHelper.java:126)
	at org.infinispan.security.impl.AuthorizationHelper.lambda$checkSubjectPermissionAndRole$164(AuthorizationHelper.java:102)
	at org.infinispan.commons.util.concurrent.jdk8backported.BoundedEquivalentConcurrentHashMapV8.computeIfAbsent(BoundedEquivalentConcurrentHashMapV8.java:3326)
	at org.infinispan.security.impl.AuthorizationHelper.checkSubjectPermissionAndRole(AuthorizationHelper.java:101)
	at org.infinispan.security.impl.AuthorizationHelper.checkPermission(AuthorizationHelper.java:73)
	at org.infinispan.security.impl.AuthorizationHelper.checkPermission(AuthorizationHelper.java:57)
	at org.infinispan.manager.DefaultCacheManager.start(DefaultCacheManager.java:634)
	at org.infinispan.manager.DefaultCacheManager.<init>(DefaultCacheManager.java:222)
	at org.infinispan.test.fwk.TestCacheManagerFactory.newDefaultCacheManager(TestCacheManagerFactory.java:378)
	at org.infinispan.test.fwk.TestCacheManagerFactory.newDefaultCacheManager(TestCacheManagerFactory.java:71)
	at org.infinispan.test.fwk.TestCacheManagerFactory.createCacheManager(TestCacheManagerFactory.java:212)
	at org.infinispan.security.CacheAuthorizationTest$1.run(CacheAuthorizationTest.java:60)
	at org.infinispan.security.CacheAuthorizationTest$1.run(CacheAuthorizationTest.java:57)
	at org.infinispan.security.Security.doAs(Security.java:118)
	at org.infinispan.security.CacheAuthorizationTest.createCacheManager(CacheAuthorizationTest.java:57)
	at org.infinispan.security.CacheAuthorizationTest.setup(CacheAuthorizationTest.java:67)
	at org.infinispan.test.SingleCacheManagerTest.createBeforeClass(SingleCacheManagerTest.java:44)

tristantarrant:
Fixed. Unfortunately I had to reinstate an additional null check since, contrary to the discussion I had above with @pruivo caches may have roles not yet defined in the Global Role mapping

danberindei:
@tristantarrant Still some failures in the integration tests:
http://ci.infinispan.org/viewLog.html?buildId=31358&buildTypeId=bt9&guest=1

tristantarrant:
Right @danberindei 
I have made some further changes to how the lifecycle of internal caches is handled. They are now all stopped by the component registry, i.e. after user caches are shutdown. Let's see what CI says

tristantarrant:
All the failures are unrelated (and they work locally)

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -1821,4 +1821,9 @@ private void resumePreviousOngoingTransaction(Transaction transaction, boolean f
    public Properties getConfigurationAsProperties() {
       return new PropertyFormatter().format(config);
    }
+
+   @Override
+   public int hashCode() {
+      return name.hashCode();
```
danberindei: Why only override `hashCode()` and not `equals()`?

core/src/main/java/org/infinispan/registry/InternalCacheRegistry.java
```diff
@@ -18,17 +18,45 @@
...
+   /**
+    * Retrieves the names of all the internal caches
+    */
    Set<String> getInternalCacheNames();
 
-   void filterInternalCaches(Set<String> names);
+   /**
+    * Removes the internal caches from the specified set of cache names
+    */
+   void filterPrivateCaches(Set<String> names);
```
danberindei: Why the name change? The javadoc still mentions internal caches...

tristantarrant: The javadoc is wrong and the actual implementation was filtering the private caches :) I'll fix this

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
tristantarrant: I'll add that

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
```diff
@@ -35,27 +38,31 @@
...
 
-   public AuthorizationHelper(GlobalSecurityConfiguration globalConfiguration, AuditContext context, String name) {
-      this(globalConfiguration, context, name, null);
+   public AuthorizationHelper(GlobalSecurityConfiguration security, AuditContext context, String name) {
+      this(security, context, name, CollectionFactory.makeBoundedConcurrentMap(10));
    }
 
    public void checkPermission(AuthorizationPermission perm) {
-      checkPermission(null, perm);
+      checkPermission(null, perm, Optional.empty());
```
danberindei: I don't really like mixing `Optional` with old-style nullable parameters, so I'd rather use `null` for both. (Or `Optional` for both, of course.)

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
```diff
@@ -88,44 +94,56 @@ private void checkSecurityManagerPermission(AuthorizationPermission perm) {
...
       if (subject != null) {
-         Integer cachedMask;
-         try {
-            cachedMask = maskCache != null ? maskCache.get(maskCacheScope, subject) : null;
-         } catch (IllegalStateException e) {
-            cachedMask = null;
+         CachePrincipalPair cpp = new CachePrincipalPair(getUserPrincipal(subject), name);
+         SubjectACL subjectACL = aclCache != null ? aclCache.computeIfAbsent(cpp, s -> {
+            return computeSubjectACL(subject, configuration);
+         }) : computeSubjectACL(subject, configuration);
```
danberindei: I think an `if` statement would be easier to read.

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
tristantarrant: I've decided to go with nulls, since they are cheaper, which in this context makes more sense.

tristantarrant:
@danberindei ready to go !

tristantarrant:
@danberindei repushed :)

core/src/main/java/org/infinispan/configuration/global/GlobalSecurityConfigurationChildBuilder.java
```diff
@@ -12,7 +12,8 @@
     */
    GlobalAuthorizationConfigurationBuilder authorization();
    /**
-    * Defines the timeout in milliseconds for which to cache user access roles
+    * Defines the timeout in milliseconds for which to cache user access roles. A value of -1 means the entries
```
wburns: I know this might be too late now.  But was there a reason we forced this to be milliseconds and didn't just do the new common method of passing a long and a TimeUnit?

core/src/main/java/org/infinispan/factories/EmptyConstructorFactory.java
```diff
@@ -27,31 +23,26 @@
  * @since 4.0
  */
 
-@DefaultFactoryFor(classes = {InboundInvocationHandler.class, RemoteCommandsFactory.class, ExternalizerTable.class,
-                              BackupReceiverRepository.class, CancellationService.class, TimeService.class})
+@DefaultFactoryFor(classes = {BackupReceiverRepository.class, CancellationService.class, ExternalizerTable.class,
+                              InboundInvocationHandler.class, RemoteCommandsFactory.class, TimeService.class})
```
wburns: Nitpicking here a bit, but it seems there are a lot of extra additions and removals in this file, when it seems we only needed some removals and 1 addition.  There was a lot of reordering of code, which afaik isn't really necessary.

core/src/main/java/org/infinispan/configuration/global/GlobalSecurityConfigurationChildBuilder.java
tristantarrant: Fixed it.

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
```diff
@@ -88,44 +93,60 @@ private void checkSecurityManagerPermission(AuthorizationPermission perm) {
...
+         return subjectACL.matches(permissionMask) && (requestedRole != null ? subjectACL.containsRole(requestedRole) : true);
+      } else {
+         return false;
+      }
+   }
+
+   private SubjectACL computeSubjectACL(Subject subject, AuthorizationConfiguration configuration) {
+      PrincipalRoleMapper roleMapper = globalConfiguration.authorization().principalRoleMapper();
+      Set<Principal> principals = subject.getPrincipals();
+      Set<String> allRoles = new HashSet<>(principals.size());
```
wburns: Couldn't there be a lot more roles than # of principals?  How many principals would a subject have normally?  I wonder if we just use the default constructor which sizes the array to 16 might be better?

core/src/main/java/org/infinispan/factories/EmptyConstructorFactory.java
tristantarrant: In the above lines I sorted the fields because I had a hard time seeing which parameters were there.

wburns: Ah gotcha you were going for alphabetical order :) - I guess we should probably do that for these larger ones moving forward.

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
tristantarrant: I would say that maybe not all principals would map to a role. The case where a principal might map to multiple roles is unlikely in my experience.

core/src/main/java/org/infinispan/security/impl/CachePrincipalPair.java
wburns: That is an interesting question, but I guess not a big deal either way.

core/src/main/java/org/infinispan/security/impl/GlobalSecurityManagerFactory.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+@Scope(Scopes.GLOBAL)
+@DefaultFactoryFor(classes = GlobalSecurityManager.class)
+public class GlobalSecurityManagerFactory extends AbstractComponentFactory implements AutoInstantiableFactory {
+
+   @Override
+   @SuppressWarnings("unchecked")
+   public <T> T construct(Class<T> componentType) {
+      if (globalConfiguration.security().authorization().enabled())
+         return (T) new GlobalSecurityManagerImpl();
```
wburns: I would recommend ```componentType.cast(new GlobalSecurityManagerImpl());```  instead.

core/src/main/java/org/infinispan/security/impl/CachePrincipalPair.java
```diff
@@ -0,0 +1,50 @@
...
+   @Override
+   public boolean equals(Object obj) {
+      if (this == obj)
+         return true;
+      if (obj == null)
+         return false;
+      if (getClass() != obj.getClass())
+         return false;
+      CachePrincipalPair other = (CachePrincipalPair) obj;
+      if (!cacheName.equals(other.cacheName))
```
wburns: Actually the cacheName not being null becomes very important here, since it would cause a NPE.

wburns:
Just some minor things, otherwise looks good :+1:

core/src/main/java/org/infinispan/security/impl/CachePrincipalPair.java
tristantarrant: I guess we should be using some form of NotNull annotations, but the intention is that the cache name should never be null

tristantarrant:
@wburns I addressed your comments (unless where noted)

core/src/main/java/org/infinispan/security/impl/CachePrincipalPair.java
wburns: K that is what I figured.  In that case https://github.com/infinispan/infinispan/pull/3726/files#diff-a1e32821ed78ef03b09079b713d7c7bbR25 doesn't need the null check is all :)

danberindei:
@tristantarrant more test failures...

tristantarrant:
@danberindei only one test failure is security related, but it's to do with JGroups SASL authentication, so not connected to this PR.

danberindei:
Integrated at long last! Thanks Tristan!

PR #3729, opened 2015-09-28
ISPN-5796

Sanne:
Target WildFly 10.0.0.CR2 for built modules

https://issues.jboss.org/browse/ISPN-5796


tristantarrant:
Pushed, thanks @Sanne

PR #3732, opened 2015-10-01
ISPN-5413 Index Affinity support

gustavonalle:
https://issues.jboss.org/browse/ISPN-5413

Early draft. 

```AffinityTest``` illustrates the current issue of locks being held after rebalancing.


Sanne:
## Configuration

Looks like we need to discuss how this should be enabled in the configuration:
this draft is using the configuration property `hibernate.search.default.sharding_strategy` but this is the *default* one, some of the indexes might override the default.

But then you make the `boolean affinityEnabled` as a field of the `QueryInterceptor`, making it a global on/off switch it seems.

I think it's ok to make it a global switch, but the configuration should be done with a specific switch; we then have to automatically replace the normal configuration of `sharding_strategy` to use the custom logic, and also to get appropriate warnings in case the user is attempting to use a different sharding strategy on top.

gustavonalle:
@Sanne The configuration to enable/disable affinity (on a per cache basis) is ```infinispan.index.affinity``` set to ```enabled```.
This property will cause the proper shard strategy to be set and the field on the query interceptor derives from that.

+1, there is some validation to be done in case affinity is turned on and another shard strategy is chosen.

gustavonalle:
Will send another one

PR #3733, opened 2015-10-05
ISPN-5801 JCache metadata expiry should be absolute

galderz:
https://issues.jboss.org/browse/ISPN-5801

* As indicated in the JCache internal metadata instance variable
  comments, both expiry and created should be absolute, so fix expiry
  parameter which until now was being passed as relative.

### DIFF
--- a/jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheLoaderAdapter.java
+++ b/jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheLoaderAdapter.java
@@ -47,11 +47,11 @@ public class JCacheLoaderAdapter<K, V> implements org.infinispan.persistence.spi
       if (value != null) {
          Duration expiry = Expiration.getExpiry(expiryPolicy, Expiration.Operation.CREATION);
          long now = ctx.getTimeService().wallClockTime(); // ms
-         if (expiry.isEternal()) {
+         if (expiry == null || expiry.isEternal()) {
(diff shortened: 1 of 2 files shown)


jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheLoaderAdapter.java
```diff
@@ -50,8 +50,9 @@ public MarshalledEntry load(Object key) throws PersistenceException {
          if (expiry.isEternal()) {
             return ctx.getMarshalledEntryFactory().newMarshalledEntry(value, value, null);
          } else {
-            JCacheInternalMetadata meta = new JCacheInternalMetadata(now,
-                  expiry.getTimeUnit().toMillis(expiry.getDurationAmount()));
+            long created = now;
```
pruivo: nitpick: created variable is not needed.

jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheLoaderAdapter.java
```diff
@@ -50,8 +50,9 @@ public MarshalledEntry load(Object key) throws PersistenceException {
          if (expiry.isEternal()) {
```
pruivo: intellij is warning me that ```expiry``` may be ```null```

galderz:
@pruivo Fixed the comments

pruivo:
looks good to me. testing and pushing...

pruivo:
integrated! thanks @galderz !

PR #3734, opened 2015-10-05
ISPN-5822 Limit the number of expansion cofactors for hybrid query

anistor:
jira: https://issues.jboss.org/browse/ISPN-5822


query/src/main/java/org/infinispan/query/dsl/embedded/impl/QueryEngine.java
```diff
@@ -51,6 +51,8 @@
  */
 public class QueryEngine {
 
+   private static final int MAX_EXPANSION_COFACTORS = 16;
```
anistor: This is not configurable ATM. I thought there are too many knobs and levers already for the poor user to handle. :)

gustavonalle:
Integrated, thanks @anistor

PR #3736, opened 2015-10-05
ISPN-5691 Verify the locking isolation and write skew check

tristantarrant:
Ensure that the locking isolation level and the write skew check
configuration are compatible and supported

https://issues.jboss.org/browse/ISPN-5691


pruivo:
looks good @tristantarrant! testing and pushing...

pruivo:
integrated! thanks @tristantarrant !

PR #3742, opened 2015-10-06
ISPN-5833 ModuleCommandExtensions should not expose deprecated type E…

Sanne:
…xtendedModuleCommandFactory

https://issues.jboss.org/browse/ISPN-5833

### DIFF
--- a/core/src/main/java/org/infinispan/commands/module/ModuleCommandExtensions.java
+++ b/core/src/main/java/org/infinispan/commands/module/ModuleCommandExtensions.java
@@ -8,7 +8,7 @@ package org.infinispan.commands.module;
  */
 public interface ModuleCommandExtensions {
 
-   ExtendedModuleCommandFactory getModuleCommandFactory();
+   ModuleCommandFactory getModuleCommandFactory();
 
    ModuleCommandInitializer getModuleCommandInitializer();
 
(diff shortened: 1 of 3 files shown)


gustavonalle:
This was attempted in https://issues.jboss.org/browse/ISPN-5601 but it needed to be reverted because of binary compatibility. Not sure what was the issue, @wburns , @rvansa ?

rvansa:
Yep, in order to keep binary compatibility between 7 and 8 for ORM we had to keep that deprecated class :-/

wburns:
Yes, @gustavonalle @Sanne unfortunately we cannot make this change until ISPN 9 :-1: Closing this for now.

Sanne:
> we had to keep that deprecated class

I didn't delete the class. Do you mean you had to keep the interface as-is?

wburns:
@Sanne yes the return type has to stay the same even

Sanne:
ok, thanks.

alanfx:
@wburns or @Sanne, would it be worthwhile to add a comment to the interface about this issue as a warning to other users? It could also state that the class will be removed and the method signature will change in ISPN 9.

Sanne:
Good point @alanfx ! I've pushed a change to add some comments on the SPI.

Sanne:
(see 669d165c0d24286f8ad224ee64114d64be41d47c)

alanfx:
Thanks, @Sanne ! :+1:

Sanne:
It's time for pr #4517 ;-)

PR #3743, opened 2015-10-06
ISPN-5797 NullPointerException in ClientListenerRegistry under load

wburns:
* Made sure null value and metadata are supported

https://issues.jboss.org/browse/ISPN-5797

### DIFF
--- a/server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
+++ b/server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
@@ -183,8 +183,10 @@ class ClientListenerRegistry(configuration: HotRodServerConfiguration) extends L
       @CacheEntryExpired
       def onCacheEvent(event: CacheEntryEvent[Bytes, Bytes]) {
          if (isSendEvent(event)) {
-            val dataVersion = event.getMetadata.version().asInstanceOf[NumericVersion].getVersion
-            sendEvent(event.getKey, event.getValue, dataVersion, event)
+            sendEvent(event.getKey, event.getValue, Option(event.getMetadata)
+              .map(_.version().asInstanceOf[NumericVersion].getVersion)
+              .getOrElse(null.asInstanceOf[Long]), event)
+
          }
       }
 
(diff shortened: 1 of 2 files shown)


server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
```diff
@@ -183,8 +183,12 @@ class ClientListenerRegistry(configuration: HotRodServerConfiguration) extends L
       @CacheEntryExpired
       def onCacheEvent(event: CacheEntryEvent[Bytes, Bytes]) {
          if (isSendEvent(event)) {
-            val dataVersion = event.getMetadata.version().asInstanceOf[NumericVersion].getVersion
-            sendEvent(event.getKey, event.getValue, dataVersion, event)
+            sendEvent(event.getKey, event.getValue, if (event.getMetadata == null) {
```
gustavonalle: Minor: the 3rd param to ```sendEvent``` could be written in a more functional way as:

```
Option(event.getMetadata)
      .map(_.version().asInstanceOf[NumericVersion].getVersion)
      .getOrElse(null.asInstanceOf[Long])
```

It'd be nicer without the ugly casts, but this cast is around in several places. It assumes ```NumericVersion``` is always used, so I wonder if there's a valid use case for setting SIMPLE versioning scheme on the server. If so, it'd break in several places. Not related to this PR anyway.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
```diff
@@ -183,8 +183,12 @@ class ClientListenerRegistry(configuration: HotRodServerConfiguration) extends L
...
+               null.asInstanceOf[Long]
+            } else {
+               event.getMetadata.version().asInstanceOf[NumericVersion].getVersion
+            }, event);
```
gustavonalle: No need for ```;```

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
wburns: Cool, I will update it.   The issue is not with having versioning enabled or not.  The problem is that there are some cases where the metadata and value are not available from an expiration event and thus we have to handle those null values.

wburns:
Updated per comments.

wburns:
This needs to also be cherry-picked into 8.0.x branch

gustavonalle:
pulling

gustavonalle:
Integrated in master and 8.0.x, thanks @wburns !

PR #3745, opened 2015-10-07
ISPN-5839 Fix unfinished transaction log message

galderz:
https://issues.jboss.org/browse/ISPN-5839

* Stream.toString does not go and convert the Streams elements into a
  String. Instead, to convert a Stream to String, a String joiner needs
  to be used.

### DIFF
--- a/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
+++ b/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
@@ -58,6 +58,7 @@ import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
 
 /**
  * Repository for {@link RemoteTransaction} and {@link org.infinispan.transaction.xa.TransactionXaAdapter}s (locally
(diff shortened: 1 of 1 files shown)


galderz:
Do not integrate this yet, something else found...

galderz:
Actually, this particular issue fixed. So all good :)

core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
```diff
@@ -606,7 +607,8 @@ private void shutDownGracefully() {
          log.unfinishedTransactionsRemain(localTransactions == null ? 0 : localTransactions.size(),
                                           remoteTransactions == null ? 0 : remoteTransactions.size());
          if (trace) log.tracef("Unfinished local transactions: %s",
-                    localTransactions.values().stream().map(AbstractCacheTransaction::getGlobalTransaction));
+            localTransactions.values().stream().map(tx -> tx.getGlobalTransaction().toString())
+               .collect(Collectors.joining(", ")));
```
danberindei: I've noticed that the output is slightly different from the line below, because `Collection.toString` puts everything the element in square brackets. Could you change it to use `Collectors.toList()` instead?

galderz:
@danberindei Fixed

wburns:
@galderz Can't we just use https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#joining-java.lang.CharSequence-java.lang.CharSequence-java.lang.CharSequence- with the bracket before and after?  This way we don't have to create another collection just to make the string.

galderz:
@wburns Good idea, fixed it accordingly.

danberindei:
Integrated, thanks Galder!

PR #3747, opened 2015-10-07
ISPN-5843 Avoid usage of deprecated code in the Lucene Directory

Sanne:
- https://issues.jboss.org/browse/ISPN-5843


tristantarrant:
Pushed, thanks

PR #3750, opened 2015-10-09
ISPN-5826 Caches should be stopped in the right order in tests

gustavonalle:
https://issues.jboss.org/browse/ISPN-5826


anistor:
Integrated. Thanks!

PR #3751, opened 2015-10-12
ISPN-5851 Split GlobalState locations into persistent and temporary

tristantarrant:
https://issues.jboss.org/browse/ISPN-5851


core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
```diff
@@ -26,7 +26,7 @@
    private final ThreadPoolConfigurationBuilder stateTransferThreadPool;
    private final ThreadPoolConfigurationBuilder asyncThreadPool;
    private final ShutdownConfigurationBuilder shutdown;
-   private final GlobalStatePersistenceConfigurationBuilder statePersistence;
+   private final GlobalStateConfigurationBuilder statePersistence;
```
danberindei: Field has the old name.

core/src/main/resources/schema/infinispan-config-8.1.xsd
```diff
@@ -535,7 +535,27 @@
...
+          <xs:documentation>
+            Defines the path where global state for this cache-container will be stored. The data stored at this location is required for graceful
+            shutdown and restore.
+          </xs:documentation>
+        </xs:annotation>
+      </xs:element>
+      <xs:element name="swap-location" type="tns:global-state-path" minOccurs="0" maxOccurs="1">
+        <xs:annotation>
+          <xs:documentation>
+            Defines the path which will be used to store temporary swap data.
```
danberindei: What is "temporary swap data"?

core/src/main/java/org/infinispan/configuration/global/GlobalStateConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+
+   public boolean enabled() {
+      return enabled.get();
+   }
+
+   public String persistentLocation() {
+      return persistentLocation.get();
+   }
+
+   public String swapLocation() {
```
danberindei: Javadoc?

core/src/main/resources/schema/infinispan-config-8.1.xsd
tristantarrant: @Sanne is planning on adding memory-mapped lucene indexes. This is where they would be placed.

danberindei:
The documentation should really explain what the "persistent" and "swap" data is... The first thing it brings to my mind is passivation, but I don't think that's the case.

core/src/main/resources/schema/infinispan-config-8.1.xsd
Sanne: an index is not "swap" nor "temporary" though ;-)

Great to have a place to store, thanks!

tristantarrant: @Sanne in your plan for off-heap indexes you mentioned both permanent and temporary indexes, so that is the intention behind this. If I misunderstood please tell me otherwise.

gustavonalle: Side note: we do have off-heap indexes already: the Lucene filesystem directory memory maps the files, and the location is configurable (defaulting to /tmp)

tristantarrant: My recommendation is then to use `swap-location` (or whatever more appropriate name we come up with for this) for these indexes instead of 'java.io.tmpdir`. Since the default is to store them there it is because they are not supposed to be durable, as they can be regenerated from the cache data anyway.

gustavonalle: actually it's not on /tmp the default, but in the current directory. I think if the user opts to store the indexes on disk, they are supposed to be durable, re-generating it involves running the massindexer and it can be a long running process

tristantarrant: Ok, persistent state path then.

Sanne: +1

tristantarrant:
I have added javadocs explaining what would get stored where. I also added attribute initializers for both properties defaulting to user.dir and java.io.tmpdir respectively.

core/src/main/java/org/infinispan/configuration/global/GlobalStateConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+    */
+   public String persistentLocation() {
+      return persistentLocation.get();
+   }
+
+   /**
+    * Returns the filesystem path where temporary state should be stored. Defaults to the value of
+    * the java.io.tmpdir system property.
+    */
+   public String swapLocation() {
```
danberindei: I would prefer `temporaryLocation()` instead, it seems to me it's more like `/tmp` than a swap partition.

core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
danberindei: @tristantarrant Do you want to keep the old name here?

tristantarrant:
@danberindei  I have s/swap/temporary/

danberindei:
@tristantarrant Looks like you missed some property files:

java.lang.RuntimeException: "WFLYCTL0158: Operation handler failed: java.util.MissingResourceException: Can't find resource for bundle java.util.PropertyResourceBundle, key datagrid-infinispan.cache-container.global-state.temporary-location"

tristantarrant:
fixed and removed all leftovers of "swap"

danberindei:
Integrated, thanks Tristan!

PR #3752, opened 2015-10-12
ISPN-5805 Allow forEach consumer to have a Cache Injected

wburns:
ISPN-5845 Update local cache stream to allow for full wrapping

https://issues.jboss.org/browse/ISPN-5805
https://issues.jboss.org/browse/ISPN-5845
https://issues.jboss.org/browse/ISPN-5902
https://issues.jboss.org/browse/ISPN-5911


core/src/main/java/org/infinispan/stream/impl/termop/AbstractForEachOperation.java
```diff
@@ -3,6 +3,7 @@
 import org.infinispan.commons.util.ByRef;
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.container.entries.ImmortalCacheEntry;
+import org.infinispan.factories.ComponentRegistry;
```
danberindei: Unnecessary import. There are a couple more unused imports in the same file.

core/src/main/java/org/infinispan/util/CacheInjectable.java
```diff
@@ -0,0 +1,15 @@
+package org.infinispan.util;
+
+import org.infinispan.Cache;
+
+/**
+ * Interface that describes how a cache can be injected into another object.  This is useful for cases such as
+ * after an object is deserialized and you must inject a Cache into it.
+ */
+public interface CacheInjectable<K, V> {
```
danberindei: Interface name sounds a bit odd to me, how about `CacheAware`?

Also, `org.infinispan.util` looks like an implementation package to me, even if it's not marked as such... I think this would go better in `org.infinispan.stream`.

danberindei:
@wburns Getting closer to moving M/R over to streams?

core/src/main/java/org/infinispan/stream/impl/termop/AbstractForEachOperation.java
wburns: Sure

core/src/main/java/org/infinispan/util/CacheInjectable.java
wburns: I wasn't planning on putting it into stream as it is not really specific to that.  I was thinking we could utilize that for other purposes as well.  But I guess if we don't think we ever will that is fine.

wburns:
@danberindei it isn't really possible to move them over at least with a full translation without M/R being quite a bit slower in many cases :( - I believe we will just remove M/R with 8.3, right @tristantarrant ?

core/src/main/java/org/infinispan/util/CacheInjectable.java
wburns: I can change it over if you wanted.  We have used *Aware in quite a few places and they seem to have different connotations where as Injectable seemed clear to me, but I guess not for others.

wburns:
Updated

tristantarrant:
@wburns @danberindei dropping in 9.0

core/src/main/java/org/infinispan/stream/impl/local/LocalIntCacheStream.java
```diff
@@ -0,0 +1,219 @@
...
+
+   @Override
+   public IntStream skip(long n) {
+      intermediateOperations.add(new SkipIntOperation(n));
+      return this;
+   }
+
+   @Override
+   public void forEach(IntConsumer action) {
+      if (action instanceof CacheAware) {
```
tristantarrant: why not factor out this code ? It's repeated in quite a few places.

core/src/main/java/org/infinispan/stream/CacheAware.java
```diff
@@ -0,0 +1,15 @@
+package org.infinispan.stream;
+
+import org.infinispan.Cache;
+
+/**
+ * Interface that describes how a cache can be injected into another object.  This is useful for cases such as
+ * after an object is deserialized and you must inject a Cache into it.
+ */
+public interface CacheAware<K, V> {
```
tristantarrant: missing @since doclet

tristantarrant:
Just a couple of comments. Looks good otherwise

wburns:
Updated

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
```diff
@@ -255,6 +254,100 @@ public void testObjForEach() {
...
+
+      @Override
+      public void injectCache(Cache<Integer, String> cache) {
+         this.cache = cache;
+      }
+
+      @Override
+      public void accept(E entry) {
+         Cache<?, ?> cache = getForEachObject(cacheOffset);
+         if (cache != null && this.cache != null && cache.getName().equals(this.cache.getName())) {
```
danberindei: Why not throw an exception here if we didn't get the right cache?

wburns: We could do that.  I am not sure how the exception will propagate, but I guess we can find out :)

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
```diff
@@ -1413,6 +1598,59 @@ public void testDoubleForEachOrdered() {
...
+      int cacheOffset = populateNextForEachStructure(cache);
+      int offset = populateNextForEachStructure(new DoubleSummaryStatistics());
+      try {
+         createStream(entrySet).mapToDouble(toDouble).forEach(new ForEachDoubleInjected(cacheOffset, offset));
+         DoubleSummaryStatistics stats = getForEachObject(offset);
+         assertEquals(2.25, stats.getAverage());
+         assertEquals(0.0, stats.getMin());
+         assertEquals(4.5, stats.getMax());
+         assertEquals(10, stats.getCount());
+         assertEquals(22.5, stats.getSum());
```
danberindei: Couldn't you have a single method that can work either with a `ForEachDoubleInjected` or with a regular `DoubleConsumer`? You could then even use [TestNG parameters](http://testng.org/doc/documentation-main.html#parameters-dataproviders) to avoid creating multiple test methods.

wburns: Unfortunately the consumer class is different for each primitive and the object type.  I could try to consolidate some of the code though.

wburns: Sorry I misread at first.

wburns:
Updated to add fail messages if no matching cache.  Also I consolidated quite a few of the methods to use a new supplier based one to get rid of all of the boilerplate creation of cache stuff I could.

danberindei:
@wburns I'm getting these failures almost every time I run the HotRod client tests on my machine:

    ExpiryTest.testGlobalExpiryPutWithFlag:46->expectExpiryAfterRequest:99->expectCachedThenExpired:103 expected:<v1> but was:<null>
    ClientListenerWithIndexingAndProtobufTest.testEventFilter:78 expected:<1> but was:<0>
    RemoteListenerWithDslFilterTest.testEventFilter:135 expected:<3> but was:<0>

wburns:
@danberindei thanks, will take a look :)

wburns:
It doesn't seem that ExpiryTest is related.  It also failed on http://ci.infinispan.org/viewLog.html?buildId=31455, I wonder if it is because of reaper.  I am still looking at the others.

wburns:
Okay I have rewritten the type converter part now to properly unwrap only after the iterator method is invoked.  This should clean up those failures.  I have tried to break this up into a few commits.  I know there is quite a bit.

Please let a test run go by first, it passed locally, so I am hoping it is all good here.

wburns:
Tests all seem to be fine now.

core/src/main/java/org/infinispan/cache/impl/SimpleCacheImpl.java
```diff
@@ -1416,15 +1416,22 @@ public boolean addAll(Collection<? extends Entry<K, V>> c) {
...
 
       @Override
       public CacheStream<Entry<K, V>> parallelStream() {
-         return new LocalEntryCacheStream(SimpleCacheImpl.this, true, null, getStreamSupplier(true), getComponentRegistry());
+         return cacheStreamCast(new LocalCacheStream<>(new EntryStreamSupplier<>(SimpleCacheImpl.this, null,
+                 getStreamSupplier(false)), true, componentRegistry));
       }
    }
 
+   // This is a hack to allow the cast to work.  Java doesn't like subtypes in generics
```
danberindei: It does, if you write `CacheStream<? extends Entry<K, V>>` :)

core/src/main/java/org/infinispan/interceptors/MarshalledValueInterceptor.java
```diff
@@ -258,7 +258,9 @@ public Object visitReplaceCommand(InvocationContext ctx, ReplaceCommand command)
       return new AbstractDelegatingEntryCacheSet<K, V>(getCacheWithFlags(cache, command), set) {
          @Override
          public CloseableIterator<CacheEntry<K, V>> iterator() {
-            return new CloseableIteratorMapper<>(super.iterator(), e -> unwrapEntry(e, ctx));
+            // We pass a null ctx, since this is always local invocation - if it was remote it would use
+            // DistributionBulkInterceptor
```
danberindei: Doesn't this mean that reusing the `ctx` parameter would also be ok, because it's always local?

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
```diff
@@ -0,0 +1,66 @@
...
+import org.infinispan.container.InternalEntryFactory;
+import org.infinispan.container.entries.CacheEntry;
+import org.infinispan.stream.impl.AbstractDelegatingCacheStream;
+import org.infinispan.stream.impl.spliterators.IteratorAsSpliterator;
+
+import java.util.Iterator;
+import java.util.Spliterator;
+
+/**
+ * Delegating stream that converts variables.  Note this stream specifically doesn't specify generics as it can have
```
danberindei: s/variables/values/ ?

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
```diff
@@ -0,0 +1,66 @@
...
+public class TypeConverterStream extends AbstractDelegatingCacheStream {
+   public TypeConverterStream(CacheStream<?> stream, TypeConverter<Object, Object, Object, Object> converter,
+           InternalEntryFactory entryFactory) {
+      super(stream);
+      this.converter = converter;
+      this.entryFactory = entryFactory;
+   }
+
+   private final TypeConverter<Object, Object, Object, Object> converter;
+   private final InternalEntryFactory entryFactory;
```
danberindei: Fields should be first

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
```diff
@@ -0,0 +1,66 @@
...
+   @Override
+   public Iterator<Object> iterator() {
+      // Note that a transformation could have been applied intermediately which means this
+      // may not be a cache entry instance at all - so we have to define it as Object first
+      return new IteratorMapper<>(super.iterator(), (java.lang.Object e) -> {
+         if (e instanceof CacheEntry) {
+            return convert((CacheEntry<?, ?>) e, converter, entryFactory);
+         } else {
+            // If it isn't a CacheEntry just unbox as is
+            // TODO: should we worry about arrays?
```
danberindei: You mean `byte[]`? I don't see any reason to worry about them here...

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
```diff
@@ -0,0 +1,66 @@
...
+   @Override
+   public Spliterator<Object> spliterator() {
+      // We rely on our iterator to do unwrapping.
+      return new IteratorAsSpliterator.Builder<>(iterator())
+              .setEstimateRemaining(super.spliterator().estimateSize())
+              .setCharacteristics(Spliterator.CONCURRENT | Spliterator.DISTINCT | Spliterator.NONNULL)
+              .get();
+   }
+
+   // TODO: add the other terminal operators as they are needed
```
danberindei: Do you expect that we'll need more terminal operators?

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -166,7 +172,15 @@ public void dataRehashed(DataRehashedEvent<K, V> event) {
       Stream<CacheEntry<K, V>> stream = (parallelStream ? cacheEntrySet.parallelStream() : cacheEntrySet.stream())
               .filterKeys(keysToInclude).filterKeySegments(segments);
       if (!keysToExclude.isEmpty()) {
-         return stream.filter(e -> !keysToExclude.contains(e.getKey()));
+         // AnyEquivalence is how HashSet works so we don't need to worry then
+         if (!(keyEquivalence instanceof AnyEquivalence)) {
+            // We have to add all the keys into an equivalent hash set to make sure we are excluding them properly
```
danberindei: I'm not sure about this... I believe it should be the caller's job to make sure they used the proper collection type when calling `filterKeys()`.

core/src/main/java/org/infinispan/stream/impl/local/EntryStreamSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+import org.infinispan.stream.impl.RemovableCloseableIterator;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+
+import java.util.Set;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+/**
+ * Created by wburns on 10/9/15.
```
danberindei: Remove?

core/src/main/java/org/infinispan/stream/impl/local/KeyStreamSupplier.java
```diff
@@ -4,40 +4,40 @@
...
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.stream.BaseStream;
 import java.util.stream.Stream;
 
 /**
- * Local cache stream that returns a stream of cache keys
- * @param <K> key type of the cache entries and resulting stream
- * @param <V> value type of the cache entries
+ * Created by wburns on 10/9/15.
```
danberindei: Remove?

danberindei:
I added some more questions, mostly about comments...

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
wburns: Unfortunately this isn't possible with how the code is currently laid out.  And actually we don't allow marshalling the EquivalentHashSet or the various Equivalence classes.  We could add that later imo.

core/src/main/java/org/infinispan/cache/impl/SimpleCacheImpl.java
wburns: I can't define the return type as that though :/  The problem is that return type for Map wants Entry which we can't change.

core/src/main/java/org/infinispan/interceptors/MarshalledValueInterceptor.java
wburns: The wording may be confusing here I guess.  But the problem is that it can also be invoked remotely - determined by the terminal operation.  However we always want this unwrapped even for the remote side.

I can try to fix this a bit.

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
wburns: Someone could return Object[] or even Collection<Object> where each of the inner ones would need unboxing.

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
wburns: Unknown at this time.  It depends if we ever want an operation that it is using compatibility (HOTROD/MEMCACHED) to use other methods other than iterator really.

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
wburns: Sure

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
wburns: Yeah I was trying not to say values since I thought it might have caused confusion with the CacheEntry at the time since it does both key and values.  But I can change it.

core/src/main/java/org/infinispan/stream/impl/local/EntryStreamSupplier.java
wburns: Not really remove, but replace with proper comments :)

core/src/main/java/org/infinispan/stream/impl/local/KeyStreamSupplier.java
wburns: Not really remove, but replace with proper comments :)

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
danberindei: Ok

core/src/main/java/org/infinispan/cache/impl/SimpleCacheImpl.java
danberindei: Well yeah, but the problem is not with generics per se, it's with how the API type and the implementation type aren't the same :)

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
danberindei: You can say keys and values...

core/src/main/java/org/infinispan/cache/impl/SimpleCacheImpl.java
wburns: Yeah but we need the CacheEntry for the implementation for AdvancedCache to allow for operations based on metadata.  I didn't want to implement both ways to save on code, and not to mention another interceptor method.

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
wburns: It isn't specific to keys and values though, since this can be used for other types besides CacheEntry like just values or whatever a map method transformed it into  :)

wburns: Maybe ``Delegating stream that converts elements or CacheEntries`` ?

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
wburns: The more I think about this should be fine.  And tbh it isn't even possible since the only intermediate operations we allow for remote invocation based streams is using our keyfilter converter framework which we control.  I can just remove this.

core/src/main/java/org/infinispan/interceptors/compat/TypeConverterStream.java
wburns: Think I should just remove it?

wburns: I just removed this for now.

wburns:
Updated from comments.

danberindei:
Integrated, thanks Will!

PR #3753, opened 2015-10-13
ISPN-5803 Extract MarshalledValues in JCache

galderz:
https://issues.jboss.org/browse/ISPN-5803

* JCache cache store loader and writers should extract values from
  marshalled values so that cache store/writer implementations can deal
  with true types instead of wrapped ones.

### DIFF
--- a/jcache/commons/src/test/java/org/infinispan/jcache/util/InMemoryJCacheLoader.java
+++ b/jcache/commons/src/test/java/org/infinispan/jcache/util/InMemoryJCacheLoader.java
@@ -1,11 +1,10 @@
 package org.infinispan.jcache.util;
 
-import org.infinispan.commons.util.CollectionFactory;
-
 import javax.cache.integration.CacheLoader;
 import javax.cache.integration.CacheLoaderException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.LongAdder;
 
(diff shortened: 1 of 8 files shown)


jcache/embedded/src/main/java/org/infinispan/jcache/embedded/MarshalledValues.java
```diff
@@ -0,0 +1,22 @@
+package org.infinispan.jcache.embedded;
+
+import org.infinispan.marshall.core.MarshalledValue;
+
+/**
+ * Helper class for dealing with Mars
```
danberindei: I think Mars would need more than just one helper class ;)

jcache/commons/src/test/java/org/infinispan/jcache/util/InMemoryJCacheLoader.java
```diff
@@ -17,10 +16,22 @@
...
 
+   public InMemoryJCacheLoader(ConcurrentMap<K, V> store) {
+      this.store = store;
+   }
+
+   public static <K, V> InMemoryJCacheLoader<K, V> create() {
+      return create(new ConcurrentHashMap<>());
+   }
+
+   public static <K, V> InMemoryJCacheLoader<K, V> create(ConcurrentMap<K, V> store) {
```
danberindei: I'm not convinced the factory methods are necessary.

danberindei:
Besides the minor comments, I'm wondering if we shouldn't do this for core loaders/writers as well. The `CacheWriter` interface already deals with `MarshalledEntry`, which has separate `getKey()` and `getKeyBytes()` methods (and the same for values/metadata), so couldn't we make `MarshalledEntry.getKey()` always strip the `MarshalledValue` wrapper?

jcache/embedded/src/main/java/org/infinispan/jcache/embedded/MarshalledValues.java
wburns: A few nukes might be enough? :)

galderz: Hahah, I'll sort that out.

jcache/commons/src/test/java/org/infinispan/jcache/util/InMemoryJCacheLoader.java
galderz: Factory methods help with avoiding the need for multiple constructors which is a pain to deal with, particularly if subclassing.

galderz:
@danberindei Not sure I understand your question...

jcache/commons/src/test/java/org/infinispan/jcache/util/InMemoryJCacheLoader.java
danberindei: Personally I never found multiple constructors to be a nuisance when subclassing, especially if all constructors delegate to the same protected implementation. Sure, you may need to write some duplicate code in the subclass, but I really don't see how the static factory methods would help.

galderz:
Dan made a good point there, closing this PR to send a new one with a different fix.

PR #3754, opened 2015-10-13
ISPN-5852 Per-cache rebalance enable/disable

tristantarrant:
https://issues.jboss.org/browse/ISPN-5852
https://issues.jboss.org/browse/ISPN-5795

### DIFF
--- a/core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
+++ b/core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
@@ -54,6 +54,7 @@ public class ClusterCacheStatus implements AvailabilityStrategyContext {
    private volatile CacheTopology stableTopology;
    private volatile AvailabilityMode availabilityMode = AvailabilityMode.AVAILABLE;
    private volatile List<Address> queuedRebalanceMembers;
+   private volatile boolean rebalancingEnabled = true;
 
    private volatile RebalanceConfirmationCollector rebalanceConfirmationCollector;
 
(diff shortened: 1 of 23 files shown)


core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -692,6 +694,17 @@ public void setRebalanceEnabled(boolean enabled) {
       }
    }
 
+   /**
+    * Only performs rebalance if the per-cache rebalancing is enabled
+    */
+   public void conditionalRebalance() {
```
danberindei: `startQueuedRebalance` already checks if rebalancing is enabled, do we really need this method?

danberindei:
Looks great @tristantarrant, just 1 small comment.

alanfx:
@tristantarrant will there still be a way to globally enable/disable rebalancing or will each cache need to be enabled/disabled individually?

tristantarrant:
@alanfx the global flag remains. However if you disable a cache rebalance and then enable the global flag, that cache will still be in disabled state.

alanfx:
ok, thanks for the details @tristantarrant, and what's wrong with cappuccino for dessert?!

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: Ok. One question I have about that is that before checking if rebalancing is enabled, it might invoke `broadcastStableTopologyUpdate`. Is that ok, or should we shuffle things around so that the check comes first ?

danberindei: I think it's ok the way it is now, we want to broadcast the stable topology when a rebalance finishes, even if rebalancing is now disabled. We're doing it here because sometimes the stable topology can change without a rebalance (e.g. a node leaves in a replicated cache), and `startQueuedRebalance` happens to be called in all cases.

tristantarrant:
@danberindei made the change you suggested and added a commit for https://issues.jboss.org/browse/ISPN-5795

danberindei:
Excellent, thanks @tristantarrant !

PR #3756, opened 2015-10-13
ISPN-5855 RemoteContinuousQueryTest fails consistently

wburns:
* Make sure expiration checks pay attention to equality

https://issues.jboss.org/browse/ISPN-5855

### DIFF
--- a/core/src/main/java/org/infinispan/expiration/impl/ExpirationManagerImpl.java
+++ b/core/src/main/java/org/infinispan/expiration/impl/ExpirationManagerImpl.java
@@ -1,12 +1,11 @@
 package org.infinispan.expiration.impl;
 
 import net.jcip.annotations.ThreadSafe;
-
 import org.infinispan.Cache;
 import org.infinispan.commons.util.Util;
+import org.infinispan.commons.util.concurrent.jdk8backported.EquivalentConcurrentHashMapV8;
 import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.container.DataContainer;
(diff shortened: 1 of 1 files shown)


danberindei:
@wburns Perhaps we should have a test in core for this?

anistor:
This needs backporting to 8.0.x too

anistor:
I see plenty of unused imports in ExpirationManagerImpl.

wburns:
@danberindei The issue cannot be in core, unless we had a custom data container equivalence with objects that don't implement equality properly.  We would have to make a very contrived test that is quite a bit confusing just to test it.

@anistor The other imports were already there, I guess I can fix it as well.  Also I was thinking when whomever integrated they would cherry-pick into 8.0.x since it should apply instantly and then I don't have to update stuff twice for review comments

danberindei:
@wburns There are also some `ClassCastException`s in the integration tests: 
http://ci.infinispan.org/viewLog.html?buildId=30924&buildTypeId=bt9&guest=1

wburns:
@danberindei Sounds like a different issue.

wburns:
@danberindei Interestingly the other tests seemed to have failed on another PR as well: http://ci.infinispan.org/project.html?projectId=Infinispan&testNameId=623317564722301945&tab=testDetails  I am thinking we should log another PR for that one as this change wouldn't cause a CCE

anistor:
@wburns @danberindei I started to investigate those. They are not related. This PR LGTM.

anistor:
Integrated in master and 8.0.x. Thanks @wburns!

PR #3761, opened 2015-10-15
Fix default value of backup failure policy in xsd schema

mcimbora:


mcimbora:
Probably not worth creating JIRA for this, if you think otherwise, please let me know.

tristantarrant:
Merged, thanks

PR #3769, opened 2015-10-21
ISPN-5875 Server must unwrap PriviledgedActionExceptions

galderz:
* Unwrapping is necessary to discover the real cause and propagate that
  back to the client.


tristantarrant:
I pushed it to both master and 8.0.x

PR #3772, opened 2015-10-22
ISPN-4912: Switch REST store to better HTTP client library

atoulme:
See [ISPN-4912](https://issues.jboss.org/browse/ISPN-4912) for discussion

### DIFF
--- a/persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
+++ b/persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
@@ -1,26 +1,20 @@
 package org.infinispan.persistence.rest;
 
-import net.jcip.annotations.ThreadSafe;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.HashSet;
+import java.util.Locale;
+import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletionService;
+import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
+
 import org.apache.commons.codec.EncoderException;
 import org.apache.commons.codec.net.URLCodec;
-import org.apache.http.Header;
-import org.apache.http.HttpEntity;
-import org.apache.http.HttpHeaders;
-import org.apache.http.HttpHost;
-import org.apache.http.HttpResponse;
-import org.apache.http.HttpStatus;
-import org.apache.http.client.HttpClient;
-import org.apache.http.client.methods.HttpDelete;
-import org.apache.http.client.methods.HttpGet;
-import org.apache.http.client.methods.HttpPut;
-import org.apache.http.entity.ByteArrayEntity;
-import org.apache.http.entity.ContentType;
-import org.apache.http.impl.client.DefaultHttpClient;
-import org.apache.http.impl.conn.PoolingClientConnectionManager;
-import org.apache.http.params.BasicHttpParams;
-import org.apache.http.params.CoreConnectionPNames;
-import org.apache.http.params.HttpParams;
-import org.apache.http.util.EntityUtils;
 import org.infinispan.commons.configuration.ConfiguredBy;
 import org.infinispan.commons.util.Util;
 import org.infinispan.container.InternalEntryFactory;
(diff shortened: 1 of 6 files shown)


danberindei:
Looks good!

There is one test utility that was using the HttpClient and now has a compilation error: 
`server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTHelper.java`

I think it can keep using HttpClient, we just have to add it as a dependency for the integration tests.

atoulme:
Will fix and update the PR. Thanks.

atoulme:
Updated. Please run the tests again?

danberindei:
@atoulme The tests run automatically, you can see the status is still "All checks have failed", but this time there are only 3 test failures :)

@galderz could you also review?

persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
```diff
@@ -316,7 +380,7 @@ public Void call() throws Exception {
                   if (!loadEntry || !loadMetadata) {
                      entry = ctx.getMarshalledEntryFactory().newMarshalledEntry(key,
                            loadEntry ? entry.getValue() : null,
-                           loadMetadata ? entry.getMetadata() : null);
+                                 loadMetadata ? entry.getMetadata() : null);
```
tristantarrant: spurious whitespace

persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
```diff
@@ -257,51 +313,59 @@ else if (timeout > 0 && timeout < 1000)
...
-         for (String stringKey = reader.readLine(); stringKey != null; stringKey = reader.readLine()) {
-            Object key = key2StringMapper.getKeyMapping(stringKey);
-            if (keyFilter == null || keyFilter.accept(key))
-               entries.add(key);
-            if (entries.size() == batchSize) {
-               final Set<Object> batch = entries;
-               entries = new HashSet<Object>(batchSize);
-               submitProcessTask(cacheLoaderTask, eacs, taskContext, batch, loadValue, loadMetadata);
+         try {
+            BufferedReader reader = new BufferedReader(new InputStreamReader(new ByteBufInputStream(((FullHttpResponse) handler.getResponse()).content())));
```
tristantarrant: The code you replaced specified a specific encoding. Is this not required here ?

tristantarrant:
@atoulme some minor comments. Otherwise looks good.

galderz:
Looks good to me, @tristantarrant's question WRT content encoding need answering.

My only question would be whether you looked into other HTTP clients, such as OkHttp. They even have benchmark that can compare Netty vs OkHttp [here](https://github.com/square/okhttp/blob/master/benchmarks/src/main/java/com/squareup/okhttp/benchmarks/NettyHttpClient.java). I'm not proposing that Netty is replaced by OkHttp, but if any testing/comparison/benchmarking has been done for this task, it'd be good to find out more details.

slaskawi:
I did some talk with Wildfly Team members about choosing an embedded REST library. 

From the comparison they did, it turned out that Undertow implementation was the fastest and there were very little differences between its competitors - Netty and Jetty. Similar results might be found on [TechEmpower](https://www.techempower.com/benchmarks/#section=data-r12&hw=peak&test=query&l=3k&p=5179l-dggb9e-0&o=4) portal. 

Having this in mind - I would vote for doing some benchmark before accepting this (or any other) PR which changes REST embedded server. 

@atoulme @tristantarrant @galderz What do you think about it? If we want to do such comparison and such change - this is the perfect moment (we're at the beginning of ISPN9 loop). If we don't want to do that - let's close this PR.

danberindei:
@slaskawi This PR isn't about the REST server, it's about the HTTP client library that the REST store uses to connect to the server. For the server side I think we're going to stay with RESTEasy over Netty, although we have encountered some problems there as well (e.g. https://issues.jboss.org/browse/RESTEASY-1205).

Since our use case is pretty standard, I don't think we need to run our own benchmarks, we just need to find some published benchmarks. Unfortunately @galderz's link to the OkHttp benchmark rotted away :)

persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
atoulme: The BufferedReader still needs to use an encoding. I'll bring it back.

atoulme:
I addressed @tristantarrant's feedback.
I didn't find a good benchmark for okHttp vs netty, but I didn't look very deep. @galderz, would you have better links?

I will look at test results and will try to address them. IIRC last time I couldn't figure it out by myself though, so I'll shout for help if needed.

slaskawi:
@danberindei Thanks for clarification.

slaskawi:
:+1: from me.

@tristantarrant @galderz @danberindei Any other comments? Or this one can be merged into ISPN 9?

galderz:
@danberindei The benchmark is still there but has been moved slightly, see [here](https://github.com/square/okhttp/blob/master/benchmarks/src/main/java/okhttp3/benchmarks/NettyHttpClient.java)

slaskawi:
Hi @atoulme ! We got rid of the most test failures in master branch, so could you please rebase this PR? We'd like to check if it doesn't introduce new failures.

slaskawi:
Update PR might be found here: https://github.com/infinispan/infinispan/pull/4214

atoulme:
Thanks @slaskawi!

PR #3773, opened 2015-10-22
ISPN-5803 Cache loaders/stores key/value params are unwrapped

galderz:
https://issues.jboss.org/browse/ISPN-5803

* Instead of having the JCache loader/writer adapters deal with
  MarshalledValue handling, make the persistence manager and
  MarshalledEntryImpl deal with the complexity of providing cache store
  implementations with the expected key/value types.
* Key instances passed directly to cache stores are now user types as
  opposed to: user types or marshalled values.
* MarshalledEntryImpl now unwraps key/value instances from
  MarshalledValue if callin getKey()/getValue() but such operations
  should normally not be called by cache stores since they normally
  should deal with byte array format, which is accessible via
  getKeyBytes()/getValueBytes and those do no
  unwrapping.
* DummyCacheStore has been fixed to avoid unnecessary
  serialization/deserialization by calling getValueBytes() in
  MarshalledEntry instead of getValue().

### DIFF
--- a/core/src/main/java/org/infinispan/filter/CollectionKeyFilter.java
+++ b/core/src/main/java/org/infinispan/filter/CollectionKeyFilter.java
@@ -3,6 +3,7 @@ package org.infinispan.filter;
 import org.infinispan.commons.marshall.AbstractExternalizer;
 import org.infinispan.commons.util.Util;
 import org.infinispan.marshall.core.Ids;
+import org.infinispan.marshall.core.MarshalledValue;
 
 import java.io.IOException;
 import java.io.ObjectInput;
@@ -33,7 +34,11 @@ public class CollectionKeyFilter<K> implements KeyFilter<K> {
 
    @Override
    public boolean accept(K key) {
-      return accept ? keys.contains(key) : !keys.contains(key);
+      // This filter is provided as callback to cache stores, so it should be
+      // able to take normal keys and verify them against seen keys which
+      // might be MarshalledValues.
+      boolean contains = keys.stream().filter(k -> k.equals(key) || MarshalledValue.unwrap(k).equals(key)).count() > 0;
+      return accept ? contains : !contains;
    }
 
    public static class Externalizer extends AbstractExternalizer<CollectionKeyFilter> {
--- a/core/src/main/java/org/infinispan/notifications/cachelistener/event/impl/EventImpl.java
+++ b/core/src/main/java/org/infinispan/notifications/cachelistener/event/impl/EventImpl.java
@@ -72,8 +72,6 @@ public class EventImpl<K, V> implements CacheEntryActivatedEvent, CacheEntryCrea
    @Override
    @SuppressWarnings("unchecked")
    public K getKey() {
-      if (key instanceof MarshalledValue)
-         key = (K) ((MarshalledValue) key).get();
       return key;
    }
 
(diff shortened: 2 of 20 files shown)


galderz:
@danberindei @wburns FYI, a reimplementation

danberindei:
There are some new test failures in some of the stores:
http://ci.infinispan.org/viewLog.html?buildId=31291&buildTypeId=bt9&guest=1

core/src/main/java/org/infinispan/filter/CollectionKeyFilter.java
```diff
@@ -33,7 +34,11 @@ public CollectionKeyFilter(Collection<? extends K> keys, boolean accept) {
 
    @Override
    public boolean accept(K key) {
-      return accept ? keys.contains(key) : !keys.contains(key);
+      // This filter is provided as callback to cache stores, so it should be
+      // able to take normal keys and verify them against seen keys which
+      // might be MarshalledValues.
+      boolean contains = keys.stream().filter(k -> k.equals(key) || MarshalledValue.unwrap(k).equals(key)).count() > 0;
```
danberindei: I'm a bit worried about performance here, as state transfer uses the `CollectionKeyFilter` and the collection is the data container itself. 

It's weird that `AdvancedCacheLoader.process()` can skip the unmarshalling for values, but not for keys... I'm not sure we can avoid unnecessary marshalling/unmarshalling with the current API, but I think keeping `keys.contains(key)` is more important.

core/src/main/java/org/infinispan/notifications/cachelistener/event/impl/EventImpl.java
```diff
@@ -175,7 +171,7 @@ public V getOldValue() {
    }
 
    public void setValue(V value) {
-      this.value = value;
+      this.value = MarshalledValue.unwrap(value);
```
danberindei: I like the lazy approach here, because the value might not be accessed by the listener. Or there may even be no listeners attached...

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -411,7 +412,7 @@ public boolean deleteFromAllStores(Object key, AccessMode mode) {
          boolean removed = false;
          for (CacheWriter w : writers) {
             if (mode.canPerform(configMap.get(w))) {
-               removed |= w.delete(key);
+               removed |= w.delete(MarshalledValue.unwrap(key));
```
danberindei: I guess this is one more place where a key-agnostic `CacheWriter` could have avoided a deserialization + serialization by dealing with the `MarshalledValue` directly. I'm starting to have some doubts about this change...

core/src/test/java/org/infinispan/persistence/dummy/DummyInMemoryStore.java
```diff
@@ -318,7 +320,11 @@ public boolean contains(Object key) {
 
    private byte[] serialize(MarshalledEntry o) {
       try {
-         return marshaller.objectToByteBuffer(new KeyValuePair(o.getValue(), o.getMetadata()));
+         // Retrieve value as bytes instead of value to avoid unnecessary
+         // unwrapping/deserialization if the value is a MarshalledValue,
+         // e.g. storeAsBinary as been enabled, and hence the value might be
```
danberindei: s/as/has/

galderz:
Dan and I discussed this and we were not fully happy with this solution. We came to the conclusion that the best way to implement this while not making too complex or avoid unnecessary serialization/deserialization would be to add overloaded methods that work with MarshalledValue instances instead of K, and with default implementations that unmarshall key. I'm closing this for now.

PR #3774, opened 2015-10-22
ISPN-5819 Add global backup site ops to container

pruivo:
Operations added:
* takeSiteOffline
* bringSiteOnline
* pushState
* cancelPushState

I'm open to suggestions about the SiteStatement in CLI :) 

https://issues.jboss.org/browse/ISPN-5819

### DIFF
--- a/cli/cli-interpreter/src/main/java/org/infinispan/cli/interpreter/statement/SiteStatement.java
+++ b/cli/cli-interpreter/src/main/java/org/infinispan/cli/interpreter/statement/SiteStatement.java
@@ -1,9 +1,5 @@
 package org.infinispan.cli.interpreter.statement;
 
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-
 import org.infinispan.Cache;
 import org.infinispan.cli.interpreter.logging.Log;
 import org.infinispan.cli.interpreter.result.EmptyResult;
@@ -11,8 +7,14 @@ import org.infinispan.cli.interpreter.result.Result;
 import org.infinispan.cli.interpreter.result.StatementException;
 import org.infinispan.cli.interpreter.result.StringResult;
 import org.infinispan.cli.interpreter.session.Session;
+import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.util.logging.LogFactory;
+import org.infinispan.xsite.GlobalXSiteAdminOperations;
 import org.infinispan.xsite.XSiteAdminOperations;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Performs operation related to Cross-Site Replication
@@ -23,8 +25,9 @@ import org.infinispan.xsite.XSiteAdminOperations;
 public class SiteStatement implements Statement {
    private static final Log log = LogFactory.getLog(SiteStatement.class, Log.class);
 
-   static enum Options {
-      OFFLINE, ONLINE, STATUS, PUSH, CANCELPUSH, CANCELRECEIVE, PUSHSTATUS, CLEARPUSHSTATUS, SENDINGSITE
+   enum Options {
+      OFFLINE, ONLINE, STATUS, PUSH, CANCELPUSH, CANCELRECEIVE, PUSHSTATUS, CLEARPUSHSTATUS, SENDINGSITE,
+      ONLINEALL, OFFLINEALL, PUSHALL, CANCELPUSHALL
    }
 
    final private SiteData siteData;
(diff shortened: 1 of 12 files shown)


cli/cli-interpreter/src/main/java/org/infinispan/cli/interpreter/statement/SiteStatement.java
```diff
@@ -23,8 +25,9 @@
 public class SiteStatement implements Statement {
    private static final Log log = LogFactory.getLog(SiteStatement.class, Log.class);
 
-   static enum Options {
-      OFFLINE, ONLINE, STATUS, PUSH, CANCELPUSH, CANCELRECEIVE, PUSHSTATUS, CLEARPUSHSTATUS, SENDINGSITE
+   enum Options {
+      OFFLINE, ONLINE, STATUS, PUSH, CANCELPUSH, CANCELRECEIVE, PUSHSTATUS, CLEARPUSHSTATUS, SENDINGSITE,
+      ONLINEALL, OFFLINEALL, PUSHALL, CANCELPUSHALL
```
danberindei: Why no underline between words, e.g. `CANCEL_PUSH`?

core/src/main/java/org/infinispan/xsite/GlobalXSiteAdminOperation.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+@Scope(Scopes.GLOBAL)
+@MBean(objectName = "ContainerXSiteAdmin", description = "Exposes tooling for handling backing up data to remote sites.")
+public class GlobalXSiteAdminOperation {
+
+   public static final String CACHE_DELIMITER = ",";
+
+   private EmbeddedCacheManager cacheManager;
+
+   private static void addIfBackupRemotely(Cache cache, List<PerCacheXSiteAdminOperation> list) {
```
danberindei: I think there should be an 'operation' in the name, maybe `addBackupOperation`?

core/src/main/java/org/infinispan/xsite/GlobalXSiteAdminOperation.java
```diff
@@ -0,0 +1,128 @@
...
+      List<PerCacheXSiteAdminOperation> operations = new ArrayList<>(cacheNames.size() + 1);
+      for (String cacheName : cacheNames) {
+         addIfBackupRemotely(cacheManager.getCache(cacheName, false), operations);
+      }
+      addIfBackupRemotely(cacheManager.getCache(), operations);
+      return operations;
+   }
+
+   private interface Operation {
+      String execute(XSiteAdminOperations operations);
```
danberindei: Executing an operation on an `operations` object is a bit too meta here, how about renaming the parameter (and the other local variables) to `mbeans`?

core/src/main/java/org/infinispan/xsite/GlobalXSiteAdminOperation.java
```diff
@@ -0,0 +1,128 @@
...
+ * A per-container (cache manager) cross-site admin operations.
+ * <p>
+ * All the operations invoked in this class will be applied to all caches which remotely backups its data.
+ *
+ * @author Pedro Ruivo
+ * @since 8.1
+ */
+@Scope(Scopes.GLOBAL)
+@MBean(objectName = "ContainerXSiteAdmin", description = "Exposes tooling for handling backing up data to remote sites.")
+public class GlobalXSiteAdminOperation {
```
danberindei: The object name and class name should probably match.

core/src/main/java/org/infinispan/xsite/GlobalXSiteAdminOperation.java
```diff
@@ -0,0 +1,128 @@
...
+      }
+      addIfBackupRemotely(cacheManager.getCache(), operations);
+      return operations;
+   }
+
+   private interface Operation {
+      String execute(XSiteAdminOperations operations);
+   }
+
+   private static class PerCacheXSiteAdminOperation {
```
danberindei: s/PerCache/Cache/ ?

core/src/test/java/org/infinispan/xsite/GlobalXSiteAdminOpsTest.java
```diff
@@ -0,0 +1,427 @@
...
+         public void defineConfiguration() {
+            ConfigurationBuilder builder = newConfiguration();
+            builder.sites().addBackup().site(siteName(1)).strategy(BackupConfiguration.BackupStrategy.SYNC);
+            builder.sites().addBackup().site(siteName(2)).strategy(BackupConfiguration.BackupStrategy.SYNC);
+            defineInSite(site(0), cacheName(), builder.build());
+            defineInSite(site(1), backupCacheName(), newConfiguration().build());
+            defineInSite(site(2), backupCacheName(), newConfiguration().build());
+         }
+      });
+      caches.put(CacheType.NO_BACKUP, new CacheConfiguration("cache4") {
```
danberindei: Since you always have the same 3 keys, wouldn't it be simpler to have a separate method for each?

danberindei:
I added some comments, @tristantarrant could you have a closer look at the SiteStatement?

cli/cli-interpreter/src/main/java/org/infinispan/cli/interpreter/statement/SiteStatement.java
pruivo: to be consistent with the other statements

core/src/test/java/org/infinispan/xsite/GlobalXSiteAdminOpsTest.java
pruivo: I think it would be more confusing. in this test, the cache name is the same in the all the sites, but in ```GlobalXSiteAdminOpsWithBackupForTest```, we have a different cache name in the backup sites.

danberindei: If that's the only difference, do we really need both tests?

pruivo: fixed!

tristantarrant:
@pruivo This is missing cache-container-level operations in the server DMR.

pruivo:
@tristantarrant isn't enough to add to ```SiteCommandHandler```?(here: https://github.com/infinispan/infinispan/pull/3774/files#diff-8cc58cf126142aaf53de827dc0f9f047R39). 
Where do I need to add then?

tristantarrant:
No @pruivo, it needs to be done like https://github.com/infinispan/infinispan/blob/master/server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/BackupSiteResource.java#L84

pruivo:
@tristantarrant and @danberindei updated! I've added the operations to the ```CacheContainerResource``` instead of ```BackupSiteResource``` since this operations are per container and not per cache. let me know what you think.

danberindei:
@tristantarrant I'll leave this one for you to review

tristantarrant:
Pushed, thanks

PR #3775, opened 2015-10-23
ISPN-5838 CDI split into remote and embedded

slaskawi:
Hey!

Please take a look at proposed solution for https://issues.jboss.org/browse/ISPN-5838

Highlights:
* CDI has been split into 3 modules - common, remote and embedded
* Common module is not designed for distribution. It's sources are included into remote and embedded module (via `build-helper-maven-plugin` - it adds another source directory to the maven build)
* This is not an *ideal* solution, because if client includes both jars - Weld will still complain about class duplicates (common part). However those complains will be harmless to the application. Unfortunately this was a trade-off for being backwards compatible (an *ideal* solution would require changing java packages (e.g. `org.infinispan.cdi.remote` and `org.infinispan.cdi.embedded`) which is not backwards compatible).
* There is also an open question regarding to OSGi manifests. Currently both embedded and remote jar exports `org.infinispan.cdi` package. OSGi bundle consumer might need to use `RequireBundle` in their manifests opposed to `RequirePackage`.
* `InfinispanExtension` class has been removed

There are lots of small changes in imports - they should be valid according to the latest IDE configuration files (I've re-imported IDE settings before doing this refactoring).

Thanks!


tristantarrant:
@slaskawi I suggest you also open a Jira to track the "ideal" fix for the class duplicates

slaskawi:
CI failures are not related

cdi/embedded/src/main/java/org/infinispan/cdi/InfinispanExtensionEmbedded.java
```diff
@@ -39,8 +39,8 @@
  * @author Pete Muir
  * @author Kevin Pollet <kevin.pollet@serli.com> (C) 2011 SERLI
  */
-public class InfinispanExtensionEmbedded {
-   private static final Log log = LogFactory.getLog(InfinispanExtensionEmbedded.class, Log.class);
+public class InfinispanExtensionEmbedded implements Extension {
+   private static final EmbeddedLog LOG = LogFactory.getLog(InfinispanExtensionEmbedded.class, EmbeddedLog.class);
```
danberindei: We never capitalize the log field, I guess references to `LOG` in the code tend to distract from the actual logic.

cdi/pom.xml
```diff
@@ -1,270 +1,48 @@
 <?xml version='1.0' encoding='UTF-8'?>
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+         xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
-   <modelVersion>4.0.0</modelVersion>
+    <modelVersion>4.0.0</modelVersion>
```
danberindei: Unnecessary indentation changes

danberindei:
@tristantarrant @slaskawi Can we say we're preserving backwards compatibility if we require the users to depend on a different Jar?

Other than that, I only have minor comments related to formatting. And I'm fine with the import order changes :)

slaskawi:
@tristantarrant An ideal solution tracked by https://issues.jboss.org/browse/ISPN-5900

cdi/embedded/src/main/java/org/infinispan/cdi/InfinispanExtensionEmbedded.java
slaskawi: done

cdi/pom.xml
slaskawi: done

tristantarrant:
@danberindei this would not impact users of the infinispan-embedded uber jar and I believe the backwards incompatibility is sufferable (i.e. no source or binary incompatibility).
@slaskawi please add a comment to documentation/src/main/asciidoc/upgrading/upgrading.adoc explaining this

slaskawi:
@tristantarrant done

danberindei:
Integrated, thanks Sebastian!

PR #3779, opened 2015-10-23
ISPN-5465 Introduce the KeyPartitioner interface

danberindei:
https://issues.jboss.org/browse/ISPN-5465
https://issues.jboss.org/browse/ISPN-5878


commons/src/main/java/org/infinispan/commons/marshall/Ids.java
```diff
@@ -8,7 +8,7 @@
  */
 public interface Ids {
    int MURMURHASH_2 = 71;
-   int MURMURHASH_2_COMPAT = 72;
+   // 72 used to be MURMURHASH_2_COMPAT
```
wburns: Is there any reason we are removing this?  I am kinda curious from an Id perspective as well?  Should we remove ```MURMURHASH_2``` as well then?

core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
```diff
@@ -374,7 +374,11 @@ public void handleRebalance(final String cacheName, final CacheTopology cacheTop
       cacheStatus.getTopologyUpdatesCompletionService().submit(new Runnable() {
          @Override
          public void run() {
-            doHandleRebalance(viewId, cacheStatus, cacheTopology, cacheName, sender);
+            try {
+               doHandleRebalance(viewId, cacheStatus, cacheTopology, cacheName, sender);
+            } catch (Throwable t) {
+               log.errorf(t, "Error handling rebalance request %s", cacheTopology);
```
wburns: Is this what we want here?  If so we should have it as a Log method then.

core/src/test/java/org/infinispan/stress/LargeClusterStressTest.java
```diff
@@ -33,7 +33,7 @@
 public class LargeClusterStressTest extends MultipleCacheManagersTest {
 
    private static final int NUM_NODES = 40;
-   private static final int NUM_CACHES = 80;
+   private static final int NUM_CACHES = 2;
```
wburns: Was this intentional?

core/src/main/resources/schema/infinispan-config-8.1.xsd
```diff
@@ -1160,15 +1160,28 @@
...
-                  The factory to use for generating the consistent hash. Must implement org.infinispan.distribution.ch.ConsistentHashFactory.
-                  This option can be used to guarantee that multiple distributed caches use exactly the same consistent hash,
-                  which for performance reasons is not guaranteed by the default consistent hash factory instance used.
-               </xs:documentation>
-            </xs:annotation>
-         </xs:attribute>
+        <xs:attribute name="consistent-hash-factory" type="xs:string">
+          <xs:annotation>
+            <xs:documentation>
+              The factory to use for generating the consistent hash.
```
wburns: Maybe we should add text saying this deprecated here as well.

wburns:
Looks good to me, just a few minor comments.

commons/src/main/java/org/infinispan/commons/marshall/Ids.java
danberindei: I'm removing the id constant because AFAIK that's what we've always done when we've removed a class/externalizer...

But MurmurHash2 wasn't deprecated, so we can't remove it yet (nor its id).

core/src/main/resources/schema/infinispan-config-8.1.xsd
danberindei: Except it's not :)
The hash function that I've deprecated was never exposed in the XML configuration.

core/src/test/java/org/infinispan/stress/LargeClusterStressTest.java
danberindei: Nope, I was testing something else at the time :)

danberindei:
@wburns Updated, but I'd like to wait until @gustavonalle also has a chance to look over it - he's the one who needs this for his index affinity work.

core/src/main/resources/schema/infinispan-config-8.1.xsd
wburns: Ah I see there was a consistentHashFactory and a hash instance

commons/src/main/java/org/infinispan/commons/marshall/Ids.java
tristantarrant: Now is a good time to deprecated MH2, I guess

danberindei: And I did :)

gustavonalle:
@danberindei I'm wondering the effect of using a custom KeyPartitioner for HotRod clients, that currently are wired to use the hash to locate segments for keys [1]. Thoughts?

[1] http://infinispan.org/docs/8.1.x/user_guide/user_guide.html#_hot_rod_hash_functions

danberindei:
@gustavonalle You're right, the HotRod client would have to access to the same custom partitioner that's defined in the server, or it won't work. Additionally, the server partitioner would see either the key either unmarshalled or marshalled as a byte[], depending on whether compatibility is enabled, and the client would have to use it on the same version of the key, or they won't match. I'm not sure how to handle that yet, as this PR was meant only for embedded mode.

core/src/main/java/org/infinispan/configuration/cache/HashConfiguration.java
```diff
@@ -138,6 +148,10 @@ public float capacityFactor() {
       return capacityFactor.get();
    }
 
+   public KeyPartitionerFactory keyPartitionerFactory() {
+      return keyPartitionerFactory.get();
+   }
+
```
gustavonalle: Can this config be mutable? Main reason is to set a custom partitioner in a ```cacheStarting``` method of a ```LifecycleManager```

gustavonalle:
@danberindei sounds good, just one question.
+1 for addressing the Hot Rod issue in another PR

core/src/main/java/org/infinispan/configuration/cache/HashConfiguration.java
danberindei: I wouldn't make it mutable, because changing it after the cache was started would make keys inaccessible - we'd need to do some rebalancing to handle this without data loss.  TBH I'm not a fan of modules modifying the configuration in general, I believe the configuration should always match what the user set.

In your case, perhaps a better solution would be to register the `KeyPartitionerFactory` (or the `KeyPartitioner`) in the component registry, and allow modules to override it there.

gustavonalle: ```cacheStarting``` is called right before the cache starts, so I think this should be fine?

gustavonalle: But I agree that for other cases this may be dangerous...
Will check the component registry stratgey

gustavonalle:
+1 for registering the partitioner in the component registry so that it can be overriden

core/src/main/java/org/infinispan/configuration/parsing/Parser80.java
```diff
@@ -1538,7 +1539,15 @@ private void parseInvalidationCache(XMLExtendedStreamReader reader, Configuratio
       for (int i = 0; i < reader.getAttributeCount(); i++) {
          String value = replaceProperties(reader.getAttributeValue(i));
          Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
-         this.parseClusteredCacheAttribute(reader, i, attribute, value, builder, baseCacheMode);
+         switch (attribute) {
+            case KEY_PARTITIONER_FACTORY: {
+               builder.clustering().hash().keyPartitionerFactory(Util.getInstance(value, holder.getClassLoader()));
```
tristantarrant: if (reader.getSchema().since(8, 2)) {
builder.clustering().hash().keyPartitionerFactory(Util.getInstance(value, holder.getClassLoader()));
} else {
throw ParseUtils.unexpectedAttribute(reader,i);
}

core/src/main/resources/schema/infinispan-config-8.1.xsd
```diff
@@ -1126,6 +1126,17 @@
...
           </xs:element>
         </xs:sequence>
+        <xs:attribute name="key-partitioner-factory" type="xs:string">
+          <xs:annotation>
+            <xs:documentation>
+              The factory to use for generating key partitioners.
+              Must implement `org.infinispan.distribution.ch.KeyPartitionerFactory`.
+              This option can be used as an alternative to grouping, to guarantee that some keys are located
+              in the same segment (and thus their primary owner is the same node).
+              Since 8.1.
```
tristantarrant: 8.2 now :)

core/src/main/java/org/infinispan/configuration/parsing/Parser80.java
tristantarrant: Also, for symmetry, server should also parse this attribute. Should we allow users to specify a custom keypartitioner on server or should we ignore it ?

gustavonalle: As discussed earlier, this PR will only cover library, we need to decide on how to support that since the HotRod client would need to use that same function

danberindei:
Finally updated! I hope the configuration change doesn't break your use case, @gustavonalle!

gustavonalle:
Thanks @danberindei !
I think your changes won't affect (much), let me test

core/src/main/java/org/infinispan/configuration/cache/HashConfigurationBuilder.java
```diff
@@ -158,6 +161,19 @@ public HashConfigurationBuilder capacityFactor(float capacityFactor) {
       return this;
    }
 
+   /**
+    * Key partitioner factory, controlling the mapping of keys to hash segments.
+    * <p>
+    * The default implementation {@code org.infinispan.distribution.ch.impl .HashFunctionPartitionerFactory},
+    * uses the hash function configured via {@link #hash(Hash)}. Future versions may ignore the hash function.
+    *
```
gustavonalle: outdated javadoc

danberindei:
Javadoc updated, including some outdated "@since 8.1" tags.

core/src/main/java/org/infinispan/distribution/ch/impl/KeyPartitionerFactory.java
```diff
@@ -0,0 +1,45 @@
...
+ * In the future, we will probably remove the hash function from the configuration and leave only the
+ * key partitioner.
+ *
+ * @author Dan Berindei
+ * @since 8.2
+ */
+@DefaultFactoryFor(classes = KeyPartitioner.class)
+public class KeyPartitionerFactory extends AbstractNamedCacheComponentFactory
+      implements AutoInstantiableFactory {
+   private KeyPartitionerFactory defaultPartitionerFactory;
```
gustavonalle: This field does not seem to be used

danberindei: Removed, thanks for catching it!

gustavonalle:
rebased and pushed, thanks @danberindei !

PR #3780, opened 2015-10-24
fixed an error in the code example of Functional Map

kazuhira-r:
Code example that uses the metadata of the Functional Map is, I think than wrong?


galderz:
Integrated, thanks @kazuhira-r :)

PR #3784, opened 2015-10-26
ISPN-5886 Upgrade WildFly to 10.0.0.CR4

tristantarrant:
https://issues.jboss.org/browse/ISPN-5886

I've also "aligned" the properties in parent/pom.xml and server/integration/versions/pom.xml so that it's easier to keep them in sync


danberindei:
Integrated, thanks Tristan!

PR #3785, opened 2015-10-27
ISPN-5882 XSite replication - take-offline.after-failures property is…

pruivo:
… ignored

https://issues.jboss.org/browse/ISPN-5882


danberindei:
Integrated, thanks Pedro!

PR #3787, opened 2015-10-28
ISPN-5889 Deal with merge views in Hot Rod

galderz:
https://issues.jboss.org/browse/ISPN-5889

* Hot Rod address cache needs to be updated when a merge happens and new
  nodes return back to the cluster, or other nodes are removed.
* The problem is that a merge is the result of a previous installed view
  where a node is removed, and if it's removed from the address cache,
  its endpoint address is also removed. So, to be able to restore a
  node's endpoint address if and when a merge happens, after removing it
  from the address cache, it needs to be kept around in a backup cache
  in case the node comes back as a result of a merge.
* Obviously, this backup address cache needs to be purged so it has been
  configured with expiration timeout of one day, so a merge can still be
  dealt with within that time.
* The backup address cache is a simple cache to reduce its footprint
  since it's just a concurrent hash map that supports expiration.

### DIFF
--- a/core/src/test/java/org/infinispan/partitionhandling/BasePartitionHandlingTest.java
+++ b/core/src/test/java/org/infinispan/partitionhandling/BasePartitionHandlingTest.java
@@ -76,7 +76,7 @@ public class BasePartitionHandlingTest extends MultipleCacheManagersTest {
    public static class PartitionDescriptor {
       int[] nodes;
 
-      PartitionDescriptor(int... nodes) {
+      public PartitionDescriptor(int... nodes) {
          this.nodes = nodes;
       }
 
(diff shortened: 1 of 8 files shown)


danberindei:
@galderz I think a better approach would be to have the node that's missing from the topology cache re-add itself after the merge, something like this: https://github.com/galderz/infinispan/compare/t_5889...danberindei:dan_pr_galderz_t_5889

galderz:
@danberindei Had a closer look and it looks good. Shall I open a new PR? Or you open a new PR for this branch of yours?

danberindei:
I've opened a new PR, https://github.com/infinispan/infinispan/pull/3798

PR #3790, opened 2015-10-28
ISPN-5896 ClusterStreamManager segment generation should use

wburns:
locatePrimaryOwnerForSegment

https://issues.jboss.org/browse/ISPN-5896

Doing a very quick local test with 4 nodes I saw about a 7% improvement in throughput when doing Stream.count calls.


wburns:
This should be cherry-picakable back to 8.0.x without a fuss as well.

danberindei:
Integrated, thanks Will!

PR #3791, opened 2015-10-29
ISPN-5897 Enhance filter_failed_tests.sh to support compressed logs

danberindei:
https://issues.jboss.org/browse/ISPN-5897

Also update failedTestLogs.sh to require a log file argument and to
use perl more.

### DIFF
--- a/bin/failedTestLogs.sh
+++ b/bin/failedTestLogs.sh
@@ -1,28 +1,30 @@
-#!/bin/sh
+#!/bin/bash
+
+FILE=$1 ; shift
+FAILED_TESTS="$*"
 
 DIR=`dirname $0`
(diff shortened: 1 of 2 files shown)


bin/failedTestLogs.sh
```diff
@@ -1,28 +1,30 @@
 #!/bin/sh
```
tristantarrant: I think things like ${FILE##*.} are not in the basic Bourne shell, so let's make this /bin/bash

danberindei: Done.

tristantarrant:
Pulling

tristantarrant:
Pushed, thanks

PR #3792, opened 2015-10-29
ISPN-5613 Replication timeouts should show the affected entries

danberindei:
https://issues.jboss.org/browse/ISPN-5613

### DIFF
--- a/core/src/main/java/org/infinispan/util/logging/Log.java
+++ b/core/src/main/java/org/infinispan/util/logging/Log.java
@@ -542,8 +542,8 @@ public interface Log extends BasicLogger {
    void unableToLockToInvalidate(Object key, Address address);
 
    @LogMessage(level = ERROR)
-   @Message(value = "Execution error", id = 136)
-   void executionError(@Cause Throwable t);
+   @Message(value = "Error executing command %s, writing keys %s", id = 136)
+   void executionError(String commandType, Collection<?> affectedKeys, @Cause Throwable t);
 
    @LogMessage(level = INFO)
    @Message(value = "Failed invalidating remote cache", id = 137)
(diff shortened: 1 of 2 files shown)


danberindei:
Please cherry-pick to 8.0.x.

core/src/main/java/org/infinispan/util/logging/Log.java
```diff
@@ -542,8 +542,8 @@ void preparedTxAlreadyExists(RecoveryAwareTransaction previous,
    void unableToLockToInvalidate(Object key, Address address);
 
    @LogMessage(level = ERROR)
-   @Message(value = "Execution error", id = 136)
-   void executionError(@Cause Throwable t);
+   @Message(value = "Error executing command %s, writing keys %s", id = 136)
+   void executionError(String commandType, Collection<Object> affectedKeys, @Cause Throwable t);
```
wburns: This seems like a lot of trace possibly to go to ERROR.  Maybe we should only write all the keys if TRACE is enabled?

wburns: Also we could just define it as Collection<?> to be a bit more flexible :)

danberindei: It's not that hard to figure out what keys were affected from other logs, if TRACE is enabled... The problem is here that users don't enable TRACE and still want to know what keys might have become inconsistent because of a replication timeout.

wburns: Fair enough.

wburns: The generic is fine to ignore if you want.

tristantarrant:
Applied to both master and 8.0.x. Thanks

PR #3797, opened 2015-10-31
fix, Functional Map Listeners code example.

kazuhira-r:


galderz:
Thanks again @kazuhira-r :)

PR #3799, opened 2015-11-02
ISPN-5903 Spring split

slaskawi:
Hey!

Please take a look at proposed solution for https://issues.jboss.org/browse/ISPN-5903

This PR is very similar to CDI split implemented a week ago: https://github.com/infinispan/infinispan/pull/3775

Thanks
Sebastian


spring/spring4/spring4-common/README.md
```diff
@@ -0,0 +1,4 @@
+Infinispan Spring support common module
+====================================
+
+This module is not meant to be used individually. It only contains common to Remote and Embedded classes which are attached as sources to those modules.
```
tristantarrant: "It only contains classes common to ..."

slaskawi:
@tristantarrant done

pom.xml
```diff
@@ -56,7 +56,9 @@
       <module>rhq-plugin</module>
       <module>osgi</module>
       <module>spring/spring</module>
-      <module>spring/spring4</module>
```
tristantarrant: Some inconsistency: spring/spring4/pom.xml is never referenced directly by any reactor and therefore it doesn't get "installed" into the repository

slaskawi: Yes, that's intentional. `spring/spring4/pom.xml` is a parent pom for Spring 4 modules which doesn't need to be built at all. It just makes things a little bit easier when importing the project into IDE.

tristantarrant: Unfortunately, if I then create an application which depends on spring4-embedded it won't find the parent pom in the repository

slaskawi: Good point - I haven't thought about it... Fix is right away.

slaskawi:
@tristantarrant done

tristantarrant:
Can you squash the commits please ?

slaskawi:
Yeah sure, any further comments or squashing and are good to go?

tristantarrant:
Looks good

tristantarrant:
Pushed, thanks

PR #3804, opened 2015-11-03
ISPN-5802 Deprecate DataContainer API as no longer pluggable

wburns:
https://issues.jboss.org/browse/ISPN-5802

~~I am not sure if we want to introduce Parser81 for this?~~  There is no harm of the message.


wburns:
Also maybe we want to add this for 8.0.2?

tristantarrant:
Parser versioning is now done inline

tristantarrant:
Pushed only to master. let's leave 8.0.x as is

PR #3805, opened 2015-11-04
ISPN-5917 + ISPN-5918 Transaction table leaks

danberindei:
ISPN-5917 Transactions rolled back early are never removed
https://issues.jboss.org/browse/ISPN-5917
ISPN-5918 Remote transactions can be registered after cache stop
https://issues.jboss.org/browse/ISPN-5918


core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
```diff
@@ -105,6 +105,7 @@
    private ConcurrentMap<GlobalTransaction, RemoteTransaction> remoteTransactions;
    private Lock minTopologyRecalculationLock;
    protected boolean clustered = false;
+   protected boolean running = false;
```
wburns: Seems we should make this volatile

wburns:
Simple suggestion.

danberindei:
Updated.

wburns:
Pulling...

wburns:

PR #3806, opened 2015-11-05
ISPN-5838 Fixed Maven reactor

slaskawi:
As @tristantarrant mentioned here: https://github.com/infinispan/infinispan/pull/3799#discussion_r43868007 we need to make sure that CDI parent pom will be published into the repository.


tristantarrant:
Pushed, thanks

PR #3808, opened 2015-11-05
ISPN-5924 and ISPN-5925 GlobalStateManager and PersistentUUID

tristantarrant:
https://issues.jboss.org/browse/ISPN-5924
https://issues.jboss.org/browse/ISPN-5925

Before I come along with the actual graceful cache shutdown / start, here are two building blocks.

The GlobalStateManager sits in the GCR lifecycle and manages global and per-cache state persistence.

The LocalUUID is the unique node identifier which survives restarts.

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/Util.java
+++ b/commons/src/main/java/org/infinispan/commons/util/Util.java
@@ -699,6 +699,10 @@ public final class Util {
          .append(HEX_VALUES.charAt((b & 0x0F)));
    }
 
+   private static void addSingleHexByte(StringBuilder buf, byte b) {
+      buf.append(HEX_VALUES.charAt(b & 0x0F));
+   }
+
    public static Double constructDouble(Class<?> type, Object o) {
       if (type.equals(Long.class) || type.equals(long.class))
          return Double.valueOf((Long) o);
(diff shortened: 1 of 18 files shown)


danberindei:
@tristantarrant Lots of failures caused by this NPE:

    Caused by: java.lang.NullPointerException
    at org.infinispan.globalstate.impl.GlobalStateManagerImpl.start(GlobalStateManagerImpl.java:71)

core/src/main/java/org/infinispan/globalstate/impl/GlobalStateManagerImpl.java
```diff
@@ -0,0 +1,132 @@
...
+      writeState(state);
+      log.globalStateWrite(state.getProperty("version"), state.getProperty("timestamp"));
+   }
+
+   @Override
+   public void writeState(ScopedPersistentState state) {
+      File stateFile = getStateFile(state.getScope());
+      try (PrintWriter w = new PrintWriter(stateFile)) {
+         state.forEach((key, value) -> {
+            w.printf("%s=%s%n", key, value);
```
danberindei: Does this mean a state value must not contain '\n', and a key must not contain '='? Perhaps the `ScopedPersistentState` javadoc should explicitly say what the allowed characters are?

core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
```diff
@@ -641,6 +652,21 @@ private int getGlobalTimeout() {
       // TODO Rename setting to something like globalRpcTimeout
       return (int) gcr.getGlobalConfiguration().transport().distributedSyncTimeout();
    }
+
+   @Override
+   public void prepareForShutdown(ScopedPersistentState state) {
+      state.setProperty("uuid", localUUID.toString());
```
danberindei: Shouldn't we write this immediately after startup, in case the node crashes?

core/src/main/java/org/infinispan/topology/LocalUUID.java
```diff
@@ -0,0 +1,76 @@
+package org.infinispan.topology;
+
+import java.util.UUID;
+
+import org.infinispan.remoting.transport.Address;
+
+/**
+ * LocalUUID. A special {@link Address} UUID whose purpose is to remain unchanged across node
```
danberindei: I don't think you need to say the class name in the javadoc :)

TBH `Local` doesn't seem relevant either, perhaps `PersistentUUID` would sound better?

core/src/main/java/org/infinispan/globalstate/impl/GlobalStateManagerImpl.java
tristantarrant: I'd rather encode them.

core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
tristantarrant: Yes, that's what I have in my second set of patches, but it makes more sense to do it here.

core/src/main/java/org/infinispan/topology/LocalUUID.java
tristantarrant: PersistentUUID is much better.

tristantarrant:
Done @danberindei 
* I have renamed LocalUUID to PersistentUUID
* I have added unicode escape/unescape to the state persister
* State is persisted immediately after the LTM has chosen a PersistentUUID

core/src/main/java/org/infinispan/globalstate/impl/GlobalStateManagerImpl.java
```diff
@@ -0,0 +1,136 @@
...
+      writeScopedState(state);
+      log.globalStateWrite(state.getProperty("version"), state.getProperty("timestamp"));
+   }
+
+   @Override
+   public void writeScopedState(ScopedPersistentState state) {
+      File stateFile = getStateFile(state.getScope());
+      try (PrintWriter w = new PrintWriter(stateFile)) {
+         state.forEach((key, value) -> {
+            w.printf("%s=%s%n", key, Util.unicodeEscapeString(value));
```
danberindei: I think the key should be escaped as well.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddressCache.java
```diff
@@ -34,11 +37,22 @@
       });
    }
 
+   public static void putAddressPersistentUUID(Object address, PersistentUUID localUUID) {
+      final Address jgAddress = ((JGroupsAddress) address).address;
+      persistentUUIDCache.put(jgAddress, localUUID);
+   }
+
+   public static PersistentUUID getLocalUUID(Object address) {
```
danberindei: s/LocalUUID/PersistentUUID/

tristantarrant:
@danberindei :
* escaped key
* added a commit to fix some leftover renames from ISPN-5851
* removed traces of localuuid :)

commons/src/main/java/org/infinispan/commons/util/Util.java
```diff
@@ -699,6 +699,10 @@ private static void addHexByte(StringBuilder buf, byte b) {
          .append(HEX_VALUES.charAt((b & 0x0F)));
    }
 
+   private static void addSingleHexByte(StringBuilder buf, byte b) {
+      buf.append(HEX_VALUES.charAt(b));
```
wburns: Guessing we may want to & with 0x0F to be safe?

tristantarrant:
Fixed @wburns

wburns:
Seems fine to me, will pull just after the previous CI run completes.

wburns:
Pulling...

wburns:
Integrated into master, thanks !

PR #3814, opened 2015-11-09
ISPN-5922 Recovery MBean showInDoubtTransactions fails because of mis…

pruivo:
…sing RecoveryManager

https://issues.jboss.org/browse/ISPN-5922

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/cache/RecoveryConfigurationBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/cache/RecoveryConfigurationBuilder.java
@@ -1,12 +1,17 @@
 package org.infinispan.configuration.cache;
 
-import static org.infinispan.configuration.cache.RecoveryConfiguration.ENABLED;
-import static org.infinispan.configuration.cache.RecoveryConfiguration.RECOVERY_INFO_CACHE_NAME;
-
-import org.infinispan.commons.CacheConfigurationException;
 import org.infinispan.commons.configuration.Builder;
 import org.infinispan.commons.configuration.attributes.AttributeSet;
 import org.infinispan.configuration.global.GlobalConfiguration;
+import org.infinispan.transaction.TransactionMode;
+import org.infinispan.transaction.TransactionProtocol;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/infinispan/configuration/cache/Configurations.java
```diff
@@ -66,4 +67,12 @@ public static boolean hasSharedCacheLoaderOrWriter(Configuration configuration)
       }
       return false;
    }
+
+   public static boolean isRecoveryEnabled(Configuration configuration) {
+      TransactionConfiguration txConfig = configuration.transaction();
+      return txConfig.transactionMode() == TransactionMode.TRANSACTIONAL &&  //tx cache
+            !txConfig.useSynchronization() && //xa enlistment
+            txConfig.transactionProtocol() != TransactionProtocol.TOTAL_ORDER && //non TO cache
+            txConfig.recovery().enabled(); //recovery enabled
```
danberindei: We already validate the first 3 conditions in `RecoveryConfigurationBuilder`, do we need another check?

pruivo: well, it cross my mind that we can have a ```Configuration``` instance without validation. You could create a new instance of ```Configuration``` by hand or using ```ConfigurationBuilder.build(false)``` (skips validation)

pruivo: do you think that I'm overthinking?

danberindei: Yeah, I don't think anyone's going to use `Configurations` with a detached configuration...

pruivo: removed!

wburns:
Changes seem fine to me.  @danberindei, did you have any objections?

danberindei:
@wburns just slipped under the radar, integrating...

danberindei:
Done, thanks Pedro!

PR #3823, opened 2015-11-12
ISPN-4956 Task Management

tristantarrant:
https://issues.jboss.org/browse/ISPN-4956
Abstract the execution of tasks into its own component
Port the scripting manager to base on tasks

Some additional features implemented here:
- support for specifying a required role for execution
- support for defining the named parameters to avoid pollution from malicious clients


tristantarrant:
There is a failure with ScriptExecIT  which I will fix. In the meantime, please review the code

tristantarrant:
Fixed. I have also ported the map/reduce tests to use streams: javascript lambdas in streams ftw !

galderz:
Looking...

core/src/main/java/org/infinispan/util/concurrent/CompletableFutures.java
```diff
@@ -36,6 +40,23 @@
...
+         try {
+            result.complete(f.get());
+         } catch (Exception e) {
+            result.cancel(false);
+         }
+      });
+      return result;
+   }
+
+   public static <T> CompletableFuture<List<T>> combine(List<CompletableFuture<T>> futures) {
```
galderz: In the functional programming world (e.g. [Haskell]([here](https://hackage.haskell.org/package/base-4.8.1.0/docs/Control-Monad.html#v:sequence))), this operation is more commonly known as `sequence` which is defined as:

    sequence :: (Traversable t, Monad m) => t (m a) -> m (t a)

^ Above is Haskell code, but substitute `Traversable` for `List`, and `Monad` for `CompletableFuture`, and what that function essentially says is that it transforms a List of CompletableFuture instances into a CompletableFuture that contains a List.

Hence, I'd suggest that you rename it to `sequence`.

core/src/main/java/org/infinispan/util/concurrent/CompletableFutures.java
```diff
@@ -36,6 +40,23 @@
       sink.setFuture(compoundSource);
    }
 
+   public static <T> CompletableFuture<T> connect(NotifyingFuture<T> source) {
```
galderz: More than `connect`, I'd name this `asCompletableFuture` or `toCompletableFuture`. This is often the terminology used in Scala implicit conversions where there are tons of methods like this, with the added benefit that the conversions are implicit rather than you having to spell out the calls.

scripting/src/main/java/org/infinispan/scripting/impl/DistributedRunner.java
```diff
@@ -34,13 +35,12 @@ private DistributedRunner() {
       DefaultExecutorService des = new DefaultExecutorService(masterCacheNode);
       try {
          List<Future<T>> tasks = des.submitEverywhere(new DistributedScript<T>(metadata));
-         List<NotifyingFuture<T>> notifyingTasks = new ArrayList<>(tasks.size());
-         for (Future<T> task : tasks) {
-               notifyingTasks.add((NotifyingFuture<T>) task);
+         List<CompletableFuture<T>> all = new ArrayList<>(tasks.size());
```
galderz: You could do this more succinctly with `tasks.stream().map(CompletableFutures::connect)`. Not sure if it'll work with `::` since you seem to need the cast...

Also, your CompletableFutures.connect, or asCompletableFuture, could take Future instead of NotifyingFuture and then users wouldn't need to do casting. Within the converter method you'd check the real type...

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
```diff
@@ -148,31 +218,56 @@ public Integer getId() {
...
-         for(Entry<MetadataProperties, String> e : object.properties.entrySet()) {
-            output.writeUTF(e.getKey().name());
-            output.writeUTF(e.getValue());
-         }
+         output.writeUTF(object.name);
+         output.writeUTF(object.extension);
+         output.writeUTF(object.mode.name());
+         writeSet(output, object.parameters);
+
+         OptionalObjectInputOutput.writeOptional(output, object.language);
```
galderz: We already have an `OptionalExternalizer`, there should not be no need to call `OptionalObjectInputOutput.writeOptional` unless I'm missing something. All you need to do is call `output.writeObject(object.language)` and the externalizer would be used.

tristantarrant: For some reason I missed that, and a much better approach than my horrible one :)

galderz:
The rest looks good, great work @tristantarrant :)

scripting/src/main/java/org/infinispan/scripting/impl/DistributedRunner.java
tristantarrant: asCompletableFuture needs NotifyingFutures because it uses the attachListener method

tristantarrant:
Applied all of your suggestions @galderz

core/src/main/java/org/infinispan/util/concurrent/CompletableFutures.java
```diff
@@ -36,6 +38,23 @@
...
+         try {
+            result.complete(f.get());
+         } catch (Exception e) {
+            result.cancel(false);
+         }
+      });
+      return result;
+   }
+
+   public static <T> CompletableFuture<List<T>> sequence(List<CompletableFuture<T>> futures) {
```
danberindei: No longer used?

core/src/main/java/org/infinispan/util/concurrent/CompletableFutures.java
danberindei: @galderz Haskell kind of goes over my head, but isn't `sequence` supposed to be a generic function that can handle any `Monad`?

tristantarrant:
@danberindei I can't seem to find your comments here, although I've received the notifications via e-mail. infinispan/scripting/src/main/java/org/infinispan/scripting/impl/DistributedRunner.java uses CompletableFutures.sequence()

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
```diff
@@ -3,57 +3,83 @@
...
-      }
+   public String extension() {
+      return extension;
+   }
 
-      @Override
-      public String toString() {
-         return s;
-      }
+   public Set<String> parameters() {
```
danberindei: I would prefer `List` to `Set` for the parameter names.

danberindei:
@tristantarrant I deleted my comment once I noticed that I had ignored the `scripting` module in IntelliJ :)

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
```diff
@@ -148,31 +217,55 @@ public Integer getId() {
...
-         for(Entry<MetadataProperties, String> e : object.properties.entrySet()) {
-            output.writeUTF(e.getKey().name());
-            output.writeUTF(e.getValue());
-         }
+         output.writeUTF(object.name);
+         output.writeUTF(object.extension);
+         output.writeUTF(object.mode.name());
+         writeSet(output, object.parameters);
+         output.writeObject(object.language);
+         output.writeObject( object.role);
```
danberindei: Extra space.

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
```diff
@@ -148,31 +217,55 @@ public Integer getId() {
...
-         }
+         builder.name(input.readUTF()).extension(input.readUTF()).mode(ExecutionMode.valueOf(input.readUTF()));
+         builder.parameters(readSet(input));
+
+         ((Optional<String>)input.readObject()).ifPresent(s -> builder.language(s));
+         ((Optional<String>)input.readObject()).ifPresent(s -> builder.role(s));
+         ((Optional<String>)input.readObject()).ifPresent(s -> builder.reducer(s));
+         ((Optional<String>)input.readObject()).ifPresent(s -> builder.collator(s));
+         ((Optional<String>)input.readObject()).ifPresent(s -> builder.combiner(s));
          return builder.build();
```
danberindei: Couldn't you use the constructor directly?

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
```diff
@@ -148,31 +217,55 @@ public Integer getId() {
...
       }
 
-   }
+      private void writeSet(ObjectOutput output, Set<String> set) throws IOException {
```
danberindei: What's the advantage over `output.writeObject(set)`?

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadataParser.java
```diff
@@ -0,0 +1,102 @@
...
+
+import org.infinispan.scripting.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+
+public class ScriptMetadataParser {
+   private static final Log log = LogFactory.getLog(ScriptMetadataParser.class, Log.class);
+   private static final String DEFAULT_SCRIPT_EXTENSION = "js";
+   private static final Pattern METADATA_COMMENT = Pattern.compile("^(?:#|//|;;)\\s*(.+)");
+   private static final Pattern METADATA_KV = Pattern
+         .compile("\\s*(\\w+)\\s*=\\s*((?:\")[^\"]*(?:\")|\\[[\\w,\\s]*\\]|[^,=\\s\"]+)\\s*,?");
```
danberindei: The group in `(?:\")` seems superfluous.

danberindei: Also, I'm not sure if it's intentional, but AFAICT there's no way to escape a `"`.

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
tristantarrant: Why ? parameter names need to be unique. Order is unimportant.

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
tristantarrant: None, I'll use writeObject()

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadataParser.java
tristantarrant: You read through the regex ? I'm impressed :)
I guess, since I'm unquoting the value in line 41, I don't need the non-capturing group.

tristantarrant:
@danberindei I have addressed some of your comments.

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
tristantarrant: This makes it more readable, otherwise you'd have a bunch of (..., input.readObject(), ...) and you wouldn't know which one is which :)

scripting/src/main/java/org/infinispan/scripting/impl/ScriptingManagerImpl.java
```diff
@@ -177,109 +175,41 @@ ScriptMetadata getScriptMetadata(String scriptName) {
...
-
    ScriptEngine getEngineForScript(ScriptMetadata metadata) {
-      String language = metadata.property(MetadataProperties.LANGUAGE);
-      if (language != null) {
-         if (scriptEnginesByLanguage.containsKey(language)) {
-            return scriptEnginesByLanguage.get(language);
-         } else {
-            ScriptEngine engine = scriptEngineManager.getEngineByName(language);
+      return metadata.language().map( (language) -> {
+         return scriptEnginesByLanguage.computeIfAbsent(language, (lang) -> {
```
danberindei: Just a matter of preference, but I like the way regular control look better when there's a lot of nesting. You could also move the `noEngineForScript` check at the end, and you'd end up with something like this:

      ScriptEngine engine;
      if (metadata.language().isPresent()) {
         engine = scriptEnginesByLanguage
               .computeIfAbsent(metadata.language().get(), scriptEngineManager::getEngineByName);
      } else {
         engine = scriptEnginesByExtension
               .computeIfAbsent(metadata.extension(), scriptEngineManager::getEngineByExtension);
      }
      if (engine == null) {
         throw log.noEngineForScript(metadata.name());
      } else {
         return engine;
      }

scripting/src/test/java/org/infinispan/scripting/ClusteredScriptingTest.java
```diff
@@ -49,17 +46,22 @@ public void testClusteredScriptExec() throws IOException, InterruptedException,
...
-         addresses.add(result);
-      }
-      assertTrue(addresses.containsAll(manager(0).getMembers()));
+      Cache<String, String> cache = cache(0);
+      loadData(cache, "/macbeth.txt");
+      loadScript(scriptingManager, "/wordCountStream.js");
+      CompletableFuture<Map<String, Long>> resultsFuture = scriptingManager.runScript("wordCountStream.js", new TaskContext().cache(cache(0)));
+      Map<String, Long> results = resultsFuture.get();
+      assertEquals(3209, results.size());
+      assertTrue(results.get("macbeth").equals(Long.valueOf(287)));
```
danberindei: You can use `assertEquals` here as well.

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
danberindei: Not if you save them into local variables...

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadataParser.java
```diff
@@ -0,0 +1,102 @@
...
+
+import org.infinispan.scripting.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+
+public class ScriptMetadataParser {
+   private static final Log log = LogFactory.getLog(ScriptMetadataParser.class, Log.class);
+   private static final String DEFAULT_SCRIPT_EXTENSION = "js";
+   private static final Pattern METADATA_COMMENT = Pattern.compile("^(?:#|//|;;)\\s*(.+)");
+   private static final Pattern METADATA_KV = Pattern
+         .compile("\\s*(\\w+)\\s*=\\s*(\"[^\"]*\"|\'[^\']*\'|\\[[\\w,\\s]*\\]|[^,=\\s\"]+)\\s*,?");
```
danberindei: TBH it's a bit harder to read after the update :)

I thought I already wrote this, but could you add a comment with some examples of the regexp is supposed to match? It would make it a little easier to read. You should probably also add some user-facing documentation about the metadata syntax, if you do maybe you could link to that instead.

scripting/src/main/java/org/infinispan/scripting/impl/ScriptingManagerImpl.java
tristantarrant: Indeed, much more compact

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
tristantarrant: And it does look better :)

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadataParser.java
tristantarrant: Adding the missing documentation

tristantarrant:
@danberindei @galderz I've addressed all of the latest suggestion.

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
danberindei: I was thinking that it would be odd if someone adds a script with parameters `[foo, bar]`, and then the script appears in the UI with parameters `[bar, foo]` - even if the order doesn't really matter.

documentation/src/main/asciidoc/user_guide/chapter-78-Scripting.adoc
```diff
@@ -11,7 +11,14 @@ Alternatively the script engine can be selected using script metadata (see below
 
 ==== Script metadata
 Script metadata is additional information about the script that the user can provide to the server to affect how a script is executed. 
-Script metadata is contained in a specially-formatted comment on the first line of the script.
+It is contained in a specially-formatted comment on the first lines of the script. Properties are specified as +key=value+ pairs, separated by commas.
+You can use several different comment styles: The '//', ';;', '#' depending on the scripting language you use. You can split metadata over multiple lines if necessary, and you can use single (') or double (") quotes to delimit your values.
```
danberindei: It looks too big for a single paragraph, I think you need to add some empty lines. Also, I'd replace "The '//', ';;', '#'" with "`//`, `;;`, or `#`,"

scripting/src/main/java/org/infinispan/scripting/impl/ScriptMetadata.java
tristantarrant: Fixed.

galderz:
Integrating...

galderz:
Integrated

PR #3828, opened 2015-11-12
ISPN-5957 Simple cache fails to start when statistics is enabled

rvansa:
https://issues.jboss.org/browse/ISPN-5957


tristantarrant:
Pushed, thanks

PR #3829, opened 2015-11-12
[8.0] ISPN-5890 NPE after a node joins a cluster and receives new views

rhusar:


pruivo:
integrated! thanks @rhusar !

PR #3830, opened 2015-11-13
ISPN-5914 Immediately close unauthenticated HotRod connections

tristantarrant:
...when uing SASL NO_ANONYMOUS mechs

https://issues.jboss.org/browse/ISPN-5914


galderz:
Integrated, thanks @tristantarrant :)

PR #3836, opened 2015-11-19
Adding class- and method-names to the key so there are no collissions.

danpfe:
When using JCache-annotations the DefaultCacheKeyGenerator exclusively looks at parameter values to form the cache key. Therefore it will be very likely that collissions occur (resulting in difficult to find ClassCastExceptions). The provided patch uses the method- and class names as additionally values to make the cache key more unique.

Might also add that I am aware that by spec this should not be an issue when no cachename is given (as it should generate a cache using the class-name), but when a cache name is given collissions may occur.

### DIFF
--- a/jcache/embedded/src/main/java/org/infinispan/jcache/annotation/DefaultCacheKeyGenerator.java
+++ b/jcache/embedded/src/main/java/org/infinispan/jcache/annotation/DefaultCacheKeyGenerator.java
@@ -12,11 +12,12 @@ import javax.enterprise.context.ApplicationScoped;
 
 /**
  * Default {@link javax.cache.annotation.CacheKeyGenerator} implementation.
- * By default all key parameters of the intercepted method compose the
+ * By default all key parameters including the intercepted method and class names compose the
  * {@link javax.cache.annotation.CacheKey}.
  *
  * @author Kevin Pollet <kevin.pollet@serli.com> (C) 2011 SERLI
  * @author Galder Zamarreño
+ * @author <a href="mailto:daniel@pfeifer.io">Daniel Pfeifer</a>
  */
 @ApplicationScoped
 public class DefaultCacheKeyGenerator implements CacheKeyGenerator {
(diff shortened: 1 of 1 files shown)


danberindei:
@danpfe Could you create an issue in JIRA as well?
https://issues.jboss.org/secure/CreateIssue!default.jspa

danberindei:
@galderz Assigning to you, since you're more familiar with the JCache specification.

galderz:
I've created a JIRA for this: [ISPN-6011](https://issues.jboss.org/browse/ISPN-6011). 

There seems to be some failures in the testsuite:

```
TestSuite: org.infinispan.integrationtests.cdijcache.interceptor (5)
CachePutInterceptorTest.testCachePut 	
CachePutInterceptorTest.testCachePutCacheKeyParam 	
CachePutInterceptorTest.testCachePutWithCacheName 	
CacheRemoveEntryInterceptorTest.testCacheRemoveEntryWithCacheKeyParam 	
CacheRemoveEntryInterceptorTest.testCacheRemoveEntryWithCacheName 	
```

@danpfe ?

danpfe:
@galderz @danberindei Sorry, was quite busy the last couple of days. I saw you guys already created a JIRA-issue. I can check up on the testsuite-failures whether they're caused by my suggested change.

danpfe:
@galderz I looked into the failing tests now and from what I can see they failed before my change, also. However, I did fix one issue with a test, now I am trying to figure out why the JCache Weld EE Container won't fire up. Will get back.

danpfe:
@galderz Sorry, just reread your previous post and realized I am digging around in the "wrong" failing tests =). I will those out instead.

danpfe:
@galderz And me again... Check the attached screenshot, I am not receiving those failures. Interestingly, I noticed that the Weld EE container doesn't always fire up when I run a regular build with maven from the command prompt (also with other tests), could that be an issue? They seem to fire up just fine with @RunWith(Arquillian.class) (i.e. JUnit-tests), but it doesn't seem to fire up when it's a TestNG-test.

screen shot 2015-12-04 at 13 38 58

I'll continue digging.

galderz:
@danpfe We run the tests from command line, which run multiple tests in parallel, so that might be the reason the tests fail. You can try running from command line using `mvn clean install -pl integrationtests/cdi-jcache-it`

slaskawi:
Hey @galderz, @danpfe! Are we going to reach a conclusion here?

danpfe:
@slaskawi Hopefully, will look into it again tonight.

danpfe:
@galderz @slaskawi Alright, finally managed to get around and made it fail locally as well. Turns out it was what I mentioned in the first post. The spec says that it goes by method params only and that is what some tests look for... Of course, I could change the tests, but I do not think that is a good idea. Since I believe it is beneficial for users of the framework to decide on whether params only are enough or if the cache key should be formed using further criteria, it is likely a better idea to contribute the patch in question as a new feature that users may chose to activate rather than have it active "by default".

slaskawi:
@danpfe Could you please rebase it?

slaskawi:
@danpfe Any update?

slaskawi:
Hey @danpfe ! I took a look at the spec and you are correct. You might be interesting in adding another issue to JCache API: https://github.com/jsr107/jsr107spec/issues.

Closing this issue for now.

PR #3840, opened 2015-11-23
ISPN-5980 Updated Scala artifact version to 2.11.7

rigazilla:
Updated Scala artifact version to 2.11.7 in the bom


tristantarrant:
Pushed, thanks

PR #3842, opened 2015-11-24
ISPN-5879 Avoid concurrent cluster switches

galderz:
Matej and I have been testing manually but I didn't have time to create a unit test for it. I'll create a JIRA for this.

* When switching clusters, update topology id of the cache involved. To
  do so, enhance TopologyInfo to deal with multiple topology ids.
* Instead of relying on individual requests to check failover validity,
  add a step in the failover cluster switch logic to check the servers
  to which it is going to failover. If none of those are available
  consider cluster switch not achieved. If any succeed, switch the
  cluster and apply new topology.
* Cluster switches can happen while clients are hammering the server, in
  which case there's a possibility that after a cluster switch old
  topology updates are received. To avoid these being applied, the
  concept of topology age has been added which is increased every time
  there's a cluster switch. When new topologies are received, the
  topology age is verified and it's from a previous cluster, this is ignored.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/TopologyInfo.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/TopologyInfo.java
@@ -39,11 +39,11 @@ public final class TopologyInfo {
    private Collection<SocketAddress> servers = new ArrayList<>();
    private Map<byte[], ConsistentHash> consistentHashes = CollectionFactory.makeMap(ByteArrayEquivalence.INSTANCE, AnyEquivalence.getInstance());
    private Map<byte[], Integer> segmentsByCache = CollectionFactory.makeMap(ByteArrayEquivalence.INSTANCE, AnyEquivalence.getInstance());
-   private volatile AtomicInteger topologyId;
+   private Map<byte[], AtomicInteger> topologyIds = CollectionFactory.makeMap(ByteArrayEquivalence.INSTANCE, AnyEquivalence.getInstance());
    private final ConsistentHashFactory hashFactory = new ConsistentHashFactory();
 
    public TopologyInfo(AtomicInteger topologyId, Collection<SocketAddress> servers, Configuration configuration) {
-      this.topologyId = topologyId;
+      this.topologyIds.put(new byte[0], topologyId);
       this.servers = servers;
       hashFactory.init(configuration);
    }
@@ -65,7 +65,8 @@ public final class TopologyInfo {
       return servers;
    }
 
-   public void updateTopology(Map<SocketAddress, Set<Integer>> servers2Hash, int numKeyOwners, short hashFunctionVersion, int hashSpace, byte[] cacheName) {
+   public void updateTopology(Map<SocketAddress, Set<Integer>> servers2Hash, int numKeyOwners, short hashFunctionVersion, int hashSpace,
+         byte[] cacheName, AtomicInteger topologyId) {
       ConsistentHash hash = hashFactory.newConsistentHash(hashFunctionVersion);
       if (hash == null) {
          log.noHasHFunctionConfigured(hashFunctionVersion);
@@ -73,10 +74,11 @@ public final class TopologyInfo {
          hash.init(servers2Hash, numKeyOwners, hashSpace);
       }
       consistentHashes.put(cacheName, hash);
-
+      topologyIds.put(cacheName, topologyId);
    }
 
-   public void updateTopology(SocketAddress[][] segmentOwners, int numSegments, short hashFunctionVersion, byte[] cacheName) {
+   public void updateTopology(SocketAddress[][] segmentOwners, int numSegments, short hashFunctionVersion,
+         byte[] cacheName, AtomicInteger topologyId) {
       if (hashFunctionVersion > 0) {
          SegmentConsistentHash hash = hashFactory.newConsistentHash(hashFunctionVersion);
          if (hash == null) {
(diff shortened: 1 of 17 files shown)


gustavonalle:
Looking (running tests)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/TopologyInfo.java
```diff
@@ -39,11 +39,11 @@
    private Collection<SocketAddress> servers = new ArrayList<>();
    private Map<byte[], ConsistentHash> consistentHashes = CollectionFactory.makeMap(ByteArrayEquivalence.INSTANCE, AnyEquivalence.getInstance());
    private Map<byte[], Integer> segmentsByCache = CollectionFactory.makeMap(ByteArrayEquivalence.INSTANCE, AnyEquivalence.getInstance());
-   private volatile AtomicInteger topologyId;
+   private Map<byte[], AtomicInteger> topologyIds = CollectionFactory.makeMap(ByteArrayEquivalence.INSTANCE, AnyEquivalence.getInstance());
```
gustavonalle: Does it need to be an ```AtomicInteger```?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/TcpTransportFactory.java
```diff
@@ -389,7 +406,7 @@ private Transport borrowTransportFromPool(SocketAddress server) {
       try {
          TcpTransport tcpTransport = pool.borrowObject(server);
          // If transport was successfully retrieved, reset viewed clusters
-         clustersViewed = 0;
+//            clustersViewed = 0;
```
gustavonalle: delete those 2 lines?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/TopologyInfo.java
galderz: Yes, I think so since there is a happens-before relationship between updating a view by a thread and another thread querying the value to send it to the server indicating its own view id. Also, there's now a check to verify that once a thread has set a view id, no other threads that might have received the same view apply it again, and this comparison is done with topology id.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/TcpTransportFactory.java
galderz: Yes, should be deleted.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/TopologyInfo.java
gustavonalle: Ok, I asked because ```TopologyInfo``` as you can see is not thread-safe, and is always supposed to be called  under a lock in ```TcpTransportFactory```

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/TcpTransportFactory.java
```diff
@@ -174,11 +180,19 @@ private void pingServers() {
             // exceptions from nodes that might not be up any more.
             if (log.isTraceEnabled())
                log.tracef(e, "Ignoring exception pinging configured servers %s to establish a connection",
-                     servers);
+                  servers);
          }
       }
    }
 
+   private void pingServers() throws Exception {
```
gustavonalle: Unused method

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/TcpTransportFactory.java
```diff
@@ -139,7 +145,7 @@ public void start(Codec codec, Configuration configuration, AtomicInteger defaul
       }
 
       if (configuration.pingOnStartup())
-         pingServers();
+         pingServersIgnoreException();
```
gustavonalle: this method calls ```topologyInfo.getServers()``` outside the lock

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/TcpTransportFactory.java
```diff
@@ -451,35 +468,104 @@ public SSLContext getSSLContext() {
...
+   }
 
-      if (clustersViewed >= clusters.size()) {
-         log.debugf("All cluster addresses viewed (number=%d) and none worked: %s", clustersViewed, clusters);
-         return false;
-      }
+   @Override
+   public int getTopologyId(byte[] cacheName) {
+      return topologyInfo.getTopologyId(cacheName);
+   }
```
gustavonalle: this method and ```createTopologyId``` interact with ```topologyInfo``` without a lock, there might be a race condition here

gustavonalle:
@galderz looks good, spotted a couple of minors  plus the ```TopologyInfo``` issue: it is not thread safe and always used under a lock inside ```TcpTransportFactory```, now it seems some operations (related to topologyId) are not using the same pattern

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/TcpTransportFactory.java
galderz: I'll remove

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/TcpTransportFactory.java
galderz: I'll fix

galderz:
Thanks @gustavonalle, I'll sort those out

gustavonalle:
pulling

galderz:
BTW, I've created [ISPN-5879](https://issues.jboss.org/browse/ISPN-5879) to add a unit test.

gustavonalle:
merged, thanks @galderz

PR #3844, opened 2015-11-24
ISPN-5984 Use Message.setFlag(short) to avoid creating a temporary array

danberindei:
https://issues.jboss.org/browse/ISPN-5984


tristantarrant:
Looks good ! Pushed.

PR #3845, opened 2015-11-24
[ISPN-000196]  Failed to recover cluster state after the current node…

treblereel:
… became the coordinator

[BZ-1283465]   Failed to recover cluster state after the current node became the coordinator
[GSS-01511582] ISPN000136 during concurrent start

based on fix https://issues.jboss.org/browse/ISPN-3395

 no upstream required


tristantarrant:
Merged.

PR #3848, opened 2015-11-24
ISPN-825 Staggered remote get

danberindei:
https://issues.jboss.org/browse/ISPN-825

Incomplete:
* The stagger timeout is fixed at 5 millis, it should be configurable
* The total timeout scheduled task should be scheduled only after
  the last request

### DIFF
--- a/core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
+++ b/core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
@@ -56,7 +56,7 @@ public class CommandAwareRpcDispatcher extends RpcDispatcher {
    private static final Log log = LogFactory.getLog(CommandAwareRpcDispatcher.class);
    private static final boolean trace = log.isTraceEnabled();
    private static final boolean FORCE_MCAST = Boolean.getBoolean("infinispan.unsafe.force_multicast");
-   private static final int STAGGER_DELAY_MILLIS = Integer.getInteger("infinispan.stagger.delay", 5);
+   private static final long STAGGER_DELAY_NANOS = TimeUnit.MILLISECONDS.toNanos(Long.getLong("infinispan.stagger.delay", 5));
 
    private final InboundInvocationHandler handler;
    private final ScheduledExecutorService timeoutExecutor;
@@ -103,7 +103,7 @@ public class CommandAwareRpcDispatcher extends RpcDispatcher {
                                                        DeliverOrder deliverOrder) {
       CompletableFuture<RspList<Response>> future;
       try {
-         if (recipients != null && mode == ResponseMode.GET_FIRST && STAGGER_DELAY_MILLIS > 0) {
+         if (recipients != null && mode == ResponseMode.GET_FIRST && STAGGER_DELAY_NANOS > 0) {
             future = new CompletableFuture<>();
             // We populate the RspList ahead of time to avoid additional synchronization afterwards
             RspList<Response> rsps = new RspList<>();
@@ -330,21 +330,23 @@ public class CommandAwareRpcDispatcher extends RpcDispatcher {
             }
          });
          if (!subFuture.isDone()) {
+            long now = System.nanoTime();
+            long delay = deadline - now;
             if (destIndex < dests.size() - 1) {
-               timeoutExecutor.schedule(() -> triggerNextFuture.complete(null),
-                     STAGGER_DELAY_MILLIS, TimeUnit.MILLISECONDS);
-            } else {
-               // this is the last request, so schedule the cancelation at deadline
-               long delay = deadline - System.nanoTime();
-               timeoutExecutor.schedule(() -> triggerNextFuture.complete(null),
-                     delay, TimeUnit.NANOSECONDS);
-            }
+               delay = Math.min(STAGGER_DELAY_NANOS, delay);
+            } // else it is the last request, so schedule the cancelation at deadline
+            timeoutExecutor.schedule(() -> triggerNextFuture.complete(null),
+                  delay, TimeUnit.NANOSECONDS);
          }
       } else {
          triggerNextFuture.complete(null);
       }
       triggerNextFuture.thenAccept(ignored -> {
          if (theFuture.isDone()) {
+            return;
+         }
+         if (System.nanoTime() >= deadline) {
+            theFuture.complete(rsps);
             return;
          }
          try {
--- a/core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
+++ b/core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
@@ -3,7 +3,6 @@ package org.infinispan.remoting.transport.jgroups;
 import org.infinispan.IllegalLifecycleStateException;
 import org.infinispan.commands.FlagAffectedCommand;
 import org.infinispan.commands.ReplicableCommand;
-import org.infinispan.commands.remote.ClusteredGetCommand;
 import org.infinispan.commons.CacheException;
 import org.infinispan.context.Flag;
 import org.infinispan.remoting.inboundhandler.DeliverOrder;
@@ -36,7 +35,6 @@ import org.jgroups.util.Rsp;
 import org.jgroups.util.RspList;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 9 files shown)


tristantarrant:
@danberindei would be interesting to see performance comparisons before/after, including some network/cpu loads

rvansa:
Running comparison benchmark...

Is the configuration (having that in programmatic API + XML) really desired, or is it better to have this fine-tuning just as a system property?

rvansa:
Here are the results (4-8 nodes in dist sync cluster, 50 threads on each node running requests in loop = without compensation for coordinated omission): http://www-rvansa.rhcloud.com/static/test_combined.html
I'd say that there's pretty good improvement, and choice of 5 ms seems right (longer delays result in lower throughput and even longer average response times). The NoStagger was executed one commit before this PR, those staggered ones use this PR with the delay configurabel through system option.

rvansa:
Added configurable timeout, and replication timeout action registered only for last request:
https://github.com/rvansa/infinispan/tree/ISPN-825

rvansa:
@danberindei One more optimization that could be used (generally, not just for staggered gets): you don't have to register the timeout action, which involves some synchronization in the delayed queue, unless you really need to do something after the timeout. Instead, you can just wrap the completable future in another, which records the deadline (now() + timeout) and translates get() to get(deadline - now()). In this wrapping future, you would only need to register an action in the scheduled executor when someone registers follow-up handler. WDYT?

danberindei:
Thanks @rvansa, this is awesome! I finally got motivated to run a JMH test on my machine to get some CPU numbers, and the number of CPU cycles per get operation went down from 267689 to 196853 (> 25%). The CPU wasn't saturated in either case because the test had only 4 threads reading from 4 nodes, but the throughput was still about 8% higher (from 11255 ops/s to 12203 ops/s).

I like the option to control this with a system setting, ideally I'd want something adaptive but it's a good middle ground to allow testing with different values. The optimization to delay the registration of the timeout task also looks good, although I guess it needs some extra checks for `timeout < stagger timeout * number of targets`.

I'm not sure about the extra optimization you're suggesting, we always have to do something when the remote get doesn't finish in time. ATM we could use wait on the `CompletedFuture` with the RPC timeout, but I'm hoping the sequential interceptors PR will get integrated eventually ;)

rvansa:
@danberindei That's true, sequential interceptors will always require some action.
I'll add the extra check to my branch.

rvansa:
Added check.

danberindei:
@rvansa Updated to include your change, with a small additional check for `timeout <= 0`.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -55,6 +56,7 @@
    private static final Log log = LogFactory.getLog(CommandAwareRpcDispatcher.class);
    private static final boolean trace = log.isTraceEnabled();
    private static final boolean FORCE_MCAST = Boolean.getBoolean("infinispan.unsafe.force_multicast");
+   private static final long STAGGER_DELAY_NANOS = TimeUnit.MILLISECONDS.toNanos(Long.getLong("infinispan.stagger.delay", 5));
```
tristantarrant: While the shorthand Type.getType() methods are nice to use, we really need to use SecurityActions to retrieve system properties.
Also, please create a documentation section to hold all the system environment properties that affect Infinispan.
I have found:
infinispan.arrays.debug
infinispan.unsafe.force_multicast
infinispan.compat
infinispan.debugDependencies

danberindei: I also found `org.infinispan.query.indexmanager.LockAcquiringBackend.MAX_QUEUE_SIZE`, used by the query module.

But this seems too much to put in this PR, I've created https://issues.jboss.org/browse/ISPN-6094

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -96,13 +99,27 @@ private boolean isValid(Message req) {
...
+         if (recipients != null && mode == ResponseMode.GET_FIRST && STAGGER_DELAY_NANOS > 0) {
+            future = new CompletableFuture<>();
+            // We populate the RspList ahead of time to avoid additional synchronization afterwards
+            RspList<Response> rsps = new RspList<>();
+            for (Address recipient : recipients) {
+               rsps.put(recipient, new Rsp<>(recipient));
+            }
+            // This isn't really documented, but some of our internal code uses timeout = 0 as no timeout.
+            long nanoTimeout = timeout > 0 ? TimeUnit.MILLISECONDS.toNanos(timeout) : Long.MAX_VALUE;
+            long deadline = System.nanoTime() + nanoTimeout;
```
wburns: We should be using TimeService.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -272,6 +289,78 @@ private SingleResponseFuture processSingleCall(ReplicableCommand command, long t
...
+               }
+            }
+         });
+         if (!subFuture.isDone()) {
+            long now = System.nanoTime();
+            long delay = deadline - now;
+            if (destIndex < dests.size() - 1) {
+               delay = Math.min(STAGGER_DELAY_NANOS, delay);
+            } // else it is the last request, so schedule the cancelation at deadline
+            timeoutExecutor.schedule(() -> triggerNextFuture.complete(null),
```
wburns: Do you think we need to retain the future returned here, in case if the stagger amount is much higher than a regular response?  In that case we could build up these timer tasks.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -272,6 +289,78 @@ private SingleResponseFuture processSingleCall(ReplicableCommand command, long t
...
+                  delay, TimeUnit.NANOSECONDS);
+         }
+      } else {
+         triggerNextFuture.complete(null);
+      }
+      triggerNextFuture.thenAccept(ignored -> {
```
wburns: Thinking we can move this lambda to a method and just invoke it directly when we know we don't need the scheduled future?

core/src/main/java/org/infinispan/remoting/transport/jgroups/SecurityActions.java
```diff
@@ -0,0 +1,63 @@
...
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+/**
+ * Privileged actions for the package
+ *
+ * Do not move. Do not change class and method visibility to avoid being called from other
+ * {@link java.security.CodeSource}s, thus granting privilege escalation to external code.
+ *
+ * @author Scott.Stark@jboss.org
```
wburns: Didn't know you changed your name :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/SecurityActions.java
```diff
@@ -0,0 +1,63 @@
...
+         }
+      };
+
+      boolean getBooleanProperty(String name);
+
+      int getIntProperty(String name, int defaultValue);
+   }
+
+   static boolean getBooleanProperty(String name) {
+      if (System.getSecurityManager() == null)
```
wburns: @tristantarrant Is this the new way of doing this?  Should the previous system property security actions use this instead?

tristantarrant: I commented above. @danberindei should be using a SecurityAction approach instead.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
danberindei: Only if we want to test it ;)

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
danberindei: You mean we're leaking timer tasks?

wburns: If we had a lot of very fast requests they will build up.  They would be cleaned up eventually, but if we can't catch up we would run out of memory.

tristantarrant:
Dan, you also need to modify the documentation, since https://github.com/infinispan/infinispan/blob/master/documentation/src/main/asciidoc/user_guide/chapter-32-Clustering_modes.adoc refers to ISPN-825

core/src/main/java/org/infinispan/remoting/transport/jgroups/SecurityActions.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @author Dan Berindei
+ * @since 8.2
+ */
+final class SecurityActions {
+   private static <T> T doPrivileged(PrivilegedAction<T> action) {
+      if (System.getSecurityManager() != null) {
+         return AccessController.doPrivileged(action);
+      } else {
+         return Security.doPrivileged(action);
```
wburns: On second thought I am not sure if we need this for simple system property retrieval...  can you confirm for sure @tristantarrant ?  It seems that system property will never check the authorized variable in Security when it is performed.

tristantarrant:
Pushed. We can think of security improvements later, but it doesn't matter much here because the properties are evaluated only once.

tristantarrant:
Thanks a lot @danberindei  and @rvansa

PR #3853, opened 2015-11-25
ISPN-5940 Added CDI Remote module to Wildfly modules

VaclavDedik:
https://issues.jboss.org/browse/ISPN-5940

### DIFF
--- a/as-modules/embedded/build.xml
+++ b/as-modules/embedded/build.xml
@@ -63,7 +63,7 @@
             <maven-resource group="org.infinispan" artifact="infinispan-cachestore-jpa" />
         </module-def>
 
-        <module-def name="org.infinispan.cdi" slot="${infinispan.slot}">
+        <module-def name="org.infinispan.cdi.embedded" slot="${infinispan.slot}">
             <maven-resource group="org.infinispan" artifact="infinispan-cdi-embedded" />
         </module-def>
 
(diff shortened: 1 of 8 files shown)


as-modules/embedded/build.xml
```diff
@@ -63,7 +63,7 @@
             <maven-resource group="org.infinispan" artifact="infinispan-cachestore-jpa" />
         </module-def>
 
-        <module-def name="org.infinispan.cdi" slot="${infinispan.slot}">
+        <module-def name="org.infinispan.cdi.embedded" slot="${infinispan.slot}">
```
slaskawi: We would need to mention the rename here: https://github.com/infinispan/infinispan/blob/master/documentation/src/main/asciidoc/upgrading/upgrading.adoc

slaskawi: @tristantarrant Can we do this kind of change in this release loop?

tristantarrant: I think at the very least we should have an org.infinispan.cdi module that acts as an alias for embedded

VaclavDedik: @tristantarrant I have added a commit that creates an alias named org.infinispan.cdi targetting org.infinispan.cdi.embedded.

tristantarrant:
Pushed, thanks

PR #3856, opened 2015-11-27
ISPN-5889 Re-add node to topology cache with overlapping partitions

danberindei:
https://issues.jboss.org/browse/ISPN-5889

Use a distributed task to check if any other member removed this node's
address, instead of relying on a CacheEntryInvalidatedEvent.

### DIFF
--- a/server/hotrod/src/main/scala/org/infinispan/server/hotrod/HotRodServer.scala
+++ b/server/hotrod/src/main/scala/org/infinispan/server/hotrod/HotRodServer.scala
@@ -1,13 +1,14 @@
 package org.infinispan.server.hotrod
 
 import logging.Log
+import org.infinispan
 import org.infinispan.AdvancedCache
 import org.infinispan.commons.marshall.Marshaller
+import org.infinispan.distexec._
 import org.infinispan.filter.{ParamKeyValueFilterConverterFactory, KeyValueFilterConverterFactory}
 import org.infinispan.notifications.Listener
-import org.infinispan.notifications.cachelistener.annotation.CacheEntryInvalidated
-import org.infinispan.notifications.cachelistener.annotation.CacheEntryRemoved
-import org.infinispan.notifications.cachelistener.event.CacheEntryEvent
+import org.infinispan.notifications.cachelistener.annotation.TopologyChanged
+import org.infinispan.notifications.cachelistener.event.TopologyChangedEvent
 import org.infinispan.notifications.cachelistener.filter.{CacheEventFilterConverterFactory, CacheEventConverterFactory, CacheEventFilterFactory}
 import org.infinispan.server.hotrod.iteration.{DefaultIterationManager, IterationManager}
 import org.infinispan.manager.EmbeddedCacheManager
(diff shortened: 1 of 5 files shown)


galderz:
Looking...

server/hotrod/src/main/scala/org/infinispan/server/hotrod/HotRodServer.scala
```diff
@@ -302,26 +303,44 @@ class HotRodServer extends AbstractProtocolServer("HotRod") with Log {
...
-            // Our address just got invalidated out of the cache, most likely because of a merge
-            // Put it back asynchronously, so we don't block the topology update
-            log.debugf("Re-adding %s to the topology cache", clusterAddress)
-            addressCache.put(clusterAddress, address)
+         val distributedExecutor = new DefaultExecutorService(addressCache)
+         var success = false
+         while (!success) {
+            try {
+               val futures = distributedExecutor.submitEverywhere(new CheckAddressTask(clusterAddress, address))
+               val everybodyHasIt = futures.forall(_.get())
```
galderz: Endless future.get calls here? :worried:

server/hotrod/src/main/scala/org/infinispan/server/hotrod/HotRodServer.scala
```diff
@@ -302,26 +303,44 @@ class HotRodServer extends AbstractProtocolServer("HotRod") with Log {
...
+      def topologyChanged(event: TopologyChangedEvent[Address, ServerAddress]): Unit = {
          if (event.isPre)
             return
 
-         if (event.getKey.equals(clusterAddress)) {
-            // Our address just got invalidated out of the cache, most likely because of a merge
-            // Put it back asynchronously, so we don't block the topology update
-            log.debugf("Re-adding %s to the topology cache", clusterAddress)
-            addressCache.put(clusterAddress, address)
+         val distributedExecutor = new DefaultExecutorService(addressCache)
```
galderz: Do you need to construct it in every callback?

galderz:
The rest looks good

server/hotrod/src/main/scala/org/infinispan/server/hotrod/HotRodServer.scala
danberindei: If you look at the `DistributedTaskPart.innerGet()` implementation, it's not actually waiting forever, because the task has a default timeout equal to `sync.replTimeout`. I could use the same timeout here, but I'm not sure it would help.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/HotRodServer.scala
danberindei: I was thinking the callbacks are going to be rare, but I can definitely move it to a field.

danberindei:
Updated: I moved the distributed executor to a field and added a comment about the future timeout.

anistor:
Looks good. Pulling ...

anistor:
Integrated. Thanks!

PR #3860, opened 2015-11-30
ISPN-5989 Use DC.keyEquivalence().hashCode() for hashing into segments

rvansa:


server/rest/src/test/scala/org/infinispan/rest/IntegrationTest.scala
```diff
@@ -689,6 +689,8 @@ class IntegrationTest extends RestServerTestBase {
          eTagFirst.equals(eTagSecond))
    }
 
+   // This test is *DISABLED* because of ISPN-5992
+   @Test(enabled = false)
```
danberindei: We usually put the JIRA reference in the `description` attribute of the `@Test` annotation, no *emphasis* required :)

danberindei: Speaking of which, is this test already broken in master, or does this commit break it?

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
```diff
@@ -98,9 +103,11 @@ public void start() throws Exception {
          log.tracef("Starting StateTransferManager of cache %s on node %s", cacheName, rpcManager.getAddress());
       }
 
+      Hash hashFunction = configuration.clustering().hash().hash();
+      componentRegistry.registerComponent(hashFunction, Hash.class);
```
danberindei: I think you need a `HashFunctionFactory`, similar to `TransportFactory`.

core/src/test/java/org/infinispan/atomic/AtomicMapAPITest.java
```diff
@@ -16,7 +16,8 @@
 public class AtomicMapAPITest extends BaseAtomicHashMapAPITest {
 
    @SuppressWarnings("UnusedDeclaration")
-   @Test(expectedExceptions = {IllegalArgumentException.class})
+   // This test is *DISABLED* due to ISPN-5988
+   @Test(expectedExceptions = {IllegalArgumentException.class}, enabled = false)
```
danberindei: Is this related to this commit, or was the test already failing randomly?

core/src/test/java/org/infinispan/api/flags/FlagsEnabledTest.java
```diff
@@ -127,9 +128,19 @@ public void testReplicateSkipCacheLoad(Method m) {
       final String v = v(m, 1);
       final String k = k(m, 1);
       cache1.withFlags(Flag.SKIP_CACHE_LOAD).put(k, v);
-      // The write-skew check tries to load it from persistence.
-      assertLoadsAndReset(cache1, isTxCache() ? 1 : 0, cache2, 0);
 
+      Address primaryLocation = cache1.getDistributionManager().getPrimaryLocation(k);
```
danberindei: In general I prefer using MagicKeys, so that the primary owner is always the same.

danberindei:
@rvansa I suppose the motivation for this was to use cached `String` hashcodes, so I would have liked to see some benchmark results in the description :)

server/rest/src/test/scala/org/infinispan/rest/IntegrationTest.scala
rvansa: The test is just fine, but the 'production' logic is broken, and this commit reveals the problem (the test crashes as it should crash).

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
rvansa: But that would require a configuration change (add the attribute) and a bit unclear semantic which has higher priorty; `hash="foo.Bar"` or `hash-factory="foo.BarFactory"`?
Is it really necessary?

core/src/test/java/org/infinispan/atomic/AtomicMapAPITest.java
rvansa: This test works only when originator is one of the owners; this commit changes the nodes where the entry is hashed to and when originator becomes non-owner, the check for locking coarseness does not work.

core/src/test/java/org/infinispan/api/flags/FlagsEnabledTest.java
rvansa: Here, I would use MagicKey, too. However, I am alerted by finding out several places where the 'production' code does not work and test do not show that as the entries are usually hashed in a 'lucky' way. This can be also source for some of the random test failures, if the cache start order is not 100% deterministic.
So pseudo-random hashing should be added to the list of ideas for improving the testsuite.

rvansa:
@danberindei Sure, I'll try to give you some numbers along with the update :)

rvansa:
@danberindei So I had somewhat higher expectations, but at least something: I've ran 4-node test with 50 threads on each node executing put()s or get()s in a loop (no pauses, no throughput deadlines), with 10000 entries being overwritten, on TCP stack (UDP in hyperion behaves in a weird way now). Note that the puts and gets test were ran independently.

Before:
95,194 gets/s, 2.10 ms MRT
263,609 gets/s, 758 us MRT

After:
99,922 puts/s, 2.01 ms MRT
269,531 gets/s, 740 us MRT

CPU usage seems too spiky to conclude on some change; for gets it was around 40%, for puts between 45 - 50% in both cases.

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
danberindei: I meant a `HashFunctionFactory` as a component factory that does the wiring for the `Hash` instance configured by the user. There's no way to configure a `TransportFactory` either, it's just there so that the `Transport` instance it returns is properly wired with its dependencies.

core/src/test/java/org/infinispan/api/flags/FlagsEnabledTest.java
danberindei: `SyncConsistentHashFactory` should become the default CH factory soon, and it will produce random mappings. 

I don't like tests that implicitly rely on a specific location for their keys either - especially if it means the test is meant to cover multiple cases, but only covers one, like with {{AtomicMapAPITest}}. 

But when the test explicitly requires knowing the specific location of its keys, as is the case here, I find it easier to keep the mapping constant and generate random keys, than to keep the key constant and decide what to do based on the random mapping. Or better yet, use a `ControlledConsistentHashFactory` so there's no randomness involved.

danberindei:
@rvansa TBH I was expecting even less of a change, because the `String.hashCode()` caching only works on the originator, not on the remote nodes. 

I guessing some of the advantage would be in skipping the `String.getBytes()` allocation... would you mind running the test again with this version of MurmurHash3: https://github.com/yonik/java_util/blob/master/src/util/hash/MurmurHash3.java ?

rvansa:
@danberindei It works on remote nodes, since the owners are looked up multiple times during one call. But I think that the UTF-8 conversion is the more expensive part, and this does not happen at all now.

tristantarrant:
Not all implementations of MH3 are the same (at least ours isn't). If we change it we need to ensure that it produces the same hashes as the old one, or the HotRod clients are going to hash "wrong".

danberindei:
@tristantarrant The changes @rvansa proposed would only affect non-`byte[]` keys, and it would not affect compatibility mode because it's already broken. The MH3 implementation I linked to be different even for `byte[]`, but I'm not proposing a switch: I just wanted to get an idea of what the effects of using our MH3 on a `String`'s chars directly would be, instead of calling `getBytes()`.

rvansa:
I don't think that particular MH3 implementation will do much difference performance-wise once it looks on all bytes. Nevertheless, why is running it on chars better than on (cached) hashCode? Just the more uniform distribution?

danberindei:
@rvansa I don't think it would be faster, but it might be the same. And would we want to change the `Hash` implementation if doesn't improve performance at all?

rvansa:
@danberindei I got lost here... this PR replaces `Hash` implementation but does not really replace MH3 implementation - that's what you wanted to try out. So, what would be the advantage of https://github.com/yonik/java_util/blob/master/src/util/hash/MurmurHash3.java when you don't expect it to be faster?

danberindei:
@rvansa The reason is that I'm always skeptical about SPI changes. Even though in this case the change is just to use an existing SPI (`Equivalence`) in a way that we haven't used it before :)

But I'm conservative by nature, and if changing just the `MurmurHash3` implementation for Strings gives us the same performance, I would prefer to do that.

rvansa:
@danberindei Updated (MagicKey use, test comments)

danberindei:
@rvansa You have a failure in `HotRodMergeTest.testNewTopologySentAfterOverlappingMerge`
```
java.lang.AssertionError: expected:<12> but was:<10>
    at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:245)
    at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:252)
    at org.infinispan.server.hotrod.test.HotRodTestingUtil$.assertHashTopology20Received(HotRodTestingUtil.scala:198)
```
Looks like you need to implement `equals(Object)` (and `hashCode()`) in all the `Equivalence` classes, because `PreferAvailabilityStatus` checks the equality of stable topology objects that have been received from different remote nodes, and are unmarshalled with different `Equivalence` instances. When they don't match, it creates a couple unneeded topology updates.

However, that doesn't cover custom `Equivalence` implementations (assuming there is such a thing), so perhaps it's better to change the `MurmurHash3.equals(Object)` implementation to skip the `equivalence.equals(that.equivalence)` check.

You also have failures in `CompatRoutingTest.testGetWithCompatibility` and `ClientAsymmetricClusterTest.testAsymmetricCluster`, but I haven't looked into those yet.

danberindei:
@rvansa Remember I said that compatibility mode doesn't matter because clients didn't target the primary owner anyway? Looks like Galder fixed that with [ISPN-5981](https://issues.jboss.org/browse/ISPN-5981) in the meantime, and `CompatRoutingTest` is testing exactly that.

There doesn't seem to be any "default" equivalence used in compatibility mode, but the good thing is that both `AnyEquivalence` and `AnyServerEquivalence` use `object.hashCode()` for everything except byte arrays. So you could fix the common case (and the test) without adding support for custom `Equivalence` on the client, just by using `AnyServerEquivalence`.

danberindei:
Now that https://github.com/infinispan/infinispan/pull/3779 is finally in, I'll try to whip up an alternative implementation that changes only `HashFunctionPartitioner`. Since the partitioner is not included in the CH on the coordinator, it should automatically fix `HotRodMergeTest`. The client will still need some changes to fix `CompatRoutingTest`.

rvansa:
Rebased. I don't have those tests failing locally, so I'll wait for CI to report failures.

rvansa:
Fixed issue with compatibility. Incorporating https://github.com/infinispan/infinispan/pull/3933 to run the CI tests with this fix as well.

galderz:
Needs rebasing

tristantarrant:
@rvansa , ISPN-6066 has already been integrated so it can be removed from the PR.

rvansa:
@danberindei Rebased - thanks.

danberindei:
@rvansa Needs another rebase because `ClientAsymmetricClusterTest` got deleted.

rvansa:
Done.

slaskawi:
Hey @rvansa - could you please rebase it?

rvansa:
@slaskawi Done, thanks.

danberindei:
Let's keep this on ice for a while, I want to try changing the key partitioner instead of the hash function.

rvansa:
@danberindei Hm, OK. Just let me know then, I am already a bit tired of rebasing over and over.

danberindei:
@rvansa That's why I changed the label, I didn't want to ask you to rebase again :)

slaskawi:
@tristantarrant @rvansa @danberindei What are we planning to do with this one... it's very old and conflicted in many files. My vote - close it and reopen if needed.

tristantarrant:
Closing.

PR #3866, opened 2015-11-30
ISPN-5948 ClusterListenerDistTxTest.testSimpleExpirationFilterNotOwner

wburns:
and ClusterListenerDistTxTest.testSimpleExpirationConverterNotOwner fail
with assertion error

* Moved notification interceptor to be below locking
* Makes sure tx notifications are properly ordered

https://issues.jboss.org/browse/ISPN-5948

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
@@ -301,6 +301,9 @@ public class TransactionConfigurationBuilder extends AbstractConfigurationChildB
             throw log.invalidLockingModeForTotalOrder(lockingMode());
          }
       }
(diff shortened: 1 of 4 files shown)


danberindei:
@wburns Should we add a warning saying listeners may not be invoked if `configuration.transaction().notifications()` is disabled?

wburns:
Sure I can add that.

wburns:
@danberindei Added an INFO message now just in case.

danberindei:
@wburns I was thinking of a WARN message that would be triggered only when cluster listeners are actually used, but the INFO message should be good enough.

galderz:
Needs rebasing?

wburns:
I rebased yesterday, guess something else came in.  Probably Dan's monster PR :)

I will fix it in just a bit.

wburns:
Rebased

danberindei:
Integrated, thanks Will!

PR #3872, opened 2015-12-02
[CI Debug] AtomicObjectFactoryTest.distributedCacheTest

tristantarrant:
This is a PR to verify why the above test fails on CI with some cleanups and additional debug. Do not integrate.


tristantarrant:
It looks as if CleanupAfterMethod fixes the test.

PR #3873, opened 2015-12-02
ISPN-5623 Retried prepare commands do not wait for backup locks

pruivo:
https://issues.jboss.org/browse/ISPN-5623

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/locking/OptimisticLockingInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/locking/OptimisticLockingInterceptor.java
@@ -61,6 +61,11 @@ public class OptimisticLockingInterceptor extends AbstractTxLockingInterceptor {
       final Collection<Object> keysToLock = command.getKeysToLock();
       ((TxInvocationContext<?>) ctx).addAllAffectedKeys(command.getAffectedKeys());
       if (!keysToLock.isEmpty()) {
+         if (command.isRetriedCommand() && ctx.isOriginLocal()) {
+            //clear backup locks for local and retried commands only. The remote commands clears the backup locks in PendingTxAction.
+            ctx.getCacheTransaction().cleanupBackupLocks();
+            keysToLock.removeAll(ctx.getLockedKeys()); //already locked!
+         }
          Collection<Object> lockedKeys = lockAllOrRegisterBackupLock(ctx, keysToLock,
                                                                      cacheConfiguration.locking().lockAcquisitionTimeout());
          if (!lockedKeys.isEmpty()) {
--- a/core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
+++ b/core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
@@ -41,6 +41,7 @@ public class DummyTransaction implements Transaction {
 
    private static final Log log = LogFactory.getLog(DummyTransaction.class);
    private static boolean trace = log.isTraceEnabled();
+   public static final String FORCE_ROLLBACK_MESSAGE = "Force rollback invoked. (debug mode)";
    private final Xid xid;
    private volatile int status = Status.STATUS_UNKNOWN;
    private final List<Synchronization> syncs;
@@ -102,7 +103,6 @@ public class DummyTransaction implements Transaction {
       checkDone("Cannot commit transaction.");
       runPrepare();
       runCommit(false);
-      throwRollbackExceptionIfAny();
    }
 
    /**
@@ -128,6 +128,11 @@ public class DummyTransaction implements Transaction {
          SystemException systemException = new SystemException("Unable to rollback transaction");
          systemException.initCause(e);
          throw systemException;
+      } catch (RollbackException e) {
+         //ignored
+         if (trace) {
+            log.trace("RollbackException thrown while rolling back", e);
+         }
       }
    }
 
@@ -282,12 +287,20 @@ public class DummyTransaction implements Transaction {
       return true;
    }
 
-   public void runCommit(boolean forceRollback) throws HeuristicMixedException, HeuristicRollbackException {
+   /**
+    * Runs the second phase of two-phase-commit protocol.
+    *
+    * If {@code forceRollback} is {@code true}, then a {@link RollbackException} is thrown with the message {@link #FORCE_ROLLBACK_MESSAGE}.
+    *
+    *
+    * @param forceRollback force the transaction to rollback.
+    */
+   public void runCommit(boolean forceRollback) throws HeuristicMixedException, HeuristicRollbackException, RollbackException {
       if (trace) {
          log.tracef("runCommit(forceRollback=%b) invoked in transaction with Xid=%s", forceRollback, xid);
       }
       if (forceRollback) {
-         markRollbackOnly(new RollbackException("Force rollback invoked. (debug mode)"));
+         markRollbackOnly(new RollbackException(FORCE_ROLLBACK_MESSAGE));
       }
 
       int notifyAfterStatus = 0;
@@ -304,6 +317,7 @@ public class DummyTransaction implements Transaction {
          notifyAfterCompletion(notifyAfterStatus);
          DummyBaseTransactionManager.setTransaction(null);
       }
+      throwRollbackExceptionIfAny(forceRollback);
    }
 
    @Override
@@ -339,8 +353,12 @@ public class DummyTransaction implements Transaction {
       return this == obj;
    }
 
-   public final void throwRollbackExceptionIfAny() throws RollbackException {
+   private void throwRollbackExceptionIfAny(boolean forceRollback) throws RollbackException {
       if (firstRollbackException != null) {
+         if (forceRollback && FORCE_ROLLBACK_MESSAGE.equals(firstRollbackException.getMessage())) {
+            //force rollback set. don't throw it.
+            return;
+         }
          throw firstRollbackException;
       }
    }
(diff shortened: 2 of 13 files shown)


core/src/main/java/org/infinispan/interceptors/locking/OptimisticLockingInterceptor.java
```diff
@@ -60,6 +60,11 @@ public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand comman
       final Collection<Object> keysToLock = command.getKeysToLock();
       ((TxInvocationContext<?>) ctx).addAllAffectedKeys(command.getAffectedKeys());
       if (!keysToLock.isEmpty()) {
+         if (command.isRetriedCommand() && ctx.isOriginLocal()) {
```
danberindei: Shouldn't we exclude `onePhaseCommit` prepares as well? At least with pessimistic locking, the prepare command isn't going to acquire any locks.

pruivo: it is not needed. This is the optimistic locking so the prepare always acquire locks.

danberindei: Oops, this is the optimistic interceptor :)

core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
```diff
@@ -102,7 +103,7 @@ public void commit() throws RollbackException, HeuristicMixedException, Heuristi
       checkDone("Cannot commit transaction.");
       runPrepare();
       runCommit(false);
-      throwRollbackExceptionIfAny();
+      //throwRollbackExceptionIfAny();
```
danberindei: Delete?

core/src/main/java/org/infinispan/transaction/xa/CacheTransaction.java
```diff
@@ -144,4 +149,20 @@
...
+
+   /**
+    * It returns a {@link CompletableFuture} that completes when the lock for the {@code key} is released.
+    *
+    * If the {@code key} is not locked by this transaction, it returns {@code null}.
+    *
+    * @param key the key.
+    * @return the {@link CompletableFuture} or {@link null} if the key is not locked by this transaction.
+    */
+   CompletableFuture<Void> getReleaseFutureForKey(Object key);
```
danberindei: Isn't this over-complicating things? `DefaultPendingLockManager` already knows that this transaction has the backup lock when calling this method, and we release all the backup locks at once, so we could use a single `CompletableFuture` for all the backup locks.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -33,7 +36,7 @@
 
 /**
  * The default implementation for {@link PendingLockManager}.
- * <p>
+ * <p/>
```
danberindei: `<p>` should enclose some content, it's not a separator :)

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -77,14 +83,14 @@ public PendingLockPromise checkPendingTransactionsForKey(TxInvocationContext<?>
          return createAndStore(getTransactionWithLockedKey(txTopologyId, key, globalTransaction),
                                globalTransaction, time, unit);
       }
-      return PendingLockPromise.NO_OP;
+      return createAndStoreDefault(globalTransaction, time, unit);
```
danberindei: I don't quite understand why you need to put these promises in the map, could you add a comment explaining that?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -288,17 +296,77 @@ private void forEachTransaction(Consumer<CacheTransaction> consumer) {
...
+         this.key = key;
+         this.keyReleased = Objects.requireNonNull(keyReleased);
+      }
+   }
+
+   private interface DefaultPendingLockPromise extends PendingLockPromise, Callable<Void> {
+
+      void await() throws InterruptedException;
+
+      PendingTransaction getTimedOutTransaction();
```
danberindei: Maybe you could move these methods to `PendingLockPromise`?

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -343,29 +411,30 @@ public long getRemainingTimeout() {
...
       }
 
       @Override
-      public Void call() throws Exception {
-         isReady();
-         return null;
-      }
-
-      private PendingTransaction getTimedOutTransaction() {
+      public PendingTransaction getTimedOutTransaction() {
```
danberindei: I think the name is misleading, because the transaction you're returning hasn't timed out... why not call it just `pendingTransaction`?

core/src/test/java/org/infinispan/jmx/MvccLockManagerMBeanTest.java
```diff
@@ -79,7 +83,12 @@ public void testNumberOfLocksAvailable() throws Exception {
 
       assertAttributeValue("NumberOfLocksHeld", 1);
       assertAttributeValue("NumberOfLocksAvailable", initialAvailable - 1);
-      tm.getTransaction().runCommit(true);
+      try {
+         tm.getTransaction().runCommit(true);
+         fail("RollbackException expected!");
```
danberindei: Why do we expect this transactions to fail?

core/src/test/java/org/infinispan/partitionhandling/OptimisticTxPartitionAndMergeDuringRollbackTest.java
```diff
@@ -61,7 +66,12 @@ public void testSplitBeforeRollback() throws Exception {
       SplitMode.BOTH_DEGRADED.split(this);
 
       transactionManager.resume(transaction);
-      transaction.runCommit(true);
+      try {
+         transaction.runCommit(true);
+         fail("RollbackException expected!");
```
danberindei: I would have liked a more specific error message, or a comment on why the commit fails.

core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
pruivo: yes it should been deleted :)

core/src/main/java/org/infinispan/transaction/xa/CacheTransaction.java
pruivo: right... you know me :) I like to complicate everything

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
pruivo: this is only implementation specific method. I don't wanna make them public outside this class.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
pruivo: of course.

core/src/test/java/org/infinispan/jmx/MvccLockManagerMBeanTest.java
pruivo: because the forceRollback is true.

danberindei: Hmmm, then I guess I have a problem with `runCommit(true)` always throwing an exception :)

It seems to be only used in tests, so couldn't the tests that use it throw the exception themselves?

pruivo: I convinced myself with following argument: without the exception, you (== or anyone who wrote some test) are not sure if the transaction rolledback because it was forced or if you have some error in the test. 
If you don't wanna write the ```try-catch``` I can do one of the following:
* create an helper method that checks if everything runs as expected
* don't throw the exception if ```forceRollback=true``` and the transaction rolledback because of it.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
pruivo: I was trying to solve a deadlock by avoiding the remote command to double check with the DefaultPendingLockManager. However, I've managed to solve it in other way and I think this code can be removed now.

core/src/test/java/org/infinispan/jmx/MvccLockManagerMBeanTest.java
danberindei: I would lean towards the 2nd alternative, I don't like the idea that the normal flow of the method is to throw an exception, or having to check the exception message.

OTOH perhaps most tests that use `runCommit(true)` don't really need it, and they can use `rollback()` instead.

pruivo:
@danberindei updated!

commons/src/main/java/org/infinispan/commons/util/Pair.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Objects;
+
+/**
+ * A pair class.
+ *
+ * @author Pedro Ruivo
+ * @since 8.1
+ */
+public class Pair<E1, E2> {
```
danberindei: We already have `KeyValuePair`

pruivo: removed!

danberindei:
Integrated, thanks Pedro!

PR #3874, opened 2015-12-02
ISPN-5885 NonTotalOrderPerCacheInvocationHandlerImpl should release l…

pruivo:
…ocks

* refactor CompositeAction class

https://issues.jboss.org/browse/ISPN-5885

### DIFF
--- a/core/src/main/java/org/infinispan/remoting/inboundhandler/action/LockAction.java
+++ b/core/src/main/java/org/infinispan/remoting/inboundhandler/action/LockAction.java
@@ -81,6 +81,13 @@ public class LockAction extends BaseLockingAction implements LockListener {
    }
 
    @Override
+   public void cleanup(ActionState state) {
+      final Object lockOwner = getLockOwner(state);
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/infinispan/remoting/inboundhandler/action/LockAction.java
```diff
@@ -81,6 +81,13 @@ public void addListener(ActionListener listener) {
    }
 
    @Override
+   public void cleanup(ActionState state) {
+      final Object lockOwner = getLockOwner(state);
+      List<Object> keysToLock = getAndUpdateFilteredKeys(state);
+      lockManager.unlockAll(keysToLock, lockOwner);
```
danberindei: Doesn't this get called when a remote `PrepareCommand` finishes successfully?

pruivo: it is a bit ugly, but the cleanup is not invoked in transactional caches.

danberindei:
Integrated, thanks Pedro!

PR #3877, opened 2015-12-03
ISPN-5883 Node can apply new topology after sending status response

danberindei:
https://issues.jboss.org/browse/ISPN-5883

Check both the sender and the view id before applying a topology update.

PR #3878, opened 2015-12-03
ISPN-5883 Node can apply new topology after sending status response

danberindei:
https://issues.jboss.org/browse/ISPN-5883

Check both the sender and the view id before applying a topology update.

Backport of https://github.com/infinispan/infinispan/pull/3877


danberindei:
No longer backporting to 8.0.x

PR #3880, opened 2015-12-04
ISPN-6014 DebuggingUnitTestNGListener handles threadTestName incorrectly

mgencur:
https://issues.jboss.org/browse/ISPN-6014

### DIFF
--- a/core/src/test/java/org/infinispan/test/fwk/DebuggingUnitTestNGListener.java
+++ b/core/src/test/java/org/infinispan/test/fwk/DebuggingUnitTestNGListener.java
@@ -3,6 +3,7 @@ package org.infinispan.test.fwk;
 import java.util.Set;
 
 import org.infinispan.manager.CacheContainer;
+import org.infinispan.test.AbstractInfinispanTest;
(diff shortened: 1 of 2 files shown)


danberindei:
@mgencur I'm surprised someone's still using `DebuggingUnitTestTestNGListener` :) I was going to remove it, because leaks are almost always going to happen in failing tests, so running once in a while with `DebuggingUnitTestTestNGListener` won't find any of them.

mgencur:
I think Sebastian is going to use it in his CI. He actually asked me to fix this problem.

slaskawi:
Hey @danberindei! I have noticed this problem during downstream build. I'm all for removing this functionality completely (if that's the simpler thing to do).

danberindei:
@slaskawi Yes, I don't think the `debug-tests` profile is really useful any more, so I'd rather remove it.

slaskawi:
@mgencur @danberindei +1. I've removed it from Jenkins job also.

danberindei:
@mgencur @slaskawi Replaced with https://github.com/infinispan/infinispan/pull/3886

PR #3881, opened 2015-12-04
ISPN-6008 Query DSL groupBy and orderBy should only accept single-valued properties

anistor:
jira: https://issues.jboss.org/browse/ISPN-6008


query/src/main/java/org/infinispan/query/logging/Log.java
```diff
@@ -94,4 +95,28 @@
...
+   IllegalStateException queryParameterNotSet(String paramName);
+
+   @Message(value = "Queries containing grouping and aggregation functions must use projections.",  id = 14021)
+   ParsingException groupingAndAggregationQueriesMustUseProjections();
+
+   @Message(value = "Cannot have aggregate functions in GROUP BY clause", id = 14022)
+   IllegalStateException cannotHaveAggregationsInGroupByClause();
+
+   @Message(value = "The property path '%s' cannot be used in the GROUP BY clause because it is multi-valued", id = 14023)
+   ParsingException multivaluedPropertyCannotBeUsedInGroupBy(String propertyPath);
```
gustavonalle: It'd be nicer to have a message like "not currently supported", since it could be added later

gustavonalle:
Looks good, just a tiny suggestion over error message

query/src/main/java/org/infinispan/query/logging/Log.java
anistor: Fixed.

gustavonalle:
pulling

gustavonalle:
Integrated, thanks @anistor !

PR #3882, opened 2015-12-04
ISPN-6008 Query DSL groupBy and orderBy should only accept single-valued properties

anistor:
jira: https://issues.jboss.org/browse/ISPN-6008


anistor:
Update error message.

gustavonalle:
pushed

PR #3885, opened 2015-12-07
ISPN-6016 Unstable HotRodRemoteCacheIT testPutAsync test

mgencur:
https://issues.jboss.org/browse/ISPN-6016

I was not able to reproduce the issue but, obviously, the test can be affected by other tests leaving entries in the server. I made it more robust.  Also saved 8 seconds in the other test.

### DIFF
--- a/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/hotrod/AbstractRemoteCacheIT.java
+++ b/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/hotrod/AbstractRemoteCacheIT.java
@@ -57,7 +57,7 @@ public abstract class AbstractRemoteCacheIT {
 
     protected RemoteCache remoteCache;
     protected static RemoteCacheManager remoteCacheManager = null;
-    protected final int ASYNC_OPS_ENTRY_LOAD = 100;
+    protected final int ASYNC_OPS_ENTRY_LOAD = 10;
 
     protected abstract List<RemoteInfinispanServer> getServers();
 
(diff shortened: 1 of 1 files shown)


danberindei:
@mgencur Are you sure this is necessary? I've been looking at the test history in CI, and there was only one failure on a PR branch that was failing all the ITs:

http://ci.infinispan.org/project.html?projectId=Infinispan&testNameId=-3265904637604539015&tab=testDetails
http://ci.infinispan.org/project.html?projectId=Infinispan&testNameId=6126279333100444085&tab=testDetails

danberindei:
@mgencur If anything, I'd change the `assertCacheEmpty()` method call `clearServer()` only once and to remove the `remoteCache.size() != 0` check.

Historically, there have been some problem with `size()`, e.g. ISPN-5755. The test used to try to clear each server separately, but I think it's much better to just call `clear()` without any check.

mgencur:
This is a random failure that annoys us within the product. And if I fix it in the product branch only, it will be lost for the next release. 

Regarding your suggestion, I don't understand it:) 
What do you mean by clearServer() only once ?
I can't see any check like the following in the test class: remoteCache.size() != 0

danberindei:
@mgencur `AbstractRemoteCacheIT.assertCacheEmpty()` calls `AbstractRemoteCacheIT.clearServer(int)` twice, once with index `0` and once with index `1`. But `clearServer(int)` ignores the argument, it just calls `clearServer()`, so you have 2 `clearServer()` calls. AFAICT one call should be enough.

There's already a difference between community and the product: in community, `clearServer()` just calls `remoteCache.clear()`, but in the product, it also checks before if the cache is already empty with `while (remoteCache.size() != 0)`. I was suggesting to make the product the same as community by removing the `while` loop.

mgencur:
OK. Thanks, Dan. I fixed the first point. The second point is not worth fixing now because the next product release will include what is in upstream now. Any changes I would make to the product only would be lost.

danberindei:
Integrated, thanks Martin!

PR #3887, opened 2015-12-08
ISPN-6017 Deprecate ReplicationQueue

rhusar:
https://issues.jboss.org/browse/ISPN-6017

I dono if @danberindei @pruivo you were serious about this, but maybe its time to really deprecate this?

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/cache/AsyncConfigurationBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/cache/AsyncConfigurationBuilder.java
@@ -17,7 +17,6 @@ import org.infinispan.util.logging.LogFactory;
  * If configured all communications are asynchronous, in that whenever a thread sends a message sent
  * over the wire, it does not wait for an acknowledgment before returning. Asynchronous configuration is mutually
  * exclusive with synchronous configuration.
- *
  */
 public class AsyncConfigurationBuilder extends AbstractClusteringConfigurationChildBuilder implements Builder<AsyncConfiguration> {
    private static final Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass(), Log.class);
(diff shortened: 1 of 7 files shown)


pruivo:
@rhusar yes I'm serious :) 
looks good to me @danberindei.

rhusar:
Thanks Pedro.

The reason why I "need" this is because we did not expose the replication queue thread pool configuration in WildFly knowing it would be deprecated. Yet it did not happen even for 8.1.0.Final. Because if we expose it and then shortly afterwards remove it, its a quite a bit of unnecessary and dull work to support that code for long time (in EAP).

danberindei:
@rhusar +1 to deprecate it. Since we now have a separate schema for 8.2, could you also change the description in `infinispan-config-8.2.xsd` and `jboss-infinispan-core_8_2.xsd`?

core/src/main/java/org/infinispan/configuration/cache/AsyncConfigurationBuilder.java
```diff
@@ -62,7 +61,10 @@ public AsyncConfigurationBuilder syncMarshalling() {
     * NOTE: Currently Infinispan will not use the object instance, but instead instantiate a new
     * instance of the class. Therefore, do not expect any state to survive, and provide a no-args
     * constructor to any instance. This will be resolved in Infinispan 5.2.0
+    *
+    * @deprecated superseded by JGroups bundler functionality
```
danberindei: Could you remove the `This will be resolved in Infinispan 5.2.0` sentence, and instead add "Since 8.2" to the `@deprecated` line?

danberindei: Actually I see we don't create a new instance any more, please remove the note paragraph completely.

rhusar:
Changes were incorporated as suggested.

pruivo:
@rhusar dont forget to update the ```xsd``` files please :)

rhusar:
Oops, done now, thanks @pruivo I totally overlooked it.

tristantarrant:
Pushed, thanks

rhusar:
Great, thanks!

PR #3888, opened 2015-12-09
Revert "ISPN-5876 Invalidate keys on commit if 2-phase"

bmaxwell:
This reverts commit 4f9556ca8d9c8c136be721de778e7a277c7d7b59.

Need to revert as this causes a change in default behavior, we need to rethink it, make an option to enable the new behavior
https://bugzilla.redhat.com/show_bug.cgi?id=1273147


slaskawi:
Hey @galderz! Could you please look into this? It seems you have broader knowledge in this area.

slaskawi:
Comments from me - it would be nice to have some JIRA + description what is the desired behavior.

galderz:
@bmaxwell Right, we could revert this but is the option to enable it needed at all? Did you verify that this change, in spite of changing the default, works as expected for the problem you had?

galderz:
We have agreed to revert this change and instead fix it by adding an system property to enable this behaviour.

galderz:
Integrated

PR #3893, opened 2015-12-14
ISPN-5981 Compatibility routing fix for clients

galderz:
https://issues.jboss.org/browse/ISPN-5981

* When compatibility is enabled, servers store data in POJO mode, hence
  clients should be routing based on the POJO version instead of the
  byte[] version.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
@@ -67,7 +67,6 @@ import org.infinispan.commons.util.Util;
  * <li><tt>infinispan.client.hotrod.force_return_values</tt>, default = false.  Whether or not to implicitly {@link org.infinispan.client.hotrod.Flag#FORCE_RETURN_VALUE} for all calls.</li>
  * <li><tt>infinispan.client.hotrod.tcp_no_delay</tt>, default = true.  Affects TCP NODELAY on the TCP stack.</li>
  * <li><tt>infinispan.client.hotrod.tcp_keep_alive</tt>, default = false.  Affects TCP KEEPALIVE on the TCP stack.</li>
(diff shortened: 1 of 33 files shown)


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
```diff
@@ -658,6 +658,7 @@ private Properties loadFromStream(InputStream stream) {
                   return null;
                }
             }
+            result.resolveCompatibility();
```
tristantarrant: Since resolveCompatibility() actually pings, this makes pingOnStartup() essentially redundant ?

galderz: Yeah, working on deprecating it...

galderz:
@tristantarrant Updated

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
```diff
@@ -650,14 +648,15 @@ private Properties loadFromStream(InputStream stream) {
...
-               }
+
+            // If ping not successful assume that the cache does not exist
+            // Default cache is always started, so don't do for it
+            if (!cacheName.equals(RemoteCacheManager.DEFAULT_CACHE_NAME) &&
+                  ping(result) == PingResult.CACHE_DOES_NOT_EXIST) {
+               return null;
             }
+
+            result.resolveCompatibility();
```
tristantarrant: You are still pinging twice. Can we collapse the cache resolution with the compatibility check in a single ping ?

galderz: True, fixing...

galderz:
Updated fix. ClientAsymmetricClusterTest randomly fails with NPE even without the ping changes, could be related to recent @danberindei changes? I've pinged him about it.

tristantarrant:
Pushed, thanks

PR #3894, opened 2015-12-14
ISPN-5988 Fine-/Coarse-grained AtomicMap info is not preserved in cache

pruivo:
https://issues.jboss.org/browse/ISPN-5988


core/src/main/java/org/infinispan/atomic/impl/AtomicHashMap.java
```diff
@@ -54,35 +53,43 @@
...
+   public static <K, V> AtomicHashMap<K, V> newInstance(Cache<Object, Object> cache, Object cacheKey, ProxyMode proxyMode) {
+      AtomicHashMap<K, V> value = new AtomicHashMap<>(proxyMode);
       Object oldValue = cache.putIfAbsent(cacheKey, value);
       if (oldValue != null) value = (AtomicHashMap<K, V>) oldValue;
       return value;
    }
 
    public AtomicHashMap() {
-      this.delegate = new FastCopyHashMap<K, V>();
+      this(new FastCopyHashMap<>(), ProxyMode.NONE);
```
danberindei: I'm not sure the `NONE` option is necessary: I removed the enum value and replaced it here with `COARSE`, and all the tests passed.

core/src/main/java/org/infinispan/atomic/impl/AtomicHashMap.java
```diff
@@ -254,4 +277,9 @@ public Integer getId() {
          return Util.<Class<? extends AtomicHashMap>>asSet(AtomicHashMap.class);
       }
    }
+
+   public enum ProxyMode {
+      NONE, FINE, COARSE;
+      public static final ProxyMode[] CACHED_VALUES = values();
```
danberindei: The same comment as for https://github.com/infinispan/infinispan/pull/3930, having this field public is confusing.

core/src/main/java/org/infinispan/atomic/impl/AtomicHashMap.java
pruivo: because that constructor is no longer used. I'll remove it.

pruivo: edit: actually, it is used in tests :)

danberindei:
Integrated, thanks Pedro!

PR #3896, opened 2015-12-15
ISPN-6024 HotRodServer should remove listeners on shutdown

danberindei:
https://issues.jboss.org/browse/ISPN-6024


tristantarrant:
Pushed, thanks

PR #3902, opened 2015-12-22
ISPN-5981 Deprecate pingOnStartup in spring4-remote component

danberindei:


tristantarrant:
Pushed to master and 8.1.x

PR #3903, opened 2015-12-22
ISPN-6033 QueryCache should not be visible to users

anistor:
Jira: https://issues.jboss.org/browse/ISPN-6033


gustavonalle:
integrated, thanks @anistor !

PR #3905, opened 2015-12-23
ISPN-6005 Avoid OOME when sending client events

galderz:
https://issues.jboss.org/browse/ISPN-6005

* When adding a client listener, if state is to be provided, do it
asynchronously to avoid locking up Netty event loop thread. This is
necessary because if the event loop thread is hanging, events cannot be
written and instead they get queued up.
* To avoid overloading clients, apply backpressure in server by checking
if the channel is writable before writing each event. For efficiency
reasons, events are batched now.

### DIFF
--- a/server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
+++ b/server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
@@ -2,8 +2,9 @@ package org.infinispan.server.hotrod
 
 import java.io.{ObjectInput, ObjectOutput}
 import java.lang.reflect.Constructor
-import java.util.concurrent.ConcurrentMap
+import java.util.concurrent._
 import java.util.concurrent.atomic.AtomicLong
+import java.util.function.{BiConsumer, Consumer, Function}
 
 import io.netty.channel.Channel
 import org.infinispan.commons.equivalence.{AnyEquivalence, ByteArrayEquivalence}
@@ -12,6 +13,7 @@ import org.infinispan.commons.marshall.{AbstractExternalizer, Marshaller}
 import org.infinispan.commons.util.CollectionFactory
 import org.infinispan.commons.util.concurrent.jdk8backported.EquivalentConcurrentHashMapV8
 import org.infinispan.container.versioning.NumericVersion
+import org.infinispan.factories.threads.DefaultThreadFactory
 import org.infinispan.metadata.Metadata
 import org.infinispan.notifications._
 import org.infinispan.notifications.cachelistener.annotation.{CacheEntryExpired, CacheEntryCreated, CacheEntryModified, CacheEntryRemoved}
@@ -41,6 +43,10 @@ class ClientListenerRegistry(configuration: HotRodServerConfiguration) extends L
    private val cacheEventFilterFactories = CollectionFactory.makeConcurrentMap[String, CacheEventFilterFactory](4, 0.9f, 16)
    private val cacheEventConverterFactories = CollectionFactory.makeConcurrentMap[String, CacheEventConverterFactory](4, 0.9f, 16)
    private val cacheEventFilterConverterFactories = CollectionFactory.makeConcurrentMap[String, CacheEventFilterConverterFactory](4, 0.9f, 16)
+
+   private val addListenerExecutor = new ThreadPoolExecutor(
+      0, 10, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue[Runnable],
+      new DefaultThreadFactory(null, 1, "add-listener-thread", null, null))
 
    def setEventMarshaller(eventMarshaller: Option[Marshaller]): Unit = {
       // Set a custom marshaller or reset to default if none
@@ -77,7 +83,7 @@ class ClientListenerRegistry(configuration: HotRodServerConfiguration) extends L
       cacheEventFilterConverterFactories.remove(name)
    }
 
-   def addClientListener(ch: Channel, h: HotRodHeader, listenerId: Bytes, cache: Cache,
+   def addClientListener(decoder: AbstractVersionedDecoder, ch: Channel, h: HotRodHeader, listenerId: Bytes, cache: Cache,
            includeState: Boolean, namedFactories: NamedFactories, useRawData: Boolean): Unit = {
       val eventType = ClientEventType.apply(namedFactories._2.isDefined, useRawData, h.version)
       val clientEventSender = ClientEventSender(includeState, ch, h.version, cache, listenerId, eventType)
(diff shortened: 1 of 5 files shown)


galderz:
All, we need to get moving with this one...

gustavonalle:
@galderz There are some test failures

galderz:
@gustavonalle Looking closer some might be related, I'll get those fixed ASAP

galderz:
@gustavonalle Tests should be fixed

anistor:
@galderz Do we want ClientEventsOOMTest to run during suite? I'd disable it and make a note.

galderz:
@anistor It does run, but with slower amount of data and sleep times. I think it's good to have such test run but we can also have a manual test that adds more data and has the original sleep times you had. WDYT?

anistor:
I'd move this to "stress" category and have it run with lots of large entries and big sleep time.

galderz:
@anistor Hmmm, the problem with stress tests is that they're manual. I'd rather have two, one lightweight, and then extend it for stress options. Why would you be against that?

anistor:
I'm not against that. Pls do it.

galderz:
Good, I'll update the PR...

gustavonalle:
Is the test in this PR able to reproduce the issue before the fix? If not, IMO it's redundant, and I'd prefer a stress test that can be run and verify that everything is fine.

anistor:
That's what I thought too, but if Galder would like to have a 'lighter' version of this test in the normal suite to be extra safe that's fine too. I'm thinking of entries just a bit larger than netty's buffers and quite few so as not to increase the duration of the suite.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
```diff
@@ -99,7 +105,28 @@ class ClientListenerRegistry(configuration: HotRodServerConfiguration) extends L
...
+                  if (cause == null) decoder.createSuccessResponse(h, null)
+                  else decoder.createErrorResponse(h, cause)
+
+               ch.writeAndFlush(resp)
+         }
+         })
+      } else {
+         cache.addListener(clientEventSender, filter.orNull, converter.orNull)
+         ch.writeAndFlush(decoder.createSuccessResponse(h, null))
+      }
```
gustavonalle: **Minor**:  there are some implicits in place in ```org.infinispan.server```'s ```package.scala``` to convert from java lambdas to scala lambdas. ```Runnable``` and ```BiConsumer``` are not there but could be added:
```scala
import java.util.function.{BiConsumer => J8BiConsumer}

implicit def asJavaBiConsumer[T, U](f: (T, U) => Unit): J8BiConsumer[T, U] = new J8BiConsumer[T, U] {
   override def accept(t: T, u: U): Unit = f(t, u)
}

implicit def asJavaRunnable(f: () => Unit): Runnable = new Runnable {
   override def run(): Unit = f()
}
```

allowing the above to be written as:

```
val cf = CompletableFuture.runAsync(() => cache.addListener(clientEventSender, filter.orNull, converter.orNull), addListenerExecutor)
```

and 

```scala
cf.whenComplete((t: Void, cause: Throwable) => {
      val resp = if (cause == null) decoder.createSuccessResponse(h, null)
               else decoder.createErrorResponse(h, cause)
      ch.writeAndFlush(resp)
      ()
 })
```

galderz:
> Is the test in this PR able to reproduce the issue before the fix?

Yes.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
galderz: Cool, I'll sort those out. 

As a FYI: using Scala Futures required more tinkering when those were used inside WildFly, so given the short amount of time to get this fixed, decided to fallback on Java.

gustavonalle:
@galderz  Apparently there are failures in the hotrod-server suite for me, let's see what CI tells

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
```diff
@@ -170,6 +197,19 @@ class ClientListenerRegistry(configuration: HotRodServerConfiguration) extends L
...
+   def findAndWriteEvents(channel: Channel): Unit = {
+      val eventSender = eventSenders.values().filter {
+         case s: BaseClientEventSender => s.hasChannel(channel)
+         case _ => false
+      }
+      eventSender.headOption.map {
+         case s: BaseClientEventSender => s.writeEventsIfPossible()
+         case _ => false
+      }
+   }
```
gustavonalle: Is it possible that more than one event sender is associated with the channel? 

Another way of writing it is replacing the two pattern matches with a single collect:

```scala
eventSenders.values().collectFirst { case s: BaseClientEventSender =>
         if(s.hasChannel(channel)) s.writeEventsIfPossible()
}
```
or using ```collect``` instead of ```collectFirst``` to assume that a channel can be part of more than one eventListener

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
```diff
@@ -216,9 +273,16 @@ class ClientListenerRegistry(configuration: HotRodServerConfiguration) extends L
...
 
-         ch.writeAndFlush(remoteEvent)
+         val waitingForFlush = !ch.isWritable
+         eventQueue.put(remoteEvent)
+
+         if (!waitingForFlush) {
+            ch.eventLoop().submit(new Runnable {
+               override def run(): Unit = writeEventsIfPossible()
+            })
+         }
```
gustavonalle: With the implicit conversion for ```Runnable```, this can be simplified to 
```
ch.eventLoop().submit(() => writeEventsIfPossible())
```

gustavonalle:
There are failures in ```hotrod-server```

server/hotrod/src/main/scala/org/infinispan/server/hotrod/Decoder2x.scala
```diff
@@ -383,9 +383,9 @@ object Decoder2x extends AbstractVersionedDecoder with ServerConstants with Log
                case _ => false
             }
             val reg = server.getClientListenerRegistry
-            reg.addClientListener(ch, h, listenerId, cache, includeState,
+            reg.addClientListener(this, ch, h, listenerId, cache, includeState,
                   (filterFactoryInfo, converterFactoryInfo), useRawData)
-            createSuccessResponse(h, null)
+            null
```
vjuranek: This causes NPE in  [HotRodTestingUtil#withClientListener](https://github.com/infinispan/infinispan/blob/master/server/hotrod/src/test/scala/org/infinispan/server/hotrod/test/HotRodTestingUtil.scala#L394) (and thus test failures)

galderz:
Looking into it

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
galderz: Neat :clap:

galderz:
All concerns have been addressed and tests should be fixed now

galderz:
Actually, still some issues pending...

galderz:
Looking good now

galderz:
Actually, one more thing! The OOME stress test is running...

galderz:
Ok, talking to Dan on IRC, we reached final decision: we're removing the stress test version since the functional test already shows the problem. Instead, I''ll ammend the functional test to optionally take sys props that can be used to easily alter the number of entries and client sleep time.

galderz:
Should be finally be ready :)

anistor:
@galderz Is it possible to avoid the wrapping into a java.util.concurrent.CompletionException? I'm referring to the same issue @vjuranek seen in RemoteListenerWithDslFilterTest.

galderz:
@anistor I'll see what it can done... but remember that this is urgent ;)

galderz:
@anistor done

anistor:
thanks! I still see ```// Regular expression adjusted to support multi-line messages```

galderz:
@anistor done

vjuranek:
I hit some more (compared to TC) test failures in HR client module, but all of them are random and don't seem to be related to the change (I was able to find most of them in TC already failing in the past).

gustavonalle:
@vjuranek Looks fine to me, I've run a few times and no failures

gustavonalle:
Integrated, thanks @galderz !

galderz:
Great, thanks all for the help sorting this out :)

PR #3909, opened 2015-12-23
ISPN-5993 ClassCastException in ProtobufMetadataManagerInterceptor fo…

anistor:
…r clear

Since clear is now non-tx we can just remove the lines of code that attempt to lock .errors key and the issue is gone.

Jira: https://issues.jboss.org/browse/ISPN-5993


danberindei:
Integrated, thanks Adrian!

danberindei:
Integrated in 8.1.x as well.

PR #3911, opened 2015-12-24
ISPN-5889 Fix cache name in HotRodMergeTest

danberindei:
RemoteCacheManager.getCache("") returns the default cache, but
EmbeddedCacheManager.getCache("") creates a new cache named "".

Since the test works even with the "" cache name, I haven't opened a new issue in JIRA.


anistor:
Looking ...

anistor:
Integrated in master and 8.1.x. Thanks @danberindei !

PR #3912, opened 2015-12-24
ISPN-5413 Index Affinity support

gustavonalle:
https://issues.jboss.org/browse/ISPN-5413

Requires Hibernate Search master

##### Overview #####
This PR introduces a new IndexManager: ```org.infinispan.query.affinity.ShardIndexManager``` that implements the index affinity as described on https://github.com/infinispan/infinispan/wiki/Index-affinity-proposal. This index manager does not have a single master, it creates 1 shard per Infinispan segment, so each node indexes owned data locally.

##### Usage #####
Currently, at a minimum it's necessary to specify the index manager name in the configuration plus the custom key partitioner in the cache. Ex:

```
ConfigurationBuilder cfg = ...
cfg.clustering().hash().keyPartitioner(new AffinityPartitioner());
cfg.indexing().index(Index.ALL).addProperty("<my-index>.indexmanager", ShardIndexManager.class.getName())
```

##### Integration #####

It should blend well with other query aspects: Mass Indexer, sync/async backend, multiple entities per cache, manual indexing, flags, etc.

##### TODO  #####

* ~~Automatic usage of SyncConsistentHash so that index caches and data caches agree on ownership~~
* The configurable ratio segments/shard is not implemented yet.
* Could use a bit more tests to have the same coverage of the InfinispanIndexManager.
* Perf. tests


core/src/main/java/org/infinispan/distribution/ch/AffinityTaggedKey.java
```diff
@@ -0,0 +1,12 @@
+package org.infinispan.distribution.ch;
+
+public interface AffinityTaggedKey {
+
+   /**
+    * This numeric id is used exclusively for storage affinity in Infinispan.
+    *
+    * @return the segment id to be used for storage, or -1 to fall back to normal owner selection.
```
slaskawi: Maybe we could throw something to indicate that we need to get back to normal owner?

slaskawi: Or maybe some isEnabled flag?

Sanne:
hi @gustavonalle , could you rebase it one latest Infinispan master please?
Does it still need Hibernate Search "master"? We can tag some preview to try start merging this (all of it or parts)

Sanne:
Shouldn't we add among the TODOs that infinispan-core should keep different hashtables for different caches in synch? (as planned, I'm not sure if that was implemented)

gustavonalle:
@Sanne rebased. 
>Does it still need Hibernate Search "master"?

It needs HSEARCH-2012 & HSEARCH-2075, currently not relased

>Shouldn't we add among the TODOs that infinispan-core should keep different hashtables for different caches in synch? (as planned, I'm not sure if that was implemented)

https://issues.jboss.org/browse/ISPN-4851 was integrated, is that what you are alluding to?

Sanne:
> https://issues.jboss.org/browse/ISPN-4851 was integrated, is that what you are alluding to?

Yes. That's great! I didn't know.

core/src/main/java/org/infinispan/distribution/ch/AffinityTaggedKey.java
gustavonalle: > Maybe we could throw something to indicate that we need to get back to normal owner? 

Throwing an exception rather than returning -1 is considerably slower, apart for polluting every caller code, I personally prefer to avoid it. Maybe it'd possible to get rid of this ```-1``` altogether and assume that this key will always have a valid segment, will check.

> Or maybe some isEnabled flag?

Could you elaborate on that? Would it be set via configuration?

Sanne:
needs yet another rebase.. hopefully the last one ;)

gustavonalle:
Rebased. 
There is one test failing ```ShardedIndexManagerMassIndexTest.testReindexing```,needs more investigation. Apparently the ```org.hibernate.search.backend.spi.BatchBackend``` is not calculating the correct shards during a Purge and Flush jobs

tristantarrant:
@gustavonalle you did not push the rebased code :)

gustavonalle:
@tristantarrant yes, I pushed, but master moved on since them, rebasing again...

gustavonalle:
@tristantarrant rebased, disabled the failing test temporarily

gustavonalle:
Rebased and fixed the failing test

Sanne:
merged!

:dancers:

PR #3913, opened 2015-12-29
ISPN-6042 NPE when using avg on repeated property and one group does not contain any value

anistor:
Jira: https://issues.jboss.org/browse/ISPN-6042


gustavonalle:
Integrated with a trivial change to fix a test failure, thanks @anistor !

PR #3916, opened 2016-01-03
hibernate-search-engine:tests to test scope.

kazuhira-r:
"hibernate-search-engine: tests" is, it'll be correct in is given to the test scope?


gustavonalle:
merged, thanks @kazuhira-r!

PR #3917, opened 2016-01-04
ISPN-6033 QueryCache should not be visible to users

anistor:
jira: https://issues.jboss.org/browse/ISPN-6033


anistor:
Integrated in 8.1.x.

PR #3918, opened 2016-01-04
ISPN-6042 NPE when using avg on repeated property and one group does …

anistor:
…not contain any value
jira: https://issues.jboss.org/browse/ISPN-6042


anistor:
Integrated in 8.1.x.

PR #3921, opened 2016-01-05
ISPN-6005 Add thread index for add listener thread

galderz:


gustavonalle:
merged, thanks @galderz !

PR #3922, opened 2016-01-06
BZ-1279757 ISPN-5937  ExpiryTest.testGlobalExpiryReplaceFlag fails randomly

mvrabel:


danberindei:
@mvrabel Are you sure this fixes the failure?

mvrabel:
@danberindei yes, I am. without the sleep : http://pastebin.com/yxnK1HHQ
with sleep :   http://pastebin.com/kSc8RFNY
I think it's because ClusterExpirationManager- > handleInStoreExpiration
@Override
   public void handleInStoreExpiration(K key) {
expiring.put(key, key);
      // Unfortunately stores don't pull the entry so we can't tell exactly why it expired and thus we have to remove
      // the entire value.  Unfortunately this could cause a concurrent write to be undone
      try {
         removeExpired(key, null, null);
      } finally {
         expiring.remove(key);
      }
Trace log  http://pastebin.com/JXk23Rw1. You can here that removeExpired  is called after the test passed

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExpiryTest.java
```diff
@@ -38,6 +39,17 @@ protected void createCacheManagers() throws Throwable {
...
    }
 
+   @AfterMethod
+   public void after() {
+      try {
+         // https://bugzilla.redhat.com/show_bug.cgi?id=1279757
+         // https://issues.jboss.org/browse/ISPN-5937
+         Thread.sleep(500);
+      } catch (InterruptedException e) {
+         e.printStackTrace();
```
slaskawi: Maybe it would be better to add this to "throws" on method signature level?

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExpiryTest.java
```diff
@@ -38,6 +39,17 @@ protected void createCacheManagers() throws Throwable {
       TestingUtil.replaceComponent(cacheManagers.get(0), TimeService.class, timeService, true);
    }
 
+   @AfterMethod
+   public void after() {
+      try {
+         // https://bugzilla.redhat.com/show_bug.cgi?id=1279757
+         // https://issues.jboss.org/browse/ISPN-5937
+         Thread.sleep(500);
```
slaskawi: Isn't there any better way? This will slow down the testsuite.

danberindei:
@mvrabel Thanks for the details, the JIRA/PR descriptions are a bit sparse :)

I guess we could work around this by using different keys in each test method instead of sleeping.

However, the comment in `ClusterExpirationManager.handleInStoreExpiration()` doesn't seem to apply here, because `ExpiryTest` doesn't have any store configured. So I'm thinking this might  be a problem in the way we are handling in-memory expiration, i.e. in the production code and not in the test.

mvrabel:
@danberindei should I create a new PR with different keys ? ... and let the developers deal with the expiration

danberindei:
@mvrabel Yes, please either update this PR or create another one with different keys. I'll create separate issues in JIRA for the problems I've noticed in the production code.

mvrabel:
@danberindei I created a new PR https://github.com/infinispan/infinispan/pull/3927

PR #3923, opened 2016-01-08
ISPN-5447 fixed spring schema usage in tests

michalszynkiewicz:
1. infinispan-spring schemas renamed to 8.2 (the number was used in spring4 pom.xml)
2. spring.schemas now pointing to proper version of schema


spring/spring/pom.xml
```diff
@@ -237,6 +237,43 @@
                     </instructions>
                 </configuration>
             </plugin>
+            <plugin>
```
slaskawi: Currently we support both Spring 3 (placed in this module) and Spring 4 (placed here https://github.com/infinispan/infinispan/tree/master/spring/spring4). Please add similar changes to Spring 4. Spring 3 support will be dropped soon.

slaskawi: Clarified... Spring 4 works a bit differently because of Spring commons package. Everything is fine.

michalszynkiewicz: spring 4 works differently. XSD is in spring4-common, while tests are in spring4-embedded.
In such situation there's no problem with classpath resolution of the XSD.

michalszynkiewicz: this turns out to be unnecessary after changing xsd version to 8.2

I will remove changes in pom.xml.

slaskawi:
:+1: Looks good. Let's wait for CI

tristantarrant:
Pushed, thanks

PR #3926, opened 2016-01-10
[8.1.x] ISPN-6049 Calls to reflection API fail with enabled security manager;…

rhusar:
… need to be run in privileged block

Jira
https://issues.jboss.org/browse/ISPN-6049

### DIFF
--- a/core/src/main/java/org/infinispan/distribution/group/GroupManagerImpl.java
+++ b/core/src/main/java/org/infinispan/distribution/group/GroupManagerImpl.java
@@ -1,7 +1,5 @@
 package org.infinispan.distribution.group;
 
-import static org.infinispan.commons.util.ReflectionUtil.invokeAccessibly;
-
 import org.infinispan.commons.util.CollectionFactory;
 import org.infinispan.commons.util.InfinispanCollections;
 import org.infinispan.commons.util.ReflectionUtil;
@@ -11,14 +9,18 @@ import org.infinispan.interceptors.locking.ClusteringDependentLogic;
 import org.infinispan.remoting.transport.Address;
 
 import java.lang.reflect.Method;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.ConcurrentMap;
 
+import static org.infinispan.commons.util.ReflectionUtil.invokeAccessibly;
+
 
 public class GroupManagerImpl implements GroupManager {
     
-    private static interface GroupMetadata {
+    private interface GroupMetadata {
         
         GroupMetadata NONE = new GroupMetadata() {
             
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/infinispan/distribution/group/GroupManagerImpl.java
```diff
@@ -46,13 +48,16 @@ public GroupMetadataImpl(Method method) {
 
         @Override
         public String getGroup(Object instance) {
-            return String.class.cast(invokeAccessibly(instance, method, Util.EMPTY_OBJECT_ARRAY));
+            PrivilegedAction<Object> action = () -> invokeAccessibly(instance, method, Util.EMPTY_OBJECT_ARRAY);
+            Object object = System.getSecurityManager() == null ? action.run() : AccessController.doPrivileged(action);
+            return String.class.cast(object);
```
tristantarrant: Instead of always invoking the lambda like that I'd rather do
if (System.getSecurityManager() == null) {
return String.class.cast(invokeAccessibly(instance, method, Util.EMPTY_OBJECT_ARRAY));
} else {
... // priv action
}

rhusar: Even given that would introduce code duplication and more LOC to manage?

tristantarrant: It saves allocating one object in the normal case and is much more readable

rhusar: Ok, let me rework that to your liking :-)

rhusar: Done, no lambdas are invoked in case security manager is not configured.

core/src/main/java/org/infinispan/notifications/impl/AbstractListenerImpl.java
```diff
@@ -179,7 +181,15 @@ protected boolean validateAndAddListenerInvocations(Object listener, AbstractInv
       Listener l = testListenerClassValidity(listener.getClass());
       boolean foundMethods = false;
       builder.setTarget(listener);
-      builder.setSubject(Security.getSubject());
+
+      Subject subject;
+      if (System.getSecurityManager() == null) {
+         subject = Security.getSubject();
```
tristantarrant: This check should be inside the Security class, not here.

tristantarrant:
Also, please this should be issued against master with a request to backport

core/src/main/java/org/infinispan/notifications/impl/AbstractListenerImpl.java
rhusar: That does not sound right. The org.infinispan.security.Security class is public so anyone can access it. Neither org.infinispan.security.Security#isTrustedClass is right, since anyone can have classes in a package org.infinispan.blabla that even jar signing won't solve.

tristantarrant: That's in fact "lightweight" security. It has no pretense of protecting you from malicious code, but it is much faster than a SecurityManager :) If one is present it will fallback to that.

rhusar:
The priority is 8.1.x for us to get into WF/EAP. For master PR, please see https://github.com/infinispan/infinispan/pull/3946.

tristantarrant:
Pushed, thanks

PR #3927, opened 2016-01-11
BZ-1279757 ISPN-5937 ExpiryTest.testGlobalExpiryReplaceFlag fails randomly

mvrabel:
This should probably also fix BZ-1293303


danberindei:
Integrated, thanks Martin!

I just removed the BZ reference from the commit comment.

PR #3928, opened 2016-01-11
ISPN-6056 Improve ReplicableCommand marshalling

pruivo:
https://issues.jboss.org/browse/ISPN-6056

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/marshall/MarshallUtil.java
+++ b/commons/src/main/java/org/infinispan/commons/marshall/MarshallUtil.java
@@ -1,7 +1,7 @@
 package org.infinispan.commons.marshall;
 
 import net.jcip.annotations.Immutable;
-import org.infinispan.commons.io.UnsignedNumeric;
+import org.infinispan.commons.util.Util;
 
 import java.io.IOException;
 import java.io.ObjectInput;
@@ -9,6 +9,7 @@ import java.io.ObjectOutput;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Objects;
+import java.util.UUID;
 
 /**
  * MarshallUtil.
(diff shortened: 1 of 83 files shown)


commons/src/main/java/org/infinispan/commons/marshall/MarshallUtil.java
```diff
@@ -21,27 +22,327 @@
...
+    * @throws IOException            If any of the usual Input/Output related exceptions occur.
+    * @throws ClassNotFoundException If the class of a serialized object cannot be found.
+    * @see {@link #marshallMap(Map, ObjectOutput)}
+    */
+   public static <K, V, T extends Map<K, V>> T unmarshallMap(ObjectInput in, MapBuilder<K, V, T> builder) throws IOException, ClassNotFoundException {
+      final int size = unmarshallInt(in);
+      if (size == NULL_VALUE) {
+         return null;
+      }
+      final T map = Objects.requireNonNull(builder, "MapBuilder must be non-null").build();
```
tristantarrant: Since we know the size the map will have, why not enhance the MapBuilder to also support an initial size to avoid resizing while unmarshalling.

tristantarrant:
@pruivo one minor comment

pruivo:
ready for review again :)

tristantarrant:
Pushed, thanks

core/src/main/java/org/infinispan/topology/CacheTopologyControlCommand.java
```diff
@@ -248,27 +253,127 @@ public byte getCommandId() {
    }
 
    @Override
-   public Object[] getParameters() {
-      return new Object[]{cacheName, (byte) type.ordinal(), sender, joinInfo, topologyId, rebalanceId, currentCH,
-            pendingCH, availabilityMode, actualMembers, throwable, viewId};
+   public void writeTo(ObjectOutput output) throws IOException {
+      MarshallUtil.marshallString(cacheName, output);
+      MarshallUtil.marshallEnum(type, output);
+      switch (type) {
```
danberindei: Selectively marshalling parameters probably makes sense more for commands that are used more often... for `CacheTopologyControlCommand`, I think serializing everything unconditionally would have been just fine.

PR #3930, opened 2016-01-12
ISPN-6061 Cache Enum.values()

pruivo:
https://issues.jboss.org/browse/ISPN-6061

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/marshall/MarshallUtil.java
+++ b/commons/src/main/java/org/infinispan/commons/marshall/MarshallUtil.java
@@ -1,14 +1,14 @@
 package org.infinispan.commons.marshall;
 
 import net.jcip.annotations.Immutable;
+import org.infinispan.commons.io.UnsignedNumeric;
 
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.util.Collection;
 import java.util.Map;
-
-import org.infinispan.commons.io.UnsignedNumeric;
+import java.util.Objects;
 
 /**
  * MarshallUtil.
@@ -18,6 +18,8 @@ import org.infinispan.commons.io.UnsignedNumeric;
  */
 @Immutable
 public class MarshallUtil {
+
+   private static final byte NULL_VALUE = -1;
 
    public static void marshallCollection(Collection<?> c, ObjectOutput out) throws IOException {
       UnsignedNumeric.writeUnsignedInt(out, c.size());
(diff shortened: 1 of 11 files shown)


tristantarrant:
Any allocation stats ?

pruivo:
allocation for 1 min run (the biggest enum we have)
```
Flag[]:          703 allocations, 50.25 MB 
```

danberindei:
@pruivo Only 703 allocations in 1 minute? Is this before or after your change?

commons/src/main/java/org/infinispan/commons/marshall/ValueMatcherMode.java
```diff
@@ -9,6 +9,8 @@
    MATCH_EXPECTED,
    MATCH_EXPECTED_OR_NEW,
    MATCH_NON_NULL,
-   MATCH_NEVER
+   MATCH_NEVER;
+
+   public static final ValueMatcherMode[] CACHED_VALUES = values();
```
danberindei: Could we not make this public? It looks weird in the auto-completion list.  I'd rather add a public method that does the lookup, i.e. `public static ValueMatcherMode valueOf(int ordinal)`.

commons/src/main/java/org/infinispan/commons/marshall/MarshallUtil.java
```diff
@@ -41,4 +41,28 @@ public static void unmarshallMap(Map<Object, Object> map, ObjectInput in) throws
...
+      try {
+         return values[input.readByte()];
+      } catch (ArrayIndexOutOfBoundsException e) {
+         throw new IOException("Unknown enum.", e);
+      }
+   }
+
+   private static boolean checkIfNull(Object obj, ObjectOutput output) throws IOException {
+      final boolean isNull = obj == null;
+      output.writeBoolean(isNull);
```
danberindei: I'm pretty sure JBoss Marshalling takes care of the `null` case by itself, because it couldn't pick the right externalizer for a `null` value, so we could just throw an exception here.

pruivo: true but we are writing the enum's ordinal and not the object ```output.writeByte(e.ordinal())```. If the enum is null, it will throw an NPE

pruivo:
rebased. 
@danberindei the allocations are before. After the changes, we have no more allocations.

commons/src/main/java/org/infinispan/commons/marshall/MarshallUtil.java
danberindei: The question is, can the enum be `null` at this point? `marshallEnum()` is only called by the externalizers, which are only called by JBoss Marshalling.

pruivo: right, it is never null on this PR, but I'll need it in the future. 

https://github.com/infinispan/infinispan/pull/3928/files#diff-90b9b3956301b89b9c2f9307d09b6fecR278

danberindei: Ok, you should have started with that :)

Still, I don't like the idea of writing an extra byte for each enum, you could write a `-1` ordinal instead.

pruivo: writing -1 looks a good idea :)

tristantarrant:
Pushed, thanks

PR #3931, opened 2016-01-12
ISPN-6062 Remove megamorphic calls when calling next method in interceptor stack

rvansa:
https://issues.jboss.org/browse/ISPN-6062


rvansa:
I have checked that non-transactional configuration has the stack inlined from InterceptorChain.invoke() up to Command.perform(). However, the performance increase is not dramatic. Without CPU fully utilized, the improvement seems to be around 0.5 - 1%, but I need to do run more tests to bring the variance down. I got some strange results (big variance between nodes) in one run with CPU more utilized (by running 32 CPU-consuming threads on each node in parallel with the test), so although in this scenario the optimization should have bigger impact, I can't confirm it now.

So, at this point this should be considered an experimental feature.

rvansa:
To explain why there's a commit changing a bunch of tests:

The implementation does not modify any of the regular interceptors, instead it creates new classes and copies (and adjusts) the code. In constructors of these new classes, all fields from the original interceptor are copied. That leads us to one limitation: if you keep a reference to the interceptor and read/write it's fields, it's not propagated to the actually used 'inlined interceptor'. The workaround is to use e.g. AtomicBoolean instead of boolean fields etc. Some reflection-based tricks the tests use (I've seen something in infinispan-extended-statistics) can't work either, but that's mostly because these are nasty breaks of encapsulation.

Another change is that a command does not react on change of interceptor stack during its execution - if you block a command in one interceptor, remove interceptor underneath and then release it, it still follows the old interceptor stack. As the 'next' field was non-volatile, this was not properly guaranteed either, but I've stumbled upon a test that got broken because of this.

rvansa:
And I know that the interceptor stack is about to change, but
a) I wanted to see the effect of megamorphic calls
b) It's possible that the optimization will be adapted to the new architecture

slaskawi:
I see this PR plays with interceptor chain. Does it fit somehow to @danberindei sequential interceptor? Or this PR is no longer valid?

rvansa:
Abandoned as it does not offer enough gains.

PR #3935, opened 2016-01-13
ISPN-6068 Upgrade to JGroups 3.6.7.Final

danberindei:
https://issues.jboss.org/browse/ISPN-6068

* Use TCP_NIO2 in the test suite.
* Remove the log4j2 workaround in CommandAwareRpcDispatcher

The test suite seems to much faster with the new version of TCP_NIO2 (and about the same speed as with regular TCP).

### DIFF
--- a/core/src/test/resources/stacks/tcp.xml
+++ b/core/src/test/resources/stacks/tcp.xml
@@ -1,7 +1,8 @@
 <config xmlns="urn:org:jgroups"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/JGroups-3.4.xsd">
-   <TCP
+        xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/JGroups-3.6.xsd">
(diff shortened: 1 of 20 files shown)


core/src/test/resources/stacks/tcp.xml
```diff
@@ -1,26 +1,27 @@
...
          port_range="30"
-         recv_buf_size="20m"
+         recv_buf_size="2m"
          send_buf_size="640k"
          max_bundle_size="31k"
          use_send_queues="true"
          enable_diagnostics="false"
          tcp_nodelay="true"
-         bundler_type="sender-sends-with-timer"
+         bundler_type="transfer-queue"
```
pruivo: isn't ```sender-sends-with-timer``` better?

danberindei: Oops, I need to revisit these config changes... I had the TCP_NIO2 changes stashed in a local branch for a long time :)

danberindei: I removed the extra changes.

pruivo:
pushing...

pruivo:

PR #3936, opened 2016-01-13
ISPN-6069 Unify the continuous query API for remote and embedded mode

anistor:
Jira: https://issues.jboss.org/browse/ISPN-6069

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ContinuousQueryImpl.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ContinuousQueryImpl.java
@@ -1,0 +1,112 @@
+package org.infinispan.client.hotrod.event;
+
+import org.infinispan.client.hotrod.RemoteCache;
+import org.infinispan.client.hotrod.annotation.*;
+import org.infinispan.client.hotrod.filter.Filters;
+import org.infinispan.client.hotrod.marshall.ProtoStreamMarshaller;
+import org.infinispan.protostream.ProtobufUtil;
+import org.infinispan.protostream.SerializationContext;
+import org.infinispan.query.dsl.Query;
+import org.infinispan.query.api.continuous.ContinuousQuery;
+import org.infinispan.query.api.continuous.ContinuousQueryListener;
+import org.infinispan.query.remote.client.ContinuousQueryResult;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * A container of continuous query listeners for a cache.
+ * <p>This class is not threadsafe.
+ *
+ * @author anistor@redhat.com
+ * @since 8.2
+ * @private
+ */
+public final class ContinuousQueryImpl<K, V> implements ContinuousQuery<K, V> {
+
+   private final RemoteCache<K, V> cache;
+
+   private final SerializationContext serializationContext;
+
+   private final List<ClientEntryListener> listeners = new ArrayList<ClientEntryListener>();
+
+   public ContinuousQueryImpl(RemoteCache<K, V> cache) {
+      if (cache == null) {
+         throw new IllegalArgumentException("cache parameter cannot be null");
+      }
+      this.cache = cache;
+      serializationContext = ProtoStreamMarshaller.getSerializationContext(cache.getRemoteCacheManager());
+   }
+
+   /**
+    * Registers a continuous query listener that uses a query DSL based filter. The listener will receive notifications
+    * when a cache entry joins or leaves the matching set defined by the query.
+    *
+    * @param listener the continuous query listener instance
+    * @param query    the query to be used for determining the matching set
+    */
+   public <C> void addContinuousQueryListener(Query query, ContinuousQueryListener<K, C> listener) {
+      ClientEntryListener eventListener = new ClientEntryListener(serializationContext, listener);
+      Object[] factoryParams = Filters.makeFactoryParams(query);
+      cache.addClientListener(eventListener, factoryParams, null);
+      listeners.add(eventListener);
+   }
+
+   public void removeContinuousQueryListener(ContinuousQueryListener<K, ?> listener) {
+      for (Iterator<ClientEntryListener> it = listeners.iterator(); it.hasNext(); ) {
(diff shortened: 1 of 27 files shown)


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ContinuousQueryImpl.java
```diff
@@ -0,0 +1,112 @@
...
+   public void removeContinuousQueryListener(ContinuousQueryListener<K, ?> listener) {
+      for (Iterator<ClientEntryListener> it = listeners.iterator(); it.hasNext(); ) {
+         ClientEntryListener l = it.next();
+         if (l.listener == listener) {
+            cache.removeClientListener(l);
+            it.remove();
+            break;
+         }
+      }
+   }
```
gustavonalle: Does this method need to be java 6 compliant?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ContinuousQueryImpl.java
```diff
@@ -0,0 +1,112 @@
...
+   public void removeAllListeners() {
+      for (ClientEntryListener l : listeners) {
+         cache.removeClientListener(l);
+      }
+      listeners.clear();
+   }
+
+   @ClientListener(filterFactoryName = Filters.CONTINUOUS_QUERY_FILTER_FACTORY_NAME,
+         converterFactoryName = Filters.CONTINUOUS_QUERY_FILTER_FACTORY_NAME,
+         useRawData = true, includeCurrentState = true)
```
gustavonalle: Not really strictly related to this PR, but WDYT about deprecating the ```KeyValueFilter``` and the ```Converter``` from ```@ClientListener``` and from continuous query, replacing both by a single ```KeyValueFilterConverter```?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ContinuousQueryImpl.java
```diff
@@ -0,0 +1,112 @@
...
+      public void handleClientCacheEntryCreatedEvent(ClientCacheEntryCustomEvent<byte[]> event) throws IOException {
+         ContinuousQueryResult cqr = ProtobufUtil.fromByteArray(serializationContext, event.getEventData(), ContinuousQueryResult.class);
+         Object key = ProtobufUtil.fromWrappedByteArray(serializationContext, cqr.getKey());
+         Object value = cqr.getValue() != null ? ProtobufUtil.fromWrappedByteArray(serializationContext, cqr.getValue()) : cqr.getProjection();
+         if (cqr.isJoining()) {
+            listener.resultJoining(key, value);
+         } else {
+            listener.resultLeaving(key);
+         }
+      }
```
gustavonalle: Nitpicking:
``` return cqr.isJoining() ? listener.resultJoining(key, value) : listener.resultLeaving(key); ```

File: query-dsl/src/main/java/org/infinispan/query/dsl/continuous/ContinuousQuery.java [hunk 4]
```diff
@@ -0,0 +1,23 @@
...
+public interface ContinuousQuery<K, V> {
+
+   <C> void addContinuousQueryListener(Query query, ContinuousQueryListener<K, C> listener);
+
+   void removeContinuousQueryListener(ContinuousQueryListener<K, ?> listener);
+
+   List<ContinuousQueryListener<K, ?>> getListeners();
+
+   void removeAllListeners();
+}
```
gustavonalle: Are ```ContinuousQuery``` and ```ContinuousQueryListener``` really part of the query dsl ? Probably they should live on ```org.infinispan.query.continuous```, or maybe the artifact ```query-dsl``` should be called ```query-commons```, wdyt?

gustavonalle:
@anistor nice refactor :+1:

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ContinuousQueryImpl.java
anistor: No.

query-dsl/src/main/java/org/infinispan/query/dsl/continuous/ContinuousQuery.java
anistor: They are not really part of the DSL. I'll move them to ```org.infinispan.query.continuous``` package. I think we could rename ```query-dsl``` module to ```query-api``` or something like that.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ContinuousQueryImpl.java
anistor: I think we can solve this easily in a backward compatible manner. If ```filterFactoryName``` is specified and the factory produces CacheEventFilterConverterFactory then we should internally do both filtering and conversion (in same step) even if no ```converterFactoryName``` is specified. Same if both attributes are specified and are equal. If both attributes are specified and are not equal then we should do separate filtering and conversion. 

This is how core already handles filtering + conversion in same step. Not sure why client events are handled a bit differently.

query-dsl/src/main/java/org/infinispan/query/dsl/continuous/ContinuousQuery.java
anistor: I remember why I did not initally move them to ```org.infinispan.query.continuous```. OSGI hates split packages. So I'm a bit out of ideas.

gustavonalle: Maybe ```org.infinispan.query.api.continuous``` ?

anistor: I moved the CQ interfaces to ```org.infinispan.query.api.continuous``` package in query-dsl module. This required some hacking of the osgi imports/exports. 

I'd also rename query-dsl module to query-api in future.

gustavonalle:
looks good @anistor, integrating...

gustavonalle:
integrated, thanks @anistor

PR #3937, opened 2016-01-14
ISPN-6073 Replace Set<Flag> by a bit set

pruivo:
https://issues.jboss.org/browse/ISPN-6073

API change: LocalFlagAffectedCommand

### DIFF
--- a/core/src/main/java/org/infinispan/commands/AbstractLocalFlagAffectedCommand.java
+++ b/core/src/main/java/org/infinispan/commands/AbstractLocalFlagAffectedCommand.java
@@ -1,8 +1,7 @@
 package org.infinispan.commands;
 
+import org.infinispan.commons.util.EnumUtil;
 import org.infinispan.context.Flag;
-
-import java.util.Set;
 
 /**
  * Base class for those local commands that can carry flags.
@@ -12,15 +11,23 @@ import java.util.Set;
  */
 public abstract class AbstractLocalFlagAffectedCommand implements LocalFlagAffectedCommand {
 
-   protected Set<Flag> flags;
+   private long flags = 0;
 
    @Override
-   public Set<Flag> getFlags() {
+   public long getFlagsBitSet() {
       return flags;
    }
 
    @Override
-   public void setFlags(Set<Flag> flags) {
-      this.flags = flags;
+   public void setFlagsBitSet(long bitSet) {
+      this.flags = bitSet;
+   }
+
+   protected final boolean hasSameFlags(LocalFlagAffectedCommand other) {
+      return this.flags == other.getFlagsBitSet();
+   }
+
+   protected final String printFlags() {
+      return EnumUtil.prettyPrintBitSet(flags, Flag.class);
    }
 }
--- a/core/src/main/java/org/infinispan/commands/CommandsFactory.java
+++ b/core/src/main/java/org/infinispan/commands/CommandsFactory.java
@@ -97,39 +97,41 @@ public interface CommandsFactory {
     * @param key key to put
     * @param value value to put
     * @param metadata metadata of entry
-    * @param flags Command flags provided by cache
+    * @param flagsBitSet Command flags provided by cache
     * @return a PutKeyValueCommand
     */
-   PutKeyValueCommand buildPutKeyValueCommand(Object key, Object value, Metadata metadata, Set<Flag> flags);
+   PutKeyValueCommand buildPutKeyValueCommand(Object key, Object value, Metadata metadata, long flagsBitSet);
 
    /**
     * Builds a RemoveCommand
     * @param key key to remove
     * @param value value to check for ina  conditional remove, or null for an unconditional remove.
-    * @param flags Command flags provided by cache
+    * @param flagsBitSet Command flags provided by cache
     * @return a RemoveCommand
     */
-   RemoveCommand buildRemoveCommand(Object key, Object value, Set<Flag> flags);
+   RemoveCommand buildRemoveCommand(Object key, Object value, long flagsBitSet);
 
    /**
     * Builds an InvalidateCommand
-    * @param flags Command flags provided by cache
+    * @param flagsBitSet Command flags provided by cache
     * @param keys keys to invalidate
     * @return an InvalidateCommand
     */
-   InvalidateCommand buildInvalidateCommand(Set<Flag> flags, Object... keys);
+   InvalidateCommand buildInvalidateCommand(long flagsBitSet, Object... keys);
 
    /**
     * Builds an InvalidateFromL1Command
+    *
+    * @param flagsBitSet Command flags provided by cache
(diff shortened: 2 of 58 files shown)


tristantarrant:
Nice. How much does this change in terms of allocation costs ?

pruivo:
@tristantarrant you have the results in the mailing list with the flight recorder snapshots. 
The resume is (1 min run):
```
Allocations (Master):
Long:           1356 allocations, 187.51 MB
RegularEnumSet: 1762 allocations, 126.45 MB

Allocations (BitSet)
Long:           1677 (+321) allocations, 211.70 (+24.19) MB
RegularEnumSet: 0 (-1762), 0 (-126.45) MB
```

danberindei:
@pruivo there's something off with those numbers, there's no way 321 Longs can use 24MB...

pruivo:
@danberindei from my little understanding in Flight Recorder, it is the TLAB allocations, not the object allocations.

danberindei:
@pruivo Ok, looks like they aren't counting all allocations, they are only sampling the allocations that don't fit in the current [TLAB][1]. I suppose they still count everything when it comes to the total size, although I'm not sure how they do it...

Anyway, the number and size of allocations per minute will depend not just on the benchmark you're running, but also on the machine you're running it on. JMH has a `gc` profiler that you can enable with `-prof gc` and that gives you the allocation rate per operation. I haven't used it enough to confirm that it's reliable for multithreaded, multi-node tests, but it's on my list of things to try :)

[1]: http://hirt.se/blog/?p=381

core/src/main/java/org/infinispan/commands/AbstractLocalFlagAffectedCommand.java
```diff
@@ -12,15 +11,23 @@
  */
 public abstract class AbstractLocalFlagAffectedCommand implements LocalFlagAffectedCommand {
 
-   protected Set<Flag> flags;
+   private long flags = 0;
 
    @Override
-   public Set<Flag> getFlags() {
+   public long getFlagsBitSet() {
```
danberindei: Couldn't you keep the old method and allocate an `EnumSet` on demand?

core/src/main/java/org/infinispan/commands/CommandsFactory.java
```diff
@@ -91,39 +90,41 @@
     * @param key key to put
     * @param value value to put
     * @param metadata metadata of entry
-    * @param flags Command flags provided by cache
+    * @param flagsBitSet Command flags provided by cache
     * @return a PutKeyValueCommand
     */
-   PutKeyValueCommand buildPutKeyValueCommand(Object key, Object value, Metadata metadata, Set<Flag> flags);
+   PutKeyValueCommand buildPutKeyValueCommand(Object key, Object value, Metadata metadata, long flagsBitSet);
```
danberindei: Based on the package, I think `CommandsFactory` is public API too.

core/src/main/java/org/infinispan/commands/functional/ReadWriteManyCommand.java
```diff
@@ -180,11 +181,21 @@ public boolean alwaysReadsExistingValues() {
    }
 
    @Override
+   public long getFlagsBitSet() {
+      return EnumUtil.EMPTY_BIT_SET;
+   }
+
+   @Override
    public void setFlags(Set<Flag> flags) {
       // TODO: Customise this generated block
```
danberindei: Remove TODO?

core/src/main/java/org/infinispan/commands/LocalFlagAffectedCommand.java
```diff
@@ -17,14 +19,22 @@
     * not be modified directly, only via the {@link #setFlags(Set)}, {@link #addFlag(Flag)} and {@link
     * #addFlags(Set)} methods.
     */
-   Set<Flag> getFlags();
+   default Set<Flag> getFlags() {
+      return EnumUtil.enumSetOf(getFlagsBitSet(), Flag.class);
+   }
+
+   long getFlagsBitSet();
```
danberindei: If this is supposed to be backwards-compatible, then you need to have a default implementation for `getFlagsBitSet()` based on `getFlags()` now, and only switch them in 9.0.

core/src/main/java/org/infinispan/commands/CommandsFactory.java
pruivo: that's why I marked as a Preview a not for 8.2 :)

core/src/main/java/org/infinispan/commands/functional/ReadWriteManyCommand.java
pruivo: No idea why the TODOs are there, but there are a bunch of methods like this. Not sure if the class is completed or not.

core/src/main/java/org/infinispan/commands/AbstractLocalFlagAffectedCommand.java
pruivo: the getFlags is not removed. It is implemented in the interface.
Yes, it was the only way I had to copy the flags to another commands (e.g. cluster get to lock control)

core/src/main/java/org/infinispan/commands/read/GetCacheEntryCommand.java
```diff
@@ -25,9 +23,8 @@
    private InternalEntryFactory entryFactory;
    private InternalCacheEntry remotelyFetchedValue;
 
-   public GetCacheEntryCommand(Object key, Set<Flag> flags, InternalEntryFactory entryFactory) {
-      this.key = key;
-      this.flags = flags;
+   public GetCacheEntryCommand(Object key, long flagsBitSet, InternalEntryFactory entryFactory) {
```
danberindei: Do we need to keep the old constructor for backwards compatibility?

core/src/main/java/org/infinispan/context/Flag.java
```diff
@@ -272,64 +270,10 @@
...
 
    /**
-    * Creates a copy of a Flag Set removing instances of FAIL_SILENTLY.
-    * The copy might be the same instance if no change is required,
-    * and should be considered immutable.
-    * @param flags
-    * @return might return the same instance
+    * Creates a copy of a Flag BitSet removing instances of FAIL_SILENTLY.
     */
-   public static Set<Flag> copyWithoutRemotableFlags(Set<Flag> flags) {
```
danberindei: We should just mark the old methods as deprecated.

core/src/main/java/org/infinispan/commands/functional/ReadWriteManyCommand.java
slaskawi: +1 for killing todo :) Comments die last :)

tristantarrant:
Pushed, thanks

PR #3938, opened 2016-01-14
ISPN-6055 Performance regression in replicated reads after ISPN-5465

danberindei:
https://issues.jboss.org/browse/ISPN-6055

### DIFF
--- a/core/src/main/java/org/infinispan/distribution/group/PartitionerConsistentHash.java
+++ b/core/src/main/java/org/infinispan/distribution/group/PartitionerConsistentHash.java
@@ -3,6 +3,7 @@ package org.infinispan.distribution.group;
 import org.infinispan.commons.hash.Hash;
 import org.infinispan.distribution.ch.ConsistentHash;
 import org.infinispan.distribution.ch.KeyPartitioner;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/distribution/group/PartitionerConsistentHash.java
```diff
@@ -77,7 +78,12 @@ public String getRoutingTableAsString() {
 
    @Override
    public boolean isKeyLocalToNode(Address nodeAddress, Object key) {
-      return ch.isSegmentLocalToNode(nodeAddress, getSegment(key));
+      // TODO Make this a method in ConsistentHash
```
tristantarrant: Why not use a default method ?

danberindei: I was hesitating about what that method should be, and I thought I was short on time...

danberindei: I added an `isReplicated()` method in `ConsistentHash` and removed the TODO.

tristantarrant:
Pushed, thanks

PR #3941, opened 2016-01-19
ISPN-3550 Allow overriding the TimeService in child classes

jmarkos:
This is a small change, needed for the Cassandra cache store testsuite.


tristantarrant:
Pushed, thanks

PR #3944, opened 2016-01-20
ISPN-6085 Avoid using sun.misc.VM

Sanne:
https://issues.jboss.org/browse/ISPN-6085

### DIFF
--- a/core/src/main/java/org/infinispan/container/entries/PrimitiveEntrySizeCalculator.java
+++ b/core/src/main/java/org/infinispan/container/entries/PrimitiveEntrySizeCalculator.java
@@ -112,20 +112,15 @@ public class PrimitiveEntrySizeCalculator extends AbstractEntrySizeCalculatorHel
       static final int MIN_CACHE_VALUE = -128;
 
       static int calculateMaxIntCache() {
-         int h = 127;
-         String integerCacheHighPropValue =
-                 sun.misc.VM.getSavedProperty("java.lang.Integer.IntegerCache.high");
-         if (integerCacheHighPropValue != null) {
-            try {
-               int i = Integer.parseInt(integerCacheHighPropValue);
-               i = Math.max(i, 127);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/container/entries/PrimitiveEntrySizeCalculator.java
```diff
@@ -112,20 +112,15 @@ protected long handleObject(Object object) {
...
-               i = Math.max(i, 127);
-               // Maximum array size is Integer.MAX_VALUE
-               h = Math.min(i, Integer.MAX_VALUE - (-128) -1);
-            } catch( NumberFormatException nfe) {
-               // If the property cannot be parsed into an int, ignore it.
-            }
+         //We start from 128 as caching numbers up to 127 is required by the JLS:
+         //see 5.1.7 Boxing Conversion
+         for (int i = 128; i < Integer.MAX_VALUE; i++) {
+            if (Integer.valueOf(i) == Integer.valueOf(i))
```
danberindei: Since `Integer.valueOf(i)` is allowed to return the same instance every time, couldn't a future version of HotSpot decide this is equivalent to `true`?

Sanne: Yes it could, but only if it in fact returns the same instance every time, like with a large cache. And you'd want to know that, right? It would be considered "cached".

Sanne: Hi Dan, sorry during my late answer reply I probably misread your question.

I think it would be illegal for the compiler to "optimise" it to that extent, unless it could guarantee that the same instance is returned for each and every value of `i`.
Implementation wise in the JVM that would imply that all Integer instances are cached within a pool, so the function would still be correct by returning MAX_INTEGER.

tristantarrant:
Pushed to both master and 8.1.x

PR #3945, opened 2016-01-22
ISPN-6089 Distributed Stream doesn't operate in parllel locally

wburns:
* Set the stream to parallel right away if a parallel stream was invoked
* Set the stream parallel or sequential when retrieving

https://issues.jboss.org/browse/ISPN-6089


tristantarrant:
Pushed to both master and 8.1.x. Thanks

PR #3946, opened 2016-01-22
[master] ISPN-6049 Calls to reflection API fail with enabled security manager;…

rhusar:
… need to be run in privileged block

Jira
https://issues.jboss.org/browse/ISPN-6049


tristantarrant:
Pushed, thanks

PR #3947, opened 2016-01-25
Lucene 5.4.1 & Hibernate Search 5.5.2.Final

Sanne:
- https://issues.jboss.org/browse/ISPN-6091
 - https://issues.jboss.org/browse/ISPN-6092

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -67,7 +67,7 @@
       <version.gnu.getopt>1.0.13</version.gnu.getopt>
       <version.hibernate.javax.persistence>1.0.0.Final</version.hibernate.javax.persistence>
       <version.hibernate.core>5.0.4.Final</version.hibernate.core>
-      <version.hibernate.search>5.5.1.Final</version.hibernate.search>
+      <version.hibernate.search>5.5.2.Final</version.hibernate.search>
       <version.javax.cache>1.0.0</version.javax.cache>
       <version.cdi>1.0-SP4</version.cdi>
       <version.jboss.marshalling>1.4.10.Final</version.jboss.marshalling>
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -150,7 +150,7 @@
       <version.leveldb>0.7</version.leveldb>
       <version.leveldbjni>1.7</version.leveldbjni>
       <version.log4j>2.3</version.log4j>
-      <version.lucene>5.3.1</version.lucene>
+      <version.lucene>5.4.1</version.lucene>
       <version.mc4j>1.2.6</version.mc4j>
       <version.mockito>1.9.5</version.mockito>
       <version.mockito_dep.objenesis>1.4</version.mockito_dep.objenesis>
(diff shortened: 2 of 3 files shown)


tristantarrant:
Because the uberjars depend on the infinispan-bom directly without going through infinispan-parent, the lucene oveerride doesn't work and we end up with infinispan-embedded-query containing the version of lucene that HS depends on (currently 5.3.1).

Sanne:
hi @tristantarrant thanks for that, I didn't notice.

But what do you suggest I should do?

Personally I keep being more and more confused by the multiple layers of "parent"s and "bom"s. I wish there was just one parent we could use for dependency management.

I guess I could add the dependencymanagement section to the uberjar modules but it all feels like lots of duplication.

BTW we also have to set the Lucene version property for both server and embedded.. every time that feels like wrong (and each time I forget, which trips some waste of time)

tristantarrant:
The idea between the split is to keep the dependency management stuff clean. Also, the server has a different parent hierarchy, and that is why you need to update it in two places. Unless we split the boms into non-overlapping sub-boms and compose them as needed.

Sanne:
> The idea between the split is to keep the dependency management stuff clean.

I don't agree with the approach but I can live with it, as long as you can suggest me what I should do to fix this PR :)

> Also, the server has a different parent hierarchy, and that is why you need to update it in two places.
I realize it has a different hierarchy, but I don't understand why they should be different: seems to introduce some redundancy (and related quality issues), while I am missing to see the benefit for the split.

tristantarrant:
I was expecting a proposal from you :) Remember that our parent also has things like pluginManagement and default "executions" (like the component.dat and OSGi generators) which have no business in a downstream consumable pom. Maybe separate all that into an infinispan "build" pom ?

tristantarrant:
Sanne, the solution is to hoist the Lucene dependencies to -bom.

Sanne:
> Sanne, the solution is to hoist the Lucene dependencies to -bom.

That implies that users will be forced to use our Lucene version, even though they might not need Query.
Doesn't seem fair?

The clean solution is to have a single parent pom file which is an *actual parent* and not just a file with a confusing name. In this parent we'll list all versions in a <dependencyManagement> block, but no actual dependencies unless you're really sure you want to force them across any module (maybe suited for some maven plugins?).

A BOM file should be a user-friendly import, not something we rely on ourselves during the build.

gustavonalle:
Why the upgrade to Lucene 5.4.1 when Hibernate Search 5.5.2.Final depends on Lucene 5.3.1? I think we should keep the Lucene versions aligned. This also avoids the whole bom/pom issue as a side effect.

Sanne:
@gustavonalle Lucene 5.4 is backwards compatible with 5.3 but includes a couple of performance fixes which did affect us.

gustavonalle:
@Sanne Unless it's a significant performance increase, I'd still keep the versions aligned, in order to avoid a fertile ground for issue in the modules, osgi, packaging, build and so on. 
http://ci.infinispan.org/viewLog.html?buildId=35193&tab=buildResultsDiv&buildTypeId=bt9&logTab=

gustavonalle:
btw, if we are integrating hsearch master in the next hours, I guess this PR can be closed?

Sanne:
ok fair enough. Sad that it has to be complex though: I was hoping to get some more feedback on the performance differences, in some cases it doubles query throughput (it was iterating the document set twice).

PR #3949, opened 2016-01-25
ISPN-6095 TransportStackConfigurationIT fails with JGroups 3.6.7.Final

danberindei:
https://issues.jboss.org/browse/ISPN-6095


tristantarrant:
Pushed, thanks

PR #3950, opened 2016-01-25
ISPN-5815 deployed tasks

michalszynkiewicz:
# Usage
To introduce a Java Server Task, a developer has to create a jar file with a service implementing org.infinispan.tasks.ServerTask.
The task is accessible by its name - the result of `getName()` method.
Cache, parameters and marshaller are injected into the task via `setTaskContext()` method.
Task developer can restrict task usages for users with a specific role only. In order to do it, `getAllowedRole()` has to implemented to return the role.

# Implementation details
During deployment, ServerTaskProcessor scans deployed modules for services implementing ServerTask interface.
ServerTasks are then registered in ServerTaskRegistry.

ServerTaskEngine is the element responsible for running server tasks. It is injected into TaskManager.
Based on ServerTask#getExecutionMode, task engine uses either LocalServerTaskRunner or DistributedServerTaskRunner to run the task.

JIRA task: https://issues.jboss.org/browse/ISPN-5815


core/src/main/java/org/infinispan/tasks/DeployedTask.java
```diff
@@ -0,0 +1,11 @@
+package org.infinispan.tasks;
+
+/**
+ * Author: Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ * Date: 1/19/16
+ * Time: 2:18 PM
+ */
+public interface DeployedTask<T> {
```
slaskawi: @tristantarrant Does this interface look correct? Do we need anything else?

parent/pom.xml
```diff
@@ -173,7 +173,7 @@
       <version.weld-se>2.2.0.SP1</version.weld-se>
 
       <version.org.wildfly>10.0.0.CR5</version.org.wildfly>
-      <version.org.wildfly.core>2.0.5.Final</version.org.wildfly.core>
+      <version.org.wildfly.core>2.0.7.Final</version.org.wildfly.core>  <!-- mstodo undo ? -->
```
slaskawi: Do we need this upgrade?

server/integration/infinispan/pom.xml
```diff
@@ -122,6 +122,11 @@
             <artifactId>metainf-services</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-tasks</artifactId>
+            <version>${version.org.infinispan}</version>  <!--   mstodo remove commented out or uncomment-->
```
slaskawi: Is it needed? I think tasks are in core right?

server/integration/infinispan/pom.xml
```diff
@@ -138,6 +143,35 @@
...
             </plugin>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>serialize_component_metadata</id>
+                        <phase>process-classes</phase>
+                        <goals>
+                            <goal>java</goal>
```
slaskawi: I think it's also unnecessary

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemAdd.java
```diff
@@ -85,11 +87,13 @@ protected void execute(DeploymentProcessorTarget processorTarget) {
             processorTarget.addDeploymentProcessor(INFINISPAN_SUBSYSTEM_NAME, Phase.POST_MODULE, ++basePriority, new CacheLoaderExtensionProcessor());
             processorTarget.addDeploymentProcessor(INFINISPAN_SUBSYSTEM_NAME, Phase.POST_MODULE, ++basePriority, new CacheWriterExtensionProcessor());
             processorTarget.addDeploymentProcessor(INFINISPAN_SUBSYSTEM_NAME, Phase.POST_MODULE, ++basePriority, new ExternalStoreExtensionProcessor());
+            processorTarget.addDeploymentProcessor(INFINISPAN_SUBSYSTEM_NAME, Phase.POST_MODULE, ++basePriority, new DeployableTaskProcessor());
```
slaskawi: Probably there should be another constant for priority here.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/DeployableTaskProcessor.java
```diff
@@ -0,0 +1,98 @@
...
+   public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
+      DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
+      Module module = deploymentUnit.getAttachment(Attachments.MODULE);
+      ServicesAttachment servicesAttachment = deploymentUnit.getAttachment(Attachments.SERVICES);
+      if (module != null && servicesAttachment != null)
+         addServices(phaseContext, servicesAttachment, module);
+   }
+
+   private void addServices(DeploymentPhaseContext ctx, ServicesAttachment servicesAttachment, Module module) {
+      List<String> implementationNames = servicesAttachment.getServiceImplementations(org.infinispan.tasks.DeployedTask.class.getName());
```
slaskawi: Import might be used here

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/DeployableTaskProcessor.java
```diff
@@ -0,0 +1,98 @@
...
+      ModuleClassLoader classLoader = module.getClassLoader();
+      for (String serviceClassName : implementationNames) {
+         try {
+            Class<? extends DeployedTask> clazz = classLoader.loadClass(serviceClassName).asSubclass(org.infinispan.tasks.DeployedTask.class);
```
slaskawi: and here

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/DeployableTaskProcessor.java
```diff
@@ -0,0 +1,98 @@
...
+      public InjectedValue<DeployedTaskRegistry> getDeployedTaskManager() {
+         return deployedTaskRegistry;
+      }
+
+      private String constructName() {
+         return extension.getName();
+      }
+
+      @Override
+      public DeployedTask getValue() throws IllegalStateException, IllegalArgumentException {
```
slaskawi: This doesn't throw those exceptions

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/DeployedTaskEngine.java
```diff
@@ -0,0 +1,43 @@
+package org.jboss.as.clustering.infinispan.task;
```
slaskawi: I'm not sure if this is the best place for Task Engine. Maybe @tristantarrant has some thoughts...

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/DeployedTaskManager.java
```diff
@@ -0,0 +1,11 @@
+//package org.jboss.as.clustering.infinispan.task;
+//
```
slaskawi: I see those are missing bits, which will be implemented soon

server/integration/testsuite/src/test/resources/config/infinispan/customtask-deployment-structure.xml
```diff
@@ -0,0 +1,8 @@
+<jboss-deployment-structure>
+    <deployment>
+        <dependencies>
```
slaskawi: Can we remove that?

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/InfinispanIntegrationMetadataFileFinder.java
```diff
@@ -0,0 +1,25 @@
+package org.jboss.as.clustering.infinispan.task;
```
slaskawi: What does it do? I'm not sure if I understand this part...

slaskawi:
All bits are falling into places. Nice work... needs a bit cleaning though and I would like to hear @tristantarrant opinion on `DeployedTaskEngine` and `DeployedTask` implementation.

parent/pom.xml
tristantarrant: Please let's stay aligned with whatever is used by the main WildFly dependency

server/integration/infinispan/pom.xml
tristantarrant: No, tasks are in a separate module

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/InfinispanIntegrationMetadataFileFinder.java
michalszynkiewicz: Without it the component metadata from *.dat file is not imported (I've just checked it)

parent/pom.xml
michalszynkiewicz: We don't need it. I will undo the change - I used it for debugging locally (with the old one I had two versions of libs)

server/integration/infinispan/pom.xml
michalszynkiewicz: If it were a sub-module of infinispan parent, simply adding 
   <properties>
         <module.skipComponentMetaDataProcessing>false</module.skipComponentMetaDataProcessing>
   </properties>
would do the trick.

Unfortunately, the "parent hierarchy" for this module is:
infinispan-server-versions <- jboss-parent

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/DeployedTaskManager.java
michalszynkiewicz: Candidates for removal

server/integration/testsuite/src/test/resources/config/infinispan/customtask-deployment-structure.xml
michalszynkiewicz: yes :)

core/src/main/java/org/infinispan/security/AuthorizationManager.java
```diff
@@ -27,4 +28,11 @@
     * has the requested permission and role. A {@link SecurityException} is thrown otherwise.
     */
    void checkPermission(AuthorizationPermission permission, String role);
+
+   /**
+    * Verifies that the {@link Subject} associated with the current {@link AccessControlContext}
+    * has the requested permission and at least one of roles.
+    * A {@link SecurityException} is thrown otherwise.
+    */
+   void checkPermission(AuthorizationPermission perm, Collection<String> roles);
```
tristantarrant: Maybe name the method to clearly indicate that it is just going to need at least one matching role, because otherwise it looks like it implies it needs all roles

michalszynkiewicz: Would checkForPermissionAndRoleAmong be a better name? ?

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemAdd.java
michalszynkiewicz: why? Can you write a bit more about it?

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/DeployableTaskProcessor.java
michalszynkiewicz: Done

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/DeployableTaskProcessor.java
michalszynkiewicz: Done

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/task/DeployableTaskProcessor.java
michalszynkiewicz: Done

core/src/main/java/org/infinispan/security/AuthorizationManager.java
tristantarrant: If we could use mathematical notation:
checkPermission∃(...) 
:)
Leave it as it is now. As long as the JavaDoc is clear.

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
```diff
@@ -61,11 +59,16 @@ public void checkPermission(AuthorizationPermission perm, String role) {
...
 
    public void checkPermission(AuthorizationConfiguration configuration, AuthorizationPermission perm) {
-      checkPermission(configuration, perm, null);
+      checkPermission(configuration, perm, (String)null);
    }
 
    public void checkPermission(AuthorizationConfiguration configuration, AuthorizationPermission perm,
-         String role) {
+                               String role) {
+      checkPermission(configuration, perm, Collections.singletonList(role));
```
tristantarrant: Unfortunately this constructs a new collection for every invocation even when we don't need to check for a role (which is the majority of cases in the code). In the above invocations it would be better to substitute (String)null with Collections.emptyList().

michalszynkiewicz: what would you say for changing this method to:

public void checkPermission(AuthorizationConfiguration configuration, AuthorizationPermission perm,
                               String role) {
      List<String> roles = role == null ? Collections.emptyList() : Collections.singletonList(role);
      checkPermission(configuration, perm, roles);
   }

tristantarrant: No need to assign to a local variable, just place the ternary evaluation in the method call

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
```diff
@@ -110,12 +113,24 @@ private boolean checkSubjectPermissionAndRole(Subject subject, AuthorizationConf
...
       } else {
          return false;
       }
    }
 
+   private boolean checkSubjectRole(SubjectACL subjectACL, Collection<String> requestedRoles) {
+      if (requestedRoles == null || requestedRoles.isEmpty()) {
+         return true;
+      }
+      for (String role : requestedRoles) {
```
tristantarrant: We really need to add support for role sorting at some point to speed up this check

michalszynkiewicz: I think it would only be slow if task was available for many roles.
subjectACL.containsRole is just a check in a hashmap.

Am I missing something?

server/integration/infinispan/pom.xml
slaskawi: Ok understood

tasks-api/src/main/java/org/infinispan/tasks/ServerTask.java
```diff
@@ -0,0 +1,26 @@
+package org.infinispan.tasks;
+
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.concurrent.Callable;
+
+/**
+ * Author: Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
```
slaskawi: Since this is a part of public API, I would suggest adding some meaningful Javadoc here (with a description how it works etc).

server/integration/testsuite/src/test/resources/config/infinispan/customtask.xml
```diff
@@ -0,0 +1,18 @@
+<subsystem xmlns="urn:infinispan:server:core:8.0" >
+    <cache-container
+            name="local"
+            default-cache="taskAccessible">
+        <local-cache
+                name="taskAccessible"
+                start="EAGER"
+                batching="false">
+            <locking
```
slaskawi: locking is probably unnecessary.

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/DistributedServerTask.java
```diff
@@ -0,0 +1,55 @@
...
+
+   public DistributedServerTask(String taskName, Map<String, ?> parameters) {
+      this.taskName = taskName;
+      this.parameters = parameters;
+   }
+
+   @Override
+   public void setEnvironment(Cache<Object, Object> cache, Set<Object> inputKeys) {
+      this.cache = cache;
+      // todo script-related code wraps a lot of things in security util calls. is this necessary here?
```
michalszynkiewicz: are these changes ok security-wise?

slaskawi: This sounds like a question for @tristantarrant

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/DistributedServerTask.java
```diff
@@ -0,0 +1,55 @@
...
+   public T call() throws Exception {
+      ServerTaskRegistry taskRegistry = componentRegistry.getComponent(ServerTaskRegistry.class);
+
+      ServerTaskWrapper<T> task = taskRegistry.getTask(taskName);
+      task.inject(prepareContext());
+      return task.run();
+   }
+
+   private TaskContext prepareContext() {
+      TaskContext context = new TaskContext();
```
slaskawi: You can use `Optional.ofNullable(...).ifPresent(...) or flatMap` constructs here.

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/DistributedServerTask.java
```diff
@@ -0,0 +1,55 @@
...
+ * Time: 1:49 PM
+ */
+public class DistributedServerTask<T> implements Serializable, DistributedCallable<Object, Object, T> {
+   private final Map<String, ?> parameters;
+   private final String taskName;
+
+   private transient GlobalComponentRegistry componentRegistry;
+   private transient Cache<Object, Object> cache;
+
+   public DistributedServerTask(String taskName, Map<String, ?> parameters) {
```
slaskawi: Just thinking... Maybe this should accept Optional of map? This would simplify this: https://github.com/infinispan/infinispan/pull/3950/files#diff-e136f12dbae088bc0d3481f744bb9c97R26

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/ServerTaskRegistry.java
```diff
@@ -0,0 +1,23 @@
...
+import org.infinispan.tasks.Task;
+
+import java.util.List;
+
+/**
+ * Author: Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ * Date: 1/22/16
+ * Time: 4:03 PM
+ */
+public interface ServerTaskRegistry {
```
slaskawi: Some meaningful javadoc here would be nice.

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/ServerTaskRegistryImpl.java
```diff
@@ -0,0 +1,57 @@
...
+@Scope(Scopes.GLOBAL)
+public class ServerTaskRegistryImpl implements ServerTaskRegistry {
+
+   @Inject
+   public void init(TaskManager taskManager, EmbeddedCacheManager cacheManager) {
+      ServerTaskEngine engine = new ServerTaskEngine(this, cacheManager);
+      ((TaskManagerImpl) taskManager).registerTaskEngine(engine);
+   }
+
+   private Map<String, ServerTaskWrapper> tasks = new HashMap<>();
```
slaskawi: What will happen if we add (and get) multiple tasks at the same time? This might happen when will have multiple deployments.

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/ServerTaskRegistryImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+   @Override
+   public List<Task> getTasks() {
+      Collection<ServerTaskWrapper> tasks = this.tasks.values();
+      return new ArrayList<>(tasks);
+   }
+
+   @Override
+   public <T> ServerTaskWrapper<T> getTask(String taskName) {
+      //noinspection unchecked
```
slaskawi: Probably an IDE comment.

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/ServerTaskRunner.java
```diff
@@ -0,0 +1,14 @@
...
+import org.infinispan.tasks.TaskContext;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Author: Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ * Date: 1/28/16
+ * Time: 9:33 AM
+ */
+public interface ServerTaskRunner {
```
slaskawi: Some javadoc might be handy.

slaskawi:
Some minor stuff found but apart from that - nice job!

@tristantarrant - any comments?

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/ServerTaskRegistryImpl.java
michalszynkiewicz: Do you mean multiple threads run in parallel?

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/DistributedServerTask.java
michalszynkiewicz: done

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/ServerTaskRegistry.java
michalszynkiewicz: done

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/ServerTaskRegistryImpl.java
michalszynkiewicz: changed to @SuppressWarnings on method level

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/ServerTaskRunner.java
michalszynkiewicz: done

core/src/main/java/org/infinispan/security/impl/AuthorizationHelper.java
tristantarrant: Let's simplify things and not bother with multiple roles: they just introduce too much complexity and affect performance negatively in the simple case

tasks-api/src/main/java/org/infinispan/tasks/ServerTask.java
```diff
@@ -0,0 +1,61 @@
...
+
+   /**
+    * Gives a set of roles, for which the task is accessible.
+    * If the task executor has at least one of the roles in the set, the task will be executed.
+    *
+    * If the collection is empty - all users can executed the task
+    *
+    * @return a set of user roles, for which the task can be executed
+    */
+   default Set<String> getAllowedRoles() {
```
tristantarrant: Convert this method's signature to Optional<String> getAllowedRole()

michalszynkiewicz: done

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/InfinispanIntegrationMetadataFileFinder.java
```diff
@@ -0,0 +1,19 @@
...
+ * Author: Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ * Date: 1/22/16
+ * Time: 2:21 PM
+ */
+@MetaInfServices
+public class InfinispanIntegrationMetadataFileFinder implements ModuleMetadataFileFinder {
+
+   @Override
+   public String getMetadataFilename() {
+      return "infinispan-server-infinispan-component-metadata.dat";
```
tristantarrant: the name of the dat file should reflect that of the artifactId.

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/DistributedServerTask.java
```diff
@@ -0,0 +1,56 @@
...
+   @Override
+   public void setEnvironment(Cache<Object, Object> cache, Set<Object> inputKeys) {
+      this.cache = cache;
+      // todo script-related code wraps a lot of things in security util calls. is this necessary here?
+      componentRegistry = cache.getCacheManager().getGlobalComponentRegistry();
+   }
+
+   @Override
+   public T call() throws Exception {
+      ServerTaskRegistry taskRegistry = componentRegistry.getComponent(ServerTaskRegistry.class);
```
tristantarrant: I think the ServerTaskRegistry should not be obtained here, but it should be injected by the runner during construction.
Also, a cache is optional paramater in the TaskContext and you should not rely on its presence here. Because of a current limitation in the DefaultExecutorService, it is not possible to have a distexec without a distributed cache, but @wburns is adding a new implementation which changes this.

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/InfinispanIntegrationMetadataFileFinder.java
michalszynkiewicz: this is ${basedir}/src/main/resources/${project.artifactId}-component-metadata.dat.

The artifactId for server/integration/infinispan is: infinispan-server-infinispan

tristantarrant: Sorry, ignore my comment. I thought this was in another module.

slaskawi:
Changes Required label was probably added by a mistake...

tristantarrant:
Pushed. Please open a Jira for the new dist stream work. Thanks a lot for working on this.

PR #3951, opened 2016-01-25
ISPN-5414 Support (named) parameters in DSL based queries - extend tests

mcimbora:
- Adding more tests to cover passing parameters by map, verify exception handling
-- TODO verify whether params with null/empty name can be passed (see testNullParamName() testEmptyParamName())
- Check whether reusing the same query with different parameters works (basic scenario, scenario using CQ & client events)

Suggestions about other areas to be covered are welcome, @anistor ?.


anistor:
Null and empty param name are not legal. But they might be (wrongly) accepted at build time and fail at run time. Not sure, will check.

Coverage is good.

I see some errors in CI in org.infinispan.client.hotrod.query due to wrong expected message id (http://ci.infinispan.org/viewLog.html?buildId=34452&buildTypeId=bt9&guest=1)

anistor:
I've added here the checks for null and empty and your tests https://github.com/infinispan/infinispan/pull/3963

anistor:
Closing this as the commits are also in https://github.com/infinispan/infinispan/pull/3963

PR #3955, opened 2016-01-26
ISPN-6105 Too many warnings logged for total order commands

danberindei:
https://issues.jboss.org/browse/ISPN-6105


tristantarrant:
Pushed, thanks.

PR #3957, opened 2016-01-26
ISPN-6100 L1 entries not always stored as L1InternalCacheEntry

danberindei:
https://issues.jboss.org/browse/ISPN-6100

DistributedClusteringLogic.commitSingleEntry wasn't using L1Metadata
for entries that should have been stored as L1 entries.


tristantarrant:
Pushed, thanks

PR #3958, opened 2016-01-26
ISPN-6101 Deprecate M/R and point to Distributed Streams

wburns:
https://issues.jboss.org/browse/ISPN-6101


tristantarrant:
Pushed, thanks

PR #3959, opened 2016-01-27
ISPN-6106 FilterConditionEndContext.between(from, to) should check ar…

anistor:
…gs are Comparable

Jira: https://issues.jboss.org/browse/ISPN-6106

### DIFF
--- a/query-dsl/src/main/java/org/infinispan/query/dsl/impl/ValueRange.java
+++ b/query-dsl/src/main/java/org/infinispan/query/dsl/impl/ValueRange.java
@@ -1,5 +1,8 @@
 package org.infinispan.query.dsl.impl;
 
+import org.infinispan.query.dsl.impl.logging.Log;
+import org.jboss.logging.Logger;
+
 /**
  * Represents a range of values starting from value {@code from} and ending at {@code to}, including or excluding
  * the interval ends as indicated by {@code includeLower} and {@code includeUpper} respectively. This is used to
@@ -10,6 +13,8 @@ package org.infinispan.query.dsl.impl;
  */
 class ValueRange {
 
+   private static final Log log = Logger.getMessageLogger(Log.class, ValueRange.class.getName());
+
    private final Object from;
 
    private final Object to;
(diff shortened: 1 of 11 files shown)


query-dsl/src/main/java/org/infinispan/query/dsl/impl/ValueRange.java
```diff
@@ -19,6 +19,12 @@
    private boolean includeUpper = true;
 
    public ValueRange(Object from, Object to) {
+      if (!(from instanceof Comparable)) {
+         throw new IllegalArgumentException("'from' must be an instance of java.lang.Comparable");
```
tristantarrant: Should go to Log.

query-dsl/src/main/java/org/infinispan/query/dsl/impl/ValueRange.java
```diff
@@ -19,6 +19,12 @@
...
+      }
+      if (!(to instanceof Comparable)) {
+         throw new IllegalArgumentException("'to' must be an instance of java.lang.Comparable");
```
tristantarrant: Should go to Log.

query-dsl/src/main/java/org/infinispan/query/dsl/impl/ValueRange.java
anistor: Yes, ideally, but there is no Log in this module yet. I'll add one and see what breaks.

tristantarrant: You can use the query module's log

anistor: query-dsl does not depend on query, so it can be used in remote mode too. we need a Log in query-dsl.

tristantarrant: no need to allocate a full 1000 range. You can carve out a chunk from query's allocation

anistor: just did that.

anistor:
I've started to use jboss-logging for the exceptions introduced in this PR. The already existing occurences (about 30 something) will be addressed separately in https://issues.jboss.org/browse/ISPN-6116/

tristantarrant:
Pushed, thanks

PR #3960, opened 2016-01-27
ISPN-6113 Make Consistent Hashes persistable to the global state

tristantarrant:
https://issues.jboss.org/browse/ISPN-6113

### DIFF
--- a/core/src/main/java/org/infinispan/distribution/ch/ConsistentHash.java
+++ b/core/src/main/java/org/infinispan/distribution/ch/ConsistentHash.java
@@ -1,7 +1,7 @@
 package org.infinispan.distribution.ch;
 
 import org.infinispan.commons.hash.Hash;
-import org.infinispan.distribution.ch.impl.HashFunctionPartitioner;
+import org.infinispan.globalstate.ScopedPersistentState;
 import org.infinispan.remoting.transport.Address;
 
 import java.util.Collection;
@@ -51,6 +51,7 @@ public interface ConsistentHash {
     * @deprecated Since 8.2, the {@code Hash} is optional - replaced in the configuration by the
     * {@code KeyPartitioner}
     */
+   @Deprecated
    default Hash getHashFunction() {
       throw new UnsupportedOperationException();
    }
@@ -172,4 +173,13 @@ public interface ConsistentHash {
     * Returns a string containing all the segments and their associated addresses.
     */
    String getRoutingTableAsString();
+
+   /**
+     * Writes this ConsistentHash to the specified scoped state.
+     *
+     * @param state the state to which this ConsistentHash will be written
+     */
+   default void toScopedState(ScopedPersistentState state) {
+      throw new UnsupportedOperationException();
+   }
 }
--- a/core/src/main/java/org/infinispan/distribution/ch/impl/DefaultConsistentHash.java
+++ b/core/src/main/java/org/infinispan/distribution/ch/impl/DefaultConsistentHash.java
@@ -6,8 +6,11 @@ import org.infinispan.commons.marshall.InstanceReusingAdvancedExternalizer;
 import org.infinispan.commons.util.Immutables;
 import org.infinispan.commons.util.Util;
 import org.infinispan.distribution.ch.ConsistentHash;
+import org.infinispan.globalstate.ScopedPersistentState;
 import org.infinispan.marshall.core.Ids;
 import org.infinispan.remoting.transport.Address;
+import org.infinispan.remoting.transport.jgroups.JGroupsAddressCache;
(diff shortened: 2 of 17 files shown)


core/src/main/java/org/infinispan/distribution/ch/ConsistentHash.java
```diff
@@ -41,6 +41,11 @@
  */
 public interface ConsistentHash {
 
+   public static final String STATE_CONSISTENT_HASH = "consistentHash";
+   public static final String STATE_HASH_FUNCTION = "hashFunction";
+   public static final String STATE_MEMBER = "member.%d";
+   public static final String STATE_MEMBERS = "members";
```
danberindei: Should these really be in the interface? Must implementers use these particular keys?

core/src/main/java/org/infinispan/distribution/ch/impl/DefaultConsistentHash.java
```diff
@@ -371,6 +414,32 @@ public String prettyPrintOwnership() {
...
    }
 
+   @Override
+   public void toScopedState(ScopedPersistentState state) {
+      state.setProperty(STATE_CONSISTENT_HASH, this.getClass().getName());
+      state.setProperty(STATE_NUM_OWNERS, numOwners);
+      state.setProperty(STATE_NUM_SEGMENTS, getNumSegments());;
+      state.setProperty(STATE_MEMBERS, members.size());
+      for (int i = 0; i < members.size(); i++) {
+         state.setProperty(String.format(STATE_MEMBER, i),
```
danberindei: Is the state expected to be empty? If the state already contains a CH's state, and the new CH has less members, we'll leave some extra members hanging around.

tristantarrant: The state will be recently created, i.e. there is no risk of pollution.

core/src/main/java/org/infinispan/distribution/ch/ConsistentHash.java
tristantarrant: I'll move them to the implementation

tristantarrant: Actually, since those are common to different CH's it's best if they are moved to a dedicated "constants" class.

tristantarrant:
@danberindei I've moved the constants.

core/src/main/java/org/infinispan/distribution/ch/ConsistentHash.java
```diff
@@ -172,4 +173,14 @@ default boolean isReplicated() {
     * Returns a string containing all the segments and their associated addresses.
     */
    String getRoutingTableAsString();
+
+   /**
+     * Writes this ConsistentHash to the specified scoped state.
+     *
+     * @param state the state to which this ConsistentHash will be written
+     */
+   /*default void toScopedState(ScopedPersistentState state) {
```
galderz: Delete?

core/src/main/java/org/infinispan/globalstate/ScopedPersistentState.java
```diff
@@ -27,6 +27,26 @@
...
+
+   /**
+    * Sets a float state property.
+    */
+   void setProperty(String format, float f);
+
+   /**
+    * Retrieves an integer state property
+    */
+   int getIntProperty(String key);
```
galderz: Not sure you need these. You could have just defined:

```
<T> T getProperty(String key);
```

And that'd cover all types...

Regardless of the API, the possibility of class cast exceptions remains with a system-property-like approach :|

galderz:
Are there any integration tests for this? The tests added are all unit tests which run without a Cache/CacheManager

tristantarrant:
No integration tests yet, in the way you intend them, since that will happen once I send the PR that actually performs the graceful start and will restore the CH into a cache.

core/src/main/java/org/infinispan/distribution/ch/ConsistentHash.java
tristantarrant: I actually intended for the default method to be in the actual PR, as we're changing an interface in a minor release and I want to preserve backwards compatibility.

core/src/main/java/org/infinispan/globalstate/ScopedPersistentState.java
tristantarrant: If I implemented it that way there would be no way for me to detect T within the method and therefore I could not perform the appropriate string -> type conversion.

tristantarrant:
Fixed the interface.

danberindei:
Integrated, thanks Tristan!

PR #3962, opened 2016-01-28
ISPN-6107 State transfer should not try to fetch lost segments

danberindei:
https://issues.jboss.org/browse/ISPN-6107

Do not try to transfer segments that were added in a topology update,
after the rebalance was started - these segments will not have a real
source anyway. And fetching new segments is not safe, if we already sent
the rebalance confirmation.

This should speed up {{ConcurrentJoinTest}}, which takes > 3 mins to shut down an 8-node cluster in the worst case.

### DIFF
--- a/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
+++ b/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
@@ -269,7 +269,16 @@ public class StateConsumerImpl implements StateConsumer {
          ownsData = false;
       }
 
-      if (isRebalance) {
+      // If a member leaves/crashes immediately after a rebalance was started, the new CH_UPDATE
+      // command may be executed before the REBALANCE_START command, so it has to start the rebalance.
+      boolean startRebalance = isRebalance;
+      if (!isRebalance) {
+         if (cacheTopology.getPendingCH() != null && this.cacheTopology.getPendingCH() == null) {
+            if (trace) log.tracef("Forcing startRebalance = true");
+            startRebalance = true;
+         }
+      }
+      if (startRebalance) {
          // Only update the rebalance topology id when starting the rebalance, as we're going to ignore any state
          // response with a smaller topology id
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -277,12 +286,13 @@ public void onTopologyUpdate(final CacheTopology cacheTopology, final boolean is
...
       }
 
-      awaitTotalOrderTransactions(cacheTopology, isRebalance);
+      awaitTotalOrderTransactions(cacheTopology, startRebalance);
 
       // Make sure we don't send a REBALANCE_CONFIRM command before we've added all the transfer tasks
       // even if some of the tasks are removed and re-added
       waitingForState.set(false);
 
+      final ConsistentHash newReadCh = cacheTopology.getReadConsistentHash();
```
pruivo: not used

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -768,6 +790,12 @@ private void requestTransactions(Set<Integer> segments, Map<Address, Set<Integer
                log.failedToRetrieveTransactionsForSegments(segments, cacheName, source, e);
                // The primary owner is still in the cluster, so we can't exclude it - see ISPN-4091
                failed = true;
+               // FIXME Temporary hack to limit the number of retries when the source has a newer topology
```
pruivo: hmmm...
I don't understand what the sleep will solve.

pruivo:
@danberindei just some minor comments.

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
danberindei: Oops, it was meant to make the failure easier to reproduce, it wasn't supposed to make it in the PR :)

danberindei:
@pruivo Updated

pruivo:
testing and pushing...

pruivo:

PR #3977, opened 2016-02-01
ISPN-5995 ParseUtils.unexpectedElement/unxpectedAttribute called but …

anistor:
…the exception is not thrown

https://issues.jboss.org/browse/ISPN-5995


tristantarrant:
Applied to both master and 8.1.x, thanks

PR #3978, opened 2016-02-01
ISPN-6051 ClusterExpirationManager missing locking

wburns:
* Added entry locking so when an entry is updated we don't
* expire a previous value

https://issues.jboss.org/browse/ISPN-6051


wburns:
Actually I need to think about this more.  I still think this is possible to get a deadlock, I need to write a test tomorrow for it.

wburns:
Actually this is fine.

galderz:
Any tests to verify this?

wburns:
It fixes the ExpiryTest mentioned in the JIRA

tristantarrant:
Pushed, thanks

PR #3979, opened 2016-02-02
ISPN-6146 XA transactions already committed in prepare as XA_RDONLY f…

rhusar:
…ail on commit with "ISPN000112: exception while committing: javax.transaction.xa.XAException"

* we maintain a status for each resource
* resources that were prepared with status XA_RDONLY
* are skipped on commit

Jira
https://issues.jboss.org/browse/ISPN-6146

Please merge to _8.1.2.Final_ as well.

### DIFF
--- a/core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
+++ b/core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
@@ -14,10 +14,13 @@ import javax.transaction.Transaction;
 import javax.transaction.xa.XAException;
 import javax.transaction.xa.XAResource;
 import javax.transaction.xa.Xid;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 import static java.lang.String.format;
 
@@ -45,7 +48,7 @@ public class DummyTransaction implements Transaction {
    private final Xid xid;
    private volatile int status = Status.STATUS_UNKNOWN;
    private final List<Synchronization> syncs;
-   private final List<XAResource> resources;
+   private final List<Map.Entry<XAResource, Integer>> resources;
    private RollbackException firstRollbackException;
 
    public DummyTransaction(DummyBaseTransactionManager tm) {
@@ -180,9 +183,9 @@ public class DummyTransaction implements Transaction {
       checkStatusBeforeRegister("resource");
 
       //avoid duplicates
-      for (XAResource otherResource : resources) {
+      for (Map.Entry<XAResource, Integer> otherResourceEntry : resources) {
          try {
-            if (otherResource.isSameRM(resource)) {
+            if (otherResourceEntry.getKey().isSameRM(resource)) {
                log.debug("Ignoring resource. It is already there.");
                return true;
             }
@@ -191,7 +194,7 @@ public class DummyTransaction implements Transaction {
          }
       }
 
-      resources.add(resource);
+      resources.add(new AbstractMap.SimpleEntry<>(resource, null));
 
       try {
          if (trace) {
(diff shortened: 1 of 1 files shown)


rhusar:
Please backport (or I can do it) to 8.1.x too if ever gets merged, thank!

core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
```diff
@@ -45,7 +47,7 @@
    private final Xid xid;
    private volatile int status = Status.STATUS_UNKNOWN;
    private final List<Synchronization> syncs;
-   private final List<XAResource> resources;
+   private final Map<XAResource, Integer> resources;
```
pruivo: I would prefer if you created another ```List``` with the prepared resources. 
I'm not sure if the method ```isSameRM()``` and ```equals()``` has to be equivalent and it can happen the ```Map``` to replace a ```XaResource``` because of ```equals()``` returns ```true```.

rhusar: BTW the 1st commit uses 2 collections, the second commit is optional to use only 1 collection.

rhusar: Hm, I am using java.util.Map#replace(K, V) so those resources cannot be replaced, the most that can happen that the badly-behaved TM would still commit twice..

pruivo: I'm a little worried about this line:
https://github.com/infinispan/infinispan/pull/3979/files#diff-9bce40d7971a6c1e61f18ac552eac0c3R196
It is possible the put to replace an existing XaResource.

rhusar: Ok, redoing this to keep using lists.

rhusar: Done, ready for review.

pruivo:
integrated! thanks @rhusar !

core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
```diff
@@ -191,7 +194,7 @@ public boolean enlistResource(XAResource resource) throws RollbackException, Ill
          }
       }
 
-      resources.add(resource);
+      resources.add(new AbstractMap.SimpleEntry<>(resource, null));
```
danberindei: What's wrong with `put(resource, null)`?

core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
```diff
@@ -247,7 +250,7 @@ public void registerSynchronization(Synchronization sync) throws RollbackExcepti
    }
 
    public Collection<XAResource> getEnlistedResources() {
-      return Collections.unmodifiableList(resources);
+      return Collections.unmodifiableList(resources.stream().map(e -> e.getKey()).collect(Collectors.toList()));
```
danberindei: `return Collections.unmodifiableCollection(resources.keys())`?

core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
rhusar: The object in question is java.util.List.

core/src/main/java/org/infinispan/transaction/tm/DummyTransaction.java
rhusar: The object in question is java.util.List.

danberindei: I see... but `getEnlistedResources()` is called all the time, e.g. in `runPrepare()`, so I'd rather not create a new list every time. I think `LinkedHashMap` might be a good compromise, if you need the resources to be kept in order. Or maybe we should just stop using `getEnlistedResources()` internally and have `endResources()` iterate on the list directly...

PR #3981, opened 2016-02-03
ISPN-5367 Cache putMap/getAll usage of SegmentAwareMap/Set performance

wburns:
investigation

* Use standard map to prevent iteration over all entries multiple times

https://issues.jboss.org/browse/ISPN-5367


wburns:
In my testing using https://gist.github.com/mcimbora/5c404d98a13b38bbecc4 and running master and the new version I saw about 10% performance increase when using 3-5 nodes.

tristantarrant:
Pushed, thanks

PR #3982, opened 2016-02-04
ISPN-6125 JMX attribute evictions is always zero

mgencur:
https://issues.jboss.org/browse/ISPN-6125

I also had another solution which would add new MBean called Eviction, similar to already existing Passivation and Activation. But we would end up with two MBeans for eviction: the already existing which still partially works and the new one). I think the current solution is less invasive and just fixes behaviour of the "evictions" attribute on Statistics MBean. So far it was registering only evictions performed explicitly by cache.evict(). Now it counts all evictions.

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/CacheMgmtInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/CacheMgmtInterceptor.java
@@ -82,7 +82,6 @@ public class CacheMgmtInterceptor extends JmxStatsCommandInterceptor {
       Object returnValue = invokeNextInterceptor(ctx, command);
       if (getStatisticsEnabled(command))
          evictions.increment();
-
       return returnValue;
    }
 
@@ -440,5 +439,8 @@ public class CacheMgmtInterceptor extends JmxStatsCommandInterceptor {
       return super.getStatisticsEnabled() && !cmd.hasFlag(Flag.SKIP_STATISTICS);
    }
 
+   public void addEvictions(long numEvictions) {
+      evictions.add(numEvictions);
+   }
 }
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/infinispan/interceptors/CacheMgmtInterceptor.java
```diff
@@ -48,7 +48,7 @@
    private final LongAdder hits = new LongAdder();
    private final LongAdder misses = new LongAdder();
    private final LongAdder stores = new LongAdder();
-   private final LongAdder evictions = new LongAdder();
+   private final AtomicLong evictions = new AtomicLong(0);
```
wburns: Is there a reason for changing this?  LongAdder tends to perform better under concurrent access.

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
```diff
@@ -350,6 +356,11 @@ public void evict(K key) {
       @Override
       public void onEntryEviction(Map<K, InternalCacheEntry<K, V>> evicted) {
          evictionManager.onEntryEviction(evicted);
+         List<CommandInterceptor> interceptors = interceptorChain.getInterceptorsWhichExtend(CacheMgmtInterceptor.class);
```
wburns: This logic seems like a better fit in the EvictionManager instead of in the data container.

wburns: Also we shouldn't do this call if jmx stats were disabled to begin with - `configuration.jmxStatistics().available()` returns false.

wburns:
I agree using the existing one is a better choice.  Just a few comments.

Thank, Martin!

mgencur:
Yeah. I overlooked that LongAdder has the "add" method too. Returned back and addressed the other comments too.

wburns:
Seems fine to me.  I am not a big fan of having to get the interceptor each time, but I am not sure of a better way.  Checking...

wburns:
Integrated, thanks @mgencur !

PR #3986, opened 2016-02-05
ISPN-6185 Server ops for task management & ISPN-6188 Server ops for script management

tristantarrant:
https://issues.jboss.org/browse/ISPN-6185
https://issues.jboss.org/browse/ISPN-6188


tristantarrant:
@vblagoje fixed the LocalDescriptions. They worked because there are some "wildcards" in the code which picked up other defs, but it is better to define them fully like this.

vblagoje:
Integrated with fcca3c7 and 7d0aa92 Thanks @tristantarrant

PR #3987, opened 2016-02-08
ISPN-6174 Avoid calling the server after iteration is done in the rem…

gustavonalle:
…ote iterator

https://issues.jboss.org/browse/ISPN-6174


tristantarrant:
Pushed, thanks

PR #3988, opened 2016-02-08
ISPN-6047 Deadlock when a prepare command is retried

pruivo:
https://issues.jboss.org/browse/ISPN-6047

### DIFF
--- a/core/src/main/java/org/infinispan/remoting/inboundhandler/action/PendingTxAction.java
+++ b/core/src/main/java/org/infinispan/remoting/inboundhandler/action/PendingTxAction.java
@@ -70,17 +70,11 @@ public class PendingTxAction extends BaseLockingAction implements PendingLockLis
          keysToLock.removeAll(context.getLockedKeys());
       }
 
-      if (keysToLock.isEmpty()) {
-         //nothing to do. nobody else was able to update the state from checking, so no need to check the CAS
-         cas(InternalState.CHECKING, InternalState.READY);
-         return ActionStatus.READY;
-      }
-
       PendingLockPromise promise = keysToLock.size() == 1 ?
             pendingLockManager.checkPendingTransactionsForKey(context, keysToLock.get(0), timeout, TimeUnit.MILLISECONDS) :
             pendingLockManager.checkPendingTransactionsForKeys(context, keysToLock, timeout, TimeUnit.MILLISECONDS);
 
-      if (promise == PendingLockPromise.NO_OP) {
+      if (promise.isReady()) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultPendingLockManager.java
```diff
@@ -308,18 +351,67 @@ private PendingTransaction(CacheTransaction cacheTransaction, Object key, Comple
...
+   private class BasePendingLockPromise implements PendingLockPromise, Callable<Void> {
+
+      protected final long expectedEndTime;
+
+      private BasePendingLockPromise(long expectedEndTime) {
+         this.expectedEndTime = expectedEndTime;
+      }
+
+      @Override
+      public Void call() throws Exception {
```
danberindei: Could you add a javadoc to say where it's used?

core/src/main/java/org/infinispan/remoting/inboundhandler/action/PendingTxAction.java
```diff
@@ -70,17 +70,11 @@ protected ActionStatus init(ActionState state) {
          keysToLock.removeAll(context.getLockedKeys());
       }
 
-      if (keysToLock.isEmpty()) {
-         //nothing to do. nobody else was able to update the state from checking, so no need to check the CAS
-         cas(InternalState.CHECKING, InternalState.READY);
-         return ActionStatus.READY;
-      }
```
danberindei: Why do we need to create a promise in this case?

core/src/test/java/org/infinispan/statetransfer/TxReplay3Test.java
```diff
@@ -0,0 +1,204 @@
...
+            try {
+               sequencer.advance(TX2_PENDING);
+            } catch (TimeoutException | InterruptedException e) {
+               throw new CacheException(e);
+            }
+         }
+      }
+   }
+
+   private static class TestConsistentHashFactory extends BaseControlledConsistentHashFactory {
```
danberindei: This class seems equivalent to `new ControlledConsistentHashFactory(0)`.

core/src/main/java/org/infinispan/remoting/inboundhandler/action/PendingTxAction.java
pruivo: yes to avoid checking for pending again when it reaches the interceptors. however, I've made some changes to avoid creating a promise and I'm storing the NO_OP instance.

danberindei:
Integrated, thanks Pedro!

PR #3989, opened 2016-02-08
ISPN-6191 Provide Externalizers for (Double|Long|Int)SummaryStatistics

wburns:
* Use reflection to read and write to fields

https://issues.jboss.org/browse/ISPN-6191


tristantarrant:
Pushed to both master and 8.1.x

PR #3991, opened 2016-02-09
[8.1.x] ISPN-6192 Memory leak caused by RPCs that complete prior to their timeout

pferraro:
https://issues.jboss.org/browse/ISPN-6192

Re-instrument singleton pattern using enum.


wburns:
Pulling..

wburns:
Integrated into 8.1.x, thanks @pferraro !

PR #3992, opened 2016-02-09
[master] ISPN-6192 Memory leak caused by RPCs that complete prior to their timeout

pferraro:
https://issues.jboss.org/browse/ISPN-6192

Re-instrument singleton pattern using enum.


wburns:
Seems fine to me, pulling.

wburns:
Integrated into master, thanks @pferraro !

PR #3993, opened 2016-02-09
[8.1.x] ISPN-6146 XA transactions already committed in prepare as XA_RDONLY f…

rhusar:
…ail on commit with "ISPN000112: exception while committing: javax.transaction.xa.XAException"

Master PR
https://github.com/infinispan/infinispan/pull/3979

Jira
https://issues.jboss.org/browse/ISPN-6146


rhusar:
This will apply cleanly, so no need for a separate PR.

PR #3995, opened 2016-02-10
fix Cluster Listener document.

kazuhira-r:
That the Cluster Listener supports CacheEntryExpired event was reflected in the document.


galderz:
Integrated, thanks!

PR #3997, opened 2016-02-11
ISPN-6199 Persist roles to the container authorization configuration only if present

tristantarrant:
https://issues.jboss.org/browse/ISPN-6199


pruivo:
integrated! thanks @tristantarrant !

PR #4000, opened 2016-02-11
ISPN-6195 CacheNotFoundResponse error when starting multiple Infinisp…

pruivo:
…an Server instances in parallel

https://issues.jboss.org/browse/ISPN-6195

this should do the trick. Also, removed unused private method and did some cleanups.


tristantarrant:
Pushed, thanks

pferraro:
@tristantarrant @pruivo Please cherry-pick this into 8.1.x.

galderz:
Thanks for the fix @pruivo :)

PR #4001, opened 2016-02-11
ISPN-6200 Fix remote iterators in compatibility mode

vjuranek:
https://issues.jboss.org/browse/ISPN-6200


tristantarrant:
Pushed, thanks

PR #3819, opened 2015-11-10
ISPN-5467 Design sequential interceptor interfaces

danberindei:
https://issues.jboss.org/browse/ISPN-5467

* SequentialInterceptor is a potentially async interceptor
* DDSequentialInterceptor extends SequentialInterceptor with
  double-dispatch support (similar to CommandInterceptor)
* SequentialInvocationContext is the new invocation context interface,
  containing support methods for the sequential interceptors
* SequentialInvocationChain is the new invocation chain interface
* SequentialInvocationChain has a new implementation, the other interfaces
  are implemented by the existing interceptor and context classes
* Log the global transaction/command invocation id as part of the MDC

PR #4002, opened 2016-02-12
ISPN-5467 Deprecate existing interceptors

danberindei:
The ISPN-5467 pull request is not quite ready, performance-wise.

This pull request only deprecates the interceptors, to alert users that the implementation may change at any time. We will still allow custom interceptors after ISPN-5467/ISPN-5468, but users depending on an existing interceptor (e.g. for positioning their custom interceptor) may need to change their code.


danberindei:
https://github.com/infinispan/infinispan/pull/3819 is now only 15% slower than master with the existing interceptors extending `CommandInterceptor` (for local mode reads - the difference is much smaller in clustered modes). So I think it would actually be ok to merge that for Beta2, and hopefully we'll get those 15% back before the Final release.

tristantarrant:
Pushed, thanks

PR #4005, opened 2016-02-15
[8.1.x] ISPN-6195 CacheNotFoundResponse error when starting multiple Infinisp…

pruivo:
…an Server instances in parallel

https://issues.jboss.org/browse/ISPN-6195

Backport to 8.1.x


tristantarrant:
Pushed

PR #4007, opened 2016-02-15
ISPN-6207 Cross-Site Replication marshalling issues

pruivo:
https://issues.jboss.org/browse/ISPN-6207

### DIFF
--- a/core/src/main/java/org/infinispan/xsite/XSiteAdminCommand.java
+++ b/core/src/main/java/org/infinispan/xsite/XSiteAdminCommand.java
@@ -104,8 +104,8 @@ public class XSiteAdminCommand extends BaseRpcCommand {
             return;
          case AMEND_TAKE_OFFLINE:
             output.writeUTF(siteName);
-            output.writeInt(afterFailures);
-            output.writeLong(minTimeToWait);
+            output.writeObject(afterFailures);
+            output.writeObject(minTimeToWait);
             return;
          case STATUS:
             return;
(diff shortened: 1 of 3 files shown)


danberindei:
@pruivo Does this help with the hung tests, http://paste.fedoraproject.org/322843/14555444/raw/ ?

core/src/main/java/org/infinispan/xsite/XSiteAdminCommand.java
```diff
@@ -104,8 +104,8 @@ public void writeTo(ObjectOutput output) throws IOException {
             return;
          case AMEND_TAKE_OFFLINE:
             output.writeUTF(siteName);
-            output.writeInt(afterFailures);
-            output.writeLong(minTimeToWait);
+            output.writeObject(afterFailures);
+            output.writeObject(minTimeToWait);
```
danberindei: I'd rather remove the `switch` and write all the fields every time.

pruivo:
@danberindei I haven't managed to reproduce the failures, but I found these issues in the meanwhile :)

core/src/main/java/org/infinispan/xsite/XSiteAdminCommand.java
pruivo: I don't see why. why waste memory with fields that aren't going to be used? you would have to allocate larger buffers and use more bandwidth...

danberindei: @pruivo because there are very few `XSiteAdminCommand` instances ever serialized, so the memory savings are negligible? The risk of someone requiring an existing field for a particular command type and forgetting to modify the switch is non-zero, and I don't think the memory savings are worth the pain to debug such a scenario.

danberindei:
Integrated, thanks Pedro!

PR #4013, opened 2016-02-16
Upgrade search components

Sanne:
A preview for @gustavonalle and @anistor 

Includes:
 - the fixes for statistics handling on Query's MBean
 - Hibernate Search upgrade
 - get Infinispan Query to compile with the latest Hibernate Search
 - Lucene Upgrade

### DIFF
--- a/query/src/main/java/org/infinispan/query/impl/InfinispanQueryStatisticsInfo.java
+++ b/query/src/main/java/org/infinispan/query/impl/InfinispanQueryStatisticsInfo.java
@@ -4,12 +4,9 @@ import org.hibernate.search.spi.SearchIntegrator;
 
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.atomic.LongAdder;
 
 /**
- * This MBean accumulates query statistics from the Hibernate Search statistics object. The only difference is this
- * statistics continue to accumulate and are not reset when the Search Factory is reconfigured. They are still reset if
- * {@code clear()} is explicitly called though.
+ * This MBean exposes the query statistics from the Hibernate Search statistics object.
  *
  * @author anistor@redhat.com
  * @since 6.1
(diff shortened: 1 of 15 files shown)


gustavonalle:
:+1: 
I'm also running the tests in my environment

gustavonalle:
I am seeing one consistent failure in:
```
 StoredIndexTest.testRestartingNode:45->assertEmailsFound:69 expected:<1> but was:<0>
```

gustavonalle:
Apart from that, the rest looks fine

query/src/main/java/org/infinispan/query/impl/InfinispanQueryStatisticsInfo.java
```diff
@@ -17,13 +17,6 @@
 public class InfinispanQueryStatisticsInfo implements InfinispanQueryStatisticsInfoMBean {
```
anistor: This class and InfinispanQueryStatisticsInfoMBean were introduced as a workaround for HSEARCH-1461. I believe they need to be removed.

Sanne: I think we can keep them. A good place to add Infinispan-specific metrics, and hopefully also clustered metrics and things related to hybrid queries?

anistor: In that case please remove the unused imports, also make getObjectLoadingExecutionAvgTime just delegate to sf.getStatistics().getObjectLoadingExecutionAvgTime() instead of actually doing the computation.
Also fix the javadoc, which no longer reflects the purpose of these classes.  Or just remove them entirely? We can add them when needed :)

anistor: getSearchQueryExecutionAvgTime should also be just delegated instead of being computed.

Sanne: ok!

anistor: There is another workaround here: https://github.com/Sanne/infinispan/blob/UpgradeSearchComponents/query/src/main/java/org/infinispan/query/backend/SearchFactoryHandler.java#L88-90

Do we still need it?

Sanne:
thanks @anistor @gustavonalle incorporated your fixes and suggestions.

I think only the actual Search version is missing ?

query/src/main/java/org/infinispan/query/impl/InfinispanQueryStatisticsInfo.java
Sanne: fixed, thanks!

anistor:
Indeed. LGTM too.

Sanne:
Pushed with the correct versions included.

 - https://issues.jboss.org/browse/ISPN-6091
 - https://issues.jboss.org/browse/ISPN-6092
 - https://issues.jboss.org/browse/ISPN-6216
 - https://issues.jboss.org/browse/ISPN-3782

gustavonalle:
pulling

gustavonalle:
merged, thanks @Sanne !

PR #4018, opened 2016-02-17
ISPN-6211 Create DMR operations to enable / disable rebalancing per c…

pruivo:
…ache and container

https://issues.jboss.org/browse/ISPN-6211


tristantarrant:
Pushed, thanks Pedro

PR #4019, opened 2016-02-17
Publishing distribution zips to the repo

michalszynkiewicz:


tristantarrant:
We don't want the distribution zip deployed to the maven repository as that would be quite wasteful.

michalszynkiewicz:
It revealed it's not necessary. Sorry.

PR #4011, opened 2016-02-16
ISPN-6214 Initial cluster size

tristantarrant:
https://issues.jboss.org/browse/ISPN-6214

Also provides a fix for 

https://issues.jboss.org/browse/ISPN-6099

PR #4020, opened 2016-02-18
ISPN-6214 Fix DistributedExecutorBadResponseFailoverTest

danberindei:
The JGroupsTransport is already started by the time we wrap it
in our CacheNotFoundResponseTransport, so we should not start it
a second time.

Caused by https://github.com/infinispan/infinispan/pull/4011


tristantarrant:
Pushed, thanks

PR #4022, opened 2016-02-19
ISPN-6246 - Create tests for Remote Script Execution

andyuk1986:
https://issues.jboss.org/browse/ISPN-6246

Added more tests increasing the code coverage as well as testing more cases of execution scripts and tasks.
Updated the Jacoco plugin and corresponding profiles for calculating the code coverage.


client/hotrod-client/src/test/java/org/infinispan/client/hotrod/AuthenticationTest.java
```diff
@@ -40,7 +40,7 @@ protected EmbeddedCacheManager createCacheManager() throws Exception {
       return cacheManager;
    }
 
-   private ConfigurationBuilder initServerAndClient() {
+   protected ConfigurationBuilder initServerAndClient() {
```
slaskawi: This method doesn't seem to be used in new tests - do we need to change it to protected?

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExecTest.java
```diff
@@ -41,6 +50,11 @@ protected void createCacheManagers() throws Throwable {
       createHotRodServers(NUM_SERVERS, new ConfigurationBuilder());
    }
 
+   @Test(expectedExceptions = HotRodClientException.class, expectedExceptionsMessageRegExp = ".*ISPN027002.*")
```
slaskawi: I would prefer not to use the numbers here. They might change over time (or in the product) and this test might start failing. Can we use some specific phrase in the exception?

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExecTest.java
```diff
@@ -103,6 +117,64 @@ public void testRemoteMapReduce() throws Exception {
...
+      String cacheName = "testRemoteMapReduce_Streams";
+      ConfigurationBuilder builder = getDefaultClusteredCacheConfig(CacheMode.DIST_SYNC, true);
+      builder.dataContainer().keyEquivalence(new AnyServerEquivalence()).valueEquivalence(new AnyServerEquivalence()).compatibility().enable().marshaller(new GenericJBossMarshaller());
+      defineInAll(cacheName, builder);
+      RemoteCache<String, String> cache = clients.get(0).getCache(cacheName);
+      RemoteCache<String, String> scriptCache = clients.get(0).getCache(SCRIPT_CACHE);
+      loadData(cache, "/macbeth.txt");
+      loadScript(scriptCache, "/wordCountStream.js");
+
+      HashMap<String, Long> results = cache.execute("wordCountStream.js", new HashMap<String, String>());
```
slaskawi: Could you please use `Map` interface here (and not the implementation)?

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExecTest.java
```diff
@@ -103,6 +117,64 @@ public void testRemoteMapReduce() throws Exception {
...
+              .compatibility().enable().marshaller(new GenericJBossMarshaller());
+      defineInAll(cacheName, builder);
+      waitForClusterToForm(cacheName);
+
+      RemoteCache<String, String> cache = clients.get(0).getCache(cacheName);
+      RemoteCache<String, String> scriptCache = clients.get(1).getCache(SCRIPT_CACHE);
+      loadData(cache, "/macbeth.txt");
+      loadScript(scriptCache, "/wordCountStream_dist.js");
+
+      ArrayList<HashMap<String, Long>> results = cache.execute("wordCountStream_dist.js", new HashMap<String, String>());
```
slaskawi: The same here... 

There are more places like this in this commit. Could you please update them also?

parent/pom.xml
```diff
@@ -1948,156 +1945,165 @@
...
-
-                                 </structure>
-
-
-                                 <html destdir="${dir.report}" />
-                              </report>
+                              <echo>Deleting existing Jacoco Report Directory.</echo>
+                              <delete dir="${dir.jacoco.report}"/>
+                              <delete dir="${dir.jacoco.merged}"/>
+                              <delete dir="${project.build.directory}/classes" />
```
slaskawi: I think this line should be removed right? We don't want to delete classes from build directory while wiping out Jacoco stuff.

scripting/src/test/java/org/infinispan/scripting/ClusteredScriptingTest.java
```diff
@@ -58,6 +52,14 @@ public void testDistExecScript() throws IOException, InterruptedException, Execu
       assertTrue(results.contains(manager(1).getAddress()));
    }
 
+   @Test(expectedExceptions = IllegalStateException.class, expectedExceptionsMessageRegExp = "^ISPN026009.*")
```
slaskawi: Could you please update this one as well? Having identifiers here might be fragile.

scripting/src/test/java/org/infinispan/scripting/ClusteredScriptingTest.java
```diff
@@ -86,26 +99,12 @@ public void testMapReduce() throws Exception {
...
-         String script = TestingUtil.loadFileAsString(is);
-         scriptingManager.addScript(fileName.replaceAll("\\/", ""), script);
-      }
+   @Test(expectedExceptions = IllegalArgumentException.class, expectedExceptionsMessageRegExp = "^ISPN026008.*")
+   public void testOnlyReduceTaskRun() throws Exception {
+      ScriptingManager scriptingManager = getScriptingManager(manager(0));
+      Cache<String, String> cache = cache(0);
+      loadData(cache, "/macbeth.txt");
+      loadScript(scriptingManager, "/wordCountReducer.js");
+      CompletableFuture<Object> future = scriptingManager.runScript("wordCountReducer.js", new TaskContext().cache(cache));
```
slaskawi: Could you tell me which line throws `IllegalArgumentException`? Maybe we need do something with this `CompletableFuture` here?

scripting/src/test/java/org/infinispan/scripting/ScriptMetadataTest.java
```diff
@@ -17,6 +18,14 @@ public void testDoubleSlashComment() throws Exception {
       assertEquals("test", metadata.name());
    }
 
+   public void testDefaultScriptExtension() throws Exception {
+      ScriptMetadata metadata = ScriptMetadataParser.parse("test", "// name=test");
+      assertEquals("test", metadata.name());
+
+      metadata = ScriptMetadataParser.parse("test.", "/* name=exampleName */");
```
slaskawi: This test could be divided into 2 separate test cases.

scripting/src/test/java/org/infinispan/scripting/ScriptingTest.java
```diff
@@ -1,51 +1,205 @@
...
+   }
 
    public void testSimpleScript() throws Exception {
-      loadScript(scriptingManager, "test.js");
       String result = (String) scriptingManager.runScript("test.js", new TaskContext().addParameter("a", "a")).get();
       assertEquals("a", result);
    }
 
-   @Test(expectedExceptions = CacheException.class, expectedExceptionsMessageRegExp = "ISPN026005:.*")
+   @Test(priority = 2, expectedExceptions = CacheException.class, expectedExceptionsMessageRegExp = "ISPN026005:.*")
```
slaskawi: It seems this test is tightly coupled with `testScriptingCacheClear` (because of the priority). Can we get rid of this dependency somehow?

scripting/src/test/java/org/infinispan/scripting/ScriptingTest.java
```diff
@@ -1,51 +1,205 @@
...
+   }
+
+   public void testSimpleScript1() throws Exception {
+      String value = "javaValue";
+      String key = "processValue";
+
+      cacheManager.getCache("test_cache").put(key, value);
+
+      CompletableFuture exec = scriptingManager.runScript("testExecWithoutProp.js");
+      while (!exec.isDone()) {
```
slaskawi: You can call `exec.get()` here (with optional timeout if you wish).

scripting/src/test/java/org/infinispan/scripting/ScriptingTest.java
```diff
@@ -1,51 +1,205 @@
...
+         Thread.sleep(1000);
+      }
+
+      assertEquals(value + ":additionFromJavascript", cacheManager.getCache("test_cache").get(key));
+   }
+
+   public void testScriptCallFromJavascript() throws Exception {
+      cacheManager.getCache().remove("a");
```
slaskawi: Do we need those 2 lines here?

tasks/src/test/java/org/infinispan/tasks/SecurityTaskManagerTest.java
```diff
@@ -0,0 +1,127 @@
...
+                    e.printStackTrace();
+                }
+
+                taskEngine.setSlowTask(new CompletableFuture<>());
+                return null;
+            }
+        });
+    }
+
+    /*public void testTaskExecutionWithoutUser() throws Exception { //<-- Throwing nullpointer if no principle is provided.
```
slaskawi: This could probably be removed

tasks/src/test/java/org/infinispan/tasks/SecurityTaskManagerTest.java
```diff
@@ -0,0 +1,127 @@
...
+                TaskExecution currentTask = currentTasks.iterator().next();
+                assertEquals(DummyTaskEngine.DummyTaskTypes.SLOW_TASK.name(), currentTask.getName());
+                assertEquals(principal, currentTask.getWho().get());
+                taskEngine.getSlowTask().complete("slow");
+
+                assertEquals(0, taskManager.getCurrentTasks().size());
+                try {
+                    assertEquals("slow", slowTask.get());
+                } catch (InterruptedException | ExecutionException e) {
+                    e.printStackTrace();
```
slaskawi: Probably `fail()` would be better here.

parent/pom.xml
```diff
@@ -1948,156 +1945,165 @@
...
                </plugin>
             </plugins>
          </build>
       </profile>
-
       <profile>
          <id>jacocoReport</id>
 
+         <activation>
+            <activeByDefault>false</activeByDefault>
```
slaskawi: This one is probably not needed.

slaskawi:
Great work @andyuk1986! There are some small things which require corrections but in general it looks good. The Jacoco part also looks very interesting!

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/AuthenticationTest.java
andyuk1986: Sorry - had a new class test extending this one, which was not included in the commit.  Added.

parent/pom.xml
andyuk1986: No, it is needed. There are 2 profiles for coverage. First one is running tests and collects the coverage files. The second one processes them. For multi-module project I couldn't implement one profile solution.

parent/pom.xml
andyuk1986: This line removes the outdated classes/ folder from infinispan/parent folder. Originally it doesn't contain classes/ folder. I am adding manually the classes from all modules there for proper jacoco code coverage report generation.

scripting/src/test/java/org/infinispan/scripting/ClusteredScriptingTest.java
andyuk1986: The IllegalArgumentException is thrown at  org.infinispan.scripting.impl.NullRunner:24 on purpose for not running reducer or combiner mode directly.

andyuk1986:
Thanks a lot @slaskawi . I have fixed all the changes that you have mentioned as well as replaced all exception IDs from expected exception messages with messages itself. 
I have replied to a few of your comments as well.

parent/pom.xml
slaskawi: Ok that makes sense.

scripting/src/test/java/org/infinispan/scripting/ClusteredScriptingTest.java
slaskawi: Right... but what do we want to do with `CompletableFuture<Object> future` reference? It has been assigned but it looks like it hasn't been used.

parent/pom.xml
slaskawi: The flag `<activeByDefault>` is actually set to false by default [1]. So why do we need to set it up here explicitly? 

[1] https://maven.apache.org/ref/3.3.9/maven-settings/settings.html

scripting/src/test/java/org/infinispan/scripting/ClusteredScriptingTest.java
andyuk1986: Ok, Got it. Fixed.

parent/pom.xml
andyuk1986: I have missed that. Thanks a lot. Fixed.

andyuk1986:
Fixed all comments. Thanks again.

slaskawi:
Integrated thanks!

PR #4027, opened 2016-02-22
ISPN-6243 Query tests involving topology change always fail

gustavonalle:
https://issues.jboss.org/browse/ISPN-6243


pruivo:
integrated! thanks @gustavonalle !

PR #4028, opened 2016-02-22
ISPN-6178 Add ability to define indexable classes via config

anistor:
jira: https://issues.jboss.org/browse/ISPN-6178


core/src/main/java/org/infinispan/configuration/cache/IndexingConfiguration.java
```diff
@@ -3,18 +3,29 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
 /**
  * Configures indexing of entries in the cache for searching.
  */
 public class IndexingConfiguration extends AbstractTypedPropertiesConfiguration {
    public static final AttributeDefinition<Index> INDEX = AttributeDefinition.builder("index", Index.NONE).immutable().build();
    public static final AttributeDefinition<Boolean> AUTO_CONFIG = AttributeDefinition.builder("autoConfig", false).immutable().build();
+   public static final AttributeDefinition<Set<Class<?>>> INDEXED_CLASSES = AttributeDefinition.builder("indexedClasses", null, (Class<Set<Class<?>>>) (Class<?>) Set.class)
```
gustavonalle: What about ```indexableEntity``` or ```indexedEntity```  instead of ```indexedClasses```?

gustavonalle:
As it goes with every config change PR, "don't forget about the server"  :wink:

anistor:
This is not ready yet. No server or xml config yet.

core/src/main/java/org/infinispan/configuration/cache/IndexingConfiguration.java
anistor: indexedEntity is good too.

gustavonalle:
@anistor Looks good, but apparently there is no test for the changes in the server's infinispan-subsystem. Minimally we could add the new attributes to https://github.com/infinispan/infinispan/blob/master/server/integration/infinispan/src/test/resources/org/jboss/as/clustering/infinispan/subsystem/subsystem-infinispan_8_2.xml?

anistor:
I was creating a test for that but had other failures to take care of ...

core/src/main/java/org/infinispan/configuration/parsing/Parser80.java
```diff
@@ -2311,35 +2311,61 @@ private void parseIndexing(final XMLExtendedStreamReader reader, final Configura
...
          }
       }
-      Properties indexingProperties = parseProperties(reader);
+
+      Properties indexingProperties = new Properties();
+      while (reader.hasNext() && (reader.nextTag() != XMLStreamConstants.END_ELEMENT)) {
+         Element element = Element.forName(reader.getLocalName());
+         switch (element) {
+            case INDEXED_ENTITY: {
+               if (reader.getAttributeCount() > 0) {
```
tristantarrant: ParseUtils.requireNoAttributes(reader)

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
```diff
@@ -470,7 +471,7 @@ public EmbeddedCacheManager startCaches(final String... cacheNames) {
       authzHelper.checkPermission(AuthorizationPermission.LIFECYCLE);
       List<Thread> threads = new ArrayList<Thread>(cacheNames.length);
       final AtomicReference<RuntimeException> exception = new AtomicReference<RuntimeException>(null);
-      for (final String cacheName : cacheNames) {
+      for (final String cacheName : new HashSet<String>(Arrays.asList(cacheNames))) {
```
tristantarrant: Why are you doing this ?

core/src/main/java/org/infinispan/configuration/parsing/Parser80.java
```diff
@@ -2311,35 +2311,61 @@ private void parseIndexing(final XMLExtendedStreamReader reader, final Configura
                throw ParseUtils.unexpectedAttribute(reader, i);
          }
       }
-      Properties indexingProperties = parseProperties(reader);
+
+      Properties indexingProperties = new Properties();
+      while (reader.hasNext() && (reader.nextTag() != XMLStreamConstants.END_ELEMENT)) {
+         Element element = Element.forName(reader.getLocalName());
+         switch (element) {
+            case INDEXED_ENTITY: {
```
tristantarrant: You are going to hate me here, but I'd rather have a <indexed-entities><indexed-entity>*</indexed-entities> form...

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
anistor: Pico-optimisation by uniquefying the cache names. InfinispanDirectoryProvider may be configured so that its caches (data/metdata/locking) are the same, so it will invoke cacheManager.startCaches with repeating cache names. A thread will be spawned for each occurrence. This does not lead to any errors but the extra threads just sit there waiting for the first to finish creating the cache.

Should probably extract that in a different commit.

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
```diff
@@ -496,8 +497,9 @@ public void run() {
          throw new CacheException("Interrupted while waiting for the caches to start");
       }
 
-      if (exception.get() != null) {
-         throw exception.get();
+      RuntimeException runtimeException = exception.get();
```
anistor: Intellij inspections keep telling me (wrongly) that the exception is likely lost and not rethrown, so I fixed it.

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
tristantarrant: The creation of a collection just to initialize a set is a bit pointless. Maybe initialize an empty set, and do `if (set.add(cacheName)) { /*startCache*/ }`

core/src/main/java/org/infinispan/configuration/parsing/Parser80.java
anistor: I hated removing this handy type autodetection too, and resisted for years, but the Rome meeting convinced me there are greater benefits.

ATM by enabling indexing and not declaring any indexed entities you still get the old autodetection behaviour. If you declare at least one type autodetection is disabled and the list of types is locked. Putting undeclared entities in cache logs warnings while in autodetect mode and adds them to the list. And is completely ignored if autodetect is off. Everything would still be queryable anyway.

core/src/main/java/org/infinispan/configuration/parsing/Parser80.java
anistor: That's what I was looking for! thanks

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
anistor: That would still be a HashSet, and all elements would end up in it eventually. We save just the creation of the Arrays.asList wrapper (that's a non copy collection btw) and an Iterator. Not good enough. I'd rather replace  ```List<Thread>``` with ```HashMap<String, Thread>```. The final bill is still cheaper, because HashSet is actually a HashMap on the inside anyway, and we get rid of the ArrayList.

anistor:
Updated.

core/src/main/java/org/infinispan/configuration/parsing/Parser80.java
anistor: Oh, I did not realise you may be actually asking for a parent ```<indexed-entities>``` element to hold them all. Is that what you intended?

tristantarrant: Yes :)

tristantarrant: I'd also like to promote some of the more common hsearch props to a real schema at some point...

anistor: Ok, I'll add the ```<indexed-entities>``` element. 

I thought by ```*``` you wanted to indicate type autodetection...

gustavonalle:
@anistor I'm wondering if the server related changes should go in a separate JIRA, since AFAICT, they won't work: if one declare the classes in the the config and start the server, it will fail simply because the classes will no be there. Probably some sort of deployment would be needed akin to filters and marshallers.

tristantarrant:
The server should definitely parse and ignore these for now.

anistor:
Added the ```indexed-entities``` wrapper element and commented out a section of code in server so the indexed entities are parsed but are actually ignored.

anistor:
I occasionally see NPEs in RemoteQueryIspnDirIT during clear, due to https://issues.jboss.org/browse/ISPN-6304, but otherwise the test results seem ok. Just started another run in CI to confirm.

tristantarrant:
pulling

tristantarrant:
Pushed, thanks Adrian

PR #4029, opened 2016-02-22
ISPN-6234 Remote Iterator resumes from wrong segments after failover

gustavonalle:
https://issues.jboss.org/browse/ISPN-6234

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/iteration/RemoteCloseableIterator.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/iteration/RemoteCloseableIterator.java
@@ -102,26 +102,16 @@ public class RemoteCloseableIterator<E> implements CloseableIterator<Entry<Objec
 
       } catch (TransportException e) {
          log.warnf(e, "Error reaching the server during iteration");
-         restartIteration(segmentKeyTracker.missedSegments());
+         startInternal(segmentKeyTracker.missedSegments());
          fetch();
       }
    }
 
-   private void restartIteration(Set<Integer> missedSegments) {
-      startInternal(missedSegments);
-   }
-
-   private void start(Set<Integer> fromSegments) {
-      IterationStartResponse startResponse = startInternal(fromSegments);
-
-      this.segmentKeyTracker = KeyTrackerFactory.create(startResponse.getSegmentConsistentHash(), startResponse.getTopologyId());
-   }
-
-   private IterationStartResponse startInternal(Set<Integer> fromSegments) {
+   private IterationStartResponse startInternal(Set<Integer> segments) {
       if (log.isDebugEnabled()) {
-         log.debugf("Staring iteration with segments %s", fromSegments);
+         log.debugf("Starting iteration with segments %s", segments);
       }
-      IterationStartOperation iterationStartOperation = operationsFactory.newIterationStartOperation(filterConverterFactory, filterParams, fromSegments, batchSize, metadata);
+      IterationStartOperation iterationStartOperation = operationsFactory.newIterationStartOperation(filterConverterFactory, filterParams, segments, batchSize, metadata);
       IterationStartResponse startResponse = iterationStartOperation.execute();
       this.transport = startResponse.getTransport();
       if (log.isDebugEnabled()) {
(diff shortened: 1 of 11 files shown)


gustavonalle:
Updated with minor refactoring

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/iteration/RemoteCloseableIterator.java
```diff
@@ -114,7 +114,7 @@ private void restartIteration(Set<Integer> missedSegments) {
    private void start(Set<Integer> fromSegments) {
```
wburns: I was a bit confused by this fromSegments.  It seems it is only ever passed ``segments``.  In this case it might be a bit more clear to remove this method and just have the ``start()`` method use this passing in ``segments`` for everything.  Or should this method be invoked elsewhere?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/iteration/SegmentKeyTracker.java
```diff
@@ -17,20 +18,21 @@
...
 
-   public SegmentKeyTracker(SegmentConsistentHash segmentConsistentHash) {
+   public SegmentKeyTracker(SegmentConsistentHash segmentConsistentHash, Set<Integer> segments) {
       int numSegments = segmentConsistentHash.getNumSegments();
       keysPerSegment = new AtomicReferenceArray<>(numSegments);
       if (log.isDebugEnabled()) log.debugf("Created SegmentKeyTracker with %d segments", numSegments);
       this.segmentConsistentHash = segmentConsistentHash;
-      IntStream.range(0, segmentConsistentHash.getNumSegments())
-              .forEach(i -> keysPerSegment.set(i, CollectionFactory.makeSet(ByteArrayEquivalence.INSTANCE)));
+      Stream<Integer> segmentStream = segments == null ?
```
wburns: I am thinking it would be better to have this an IntStream instead.

``Stream<Integer> case``
segments is null we have to box N values and unbox N values in the forEach consumer
segments is not null we have to unbox N values in the forEach consumer

``IntStream case``
segments is non null we have 0 unboxing
segment is not null we have 1 unboxing in a call to ``mapToInt``

In either case it should be better.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/IterationNextOperation.java
```diff
@@ -90,9 +90,8 @@ protected IterationNextOperation(Codec codec, int flags, byte[] cacheName, Atomi
                entries.add(new SimpleEntry<>(unmarshall(key, status), (E) value));
             }
          }
-         segmentKeyTracker.segmentsFinished(finishedSegments);
-
       }
+      segmentKeyTracker.segmentsFinished(finishedSegments);
```
wburns: Ah another bug if no entries were in those segments?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/iteration/RemoteCloseableIterator.java
gustavonalle: Yes, this can be removed, it's probably a left over from early versions. The same for ```restartIteration```

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/iteration/SegmentKeyTracker.java
gustavonalle: :+1:

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/IterationNextOperation.java
gustavonalle: This is to allow the client to receive completed segments even if no entries were pulled out of the remote iteration

wburns:
Looks good, pulling

wburns:
Integrated into master, thanks @gustavonalle !

PR #4031, opened 2016-02-22
ISPN-6249 Streams filter key segments with using count can cause

wburns:
deadlock issues

* Make sure processed segments are checked when completing

https://issues.jboss.org/browse/ISPN-6249


gustavonalle:
integrated, thanks @wburns !

PR #4032, opened 2016-02-23
ISPN-6210 Make the ScriptingManager use an unwrapped cache when security is enabled

tristantarrant:
https://issues.jboss.org/browse/ISPN-6210


scripting/src/main/java/org/infinispan/scripting/impl/ScriptingManagerImpl.java
```diff
@@ -69,7 +69,7 @@ public void initialize(final EmbeddedCacheManager cacheManager, InternalCacheReg
 
    Cache<String, String> getScriptCache() {
       if (scriptCache == null) {
-         scriptCache = cacheManager.getCache(SCRIPT_CACHE);
+         scriptCache = SecurityActions.getUnwrappedCache(cacheManager.getCache(SCRIPT_CACHE));
```
wburns: Unfortunately I am not familiar with the chain of invocation here, but since the SecurityActions doesn't use a Privileged block does that mean the original invocation of this requires ADMIN privileges?

tristantarrant: Yes, but since this happens in the server initialization, that's exactly what happens

gustavonalle:
Pulling

gustavonalle:
Integrated, thanks @tristantarrant !

PR #4033, opened 2016-02-23
ISPN-6254 State transfer hangs after CacheNotFoundResponse

danberindei:
https://issues.jboss.org/browse/ISPN-6254

Only call SemaphoreCompletionService.continueTaskInBackground() if
the START_STATE_TRANSFER command was successful.


tristantarrant:
Pushed, thanks

PR #4034, opened 2016-02-24
ISPN-6258 Avoid NPE when subject has no principals

vjuranek:
https://issues.jboss.org/browse/ISPN-6258


slaskawi:
There are some new failures, retriggering CI.

slaskawi:
Integrated thanks!

PR #4035, opened 2016-02-24
ISPN-6255 Add log4j2 configuration files for CI

danberindei:
https://issues.jboss.org/browse/ISPN-6255


slaskawi:
How about creating a new module (test-commons?) and placing this file there? The next step would be to add this newly created module in code in test scope.

This way the file will be on the classpath and we will not use any profiles (which are a big hackish).

slaskawi:
We decided to push it as is.... integrating.

slaskawi:
Integrated thanks!

PR #4037, opened 2016-02-24
ISPN-6261 When iterate over all caches, take care about secured caches

vjuranek:
https://issues.jboss.org/browse/ISPN-6261
https://issues.jboss.org/browse/ISPN-6515

### DIFF
--- a/core/src/main/java/org/infinispan/stats/impl/CacheContainerStatsImpl.java
+++ b/core/src/main/java/org/infinispan/stats/impl/CacheContainerStatsImpl.java
@@ -1,8 +1,12 @@
 package org.infinispan.stats.impl;
 
+import java.util.HashSet;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
+import org.infinispan.AdvancedCache;
+import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.jmx.JmxStatisticsExposer;
 import org.infinispan.jmx.annotations.DataType;
@@ -62,11 +66,7 @@ public class CacheContainerStatsImpl implements CacheContainerStats, JmxStatisti
    @Override
    public void resetStatistics() {
       if (getStatisticsEnabled()) {
-         for (String cn : cm.getCacheNames()) {
-            if (cm.cacheExists(cn)) {
-               cm.getCache(cn).getAdvancedCache().getStats().reset();
-            }
-         }
+         getEnabledStats().forEach( stats -> stats.reset());
          resetNanoseconds.set(timeService.time());
       }
    }
@@ -95,13 +95,11 @@ public class CacheContainerStatsImpl implements CacheContainerStats, JmxStatisti
    protected long calculateAverageReadTime() {
       long totalAverageReadTime = 0;
       int includedCacheCounter = 0;
-      for (String cn : cm.getCacheNames()) {
-         if (cm.cacheExists(cn)) {
-            long averageReadTime = cm.getCache(cn).getAdvancedCache().getStats().getAverageReadTime();
-            if (averageReadTime > 0) {
-               includedCacheCounter++;
-               totalAverageReadTime += averageReadTime;
-            }
+      for (Stats stats : getEnabledStats()) {
+         long averageReadTime = stats.getAverageReadTime();
+         if (averageReadTime > 0) {
+            includedCacheCounter++;
+            totalAverageReadTime += averageReadTime;
          }
       }
       if (includedCacheCounter > 0) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/infinispan/stats/impl/CacheContainerStatsImpl.java
```diff
@@ -64,7 +64,11 @@ public void resetStatistics() {
       if (getStatisticsEnabled()) {
          for (String cn : cm.getCacheNames()) {
             if (cm.cacheExists(cn)) {
-               cm.getCache(cn).getAdvancedCache().getStats().reset();
+               try {
+                  cm.getCache(cn).getAdvancedCache().getStats().reset();
+               } catch (SecurityException e) {
+                  // no-op, skipping this cache
```
danberindei: Couldn't we avoid the `SecurityException` altogether?

vjuranek: I see two possibilities how to avoid it:
* check, if the user has access to the cache before accessing it, but AFAICT in case user has access to it, it would be checked twice (before accessing the cache and during accessing the cache), so this wouldn't be very efficient approach (especially when assuming that in most of the cases user has correct permissions and has access to the case)
* use `SecurityActions` to access the cache without checks, but my understanding is that also statistics should respect security setup and shouldn't access caches without required permissions (i.e. IMHO this is not an option at all, but maybe I'm wrong)

danberindei: I was thinking that we could use a proper user to call `CacheContainerStatsImpl.resetStatistics()`, and this method shouldn't have to worry about security at all.

I don't really like the idea of skipping caches we don't have access to, because the user gets no feedback why he's seeing partial statistics. @tristantarrant WDYT?

vjuranek: > I was thinking that we could use a proper user
I'm not sure what actually you mean by proper user. If one cache has user A as an admin and another cache user B, who is the proper user in this case? As mentioned above, this can be solved by {{SecurityActions}} (i.e. skipping security checks), but I'm really not sure we want this, IMHO user which hasn't access to the cache should see its statistics at all and shouldn't be even aware such cache exists. Anyway, lets wait what @tristantarrant thinks about it.

tristantarrant: I agree with @danberindei: a user has no way of knowing which caches would be included / excluded. I think also that, when security is enabled, only a cache ADMIN would have access to stats ops anyway.

vjuranek: sorry, @tristantarrant , I probably don't understand, so what the correct solution should look like?

> a user has no way of knowing which caches would be included / excluded

exactly, this was original motivation why I started this PR

>  I think also that, when security is enabled, only a cache ADMIN would have access to stats ops anyway.

this is obviously not true, at least in current implementation. Statistics is touched in various places - see linked jira. As an ADMIN of the cache, I get exception when accessing the cache via CLI, as statistics try to update all caches, including internal caches, in this cache script cache and fails as I hadn't `__scriptmanager` role. 

It's also possible we should change statistics implementation (haven't looked on statistics yet, so I have currently no opinion on it), but if we want to iterate over all caches in statistics, I don't see any other solution that skip caches which I haven't access to (or skip security check at all)

danberindei: It's definitely a problem if you're trying to read/reset the statistics of one cache and Infinispan reads/resets the statistics of all caches. The stack trace in the JIRA description doesn't seem right, so I can't tell what attribute the CLI was trying to read, maybe you can add some more information about reproducing this?

My thinking was that these are container-level attributes/operations, so whoever tries to use them should be a global ADMIN. That means it's ok to fail if the user is not a global ADMIN, even if we have a single secure cache and the user is an ADMIN on that cache. (However, the cache-level ADMIN should be able to read/reset the statistics on that particular cache.) 

We can do this by checking the permissions once, before we start iterating on the caches, and then using a `PrivilegedAction` to skip the security checks of the individual caches.

I'm not sure how this is exposed in the CLI, so let me know if any of this make sense @vjuranek @tristantarrant !

vjuranek: sorry for poor description, I've updated [ISPN-6261](https://issues.jboss.org/browse/ISPN-6261) with steps how to reproduce this issue. Let me know if it's not sufficient or anything is unclear. Basically, I don't do anything special, just start CLI, navigate to cache container and run `ls` there

slaskawi:
Hey guys! What is the status of this PR?

tristantarrant:
The real fix should go into the "entry" points: these should wrap the calls in PrivilegedActions, assuming the "invoker" has sufficient privileges.
For JMX, this means that the Platform MBean Server should provide an appropriate JMXPrincipal and probably the ResourceDMBean could wrap it in a Subject.
For DMR, this should be handled by the Command operations.

slaskawi:
@vjuranek Could you tell me if you plan to introduce changes suggested by Tristan?

vjuranek:
@slaskawi yes, but not sure about the time scale - haven't looked on it yet, but I expect it will be little bit more expensive than these few line of code, so there could be some delay, based on my time capabilities. If you need this to be fixed ASAP, feel free to close and implement in another PR.

slaskawi:
@vjuranek No worries - take your time... I'm just doing daily routine looking for inactive PRs

vjuranek:
@tristantarrant so I finally find some time to look on this again and I'm not sure I understand the changes you've proposed. Namely I don't understand how this could help, as the principals is already propagated to the caller and the rights are checked. 

Looking into it, IMHO the issue is that `___script_cache` is visible to users (therefore is touched by statistics) **and** it defines its own specific role (`___script_manager`), which can touch the cache (don't use any of user defined roles).

tristantarrant:
Internal caches have stats disabled: https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/registry/impl/InternalCacheRegistryImpl.java#L59

vjuranek:
Thanks for pointing this out @tristantarrant . So the issue is that these stat ops. don't take care if stats are enabled on cache level and run on all caches. PR updated and should address this issue and also ISPN-6515 which appeared after fixing stats.

galderz:
Integrated, thanks @vjuranek :)

PR #4042, opened 2016-02-25
ISPN-6269 Avoid security exception during startup of REST server

vjuranek:
https://issues.jboss.org/browse/ISPN-6269


tristantarrant:
Pushed thanks

PR #4043, opened 2016-02-25
ISPN-6271 Stream filterKeys reads keys not owned remotely when it

wburns:
shouldn't

* Make sure key filter uses local cache to look up on each node

https://issues.jboss.org/browse/ISPN-6271


anistor:
Looks good. Pulling ...

anistor:
Integrated. Thanks @wburns !

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
```diff
@@ -1996,4 +1996,17 @@ public void testKeySegmentFilter() {
       assertEquals(realCount.get(), createStream(entrySet).filterKeySegments(
               IntStream.range(0, segments).boxed().collect(Collectors.toSet())).count());
    }
+
```
anistor: There were some unused imports I removed before merging this.

PR #4045, opened 2016-02-25
ISPN-6274- Create tests for Remote Script Execution - Server Tasks deployment

andyuk1986:
Added tests covering Server Task Execution over Hotrod Client functionality - ISPN-5815. 
Tests are executed in standalone mode.

### DIFF
--- a/server/integration/testsuite/pom.xml
+++ b/server/integration/testsuite/pom.xml
@@ -118,6 +118,11 @@
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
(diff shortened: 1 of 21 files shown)


tristantarrant:
Thanks, @andyuk1986.
I would however like new clustered tests to use domain mode by default.

andyuk1986:
Agree @tristantarrant , we will migrate a few tests to domain mode. But I think these tests are necessary  - I have found and reported a few bugs based on them.

tristantarrant:
Pushed, thanks

PR #4046, opened 2016-02-25
ISPN-6248 - No Marshaller available in Server Task while executing in clustered mode

andyuk1986:


danberindei:
@andyuk1986 There is a {{DistributedCacheUsingTask}} in infinispan-server-testsuite logging the marshaller, could you change it to assert the marshaller is non-null instead?
There's also a {{LocalTestServerTask}} that apparently uses the marshaller to deserialize a key and value, could you make sure that test really executes the task and verifies the result?

andyuk1986:
@danberindei I have added assert for marshaller to be non-null for both DistributedCacheUsingTask and LocalTestServerTask. The rest test verifications are done in corresponding integration tests - LocalServerTaskIT and DistributedServerTaskIT.
Also made a few changes to the 2 test cases which were dependent on this pull request.

tristantarrant:
Pushed, thanks @andyuk1986

PR #4048, opened 2016-02-29
ISPN-6277  Avoid NullPointerException errors in the core test suite logs

danberindei:
https://issues.jboss.org/browse/ISPN-6277

Small fixes to avoid NPEs in the core test suite logs


core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
```diff
@@ -118,12 +118,12 @@ public void inject(Transport transport,
       this.cacheManagerNotifier = cacheManagerNotifier;
       this.cacheManager = cacheManager;
       this.timeService = timeService;
+      this.eventLog = eventLogManager.getEventLogger();
```
tristantarrant: while you still may want to grab the eventlogmanager here, retrieving the actual eventlogger should still happen in the start() method, since the server event logger is different

danberindei: So the server event logger replaces the regular event logger during startup?

tristantarrant: https://github.com/infinispan/infinispan/blob/master/server/integration/event-logger/src/main/java/org/infinispan/server/eventlogger/LifecycleCallbacks.java#L42

core/src/test/java/org/infinispan/manager/ClusterExecutorTest.java
```diff
@@ -0,0 +1,327 @@
...
+import static org.infinispan.test.TestingUtil.withCacheManagers;
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertTrue;
+import static org.testng.AssertJUnit.fail;
+
+/**
+ * @author Will Burns
+ * @since 8.2
+ */
+@Test(groups = {"functional", "smoke"}, testName = "manager.CacheManagerTest")
```
wburns: Thanks @danberindei :)  I hadn't gotten around to moving this.  Also we can update the testName

core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
danberindei: I'll roll back my changes for now, since https://github.com/infinispan/infinispan/pull/4053 has a more comprehensive solution.

core/src/test/java/org/infinispan/manager/ClusterExecutorTest.java
wburns: Just to make sure this wasn't overlooked the testName needs to be changed to manager.ClusterExecutorTest

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
```diff
@@ -661,43 +665,37 @@ private void terminate(String cacheName) {
...
-               globalComponentRegistry.getComponent(CacheManagerJmxRegistration.class).stop();
-               globalComponentRegistry.stop();
 
-            } else {
-               log.trace("Ignore call to stop as the cache manager is stopping");
-            }
+      synchronized (this) {
+         if (stopping) {
+            log.trace("Ignore call to stop as the cache manager is stopping");
+            return;
```
wburns: seems we can remove the return and just have the rest of the code in an else block

wburns:
Seems fine to me, just a couple minor points.

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
danberindei: Personally, I like the return style better, I try to avoid nesting as much as possible.

core/src/test/java/org/infinispan/manager/ClusterExecutorTest.java
danberindei: Updated @wburns! I just overlooked the comment, but clarifying things is never a bad idea ;)

gustavonalle:
@danberindei There are some security IT tests that fail only on builds from this PR, for example 
http://ci.infinispan.org/project.html?projectId=Infinispan&testNameId=7862150477724764799&tab=testDetails&guest=1

tristantarrant:
Doesn't matter what CI says, I'm still seeing many failurs on the security ITs: 
Results :

Tests in error: 
  KrbLdapAuthenticationIT>AbstractAuthentication.tearDown:116 » PrivilegedAction
  KrbLdapAuthenticationIT>AbstractAuthentication.tearDown:116 » PrivilegedAction
  KrbLdapAuthenticationIT>AbstractAuthentication.tearDown:116 » PrivilegedAction
  KrbLdapAuthenticationIT>AbstractAuthentication.tearDown:116 » PrivilegedAction
  KrbLdapAuthenticationIT>AbstractAuthentication.tearDown:116 » PrivilegedAction
  LdapAuthenticationIT>AbstractAuthentication.tearDown:116 » PrivilegedAction
  LdapAuthenticationIT>AbstractAuthentication.tearDown:116 » PrivilegedAction
  LdapAuthenticationIT>AbstractAuthentication.tearDown:116 » PrivilegedAction
  LdapAuthenticationIT>AbstractAuthentication.tearDown:116 » PrivilegedAction
  LdapAuthenticationIT>AbstractAuthentication.tearDown:116 » PrivilegedAction

Tests run: 52, Failures: 0, Errors: 10, Skipped: 0

danberindei:
The failures in `KrbLdapAuthenticationIT` and `LdapAuthenticationIT` were because the global ACL cache needs to be stopped after all the secure caches, but there was no explicit dependency between them. Instead, `DefaultCacheManager` didn't stop internal caches at all, which was fine in this case because the ACL cache is a simple-cache. I changed `AuthorizationManagerImpl` do add a dependency from each secure cache to the ACL cache.

tristantarrant:
security-it now is fine, but RolePermissionNoACLCacheTest>SingleCacheManagerTest.createBeforeClass:44->RolePermissionTest.setup:44 » PrivilegedAction
 is failing consistently

tristantarrant:
RolePermissionNoACLCacheTest fails consistently

danberindei:
Sorry about that @tristantarrant, should be fixed now.

tristantarrant:
Thanks, pushed

core/src/test/java/org/infinispan/distribution/rehash/NonTxStateTransferOverwritingValue2Test.java
```diff
@@ -200,22 +200,27 @@ public void commitEntry(CacheEntry entry, Metadata metadata, FlagAffectedCommand
...
-               @Override
-               public void commit(DataContainer container, Metadata metadata) {
-                  checkPoint.trigger("pre_commit_entry_" + getKey() + "_from_" + source);
-                  try {
-                     checkPoint.awaitStrict("resume_commit_entry_" + getKey() + "_from_" + source, 10, SECONDS);
-                  } catch (InterruptedException e) {
-                     throw new RuntimeException(e);
-                  } catch (TimeoutException e) {
-                     throw new RuntimeException(e);
+            if (entry instanceof ClearCacheEntry) {
```
rvansa: Just noticed this in master... is there any reason for the second instanceof ClearCacheEntry check?

PR #4050, opened 2016-02-29
ISPN-6280 Moved JCacheFailoverIT to it's own testsuite

slaskawi:
https://issues.jboss.org/browse/ISPN-6280

Provided fix for this issue is not trivial and requires many small changes:
* `JCacheFailoverIT` was moved to integrationtests since it depends on server assembly
* Since most of the latest tests depend on Uber Jars, `JCacheFailoverIT` was put in `all-remote-it`
* It turned out that remote uber jar didn't have `wildfly-controller` dependency and it was causing errors during the test
* Finally it turned out that XNIO provider has wrong reference from services (it's package has been relocated). After investigation it turned out that [MSHADE-182](https://issues.apache.org/jira/browse/MSHADE-182) was the root cause of the error. A temporary fix was applied and proper JIRA for plugin upgrade was created.


gustavonalle:
> JCacheFailoverIT was moved to integrationtests since it depends on server assembly

You don't necessarily need to move the test because it depends on the server assembly, it'd be enough to add a dependency to:
```
<dependency>
    <groupId>org.infinispan.server</groupId>
    <artifactId>infinispan-server-build</artifactId>
    <version>${project.version}</version>
    <type>zip</type>
    <scope>test</scope>
</dependency>
```

and extract it accordingly in the pom with the unpack plugin, maven will always build jcache after the server. This is how it was working until recently

> It turned out that remote uber jar didn't have wildfly-controller dependency and it was causing errors during the test

So the remote uber jar is going to depend on a test artifact? This sounds strange, since in theory uberjars are "uber" :)

Btw, I think the JIRA issue is too generic named "Fix unstable tests", maybe it'd be better to narrow down to the test itself?

slaskawi:
Adding ISPN server distribution as a dependency to JCache Remote module doesn't sound right. The failover test uses 2 ISPN servers and turns them on and off. This makes it fall into the integration test category, so I moved it to proper module (of course assuming that my understanding is correct).

However there's a good point - server assembly dependency is missing from the testsuite itself. I'll modify the pom shortly.

Let me clarify the Uber Jar issue - since `wildfly-controller` was not present in the Remote Uber Jar - JCache Remote using it was broken (you can integrate you application either using small or uber jars, the former was fine, the latter was suffering issues). It was a bit unclear when I wrote that this is needed to get test running - it was needed to fix the Remote JCache integration with Uber Jars.

Yeah - I think you are right - this is so big that it deserves new JIRA.

Adding @galderz and @mgencur to the loop to get some more feedback.

slaskawi:
Thanks a lot for the review!

I updated the JIRA, added server build dependency, rebased and repushed.

integrationtests/all-remote-it/pom.xml
```diff
@@ -54,10 +43,81 @@
...
+                  <configuration>
+                     <redirectTestOutputToFile>true</redirectTestOutputToFile>
+                  </configuration>
+               </execution>
+            </executions>
+         </plugin>
+         <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-resources-plugin</artifactId>
+            <executions>
```
gustavonalle: Since the ```infinispan-server-build``` is now a dependency, it should be unpacked instead of using relatives paths to the server

gustavonalle:
Running the suite locally, I got this:
```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 26.074 sec <<< FAILURE! - in org.infinispan.all.remote.jcache.JCacheFailoverIT
testRemoteListener on testRemoteListener(org.infinispan.all.remote.jcache.JCacheFailoverIT)(org.infinispan.all.remote.jcache.JCacheFailoverIT)  Time elapsed: 24.805 sec  <<< FAILURE!
org.infinispan.client.hotrod.exceptions.TransportException: Could not fetch transport
        at sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
        at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:717)
        at sun.nio.ch.SocketAdaptor.connect(SocketAdaptor.java:111)
        at org.infinispan.client.hotrod.impl.transport.tcp.TcpTransport.<init>(TcpTransport.java:68)
        at org.infinispan.client.hotrod.impl.transport.tcp.TransportObjectFactory.makeObject(TransportObjectFactory.java:37)
        at org.infinispan.client.hotrod.impl.transport.tcp.TransportObjectFactory.makeObject(TransportObjectFactory.java:16)
        at infinispan.org.apache.commons.pool.impl.GenericKeyedObjectPool.borrowObject(GenericKeyedObjectPool.java:1220)
        at org.infinispan.client.hotrod.impl.transport.tcp.TcpTransportFactory.borrowTransportFromPool(TcpTransportFactory.java:403)
        at org.infinispan.client.hotrod.impl.transport.tcp.TcpTransportFactory.getTransport(TcpTransportFactory.java:247)
        at org.infinispan.client.hotrod.impl.operations.AbstractKeyOperation.getTransport(AbstractKeyOperation.java:44)
        at org.infinispan.client.hotrod.impl.operations.RetryOnFailureOperation.execute(RetryOnFailureOperation.java:53)
        at org.infinispan.client.hotrod.impl.RemoteCacheImpl.get(RemoteCacheImpl.java:493)
        at org.infinispan.jcache.remote.RemoteCacheWrapper.get(RemoteCacheWrapper.java:66)
        at org.infinispan.jcache.AbstractJCache.put(AbstractJCache.java:225)
        at org.infinispan.jcache.remote.JCache.put(JCache.java:137)
        at org.infinispan.all.remote.jcache.JCacheFailoverIT.testRemoteListener(JCacheFailoverIT.java:82)
```

slaskawi:
hmmm interesting error.. I'll have a look

slaskawi:
Thanks for the reivew. I updated the PR. Let's see what CI thinks about it.

gustavonalle:
@slaskawi There's a suspicious osgi failure on CI on some builds of this PR:
http://ci.infinispan.org/viewLog.html?buildId=36997&buildTypeId=bt9&guest=1#testNameId6998731395864049875&guest=1

slaskawi:
Looking at OSGi failure...

slaskawi:
I applied the fix but I can't reproduce the exact failure... Marking this one as a Preview and working with CI.

danberindei:
Thanks Sebastian, integrated!

About the test failure, maybe you couldn't reproduce it because you already had the infinispan-remote uber-jar from a previous run?

PR #4054, opened 2016-02-29
ISPN-6227 ContinuousQueryMultipleCachesDistTest fails randomly

anistor:
https://issues.jboss.org/browse/ISPN-6227

### DIFF
--- a/core/src/test/java/org/infinispan/test/AbstractCacheTest.java
+++ b/core/src/test/java/org/infinispan/test/AbstractCacheTest.java
@@ -3,13 +3,10 @@ package org.infinispan.test;
 import org.infinispan.Cache;
 import org.infinispan.configuration.cache.CacheMode;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
-import org.infinispan.distribution.ch.ConsistentHashFactory;
 import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.test.fwk.CleanupAfterMethod;
 import org.infinispan.test.fwk.CleanupAfterTest;
 import org.infinispan.test.fwk.TestCacheManagerFactory;
-import org.infinispan.util.ControlledConsistentHashFactory;
-import org.infinispan.util.ReplicatedControlledConsistentHashFactory;
 
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


wburns:
Seems fine to me integrating.

gustavonalle:
pulling

wburns:

danberindei:
@anistor How does this fix the failures?

anistor:
testCQCacheLeavesAndJoins calls killMember and then addClusterEnabledCacheManager during the test and does not wait for the new node to finish joining, which is ok, this should not cause and does not cause problems yet.

The next test will try to add a clustered listener but the joining node still did not finish joining (most of the times). The replicated callable is sent to the joining node too and DefaultExecutorService will complain that the node is not a member (SuspectException). I believe this is a known issue with clustered listeners, unrelated to this test, but could not find its jira. To fix this test either dispose cache managers between tests or add waitForClusterToForm() right after the addition of the new member (but that would be a bit silly).

I've extracted the relevant code to reproduce the SuspectException here https://gist.githubusercontent.com/anistor/ec4d1994d0d2c3b68cf2/raw/c671b0b8a59c2f20d45f378af31be191c72f3876/gistfile1.java

danberindei:
I guess in this case it doesn't matter because the test only has 2 methods, but in general I'd much rather have each test method clean up after itself (e.g. by waiting for the node to finish starting) than having to spend a lot of time to re-create the cluster for each test method.

anistor:
Btw, re the known issue with SuspectException at cluster listener registration, I still cannot find it in jira. Have the id?

danberindei:
I'm not aware of any open issues about cluster listener registration...

danberindei:
Your test seems to fail because {{CacheNotifierImpl.addListener()}} sends a distributed task to all the nodes in the cluster ({{DefaultCacheManager.getMembers()}}, but then {{DefaultExecutorService.submit()}} checks that the target node is a member of the cache topology. After the ISPN-6214 changes, creating the cache manager is enough to start the transport and join the JGroups cluster, so adding a cluster listener before starting the cache on the new node will trigger the SuspectException.

anistor:
Indeed. If no jira yet I'll add one now.

PR #4061, opened 2016-03-02
Fixed a few outdated comments on javadoc

wburns:


wburns:
Actually I will add it to the original executor JIIRA

PR #4065, opened 2016-03-02
ISPN-6302 Ignore DistributedServerTaskDeploymentIT

slaskawi:
https://issues.jboss.org/browse/ISPN-6302

Unfortunately I couldn't find any solution for fixing this test. Applying a work around and returning back to the original author.


anistor:
Integrated in master. Thanks @slaskawi !

PR #4066, opened 2016-03-02
ISPN-6306 Distributed Streams should also be used for Replication Cache

wburns:
* Replication now uses distributed streams by default
** Iterator/Spliterator method will always be local
* Cleaned up tests a bit

https://issues.jboss.org/browse/ISPN-6306


wburns:
Please wait on this one, not sure if this is causing an issue with rehash stress test or not yet.

wburns:
The issue is unrelated and will be handled in https://issues.jboss.org/browse/ISPN-6308

tristantarrant:
Pushed, thanks

PR #4067, opened 2016-03-03
ISPN-6307 Add datatype script parameter

galderz:
https://issues.jboss.org/browse/ISPN-6307

* Datatype script metadata parameter is designed for remote clients
  working purely with data in a particular format, e.g. JS clients
  with UTF-8 Strings.

### DIFF
--- a/documentation/src/main/asciidoc/user_guide/chapter-78-Scripting.adoc
+++ b/documentation/src/main/asciidoc/user_guide/chapter-78-Scripting.adoc
@@ -41,6 +41,13 @@ The following metadata property keys are available
 * collator: this property is only valid for +mapper+ scripts, and specifies the name of another script that will be used for the collation phase of a map/reduce job. This property will be removed in Infinispan 9.
 * combiner: this property is only valid for +mapper+ scripts, and specifies the name of another script that will be used for the combine phase of a map/reduce job. This property will be removed in Infinispan 9.
 * parameters: an array of valid parameter names for this script. Invocations which specify parameter names not included in this list will cause an exception.
+* datatype: optional property providing information, in the form of
+Media Types (also known as MIME) about the type of the data stored in the
+caches, as well as parameter and return values. Currently it only accepts a
+single value which is `text/plain; charset=utf-8`, indicating that data is
+String UTF-8 format. This metadata parameter is designed for remote clients
+that only support a particular type of data, making it easy for them to
+retrieve, store and work with parameters.
 
 Since the execution mode is a characteristic of the script, nothing special needs to be done on the client to invoke scripts in different modes.
 
--- a/scripting/src/main/java/org/infinispan/scripting/impl/DataType.java
+++ b/scripting/src/main/java/org/infinispan/scripting/impl/DataType.java
@@ -1,0 +1,100 @@
+package org.infinispan.scripting.impl;
+
+import org.infinispan.commons.CacheException;
+import org.infinispan.commons.marshall.Marshaller;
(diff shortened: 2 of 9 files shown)


scripting/src/main/java/org/infinispan/scripting/impl/DataType.java
```diff
@@ -0,0 +1,78 @@
+package org.infinispan.scripting.impl;
+
+import org.infinispan.commons.CacheException;
+import org.infinispan.commons.marshall.Marshaller;
+
+import java.io.UnsupportedEncodingException;
+import java.util.Optional;
+
+public enum DataType {
+   UTF8(new Utf8Transformer()),
```
tristantarrant: On the JIRA I mentioned that probably we should use the media type format as defined by IANA: http://www.iana.org/assignments/media-types/media-types.xhtml

galderz: Agreed to use `text/plain; charset=utf-8` in this case.

galderz:
Some compilation/test errors still present, so looking into those...

galderz:
Updated to accept mime data type and fixed compilation error. ExecTest failing is not related to this PR.

scripting/src/main/java/org/infinispan/scripting/impl/DataType.java
```diff
@@ -0,0 +1,105 @@
...
+         try {
+            return obj instanceof String ? ((String) obj).getBytes("utf8") : obj;
+         } catch (UnsupportedEncodingException e) {
+            throw new CacheException(e);
+         }
+      }
+
+      private static String asString(Object v) {
+         try {
+            return new String((byte[]) v, "utf8");
```
tristantarrant: Use the new String(null, Charset.forName("UTF-8")); form to avoid the try/catch. Constify the UTF-8 Charset elsewhere for extra points ;)

documentation/src/main/asciidoc/user_guide/chapter-78-Scripting.adoc
```diff
@@ -41,6 +41,13 @@ The following metadata property keys are available
 * collator: this property is only valid for +mapper+ scripts, and specifies the name of another script that will be used for the collation phase of a map/reduce job. This property will be removed in Infinispan 9.
 * combiner: this property is only valid for +mapper+ scripts, and specifies the name of another script that will be used for the combine phase of a map/reduce job. This property will be removed in Infinispan 9.
 * parameters: an array of valid parameter names for this script. Invocations which specify parameter names not included in this list will cause an exception.
+* datatype: optional property providing information, in the form of
```
tristantarrant: @rigazilla you want to take note and implement support for this in the C++ client too (simpler than the JBasicMarshaller)

tristantarrant:
Looks good, @galderz 
1 minor comment

documentation/src/main/asciidoc/user_guide/chapter-78-Scripting.adoc
galderz: Yeah, that would be a good idea, but @rigazilla remember one thing: you still need the JBasicMarshaller to be able to store the script remotely using the Hot Rod protocol at least with the default marshaller on the server. If the script can be stored using CLI operations, then we could get rid of JBasicMarshaller altogether but dunno if that's possible...

galderz:
@tristantarrant Updated

tristantarrant:
Pushed, thanks

PR #4068, opened 2016-03-03
ISPN-6311 - ExecTest.testRemoteMapReduceWithStreams test is  failing

andyuk1986:
Disabled as the test depends on bug ISPN-6173

### DIFF
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExecTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExecTest.java
@@ -117,6 +117,7 @@ public class ExecTest extends MultiHotRodServersTest {
       assertTrue(results.get("macbeth").equals(Double.valueOf(287)));
    }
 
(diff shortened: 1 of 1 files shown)


anistor:
@andyuk1986 ISPN-6311 was created just to disable a test?

wburns:
Also to note @andyuk1986 is that REPL will be the same once #4066 is integrated.

andyuk1986:
@anistor at the moment yes, as that test was failing in CI, and I didn't know another pretty way for performing this pull request. But seems that, as soon as #4066 is integrated I would need to disable more tests until the bug ISPN-6173 is fixed.

anistor:
In such cases creating a PR linked to the issue in question (ISPN-6173) is enough, provided there's a note that this PR just disables the failing test rather than fixing it, so the integrator would not mistakenly close the issue. No need to create a separate jira just for disabling a test.

anistor:
Integrated in master. Thanks @andyuk1986 !

andyuk1986:
@anistor OK, got it! Thanks a lot :)

PR #4071, opened 2016-03-03
ISPN-6314 ClusterTopologyManager.waitForJoinerView can wait forever

pruivo:
https://issues.jboss.org/browse/ISPN-6314

### DIFF
--- a/core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
+++ b/core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
@@ -1,6 +1,7 @@
 package org.infinispan.topology;
 
 
+import net.jcip.annotations.GuardedBy;
 import org.infinispan.commands.ReplicableCommand;
 import org.infinispan.commons.CacheException;
 import org.infinispan.commons.util.CollectionFactory;
@@ -32,13 +33,11 @@ import org.infinispan.remoting.rpc.ResponseMode;
 import org.infinispan.remoting.transport.Address;
 import org.infinispan.remoting.transport.Transport;
 import org.infinispan.remoting.transport.jgroups.SuspectException;
-import org.infinispan.util.TimeService;
 import org.infinispan.util.concurrent.TimeoutException;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 import org.infinispan.util.logging.events.EventLogCategory;
 import org.infinispan.util.logging.events.EventLogManager;
-import org.infinispan.util.logging.events.EventLogger;
 
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
```diff
@@ -460,15 +460,17 @@ private void waitForJoinerView(Address joiner, int joinerViewId, long timeout)
...
+               remainingTime = timeService.remainingTime(endTime, TimeUnit.MILLISECONDS);
+               if (remainingTime == 0 || !clusterManagerStatus.isRunning())
                   break;
-               clusterManagerLock.wait(timeService.remainingTime(endTime, TimeUnit.MILLISECONDS));
+               clusterManagerLock.wait(remainingTime);
             }
          }
-         if (timeService.isTimeExpired(endTime)) {
+         if (remainingTime == 0) { //zero is only returned by TimeService.remainingTime() when it expires.
             throw new TimeoutException("Timed out waiting for view " + joinerViewId);
```
tristantarrant: i18n ?

core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
```diff
@@ -460,15 +460,17 @@ private void waitForJoinerView(Address joiner, int joinerViewId, long timeout)
...
-               if (timeService.isTimeExpired(endTime) || !clusterManagerStatus.isRunning())
+               remainingTime = timeService.remainingTime(endTime, TimeUnit.MILLISECONDS);
+               if (remainingTime == 0 || !clusterManagerStatus.isRunning())
                   break;
-               clusterManagerLock.wait(timeService.remainingTime(endTime, TimeUnit.MILLISECONDS));
+               clusterManagerLock.wait(remainingTime);
             }
          }
-         if (timeService.isTimeExpired(endTime)) {
+         if (remainingTime == 0) { //zero is only returned by TimeService.remainingTime() when it expires.
```
slaskawi: Would it be possible to move this throw into the while loop (and rearrange lines 468-470)? This way it would be a little bit more readable.

danberindei:
Maybe we should use a {{Condition}} instead?

danberindei:
Thanks for finding this BTW, I also looked into `ConcurrentStartTest` hanging but I only found https://issues.jboss.org/browse/ISPN-6322 :)

pruivo:
Updated! I'm using a Condition now and rework a little bit the loop.

danberindei:
Integrated, thanks Pedro!

danberindei:
The log message was missing the id. I added the message id and pushed it directly.

pruivo:
thanks @danberindei

PR #4073, opened 2016-03-03
ISPN-6315 Add test for remote query using the client as-modules with annotation based protostream marshallers

anistor:
jira: https://issues.jboss.org/browse/ISPN-6315
https://issues.jboss.org/browse/ISPN-6312

This is for 8.1.x.

### DIFF
--- a/integrationtests/as-integration-client/pom.xml
+++ b/integrationtests/as-integration-client/pom.xml
@@ -72,6 +72,18 @@
             <groupId>org.infinispan</groupId>
             <artifactId>infinispan-client-hotrod</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.infinispan</groupId>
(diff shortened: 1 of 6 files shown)


integrationtests/as-integration-client/pom.xml
```diff
@@ -72,6 +72,18 @@
             <groupId>org.infinispan</groupId>
             <artifactId>infinispan-client-hotrod</artifactId>
         </dependency>
+        <dependency>
```
slaskawi: Just checking.. do we need to add any of those dependencies to the Uber Jars?

anistor: @slaskawi These are already in the remote client uberjar. But are they tested? It seems ```RemoteAllTest``` is all we have atm.

tristantarrant:
Pushed, thanks

PR #4075, opened 2016-03-04
ISPN-6320 Improve stability of ClientEventsOOMTest

slaskawi:
https://issues.jboss.org/browse/ISPN-6320

My local runs shows that sometimes this test takes pretty long (since JVM struggles for life in this test, such things might happen). I think this is the root cause why it fails [1].

[1] http://ci.infinispan.org/project.html?projectId=Infinispan&testNameId=5048793009923214371&tab=testDetails


anistor:
Integrated. Thanks @slaskawi !

PR #4079, opened 2016-03-04
ISPN-6308 Stream iterator method can fail sometimes with a node failure

wburns:
* Fixed bug introduced by ISPN-6127
* LocalStreamManager stored id only now the id also has node name

https://issues.jboss.org/browse/ISPN-6308


tristantarrant:
Pushed, cheers !

PR #4082, opened 2016-03-07
ISPN-6329 Added ISPN Core to Remote Uber Jar

slaskawi:
https://issues.jboss.org/browse/ISPN-6329

Required for Remote Cache Store use cases.


slaskawi:
We agreed with Tristan that this change does not make sense

PR #4085, opened 2016-03-07
ISPN-6307 Adjust server task engine

galderz:
https://issues.jboss.org/browse/ISPN-6307

* Since the introduction of the datatype metadata header, it makes sense
  for the task engine to take ownership of the marshalling/unmarshalling
  of parameters and return types, as such, the server task engine needed
  adjusting to this new reality.
* One of the server task engine tests had issues with this assumption,
  where a remote server task was then running a JS file locally. This
  should be seen as an edge case (e.g. why not execute the JS file
  remotely?) and hence it makes sense for the task to do some extra work
  in this case.


tristantarrant:
Pushed, thanks

PR #4086, opened 2016-03-07
ISPN-6332 L1CachingIT random failures

danberindei:
https://issues.jboss.org/browse/ISPN-6332

Use more keys, and use Assume to ignore the test if all keys land on the
same node.


tristantarrant:
Pushed, thanks

PR #4087, opened 2016-03-07
ISPN-6323 Cluster startup hangs indefinitely in client-server mode

anistor:
* avoid creation of remote query components for internal caches because the starting of __protobuf_metadata cache may lead to deadlock

Jira: https://issues.jboss.org/browse/ISPN-6323


tristantarrant:
Pushed, thanks

PR #4089, opened 2016-03-08
ISPN-6306 Dist Stream iterator can return with no results in some cases

wburns:
* Closing timing issue by checking queue again after completion

https://issues.jboss.org/browse/ISPN-6335

### DIFF
--- a/core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
+++ b/core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
@@ -5,7 +5,6 @@ import org.infinispan.commons.CacheException;
 import org.infinispan.commons.marshall.Externalizer;
 import org.infinispan.commons.marshall.SerializeWith;
 import org.infinispan.commons.util.CloseableIterator;
-import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.distribution.DistributionManager;
 import org.infinispan.distribution.ch.ConsistentHash;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -733,13 +733,22 @@ public R get() {
...
                }
                return null;
             }
             nextLock.lock();
             try {
                boolean interrupted = false;
-               while ((entry = queue.poll()) == null && !completed.get()) {
+               // We have to poll queue after completed just in case if we completed and sent an item to the
+               // queue and completed iterator that we don't lose the item
+               while ((entry = queue.poll()) == null && !completed.get() && (entry = queue.poll()) == null) {
```
anistor: poll twice in a row?

anistor: I understand the reason for it but was wondering if there's maybe a way to write it in a less weird way than relying on short-circuit evaluation.

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -733,13 +733,22 @@ public R get() {
             if (completed.get()) {
```
anistor: I think I saw some unused imports in DistributedCacheStream.

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
danberindei: +1, one assignment in a `while` condition is already too much :)

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -733,13 +733,22 @@ public R get() {
...
+                  return entry;
                }
                return null;
             }
             nextLock.lock();
             try {
                boolean interrupted = false;
-               while ((entry = queue.poll()) == null && !completed.get()) {
+               // We have to poll queue after completed just in case if we completed and sent an item to the
+               // queue and completed iterator that we don't lose the item
```
danberindei: Could you clarify this a bit? I'm still not sure what's going on with the completed iterator :)

wburns: I tried to explain this in more detail at https://issues.jboss.org/browse/ISPN-6335.  Did you want me to move this over to here?

wburns: Also am adding some more wording to describe it.

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
wburns: Removed Configuration from imports.

wburns:
Updated, hopefully it is better :)

danberindei:
Integrated, thanks Will!

PR #4092, opened 2016-03-08
ISPN-6323 Cluster startup hangs indefinitely in client-server mode

anistor:
https://issues.jboss.org/browse/ISPN-6323

 ___protobuf_metadata cache should be started lazily.


wburns:
Integrated into 8.1.x, thanks @anistor !

PR #4095, opened 2016-03-09
ISPN-6342 Fix ConfigurationUnitTest

slaskawi:
https://issues.jboss.org/browse/ISPN-6342

### DIFF
--- a/core/src/main/resources/schema/infinispan-config-9.0.xsd
+++ b/core/src/main/resources/schema/infinispan-config-9.0.xsd
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified"
-           version="1.0" targetNamespace="urn:infinispan:config:8.2"
-           xmlns:tns="urn:infinispan:config:8.2" xmlns:xs="http://www.w3.org/2001/XMLSchema">
+           version="1.0" targetNamespace="urn:infinispan:config:9.0"
+           xmlns:tns="urn:infinispan:config:9.0" xmlns:xs="http://www.w3.org/2001/XMLSchema">
 
   <xs:element name="infinispan" type="tns:infinispan">
     <xs:annotation>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -53,7 +53,7 @@
 
    <properties>
       <infinispan.codename>Chakra</infinispan.codename>
-      <infinispan.core.schema.version>8.2</infinispan.core.schema.version>
+      <infinispan.core.schema.version>9.0</infinispan.core.schema.version>
 
       <version.java>1.8</version.java>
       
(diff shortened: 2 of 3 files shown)


core/src/main/resources/schema/infinispan-config-9.0.xsd
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified"
-           version="1.0" targetNamespace="urn:infinispan:config:8.2"
-           xmlns:tns="urn:infinispan:config:8.2" xmlns:xs="http://www.w3.org/2001/XMLSchema">
+           version="1.0" targetNamespace="urn:infinispan:config:9.0"
```
wburns: Confused as to how this would fix a test.  We haven't branched yet so we shouldn't be adding 9.0 changes yet afaik.

danberindei: @wburns The version number in master is 9.0.0-SNAPSHOT, and `ConfigurationUnitTest.testSchema()` tries to load `infinispan-config-<major>.<minor>.xsd`. So we probably don't need the namespace changes yet, but we need the new schema file.

core/src/main/resources/schema/infinispan-config-9.0.xsd
```diff
@@ -1178,7 +1178,7 @@
               Must implement `org.infinispan.distribution.ch.KeyPartitioner`.
               A custom key partitioner can be used as an alternative to grouping, to guarantee that some keys
               are located in the same segment (and thus their primary owner is the same node).
-              Since 8.2.
+              Since 9.0.
```
danberindei: 8.2 seems fine here.

core/src/main/resources/schema/infinispan-config-9.0.xsd
wburns: Sounds like we need to branch then :)  Unless we aren't going to do anything more for 8.2.

core/src/main/resources/schema/infinispan-config-9.0.xsd
```diff
@@ -1258,7 +1258,7 @@
...
```
wburns: Can be removed here as well.

core/src/main/resources/schema/infinispan-config-9.0.xsd
slaskawi: All right... I'll leave it than

core/src/main/resources/schema/infinispan-config-9.0.xsd
slaskawi: Ok, I'll leave it as is

slaskawi:
Sorry I missed a couple of bits. Now it should be fine.

danberindei:
Integrated, thanks Sebastian!

PR #4097, opened 2016-03-09
ISPN-6341 + ISPN-5507

danberindei:
ISPN-6341 StateTransferManager should be the first component to stop
 	ISPN-5507 Shutdown shouldn't wait for remote transactions

https://issues.jboss.org/browse/ISPN-6341
https://issues.jboss.org/browse/ISPN-5507

### DIFF
--- a/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
+++ b/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
@@ -14,6 +14,7 @@ import org.infinispan.commons.util.concurrent.jdk8backported.EquivalentConcurren
 import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.configuration.cache.Configurations;
 import org.infinispan.context.InvocationContextFactory;
+import org.infinispan.factories.annotations.ComponentName;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.factories.annotations.Start;
 import org.infinispan.factories.annotations.Stop;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
```diff
@@ -596,49 +602,75 @@ private void calculateMinTopologyId(int idOfRemovedTransaction) {
...
-            if (clustered) {
-               log.debugf("Interrupted waiting for on-going transactions to finish. %s local transactions and %s remote transactions", localTransactions.size(), remoteTransactions.size());
-            } else {
-               log.debugf("Interrupted waiting for %s on-going transactions to finish.", localTransactions.size());
+            log.debugf("Interrupted waiting for %d on-going local transactions to finish.", localTransactions.size());
+         }
+      }
+
+      if (executorService != null) {
+         Future<?> remoteTxsFuture = executorService.submit(new Runnable() {
```
wburns: I am curious since this is a single thread executor service it seems this could get delayed from another node shutting down?  It seems we would want to butt in front?  To be honest I am not as familiar with the transaction code to know for sure.

wburns:
Looks good to me, just had the 1 question.

core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
danberindei: Yeah, I didn't want a new executor because one thread per cache is already too much here. I guess it would have been better to use one of the global thread pools, I'll make the change.

danberindei:
Updated to use the global TIMEOUT_SCHEDULE_EXECUTOR for checking remote transactions.

wburns:
Looks like test failures are related to something else, pulling..

wburns:
Integrated into 8.2.x and master, thanks @danberindei !

PR #4098, opened 2016-03-09
ISPN-6043 Log the stacktrace if a listener fails

jmarkos:
Having only the exception message in the log is often not enough, i.e.

```
ERROR [org.infinispan.notifications.cachemanagerlistener.CacheManagerNotifierImpl] (sc-main) org.infinispan.commons.CacheListenerException: ISPN000280: Caught exception [java.lang.NullPointerException] while invoking method [public synchronized void org.radargun.service.InfinispanClustered.viewChanged(org.infinispan.notifications.cachemanagerlistener.event.ViewChangedEvent)] on listener instance: org.radargun.service.InfinispanClustered@78338b0a
```

### DIFF
--- a/core/src/main/java/org/infinispan/notifications/cachemanagerlistener/CacheManagerNotifierImpl.java
+++ b/core/src/main/java/org/infinispan/notifications/cachemanagerlistener/CacheManagerNotifierImpl.java
@@ -133,7 +133,7 @@ public class CacheManagerNotifierImpl extends AbstractListenerImpl<Event, Listen
       } catch (Exception x) {
          // Only cache entry-related listeners should be able to throw an exception to veto the operation.
          // Just log the exception thrown by the invoker, it should contain all the relevant information.
-         log.error(x);
+         log.failedInvokingCacheManagerListener(x);
       }
    }
 
--- a/core/src/main/java/org/infinispan/util/logging/Log.java
+++ b/core/src/main/java/org/infinispan/util/logging/Log.java
@@ -1408,4 +1408,8 @@ public interface Log extends BasicLogger {
 
    @Message(value = "The configured entity class %s is not indexable. Please remove it from the indexing configuration.", id = 404)
    CacheConfigurationException classNotIndexable(String className);
+
+   @LogMessage(level = ERROR)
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/infinispan/notifications/cachemanagerlistener/CacheManagerNotifierImpl.java
```diff
@@ -133,7 +135,10 @@ private void invokeListener(ListenerInvocation listener, EventImpl e) {
       } catch (Exception x) {
          // Only cache entry-related listeners should be able to throw an exception to veto the operation.
          // Just log the exception thrown by the invoker, it should contain all the relevant information.
-         log.error(x);
```
wburns: The reason the stack trace isn't logged is because it is using the error(Object) method which just invokes toString.  If you instead call log.error(String, Throwable) method it will print the stack trace.

Although I wonder if we want to only print the stack trace if tracing is enabled and leave it as is for when tracing is not?

danberindei: -1 to print the stack trace only if trace is enabled. If the exception is harmless and shouldn't be logged, the listener can catch it itself.

wburns: Alright, I just didn't know cause logging an entire stack trace to error can spam the output quite a bit possibly (even with no trace configuration), but I am open to either way.

danberindei: JGroups has a neat trick, refusing to log repeated messages more than once a second (I think). Unfortunately, it also has this habit of only logging the exception message, and it makes it very hard to figure out what's going on some of the time.

BTW, since this is an error message, it should be i18n-ed in the `Log` interface.

jmarkos:
Updated (feel free to change the log message).

wburns:
Seems fine to me. @danberindei ?

danberindei:
Ok, integrated. Thanks Jakub!

danberindei:
@jmarkos JIRA reference seems wrong, though :)

PR #4100, opened 2016-03-09
ISPN-6272 Extend CacheStream interface to allow for Serializable lambdas

wburns:
* Initial changes for CacheStream
* Removed unneeded casts from tests

https://issues.jboss.org/browse/ISPN-6272


wburns:
This PR only has changes related to CacheStream the primitive versions will come in https://issues.jboss.org/browse/ISPN-6349

wburns:
There are a few test failures I need to address
http://ci.infinispan.org/viewLog.html?buildId=37251&buildTypeId=bt9&guest=1

wburns:
Test is fixed and added upgrade documentation to state class moved packages.

core/src/main/java/org/infinispan/CacheStream.java
```diff
@@ -180,12 +184,21 @@
...
+    * @param action consumer to be ran for each element in the stream
     */
    @Override
    void forEach(Consumer<? super R> action);
 
    /**
+    * Same as {@link CacheStream#forEach(Consumer)} except that the Consumer must also implement Serializable.
+    * <p>
+    * This method will be used automatically by lambdas, which prevents users from having to manually cast to
+    * a Serializable lambda.
```
danberindei: I'd say `Serializable & Consumer` instead of "Serializable lambda".

core/src/main/java/org/infinispan/CacheStream.java
```diff
@@ -180,12 +184,21 @@
...
-    * @param action
+    * @param action consumer to be ran for each element in the stream
     */
    @Override
    void forEach(Consumer<? super R> action);
 
    /**
+    * Same as {@link CacheStream#forEach(Consumer)} except that the Consumer must also implement Serializable.
+    * <p>
+    * This method will be used automatically by lambdas, which prevents users from having to manually cast to
```
danberindei: s/prevents/saves/ ?

core/src/main/java/org/infinispan/CacheStream.java
wburns: Yeah, I was trying to limit how much stuff I had to change on each copy/paste :)  I kinda figured the first paragraph would be fine then.

core/src/main/java/org/infinispan/CacheStream.java
wburns: Sounds better, yeah

core/src/main/java/org/infinispan/CacheStream.java
danberindei: I see what you mean... I guess you could say "The compiler will pick this overload for lambda parameters, making them `Serializable`."

wburns: Sure I can do that.

core/src/main/java/org/infinispan/CacheStream.java
```diff
@@ -291,4 +341,297 @@
...
+    * Same as {@link CacheStream#allMatch(Predicate)} except that the Predicate must also implement Serializable.
+    * <p>
+    * This method will be used automatically by lambdas, which prevents users from having to manually cast to
+    * a Serializable lambda.
+    * @param predicate a non-interfering, stateless
+    *                  predicate to apply to elements of this stream that is serializable
+    * @return {@code true} if either all elements of the stream match the
+    * provided predicate or the stream is empty, otherwise {@code false}
+    */
+   boolean allMatch(SerializablePredicate<? super R> predicate);
```
danberindei: I'd linke to keep the serializable and non-serializable overloads grouped.

core/src/main/java/org/infinispan/CacheStream.java
```diff
@@ -291,4 +341,297 @@
...
+    * a Serializable lambda.
+    * @param comparator a non-interfering, stateless
+    *                   {@code Comparator} to compare elements of this stream that is also serializable
+    * @return an {@code Optional} describing the minimum element of this stream,
+    * or an empty {@code Optional} if the stream is empty
+    */
+   Optional<R> min(SerializableComparator<? super R> comparator);
+
+   /**
+    * Same as {@link CacheStream#reduce(BinaryOperator)}  except that the BinaryOperator must also implement Serializable.
```
danberindei: Extra space before "except" ;)

core/src/main/java/org/infinispan/CacheStream.java
wburns: I tried to keep them grouped and then it did by order of the original interface.  Do you mean you want me to keep all methods that have both in 1 spot and the ones where I didn't have to in another?

Btw the ones that aren't are ones that don't have a Stream return type.

core/src/main/java/org/infinispan/CacheStream.java
wburns: Got the iron sight going here :+1:

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
```diff
@@ -419,8 +423,7 @@ public void testObjMin() {
       CacheSet<Map.Entry<Integer, String>> entrySet = cache.entrySet();
 
       assertEquals(Integer.valueOf(0),
-              createStream(entrySet).min((Serializable & Comparator<Map.Entry<Integer, String>>)
-                      (e1, e2) -> Integer.compare(e1.getKey(), e2.getKey())).get().getKey());
+              createStream(entrySet).min((e1, e2) -> Integer.compare(e1.getKey(), e2.getKey())).get().getKey());
```
danberindei: I wonder if the comparator would be serializable if you used `min(Comparator.comparing(Map.Entry::getKey())`...

wburns: It will not unfortunately :(

wburns: Do we want a CacheComparator class maybe that defines this as being argument as Serializable and just passes it to Comparator class maybe? :)

core/src/main/java/org/infinispan/CacheStream.java
danberindei: Nope, I thought you had 2 overloads for all the methods. I think I had a C++ flashback and expected that the method from `Stream` would be hidden by the overload in `CacheStream`, so you'd have to re-declare it here :)

wburns: Ah gotcha, yeah we may have to bring them all up some day just for Javadoc reasons, heh.

wburns:
Talking with Dan also think I will move all of the new Serializable* classes into `org.infinispan.util.function` to mirror `java.util.function` package structure.

wburns:
Alright everything should be updated from comments.

danberindei:
Thanks Will, integrated!

PR #4102, opened 2016-03-09

anistor:
…gration tests

* define rest and memcachedCache to avoid being created based on the default config (which would fail due to classloading of InfinispanIndexManager)
* remove module attribute from cache definitions that do not need it

### DIFF
--- a/server/integration/testsuite/src/test/java/org/infinispan/server/test/configs/ExampleConfigsIT.java
+++ b/server/integration/testsuite/src/test/java/org/infinispan/server/test/configs/ExampleConfigsIT.java
@@ -25,7 +25,6 @@ import org.infinispan.arquillian.core.RemoteInfinispanServer;
 import org.infinispan.arquillian.core.RemoteInfinispanServers;
 import org.infinispan.arquillian.core.RunningServer;
 import org.infinispan.arquillian.core.WithRunningServer;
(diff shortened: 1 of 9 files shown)


slaskawi:
Looks good to me. Plus a green CI :)

anistor:
Found some more ```WFLYCTL0180: Services with missing/unavailable dependencies" => ["jboss.datagrid-infinispan-endpoint.memcached.memcached-connector is missing [jboss.datagrid-infinispan.clustered.memcachedCache]"]}``` in other tests. Will update the PR with more config fixes.

tristantarrant:
@anistor I believe those messages are not really affecting the test results, but it's because some of the xslt transformations on the config files are creating partially invalid configurations (which are not being tested)

tristantarrant:
@slaskawi  and CI was green, because only 3392 tests were executed

anistor:
They do not fail the tests. But are extremely annoying; the REST failure is in red in my console :)

tristantarrant:
We should make them a subtle shade of pink :)

anistor:
Updated. ATM there are only 2 instances of the (benign) memcached connector startup errors left, during execution of ExampleConfigsIT. This level of RED is good enough for now.

danberindei:
Integrated, thanks Adrian!

PR #4104, opened 2016-03-09

anistor:
…gration tests

* define rest and memcachedCache to avoid being created based on the default config (which would fail due to classloading of InfinispanIndexManager)
* remove module attribute from cache definitions that do not need it


danberindei:
Integrated, thanks Adrian!

PR #4105, opened 2016-03-09
ISPN-6352 ClusterExecutor.submitConsumer swallows exceptions thrown by

wburns:
the consumer

* Fixed exception handling for submitConsumer method

https://issues.jboss.org/browse/ISPN-6352


danberindei:
Integrated, thanks Will!

PR #4107, opened 2016-03-10
Fix a classloading issue if HR client tests are used with EAP modules

jmarkos:
This is a fix needed for product testing.


danberindei:
@jmarkos Need a JIRA reference :)

Also, I think the "proper" fix would be to set the classloader in the `RemoteCacheManager` configuration (`ConfigurationBuilder.classLoader(cl)`). If that doesn't work, then this is a bug in the production code, not in the test.

jmarkos:
@danberindei I looked around, and I don't see the classloader of RemoteCacheManager being used for anything. It was last used in the 5.2.x branch, or I'm missing something.

danberindei:
@jmarkos Ouch! True, it wasn't very encouraging when I didn't find any javadoc for `ConfigurationBuilder.classLoader()`, but there was only one thing I could imagine it being for: unmarshalling. You're right, it's not used at all, so I've created https://issues.jboss.org/browse/ISPN-6367

danberindei:
I added a comment referencing ISPN-6367 and I integrated it. Thanks Jakub!

PR #4117, opened 2016-03-12
ISPN-6371 Limit unboxing of Integer values in Distributed Streams

wburns:
* BitSet used to unbox only times equal to segments
* Unbox value manually to prevent unbox twice

https://issues.jboss.org/browse/ISPN-6371

### DIFF
--- a/core/src/main/java/org/infinispan/stream/impl/local/EntryStreamSupplier.java
+++ b/core/src/main/java/org/infinispan/stream/impl/local/EntryStreamSupplier.java
@@ -10,6 +10,7 @@ import org.infinispan.stream.impl.RemovableCloseableIterator;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
+import java.util.BitSet;
 import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
@@ -20,6 +21,7 @@ import java.util.stream.Stream;
  */
 public class EntryStreamSupplier<K, V> implements AbstractLocalCacheStream.StreamSupplier<CacheEntry<K, V>> {
    private static final Log log = LogFactory.getLog(EntryStreamSupplier.class);
+   private static final boolean trace = log.isTraceEnabled();
 
    private final Cache<K, V> cache;
    private final ConsistentHash hash;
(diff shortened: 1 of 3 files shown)


wburns:
Writing a custom JMH benchmark with 5 nodes and using count I saw ~ 5% increase in throughput.  This was pretty consistent from 2,500 entries to 250,000

The big difference is that instead of unboxing Integers N times where N is the # of entries it only unboxes M times where M is the number of segments targeted for that node.

wburns:
To note this does cause less performance when the number of entries in the cache is rather low.  For example when I had only 25 entries on 5 nodes with numOwners=2, which means ~10 entries per node, I saw a 2% drop in performance.  If we changed this to pass around BitSet everywhere this would not be an issue.

core/src/main/java/org/infinispan/stream/impl/local/EntryStreamSupplier.java
```diff
@@ -44,7 +45,9 @@ public EntryStreamSupplier(Cache<K, V> cache, ConsistentHash hash, Supplier<Stre
       }
       if (segmentsToFilter != null && hash != null) {
          log.tracef("Applying segment filter %s", segmentsToFilter);
-         stream = stream.filter(k -> segmentsToFilter.contains(hash.getSegment(k.getKey())));
+         BitSet bitSet = new BitSet(hash.getNumSegments());
```
anistor: BitSet seems like a good idea. I think it's worth changing StreamSupplier.buildStream to take a BitSet for segmentsToFilter.  I'd also add a check for log.isTraceEnabled() to avoid the toString().

anistor: And LocalCacheStream.filterKeySegments could continue to take a Set and convert it internally to BitSet.

wburns: I will think it over a bit more, this seemed like the easiest place since it is also done lazily.  Unless you are saying so the BitSet creation is only done in one place?

Also I am not sure what toString you are talking about.  The tracef doesnt' do a toString unless trace is enabled.  However I can a trace variable that will avoid the logging level lookup.

anistor: Was thinking AbstractLocalCacheStream.segmentsToFilter could become a BitSet.

Forget about toString.

wburns: Oh I see, sure I can do that.

wburns: Hrmm unfortunately it requires changing quite a few classes cause I want to have access to the hash to properly initialize the BitSet to the correct size and avoid any resizings.

anistor: Then it's ok the way it is now.

wburns: To be honest the way I want this is to eventually convert this over to using a BitSet completely, but I don't know if that is feasible until we also change the hash to return BitSets instead of Set<Integer> or at least some sort of native collection type so we can avoid these unboxings all over the place.

wburns: I even tried using Trove at one point and it seems to be the same performance as BitSet so there was no reason to use it really.

wburns: I will submit the trace changes though.

wburns:
Updated to add trace blocks.

anistor:
Pulling ..

anistor:
Integrated. Thanks @wburns !

PR #4118, opened 2016-03-13
ISPN-6302 -  DistributedServerTaskDeploymentIT fails

andyuk1986:
Changed the name of the jars to be deployed. I assume that the jars with the same name are not successfully deleted from the other test execution which leads to this test's fail.


danberindei:
@andyuk1986 Unless this failure is Windows-specific, I can't think of a reason why the jars could not be deleted.

andyuk1986:
I have no idea @danberindei , but as you can see from the CI build, renaming the jar resolved the issue.

slaskawi:
Integrating. 

@andyuk1986 - Could you please put url to JIRA in the PR description next time? It helps navigating though JIRA/Github.

slaskawi:
Integrated thanks!

andyuk1986:
@slaskawi sure will do :) Thank you :)

PR #4121, opened 2016-03-14

anistor:
…gration tests

* make the rest cache startup eager to avoid CNFE
* fix max-entries vs size
* use local caches for REST and memcached if the cache container is local

jira: https://issues.jboss.org/browse/ISPN-6353

This is take 2. The previous PR for this issue has introduced a failure in EvictionStrategyIT and also did not really solve the problem of the missing rest or memcached caches because the newly added ones failed to start anyway since they were replicated and the container was local.


slaskawi:
Pulling

slaskawi:
Integrated. Thanks!

PR #4122, opened 2016-03-14

anistor:
…gration tests

* make the rest cache startup eager to avoid CNFE
* fix max-entries vs size
* use local caches for REST and memcached if the cache container is local


slaskawi:
Pulling

slaskawi:
Integrated thanks!

PR #4130, opened 2016-03-15
ISPN-6380 - Fixed the coverage and jacocoReport profiles.

andyuk1986:
The tracking JIRA is: https://issues.jboss.org/browse/ISPN-6380

In the pom.xml you can see the big list of src folders which is added to sources of maven. This is done for jacoco, so that the generated report can also show the covered and uncovered lines directly on the class files. 

The build-helper-maven-plugin does not support the src folders inclusion via wildcards or include/exclude tags. If this is ok, I will file an issue in plugins repo, and as soon as it is implemented will change my code correspondingly.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -190,6 +190,8 @@
       <version.pax.url>2.2.0</version.pax.url>
       <version.karaf>3.0.4</version.karaf>
       <version.pax.exam>4.5.0</version.pax.exam>
+      <version.ant-nodeps>1.8.1</version.ant-nodeps>
+      <version.ant-contrib>1.0b3</version.ant-contrib>
       <dir.ispn>../</dir.ispn>
       <dir.jacoco>${session.executionRootDirectory}/jacoco/</dir.jacoco>
       <dir.jacoco.merged>../jacoco/merged/</dir.jacoco.merged>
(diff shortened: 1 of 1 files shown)


andyuk1986:
@slaskawi as we have agreed fixed and sent a pull request. Could you please have a look when you have time?

parent/pom.xml
```diff
@@ -2103,6 +2133,71 @@
...
+                        <configuration>
+                           <sources>
+                              <source>${dir.ispn}/atomic-factory/src/main/java</source>
+                              <source>${dir.ispn}/cdi/common/src/main/java</source>
+                              <source>${dir.ispn}/cdi/embedded/src/main/java</source>
+                              <source>${dir.ispn}/cdi/remote/src/main/java</source>
+                              <source>${dir.ispn}/checkstyle/src/main/java</source>
+                              <source>${dir.ispn}/cli/cli-client/src/main/java</source>
+                              <source>${dir.ispn}/cli/cli-interpreter/src/main/java</source>
+                              <source>${dir.ispn}/client/hotrod-client/src/main/java</source>
```
slaskawi: Could you tell me why do we need this part?

It might be really hard to maintain it properly... I bet someone will create a new module and forget to add it here... IMO we need some other approach

parent/pom.xml
```diff
@@ -2103,6 +2133,71 @@
                </plugin>
 
                <plugin>
+                  <groupId>org.codehaus.mojo</groupId>
+                  <artifactId>build-helper-maven-plugin</artifactId>
+                  <version>1.7</version>
```
slaskawi: This version should not be needed here... (already defined in `//build/pluginManagement`)

parent/pom.xml
andyuk1986: @slaskawi I have wrote about that in the comment above while sending pull request: 

In the pom.xml you can see the big list of src folders which is added to sources of maven. This is done for jacoco, so that the generated report can also show the covered and uncovered lines directly on the class files.

The build-helper-maven-plugin does not support the src folders inclusion via wildcards or include/exclude tags. If this is ok, I will file an issue in plugins repo, and as soon as it is implemented will change my code correspondingly.

If the new module is created and the dev forgot to add it here, it will not cause any problem. The result would be only that the generated jacoco report will show the coverage for the corresponding package on method level and not directly on the source code. 

You can see it in the comment attachment: the Report_with_source1.png and Report_with_source2.png files show the generated jacoco report for specified class when the source for that module was available above.; the Report_with_missing_source.png shows the generated report for the same class in case when the source for that class is not found during jacoco profile run (as you can see the method names are not clickable so you cannot view the coverage directly in the code, to see what is covered and what is not).
report_with_missing_source
report_with_source1
report_with_source2

parent/pom.xml
andyuk1986: Oops, haven't seen that. Fixed.

parent/pom.xml
slaskawi: Thanks for the explanation - I missed it.

I've been thinking about this and I think we should introduce Sonar analysis into the CI loop (it appeared a couple of times on the mailing list). So how about merging it as is for now and if we succeed with Sonar integration - we'll remove whole Jacoco reporting part (Sonar has build-in coverage visualizer). 

@andyuk1986 @tristantarrant @mgencur WDYT?

tristantarrant: +1

andyuk1986: Ok

slaskawi:
Pulling

slaskawi:
Integrated thanks!

PR #4133, opened 2016-03-16
ISPN-6322 Infinispan can miss incoming commands with JGroupsChannelLokup

danberindei:
Also fixes ISPN-6384 Fix JGroupsTransport.invokeRemotelyAsync timeout handling.
The ISPN-6322 commit adds a new test for JGroupsChannelLookup concurrent start, so there's another commit to stop using JGroupsChannelLookup in ConcurrentStartTest.

https://issues.jboss.org/browse/ISPN-6322
https://issues.jboss.org/browse/ISPN-6384
https://issues.jboss.org/browse/ISPN-5495

Please cherry-pick to 8.2.x as well.

### DIFF
--- a/core/src/test/java/org/infinispan/statetransfer/ConcurrentStartChanelLookupTest.java
+++ b/core/src/test/java/org/infinispan/statetransfer/ConcurrentStartChanelLookupTest.java
@@ -1,0 +1,125 @@
+package org.infinispan.statetransfer;
+
+import org.infinispan.Cache;
+import org.infinispan.configuration.cache.CacheMode;
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.configuration.global.GlobalConfigurationBuilder;
+import org.infinispan.lifecycle.ComponentStatus;
+import org.infinispan.manager.DefaultCacheManager;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.test.MultipleCacheManagersTest;
+import org.infinispan.test.fwk.CleanupAfterMethod;
+import org.infinispan.test.fwk.JGroupsConfigBuilder;
+import org.infinispan.test.fwk.TestResourceTracker;
+import org.infinispan.test.fwk.TransportFlags;
+import org.jgroups.JChannel;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.concurrent.Future;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.infinispan.test.TestingUtil.blockUntilViewsReceived;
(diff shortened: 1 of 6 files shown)


core/src/test/java/org/infinispan/statetransfer/ConcurrentStartChanelLookupTest.java
```diff
@@ -0,0 +1,118 @@
...
+import static org.infinispan.test.TestingUtil.blockUntilViewsReceived;
+import static org.infinispan.test.TestingUtil.extractGlobalComponentRegistry;
+import static org.infinispan.test.TestingUtil.waitForRehashToComplete;
+import static org.testng.AssertJUnit.assertEquals;
+
+/**
+ * Tests concurrent startup of replicated and distributed caches
+ *
+ * @author Dan Berindei
+ * @since 7.2
```
slaskawi: 7.2 ? :)

core/src/test/java/org/infinispan/statetransfer/ConcurrentStartChanelLookupTest.java
```diff
@@ -0,0 +1,118 @@
...
+      EmbeddedCacheManager cm2 = createCacheManager(name2, ch2);
+
+      try {
+         assertEquals(ComponentStatus.INSTANTIATED, extractGlobalComponentRegistry(cm1).getStatus());
+         assertEquals(ComponentStatus.INSTANTIATED, extractGlobalComponentRegistry(cm2).getStatus());
+
+         log.debugf("Channels created. Starting the caches");
+         Future<Object> repl1Future = fork(() -> manager(eagerManager).getCache("repl"));
+
+         Thread.sleep(1000);
```
slaskawi: Would it be possible to get rid of this sleep?

core/src/test/java/org/infinispan/statetransfer/ConcurrentStartChanelLookupTest.java
danberindei: Copy-paste :)

core/src/test/java/org/infinispan/statetransfer/ConcurrentStartChanelLookupTest.java
danberindei: It's tricky, because we don't know exactly when JGroups sends the message, and in some cases it can even retransmit messages that were sent right before the join. But I could reduce the timeout :)

There is another option, we could emulate the lost commands by intercepting incoming commands on the second node and discarding the commands there. Still, I wanted this test to "document" how injected channels behave.

tristantarrant:
Pushed to master and 8.2.x. Thanks

PR #4134, opened 2016-03-16
ISPN-6385 Improved stability of TwoNodesTests

slaskawi:
https://issues.jboss.org/browse/ISPN-6385

master and 8.2.x please

### DIFF
--- a/lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/ClusterTestHelper.java
+++ b/lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/ClusterTestHelper.java
@@ -1,6 +1,5 @@
 package org.infinispan.hibernate.search;
 
-import junit.framework.AssertionFailedError;
 import org.hibernate.cfg.Environment;
 import org.hibernate.search.engine.spi.EntityIndexBinding;
 import org.hibernate.search.indexes.spi.DirectoryBasedIndexManager;
@@ -13,6 +12,7 @@ import org.infinispan.remoting.transport.Address;
 
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Helpers to setup several instances of Hibernate Search using clustering to connect the index, and sharing the same H2
@@ -80,22 +80,31 @@ public final class ClusterTestHelper {
    }
 
    /**
-    * Wait some time for the cluster to form
+    * delegates {@link #waitMembersCount(FullTextSessionBuilder, Class, int, long, TimeUnit)} with 10s.
     */
    public static void waitMembersCount(FullTextSessionBuilder node, Class<?> entityType, int expectedSize) {
+      waitMembersCount(node, entityType, expectedSize, 10, TimeUnit.SECONDS);
+   }
+
+   /**
+    * Wait some time for the cluster to form
+    *
+    * @param node Node to be checked
+    * @param entityType An entity for check
+    * @param expectedSize Expected size of the cluster
+    * @param timeout Desired timeout
+    * @param timeoutUnit Timeout units
+     */
+   public static void waitMembersCount(FullTextSessionBuilder node, Class<?> entityType, int expectedSize, long timeout, TimeUnit timeoutUnit) {
+      long endTime = System.currentTimeMillis() + TimeUnit.MILLISECONDS.convert(timeout, timeoutUnit);
       int currentSize = 0;
-      int loopCounter = 0;
-      while (currentSize < expectedSize) {
-         try {
-            Thread.sleep(10);
-         } catch (InterruptedException e) {
-            throw new AssertionFailedError(e.getMessage());
+      do {
+         if (System.currentTimeMillis() > endTime) {
+            throw new AssertionError("Timeout when waiting for desired number of nodes. Expected: " + expectedSize + ", got: " + currentSize);
          }
+         Thread.yield();
          currentSize = clusterSize(node, entityType);
-         if (loopCounter > 200) {
-            throw new AssertionFailedError("timeout while waiting for all nodes to join in cluster");
-         }
-      }
+      } while(currentSize != expectedSize);
    }
 
    /**
--- a/lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/TwoNodesTest.java
+++ b/lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/TwoNodesTest.java
@@ -60,6 +60,7 @@ public class TwoNodesTest {
       } finally {
          nodeC.close();
       }
+      waitMembersCount(nodea, SimpleEmail.class, 2);
       assertEquals(2, clusterSize(nodea, SimpleEmail.class));
       verifyNodeSeesUpdatedIndex(nodea);
       verifyNodeSeesUpdatedIndex(nodeb);


lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/TwoNodesTest.java
```diff
@@ -60,6 +60,7 @@ public void testSomething() {
       } finally {
          nodeC.close();
       }
+      waitMembersCount(nodea, SimpleEmail.class, 2);
```
wburns: I am personally fine with the fix.  However I am thinking we should change the waitMembersCount method to pass in a time out or something though.  It seems like this method could loop forever in a bad case :-1:

slaskawi: This method is based on a counter: https://github.com/infinispan/infinispan/blob/fd9410e09233baaab91861b96d1210bd4c6728ad/lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/ClusterTestHelper.java#L96

However I totally agree - it should be based on time.

wburns: Actually about that if you look closer `loopCounter` is never incremented.  I was originally saying this because a time based solution is better, but if you want just changing it to do a post increment I would be fine with that as well.

Sanne: fix sounds good, but it's a great example of the worst of Infinispan in terms of usability: people starting real nodes on real machines should also invoke such testing helpers?

I shouldn't need to apply such tricks, but to pass some parameter at boot time which expresses how many nodes I'm expecting to have in my cluster: that something that everyone needs.

slaskawi: Comments addressed and rebased.

Sanne:
I would have merged the original one patch :)

Could you split this new patch? "Improving stability of TwoNodesTests" shouldn't change the semantics of the core test utils, I'd track that as an independent improvement.

wburns:
Splitting this into 2 commits seems fine to me.  We can even have another bug to address the issue of the counter never being incremented.  Then if we want I say we can improve the time part later.

gustavonalle:
I'm getting another intermittent error:

```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 7.761 sec <<< FAILURE! - in org.infinispan.hibernate.search.TwoNodesTest
testSomething(org.infinispan.hibernate.search.TwoNodesTest)  Time elapsed: 7.755 sec  <<< FAILURE!
java.lang.AssertionError: Timeout when waiting for desired number of nodes. Expected: 2, got: 0
        at org.infinispan.hibernate.search.ClusterTestHelper.waitMembersCount(ClusterTestHelper.java:103)
        at org.infinispan.hibernate.search.ClusterTestHelper.waitMembersCount(ClusterTestHelper.java:86)
        at org.infinispan.hibernate.search.TwoNodesTest.setUp(TwoNodesTest.java:97)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:24)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:229)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:309)
```

slaskawi:
Could you please review https://github.com/infinispan/infinispan/pull/4147

I created a generic waiting utility and I would like to use it here also. Once it gets integrated, I'll update this PR.

slaskawi:
Change is now divided into 2 parts (stabilizing the test and improving `waitMembersCount`). Everything is rebased.

Sanne:
great, thanks I'm merging it!

lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/ClusterTestHelper.java
```diff
@@ -80,22 +80,31 @@ public static FullTextSessionBuilder createClusterNode(Set<Class<?>> entityTypes
...
-      while (currentSize < expectedSize) {
-         try {
-            Thread.sleep(10);
-         } catch (InterruptedException e) {
-            throw new AssertionFailedError(e.getMessage());
+      do {
+         if (System.currentTimeMillis() > endTime) {
+            throw new AssertionError("Timeout when waiting for desired number of nodes. Expected: " + expectedSize + ", got: " + currentSize);
          }
+         Thread.yield();
```
danberindei: Busy wait for 10 seconds sounds pretty bad. It would be better to sleep for 100ms between iterations, or even better to use `TestingUtil.blockUntilViewReceived()`.

lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/TwoNodesTest.java
```diff
@@ -60,6 +60,7 @@ public void testSomething() {
...
       assertEquals(2, clusterSize(nodea, SimpleEmail.class));
```
danberindei: We already waited for the cluster size to become 2, no need to assert the same thing again.

lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/TwoNodesTest.java
danberindei: @Sanne people starting real nodes on real machines can already use `configuration.transport.initialClusterSize`. But the test kills a node and for some reason needs to check that the view is updated on the other nodes -- I don't see why users would do this.

PR #4135, opened 2016-03-16
ISPN-6275 Avoid invalidating transport twice

galderz:
* On invalidate simply mark transport as invalid instead of invalidating
  it from the pool. When the transport is released, if the transport is
  invalid, it'll get evicted from the pool.


tristantarrant:
Pushed, thanks

PR #4137, opened 2016-03-17
ISPN-6389 Warn when duplicated classes are found

slaskawi:
https://issues.jboss.org/browse/ISPN-6389

Performs an async "fire and forget" check if there are duplicated classes found on the classpath. If so - prints out a warning.


slaskawi:
@tristantarrant Could you please review it?

slaskawi:
rebased

tristantarrant:
It seems a bit overkill. I would make this simpler (and faster), by checking jar manifests against a list of incompatible combinations. This would avoid scanning the whole classpath and just relying on classloader.getResources()

slaskawi:
hmmm neat idea... let me check if that's possible...

slaskawi:
@tristantarrant Implemented as you suggested

slaskawi:
Rebased

slaskawi:
Rebased

tristantarrant:
Pushde to master and 8.2.x

PR #4138, opened 2016-03-17
ISPN-6239 InitialClusterSizeTest.testInitialClusterSizeFail random failures

danberindei:
Increases the timeout in InitialClusterSizeTest and fixes some issues in TEST_PING and JGroupsTransport.

https://issues.jboss.org/browse/ISPN-6239
https://issues.jboss.org/browse/ISPN-6390
https://issues.jboss.org/browse/ISPN-6391

Please cherry-pick to 8.2.x as well.

### DIFF
--- a/core/src/test/java/org/infinispan/remoting/transport/InitialClusterSizeTest.java
+++ b/core/src/test/java/org/infinispan/remoting/transport/InitialClusterSizeTest.java
@@ -1,21 +1,27 @@
 package org.infinispan.remoting.transport;
 
-import static org.testng.AssertJUnit.assertEquals;
-import static org.testng.AssertJUnit.assertTrue;
-
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-
+import org.apache.commons.math.FieldElement;
 import org.infinispan.commons.CacheException;
 import org.infinispan.configuration.cache.CacheMode;
 import org.infinispan.configuration.global.GlobalConfigurationBuilder;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.manager.EmbeddedCacheManagerStartupException;
+import org.infinispan.test.Exceptions;
 import org.infinispan.test.MultipleCacheManagersTest;
 import org.infinispan.test.fwk.CleanupAfterMethod;
 import org.infinispan.test.fwk.TestCacheManagerFactory;
 import org.infinispan.test.fwk.TransportFlags;
 import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.fail;
 
 @Test(testName = "transport.InitialClusterSizeTest", groups = "functional")
 @CleanupAfterMethod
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -237,26 +235,24 @@ public int getViewId() {
 
    @Override
    public void waitForView(int viewId) throws InterruptedException {
-      waitForView(viewId, 0, TimeUnit.SECONDS);
+      waitForView(viewId, Long.MAX_VALUE, TimeUnit.NANOSECONDS);
```
slaskawi: Could we use seconds/milliseconds here? It would be much easier to read...

core/src/test/java/org/infinispan/remoting/transport/InitialClusterSizeTest.java
```diff
@@ -47,22 +53,29 @@ public void testInitialClusterSize() throws ExecutionException, InterruptedExcep
...
+         try {
+            // JGroupsTransport only starts counting down on initialClusterTimeout *after* it connects.
+            // The initial connection may take take 3 seconds (GMS.join_timeout) because of JGRP-2028
+            // Shutdown may also take 2 seconds (GMS.view_ack_collection_timeout) because of JGRP-2030
+            future.get(CLUSTER_TIMEOUT_SECONDS + 5, TimeUnit.SECONDS);
+            fail("Should have thrown an exception");
+         } catch (ExecutionException ee) {
+            Exceptions.assertException(EmbeddedCacheManagerStartupException.class, ee.getCause());
+            Exceptions.assertException(CacheException.class,
+                  org.infinispan.util.concurrent.TimeoutException.class, "ISPN000399:.*",
```
slaskawi: Can we use something else than number here? Those can be problematic in long term maintenance in the product (after merging/cherry-picking commits)

core/src/test/java/org/infinispan/test/Exceptions.java
```diff
@@ -30,7 +30,8 @@ public static void assertException(Class<? extends Throwable> exceptionClass, St
       assertException(exceptionClass, t);
       Pattern pattern = Pattern.compile(messageRegex);
       if (!pattern.matcher(t.getMessage()).matches()) {
-         fail("Wrong exception message: " + t.getMessage());
+         t.printStackTrace();
```
slaskawi: How about throwing something here? This way Surefire will print out the stacktrace.

slaskawi:
Only some minor things... otherwise looks ok to me.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
danberindei: How much easier? If I used seconds, I couldn't use `Long.MAX_VALUE`, because it would be transformed into a negative number of nanoseconds.

core/src/test/java/org/infinispan/remoting/transport/InitialClusterSizeTest.java
danberindei: Does the product use different message ids? Why?

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
wburns: If we are using TimeUnit properly this isn't possible.  Although tbh I am fine with NANOSECONDS anyways.

core/src/test/java/org/infinispan/test/Exceptions.java
danberindei: Forgot about this one, I intended to throw `AssertionError` directly instead of using `fail()`.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
danberindei: @wburns Interesting, I didn't realize `TimeUnit` conversions have overflow protection.

slaskawi: I'm more interested why we need `Long.MAX_VALUE` (which I understand as wait forever right)?

core/src/test/java/org/infinispan/remoting/transport/InitialClusterSizeTest.java
slaskawi: Probably because of a long terms maintenance and different release cycles. I've been fixing those issues in the past, so I would advice being cautious.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
danberindei: It's the same as the `waitForView(0, SECONDS)` we had before, and I'd argue it's easier to read because we don't have to guess that `0` means wait forever.

core/src/test/java/org/infinispan/remoting/transport/InitialClusterSizeTest.java
danberindei: I'd rather fix the product so that it uses the same message ids :)

slaskawi:
Pulling

slaskawi:
Pushed to master branch. `8.2.x` needs a rebase. Would you open another PR just for that?

slaskawi:
Integrated thanks!

PR #4141, opened 2016-03-17
ISPN-3351 Graceful shutdown/restart

tristantarrant:
https://issues.jboss.org/browse/ISPN-3351

What this does:
- it adds a Cache.shutdown() method which performs clustered cache ordered shutdown, by stopping rebalance and shutting down the cache on all nodes
- if (GlobalStatePersistence is enabled) && (cache is persistent) the above also flushes caches to store and persists state (CH)
- on cache startup, if state is present, the initial CH won't be installed until all previous nodes have rejoined
- nodes without state won't be allowed to join clusters with state until the cluster has formed
- nodes with state will never be allowed to join a cluster without state


tristantarrant:
This is now ready to review. Still investigating a pair of failures.

tristantarrant:
I have fixed the failures.

core/src/main/java/org/infinispan/Cache.java
```diff
@@ -334,4 +334,13 @@
     */
    @Override
    void clear();
+
+   /**
+    * Performs a controlled shutdown of the cache. This method differs from {@link #stop()} only in clustered modes,
```
danberindei: It would be nice to know _how_ the method is different from `stop()` for clustered caches...

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -897,14 +902,54 @@ public void start() {
...
+
+   private void performImmediateShutdown() {
+      if (log.isDebugEnabled())
+         log.debugf("Stopping cache %s on %s", getName(), getCacheManager().getAddress());
       componentRegistry.stop();
    }
 
+   public void performGracefulShutdown() {
+      // Perform any orderly shutdown operations here
+      Optional<PassivationManager> passivationManager = componentRegistry.getOptionalComponent(PassivationManager.class);
```
danberindei: The `PassivationManager` doesn't seem optional to me, it's always created by `EmptyConstructorNamedCacheFactory`. True, it may not be created if no other component depends on it, but the `getOptionalComponent()` method name gives the wrong impression. And TBH I like the lambda-less version better anyway :)

core/src/main/java/org/infinispan/globalstate/ScopedPersistentState.java
```diff
@@ -50,4 +50,9 @@
     * Performs the specified action on every entry of the state
     */
    void forEach(BiConsumer<String, String> action);
+
+   /**
+    * Returns the checksum of the common properties (i.e. those not prefixed with @)
```
danberindei: `@version` doesn't seem that un-common

tristantarrant: Sorry, I need to reword that: the @ props are those that need to be excluded from the checksum computation (e.g. timestamp, version)

tristantarrant:
@danberindei I've addressed the comments.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -897,14 +902,55 @@ public void start() {
...
+         description = "Shuts down the cache",
+         displayName = "Shuts down the cache"
+   )
+   public void shutdown() {
+      if (log.isDebugEnabled())
+         log.debugf("Shutting down cache %s on %s", getName(), getCacheManager().getAddress());
+      if (!stopping) {
+         synchronized (this) {
+            // DCL to make sure that only one thread calls stop at one time,
+            // and any other calls by other threads are ignored.
```
danberindei: Why DCL? `shutdown()` calls can't be that frequent.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -897,14 +902,55 @@ public void start() {
...
+            if (!stopping && componentRegistry.getStatus() == ComponentStatus.RUNNING) {
+               stopping = true;
+               requestGracefulShutdown();
+            }
+         }
+      }
+   }
+
+   private void requestGracefulShutdown() {
```
danberindei: Shouldn't this be `requestClusterWideShutdown()`? `performGracefulShutdown()` does something quite different.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
tristantarrant: I followed the same logic used by DefaultCacheManager.stop()

core/src/main/java/org/infinispan/globalstate/impl/ScopedPersistentStateImpl.java
```diff
@@ -61,4 +62,29 @@ public void forEach(BiConsumer<String, String> action) {
...
+   public int hashCode() {
+      int result = scope != null ? scope.hashCode() : 0;
+      result = 31 * result + (state != null ? state.hashCode() : 0);
+      return result;
+   }
+
+   @Override
+   public int getChecksum() {
+      int result = scope != null ? scope.hashCode() : 0;
+      result += state.entrySet().stream().filter(e -> !e.getKey().startsWith("@")).mapToInt(Map.Entry::hashCode).sum();
```
danberindei: Do we want to depend on the `hashCode()` implementation in `Properties`?

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
tristantarrant: Right, better make it clear.

core/src/main/java/org/infinispan/globalstate/impl/ScopedPersistentStateImpl.java
```diff
@@ -15,6 +15,7 @@
 public class ScopedPersistentStateImpl implements ScopedPersistentState {
    private final String scope;
    private final Map<String, String> state;
+   private int checksum;
```
danberindei: Never used

core/src/main/java/org/infinispan/globalstate/impl/ScopedPersistentStateImpl.java
danberindei: Oops, I confused the comment about the file format in `ScopedPersistentState` with the actual implementation. Still, I think it's worth implementing the checksum so it doesn't depend on the `Map` implementation.
Also, we care about the order of properties, but this checksum doesn't detect when two entries are swapped.

tristantarrant: I'm relying on the javadoc for Map.Entry: The hash code of a map entry <tt>e</tt> is defined to be
*     (e.getKey()==null   ? 0 : e.getKey().hashCode()) ^
*     (e.getValue()==null ? 0 : e.getValue().hashCode())

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
danberindei: Nope :)
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/manager/DefaultCacheManager.java#L677

core/src/main/java/org/infinispan/globalstate/impl/ScopedPersistentStateImpl.java
danberindei: Ok, didn't realize that's specced. Do we care about reordered entries?

tristantarrant: No, order doesn't matter. It's just there for readability :)

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
```diff
@@ -120,7 +130,9 @@ public void rebalance(CacheTopology cacheTopology) {
       }, partitionHandlingManager);
 
       if (trace) {
-         log.tracef("StateTransferManager of cache %s on node %s received initial topology %s", cacheName, rpcManager.getAddress(), initialTopology);
+         if (initialTopology != null) {
```
danberindei: This check doesn't seem necessary

tristantarrant: I was trying to avoid a potentially confusing message in the logs, but I guess "initial topology null" is unambiguous

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -172,6 +184,21 @@ private boolean addMember(Address joiner, CacheJoinInfo joinInfo) {
             return false;
          }
 
+         // Validate if the member is allowed to join
+         if (persistentState.isPresent()) {
```
danberindei: So the cache's persistent state must already exist on the coordinator? What if one of the nodes' disk is lost, and that happens to be the first node to start?

danberindei: Doesn't have to be after a cluster restart either, the node could start up in a separate partition and then merge with a running cluster, then become the coordinator of the merged cluster.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -571,22 +598,68 @@ public CacheStatusResponse doJoin(Address joiner, CacheJoinInfo joinInfo) throws
...
-            // This node was the first to join. We need to install the initial CH
-            CacheTopology initialTopology = createInitialCacheTopology();
+         if (status == ComponentStatus.INSTANTIATED) {
+            if (persistentState.isPresent()) {
+               if (trace) log.tracef("Node %s joining. Attempting to reform previous cluster", joiner);
+               // We can only allow this to proceed if we have a complete cluster
+               CacheTopology topology = restoreCacheTopology(persistentState.get(), joinInfo.getPersistentStateChecksum());
+               if (topology != null) {
+                  // Change our status
+                  status = ComponentStatus.RUNNING;
```
danberindei: Shouldn't we keep it `INITIALIZING` until we have all the members?

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: All of the original members need to be part of the new cluster. All of them
persist state locally. If you lose local state then the cluster won't start
and you need to perform manual repair

danberindei: But you can have new nodes, right? And one of those new nodes could become the coordinator.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -571,22 +598,68 @@ public CacheStatusResponse doJoin(Address joiner, CacheJoinInfo joinInfo) throws
...
+                  clusterTopologyManager.broadcastTopologyUpdate(cacheName, topology, availabilityMode, isTotalOrder(), isDistributed());
+                  clusterTopologyManager.broadcastStableTopologyUpdate(cacheName, topology, isTotalOrder(), isDistributed());
```
danberindei: I think you can skip the broadcast, since the cache topology doesn't change for the existing members. In fact, since you only use the persistent state from the coordinator, you could restore the topology in the constructor and do nothing here.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: I have added another test to both Dist and Repl tests which reverses the order in which the cachemanagers are initialized. This means that the view order will be reversed on restart and the coordinator will be different, just to make things a bit more realistic.
The persisted state is identical between all nodes in an existing cluster (just the timestamps will differ) so it doesn't matter who is coordinator and who is a humble member. The conditions I want to avoid are nodes without state joining "stateful" clusters before the initial formation, and nodes with state joining stateless clusters AT ANY TIME (as that might corrupt their local state).
If all conditions are met, and the initial cluster has formed, new (stateless) nodes may join the cluster at will.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: The topology will be != null IFF all members have joined (not less, not more).

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: I have removed the broadcastStableTopologyUpdate invocation. I don't understand the rest of your sentence :)

core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
```diff
@@ -408,7 +412,11 @@ private ClusterCacheStatus initCacheStatusIfAbsent(String cacheName) {
          } else {
             availabilityStrategy = new PreferAvailabilityStrategy(eventLogManager);
          }
-         return new ClusterCacheStatus(cacheName, availabilityStrategy, this, transport);
+         Optional<GlobalStateManager> globalStateManager = cacheManager.getGlobalComponentRegistry().getOptionalComponent(GlobalStateManager.class);
```
danberindei: Why not use injection?

core/src/test/java/org/infinispan/globalstate/AbstractGlobalStateRestartTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+      // Simulate a restart
+      JGroupsAddressCache.flushAddressCaches();
+
+      // Recreate the cluster
+      createStatefulCacheManagers(false, extraneousNodePosition);
+
+      // Healthy cluster
+      switch (extraneousNodePosition) {
+         case -1: {
```
danberindei: These case code could be in the calling methods, which could in turn be moved to the base class.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
danberindei: Ok, if the wrong node happens to start first, the `getCache()` fails but no state/data is lost, and the user can restart the cluster again. I guess it's just a matter of how good our error messages are going to be ;)

I'm a bit more concerned about nodes without state (or just with a different state) becoming coordinator (via merge) while a cluster with state is starting up. E.g. some nodes already sent their join requests to the old coordinator and got the ok to join, but the new coordinator is not going to try and restore that state. I'm not sure how that will play out with the current code, but I think we need some handling of state during cluster recovery, if only to log better error messages.

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
danberindei: Looking at `ClusterCacheStatus.doJoin`, it doesn't look like the topology could be null...

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
danberindei: Ok

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
danberindei: Actually I put the comment on the wrong line, I was looking at the `else` branch, and at the time I was assuming that you're sending the response with the initial topology and with `DEGRADED_MODE` to the first node that joined. 

Now that I've looked closer into `restoreCacheTopology`, I think you do need both broadcasts.

danberindei:
Test failures: http://ci.infinispan.org/viewLog.html?buildId=38274&buildTypeId=bt9&guest=1

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -571,22 +598,67 @@ public CacheStatusResponse doJoin(Address joiner, CacheJoinInfo joinInfo) throws
...
    }
 
+   protected CacheTopology restoreCacheTopology(ScopedPersistentState state, Optional<Integer> persistentStateHash) {
+      if (trace) log.tracef("Attempting to restore CH for cache %s", cacheName);
+
+      try {
+         ConsistentHash persistedCH = joinInfo.getConsistentHashFactory().fromPersistentState(state);
+
+         if (!getExpectedMembers().containsAll(persistedCH.getMembers())) {
+            throw new IllegalStateException();
```
danberindei: Why not `return null`?

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -571,22 +598,67 @@ public CacheStatusResponse doJoin(Address joiner, CacheJoinInfo joinInfo) throws
...
+            throw log.extraneousMembersJoinRestoredCache(extraneousMembers, cacheName);
+         }
+         CacheTopology initialTopology = new CacheTopology(0, 0, persistedCH, null, persistedCH.getMembers());
+         setCurrentTopology(initialTopology);
+         setStableTopology(initialTopology);
+         rebalancingEnabled = true;
+         availabilityMode = AvailabilityMode.AVAILABLE;
+         return initialTopology;
+      } catch (IllegalStateException e) {
+         if (trace) log.tracef("Could not restore CH for cache %s", cacheName);
```
danberindei: Shouldn't this be at least a warning?

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -571,22 +598,67 @@ public CacheStatusResponse doJoin(Address joiner, CacheJoinInfo joinInfo) throws
...
+         topologyBeforeRebalance = getCurrentTopology();
+         // Only trigger availability strategy if we have a topology installed
+         if (topologyBeforeRebalance != null)
+            availabilityStrategy.onJoin(this, joiner);
       }
 
       return new CacheStatusResponse(null, topologyBeforeRebalance, stableTopology, availabilityMode);
    }
 
+   protected CacheTopology restoreCacheTopology(ScopedPersistentState state, Optional<Integer> persistentStateHash) {
```
danberindei: `persistentStateHash` isn't used.

core/src/test/java/org/infinispan/globalstate/AbstractGlobalStateRestartTest.java
```diff
@@ -0,0 +1,212 @@
...
+      this.cacheManagers.clear();
+
+      // Simulate a restart
+      JGroupsAddressCache.flushAddressCaches();
+
+      // Recreate the cluster
+      createStatefulCacheManagers(false, extraneousNodePosition, reverse);
+      if(reverse) {
+         Map<JGroupsAddress, PersistentUUID> reversed = new LinkedHashMap<>();
+         reverseLinkedMap(addressMappings.entrySet().iterator(), reversed);
```
danberindei: Couldn't we rely on the node name instead? Depending on the ordering seems a bit fragile to me.

core/src/test/java/org/infinispan/globalstate/AbstractGlobalStateRestartTest.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.Set;
+import java.util.TreeMap;
+
+import static org.testng.AssertJUnit.*;
+
+public abstract class AbstractGlobalStateRestartTest extends MultipleCacheManagersTest {
+
+   protected abstract int getClusterSize();
+
+   public static int DATA_SIZE = 1000;
```
danberindei: Since this isn't a stress test, I think 100 entries should be more than enough.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: I don't think it's a warning: most of the time this message will be printed as part of the normal join sequence: once for every node join until the original topology/CH can be restored.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: The coordinator won't give the ok until ALL required nodes are in the initial cluster. So if the original cluster had 4 nodes on graceful shutdown, and when restarting they form two two-nodes clusters, neither coordinator will allow the join to proceed. Once the two clusters merge, the new coordinator will have all the required nodes.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: ConsistenthHashFactory.fromPersistentState(state) also throws an IllegalStateException() if it can't reinstate the CH. Mostly this happens because a "PersistentUUID" is not available in the jgroups address cache (i.e. a node hasn't joined the jgroups view yet). So "symmetry" is my driver here :)

core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
tristantarrant: I didn't want to waste a field for storing a component which I only need once to retrieve the persistent state

core/src/test/java/org/infinispan/globalstate/AbstractGlobalStateRestartTest.java
tristantarrant: Right, trace logging storm :)

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: Right

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: Readded

tristantarrant:
@danberindei testsuite should be fixed. I've adopted the JGroupsAddressCache *hack* we discussed in the tests and will come up with a better solution for the persistent uuid cache in a subsequent PR.

tristantarrant:
Need to rework the persistentuuid stuff.

tristantarrant:
@danberindei  this is now ready. The PersistentUUIDManager is there as well as the CH.remapAddresses() API addition which I could not avoid.

danberindei:
@tristantarrant  WriteSkewCacheLoaderFunctionalTest.testPreloadingInTransactionalCache and  SyncReplTest.testReplicateToNonExistentCache are still failing.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
danberindei: Sorry Tristan, but I hate it when we throw exceptions for stuff that we expect to happen *every single time*. Someone who reads this for the first time (like me, every time I read it it's like the first time ;) ) won't suspect that you're expecting another `IllegalStateException` in addition to the one thrown by this method.

tristantarrant:
Found the reason for the failures. CTM was not removing ClusterCacheStatus when all nodes had left. I'll be pushing a fix shortly.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
tristantarrant: Perfectly reasonable. Let me fix it.

tristantarrant:
This is now ready.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
danberindei: I'm just worried that if the merge coordinator doesn't know the persistent uuids, it doesn't have any way to signal that to the nodes that already sent the join request to the old coordinator, so they'll time out instead of reporting an error immediately.

danberindei:
Integrated, thanks Tristan!

As a next step, I think we should add another test to restart the cluster with a merge, both with the merge coordinator one of the initial members, and with it being a new node without persistent state.

PR #4142, opened 2016-03-17
Hot Rod client/server related deprecations/removals

galderz:
This PR does not include server integration removals.


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/logging/LogFactory.java
```diff
@@ -19,20 +18,4 @@ public static Log getLog(Class<?> clazz) {
       return Logger.getMessageLogger(logClass, clazz.getName());
    }
 
-   public static void pushNDC(String cacheName, boolean isTrace) {
-      // Disabled NDC usage until https://issues.jboss.org/browse/JBLOGGING-106 is fixed or
```
wburns: These can actually be added back in now.

galderz: These were not in use before, so agreed to remove them for now.

galderz:
Rebased

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
```diff
@@ -55,7 +55,7 @@
  * @author Mircea.Markus@jboss.com
  * @since 4.1
  */
-public class RemoteCacheManager implements BasicCacheContainer {
```
slaskawi: There is a small dichotomy here... 

`EmbeddedCacheManager` is an interface and for remote one we use `RemoteCacheContainer`. I think it would make sense to rename `RemoteCacheContainer` -> `RemoteCacheManager` and `RemoteCacheManager` -> `Default?RemoteCacheManager`. If you won't do it now - we'll stick to this dichotomy for whole ISPN 9 time frame.

slaskawi:
@galderz When you'll be rebasing - please have a look at my comment :) https://github.com/infinispan/infinispan/pull/4142#discussion_r56683134 

Even though it will cost much more work... I think it will be worth it.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
galderz: Changing an class to become an interface would be problematic for existing users since that'd create a binary compatibility issue which we can work around it. So, although we're free to do the change you suggest, you'd make it more difficult for users to upgrade. So -1 from me.

tristantarrant: -1 from me too

galderz: Btw, we had a very similar issue with org.infinispan.commons.util.FileLookup. It used to be an interface and we let it become a class in 7.0, and in 7.1 (and 7.0.3) we had to revert that back so that other projects could easily use different Infinispan versions without having to recompile them.

galderz:
Rebased :)

tristantarrant:
[ERROR] /home/tst/Work/JBoss/infinispan/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/near/LazyFailoverNearCacheTest.java:[10,48] cannot find symbol
  symbol: class EagerFailoverNearCacheTest

galderz:
File got on the way during yet another rebase...

galderz:
Should be fixed now

tristantarrant:
Pushed, thanks

PR #4145, opened 2016-03-18
ISPN-6370 Expose DMR operation to find out if a cache is being ignored

gustavonalle:
https://issues.jboss.org/browse/ISPN-6370

```
[standalone@localhost:9990 /] /subsystem=datagrid-infinispan-endpoint:is-ignored-
all-endpoints(cache-names=["cache1","cache2"])
{
    "outcome" => "success",
    "result" => {
        "cache1" => true,
        "cache2" => false
    }
}
```


vblagoje:
Great stuff @gustavonalle

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/CacheIgnoreStatusHandler.java
```diff
@@ -0,0 +1,47 @@
...
+import org.jboss.as.controller.registry.Resource;
+import org.jboss.as.controller.registry.Resource.ResourceEntry;
+import org.jboss.dmr.ModelNode;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.infinispan.server.endpoint.subsystem.ModelNodeUtils.contains;
+
+/**
```
slaskawi: Please add some meaningful javadoc or remove it.

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/CacheIgnoreStatusHandler.java
```diff
@@ -0,0 +1,47 @@
...
+ * @author gustavonalle
+ * @since 9.0
+ */
+public class CacheIgnoreStatusHandler implements OperationStepHandler {
+
+   public static final CacheIgnoreStatusHandler INSTANCE = new CacheIgnoreStatusHandler();
```
slaskawi: How about using 1 element enum here?

server/integration/endpoint/src/test/java/org/infinispan/server/endpoint/EndpointSubsystemOperationsTestCase.java
```diff
@@ -50,48 +46,40 @@ private void executeOp(String operationName, String... parameters) {
...
    @Test
    public void testIgnoreCaches() throws Exception {
+      assertCachesNotIgnored("cache1", "cache2", "cache3");
+      assertCachesNotIgnored("whatever");
+
       executeOp("ignore-cache-all-endpoints", "cacheA", "cacheB", "cacheC");
       assertCachesIgnored("cacheA", "cacheB", "cacheC");
 
       executeOp("unignore-cache-all-endpoints", "cacheA");
+      assertCachesNotIgnored("cacheA");
```
slaskawi: Those seem to be separate test cases... I think it might be a good idea to put them into separate methods.

slaskawi:
Only some minor stuff found. Otherwise looks good to me.

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/CacheIgnoreStatusHandler.java
gustavonalle: added

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/CacheIgnoreStatusHandler.java
gustavonalle: What would it be the advantage?

server/integration/endpoint/src/test/java/org/infinispan/server/endpoint/EndpointSubsystemOperationsTestCase.java
gustavonalle: The test case here is manipulating the ignore list, so it needs to use the 3 operations (ignore, unignore and isIgnored) in the same method

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/CacheIgnoreStatusHandler.java
slaskawi: Here is an interesting article about it: http://www.informit.com/articles/article.aspx?p=1216151&seqNum=3

server/integration/endpoint/src/test/java/org/infinispan/server/endpoint/EndpointSubsystemOperationsTestCase.java
slaskawi: Ok, understood.

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/CacheIgnoreStatusHandler.java
danberindei: I never was sure why you'd need to "enforce" the singleton pattern, but if you do, I'd much rather make the default constructor private than make the class an enum.

anistor: +1 for private constructor.

tristantarrant:
Pushed, thanks

PR #4146, opened 2016-03-18
ISPN-6399 Timeout updating the JGroups view after killing one node

danberindei:
Work around JGRP-2028 by disabling view bundling.

https://issues.jboss.org/browse/ISPN-6399


tristantarrant:
Pushed

PR #4147, opened 2016-03-18
ISPN-6401 Improve stability of ClientListenerLeakTest#testNoLeaksAfte…

slaskawi:
https://issues.jboss.org/browse/ISPN-6401

It sometimes happens that the Thread Scheduler has not enough time to kill all the threads (Client event listener threads work in an executor). All we need to do is to give it a bit more time.

I also provided a generic waiting utility.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ClientListenerNotifier.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ClientListenerNotifier.java
@@ -33,6 +33,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -67,7 +68,16 @@ public class ClientListenerNotifier {
    }
 
    public static ClientListenerNotifier create(Codec codec, Marshaller marshaller) {
-      return new ClientListenerNotifier(Executors.newCachedThreadPool(), codec, marshaller);
+      ExecutorService executor = Executors.newCachedThreadPool(getRestoreThreadNameThreadFactory());
+      return new ClientListenerNotifier(executor, codec, marshaller);
+   }
+
+   private static ThreadFactory getRestoreThreadNameThreadFactory() {
+      return r -> new Thread(() -> {
+         final String originalName = Thread.currentThread().getName();
+         r.run();
+         Thread.currentThread().setName(originalName);
+      });
    }
 
    public Marshaller getMarshaller() {


client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerLeakTest.java
```diff
@@ -38,7 +41,10 @@ public void call() {
             eventListener.expectOnlyCreatedEvent(1, cache(cacheName));
          }
       });
-      detectThreadLeaks(".*Client-Listener-ClientListenerLeakTest-.*");
+      WaitFor
+              .condition(() -> listThreadsMatchingRegexp(".*Client-Listener-ClientListenerLeakTest-.*").size() == 0)
+              .atMost(10, TimeUnit.MILLISECONDS)
+              .await();
```
gustavonalle: There's a method already for this on ```AbstractInfinispanTest#eventually```, could it be used here?

wburns: I agree we should use eventually, @danberindei was even nice enough to add some new Java 8 based methods for it too :)

slaskawi: Thanks for the hint! I was grepping code for "wait, waiter..." etc and I missed `eventually` :)

gustavonalle: :smile: I guess the best grep would be ```Thread.sleep```

gustavonalle:
The fix is fine, I cannot seem reproduce the failures anymore

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerLeakTest.java
danberindei: @slaskawi My question is do we need this? Shouldn't the `RemoteCacheManager` shut down the listener threads synchronously?

slaskawi: I believe so...

Those threads are killed [here](https://github.com/infinispan/infinispan/blob/b4f09f9a59dec2d9ee402b857e61d4d280d16e91/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ClientListenerNotifier.java#L216). Occasionally it might happen that the scheduler doesn't have enough time to clean everything out . Adding this small amount of sleep made it stable on my local environment.

Alternative idea would be to wait until all threads are really dead in the `ClientListenerNotifier#stop` method (but I'm not sure if this is a good idea).

gustavonalle: According to https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#shutdown  "This method does not wait for previously submitted tasks to complete execution. Use awaitTermination to do that."

danberindei: @slaskawi, the `stop()` method already waits for 5 seconds, does a 10ms extra wait really make a difference? Or is the way we check for leaked threads different from the way `awaitTermination()` checks if the threads are done, and perhaps we should change that?

anistor: So we wait 5 seconds for the threads to (hopefully) terminate in ClientListenerNotifier.stop. And then we wait 10 milliseconds more before triggering the alarm? I'd rather fix this by interrupting those threads in ClientListenerNotifier.stop.

danberindei: @slaskawi Ok, so here's my take on it: the last worker thread to finish changes the `ThreadPoolExecutor`'s state to `TERMINATED`, then notifies the thread that's blocked in `awaitTermination()`, and then runs a couple more instructions in `processWorkerExit()`. If the worker thread is preempted during those instructions, the thread leak detector will see it as live and the test will fail.

I've been discussing with @anistor and we think the proper way to make the test deterministic is to change `CacheNotifierListener.EventDispatcher.run()` to restore the old thread name in a finally block. Changing `stop()` to use `shutdownNow()` instead of `shutdown()` is optional :)

anistor: I did that here https://github.com/infinispan/jdg/pull/849/files#diff-bb42a970aa86d3c5ddd10e7cbef8460cL236, unfortunately that PR had other issues, and I had to roll it back.

slaskawi: Let me answer your comments before proposing solution:

- Unfortunately not all caught threads are in `TERMINATED` state. Some of them are in `RUNNABLE` state and they were waiting for `Unsafe.unpark` method (at least this happens on my JVM `Java version: 1.8.0_40, vendor: Oracle Corporation`). So the simplest possible trick - adding `Thread.isAlive` to `if (thread.getName().matches(regexp))` didn't work. 
- I think `shutdown()` is fine there... I noticed that the test doesn't wait for terminating clients. It's more about the method that we use to detect leaked threads. 

Looking at your proposed changes, I have implemented another fix. In order to decouple reverting the original thread name and the work to be done - I used `ThreadFactory` interface and passed it to the executor.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ClientListenerNotifier.java
```diff
@@ -67,7 +68,16 @@ protected ClientListenerNotifier(ExecutorService executor, Codec codec, Marshall
...
-      return new ClientListenerNotifier(Executors.newCachedThreadPool(), codec, marshaller);
+      ExecutorService executor = Executors.newCachedThreadPool(getRestoreThreadNameThreadFactory());
+      return new ClientListenerNotifier(executor, codec, marshaller);
+   }
+
+   private static ThreadFactory getRestoreThreadNameThreadFactory() {
+      return r -> new Thread(() -> {
+         final String originalName = Thread.currentThread().getName();
+         r.run();
+         Thread.currentThread().setName(originalName);
```
anistor: no ```finally``` for this?

slaskawi: Good point. Thanks for spotting this

anistor: Overriding the default thread factory looses some of the initial behaviour. The original TF ensures the thread priority is normal instead of being inherited from the caller. Do we care?

slaskawi: I would say no, but let's wait a bit for some other opinions...

danberindei: None of the other thread factories in Infinispan seem to care about priorities, so I think we can ignore it here as well.

anistor:
Added ```finally``` and pulled in master and 8.2.x. thanks @slaskawi !

PR #4149, opened 2016-03-18
Fix some checkstyle warnings

anistor:


server/integration/testsuite/src/test/java/org/infinispan/server/test/client/hotrod/HotRodCustomMarshallerEventIT.java
```diff
@@ -169,9 +169,8 @@ public boolean equals(Object o) {
 
             Id id1 = (Id) o;
 
-            if (id != id1.id) return false;
+            return id == id1.id;
```
tristantarrant: this one is amazing :)

anistor: Indeed. Fixing these is just as satisfying and addictive as popping bubble wrap.

slaskawi:
Pushed thanks!

PR #4153, opened 2016-03-21
ISPN-6412 Stabilize RESTReplicationIT#* tests

slaskawi:
https://issues.jboss.org/browse/ISPN-6412

Let me run a couple of private builds before integrating this one. This test was failing because it needs a synced, replicated cache to work on.

To master and 8.2.x please

### DIFF
--- a/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTHelper.java
+++ b/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTHelper.java
@@ -46,6 +46,7 @@ public class RESTHelper {
     public static final String KEY_C = "c";
 
     private static final String DATE_PATTERN_RFC1123 = "EEE, dd MMM yyyy HH:mm:ss zzz";
+    public static final String CACHE_NAME = "testcache";
 
     private static int port = 8080;
     private static List<Server> servers = new ArrayList<Server>();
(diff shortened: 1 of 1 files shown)


danberindei:
@slaskawi `RESTReplicationIT(replmode-udp).testReplicationPost` is still failing:
http://ci.infinispan.org/viewLog.html?buildId=37693&buildTypeId=bt9&guest=1

TBH I couldn't figure out what cache definitions are being used by this test, so I don't know if your changes would help. But I'm guessing that the default cache is also replicated/distributed, otherwise the test would never pass.

slaskawi:
Hmmmm this is a tough one... I can't figure out what's wrong with it... Locally everything is OK...

slaskawi:
@mgencur Maybe you have any idea how to debug this?

mgencur:
I was trying to reproduce it locally with no success. Maybe we could enable trace logging in TeamCity for a while until this happens again? Not sure what else to do except turn on trace logging and see what happened. This is a random failure.

danberindei:
@mgencur @slaskawi There is already a build with TRACE logging enabled in TeamCity, but it's much slower than the regular build and it also doesn't work for the server logs. 

If `-Dtrace=org.infinispan` works for this test, you can add it to the `maven_opts.log.debug` parameter in the Infinispan project settings. You can also run ad-hoc builds with the `Run|...` button and modify the parameter there.

danberindei:
@mgencur @slaskawi Is there any logging that we could add to the test itself to help debugging?

slaskawi:
Unfortunately I didn't manage to find anything useful... This issue seems to be really weird. Closing this PR.

PR #4155, opened 2016-03-21
ISPN-6414 Hybrid query + compat mode applies type conversion too earl…

anistor:
…y (perf issue)

* also change indexing of protobuf's bools to string instead of integer

Jira: https://issues.jboss.org/browse/ISPN-6414
        https://issues.jboss.org/browse/ISPN-6416

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/QueryOperation.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/QueryOperation.java
@@ -86,7 +86,7 @@ public class QueryOperation extends RetryOnFailureOperation<QueryResponse> {
             EnumMarshaller encoder = (EnumMarshaller) remoteQuery.getSerializationContext().getMarshaller(value.getClass());
             value = encoder.encode((Enum) value);
          } else if (value instanceof Boolean) {
-            value = (Boolean) value ? 1 : 0;
+            value = value.toString();
          } else if (value instanceof Date) {
             value = ((Date) value).getTime();
          }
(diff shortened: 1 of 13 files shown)


anistor:
rebased.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/QueryOperation.java
```diff
@@ -86,7 +86,7 @@ protected QueryResponse executeOperation(Transport transport) {
             EnumMarshaller encoder = (EnumMarshaller) remoteQuery.getSerializationContext().getMarshaller(value.getClass());
             value = encoder.encode((Enum) value);
          } else if (value instanceof Boolean) {
-            value = (Boolean) value ? 1 : 0;
+            value = value.toString();
```
tristantarrant: What's the benefit ?

anistor: "Booleans values used to be translated to 0/1 literals internally by the DSL builder in remote mode so the underlying Jpql query would differ for the same query in embedded vs remote mode which is not ideal."

gustavonalle: From Lucene perspective, having indexing a boolean as two different terms instead of a number is a better choice. Indexing it as a number implies storing the field as a trie structure, and it requires more work to query it.

anistor: That too. And since it impacts index format I wanted to do it early. Right now it does not matter much, but it will if we decide later to expose the jpql string. We should ensure the same query string works both in embedded and remote.
ATM we still have enum and date which are handled in a particular way for protobuf ...

tristantarrant:
Pushed to master, thanks

PR #4157, opened 2016-03-21
[ISPN-6415] Allow duplicate JMX domains in JPA cache store testsuite

Holmistr:
See: https://issues.jboss.org/browse/ISPN-6415


tristantarrant:
Pushed, thanks

PR #4160, opened 2016-03-22
ISPN-6405 Persistence configuration with clustered cache can cause

wburns:
duplicate expiration messages

* Fixed issue where reaper caused 2 expirations in cluster cache

https://issues.jboss.org/browse/ISPN-6405

### DIFF
--- a/core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
+++ b/core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
@@ -118,8 +118,8 @@ public class RemoveExpiredCommand extends RemoveCommand {
    public String toString() {
       return "RemoveExpiredCommand{" +
               "key=" + key +
-              "value=" + value +
-              "lifespan=" + lifespan +
+              ", value=" + value +
+              ", lifespan=" + lifespan +
               '}';
    }
 
--- a/core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
+++ b/core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
@@ -128,7 +128,12 @@ public class RemoveExpiredCommand extends RemoveCommand {
       output.writeObject(commandInvocationId);
       output.writeObject(key);
       output.writeObject(value);
-      output.writeLong(lifespan);
+      if (lifespan != null) {
+         output.writeBoolean(true);
+         output.writeLong(lifespan);
+      } else {
+         output.writeBoolean(false);
+      }
    }
 
    @Override
@@ -136,7 +141,12 @@ public class RemoveExpiredCommand extends RemoveCommand {
       commandInvocationId = (CommandInvocationId) input.readObject();
       key = input.readObject();
       value = input.readObject();
-      lifespan = input.readLong();
+      boolean lifespanProvided = input.readBoolean();
+      if (lifespanProvided) {
+         lifespan = input.readLong();
+      } else {
+         lifespan = null;
+      }
    }
 
    @Override
--- a/core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
+++ b/core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
@@ -659,48 +659,42 @@ public class SingleFileStore<K, V> implements AdvancedLoadWriteStore<K, V> {
    
    @Override
    public void purge(Executor threadPool, final PurgeListener task) {
-
-      threadPool.execute(new Runnable() {
-         @Override
-         public void run() {
-            long now = timeService.wallClockTime();
-            List<KeyValuePair<Object, FileEntry>> entriesToPurge = new ArrayList<KeyValuePair<Object, FileEntry>>();
-            synchronized (entries) {
-               for (Iterator<Map.Entry<K, FileEntry>> it = entries.entrySet().iterator(); it.hasNext(); ) {
-                  Map.Entry<K, FileEntry> next = it.next();
-                  FileEntry fe = next.getValue();
-                  if (fe.isExpired(now)) {
-                     it.remove();
-                     entriesToPurge.add(new KeyValuePair<Object, FileEntry>(next.getKey(), fe));
-                  }
-               }
-            }
-
-            resizeLock.readLock().lock();
-            try {
-               for (Iterator<KeyValuePair<Object, FileEntry>> it = entriesToPurge.iterator(); it.hasNext(); ) {
-                  KeyValuePair<Object, FileEntry> next = it.next();
-                  FileEntry fe = next.getValue();
-                  if (fe.isExpired(now)) {
-                     it.remove();
-                     try {
-                        free(fe);
-                     } catch (Exception e) {
-                        throw new PersistenceException(e);
-                     }
-                     if (task != null) task.entryPurged(next.getKey());
-                  }
-               }
-               
-               // Disk space optimizations
-               synchronized (freeList) {
-                 processFreeEntries();
-               }
-            } finally {
-               resizeLock.readLock().unlock();
+      long now = timeService.wallClockTime();
+      List<KeyValuePair<Object, FileEntry>> entriesToPurge = new ArrayList<KeyValuePair<Object, FileEntry>>();
+      synchronized (entries) {
+         for (Iterator<Map.Entry<K, FileEntry>> it = entries.entrySet().iterator(); it.hasNext(); ) {
+            Map.Entry<K, FileEntry> next = it.next();
+            FileEntry fe = next.getValue();
+            if (fe.isExpired(now)) {
+               it.remove();
+               entriesToPurge.add(new KeyValuePair<Object, FileEntry>(next.getKey(), fe));
             }
          }
-      });
+      }
+
+      resizeLock.readLock().lock();
+      try {
+         for (Iterator<KeyValuePair<Object, FileEntry>> it = entriesToPurge.iterator(); it.hasNext(); ) {
+            KeyValuePair<Object, FileEntry> next = it.next();
+            FileEntry fe = next.getValue();
+            if (fe.isExpired(now)) {
+               it.remove();
+               try {
+                  free(fe);
+               } catch (Exception e) {
+                  throw new PersistenceException(e);
+               }
+               if (task != null) task.entryPurged(next.getKey());
+            }
+         }
+
+         // Disk space optimizations
+         synchronized (freeList) {
+           processFreeEntries();
+         }
+      } finally {
+         resizeLock.readLock().unlock();
+      }
    }
 
    @Override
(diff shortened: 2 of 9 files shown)


slaskawi:
Looks good to me

slaskawi:
Could you please rebase it? I would like to make sure those failures will go away...

wburns:
Rebased

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
```diff
@@ -118,8 +118,8 @@ public byte getCommandId() {
    public String toString() {
       return "RemoveExpiredCommand{" +
               "key=" + key +
-              "value=" + value +
-              "lifespan=" + lifespan +
+              " value=" + value +
+              " lifespan=" + lifespan +
```
danberindei: Might as well add a comma

wburns: Sure

core/src/test/java/org/infinispan/expiration/impl/ExpirationSingleFileStoreDistListenerFunctionalTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.testng.annotations.Test;
+
+@Test(groups = "functional", testName = "expiration.impl.ExpirationSingleFileStoreDistListenerFunctionalTest")
+public class ExpirationSingleFileStoreDistListenerFunctionalTest extends ExpirationStoreListenerFunctionalTest {
+   @Override
+   protected void configure(ConfigurationBuilder config) {
+      config
+              // We set the wakeup interval to max value.  If we have a store we force the reaper to be enabled, but
+              // for this test we can get false positives if the reaper runs at the same time as a manual expiration
+              // process
```
danberindei: How about "Prevent the reaper from running, reaperEnabled(false) doesn't work when a store is present"?

wburns: okay

wburns:
Updated with the comments @danberindei.  I didn't squash the commit so you can see the diff easily.  Still needs to be fixed up on a rebase.

core/src/test/java/org/infinispan/expiration/impl/ExpirationStoreListenerFunctionalTest.java
```diff
@@ -56,18 +57,20 @@ public void testExpirationOfStoreWhenDataNotInMemory() throws Exception {
...
-      assertNotNull(event.getMetadata());
+      if (TestingUtil.getCacheLoader(cache) instanceof AdvancedCacheExpirationWriter) {
+         assertEquals("v", event.getValue());
+         assertNotNull(event.getMetadata());
+      }
    }
 
    private void assertExpiredEvents(int count) {
-      assertEquals(count, listener.getInvocationCount());
+      eventually("Last count was :" + listener.getInvocationCount(), () -> count == listener.getInvocationCount());
```
danberindei: We have `eventuallyEquals` for this

wburns: I didn't realize the trace message was what I wanted. Thanks.

wburns:
Comments address, still will need to be rebased before integration.

wburns:
Something changed that is causing this to fail now, not sure what since nothing has changed in this PR by itself that would have caused it, looking..

wburns:
Looks like the failures is caused by https://github.com/infinispan/infinispan/commit/d60bdc6ad4cb62772dade4b12afd92cddd0a6af0#diff-719e44d55c697435cc2ae65da78d4a35 since the purge is no longer synchronous.

wburns:
I don't know how I didn't see the failures in my tests before, but it is caused due to the SingleFileStore purging entries asynchronously.  And to be honest there is no way to really fix this in a nice way, besides tests.

The API doesn't define a Future or anything for the caller to return to block to wait on.  This would mean that could theoretically get a pile up of expiration threads if there was a low reaper wait time and an expiration for a cache store that would take a while.

wburns:
Will just make the calling thread block until the expiration thread is completed.  Unfortunately this means we are using a thread for nothing, while it could have just been processing the expired entries directly.

danberindei:
@wburns I see some of the stores use a `ExecutorCompletionService` to make sure the purge is done by the time the method returns, we could do the same in `SingleFileStore`. And we should make it clear in the `purge()` javadoc that we expect it to be finished when it returns.

We could enforce this by wrapping the persistence executor and counting started/finished tasks in the wrapper, but I'm not sure if we should do it in production or just for these tests.

wburns:
@danberindei That is what I ended up doing, re: the `ExecutorAllCompletionService`.  I will add the javadoc mention though.

wburns:
Added another commit that should fix that all up, let me know what you think @danberindei

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -663,47 +666,53 @@ private void mergeFreeEntries(List<FileEntry> entries) {
...
+               KeyValuePair<Object, FileEntry> next = it.next();
+               FileEntry fe = next.getValue();
+               if (fe.isExpired(now)) {
+                  it.remove();
+                  try {
+                     free(fe);
+                  } catch (Exception e) {
+                     throw new PersistenceException(e);
                   }
+                  if (task != null) task.entryPurged(next.getKey());
```
danberindei: I'd rather run only the PurgeListener on the Executor.

wburns: I had thought about that.  My worry is that the listener may not do anything, so it would end up causing a context switch with no benefit.  I am thinking I can submit them in batches using a parallel stream though.  Check it out after I resubmit to let me know what you think.

wburns:
Updated to use persistence thread pool for the notifications and rebased.

core/src/main/java/org/infinispan/executors/ParallelExecutorHelper.java
```diff
@@ -0,0 +1,59 @@
...
+    * @param executor the executor to submit the tasks to
+    * @param <T> the type of object being processed
+    * @return future that signifies completion of all submitted tasks
+    */
+   public static <T> CompletableFuture<Void> forEach(Spliterator<? extends T> spliterator,
+           Consumer<? super T> consumer, Executor executor) {
+      long size = spliterator.getExactSizeIfKnown();
+      // If the size is unknown we use the builder method to prevent multiple resizings
+      if (size == -1) {
+         // Use stream builder as it requires less resizing than ArrayList if the spliterator is large
```
danberindei: I don't think keeping all the entries/tasks in memory is a good idea anyway, even if it requires less resizing.

danberindei: I don't think keeping all the entries/tasks in memory is a good idea anyway, even if it requires less resizing.
That's the reasoning behind `ExecutorAllCompletionService`, to avoid keeping all the futures in memory.

core/src/main/java/org/infinispan/executors/ParallelExecutorHelper.java
```diff
@@ -0,0 +1,59 @@
...
+         List<CompletableFuture<Void>> list = new ArrayList<>((int) size);
+         list.add(forEach(spliterator, consumer, executor, list::add));
+         return CompletableFuture.allOf((CompletableFuture<Void>[]) list.toArray(new CompletableFuture[list.size()]));
+      }
+   }
+
+   private static <T> CompletableFuture<Void> forEach(Spliterator<? extends T> spliterator,
+           Consumer<? super T> consumer, Executor executor, Consumer<CompletableFuture<Void>> builder) {
+      Spliterator<? extends T> split;
+      while ((split = spliterator.trySplit()) != null) {
```
danberindei: Shouldn't you stop splitting when the spliterator reaches a certain size to batch consumer calls? Looking at the `ArrayListSpliterator` code, it only stops splitting when it has a single element.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -35,33 +37,33 @@
...
+ * <p>
+ * So, the space taken by this cache store is both the space in the file itself plus the in-memory index with the keys
+ * and their file positions. With this in mind and to avoid the cache store leading to OutOfMemoryExceptions, you can
+ * optionally configure the maximum number of entries to maintain in this cache store, which affects both the size of
+ * the file and the size of the in-memory index. However, setting this maximum limit results in older entries in the
+ * cache store to be eliminated, and hence, it only makes sense configuring a maximum limit if Infinispan is used as a
+ * cache where loss of data in the cache store does not lead to data loss, and data can be recomputed or re-queried from
+ * the original data source.
+ * <p>
+ * This class is fully thread safe, yet allows for concurrent load / store of individual cache entries.
```
danberindei: Did you actually change anything here?

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -219,8 +220,8 @@ private void rebuildIndex() throws Exception {
    }
 
    /**
-    * The base class implementation calls {@link #load(Object)} for this, we can do better because
-    * we keep all keys in memory.
+    * The base class implementation calls {@link #load(Object)} for this, we can do better because we keep all keys in
+    * memory.
```
danberindei: Better configure IntelliJ to only reformat changed lines :)

wburns: Yeah I noticed this too late. I can fix this and put it in another commit.

core/src/main/java/org/infinispan/executors/ParallelExecutorHelper.java
wburns: Hrmm I must be misremembering how the spliterators are implemented. I will
take a closer look and remedy.

core/src/main/java/org/infinispan/executors/ParallelExecutorHelper.java
wburns: Hrmm, when the `ExecutorAllCompletionService` was added it was more about not having to keep track of # of executed tasks in the user code.  The futures not being retained for the duration was just a side effect.

And the comment here is not talking about ExecutorAllCompletionService but rather the fact it is using a Stream.Builder instead of an ArrayList.

I can try to convert this over to use the EACS, but I am not sure if there is a nice way to also return a `CompletableFuture` as well since I won't have a thread to complete the Future when all tasks are completed since there is no guarantee that a user thread is available to check it.

wburns:
Okay I have just changed it to not use the thread pool with the single file store for now.  Also I added in ISPN-6071 since it needs the new test that I have with a slight change.

wburns:
We can cherry pick this into 8.2.x, but I will make a separate PR after this is integrated for 8.1.x for ISPN-6071.

danberindei:
@wburns Some new test failures:
http://ci.infinispan.org/viewLog.html?buildId=38345&buildTypeId=bt9&guest=1

wburns:
@danberindei I have rebased and I haven't been able to reproduce this issue at all yet.

tristantarrant:
Pushed, thanks. Can we have an 8.2.x fix please ?

PR #4162, opened 2016-03-22
ISPN-6424 Distributed Executor get can propagate exception if interrupt

wburns:
returns before cancellation

* Make sure we only send interrupt if we are able to interrupt locally

https://issues.jboss.org/browse/ISPN-6424

### DIFF
--- a/core/src/main/java/org/infinispan/distexec/DefaultExecutorService.java
+++ b/core/src/main/java/org/infinispan/distexec/DefaultExecutorService.java
@@ -981,10 +981,10 @@ public class DefaultExecutorService extends AbstractExecutorService implements D
 
       @Override
       public boolean cancel(boolean mayInterruptIfRunning) {
-         if (!isCancelled()) {
+         if (!isCancelled() && super.cancel(true)) {
             CancelCommand ccc = factory.buildCancelCommandCommand(distCommand.getUUID());
             rpc.invokeRemotely(Collections.singletonList(getExecutionTarget()), ccc, rpc.getDefaultRpcOptions(true));
-            return super.cancel(true);
+            return true;
          } else {
             //already cancelled
             return false;
@@ -1023,7 +1023,7 @@ public class DefaultExecutorService extends AbstractExecutorService implements D
 
       @Override
       public synchronized boolean cancel(boolean mayInterruptIfRunning) {
-         if (!isCancelled()) {
+         if (!isCancelled() && super.cancel(true)) {
             CancelCommand ccc = factory.buildCancelCommandCommand(distCommand.getUUID());
             ccc.init(cancellationService);
             try {
@@ -1031,7 +1031,7 @@ public class DefaultExecutorService extends AbstractExecutorService implements D
             } catch (Throwable e) {
                log.couldNotExecuteCancellationLocally(e.getLocalizedMessage());
             }
-            return super.cancel(true);
+            return true;
          } else {
             //already cancelled
             return false;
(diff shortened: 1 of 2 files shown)


danberindei:
@wburns `CompletableFuture.cancel(true)` actually ignores the `mayInterruptIfRunning` parameter, because it doesn't have an associated thread to interrupt.

wburns:
@danberindei you are saying just to change it from ``super.cancel(true);`` to ``super.cancel(false);``?  Cause it doesn't matter either way as you just mentioned.

Or are you saying we shouldn't use ``CompletableFuture`` for this at all since the javadoc says this?  Or maybe we should add a mention to the ``DistributedExecutorService``.

I personally think just adding the javadoc would be best.

danberindei:
@wburns I was thinking we could keep track of the thread ourselves, and use `Thread.interrupt()`. But since we now return a `CompletableFuture`, and users don't (or shouldn't) expect `CompletableFuture.cancel(true)` to do anything, maybe we should give up on interrupting threads altogether :)

wburns:
Added documentation regarding interruption and rebased.

core/src/main/java/org/infinispan/distexec/DistributedExecutorService.java
```diff
@@ -25,12 +25,18 @@
...
- *
  * Note that due to potential task migration to other nodes every {@link Callable},
  * {@link Runnable} and/or {@link DistributedCallable} submitted must be either {@link Serializable}
  * or {@link Externalizable}. Also the value returned from a callable must be {@link Serializable}
  * or {@link Externalizable}. Unfortunately if the value returned is not serializable then a
  * {@link NotSerializableException} will be thrown.
+ * <p>
+ *
+ * All {@link CompletableFuture} returned to the caller may be cancelled normally, however if interruption is desired
+ * to stop an in process task only the initial <b>CompletableFuture</b> may be used.  Any chained
```
danberindei: I wouldn't say anything about "normal" cancellation here, just that calling with `mayInterruptIfRunning == true` will interrupt the task on all the nodes where it is running, and link to `Thread#interrupt()`. Possibly also mention what  exception the user should expect when the task is cancelled, and that cancelling dependent `CompletableFutures` will not cancel this future (e.g. from `CompletableFuture.allOf()`).

core/src/main/java/org/infinispan/distexec/DefaultExecutorService.java
```diff
@@ -980,10 +980,10 @@ public void execute() {
 
       @Override
       public boolean cancel(boolean mayInterruptIfRunning) {
-         if (!isCancelled()) {
+         if (!isCancelled() && super.cancel(true)) {
```
danberindei: I think you can remove `isCancelled()`.

wburns: Unfortuantely `super.cancel(true)` will return true even if called the second time.  It is a weird behavior change from regular Future I know :-1:

core/src/main/java/org/infinispan/distexec/DistributedExecutorService.java
danberindei: Thinking about it some more, "dependent CF" probably isn't super clear either, although an example would probably help. But the distinction between "the initial CF" and "all CFs returned" is too murky, especially for methods that return a collection of CFs.

wburns:
I changed the documentation a bit, unfortunately the other comment needs to stay as is (inconsistent behaviors).

danberindei:
Integrated, thanks Will!

PR #4165, opened 2016-03-22
ISPN-6356 Wrap removal of default HR liteners into security actions

vjuranek:
https://issues.jboss.org/browse/ISPN-6356

### DIFF
--- a/core/src/main/java/org/infinispan/security/actions/RemoveListenerAction.java
+++ b/core/src/main/java/org/infinispan/security/actions/RemoveListenerAction.java
@@ -1,0 +1,29 @@
+package org.infinispan.security.actions;
+
+import java.security.PrivilegedAction;
+
+import org.infinispan.notifications.Listenable;
+
+/**
+ * RemoveListenerAction
+ *
+ * @author vjuranek
+ * @since 9.0
+ */
+public class RemoveListenerAction implements PrivilegedAction<Void> {
+
+    private final Listenable listenable;
+    private final Object listener;
+
+    public RemoveListenerAction(Listenable listenable, Object listener) {
+        this.listenable = listenable;
+        this.listener = listener;
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/security/actions/RemoveListenerAction.java
```diff
@@ -0,0 +1,29 @@
...
+    private final Object listener;
+
+    public RemoveListenerAction(Listenable listenable, Object listener) {
+        this.listenable = listenable;
+        this.listener = listener;
+    }
+
+    @Override
+    public Void run() {
+        listenable.removeListener(listener);
```
slaskawi: Perhaps we could add some not null checks here? I think it would be the best to ignore this instruction when `listenable` is null.

slaskawi:
I think it would be good to have some test for it... What do you think?

vjuranek:
> I think it would be good to have some test for it

test for what actually? The functionality is basically tested in many HR server tests (these listeners are added automatically) and this PR just removes warning from log - not sure how to test it.

core/src/main/java/org/infinispan/security/actions/RemoveListenerAction.java
vjuranek: instead of doing nothing when `listenable` is `null`, we should at least log it, as there's probably something wrong, `null` is definitely expected here. Looking into other security actions, non of them contains logging or other checks - I guess it's because of performance, as these actions can be used quite frequently - is this correct @tristantarrant ?

If I'm wrong and there are no performance concerns, no problem to add it.

slaskawi:
I took another look at the stacktrace in the JIRA and I think I now understand what's going on... integrating into both - master and 8.2.x

slaskawi:
Ok, it has been integrated previously... Thanks Vojtech!

PR #4171, opened 2016-03-23
ISPN-6429 Duplicated boolean projections are broken in compat mode

anistor:
https://issues.jboss.org/browse/ISPN-6429

### DIFF
--- a/server/integration/feature-pack/src/main/resources/modules/system/layers/base/org/infinispan/objectfilter/main/module.xml
+++ b/server/integration/feature-pack/src/main/resources/modules/system/layers/base/org/infinispan/objectfilter/main/module.xml
@@ -9,7 +9,6 @@
       <module name="org.infinispan.query.dsl"/>
       <module name="org.antlr.antlr-runtime"/>
       <module name="org.hibernate.hql.parser"/>
-      <module name="org.antlr.antlr-runtime"/>
       <module name="org.infinispan.protostream"/>
       <module name="org.jboss.logging" />
    </dependencies>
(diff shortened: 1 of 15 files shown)


server/integration/feature-pack/src/main/resources/modules/system/layers/base/org/infinispan/objectfilter/main/module.xml
```diff
@@ -9,7 +9,6 @@
       <module name="org.infinispan.query.dsl"/>
       <module name="org.antlr.antlr-runtime"/>
       <module name="org.hibernate.hql.parser"/>
-      <module name="org.antlr.antlr-runtime"/>
```
slaskawi: Is this intentional? If so, perhaps we could delete `antlr` from some WF modules (`as-modules/*`)?

slaskawi:
The changes look good to me. You could squash all the commits though (but it's a minor thing)

server/integration/feature-pack/src/main/resources/modules/system/layers/base/org/infinispan/objectfilter/main/module.xml
anistor: We cannot delete antlr, unfortunatelly. But here it was imported twice so I removed one.

anistor:
I'll squash the two commits related to ISPN-6429. I initially chose to separate the test and the fix so the reviewer can easily revert the fix and see the test failing.

server/integration/feature-pack/src/main/resources/modules/system/layers/base/org/infinispan/objectfilter/main/module.xml
slaskawi: ooh yeah - I'm blind... sorry

slaskawi:
Ok, great! Thanks!

anistor:
Updated.

slaskawi:
pushed thanks!

PR #4177, opened 2016-03-23
ISPN-6432 Remove deprecated MurmurHash2

danberindei:
https://issues.jboss.org/browse/ISPN-6432

Also remove the client ConsistentHashV1 that was using it,
but was not used itself since 5.0.


tristantarrant:
Pushed to master, thanks

PR #4178, opened 2016-03-23
ISPN-6434 Remove namedCache from clustered config test

vjuranek:
https://issues.jboss.org/browse/ISPN-6434

### DIFF
--- a/server/integration/testsuite/src/test/java/org/infinispan/server/test/configs/ExampleConfigsIT.java
+++ b/server/integration/testsuite/src/test/java/org/infinispan/server/test/configs/ExampleConfigsIT.java
@@ -577,12 +577,10 @@ public class ExampleConfigsIT {
         delete(fullPathKey(KEY_A));
         delete(fullPathKey(KEY_B));
         delete(fullPathKey(KEY_C));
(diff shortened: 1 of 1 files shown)


slaskawi:
Those [2 failures](http://ci.infinispan.org/viewLog.html?buildId=37887&buildTypeId=bt9&guest=1) seems unrelated but retriggering CI to be sure.

slaskawi:
The change looks good to me

danberindei:
Integrated, thanks Vojtech!

However, the test wasn't failing, so I'm wondering how useful it is...

vjuranek:
@danberindei It's an issue for JDG, see [JDG-124](https://issues.jboss.org/browse/JDG-124). In ISPN it's not an issue as it still has older dependency on RestEasy (better to say this is just my suspicion which I didn't investigate much). I hope dependency update in ISPN will be done correctly and this issue won't pop up, but anyway, test delete from non-exiting cache in this place is confusing (at least for me :smile: ). I moved test for delete from non-existing place into (IMHO) better place as part of PR #4179.

danberindei:
@vjuranek You're right, it was confusing to me as well :)

PR #4183, opened 2016-03-25
ISPN-6439 Stabilize ExampleConfigsIT#testLevelDBCacheStoreConfig

slaskawi:
https://issues.jboss.org/browse/ISPN-6439

Now it uses the same key references for validation

To both master and `8.2.x` please

### DIFF
--- a/server/integration/testsuite/src/test/java/org/infinispan/server/test/configs/ExampleConfigsIT.java
+++ b/server/integration/testsuite/src/test/java/org/infinispan/server/test/configs/ExampleConfigsIT.java
@@ -1,15 +1,5 @@
 package org.infinispan.server.test.configs;
 
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.ObjectInputStream;
-import java.net.Inet6Address;
-import java.util.ArrayList;
-import java.util.List;
-
-import javax.management.ObjectName;
-import javax.servlet.http.HttpServletResponse;
-
 import org.apache.http.HttpResponse;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPut;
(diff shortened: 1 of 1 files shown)


server/integration/testsuite/src/test/java/org/infinispan/server/test/configs/ExampleConfigsIT.java
```diff
@@ -653,13 +668,17 @@ private void doPutGetWithExpiration(RemoteInfinispanMBeans s1, RemoteInfinispanM
...
-            assertEquals("v" + i, c.get("k" + i));
-        }
-        assertEquals(10, s.cache.getNumberOfEntries());
+
+        final int maxRange = 10;
+        Map<String, String> testMap = new HashMap<>();
+        IntStream.range(0, maxRange)
+                .forEach(number -> testMap.put("k" + number, "v" + number));
+
+        c.putAll(testMap);
```
danberindei: Should the test method be renamed, if it's doing `putAll` instead of `put`?

server/integration/testsuite/src/test/java/org/infinispan/server/test/configs/ExampleConfigsIT.java
```diff
@@ -653,13 +668,17 @@ private void doPutGetWithExpiration(RemoteInfinispanMBeans s1, RemoteInfinispanM
...
+
+        assertEquals(maxRange, s.cache.getNumberOfEntries());
+        c.retrieveEntries(null, maxRange)
+                .forEachRemaining((e) -> assertTrue(testMap.containsValue(e.getValue())));
```
danberindei: I'd rather use `assertEquals(e.getValue(), testMap.get(e.getKey())`.

server/integration/testsuite/src/test/java/org/infinispan/server/test/configs/ExampleConfigsIT.java
danberindei: Considering that the test also verifies the cache contents with `retrieveEntries()` instead of `get()`, I'd argue that this isn't fixing the existing test, it's just testing something different.

danberindei:
@slaskawi Could you explain what you mean by "the same key references for validation" in the commit message?

server/integration/testsuite/src/test/java/org/infinispan/server/test/configs/ExampleConfigsIT.java
slaskawi: Yes, I think you are right....

slaskawi:
@danberindei I though this could be connected some weird String interning thing... So I'm using exactly the same String references for putting them in a cache and checking if they are valid

danberindei:
@slaskawi We never use `String.intern()` in our code, and the JVM only interns string literals, never the result of a concatenation.

slaskawi:
hmmm in that case this whole PR does not make much sense... Let me close it and reopen if I find a better solution.

PR #4185, opened 2016-03-25
ISPN-6437 InfinispanLock.LockPlaceHolder sometimes doesn't invoke its listeners

pruivo:
Just cleaning up the test and add more info to the InfinispanLock

https://issues.jboss.org/browse/ISPN-6437

### DIFF
--- a/core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
+++ b/core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
@@ -312,11 +312,12 @@ public class InfinispanLock {
 
       @Override
       public void lock() throws InterruptedException, TimeoutException {
-         while (true) {
-            switch (lockState) {
+         do {
+            LockState currentState = lockState;
+            switch (currentState) {
                case WAITING:
                   checkTimeout();
-                  await(notifier, timeService.remainingTime(timeout, TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
+                  await(notifier, timeService.remainingTime(timeout, TimeUnit.NANOSECONDS), TimeUnit.NANOSECONDS);
                   break;
                case ACQUIRED:
                   return; //acquired!
@@ -329,9 +330,9 @@ public class InfinispanLock {
                   cleanup();
                   throw new DeadlockDetectedException("DeadLock detected");
                default:
-                  throw new IllegalStateException("Unknown lock state: " + lockState);
+                  throw new IllegalStateException("Unknown lock state: " + currentState);
             }
-         }
+         } while (true);
       }
 
       @Override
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/util/concurrent/locks/impl/InfinispanLock.java
```diff
@@ -342,14 +346,14 @@ public void addListener(LockListener listener) {
       @Override
       public void cancel(LockState state) {
          checkValidCancelState(state);
-         out:
          do {
             LockState currentState = lockState;
             switch (currentState) {
                case WAITING:
                   if (casState(LockState.WAITING, state)) {
+                     onCanceled(this);
```
danberindei: This looks like a fix to me, not just added logging :)

pruivo: the ```onCanceled``` was outside of the ```while```. I just moved it inside and replace it ```break``` with a ```return```

danberindei: I was thinking about moving `onCancelled()` before `notifyListeners()`. But I guess you're right, because I was able to reproduce the test failure again :)

danberindei:
@pruivo It's pretty easy for me to reproduce the failure on my machine with `@Test(invocationCount = 1000)`, so I'd rather not integrate this without a proper fix.

pruivo:
@danberindei updated!

extended-statistics/src/test/java/org/infinispan/stats/topK/LocalTopKeyTest.java
```diff
@@ -79,6 +79,7 @@ public void testGet(Method method) {
       assertWriteSkew(key2, 0);
    }
 
+   @Test(invocationCount = 1000)
```
danberindei: Considering this isn't a stress test, I wouldn't commit this. Maybe with `invocationCount = 100`...

extended-statistics/src/test/java/org/infinispan/stats/topK/LocalTopKeyTest.java
```diff
@@ -158,33 +159,45 @@ private CacheUsageInterceptor getTopKey() {
...
-      } else {
-         actual = getTopKey().getLocalTopPuts().get(key);
-      }
-      Assert.assertEquals(actual == null ? 0 : actual, expected, "Wrong number of accesses");
+      final CacheUsageInterceptor topK = getTopKey();
+      eventuallyEquals("Wrong number of accesses.", expected, () -> {
+         Long value = readAccesses ?
+               topK.getLocalTopGets().get(key) :
+               topK.getLocalTopPuts().get(key);
+         return value == null ? 0 : value;
```
danberindei: `getOrDefault(key, 0)`?

extended-statistics/src/test/java/org/infinispan/stats/topK/LocalTopKeyTest.java
pruivo: that was a mistake :)

pruivo:
updated again @danberindei!

danberindei:
Thanks Pedro, integrated!

PR #4189, opened 2016-03-30
ISPN-6404 Added missing schema files

slaskawi:
https://issues.jboss.org/browse/ISPN-6404

I took all files from WF10 distribution and placed them in the feature pack. That's why the changes are so big.


slaskawi:
@tristantarrant Could you please have a look at [my comment on the jira?](https://issues.jboss.org/browse/ISPN-6404?focusedCommentId=13183655&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13183655)

tristantarrant:
I was hoping for the schemas to be extracted from the modules, like wildfly does it. Bumping the wildfly build tools to 1.1.6.Final (and modifying keep-schemas.txt) does get us somewhere there, but it is still missing our own schemas...

slaskawi:
Ok, I'll have a look and implement necessary changes.

slaskawi:
@tristantarrant comments addressed. Now we use `wildfly-server-provisioning-maven-plugin` `1.1.6` with updated `extract-schemas-groups`. All schemas are unpacked from the modules.

tristantarrant:
Pushed, thanks

PR #4190, opened 2016-03-30
ISPN-5683 Turned on tab validation

slaskawi:
https://issues.jboss.org/browse/ISPN-5683

Automatic correction performed using:
```
find . -name '*.properties' ! -type d -exec bash -c 'expand -t 3 "$0" > /tmp/e && mv /tmp/e "$0"' {} \;
find . -name '*.java' ! -type d -exec bash -c 'expand -t 3 "$0" > /tmp/e && mv /tmp/e "$0"' {} \;
```


slaskawi:
Marking as urgent since conflicting with other PR is more than likely :)

tristantarrant:
Pushed to master, thanks

PR #4196, opened 2016-03-30
ISPN-6364 Report test failures on the fly in modules using JUnit

danberindei:
https://issues.jboss.org/browse/ISPN-6364

* Add a JUnit listener for JUnit-based modules
* Put the TestNG/JUnit progress reporters in a separate module.


commons-test/src/main/java/org/infinispan/commons/test/TestSuiteProgress.java
```diff
@@ -0,0 +1,78 @@
...
+   static void testFinished(String name) {
+      String message = "Test succeeded: " + name;
+      consoleLog(message);
+      log.info(message);
+      succeeded.incrementAndGet();
+      printStatus();
+   }
+
+   static void testFailed(String name, Throwable exception) {
+      String message = "Test starting: " + name;
```
alanfx: "Test starting: " -> "Test failed: "?

bom/pom.xml
```diff
@@ -162,6 +162,11 @@
          </dependency>
          <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>infinispan-commons-test</artifactId>
```
slaskawi: +1000 for creating new commons test module. I heart was bleeding when I saw core's test jar included everywhere only because of the listeners :)

danberindei: @slaskawi I think it's more because of `TestCacheManagerFactory` and `AbstractInfinispanTest` than because of the listeners :)

galderz:
Integrated

PR #4200, opened 2016-03-31
ISPN-6452 Use JBoss Marshaller OSGi in client EAP module

slaskawi:
https://issues.jboss.org/browse/ISPN-6452

Since we use JBoss Marshaller OSGi as a runtime/compiletime dependency - we should also use it in EAP modules.


slaskawi:
Rebased

tristantarrant:
Thanks @slaskawi , pushed

PR #4201, opened 2016-03-31
ISPN-6458 DistTopKeyTest.testLockLockFailed random failures

pruivo:
* Due to async nature of top-k, use eventually for asserts

I've moved all the asserts to TestingUtil in extended stats module, to be shared between local and clustered tests.
Lets hope this would fix all the random failures in top key :)

https://issues.jboss.org/browse/ISPN-6458

### DIFF
--- a/extended-statistics/src/test/java/org/infinispan/stats/AbstractTopKeyTest.java
+++ b/extended-statistics/src/test/java/org/infinispan/stats/AbstractTopKeyTest.java
@@ -1,0 +1,133 @@
+package org.infinispan.stats;
+
+import org.infinispan.Cache;
+import org.infinispan.commands.tx.PrepareCommand;
+import org.infinispan.context.impl.TxInvocationContext;
+import org.infinispan.distribution.DistributionManager;
+import org.infinispan.interceptors.TxInterceptor;
+import org.infinispan.interceptors.base.CommandInterceptor;
+import org.infinispan.stats.topK.CacheUsageInterceptor;
+import org.infinispan.test.MultipleCacheManagersTest;
+
+import java.util.List;
+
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
```diff
@@ -91,16 +91,16 @@ protected void eventuallyEquals(int expected, IntSupplier supplier) {
             () -> expected == supplier.getAsInt());
    }
 
-   protected void eventuallyEquals(String message, long expected, LongSupplier supplier) {
+   public static void eventuallyEquals(String message, long expected, LongSupplier supplier) {
```
danberindei: I don't like mixing static and non-static methods in the same class, especially when they have so similar names, so I'd rather move these to a separate class, e.g. `Eventually`. 

Or, if you don't want to affect the core tests, you could just replace your `TestingUtil` with an `AbstractStatisticsTest` extending `AbstractInfinispanTest`.

pruivo: The second option doesn't work since one test extends ```SingleCacheManagerTest``` and the other ```MultipleCacheManagersTest```

changing to a separate class looks better to me, but I prefer to do that in another JIRA than full this one with changes everywhere :)

danberindei: You can extend `MultipleCacheManagersTest` in both tests, there's no rule that non-clustered tests have to use `SingleCacheManagerTest`.

pruivo:
updated!

danberindei:
Integrated, thanks Pedro!

PR #4205, opened 2016-04-01
ISPN-6461 Add Spring common content into remote and embedded

slaskawi:
https://issues.jboss.org/browse/ISPN-6461

This could go to master and `8.2.x` however this is not a must have. The workaround is really simple - just add `spring4-commons` to you app.

### DIFF
--- a/spring/spring4/spring4-common/pom.xml
+++ b/spring/spring4/spring4-common/pom.xml
@@ -25,25 +25,6 @@
     </dependencies>
 
     <build>
-        <resources>
-            <resource>
-                <directory>${project.basedir}/src/main/resources</directory>
-                <filtering>false</filtering>
-                <includes>
-                    <include>**/infinispan-spring-${infinispan.core.schema.version}.xsd</include>
-                </includes>
-            </resource>
-            <resource>
-                <directory>${project.basedir}/src/main/resources</directory>
-                <filtering>false</filtering>
-                <includes>
-                    <include>**/*</include>
-                </includes>
-                <excludes>
-                    <exclude>**/infinispan-spring-*.xsd</exclude>
-                </excludes>
-            </resource>
-        </resources>
         <plugins>
             <plugin>
                 <groupId>org.apache.felix</groupId>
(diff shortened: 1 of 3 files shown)


gustavonalle:
@slaskawi the branch name says it all: ```Spring-commons-can-not-be-optional```, then why in ```spring4-embedded``` and ```spring4-remote``` there's:
```
<dependency>
       <artifactId>infinispan-spring4-common</artifactId>
        <optional>true</optional>
</dependency>
```

To me it sounds more natural to make it a ```compile``` dependency than manually copying the resources from it

slaskawi:
Sorry for confusion... the old idea was to make it compile time dependency. However I implemented it slightly differently - with Build Helper plugin which includes all source files and resources from common into the embedded and remote.

gustavonalle:
> Build Helper plugin which includes all source files and resources from common into the embedded and remote.

Going back to the original question, why not simply make ```commons```  a compile dependency?

slaskawi:
Mainly because I wanted to introduce as few dependencies as possible on the client side. Having additional `commons` module adds only a noise.

gustavonalle:
> Mainly because I wanted to introduce as few dependencies as possible on the client side

commons jar is extra tiny, and all of the build tools will resolve it automatically.

slaskawi:
We decided to go for having `spring-common` as a compile time dep.

danberindei:
@slaskawi I tried to find the dev-list thread about this to refresh my memory, but I couldn't find anything with spring-common in it. Weren't we discussing adding the same source in 2 modules in the context of supporting both spring 3 and spring 4 from the same sources?

gustavonalle:
@danberindei I believe you're referring to http://lists.jboss.org/pipermail/infinispan-dev/2014-October/015560.html
But lots of things changed since then: spring3 is gone, spring module was split into embedded and remote, and commons was introduced between them

slaskawi:
@danberindei @gustavonalle We've been discussing it on the internal mailing list. Look for thread "Packaging proposal".

slaskawi:
Rebased and addressed comments

gustavonalle:
integrated, thanks @slaskawi !

PR #4210, opened 2016-04-04
ISPN-6463 Enable Checkstyle during build

slaskawi:
https://issues.jboss.org/browse/ISPN-6463

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -1213,16 +1213,8 @@
                   <failsOnError>true</failsOnError>
                   <violationSeverity>error</violationSeverity>
                   <includeTestSourceDirectory>true</includeTestSourceDirectory>
+                  <sourceDirectory>${project.build.sourceDirectory}</sourceDirectory>
                </configuration>
-               <executions>
-                  <execution>
-                     <id>checkstyle</id>
-                     <phase>process-sources</phase>
-                     <goals>
-                        <goal>checkstyle</goal>
-                     </goals>
-                  </execution>
-               </executions>
              </plugin>
              <plugin>
                 <groupId>org.scala-tools</groupId>
@@ -1345,7 +1337,19 @@
                <buildOutputDirectory>${basedir}/eclipse-output</buildOutputDirectory>
             </configuration>
          </plugin>
-
+         <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-checkstyle-plugin</artifactId>
+            <executions>
+               <execution>
+                  <id>checkstyle</id>
+                  <phase>verify</phase>
+                  <goals>
+                     <goal>checkstyle</goal>
+                  </goals>
+               </execution>
+            </executions>
+         </plugin>
          <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-idea-plugin</artifactId>
--- a/checkstyle/pom.xml
+++ b/checkstyle/pom.xml
@@ -4,9 +4,9 @@
    <modelVersion>4.0.0</modelVersion>
    <parent>
       <groupId>org.infinispan</groupId>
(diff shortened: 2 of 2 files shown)


slaskawi:
Rebased

parent/pom.xml
```diff
@@ -1513,7 +1505,19 @@
...
             </configuration>
          </plugin>
-
+         <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-checkstyle-plugin</artifactId>
+            <executions>
+               <execution>
+                  <id>checkstyle</id>
+                  <phase>validate</phase>
```
gustavonalle: ```validate``` comes before ```compile```, I think it'd be better to associate with ```verify```

gustavonalle:
@slaskawi Looks good, apart from the lifecycle phase.
Btw, all checks are commented out on ```checkstyle.xml```, so the plugin is not doing anything useful.

parent/pom.xml
slaskawi: Exactly.. the main though was that if it doesn't fit the rules, it's not worth compiling. What would be the argument for running it in `verify`?

slaskawi:
Yes, most of them are disabled. I'm turning them on gradually

parent/pom.xml
gustavonalle: My rationale is if the code is not compiling, I'd like to get the compile error first instead of an eventual checkstyle error

slaskawi: hmmm yeah it also makes sense... when I think about it again... your approach might be slightly more friendly for those who do not have properly configured IDE... Let's go for it...

gustavonalle:
@slaskawi LGTM, but you added "Changed Required" label?

slaskawi:
whoops missed the lablel button...

slaskawi:
Rebased

gustavonalle:
I believe there's a circularity here: 
the module ```infinispan-checkstyle``` inherits from ```infinispan-parent```, but ```infinispan-parent``` has the checkstyle plugin declared both in the ```pluginManagement``` and in the ```plugin```sections, making ```infinispan-parent``` also a dependant of ```infinispan-checkstyle```.


$ rm -Rf $HOME/.m2/repository/org/infinispan/infinispan-checkstyle
$ mvn -o clean install  -DskipTests=true -s maven-settings.xml
[INFO] Reactor Build Order:
[INFO] 
[INFO] Infinispan BOM
[INFO] Infinispan Common Parent
[INFO] Infinispan Checkstyle Rules
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:checkstyle (checkstyle) on project infinispan-commons: An error has occurred in Checkstyle report generation. Failed during checkstyle execution: Unable to find configuration file at location: checkstyle.xml: Could not find resource 'checkstyle.xml'. -> [Help 1]

To solve the issue, either ```infinispan-parent``` should be prevented to run checkstyle or ```infinispan-checkstyle``` should not inherit from parent

danberindei:
@gustavonalle @slaskawi I'd rather change infinispan-checkstyle to not inherit infinispan-parent.

gustavonalle:
@danberindei @slaskawi the other alternative would be to keep checkstyle only on ```pluginManagement``` and enable it on a per module basis, that should work as well

slaskawi:
ooh I missed that... Running with `-o` is really neat here... Let me think about some alternatives and get back to it shortly. Currently I'm leaning towards to @danberindei's idea.

slaskawi:
I changed checkstyle module's parent from `infinispan-parent` to `infinispan-bom`

gustavonalle:
Looks good, pulling...

gustavonalle:
integrated, thanks @slaskawi!

PR #4211, opened 2016-04-04
ISPN-6379 Add event logs for backup sites failures

pruivo:
https://issues.jboss.org/browse/ISPN-6379

### DIFF
--- a/core/src/main/java/org/infinispan/util/logging/events/Messages.java
+++ b/core/src/main/java/org/infinispan/util/logging/events/Messages.java
@@ -71,4 +71,10 @@ public interface Messages {
 
    @Message(value = "After merge (or coordinator change), the coordinator failed to recover cluster. Cluster members are %s.", id = 100004)
    String clusterRecoveryFailed(Collection<Address> members);
+
+   @Message(value = "Site '%s' is online.", id = 100005)
+   String siteOnline(String siteName);
+
+   @Message(value = "Site '%s' is offline.", id = 100006)
+   String siteOffline(String siteName);
 }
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/infinispan/util/logging/events/Messages.java
```diff
@@ -71,4 +71,10 @@
 
    @Message(value = "After merge (or coordinator change), the coordinator failed to recover cluster. Cluster members are %s.")
    String clusterRecoveryFailed(Collection<Address> members);
+
+   @Message(value = "Site '%s' is online.", id = 100004)
```
anistor: Where does this id come from? Last messages have no id. Others have ids in a completely different range. Maybe @danberindei or @tristantarrant know?

pruivo: because the messages start with id 100000.
the messages with id 300+ was moved from the ```Log``` as it is in the comment:
```
// Messages between 312 and 320 have been moved to the org.infinispan.util.logging.events.Messages class
```
the ```eventLog*``` has no id because it is used to build the final message (from what I understand)
finally, I have no idea why the last message has no id.

anistor: OK, but clusterRecoveryFailed should also get an id. Could you fix that please?

pruivo: @anistor https://github.com/infinispan/infinispan/pull/4216

pruivo:
updated!

tristantarrant:
Pushed, thanks.

PR #4212, opened 2016-04-04
ISPN-5354 HotRod putAll/getAll send proper CH routing

gdarmont:
Hi Infinispan team,

This PR is an implementation proposal for ISPN-5354, where HotRod client splits one PutAll/GetAll operation into N operations where N is the number of nodes in the remote cluster.

__Characteristics of the proposed implementation :__
* Previous GetAllOperation/PutAllOperation are now used as sub-operations for GelAllParallelOperation/PutAllParallelOperation. This allows to rely on existing retry behavior when an operation fails.
* Uses a dedicated ExecutorService named “parallelExecutorService“
  * parallelExecutorService parameters are duplicated from existing asyncExecutorService
  * Default pool size is 32
* Behavior :
  * Only one target node : call is simply done in the caller thread
  * Multiple target nodes : create multiple operations that uses the parallelExecutorService with a CompletionService
  * When a sub-operation fails, all remaining operations are cancelled and a ParallelOperationException is thrown.

This PR lacks :
* Documentation on the new behavior
* Unit tests that could enforce the fact that each cluster node is requested

Hope this helps,
Guillaume

-------------

Some results from tests ran on my machine, using a 3 nodes hotrod cluster.
8.2.0.Final

Iteration :  196 Put all time : 393 ms
Iteration :  197 Put all time : 397 ms
Iteration :  198 Put all time : 355 ms
Iteration :  199 Put all time : 394 ms
Iteration :  200 Put all time : 391 ms

Iteration :  196 Get all time :  19 ms
Iteration :  197 Get all time :  21 ms
Iteration :  198 Get all time :  21 ms
Iteration :  199 Get all time :  23 ms
Iteration :  200 Get all time :  19 ms

Entries count per operation : 10000. Iterations : 200
Total time (ms) 83460

-----------------------

9.0.0 SNAPSHOT

Iteration :  196 Put all time : 210 ms
Iteration :  197 Put all time : 149 ms
Iteration :  198 Put all time : 218 ms
Iteration :  199 Put all time : 148 ms
Iteration :  200 Put all time : 154 ms

Iteration :  196 Get all time :   9 ms
Iteration :  197 Get all time :   9 ms
Iteration :  198 Get all time :   9 ms
Iteration :  199 Get all time :   8 ms
Iteration :  200 Get all time :   8 ms

Entries count per operation : 10000. Iterations : 200
Total time (ms) 38037


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
```diff
@@ -69,6 +69,7 @@
    private Marshaller marshaller;
    protected TransportFactory transportFactory;
    private ExecutorService asyncExecutorService;
+   private ExecutorService parallelExecutorService;
```
gustavonalle: Instead of having two executor services in the client, I'd prefer deprecating the ```asyncExecutorService``` and create a new config ```executorService``` that will handle both async and parallel operations. This way the client will have a single executor.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/exceptions/ParallelOperationException.java
```diff
@@ -0,0 +1,19 @@
+package org.infinispan.client.hotrod.exceptions;
```
gustavonalle: @gdarmont Please make sure your IDE is properly configured to follow out code style, as described in http://infinispan.org/docs/8.0.x/contributing/contributing.html#_setting_up_your_ide

gustavonalle:
Hi @gdarmont, nice perf improvements :+1: 
Apart from the extra executor, LGTM
@wburns any comments?

gdarmont:
Thanks for the review. I'll make the requested changes.

wburns:
@gdarmont This looks pretty good :)  I like the use of the framework to do either operation.  The only reason this functionality wasn't added before was because I wanted to implement this once we had an async client, since we wouldn't need a thread per node :+1:  But even without it the abstraction for the implementations would be useful either way.

@galderz are you alright with the threading?  If not this seems fine with me given the rework.

galderz:
I'm happy with Gustavo's suggestion to reuse the the async executor.

tristantarrant:
Thanks very much @gdarmont, I've squashed the commits and pushed to both master and 8.2.x

gdarmont:
@tristantarrant Glad I could help. 
As @gustavonalle suggested in his first comment, I've made a commit ([here](https://github.com/gdarmont/infinispan/commit/1020fa25ea531b94728e2228f8f26b325f93eaae)) where the exsiting `asyncExecutorService` is deprecated in favor of a simple `executorService`.
API are adapted with the new name and all previous public API are kept for compatibility but deprecated.

I could make another PR if you are interested in such refactoring.

tristantarrant:
Yes, this should be under a separate Jira, since we only would do this in master

PR #4217, opened 2016-04-07
ISPN-6471 Avoid blocking Netty IO threads, delegating to a separate executor

gustavonalle:
https://issues.jboss.org/browse/ISPN-6471

PREVIEW ONLY

Working on bechmark tests

### DIFF
--- a/server/core/src/main/scala/org/infinispan/server/core/transport/NettyChannelInitializer.scala
+++ b/server/core/src/main/scala/org/infinispan/server/core/transport/NettyChannelInitializer.scala
@@ -1,11 +1,13 @@
 package org.infinispan.server.core.transport
 
+import javax.net.ssl.SSLEngine
+
+import io.netty.channel.{Channel, ChannelInitializer, ChannelOutboundHandler}
+import io.netty.handler.ssl.SslHandler
+import io.netty.util.concurrent.EventExecutorGroup
+import org.infinispan.commons.util.SslContextFactory
 import org.infinispan.server.core.ProtocolServer
 import org.infinispan.server.core.configuration.SslConfiguration
-import javax.net.ssl.SSLEngine
-import org.infinispan.commons.util.SslContextFactory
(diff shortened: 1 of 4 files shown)


server/core/src/main/scala/org/infinispan/server/core/transport/NettyChannelInitializer.scala
```diff
@@ -16,17 +18,17 @@ import io.netty.handler.ssl.SslHandler
...
+                              encoder: ChannelOutboundHandler, executor: EventExecutorGroup)
+        extends ChannelInitializer[Channel] {
 
    override def initChannel(ch: Channel): Unit = {
       val pipeline = ch.pipeline
       val ssl = server.getConfiguration.ssl
       if (ssl.enabled())
-         pipeline.addLast("ssl", new SslHandler(createSslEngine(ssl)))
-      pipeline.addLast("decoder", server.getDecoder)
+         pipeline.addLast(executor, "ssl", new SslHandler(createSslEngine(ssl)))
```
wburns: Do we want this for ssl as well?

wburns:
I was also planning on doing this with my changes.  It would be nice to see what benchmarks you are using, as I will need them for other purposes as well :)

server/core/src/main/scala/org/infinispan/server/core/transport/NettyChannelInitializer.scala
gustavonalle: Yes, any concerns?

gustavonalle:
There directory memory failures in the ```hotrod-client``` suite, will investigate

@wburns The idea is to use a JMH benchmark hammering 1 server, doing a mix of fast ops (get/put) and potentially blocking ops (script, file system based indexing)

gustavonalle:
Closing since changes are incorporated in https://github.com/infinispan/infinispan/pull/4240

PR #4218, opened 2016-04-07
ISPN-6468 SyncReplExtendedStatisticTest.​testReplace is unstable

pruivo:
* mark testReplace as unstable


anistor:
I've added back the description pointing to the issue and pulled it. Thanks @pruivo !

PR #4221, opened 2016-04-08
ISPN-6475 Remove the async log4j2 appender in the test configurations

danberindei:
https://issues.jboss.org/browse/ISPN-6475


tristantarrant:
Pushed to master and 8.2.x

PR #4219, opened 2016-04-07
ISPN-6472 Add server-side SASL EXTERNAL mech for client-certificate authentication

tristantarrant:
https://issues.jboss.org/browse/ISPN-6472
https://issues.jboss.org/browse/ISPN-6473

PR #4222, opened 2016-04-08
ISPN-6438 Configure client SSL with properties

galderz:


tristantarrant:
Looks good. This also needs adding to the docs and possibly an "secure jcache" example. Once https://github.com/infinispan/infinispan/pull/4219 is done we can provide one.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/configuration/ConfigurationTest.java
```diff
@@ -155,30 +270,52 @@ private void assertServer(String host, int port, ServerConfiguration serverCfg)
 
    private void validateConfiguration(Configuration configuration) {
       assertEquals(2, configuration.servers().size());
-      assertEquals(SomeAsyncExecutorFactory.class, configuration.asyncExecutorFactory().factoryClass());
-      assertEquals(SomeRequestBalancingStrategy.class, configuration.balancingStrategyClass());
-      assertEquals(SomeTransportfactory.class, configuration.transportFactory());
+      assertEqualsConfig(SomeAsyncExecutorFactory.class, ASYNC_EXECUTOR_FACTORY, configuration);
```
danberindei: Considering the OPTIONS map is only used here, wouldn't it have been simpler to "inline" the lambdas?

tristantarrant:
We should probably also support "injecting" a pre-built SSLContext

tristantarrant:
A few more properties, to enable authentication as well as encryption please:
authenticationEnabled (a boolean)
saslMechanism (a string, should support multiple values which could be comma-separated)
callbackHandler (an instance of javax.security.auth.callback.CallbackHandler)
serverName (a string)
clientSubject (an instance of javax.security.auth.Subject)
saslProperties (a map<String, String>, need to think how to represent this as a string, an instance would be fine)

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/configuration/ConfigurationTest.java
galderz: The idea behind it is to be able to reuse them potentially in other client config tests.

galderz:
@tristantarrant Authentication properties have been added.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/Configuration.java
```diff
@@ -202,4 +206,97 @@ public String toString() {
...
+      properties.setProperty(ConfigurationProperties.SERVER_LIST, servers.toString());
+
+      properties.setProperty(ConfigurationProperties.USE_SSL, Boolean.toString(security.ssl().enabled()));
+
+      if (security.ssl().keyStoreFileName() != null)
+         properties.setProperty(ConfigurationProperties.KEY_STORE_FILE_NAME, security.ssl().keyStoreFileName());
+
+      if (security.ssl().keyStorePassword() != null)
+         properties.setProperty(ConfigurationProperties.KEY_STORE_PASSWORD, new String(security.ssl().keyStorePassword()));
+
```
tristantarrant: I just pushed ISPN-6497 which adds one extra property: keyStoreCertificatePassword
Can you please add it @galderz. I promise it's the last one ;)

tristantarrant: I need to rectify my statement. Can you also make the callbackHandler with both with a String (and therefore initialize it with reflection) or an instance ?

galderz:
@tristantarrant +1 to both comments, they're both included now.

tristantarrant:
Pushed, thanks

PR #4224, opened 2016-04-08
ISPN-6477 Security tests always log a warning during cache manager stop

danberindei:
https://issues.jboss.org/browse/ISPN-6477


tristantarrant:
Pushed, thanks

PR #4227, opened 2016-04-11
ISPN-6457 Access protected interal caches over HotRod only with authz…

tristantarrant:
… or over loopback

https://issues.jboss.org/browse/ISPN-6457

### DIFF
--- a/server/hotrod/src/test/scala/org/infinispan/server/hotrod/HotRodSingleClusteredNonLoopbackTest.scala
+++ b/server/hotrod/src/test/scala/org/infinispan/server/hotrod/HotRodSingleClusteredNonLoopbackTest.scala
@@ -1,24 +1,24 @@
 package org.infinispan.server.hotrod
 
-import test.{HotRodClient, TestErrorResponse}
 import java.lang.reflect.Method
 import java.util.EnumSet
 
-import org.infinispan.server.hotrod.OperationStatus._
-import org.infinispan.test.MultipleCacheManagersTest
-import test.HotRodTestingUtil._
-import org.infinispan.test.AbstractCacheTest._
-import org.testng.annotations.{AfterClass, BeforeClass, Test}
-import org.infinispan.server.core.test.ServerTestingUtil._
-import org.infinispan.test.fwk.TestCacheManagerFactory
 import org.infinispan.configuration.cache.{CacheMode, ConfigurationBuilder}
-import org.infinispan.commons.equivalence.ByteArrayEquivalence
+import org.infinispan.manager.EmbeddedCacheManager
 import org.infinispan.registry.InternalCacheRegistry
-import org.infinispan.server.hotrod.configuration.HotRodServerConfiguration
+import org.infinispan.server.core.test.ServerTestingUtil._
+import org.infinispan.server.hotrod.OperationStatus._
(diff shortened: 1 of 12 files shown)


server/hotrod/src/test/scala/org/infinispan/server/hotrod/HotRodSingleClusteredNonLoopbackTest.scala
```diff
@@ -0,0 +1,67 @@
...
+         getDefaultClusteredCacheConfig(CacheMode.REPL_SYNC, false))
+      cm.defineConfiguration(cacheName, builder.build())
+
+   }
+
+   @BeforeClass(alwaysRun = true)
+   @Test(enabled=false) // Disable explicitly to avoid TestNG thinking this is a test!!
+   override def createBeforeClass() {
+      super.createBeforeClass()
+      val iface = findNetworkInterfaces(false).next
```
tristantarrant: caveat: I'm binding the server to the first non-loopback interface to test this

galderz:
Needs rebasing

galderz:
-1 if I'm understanding this correctly:

The Javascript client needs to have the ability to remotely store scripts and currently I'm using that by putting those scripts directly in the script cache via a cache.put. It seems like your change will disable this?

If so, how is the Javascript client supposed to load scripts remotely?

tristantarrant:
The change doesn't affect behaviour when connecting over a loopback interface (e.g. localhost). So if your tests go over 127.0.0.1 or ::1 that will be fine. In all other cases, authorization must be enabled. You will then need either an authenticated client or go through the console/CLI.

galderz:
> The change doesn't affect behaviour when connecting over a loopback interface (e.g. localhost). So if your tests go over 127.0.0.1 or ::1 that will be fine. In all other cases, authorization must be enabled. You will then need either an authenticated client or go through the console/CLI.

Authentication has not yet been implemented in JS client so console/CLI would be the only alternative.

I had added a method to the JS client precisely to make it easier to add a script remotely (see [here](https://github.com/infinispan/js-client/blob/master/lib/infinispan.js#L322)). Should this remain? I guess so. For now it'll only work when calling from localhost as you said, and eventually it should work from anywhere once authentication is in place.

Btw, the documentation says:

> 14.5.1. Installing scripts
> Scripts are stored in a special script cache, named '___script_cache'. Adding a script is therefore as simple as put+ting it into the cache itself. If the name of the script contains a filename extension, e.g. +myscript.js, then that extension determines the engine that will be used to execute it. Alternatively the script engine can be selected using script metadata (see below).

Not only does not mention how to install scripts via console/CLI, but this whole section needs to be updated and I see no such changes in this PR.

galderz:
Documentation to come later, integrating...

galderz:
Actually, testing first

galderz:
Testing looks good, integrating...

alanfx:
> The change doesn't affect behaviour when connecting over a loopback interface (e.g. localhost). So if your tests go over 127.0.0.1 or ::1 that will be fine.

Doesn't this still leave a security hole? (Asking for a friend!)

tristantarrant:
@alanfx I've consulted with Chess on this and he's ok with it.

alanfx:
👍 I'll take his word for it, then!

PR #4230, opened 2016-04-12
ISPN-6482 Allow setup only truststore in HR client config

vjuranek:
https://issues.jboss.org/browse/ISPN-6482

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/SslConfigurationBuilder.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/SslConfigurationBuilder.java
@@ -101,9 +101,6 @@ public class SslConfigurationBuilder extends AbstractSecurityConfigurationChildB
    public void validate() {
       if (enabled) {
          if (sslContext == null) {
-            if (keyStoreFileName == null) {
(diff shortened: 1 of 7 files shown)


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/SslConfigurationBuilder.java
```diff
@@ -101,9 +101,6 @@ public SslConfigurationBuilder trustStorePassword(char[] trustStorePassword) {
    public void validate() {
       if (enabled) {
          if (sslContext == null) {
-            if (keyStoreFileName == null) {
-               throw log.noSSLKeyManagerConfiguration();
```
tristantarrant: You can remove the log now too.

vjuranek: looked to me strange to have gaps in log numbers :smile: . There's more log messages which seem to be unused, I removed them all in a additional commit (and created https://issues.jboss.org/browse/ISPN-6492)

tristantarrant:
Thanks, I've pushed it.

PR #4232, opened 2016-04-13
ISPN-6477 Unregister cache mbean even if cache is terminated

tristantarrant:
https://issues.jboss.org/browse/ISPN-6477


danberindei:
Integrated, but the warning reappeared so I'm keeping the issue open.

PR #4233, opened 2016-04-13
ISPN-6488 Convert CacheRpcCommands to use a pre-computed ByteStrings

tristantarrant:
https://issues.jboss.org/browse/ISPN-6488

WARNING: this incarnation is backwards incompatible since it changes the CacheRpcCommand to _only_ take a ByteString instead of a String. I did this to make sure I didn't miss any existing internal commands, but I will re-add the String constructors once we are happy

### DIFF
--- a/core/src/main/java/org/infinispan/commands/tx/VersionedCommitCommand.java
+++ b/core/src/main/java/org/infinispan/commands/tx/VersionedCommitCommand.java
@@ -3,6 +3,7 @@ package org.infinispan.commands.tx;
 import org.infinispan.commons.marshall.MarshallUtil;
 import org.infinispan.container.versioning.EntryVersionsMap;
 import org.infinispan.transaction.xa.GlobalTransaction;
+import org.infinispan.util.ByteString;
 
 import java.io.IOException;
 import java.io.ObjectInput;
@@ -20,14 +21,14 @@ public class VersionedCommitCommand extends CommitCommand {
    private EntryVersionsMap updatedVersions;
 
    public VersionedCommitCommand() {
-      super("");
+      super(null);
    }
 
-   public VersionedCommitCommand(String cacheName, GlobalTransaction gtx) {
+   public VersionedCommitCommand(ByteString cacheName, GlobalTransaction gtx) {
       super(cacheName, gtx);
    }
 
-   public VersionedCommitCommand(String cacheName) {
+   public VersionedCommitCommand(ByteString cacheName) {
       super(cacheName);
    }
 
--- a/core/src/main/java/org/infinispan/factories/GlobalComponentRegistry.java
+++ b/core/src/main/java/org/infinispan/factories/GlobalComponentRegistry.java
@@ -27,6 +27,7 @@ import org.infinispan.registry.impl.InternalCacheRegistryImpl;
 import org.infinispan.remoting.transport.Transport;
 import org.infinispan.topology.ClusterTopologyManager;
 import org.infinispan.topology.LocalTopologyManager;
+import org.infinispan.util.ByteString;
 import org.infinispan.util.ModuleProperties;
 import org.infinispan.util.TimeService;
 import org.infinispan.util.logging.Log;
@@ -83,7 +84,7 @@ public class GlobalComponentRegistry extends AbstractComponentRegistry {
 
    final Collection<ModuleLifecycle> moduleLifecycles;
 
-   final ConcurrentMap<String, ComponentRegistry> namedComponents = new ConcurrentHashMap<String, ComponentRegistry>(4);
+   final ConcurrentMap<ByteString, ComponentRegistry> namedComponents = new ConcurrentHashMap<ByteString, ComponentRegistry>(4);
 
    protected final WeakReference<ClassLoader> defaultClassLoader;
 
@@ -196,15 +197,20 @@ public class GlobalComponentRegistry extends AbstractComponentRegistry {
 
    public final ComponentRegistry getNamedComponentRegistry(String name) {
       //no need so sync this method as namedComponents is thread safe and correctly published (final)
+      return getNamedComponentRegistry(ByteString.fromString(name));
+   }
+
+   public final ComponentRegistry getNamedComponentRegistry(ByteString name) {
+      //no need so sync this method as namedComponents is thread safe and correctly published (final)
       return namedComponents.get(name);
    }
 
    public synchronized final void registerNamedComponentRegistry(ComponentRegistry componentRegistry, String name) {
-      namedComponents.put(name, componentRegistry);
+      namedComponents.put(ByteString.fromString(name), componentRegistry);
    }
 
    public synchronized final void unregisterNamedComponentRegistry(String name) {
-      namedComponents.remove(name);
+      namedComponents.remove(ByteString.fromString(name));
    }
 
    public synchronized final void rewireNamedRegistries() {
--- a/core/src/main/java/org/infinispan/util/ByteString.java
+++ b/core/src/main/java/org/infinispan/util/ByteString.java
@@ -1,0 +1,78 @@
+package org.infinispan.util;
+
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+/**
+ * A simple class which encapsulates a byte[] representation of a String using a predefined encoding (currently UTF-8).
+ * This avoids repeated invocation of the expensive {@link ObjectOutput#writeUTF(String)} on marshalling
+ *
+ * @author Tristan Tarrant
+ * @since 9.0
+ */
+public class ByteString {
+   public static final Charset CHARSET = StandardCharsets.UTF_8;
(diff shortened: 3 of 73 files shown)


tristantarrant:
We might NOT want the String constructors.

core/src/main/java/org/infinispan/util/ByteString.java
```diff
@@ -0,0 +1,90 @@
...
+import java.nio.charset.Charset;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * @author Tristan Tarrant
+ * @since 9.0
+ */
+public class ByteString {
```
galderz: Why do you need this wrapper class? Can't you just work with byte[] directly? What's stopping you?

core/src/main/java/org/infinispan/util/ByteString.java
```diff
@@ -0,0 +1,90 @@
...
+      public void writeObject(ObjectOutput output, ByteString object) throws IOException {
+         output.writeByte(object.b.length);
+         if (object.b.length > 0) {
+            output.write(object.b);
+         }
+      }
+
+      @Override
+      public ByteString readObject(ObjectInput input) throws IOException, ClassNotFoundException {
+         short l = input.readByte();
```
slaskawi: Could we use `len` here or something more descriptive? `l` looks like '1' and it doesn't look good in `new byte[1]` :)

core/src/main/java/org/infinispan/util/ByteString.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Set;
+
+/**
```
slaskawi: Could you please add some meaningful javadoc here?

slaskawi:
If I understand this correctly - we are using `ByteString`s everywhere to make everybody aware that some chars might be lost during the conversion? That's the argument why we shouldn't use String all around and convert them to `ByteString`s just before sending through JGroups?

core/src/main/java/org/infinispan/util/ByteString.java
tristantarrant: To abstract the logic. As @slaskawi says below, we might decide we want to use a different encoding, hashing, etc down the line

galderz: Different encoding/hashing...etc can be achieved with a static class that gets the byte[] and produces what you want. I'm not 100% sure we really need a wrapper here, but feel free to go ahead with it, we can always change later.

The only real problem we have with byte[] is equality and hashcode but that's already sorted out with the equivalence functions that are separated out.

tristantarrant:
@slaskawi, no the argument is that String.writeUTF is an extremely expensive op that currently is performed for every CacheRpcCommand. Precomputing a byte[] representation of it once (at Cache construction) ensures we skip that.

core/src/main/java/org/infinispan/util/ByteString.java
tristantarrant: Done

tristantarrant:
@slaskawi  addressed your comments
@galderz I still prefer encapsulation. But if anybody else has arguments against the ByteString class please voice them

slaskawi:
Looks good to me.

slaskawi:
I see CI has failed; let's rerun it.

core/src/main/java/org/infinispan/util/ByteString.java
```diff
@@ -0,0 +1,93 @@
...
+      }
+
+      @Override
+      public ByteString readObject(ObjectInput input) throws IOException, ClassNotFoundException {
+         short len = input.readByte();
+         if (len == 0)
+            return EMPTY;
+
+         byte[] b = new byte[len];
+         input.read(b);
```
pruivo: ```readFully``` is more correct than ```read```

core/src/main/java/org/infinispan/factories/GlobalComponentRegistry.java
```diff
@@ -84,6 +85,7 @@
    final Collection<ModuleLifecycle> moduleLifecycles;
 
    final ConcurrentMap<String, ComponentRegistry> namedComponents = new ConcurrentHashMap<String, ComponentRegistry>(4);
+   final ConcurrentMap<ByteString, ComponentRegistry> byteNamedComponents = new ConcurrentHashMap<ByteString, ComponentRegistry>(4);
```
pruivo: you can keep only the ```byteNamedComponents``` map. ```getNamedComponentRegistry(String)``` is only invoked when a cache is started or stopped, so you can afford the cost of converting the ```String``` => ```ByteString``` in that cases.

core/src/main/java/org/infinispan/util/ByteString.java
```diff
@@ -0,0 +1,93 @@
...
+   public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+      ByteString that = (ByteString) o;
+      return Arrays.equals(b, that.b);
+   }
+
+   @Override
+   public String toString() {
+      return new String(b, CHARSET);
```
pruivo: I'm a little concern about this. wouldn't be better to cache the string too?

core/src/main/java/org/infinispan/commands/tx/VersionedCommitCommand.java
```diff
@@ -20,14 +21,14 @@
    private EntryVersionsMap updatedVersions;
 
    public VersionedCommitCommand() {
-      super("");
+      super(ByteString.emptyString());
```
danberindei: Might as well use `null`, almost all commands use `super(null)` in the "CommandIdUniquenessTest constructor".

core/src/main/java/org/infinispan/util/ByteString.java
danberindei: AFAICT this is only used in `LogFactory.pushNDC(ByteString, boolean)`, and that is only used for `StateRequestCommand`, and `StateResponseCommand`, so it shouldn't be a problem. 

And even those usages could be eliminated by calling `LogFactory.pushNDC()` from the inbound invocation handler for all commands, instead of `InvocationContextInterceptor` for some and `perform()` for others.

query/src/main/java/org/infinispan/query/indexmanager/AbstractUpdateCommand.java
```diff
@@ -70,14 +71,15 @@ public boolean isReturnValueExpected() {
     */
    @Override
    public void fetchExecutionContext(CommandInitializer ci) {
-      if (ci.getCacheManager().cacheExists(cacheName)) {
-         Cache cache = ci.getCacheManager().getCache(cacheName);
+      String name = cacheName.toString();
```
danberindei: I guess I spoke to soon about `ByteString.toString()` not being used... I guess here we need a way for the per-cache inbound invocation handler to pass the `Cache` to the command.

danberindei:
@tristantarrant My ideal solution would be to remove the cache name from `CacheRpcCommand` altogether, and to have only the `RpcManager` and the `PerCacheInboundInvocationHandler` deal with cache references (be they actual names, hashes, or something else). But that would be a "slightly" bigger change :) 

At this point, I like having a `ByteString` reference instead of a raw `byte[]`, I'd just use static methods for serialization/deserialization instead of a full externalizer.

core/src/main/java/org/infinispan/commands/tx/VersionedCommitCommand.java
tristantarrant: Done

query/src/main/java/org/infinispan/query/indexmanager/AbstractUpdateCommand.java
tristantarrant: I am caching the String on first invocation of toString() now

core/src/main/java/org/infinispan/util/ByteString.java
tristantarrant: Done

core/src/main/java/org/infinispan/factories/GlobalComponentRegistry.java
tristantarrant: Done

tristantarrant:
Dropped the Externalizer and introduced the two static methods.

pruivo:
looks good to me :+1:

core/src/main/java/org/infinispan/marshall/core/ExternalizerTable.java
```diff
@@ -106,6 +106,7 @@
 import org.infinispan.transaction.xa.recovery.RecoveryAwareDldGlobalTransaction;
 import org.infinispan.transaction.xa.recovery.RecoveryAwareGlobalTransaction;
 import org.infinispan.transaction.xa.recovery.SerializableXid;
+import org.infinispan.util.ByteString;
```
danberindei: No longer needed ;)

tristantarrant:
@danberindei removed the import

danberindei:
Integrated, thanks Tristan!

PR #4237, opened 2016-04-13
ISPN-6497 Allow keystore and cert. private key passwds to be different

vjuranek:
https://issues.jboss.org/browse/ISPN-6497


tristantarrant:
Nice one, thanks @vjuranek

PR #4240, opened 2016-04-14
Ispn 5342 Refactor Hot Rod Server Code

wburns:
This in preview still.  I have to clean up some comments and verify performance is what it should be first.  The client event stress test works flawlessly for me though.

https://issues.jboss.org/browse/ISPN-5342
https://issues.jboss.org/browse/ISPN-6514


server/hotrod/src/test/scala/org/infinispan/server/hotrod/test/HotRodTestingUtil.scala
```diff
@@ -91,6 +94,19 @@ object HotRodTestingUtil extends Log {
...
          }
+
+         override def getInitializer: ChannelInitializer[Channel] = {
+            // Pass by name since we have circular dependency
+            def getTransport() = {
+               transport
+            }
+            if (configuration.idleTimeout > 0)
+               new HotRodChannelInitializer(this, getTransport(), getEncoder, "test")
+                 with TimeoutEnabledChannelInitializer with SingleByteFrameDecoderChannelInitializer
```
wburns: Need to add a system property to only enable SingleByteFrameDecoderChannelInitializer for tests we want to verify decoder is working.  For perf tests this will be a big burden.

wburns: It might be better to add a HotRodTestingUtil method that is just like the current one but you can call either based on if this decoder is present or not.

server/core/src/main/scala/org/infinispan/server/core/UnsignedNumeric.java
```diff
@@ -0,0 +1,91 @@
...
+      return i;
+   }
+
+   /**
+    * Same as {@link #readUnsignedInt(ByteBuf)} except that it returns an <code>OptionalInt</code> that can be used
+    * when the byte buf is not used in blocking mode.  If an empty value is returned the reader index will be reset.
+    * @param in the byte buf to read from
+    * @return an optional showing whether or not the int could be read
+    */
+   public static OptionalInt readOptionalUnsignedInt(ByteBuf in) {
```
gustavonalle: Not used?

server/core/src/main/scala/org/infinispan/server/core/UnsignedNumeric.java
```diff
@@ -0,0 +1,91 @@
...
+    */
+   public static long readUnsignedLong(ByteBuf in) {
+      byte b = in.readByte();
+      long i = b & 0x7F;
+      for (int shift = 7; (b & 0x80) != 0; shift += 7) {
+         b = in.readByte();
+         i |= (b & 0x7FL) << shift;
+      }
+      return i;
+   }
```
gustavonalle: Not used?

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
```diff
@@ -57,6 +59,151 @@ object ExtendedByteBuf {
...
+     } else {
+       val bytes = new Array[Byte](length)
+       bf.readBytes(bytes)
+       Some(bytes)
+     }
+  }
+
+  def readMaybeSignedInt(bf: ByteBuf): Option[Int] = {
+    readMaybeVInt(bf).map(i => SignedNumeric.decode(i))
+  }
```
gustavonalle: [Minor] simplify to 
```  
def readMaybeSignedInt(bf: ByteBuf): Option[Int] = readMaybeVInt(bf).map(SignedNumeric.decode)
```

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
```diff
@@ -57,6 +59,151 @@ object ExtendedByteBuf {
...
+
+   def readMaybeOptRangedBytes(bf: ByteBuf): Option[Option[Array[Byte]]] = {
```
gustavonalle: Can it return a ```Option[Array[Byte]]``` instead of a  ```Option[Option[Array[Byte]]]```?

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
```diff
@@ -57,6 +59,151 @@ object ExtendedByteBuf {
...
+      val l = readMaybeSignedInt(bf)
+      if (l.isDefined) {
```
gustavonalle: It'd be nicer to use the ```map```operator in ```Option[Int]```

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
```diff
@@ -57,6 +59,151 @@ object ExtendedByteBuf {
...
+         val length = l.get
+         if (length < 0) Some(None) else readMaybeRangedBytes(bf, length).map(b => Some(b))
```
gustavonalle: ```Some(None)``` could be avoided if the return type is flattened (see above comment)

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
```diff
@@ -57,6 +59,151 @@ object ExtendedByteBuf {
...
+    * Reads a string if possible.  If not present the reader index is reset to the last mark.
+    * @param bf
+    * @return
+    */
+   def readMaybeString(bf: ByteBuf): Option[String] = {
+      val bytes = readMaybeRangedBytes(bf)
+      bytes.map(b => if (b.isEmpty) "" else new String(b, CharsetUtil.UTF_8))
+   }
+
+  def readMaybeOptString(bf: ByteBuf): Option[Option[String]] = {
```
gustavonalle: Same as before, does it need to return two layers of ```Option``` ?

server/core/src/main/scala/org/infinispan/server/core/transport/NettyTransport.scala
```diff
@@ -30,14 +30,15 @@ import io.netty.buffer.PooledByteBufAllocator
...
+                     address: InetSocketAddress, val configuration: ProtocolServerConfiguration,
+                     threadNamePrefix: String, cacheManager: EmbeddedCacheManager)
         extends Transport with Log {
 
    private val serverChannels = new DefaultChannelGroup(threadNamePrefix + "-Channels", ImmediateEventExecutor.INSTANCE)
    val acceptedChannels = new DefaultChannelGroup(threadNamePrefix + "-Accepted", ImmediateEventExecutor.INSTANCE)
 
    private val masterGroup = new NioEventLoopGroup(1, new DefaultThreadFactory(threadNamePrefix + "ServerMaster"))
-   private val workerGroup = new NioEventLoopGroup(configuration.workerThreads, new DefaultThreadFactory(threadNamePrefix + "ServerWorker"))
+   private val workerGroup = new NioEventLoopGroup(0, new DefaultThreadFactory(threadNamePrefix + "ServerWorker"))
```
gustavonalle: having this as ```0``` defaults event loop threads to ```Runtime.getRuntime().availableProcessors() * 2``` or ```io.netty.eventLoopThreads``` sys prop, but then ```configuration.workerThreads``` is ignored.

server/core/src/main/scala/org/infinispan/server/core/transport/StatsChannelHandler.java
```diff
@@ -0,0 +1,68 @@
+package org.infinispan.server.core.transport;
```
gustavonalle: another ```.java``` inside ```src/main/scala```? Soon we will need to rename this folder :smile:

server/core/src/test/scala/org/infinispan/server/core/AbstractProtocolServerTest.scala
```diff
@@ -65,6 +64,10 @@ class AbstractProtocolServerTest {
          super.startInternal(configuration, cacheManager)
       }
 
+      override def getInitializer(): ChannelInitializer[Channel] = {
+         return null;
+      }
+
```
gustavonalle: Nitpick:  
``` override def getInitializer = null ```

File: server/hotrod/src/main/scala/org/infinispan/server/hotrod/transport/HotRodChannelInitializer.scala [hunk 12]
```diff
@@ -0,0 +1,29 @@
...
+   override def initChannel(ch: Channel): Unit = {
+      super.initChannel(ch)
+      // Any inbound handler after this point should really be using this executionGroup
+      val executionGroup = new DefaultEventExecutorGroup(transport.configuration.workerThreads,
+         new DefaultThreadFactory(threadNamePrefix + "ServerHandler"))
+      if (server.getConfiguration.authentication().enabled()) {
+         ch.pipeline().addLast(executionGroup, "authentication", new AuthenticationHandler(server))
+      }
+      ch.pipeline.addLast(executionGroup, "handler", new ContextHandler(server, transport))
+      ch.pipeline.addLast(executionGroup, "exception", new HotRodExceptionHandler)
```
gustavonalle: :+1:  I will close https://github.com/infinispan/infinispan/pull/4217 then since you include the extra executor here.

server/hotrod/src/main/java/org/infinispan/server/hotrod/ContextHandler.java
```diff
@@ -0,0 +1,204 @@
...
+            if (server.getClientListenerRegistry().removeClientListener(listenerId, msg.cache())) {
+               writeResponse(msg, ctx.channel(), msg.decoder().createSuccessResponse(h, null));
+            } else {
+               writeResponse(msg, ctx.channel(), msg.decoder().createNotExecutedResponse(h, null));
+            }
+            break;
+         case IterationStartRequest:
+            Tuple4<Option<byte[]>, Option<Tuple2<String, scala.collection.immutable.List<byte[]>>>, Integer, Boolean> iterationStart =
+                    (Tuple4<Option<byte[]>, Option<Tuple2<String, scala.collection.immutable.List<byte[]>>>, Integer, Boolean>) msg.operationDecodeContext();
+
```
gustavonalle: I think after this PR is integrated we should consider gradually move the scala sources to java, otherwise we end up with those ugly casts everywhere.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodExceptionHandler.java
```diff
@@ -0,0 +1,44 @@
...
+      log.debug("Exception caught", t);
+      if (t instanceof DecoderException) {
+         t = t.getCause();
+      }
+      if (t instanceof HotRodException) {
+         // HotRodException is already translated to response
+         ch.writeAndFlush(((HotRodException) t).response());
+      } else {
+         ch.writeAndFlush(decodeCtx.createExceptionResponse(t));
+      }
```
gustavonalle: According to Netty's best practices, when doing a write and flush and not interested in the future, it's better to use ```ch.writeAndFlush(whatever, ch.voidPromise())``` to reduce allocation rates since it reuses the same promise. I wonder if we could adopt it a rule on all writes without breaking anything

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
```diff
@@ -195,9 +195,12 @@ class ClientListenerRegistry(configuration: HotRodServerConfiguration) extends L
...
    def findAndWriteEvents(channel: Channel): Unit = {
-      eventSenders.values().collectFirst { case s: BaseClientEventSender =>
-         if(s.hasChannel(channel)) s.writeEventsIfPossible()
-      }
+      // Make sure we write any event in main event loop
+      channel.eventLoop().execute(new Runnable {
+         override def run(): Unit = eventSenders.values().collectFirst {
+            case s: BaseClientEventSender => if (s.hasChannel(channel)) s.writeEventsIfPossible()
+         }
+      })
```
gustavonalle: @galderz I think that with this change the ClientStressTest would pass on master, wouldn't it?

server/hotrod/src/test/scala/org/infinispan/server/hotrod/test/HotRodTestingUtil.scala
gustavonalle: +1 for the second option

gustavonalle:
@wburns Looking good :+1:

server/core/src/main/scala/org/infinispan/server/core/transport/NettyChannelInitializer.scala
```diff
@@ -15,12 +18,14 @@ import io.netty.handler.ssl.SslHandler
...
  */
-class NettyChannelInitializer(server: ProtocolServer,
-                                  encoder: ChannelOutboundHandler)
+class NettyChannelInitializer(server: ProtocolServer, transport: => NettyTransport,
+                                       encoder: ChannelOutboundHandler)
       extends ChannelInitializer[Channel] {
 
    override def initChannel(ch: Channel): Unit = {
       val pipeline = ch.pipeline
+//      pipeline.addLast("netty-logging", new LoggingHandler(LogLevel.ERROR))
```
slaskawi: Just a heads up - there will be a clash with my PR here: https://github.com/infinispan/infinispan/pull/4247

server/core/src/main/scala/org/infinispan/server/core/UnsignedNumeric.java
wburns: I added it to mirror the other UnsignedNumeric in core since we had both Int and Long.  I figured it is easiest to add it now when I was actually thinking about it rather than later, and it can be used by others if needed.

server/core/src/main/scala/org/infinispan/server/core/UnsignedNumeric.java
wburns: Same thing as the previous.

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
wburns: Sounds good.  Some of the lambda stuff from Java 8 seems to work in scala and others didn't.

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
wburns: Unfortuantely no.  The problem is that we have an optional range of bytes and then we also have the possibility that the bytes weren't provided, so we have a double layered option.  We can't handle this with a single Option cause we need a "not enough bytes state", "no byte array provided" and a "byte array provided state"

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
wburns: I agree, the problem is we can't :-1: I later found I could implement this using flatMap such as https://github.com/infinispan/infinispan/pull/4240/files#diff-a2d4a96cb76d16183df8b36680404637R346, if you want I can do that instead?  To be honest I am not sure which one is nicer.

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
wburns: This is handled by the other comments.

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
wburns: Unfortunately yes, check other comment.

server/core/src/main/scala/org/infinispan/server/core/transport/NettyTransport.scala
wburns: :)

server/core/src/main/scala/org/infinispan/server/core/transport/StatsChannelHandler.java
wburns: The reason I did so was because the whole static variable thing in scala is not a very nice solution, imo.  I have another handler that I will be adding that will use these variables.

server/core/src/test/scala/org/infinispan/server/core/AbstractProtocolServerTest.scala
wburns: Sure

server/hotrod/src/main/scala/org/infinispan/server/hotrod/transport/HotRodChannelInitializer.scala
wburns: That was my thought :+1:

server/hotrod/src/main/java/org/infinispan/server/hotrod/ContextHandler.java
wburns: I agree.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodExceptionHandler.java
wburns: Yeah we should make sure to change all of them then I am guessing.

gustavonalle: Yes, all of them, to avoid avoid unnecessary allocations every time flush is called

wburns: Thinking I should log another JIRA for that, I can include it in this PR though.

gustavonalle: np

wburns: https://issues.jboss.org/browse/ISPN-6514

server/core/src/main/scala/org/infinispan/server/core/transport/NettyChannelInitializer.scala
wburns: Thanks, will keep it in mind.

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
galderz: Hmmm, in that case I'd try to create a sealed trait and 3 case classes that represent the 3 different possible states. Sealed traits are often used as alternative to enums (see [here](http://stackoverflow.com/questions/11203268/what-is-a-sealed-trait))

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
galderz: The norm is to avoid checking whether an Option is defined and instead use high order functions. The only situation where I'd check if Option is defined is if using high order functions would confuse the reader, but map/flatMap/filter/foldR/foldL are common enough that any Scala developer should know about them.

wburns: I agree they are the norm, the problem is the logic for the whole segment was quite a bit confusing imo and took quite a bit of tweaking to get it right.  It might just be because I wasn't used to scala as much though.

server/hotrod/src/main/java/org/infinispan/server/hotrod/ContextHandler.java
galderz: Yup, I'm fine with that too. Bye bye Scala 😢 

^ Not that fuzzed actually 😄

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
galderz: @gustavonalle Do you read my emails? 😉 

Indeed, as mentioned in the email, this is the change that made the difference.

galderz: Actually, this change, together with separating cache ops to different executor (your idea), so both ideas combined 😄

server/hotrod/src/main/scala/org/infinispan/server/hotrod/Decoder2x.scala
```diff
@@ -234,209 +217,185 @@ object Decoder2x extends AbstractVersionedDecoder with ServerConstants with Log
...
-         for (i <- 0 until numParams) params += readRangedBytes(buffer)
-         params.toList
-      } else List.empty
+   private def readOptionalParams(buffer: ByteBuf): Option[List[Bytes]] = {
+      val numParams = readMaybeByte(buffer)
+      numParams.map(p => {
+         if (p > 0) {
+            val params = ListBuffer[Bytes]()
+            @tailrec def readParams(buf: ListBuffer[Bytes]): Boolean = {
+               if (readMaybeRangedBytes(buffer).exists(param => {
```
galderz: Taking in account what I said above WRT sealed traits, I think it'd make this whole section a lot more readable by pattern matching on the 3 possible outcomes.

galderz:
Great work @wburns, minor comments except for suggesting using sealed traits to make it easier to distinguish between different possibilities.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
gustavonalle: @galderz sorry, I read the email after commenting here :smile:

wburns:
Rebased, addressed comments and added a trait for the Option[Option[Array[Byte]]] to reduce complexity.

Hoping CI runs better this time, got some socket in use errors last time.  I ran the full test suite locally and haven't encountered these so I don't know if they are CI server related or not.

wburns:
Looks like the auth changes need to be tweaked still

http://ci.infinispan.org/viewLog.html?buildId=38802&buildTypeId=bt9&guest=1

wburns:
Authentication tests should now be fixed.  Basically all cache operations should be done in the ContextHandler and not in the Encoder or Decoder as they won't have authorization.

wburns:
Tests should be all good now.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/OperationResponse.scala
```diff
@@ -52,32 +51,9 @@ object OperationResponse extends Enumeration {
...
-         case AuthRequest => AuthResponse
-         case AddClientListenerRequest => AddClientListenerResponse
-         case RemoveClientListenerRequest => RemoveClientListenerResponse
-         case ExecRequest => ExecResponse
-         case PutAllRequest => PutAllResponse
-         case GetAllRequest => GetAllResponse
-      }
+   def toResponse(request: NewHotRodOperation): OperationResponse = {
+      // Go to java so switch case will be optimized properly
+      OperationResponseJava.operationToResponse(request).asInstanceOf[OperationResponse]
```
tristantarrant: This is hilarious

wburns: Yeah scala Enumerations do not play nicely with Java at all :-1:

server/hotrod/src/main/scala/org/infinispan/server/hotrod/NewHotRodOperation.java
```diff
@@ -0,0 +1,150 @@
+package org.infinispan.server.hotrod;
+
+/**
+ * Enumeration defining all of the possible hotrod operations
+ *
+ * @author wburns
+ * @since 9.0
+ */
+public enum NewHotRodOperation {
```
gustavonalle: Can it just be called ```HotRodOperation```?

wburns: Yeah sorry I can rename this and the Decoder, forgot to :)

server/hotrod/src/main/java/org/infinispan/server/hotrod/NewHotRodDecoder.java
```diff
@@ -0,0 +1,321 @@
...
+import java.util.OptionalLong;
+import java.util.function.Predicate;
+
+/**
+ * Decoder that will decode hotrod messages and then send a {@link CacheDecodeContext} down the pipeline.
+ *
+ * @author wburns
+ * @since 9.0
+ */
+public class NewHotRodDecoder extends ByteToMessageDecoder {
```
gustavonalle: Can it be name ```HotRodDecoder```?

server/hotrod/src/main/java/org/infinispan/server/hotrod/ResponseWriting.java
```diff
@@ -0,0 +1,52 @@
...
+import org.infinispan.server.hotrod.logging.JavaLog;
+import scala.Tuple2;
+
+import java.security.PrivilegedActionException;
+
+/**
+ * Static helper to provide common way of writing response to channel
+ *
+ * @author wburns
+ * @since 4.0
```
gustavonalle: 9.0

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientListenerRegistry.scala
```diff
@@ -211,16 +214,16 @@ class ClientListenerRegistry(configuration: HotRodServerConfiguration) extends L
            extends BaseClientEventSender(ch, listenerId, version, targetEventType)
 
    private abstract class BaseClientEventSender(ch: Channel, listenerId: Bytes, version: Byte, targetEventType: ClientEventType) {
-      val eventQueue = new LinkedBlockingQueue[AnyRef](100)
+      val eventQueue = new LinkedBlockingQueue[AnyRef](50)
```
gustavonalle: Should we make this configurable since we are here? :smile:

wburns: Whoops I didn't mean to change that, let me change it back.  We can cover this in a different PR if we want to make it configurable.

gustavonalle:
@wburns LGTM, nice work :clap:

wburns:
Rebasing, unfortunately Tristan's commit has made this all conflicted now.

server/core/src/main/scala/org/infinispan/server/core/transport/ExtendedByteBuf.scala
```diff
@@ -72,4 +214,25 @@ object ExtendedByteBuf {
...
+  }
+
+  case object MoreBytesForString extends MaybeStringTrait
+  case object StringNotPresent extends MaybeStringTrait
+
+  final case class PresentString(value: String) extends MaybeStringTrait {
+    override def getValue(): String = value
+  }
+
+  sealed trait MaybeBytesTrait {
```
galderz: This is a bit repetitive, you could take advantage of generics to define a single type (e.f. `Data[T]`) and then create `Data[Bytes]` and `Data[String]` versions, but we can always refactor this later on.

galderz:
@wburns Could you rebase and reword my commit to add stress test to include ISPN-6478 in the title?

wburns:
Changed commit message to have JIRA

gustavonalle:
pulling

gustavonalle:
integrated, thanks @wburns !

PR #4245, opened 2016-04-15
ISPN-6526 Transaction doesn't rollback on throwable

d0k1:
https://issues.jboss.org/browse/ISPN-6526

### DIFF
--- a/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
+++ b/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
@@ -88,6 +88,7 @@ import javax.transaction.Transaction;
 import javax.transaction.TransactionManager;
 import javax.transaction.xa.XAResource;
 
+import java.lang.Throwable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


slaskawi:
Hey @d0k1! Could you tell me when and how did you encounter for this problem? And of course what `Throwable` got through this try/catch statement?

d0k1:
Hi @slaskawi, thanks for reply.
So, about my problem. Sometimes thread can get a `java.lang.ThreadDeath` or `java.lang.OutOfMemoryError`. And once upon a time, one thread got such throwable and left object in cache locked until JVM was restarted. So, I thought it might be not bad to handle such case. 
I suppose when IS rollbacks current tx, it would release any lock it holds. Is that correct?

slaskawi:
@d0k1 That makes sense to me. Let's also ask @pruivo for opinion.

pruivo:
@slaskawi looks good to me.
@d0k1 do you mind to open a JIRA (https://issues.jboss.org/secure/Dashboard.jspa) and update the commit message?

d0k1:
@pruivo I've made an JIRA issue (https://issues.jboss.org/browse/ISPN-6526), modified commit message, made a rebase. And modified commit's code a little, to get it more readable (just import `java.lang.Throwable`).
Is there anything else to do?

slaskawi:
Pulling

slaskawi:
Integrated thanks!

PR #4246, opened 2016-04-15
ISPN-6411 Do not marshall byte[] returns for exec

galderz:
https://issues.jboss.org/browse/ISPN-6411


tristantarrant:
Pushed to master

PR #4249, opened 2016-04-18
ISPN-5900 CDI split up (common, remote and embedded)

slaskawi:
https://issues.jboss.org/browse/ISPN-5900

This is the final part of CDI refactoring - the split between embedded, remote and common. Now all modules have their own packages so there will be no clashes in OSGi. 

However CDI embedded integration is not backwards compatible when using JBoss Modules (one will need to switch from `org.infinispan.cdi` to `org.infinispan.cdi.embedded` dependency declaration).


galderz:
Needs rebasing

galderz:
> However CDI embedded integration is not backwards compatible when using JBoss Modules (one will need to switch from org.infinispan.cdi to org.infinispan.cdi.embedded dependency declaration).

^ Has this been agreed? I don't recall a thread in infinispan-dev list.

slaskawi:
Yes it was discussed long time ago but no official summary was been sent. Probably this would be a good moment to do it.

The best summary might be found in [BZ1266832](https://bugzilla.redhat.com/show_bug.cgi?id=1266832#c9). So during the split of CDI module we moved all remote stuff to `org.infinispan.cdi.remote` package and left common and embedded in `org.infinispan.cdi` (we didn't want to break embedded cache use cases). However in that case we have `infinispan-common` and `infinispan-embedded` exposing the same package, which might break OSGi use cases. So in order to do things correctly, we need to shuffle packages one more time.

//cc @tristantarrant

slaskawi:
Rebasing

slaskawi:
*[as-modules/client/pom.xml, line 34 \[r1\]](https://reviewable.io:443/reviews/infinispan/infinispan/4249#-KFibFbn5lenVC6tVaw3:-KFibFbn5lenVC6tVaw4:1664162474) ([raw file](https://github.com/infinispan/infinispan/blob/1c0a165c92d33c685cfb9f44b6826b4f39f7804d/as-modules/client/pom.xml#L34)):*
Just testing, please ignore this comment

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/infinispan/infinispan/4249)*

danberindei:
@slaskawi Can't we keep the old annotations as deprecated? It would be nice to have a major version with a documented alternative in place before forcing the users to change their code.

slaskawi:
@danberindei Those annotations should be used only internally (the code was refactored at some point between ISPN 7 and 8 if I recall correctly). Now the clients need to produce Configuration or Embedded/RemoteCacheManager, and that's all. [Here is a nice example.](https://github.com/infinispan/infinispan/blob/master/cdi/embedded/src/test/java/org/infinispan/cdi/test/cachemanager/registration/Config.java)

danberindei:
@slaskawi I see, It's fine by me then.

tristantarrant:
Two cdi/jcache errors: 
CacheRemoveAllInterceptorTest.arquillianBeforeTest 
CacheResultInterceptorTest.arquillianBeforeTest

slaskawi:
@tristantarrant After the rebase I couldn't reproduce those failures. Could you please make sure that you rebuilt all modules? Server build needs to be in sync with updated CDI integration modules.

tristantarrant:
I'm getting consistent failures in JCacheFailoverIT.testRemoteListener:63 » Transport Could not fetch transport

slaskawi:
Those failures were probably fixed by https://github.com/infinispan/infinispan/pull/4258

slaskawi:
Rebased

tristantarrant:
Pushed to master. Thanks

PR #4250, opened 2016-04-18
ISPN-6511 Allocate less memory during RPC response handling

danberindei:
https://issues.jboss.org/browse/ISPN-6511

Porperly size the map of responses.


galderz:
Integrated, thanks Dan!

PR #4251, opened 2016-04-18
ISPN-6513 Inconsistencies in Maven pom files

mgencur:
https://issues.jboss.org/browse/ISPN-6513

* Removed repository definitions from pom files as this is generally not recommended
* Added missing modules to dependency management section in the bom file

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -141,6 +141,11 @@
          </dependency>
          <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>infinispan-cdi-common</artifactId>
+            <version>${project.version}</version>
+         </dependency>
+         <dependency>
+            <groupId>${project.groupId}</groupId>
             <artifactId>infinispan-cdi-embedded</artifactId>
             <version>${project.version}</version>
          </dependency>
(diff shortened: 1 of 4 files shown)


bom/pom.xml
```diff
@@ -166,6 +181,56 @@
          </dependency>
          <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>infinispan-distexec-demo</artifactId>
+            <version>${project.version}</version>
+         </dependency>
+         <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>infinispan-gridfs-webdav</artifactId>
+            <version>${project.version}</version>
```
tristantarrant: We should not include demos and examples in the bom.

mgencur: Why? When someone wants to try the demo they probably want the right version of dependencies...

tristantarrant: If they want to build the demo, they build it as part of the tree, not import it as a dependency.

slaskawi: Actually Martin has a point here. With our [new SNAPSHOT deployment](https://repository.jboss.org/nexus/content/repositories/snapshots/org/infinispan/infinispan-bom/9.0.0-SNAPSHOT) (which deploys new SNAPSHOTs whenever we have a stable build) it is perfectly legal to build demos without other modules (of course whether it's sensible or not is another story...)

mgencur:
Updated.

tristantarrant:
Pushed, thanks

PR #4255, opened 2016-04-21
SNI support

slaskawi:
https://issues.jboss.org/browse/ISPN-5721

Tasks: 
* https://issues.jboss.org/browse/ISPN-6517
* https://issues.jboss.org/browse/ISPN-6520
* https://issues.jboss.org/browse/ISPN-6523
* https://issues.jboss.org/browse/ISPN-6524
* https://issues.jboss.org/browse/ISPN-6525

Highlights
* Netty upgrade to `4.1.0.CR7`
* Added `SniHandler` to Hotrod server with proper configuration bits:
```
builder.ssl.enable()
   .sniHostName("www.example.com")
      .keyStoreFileName("some_file")
      .keyStorePassword("password")
   .sniHostName("www.example2.com")
      ...
```
* Added Hotrod server XML configuration:
```
<hotrod-connector name="hotrod5" socket-binding="hotrod" cache-container="default">
    <authentication security-realm="other">
         <sni host-name="example.com" security-realm="other" />
         <sni host-name="example2.com" security-realm="other2" />
     </authentication>
</hotrod-connector>
```
* The client decides to which SNI domain it wants to connect using [SSLParameters#setServerNames] (https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLParameters.html#setServerNames-java.util.List-) method. 
* Added SNI to the client
```
 builder.security().ssl()
   .sslContext(cont)
   .sniHostName("sni")
   .enable();
```


slaskawi:
Fixed test failures

server/hotrod/src/test/resources/create_certs.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+echo "==== Removing old files ===="
+find . -name "*.jks" -exec rm -rf {} \;
+
+echo "==== Generating private keys ===="
+keytool -genkey -noprompt -trustcacerts -keyalg RSA -alias "default" -dname "CN=HotRod_1, OU=Infinispan, O=JBoss, L=Red Hat, ST=World, C=WW" -keypass "secret" -storepass "secret" -keypass "secret2" -keystore "password_server_keystore.jks"
```
tristantarrant: While we're here, we might as well introduce a more articulate test keystore using the following hierarchy
Root CA > CA > Certificates*

slaskawi: I have a mixed feelings about it. On one hand - yes, this test would be really descriptive but on the other hand we should test Infinispan bits not OpenJDK SSL support. 

Anyway, I created https://issues.jboss.org/browse/ISPN-6541

@vjuranek any thoughts?

vjuranek: well, I have no strong opinion, but I'd incline to @tristantarrant proposal. Tests are sometimes source of knowledge how to configure some functionality and for security setup, which is quite often source of confusion, complex/descriptive tests can have this added value that in case of questions/confusion we can easily point users to the test suite to explain how to use/setup given feature.

server/integration/endpoint/src/main/resources/schema/jboss-infinispan-endpoint_9_0.xsd
```diff
@@ -365,6 +368,26 @@
         </xs:attribute>
     </xs:complexType>
 
+    <xs:complexType name="sni">
```
slaskawi: Critical thing to review

server/integration/endpoint/src/test/resources/org/infinispan/server/endpoint/endpoint-9.0.xml
```diff
@@ -18,6 +18,12 @@
         </sasl>
       </authentication>
    </hotrod-connector>
+    <hotrod-connector name="hotrod5" socket-binding="hotrod" cache-container="default">
```
slaskawi: And another critical thing to look at

slaskawi:
Rerunning CI, I'm 100% sure I fixed those failures... maybe I didn't commit something...

server/integration/endpoint/src/test/resources/org/infinispan/server/endpoint/endpoint-9.0.xml
slaskawi: I've been playing with this for a while and I think the SNI part should be moved under `authentication`.

slaskawi:
Clarified - The SNI configuration must be placed under encryption because we enable SSL there. I will open another PR with previous version (and close this one for historical reasons).

//cc @tristantarrant

PR #4257, opened 2016-04-22
ISPN-6513 Updated repositories to public-jboss

slaskawi:
This is an additional fix to https://github.com/infinispan/infinispan/pull/4251

Some of the artifacts (like [1]) are present only in `public-jboss` repository.

[1] https://repository.jboss.org/nexus/content/groups/public-jboss/com/thoughtworks/xstream/xstream/1.4.9/

### DIFF
--- a/maven-settings.xml
+++ b/maven-settings.xml
@@ -13,7 +13,7 @@
                 <repository>
                     <id>jboss-public-repository-group</id>
                     <name>JBoss Public Maven Repository Group</name>
-                    <url>https://repository.jboss.org/nexus/content/groups/public</url>
+                    <url>https://repository.jboss.org/nexus/content/groups/public-jboss</url>
                     <layout>default</layout>
                     <releases>
                         <enabled>true</enabled>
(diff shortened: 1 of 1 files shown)


slaskawi:
BTW I have updated CI agents to use `public-jboss` to unblock builds

danberindei:
@slaskawi Couldn't we remove the dependency on xstream instead? :D

danberindei:
Integrated, thanks Sebastian!

mgencur:
Hey, did you know that the Creaper dependency is only in the "public" repository?

danberindei:
@slaskawi @mgencur I cleaned up my local repository and I was able to build everything with the "public" repository, so I think it's safe to revert this. xstream is definitely present in Maven Central, so we don't need an additional repository for that one. Speaking of which, didn't Maven Central have a rule that all dependencies of artifacts uploaded there (e.g. infinispan) must not depend on artifacts that are not in Maven Central?

PR #4263, opened 2016-04-25
ISPN-6406 Null exec return is empty byte[]

galderz:
https://issues.jboss.org/browse/ISPN-6406

* A Java client will return null when encountering byte[].


server/hotrod/src/main/java/org/infinispan/server/hotrod/ContextHandler.java
```diff
@@ -99,7 +99,8 @@ protected void channelRead0(ChannelHandlerContext ctx, CacheDecodeContext msg) t
             byte[] result = (byte[]) taskManager.runTask(execContext.name(),
                     new TaskContext().marshaller(marshaller).cache(msg.cache()).parameters(execContext.params())).get();
             writeResponse(msg, ctx.channel(),
-                    new ExecResponse(h.version(), h.messageId(), h.cacheName(), h.clientIntel(), h.topologyId(), result));
+                    new ExecResponse(h.version(), h.messageId(), h.cacheName(), h.clientIntel(), h.topologyId(),
+                          result == null ? new byte[]{} : result));
```
wburns: Have to admit I am struggling if we should put this type of change in the Encoder instead, but this should be fine.

wburns:
Pulling...

wburns:

PR #4264, opened 2016-04-25
ISPN-20/ISPN-701: JDBC re-factoring and Upsert support

ryanemerson:
https://issues.jboss.org/browse/ISPN-20
https://issues.jboss.org/browse/ISPN-701

In this solution, upsert is enabled by default for JdbcStringBasedStores. Upsert is only ignored if the user explicitly requests it in the configuration, or if the selected DB and version is known not to support the operation (e.g. Postgres versions < 9.5).


server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/Attribute.java
```diff
@@ -67,6 +67,9 @@
     DATASOURCE(ModelKeys.DATASOURCE),
     DEFAULT_CACHE(ModelKeys.DEFAULT_CACHE),
     @Deprecated DEFAULT_CACHE_CONTAINER("default-cache-container"),
+    DB_DISABLE_UPSERT(ModelKeys.DB_DISABLE_UPSERT),
```
tristantarrant: Not sure I like having a dedicated configuration option for this.
Better have it as a "known" property (to be specified with the <properties> element).

ryanemerson: Sure, that makes sense. Is there a list of known System properties anywhere? Or a class/convention for declaring them?

ryanemerson:
@tristantarrant Upsert is now disabled via ```<property name="jdbc.upsert.disabled">true</property>```

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/table/management/TableManagerFactory.java
```diff
@@ -0,0 +1,162 @@
...
+
+      if (LOG.isDebugEnabled())
+         LOG.debugf("Guessing database dialect as '%s'.  If this is incorrect, please specify the correct dialect using the 'dialect' attribute in your configuration.  Supported database dialect strings are %s", databaseType, Arrays.toString(DatabaseType.values()));
+      return new DbMetaData(databaseType, majorVersion, minorVersion, disableUpsert);
+   }
+
+   private static DatabaseType guessDialect(String name) {
+      DatabaseType type = null;
+      if (name != null) {
+         if (name.toLowerCase().contains("mysql")) {
```
tristantarrant: We can convert this to use a switch statement with strings

tristantarrant:
You also need to amend the XSDs both for embedded and for server.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/table/management/TableManagerFactory.java
ryanemerson: I don't think we can use a Switch statement here, as we are testing that the String contains another String, not direct equality.

tristantarrant: Right. You might want to extract name.toLowerCase() to a local var

tristantarrant:
Thanks, pushed to master

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/table/management/PostgresTableManager.java
```diff
@@ -0,0 +1,73 @@
...
+         deleteRowSql = String.format("DELETE FROM %s WHERE %s = cast(? as %s)",
+                                      getTableName(), config.idColumnName(), config.idColumnType());
+      }
+      return deleteRowSql;
+   }
+
+   @Override
+   public boolean isUpsertSupported() {
+      // ON CONFLICT added in Postgres 9.5
+      return super.isUpsertSupported() && metaData.getMajorVersion() > 8 && metaData.getMinorVersion() > 4;
```
danberindei: @ryanemerson Isn't this going to fail with Postgres 10.0?

ryanemerson: @danberindei Argh good point! I will issue a new PR to fix this and SqlLiteTableManager

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/table/management/SQLiteTableManager.java
```diff
@@ -0,0 +1,34 @@
...
+   private static final Log LOG = LogFactory.getLog(SQLiteTableManager.class, Log.class);
+
+   SQLiteTableManager(ConnectionFactory connectionFactory, TableManipulationConfiguration config, DbMetaData metaData) {
+      super(connectionFactory, config, metaData, LOG);
+   }
+
+   @Override
+   public boolean isUpsertSupported() {
+      // OR/ON CONFLICT introduced in 3.8.11
+      return super.isUpsertSupported() && metaData.getMajorVersion() > 2 && metaData.getMinorVersion() > 7;
```
danberindei: This will return false with version 4.0. And since the functionality was not available in version 3.8, you should also check the micro version - or if not possible, avoid supporting any 3.8.x version.

And finally, I would prefer `>=` comparisons, so that you don't need a comment to explain what version you actually want.

PR #4265, opened 2016-04-25
ISPN-6542 Subclasses of AbstractWriteKeyCommand don't marshal the Par…

rvansa:
…ams params field

* marshall params
* pass params to all write/read-write functional commands
* don't ignore some commands in CacheWriterInterceptor
* add Dummy store to functional API tests

### DIFF
--- a/core/src/main/java/org/infinispan/functional/impl/Params.java
+++ b/core/src/main/java/org/infinispan/functional/impl/Params.java
@@ -3,10 +3,17 @@ package org.infinispan.functional.impl;
 import org.infinispan.commons.api.functional.Param;
 import org.infinispan.commons.api.functional.Param.FutureMode;
 import org.infinispan.commons.api.functional.Param.PersistenceMode;
+import org.infinispan.commons.marshall.AdvancedExternalizer;
 import org.infinispan.commons.util.Experimental;
+import org.infinispan.marshall.core.Ids;
 
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Supplier;
@@ -31,6 +38,10 @@ public final class Params {
       FutureMode.defaultValue(),
       PersistenceMode.defaultValue(),
    };
+   // TODO: as Params are immutable and there's only limited number of them,
+   // there could be a table with all the possible combinations and we
+   // wouldn't have to allocate at all
+   private static final Params DEFAULT_INSTANCE = new Params(DEFAULTS);
 
    final Param<?>[] params;
 
@@ -77,7 +88,7 @@ public final class Params {
    }
 
    public static Params create() {
-      return new Params(DEFAULTS);
+      return DEFAULT_INSTANCE;
    }
 
    public static Params from(Param<?>... ps) {
@@ -108,4 +119,35 @@ public final class Params {
       }
    }
 
+   public static class Externalizer implements AdvancedExternalizer<Params> {
+      public static Externalizer INSTANCE = new Externalizer();
+
+      @Override
+      public Set<Class<? extends Params>> getTypeClasses() {
+         return Collections.singleton(Params.class);
+      }
+
+      @Override
+      public Integer getId() {
+         return Ids.PARAMS;
+      }
+
+      @Override
+      public void writeObject(ObjectOutput output, Params params) throws IOException {
+         // There's no point in sending FutureMode over wire
+         output.writeInt(((PersistenceMode) params.get(PersistenceMode.ID).get()).ordinal());
+      }
+
+      @Override
+      public Params readObject(ObjectInput input) throws IOException, ClassNotFoundException {
+         PersistenceMode persistenceMode = PersistenceMode.values()[input.readInt()];
+         if (persistenceMode == PersistenceMode.defaultValue()) {
+            return DEFAULT_INSTANCE;
+         } else {
+            Param[] params = Arrays.copyOf(DEFAULTS, DEFAULTS.length);
+            params[PersistenceMode.ID] = persistenceMode;
+            return new Params(params);
+         }
+      }
+   }
 }
--- a/core/src/main/java/org/infinispan/marshall/core/ExternalizerTable.java
+++ b/core/src/main/java/org/infinispan/marshall/core/ExternalizerTable.java
@@ -67,6 +67,7 @@ import org.infinispan.filter.KeyValueFilterAsKeyFilter;
 import org.infinispan.filter.NullValueConverter;
 import org.infinispan.functional.impl.EntryViews;
 import org.infinispan.functional.impl.MetaParams;
+import org.infinispan.functional.impl.Params;
 import org.infinispan.marshall.exts.*;
 import org.infinispan.metadata.EmbeddedMetadata;
 import org.infinispan.metadata.impl.InternalMetadataImpl;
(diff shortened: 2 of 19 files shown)


rvansa:
@galderz Please review.

ksobolew:
FYI: there's one more command that is not handled in CacheWriterInterceptor but should be.

diff --git a/core/src/main/java/org/infinispan/interceptors/CacheWriterInterceptor.java b/core/src/main/java/org/infinispan/interceptors/CacheWriterInterceptor.java
index c9672a4..f6a0b5f 100755
--- a/core/src/main/java/org/infinispan/interceptors/CacheWriterInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/CacheWriterInterceptor.java
@@ -10,6 +10,7 @@
 import org.infinispan.commands.functional.ReadWriteManyCommand;
 import org.infinispan.commands.functional.ReadWriteManyEntriesCommand;
 import org.infinispan.commands.functional.WriteOnlyKeyCommand;
+import org.infinispan.commands.functional.WriteOnlyKeyValueCommand;
 import org.infinispan.commands.functional.WriteOnlyManyCommand;
 import org.infinispan.commands.functional.WriteOnlyManyEntriesCommand;
 import org.infinispan.commands.tx.CommitCommand;
@@ -230,6 +231,11 @@ public Object visitWriteOnlyKeyCommand(InvocationContext ctx, WriteOnlyKeyComman
       return visitWriteCommand(ctx, command);
    }
 
+   @Override
+   public Object visitWriteOnlyKeyValueCommand(InvocationContext ctx, WriteOnlyKeyValueCommand command) throws Throwable {
+      return visitWriteCommand(ctx, command);
+   }
+
    private <T extends DataWriteCommand & ParamsCommand> Object visitWriteCommand(InvocationContext ctx, T command) throws Throwable {
       Object retval = invokeNextInterceptor(ctx, command);
       if (!isStoreEnabled(command) || ctx.isInTxScope() || !command.isSuccessful()) return retval;

rvansa:
@ksobolew thanks, an oversight.

core/src/main/java/org/infinispan/commands/functional/ManyDataCommand.java
```diff
@@ -0,0 +1,10 @@
+package org.infinispan.commands.functional;
+
+import java.util.Set;
+
+/**
+ * @author Radim Vansa &lt;rvansa@redhat.com&gt;
```
slaskawi: Some meaningful docs here could be useful.

slaskawi:
I would only request this one small javadoc to be added. And could you please rebase it? I'm not sure if those CI failures are related or not.

Could you tell me what are the performance implication of this? I'm especially interested how much do we gain (?) with ordinal and how much do we loose with sending an object.

core/src/main/java/org/infinispan/commands/functional/ManyDataCommand.java
rvansa: It's all it is, just a superinterface for the functional commands that operate on multiple keys.

rvansa:
@slaskawi I've rebased it this morning and Github tells me there're no conflicts. The CI failures are IMO unrelated, this affects only Functional API tests.

I haven't benchmarked anything, I just thought about the most effective way to read/write the params (space-wise), given that those already have their disctinct IDs (so we don't have to register another ones for AdvancedExternalizers). So we can spare the lookup as well.
I don't think the actual effect will be visible.

core/src/main/java/org/infinispan/commands/functional/ManyDataCommand.java
danberindei: @rvansa That would work as a javadoc :)

danberindei:
@rvansa looks good to me
@galderz do you want to have a look as well?

rvansa:
@galderz commented on some things in #4303, I'll update this (and remove the ManyDataCommand)

commons/src/main/java/org/infinispan/commons/api/functional/Param.java
```diff
@@ -44,6 +44,29 @@
    P get();
 
    /**
+    * Should this paramater go over the wire?
+    */
+   SerializeMethod serializeMethod();
```
galderz: This is not how we deal with an objects serialization/marshalling characteristics. Instead, if a param instance needs to be serialized, an Externalizer should be provided for it (examples in the codebase) and then the Externalizer it's either caught by annotation processing (e.g. @SerializedWith) or it's preregistered. The externalizer itself can decide how it serialize it using whichever method it wants. A param implementation, if you want something lazy, can implement Serializable and that's it. So, there's no need to add anything to the Param interface, which has been designed with the bare minimum required to do its job.

commons/src/main/java/org/infinispan/commons/api/functional/Param.java
```diff
@@ -44,6 +44,29 @@
...
+
+   /**
+    * The way how this object should be serialized;
+    */
+   enum SerializeMethod {
```
galderz: Also, whether the param itself should be sent over the wire or not is not down the parameter itself, but more to the command being executed. The command itself must decide whether it includes certain params inside its externalizer payload. Params themselves do not live on their own.

commons/src/main/java/org/infinispan/commons/api/functional/Param.java
galderz: Cross posting what I said in #4303. My feeling is that overall this method and enum pollute the interface itself when there's some clear ways to avoid the issue without the need to add extra methods.

galderz: Sorry, we're already discussing this in #4303, let's continue there.

ksobolew: I'm afraid I'm responsible for the mixup :) I'll explain what happened so other can learn from my mistakes ;) Basically I should've created separate PR for the backport of this PR to 8.2.x, and then base my "proposed fix" on that. But at first I thought they could be independent of each other so I only created one PR. Later on this proved to be untrue. I did create separate branches for both but it was too late to change the base branch of the PR and the backport shows as part of my PR.

rvansa:
@danberindei @galderz @slaskawi @ksobolew Updated.

commons/src/main/java/org/infinispan/commons/api/functional/Param.java
rvansa: @ksobolew Don't worry about the mixup, as long as you keep those as separate commits, so you can easily rebase your changes. And thanks for the good work :)

galderz:
@rvansa Thanks! Looking... :)

galderz:
Looks great @rvansa, apologies for the delay getting around to reviewing this. Integrating...

galderz:
Integrated, thanks @rvansa!

core/src/main/java/org/infinispan/functional/impl/Params.java
```diff
@@ -108,4 +119,35 @@ public static Params from(Param<?>... ps) {
...
+
+      @Override
+      public void writeObject(ObjectOutput output, Params params) throws IOException {
+         // There's no point in sending FutureMode over wire
+         output.writeInt(((PersistenceMode) params.get(PersistenceMode.ID).get()).ordinal());
+      }
+
+      @Override
+      public Params readObject(ObjectInput input) throws IOException, ClassNotFoundException {
+         PersistenceMode persistenceMode = PersistenceMode.values()[input.readInt()];
```
danberindei: `values()` should be cached, because it allocates a new array every time.

core/src/main/java/org/infinispan/marshall/core/ExternalizerTable.java
```diff
@@ -384,6 +385,8 @@ private void loadInternalMarshallables() {
       addInternalExternalizer(new MarshallableFunctionExternalizers.LambdaWithMetasExternalizer());
       addInternalExternalizer(new MarshallableFunctionExternalizers.SetValueIfEqualsReturnBooleanExternalizer());
       addInternalExternalizer(new PersistentUUID.Externalizer());
+
+      addInternalExternalizer(Params.Externalizer.INSTANCE);
```
danberindei: If you always call the externalizer's methods directly, this registration will never be used. So you could remove the `Params.Externalizer` class altogether and make the methods static in `Params`.

core/src/main/java/org/infinispan/functional/impl/Params.java
rvansa: Oops, haven't realized this. I'll issue fixup PR.

core/src/main/java/org/infinispan/marshall/core/ExternalizerTable.java
rvansa: I've expected that under some circumstances it could get serialized, but ATM it would make more sense to omit the registration.

PR #4266, opened 2016-04-25
ISPN-6545 Avoid leaking HR server req details

galderz:
https://issues.jboss.org/browse/ISPN-6545

### DIFF
--- a/server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodDecoder.java
+++ b/server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodDecoder.java
@@ -28,8 +28,9 @@ public class HotRodDecoder extends ByteToMessageDecoder {
    private final EmbeddedCacheManager cacheManager;
    private final NettyTransport transport;
    private final Predicate<? super String> ignoreCache;
+   private final HotRodServer server;
 
-   final CacheDecodeContext decodeCtx;
+   CacheDecodeContext decodeCtx;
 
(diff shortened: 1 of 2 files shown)


server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodDecoder.java
```diff
@@ -49,7 +51,7 @@ public NettyTransport getTransport() {
    }
 
    void resetNow() {
-      decodeCtx.resetParams();
```
wburns: We may as well remove this method while we are at it.

wburns:
Changes are fine with me, but these are also the ones I suggested.  If anyone wants to comment please do otherwise I will integrated tomorrow.

gustavonalle:
Changes looks good, but is the node.js client the only way to reproduce it?

wburns:
@gustavonalle unfortunately yes.  It only affected an async clien, where it was sending multiple requests on the same socket.

galderz:
@gustavonalle We'd need an async Java client to replicate indeed.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodDecoder.java
galderz: Sure

galderz:
@wburns Updated

wburns:
Pulling..

wburns:

PR #4267, opened 2016-04-26
ISPN-6505 Remote continuous query on a cache with compat mode enabled leads to CCE

anistor:
https://issues.jboss.org/browse/ISPN-6505


tristantarrant:
Pushed to master. Thanks.

PR #4272, opened 2016-04-27
ISPN-6300 Add missing bindings to dist remote exec

galderz:
* Added cacheManager and context parameters as bindings for distributed
  remote task execution.
* Marshaller not currently passed since the marshaller in the context
  can't easily be initialised in the distributed tasks.
* Added tests to verify that bindings are correctly set for normal
  remote execution and typed remote execution.

Q. Plugging in Marshaller for dist script execution is tricky, any ideas?


tristantarrant:
Pushed, thanks

PR #4273, opened 2016-04-27
ISPN-6547 Not all classes should be added to bean manager

slaskawi:
https://issues.jboss.org/browse/ISPN-6547


danberindei:
Sounds good, integrating.

danberindei:
Pushed, thanks Sebastian!

PR #4277, opened 2016-04-27
ISPN-6556 SegmentListenerNotifier.completeSegmentsNoResults is not

wburns:
called for batches without values

* Made sure empty set is returned so we notify empty segments


wburns:
This would need to be backported to 8.2.x and possibly 8.1.x.

tristantarrant:
Pushed to master, 8.2.x and 8.1.x. Thanks

PR #4278, opened 2016-04-27
Parse server list with spaces properly

vjuranek:
https://issues.jboss.org/browse/ISPN-6560


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.java
```diff
@@ -94,7 +94,7 @@ public ClusterConfigurationBuilder addCluster(String clusterName) {
    @Override
    public ConfigurationBuilder addServers(String servers) {
       for (String server : servers.split(";")) {
```
tristantarrant: You could also have used "\s*;\s*" and skip the trim().

tristantarrant:
Pushed to master, thanks

PR #4279, opened 2016-04-28
Ispn 5721 sni support

slaskawi:
https://issues.jboss.org/browse/ISPN-5721

Tasks: 
* https://issues.jboss.org/browse/ISPN-6517
* https://issues.jboss.org/browse/ISPN-6520
* https://issues.jboss.org/browse/ISPN-6523
* https://issues.jboss.org/browse/ISPN-6524
* https://issues.jboss.org/browse/ISPN-6525

Highlights
* Netty upgrade to `4.1.0.CR7`
* Added `SniHandler` to Hotrod server with proper configuration bits:
```
builder.ssl.enable()
   .sniHostName("www.example.com")
      .keyStoreFileName("some_file")
      .keyStorePassword("password")
   .sniHostName("www.example2.com")
      ...
```
* Added Hotrod server XML configuration:
```
<hotrod-connector name="hotrod5" socket-binding="hotrod" cache-container="default">
    <encryption security-realm="other">
         <sni host-name="example.com" security-realm="other" />
         <sni host-name="example2.com" security-realm="other2" />
     </encryption>
</hotrod-connector>
```
I was considering moving it under `<authentication>` but it turns out that SSL is turned on only with `<encryption>`. So I would have to do a some dirty tricks with the configuration.
* The client decides to which SNI domain it wants to connect using [SSLParameters#setServerNames] (https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLParameters.html#setServerNames-java.util.List-) method. 
* Added SNI to the client
```
 builder.security().ssl()
   .sslContext(cont)
   .sniHostName("sni")
   .enable();
```


slaskawi:
The failure seems unrelated and I can't reproduce it locally.

slaskawi:
Let's confirm if that fits into Router design doc: https://github.com/infinispan/infinispan/wiki/Multi-tenancy-for-Hotrod-Server

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/EndpointSubsystemReader_9_0.java
```diff
@@ -559,9 +559,42 @@ private void parseEncryption(XMLExtendedStreamReader reader, ModelNode connector
...
+            }
+         }
       }
+
+      if(!skipTagCheckAtTheEnd)
+         ParseUtils.requireNoContent(reader);
+   }
+
+   private void parseSni(final XMLExtendedStreamReader reader, final ModelNode encryption, final List<ModelNode> operations) throws XMLStreamException {
+      ParseUtils.requireAttributes(reader, Attribute.HOST_NAME.getLocalName(), Attribute.SECURITY_REALM.getLocalName());
```
tristantarrant: Why require to specify a security realm ? can't the realm specified on the `<encryption>` element be used by default ?

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/ProtocolServerService.java
```diff
@@ -152,6 +165,10 @@ public synchronized void start(final StartContext context) throws StartException
       }
    }
 
+   private boolean isSniEnabled() {
+      return sniDomains.keySet().size() > 0;
```
tristantarrant: return !sniDomains.isEmpty();

server/core/src/main/scala/org/infinispan/server/core/configuration/SslConfigurationBuilder.java
```diff
@@ -1,32 +1,39 @@
...
-   private SSLContext sslContext;
-   private String trustStoreFileName;
-   private char[] trustStorePassword;
-
-   SslConfigurationBuilder() {}
+   private SslEngineConfigurationBuilder defaultDomainConfigurationBuilder = new SslEngineConfigurationBuilder(this);
+   private Map<String, SslEngineConfigurationBuilder> sniDomains;
+
+   SslConfigurationBuilder(ProtocolServerConfigurationChildBuilder<T, S> parentConfigurationBuilder) {
+      this.parentConfigurationBuilder = parentConfigurationBuilder;
```
tristantarrant: If I understood this correctly, additional domains create a new SslConfigurationBuilder (see sniHostName(String domain) below) but each one defines a HashMap and a default domain.
Instead I believe only the first SslConfigurationBuilder should create the map which needs to be shared among all instances.

tristantarrant:
@slaskawi a few changes needed

server/core/src/main/scala/org/infinispan/server/core/configuration/SslConfigurationBuilder.java
slaskawi: Actually this should work correctly since I created additional builder for SNI called `SslEngineConfigurationBuilder` (because the SNI Host Name belongs to `SSLEngine` when looking at the implementatin). So it actually works as you described, we have only one map (in `SslConfigurationBuilder`) and multiple `SslEngineConfigurationBuilder`s.

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/ProtocolServerService.java
slaskawi: Done.

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/EndpointSubsystemReader_9_0.java
slaskawi: Ok, that sounds like a reasonable idea. Let me implement it.

slaskawi: Implemented in the `ProtocolServerService`.

tristantarrant:
Pushed to master (with a minor logging fix). Thanks

PR #4287, opened 2016-04-29
ISPN-6518 GlobalTransaction objects are not cleared properly

pruivo:
https://issues.jboss.org/browse/ISPN-6518


tristantarrant:
Pushed to master and 8.2.x. Create a PR for 8.1.x together with ISPN-6047

PR #4288, opened 2016-04-29
SkipOperation should use 'skip' rather than 'limit' on the stream

dahankzter:
Perhaps I misunderstood the idea here but probably SkipOperation should use skip rather than limit on the stream at hand.

### DIFF
--- a/core/src/test/java/org/infinispan/stream/BaseStreamTest.java
+++ b/core/src/test/java/org/infinispan/stream/BaseStreamTest.java
@@ -1499,9 +1499,7 @@ private static class ForEachIntInjected implements IntConsumer,
       for (int i = 0; i < range; i++) {
          List<Double> res = createStream(keySet)
                  .sorted().skip(i).collect(Collectors.toList());
-         for (int j = 0; j < range - res.size(); j++) {
-            assertFalse(res.contains(j));
-         }
+         assertEquals(0, IntStream.range(0, range - res.size()).mapToDouble(v -> v).filter(res::contains).count());
       }
    }
 
@@ -1515,9 +1513,7 @@ private static class ForEachIntInjected implements IntConsumer,
       for (int i = 0; i < range; i++) {
          List<Integer> res = createStream(keySet)
                  .sorted().skip(i).collect(Collectors.toList());
-         for (int j = 0; j < range - res.size(); j++) {
-            assertFalse(res.contains(j));
-         }
+         assertEquals(0, IntStream.range(0, range - res.size()).filter(res::contains).count());
       }
    }
 
@@ -1532,9 +1528,7 @@ private static class ForEachIntInjected implements IntConsumer,
          List<Double> res = createStream(keySet)
                  .sorted().limit(i).collect(Collectors.toList());
          assertEquals(i, res.size());
-         for (int j = res.size(); j < range; j++) {
-            assertFalse(res.contains(j));
-         }
+         assertEquals(0, IntStream.range(res.size(), range).mapToDouble(v -> v).filter(res::contains).count());
       }
    }
 
@@ -1549,9 +1543,8 @@ private static class ForEachIntInjected implements IntConsumer,
          List<Integer> res = createStream(keySet)
                  .sorted().limit(i).collect(Collectors.toList());
          assertEquals(i, res.size());
-         for (int j = res.size(); j < range; j++) {
-            assertFalse(res.contains(j));
-         }
+         assertEquals(0, IntStream.range(res.size(), range).filter(res::contains).count());
+
       }
    }
 
--- a/core/src/test/java/org/infinispan/stream/BaseStreamTest.java
+++ b/core/src/test/java/org/infinispan/stream/BaseStreamTest.java
@@ -1493,16 +1493,16 @@ private static class ForEachIntInjected implements IntConsumer,
       Cache<Double, String> cache = getCache(0);
       int range = 10;
       // First populate the cache with a bunch of values
-      IntStream.range(0,range).mapToDouble(value -> value).forEach(i -> cache.put(i, i + "-value"));
+      IntStream.range(0, range).mapToDouble(value -> value).forEach(i -> cache.put(i, i + "-value"));
       CacheSet<Map.Entry<Double, String>> entrySet = cache.entrySet();
 
       int cnt = range;
-      for(int i = 0;i<cnt;i++){
-         List<Map.Entry<Double,String>> res = createStream(entrySet)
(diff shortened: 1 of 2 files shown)


tristantarrant:
Looks good, please create a Jira and include it in the commit message.

wburns:
Good catch. Copy paste failure at its best.

dahankzter:
Added the Jira issue number to the commit.

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
```diff
@@ -1489,6 +1489,48 @@ public void testDoubleAllMatch() {
...
+      }
+   }
+
+   public void testLimit() {
+      Cache<Double, String> cache = getCache(0);
+      int range = 10;
+      // First populate the cache with a bunch of values
+      IntStream.range(0,range).mapToDouble(value -> value).forEach(i -> cache.put(i, i + "-value"));
+      CacheSet<Map.Entry<Double, String>> entrySet = cache.entrySet();
+      List<Map.Entry<Double,String>> total = createStream(entrySet)
```
wburns: Unused?

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
```diff
@@ -1489,6 +1489,48 @@ public void testDoubleAllMatch() {
...
 
+   public void testSkip() {
+      Cache<Double, String> cache = getCache(0);
+      int range = 10;
+      // First populate the cache with a bunch of values
+      IntStream.range(0,range).mapToDouble(value -> value).forEach(i -> cache.put(i, i + "-value"));
+      CacheSet<Map.Entry<Double, String>> entrySet = cache.entrySet();
+
+      int cnt = range;
+      for(int i = 0;i<cnt;i++){
```
wburns: Formatting is off, please apply the formatter found at https://github.com/infinispan/infinispan/tree/master/ide-settings

dahankzter: Ah yes my IntelliJ settings are probably off.

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
```diff
@@ -1489,6 +1489,48 @@ public void testDoubleAllMatch() {
...
+         List<Map.Entry<Double,String>> res = createStream(entrySet)
+                 .sorted((o1, o2) -> o1.getKey().compareTo(o2.getKey()))
+                 .skip(i).collect(Collectors.toList());
+         for(int j = 0;j< cnt - res.size();j++){
+            String targetValue = (double)j + "-value";
```
wburns: Also unneeded cast to double here and below.

wburns:
@dahankzter looks great, thanks for the contribution.

Normally I would add a test for the primitive versions as well, but this should be fine for this PR.

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
dahankzter: I don't think it is unneeded? The format of the elements in the cache is "1.0-value" and with an int it would not be comparable.

wburns: Ah I see it is because the test passes without it.  Maybe instead of checking to make sure no values exist with those we should be doing the inverse instead.

dahankzter:
Perhaps I should just work with the keys? The key set should suffer from the same thing and the test would probably be much cleaner.

What do you think?

wburns:
If you invert it to assert that the values are present it would catch what I was seeing.  And actually it would make the test a lot more correct, cause unfortunately testing for it to not exist means it could exist for other reasons :)

dahankzter:
Are you sure? The "skipped" values should absolutely not be present and checking for the existence of the remaining can not account for that right? Or perhaps enough with that and a size check?

dahankzter:
I switched to using the keys. The tests are much clearer and it shouldn't really matter if the keys or values are used for checks right?

wburns:
With a size check you can guarantee the proper elements are returned since you know the starting entries and which ones would have been skipped or limited since you have sorted them.

wburns:
But you are doing that as well now with the new changes :)

dahankzter:
Am I? I just tried to clean it up a bit. :)

dahankzter:
Are you happy with the current state? 

Can this go out in an 8.x release?
We would prefer to remove our somewhat ugly hack that we have now as soon as possible.

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
```diff
@@ -1489,6 +1489,65 @@ public void testDoubleAllMatch() {
...
+   public void testDoubleSkip() {
+      Cache<Double, String> cache = getCache(0);
+      int range = 10;
+      // First populate the cache with a bunch of values
+      IntStream.range(0, range).mapToDouble(value -> value).forEach(i -> cache.put(i, i + "-value"));
+      CacheSet<Double> keySet = cache.keySet();
+
+      for (int i = 0; i < range; i++) {
+         List<Double> res = createStream(keySet)
+                 .sorted().skip(i).collect(Collectors.toList());
```
wburns: Need to just add a `assertEquals(i, range - res.size());` and these should be good.

wburns:
@dahankzter what version were you looking at?  We are only maintaining 8.2.x and if critical 8.1.x.

dahankzter:
Added similar size checks for skips.

dahankzter:
We are using the latest 8.2

wburns:
I am updating the tests a bit more.  I actually found I  had a similar copy paste error for LongStream limit operation.  So I will fix that in this commit as well.  We will get this into 8.2.x though.  I dont' know when the next build is though.

dahankzter:
Great! I look forward to the release.

wburns:
@dahankzter BTW, if you don't mind, how you are guys using Distributed Streams?  And if you had any operators you might like to see would be great for some feedback there.  With the recent refactoring in Infinispan 9 we are able to add new operations like I did with [1] and I am always looking for new ones people might want.

[1] https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/CacheStream.java#L230

wburns:
This one is replaced by #4294

dahankzter:
We are mostly using it for whatever we would use streams in any other collections. Perhaps doing a search and then further customize sort and apply a filter on the returned entities. Convenience things more than any specifically distributed aspect.

PR #4291, opened 2016-04-29
ISPN-6574 Implement JCache remote getCache()

galderz:
https://issues.jboss.org/browse/ISPN-6574

* Provide a proper implementation for JCache remote getCache() method so
  that it verifies whether any remote caches exist and then wrap those
  inside a JCache instance and register it into the caches collection.


tristantarrant:
Pushed to master and 8.2.x. Thanks

PR #4301, opened 2016-05-04
ISPN-6491 Remove total order thread pool

pruivo:
https://issues.jboss.org/browse/ISPN-6491


slaskawi:
Shouldn't we mention in the docs that we are deprecating this? That's the only bit I'm missing in this PR, otherwise looks ok to me.

pruivo:
@slaskawi where should I mention it? 
the thread pool was deprecated (and no longer used) since 7.1. I just left the configuration and parser, that is removed in this PR.

wburns:
@pruivo I normally mention changes like this at https://github.com/infinispan/infinispan/blob/master/documentation/src/main/asciidoc/upgrading/upgrading.adoc

slaskawi:
@wburns Yes, that's exactly what I mean

pruivo:
done!

danberindei:
Integrated, thanks Pedro!

PR #4303, opened 2016-05-04
ISPN-6576: backport to 8.2.x

ksobolew:
The tests are a little different than the ones attached to the issue - there are fewer classes and the tests in transactional mode are disabled (because of ISPN-6573). The fix is almost too simple to make me confident it's the right one :)

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/api/functional/Param.java
+++ b/commons/src/main/java/org/infinispan/commons/api/functional/Param.java
@@ -111,6 +111,7 @@ public interface Param<P> {
       PERSIST, SKIP;
 
       public static final int ID = ParamIds.PERSISTENCE_MODE_ID;
+      private static final PersistenceMode[] CACHED_VALUES = values();
 
       @Override
       public int id() {
@@ -127,6 +128,10 @@ public interface Param<P> {
        */
       public static PersistenceMode defaultValue() {
          return PERSIST;
+      }
+
+      public static PersistenceMode valueOf(int ordinal) {
+         return CACHED_VALUES[ordinal];
       }
    }
 
--- a/core/src/main/java/org/infinispan/functional/impl/Params.java
+++ b/core/src/main/java/org/infinispan/functional/impl/Params.java
@@ -3,17 +3,14 @@ package org.infinispan.functional.impl;
 import org.infinispan.commons.api.functional.Param;
 import org.infinispan.commons.api.functional.Param.FutureMode;
 import org.infinispan.commons.api.functional.Param.PersistenceMode;
-import org.infinispan.commons.marshall.AdvancedExternalizer;
+import org.infinispan.commons.marshall.MarshallUtil;
 import org.infinispan.commons.util.Experimental;
-import org.infinispan.marshall.core.Ids;
 
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
-import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Supplier;
@@ -119,35 +116,19 @@ public final class Params {
       }
    }
 
-   public static class Externalizer implements AdvancedExternalizer<Params> {
-      public static Externalizer INSTANCE = new Externalizer();
+   public static void writeObject(ObjectOutput output, Params params) throws IOException {
+      // There's no point in sending FutureMode over wire
+      MarshallUtil.marshallEnum((PersistenceMode) params.get(PersistenceMode.ID).get(), output);
+   }
 
-      @Override
-      public Set<Class<? extends Params>> getTypeClasses() {
-         return Collections.singleton(Params.class);
-      }
-
-      @Override
-      public Integer getId() {
-         return Ids.PARAMS;
-      }
-
-      @Override
-      public void writeObject(ObjectOutput output, Params params) throws IOException {
-         // There's no point in sending FutureMode over wire
-         output.writeInt(((PersistenceMode) params.get(PersistenceMode.ID).get()).ordinal());
-      }
-
-      @Override
-      public Params readObject(ObjectInput input) throws IOException, ClassNotFoundException {
-         PersistenceMode persistenceMode = PersistenceMode.values()[input.readInt()];
-         if (persistenceMode == PersistenceMode.defaultValue()) {
-            return DEFAULT_INSTANCE;
-         } else {
-            Param[] params = Arrays.copyOf(DEFAULTS, DEFAULTS.length);
-            params[PersistenceMode.ID] = persistenceMode;
-            return new Params(params);
-         }
+   public static Params readObject(ObjectInput input) throws IOException, ClassNotFoundException {
+      PersistenceMode persistenceMode = MarshallUtil.unmarshallEnum(input, PersistenceMode::valueOf);
+      if (persistenceMode == PersistenceMode.defaultValue()) {
+         return DEFAULT_INSTANCE;
+      } else {
+         Param[] params = Arrays.copyOf(DEFAULTS, DEFAULTS.length);
+         params[PersistenceMode.ID] = persistenceMode;
+         return new Params(params);
       }
    }
 }
--- a/core/src/main/java/org/infinispan/functional/impl/Params.java
+++ b/core/src/main/java/org/infinispan/functional/impl/Params.java
@@ -3,10 +3,17 @@ package org.infinispan.functional.impl;
 import org.infinispan.commons.api.functional.Param;
 import org.infinispan.commons.api.functional.Param.FutureMode;
 import org.infinispan.commons.api.functional.Param.PersistenceMode;
+import org.infinispan.commons.marshall.AdvancedExternalizer;
 import org.infinispan.commons.util.Experimental;
+import org.infinispan.marshall.core.Ids;
 
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Supplier;
@@ -31,6 +38,10 @@ public final class Params {
       FutureMode.defaultValue(),
       PersistenceMode.defaultValue(),
    };
+   // TODO: as Params are immutable and there's only limited number of them,
+   // there could be a table with all the possible combinations and we
+   // wouldn't have to allocate at all
+   private static final Params DEFAULT_INSTANCE = new Params(DEFAULTS);
 
    final Param<?>[] params;
 
@@ -77,7 +88,7 @@ public final class Params {
    }
 
    public static Params create() {
-      return new Params(DEFAULTS);
+      return DEFAULT_INSTANCE;
    }
 
    public static Params from(Param<?>... ps) {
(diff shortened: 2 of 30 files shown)


rvansa:
Do you think it would be possible to just add the test as cases to the existing tests, rather than adding a completely new ones? And you can drop the tx-support in the tests, that can be added later when it's implemented.

rvansa:
(and btw., when updating PR, you don't have to close the old one and create new, just do a force-push into your branch).

ksobolew:
Yeah, I closed the PR accidentally when I was trying to put these commits on a separate branch :)

As for merging these tests with some existing ones - well, I don't know. They need some specific configuration, but I'll see what I can do. (Oh, and I can't seem to get TestNG to work right, *sigh*)

ksobolew:
I don't think I can find another home for these tests. I removed the transactional mode tests and simplified the rest a little (at the cost of parameterization of AbstractFunctionalTest). And I fixed them so that they actually work :)

ksobolew:
Ah, one more thing: the tests will pass only when ISPN-6542 is fixed; until then they'll throw a NPE. I forgot that I had the proposed fix for that committed on my local branch.

ksobolew:
Another bunch of tests and fixes. In addition to the simple ReadWriteMap.eval() the tests now also exercise all other types of functional Maps and operations on them. Note that it still requires the fix to ISPN-6542; see also https://github.com/infinispan/infinispan/pull/4265#issuecomment-217165678 which I originally did for this branch but it depends on rvansa's commit there.

ksobolew:
Argh, I see that the ManyDataCommand interface was also added by rvansa for ISPN-6542 and it's missing here, so it does not compile. Oh well...

ksobolew:
Rebased on top of back-ported ISPN-6542.

galderz:
Hey @ksobolew, thanks a lot for sending the PR. I'll have a look at the PR tomorrow and get back to you. Thanks for getting involved with the project 👏 

@rvansa Thanks for the help with this 😃

ksobolew:
@galderz You're walcome :) I'm sort of familiar with the codebase from the JBoss Cache days (we're finally migrating to Infinispan!) so I could just dive in and try to fix the problem myself

commons/src/main/java/org/infinispan/commons/api/functional/Param.java
```diff
@@ -44,6 +44,29 @@
...
 
    /**
+    * Should this paramater go over the wire?
+    */
+   SerializeMethod serializeMethod();
+
+   /**
+    * The way how this object should be serialized;
+    */
+   enum SerializeMethod {
```
galderz: This is not how we deal with an objects serialization/marshalling characteristics. Instead, if a param instance needs to be serialized, an Externalizer should be provided for it (examples in the codebase) and then the Externalizer it's either caught by annotation processing (e.g. @SerializedWith) or it's preregistered. The externalizer itself can decide how it serialize it using whichever method it wants. A param implementation, if you want something lazy, can implement Serializable and that's it. So, there's no need to add anything to the Param interface, which has been designed with the bare minimum required to do its job.

galderz: Also, whether the param itself should be sent over the wire or not is not down the parameter itself, but more to the command being executed. The command itself must decide whether it includes certain params inside its externalizer payload. Params themselves do not live on their own.

ksobolew: Note that this is part of back-ported #4265 by @rvansa

core/src/main/java/org/infinispan/commands/functional/ManyDataCommand.java
```diff
@@ -0,0 +1,10 @@
+package org.infinispan.commands.functional;
+
+import java.util.Set;
+
+/**
+ * @author Radim Vansa &lt;rvansa@redhat.com&gt;
+ */
+public interface ManyDataCommand<K> {
```
galderz: I don't really see the need to add yet another interface to the huge chain of *Command of interfaces we already have. What I mean by that is this: Each AbstractWriteManyCommand implementation could have a getKeys() method inside, that does not depend on an interface, and when it arrives to the interceptor level, each instance of the command, for which we have a visit* call, could call its own getKeys() and pass that to the method.

In fact, I think adding this interface makes the code more complex unnecessarily by having to use `&` structures to find commands that are X and Y...etc.

Truth to be said, I did the same error with ParamsCommand, and I can see now how this could get out for unnecessary reasons really.

core/src/main/java/org/infinispan/functional/impl/Params.java
```diff
@@ -108,4 +115,50 @@ public static Params from(Param<?>... ps) {
...
+
+      @Override
+      public Integer getId() {
+         return Ids.PARAMS;
+      }
+
+      @Override
+      public void writeObject(ObjectOutput output, Params object) throws IOException {
+         for (Param param : object.params) {
+            switch (param.serializeMethod()) {
```
galderz: I see what you are doing here but this is brittle IMO. Simply have each Param have its own Externalizer and that should work without any further complications nor the need to add an extra method to the interface. Besides, I don't expect that many params to make this is really that useful. IOW, I expect the number of Param impls to be in the tens and not in the hundreds :)

core/src/main/java/org/infinispan/interceptors/ClusteredCacheLoaderInterceptor.java
```diff
@@ -41,24 +41,35 @@ private void startClusteredCacheLoaderInterceptor() {
    }
 
    @Override
-   protected boolean skipLoadForWriteCommand(WriteCommand cmd, Object key, InvocationContext ctx) {
+   protected boolean skipLoadForWriteCommand(WriteCommand cmd, Object key, InvocationContext ctx, boolean functionalReadWrite) {
```
galderz: I'm not sure I like this boolean parameter either. It makes the method quite complex adding further if statements and it's already quite difficult to read. I think i'd rather add some code duplication by having functional-specific methods which are more focused, smaller, and easier to read.

Same thing for other methods that have been extended to have `functionalReadWrite` boolean parameters

ksobolew: Right. I was not sure about that either. Earlier versions (still immortalized by earlier commits in this PR) used instanceof checks deeper in the call stack but I was even less sure about those :) Would these checks be more acceptable or should I go the functional-specific methods way?

core/src/main/java/org/infinispan/functional/impl/Params.java
rvansa: @galderz So you think that Params should already know about all possible Param implementations (it does, through defaults), and serialize them one-by-one? That's what I've tried to avoid, by delegating the information about the params to themselves.
I want to spare the bytes and lookups in marshaller when the params always have the same order (according to their ids) and some of them don't make sense to be serialized at all (such as the future mode which is purely local thing).

core/src/main/java/org/infinispan/commands/functional/ManyDataCommand.java
rvansa: That's a valid point. I'll update #4265 accordingly.

danberindei: @galderz I don't really agree this makes the code more complex. There's a lot of duplicate code, e.g. in `NonTxDistributionInterceptor`, that only exists because we don't have an interface with `getKeys()` on it.

commons/src/main/java/org/infinispan/commons/api/functional/Param.java
galderz: Cross posted the comments to that PR.

core/src/main/java/org/infinispan/commands/functional/ManyDataCommand.java
galderz: I agree there's duplicate code and to be honest that class could do with some refactoring to avoid that. But I don't know for sure if your point is about duplication between different functional command visit impls or between functional command visit impls and the others.

In any case, I'm happy for such ManyDataCommand interface to exist if it's really going to reduce some code in that class too.

core/src/main/java/org/infinispan/functional/impl/Params.java
galderz: @rvansa Param implementations will be provided by us, it's like flags really. Params control how the internals work, e.g. whether to check the cache store or not...etc, so really they are something we control. I don't expect users to implement Param instances. On the other hand, MetaParam instances can easily be provided by the user since they're stored along with the entry (like Metadata) so we have less control there.

I can see the utility of having a single externalizer for all Params, but without your extra method you'd require some kind of lookup. Besides, I don't think your method would work for Param instances that carry values.

One externalizer per Param would just work fine IMO.

core/src/main/java/org/infinispan/interceptors/ClusteredCacheLoaderInterceptor.java
galderz: I think I'd go for functional-specific methods.

core/src/main/java/org/infinispan/functional/impl/Params.java
rvansa: MetaParams are different thing; I haven't touched those. I understand why these need externalizer per MetaParam.

But as for Params, we control those, so we can handle them in single Params.Externalizer. I'll push update to the other PR, so please let's move the discussion there.

galderz:
@ksobolew Radim's changes are in, so make sure you rebase with latest changes he made :)

ksobolew:
@galderz I'm on it!

ksobolew:
Updated and rebased.

core/src/main/java/org/infinispan/interceptors/CacheLoaderInterceptor.java
```diff
@@ -415,7 +444,12 @@ private boolean skipLoad(FlagAffectedCommand cmd, Object key, InvocationContext
       }
 
       boolean skip;
-      if (cmd instanceof WriteCommand) {
+      if (cmd instanceof AbstractWriteKeyCommand || cmd instanceof AbstractWriteManyCommand) {
```
ksobolew: Since the line 452 (formerly 418) uses instanceof, I thought I might as well add some more tests like this. I'm not etirely confident that testing for Abstract* classes is the way to go - maybe I should add a marker interface FunctionalCommand?

rvansa:
Sorry @ksobolew, @danberindei pointed out some shortcomings so I've added one more commit as #4359

ksobolew:
Any progress on this?

galderz:
@ksobolew Sorry for the delay, I'll have a look into this later on today.

galderz:
Why is this PR against 8.2.x? We're only fixing critical stuff in 8.x. This fix should be sent first for master and then we can backport if necessary. I'll do a quick review in the mean time but it's confusing since ISPN-6542 is already fixed in master and don't want to go over the same code again. Plus, the fact other fixes are quite spread (would it make sense to squash them?) does not make review easy :|.

core/src/test/java/org/infinispan/functional/FunctionalDistributionTest.java
```diff
@@ -0,0 +1,108 @@
...
+               try {
+                  Thread.sleep(10);
+               } catch (InterruptedException e) {
+                  Thread.currentThread().interrupt();
+               }
+               return entry.set(entry.find().orElse(0) + 1);
+            }).join();
+
+      // since we're in async mode, we need to wait before the replication finishes
+      Thread.sleep(100); // TODO: find a better way?
```
galderz: Minor thing, TestingUtil.sleepThread() sleeps and does not throw InterruptedException.

galderz:
Looks good otherwise, thanks a lot @ksobolew 😃. Please send a fix for master and try, if possible, to squash the commits if it makes sense to do so.

ksobolew:
@galderz Yes, sorry about the mixup, I should've sent a PR with the 6542 backport first (it's #4357 , BTW), then base this PR on that. But you're right, I also should've familiarized myself more with the procedures :) I'm working with 8.2.x on my application so that's where I fix the problems I encounter. I'll rebase it on top of master and push again soon, thanks!

ksobolew:
Rebased as a backport of #4378 to 8.2.x

@galderz I hope that this falls under "backport if necessary" :) This is pretty critical for me because the Functional API does not work at all and I'm relying on it.

galderz:
Integrated, thanks a million @ksobolew 😃

PR #4321, opened 2016-05-09
ISPN-5883 Node can apply new topology after sending status response

danberindei:
https://issues.jboss.org/browse/ISPN-5883

Change LocalTopologyManagerImpl to synchronize on `runningCaches` more. 

Note that this only changes the way topologies are updated in `LocalTopologyManagerImpl` - caches can still process topology updates in parallel, without holding a global lock.


tristantarrant:
Pushed to master, thanks

PR #4322, opened 2016-05-09
ISPN-6581 Typed remote exec in binary cache

galderz:
https://issues.jboss.org/browse/ISPN-6581

* Remote execution for typed scripts should be done in binary cache so
  that remote events can be correctly propagated. So, data conversion
  for non-compatibility mode caches is provided by a new class called
  DataTypedCache and it comes with a corresponding
  DataTypedCacheManager.
* When compatibility mode is enabled, the inner layer already provides
  data conversion layer. The problem with compatiblity mode is that it
  currently needs server-side configuration as opposed to a simple
  script header definition. Once encoding is better handled, we should
  be able to find a better solution for both.
* When data is inserted directly by the script, no version information
  is currently provided since that's managed by Hot Rod, so code had to
  be adjusted to deal with this situation. This might change in the
  future.
* Simplified EventLogListener code to check version information using
  purely the remote interface, just like other language target clients
  will use. This uncovered issues with dealing with version-less
  entries.
* Ported EventLogListener code to using Java 8 to make the code a lot
  clearer and cleaner to read.

### DIFF
--- a/scripting/src/main/java/org/infinispan/scripting/impl/ScriptingManagerImpl.java
+++ b/scripting/src/main/java/org/infinispan/scripting/impl/ScriptingManagerImpl.java
@@ -165,10 +165,13 @@ public class ScriptingManagerImpl implements ScriptingManager {
          .orElseGet(() -> new SimpleBindings());
 
       SimpleBindings systemBindings = new SimpleBindings();
-      systemBindings.put(SystemBindings.CACHE_MANAGER.toString(), cacheManager);
+      DataTypedCacheManager cm = new DataTypedCacheManager(metadata.dataType(), context.getMarshaller(), cacheManager);
+      systemBindings.put(SystemBindings.CACHE_MANAGER.toString(), cm);
       systemBindings.put(SystemBindings.SCRIPTING_MANAGER.toString(), this);
       context.getCache().ifPresent(cache -> {
-         systemBindings.put(SystemBindings.CACHE.toString(), cache);
+         Cache<?, ?> c = cm.getCacheConfiguration(cache.getName()).compatibility().enabled()
+               ? cache : new DataTypedCache<>(cm, cache);
+         systemBindings.put(SystemBindings.CACHE.toString(), c);
       });
 
       context.getMarshaller().ifPresent(marshaller -> {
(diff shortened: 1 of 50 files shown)


galderz:
It seems like there's some errors in testsuite, looking:
http://ci.infinispan.org/viewLog.html?buildId=39476&tab=buildResultsDiv&buildTypeId=bt9

galderz:
Updated with fix for failing tests.

scripting/src/main/java/org/infinispan/scripting/impl/ScriptingManagerImpl.java
```diff
@@ -165,10 +165,13 @@ public boolean containsScript(String taskName) {
          .orElseGet(() -> new SimpleBindings());
 
       SimpleBindings systemBindings = new SimpleBindings();
-      systemBindings.put(SystemBindings.CACHE_MANAGER.toString(), cacheManager);
+      DataTypedCacheManager cm = new DataTypedCacheManager(metadata.dataType(), context.getMarshaller(), cacheManager);
+      systemBindings.put(SystemBindings.CACHE_MANAGER.toString(), cm);
       systemBindings.put(SystemBindings.SCRIPTING_MANAGER.toString(), this);
       context.getCache().ifPresent(cache -> {
-         systemBindings.put(SystemBindings.CACHE.toString(), cache);
+         Cache<?, ?> c = cm.getCacheConfiguration(cache.getName()).compatibility().enabled()
```
galderz: This calls `cm.getCacheConfiguration(cache.getName()).compatibility().enabled()` instead of `cache.getCacheConfiguration().compatibility().enabled()` because the latter requires ADMIN rights in secured caches whereas the former works.

The bigger problem is that even if the cache is secured, we should be able to call what we need to call without problems. Maybe this can be implemented in a better way?

galderz:
Updated the comment with more details about the fix.

scripting/src/main/java/org/infinispan/scripting/impl/ScriptingManagerImpl.java
tristantarrant: Aside from the fact that this should be fixed by protecting the other entry point as well, cm.getCacheConfiguration() doesn't probably do what you want since you can start a cache using a configuration with a different name

tristantarrant:
Pushed to master.

PR #4324, opened 2016-05-09
ISPN-6610 Stop using forkInvocationSync

danberindei:
https://issues.jboss.org/browse/ISPN-6610

Interceptors must use onReturn() for things they need
to do after command.perform(). While forkInvocationSync()
still works, it doesn't allow asynchronous execution.

* Fix forkInvocation-related problems in BaseSequentialInvocationContext.
* Make BaseSequentialInvocationContext.invokeInterceptors invoke
  visitCommand in sequence for all interceptors, to allow inlining.
* Move visit notification from CallInterceptor to EntryWrappingInterceptor

With the invokeInterceptors unrolled loop and all the interceptors inlined, the performance in local caches is similar to the performance with forkInvocationSync. Performance in clustered scenarios should be better once we enable asynchronous execution.

### DIFF
--- a/core/src/main/java/org/infinispan/context/AsyncInvocationContext.java
+++ b/core/src/main/java/org/infinispan/context/AsyncInvocationContext.java
@@ -2,7 +2,7 @@ package org.infinispan.context;
 
 import org.infinispan.commands.VisitableCommand;
 import org.infinispan.commons.util.Experimental;
-import org.infinispan.interceptors.SequentialInterceptor;
+import org.infinispan.interceptors.AsyncInterceptor;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
@@ -16,7 +16,7 @@ import java.util.function.Function;
  * @author Dan Berindei
  * @since 9.0
  */
-public interface SequentialInvocationContext {
+public interface AsyncInvocationContext {
    /**
     * Execute a callback after the rest of the interceptors in the chain have finished executing.
     *
@@ -29,12 +29,12 @@ public interface SequentialInvocationContext {
     * Returning {@code null} will preserve the previous result or exception (equivalent to a
     * {@code finally} block in an old-style interceptor).
     */
-   CompletableFuture<Void> onReturn(SequentialInterceptor.ReturnHandler returnHandler);
+   CompletableFuture<Void> onReturn(AsyncInterceptor.ReturnHandler returnHandler);
 
    /**
     * Equivalent to {@code null}, when returned by
-    * {@link SequentialInterceptor#visitCommand(InvocationContext, VisitableCommand)} or
-    * {@link SequentialInterceptor.ReturnHandler#handle(InvocationContext, VisitableCommand, Object, Throwable)}.
+    * {@link AsyncInterceptor#visitCommand(InvocationContext, VisitableCommand)} or
+    * {@link AsyncInterceptor.ReturnHandler#handle(InvocationContext, VisitableCommand, Object, Throwable)}.
     *
     * <p>Unlike {@code null}, it can be used with {@link CompletableFuture#thenCompose(Function)}.
     * And unlike {@code CompletableFuture.completedFuture(null)}, it will not allocate a new object.</p>
@@ -65,14 +65,14 @@ public interface SequentialInvocationContext {
     * @param newCommand The command to fork
     * @param forkReturnHandler The return callback
     */
-   CompletableFuture<Void> forkInvocation(VisitableCommand newCommand, SequentialInterceptor.ForkReturnHandler forkReturnHandler);
+   CompletableFuture<Void> forkInvocation(VisitableCommand newCommand, AsyncInterceptor.ForkReturnHandler forkReturnHandler);
 
    /**
     * Fork the command invocation, and execute the remaining interceptors (and their return handlers)
     * synchronously.
     *
     * <p>Usually it is easier to use than
-    * {@link #forkInvocation(VisitableCommand, SequentialInterceptor.ForkReturnHandler)}.
+    * {@link #forkInvocation(VisitableCommand, AsyncInterceptor.ForkReturnHandler)}.
     * However, it is not recommended, because any asynchronous work in the remaining interceptors will block
     * the calling thread.</p>
     *
@@ -81,5 +81,5 @@ public interface SequentialInvocationContext {
     * @param newCommand The command to fork
     */
    @Experimental
-   Object forkInvocationSync(VisitableCommand newCommand) throws InterruptedException, Throwable;
+   Object forkInvocationSync(VisitableCommand newCommand) throws Throwable;
 }
--- a/core/src/main/java/org/infinispan/context/impl/AbstractTxInvocationContext.java
+++ b/core/src/main/java/org/infinispan/context/impl/AbstractTxInvocationContext.java
@@ -2,7 +2,6 @@ package org.infinispan.context.impl;
 
 import org.infinispan.commands.write.WriteCommand;
 import org.infinispan.container.entries.CacheEntry;
-import org.infinispan.interceptors.SequentialInterceptorChain;
 import org.infinispan.remoting.transport.Address;
 import org.infinispan.transaction.impl.AbstractCacheTransaction;
 import org.infinispan.transaction.xa.GlobalTransaction;
@@ -25,8 +24,7 @@ public abstract class AbstractTxInvocationContext<T extends AbstractCacheTransac
 
    private final T cacheTransaction;
 
-   protected AbstractTxInvocationContext(T cacheTransaction, Address origin,
-                                         SequentialInterceptorChain interceptorChain) {
+   protected AbstractTxInvocationContext(T cacheTransaction, Address origin) {
       super(origin);
       if (cacheTransaction == null) {
          throw new NullPointerException("CacheTransaction cannot be null");
--- a/core/src/main/java/org/infinispan/context/impl/LocalTxInvocationContext.java
+++ b/core/src/main/java/org/infinispan/context/impl/LocalTxInvocationContext.java
@@ -6,7 +6,6 @@ import org.infinispan.transaction.impl.LocalTransaction;
 import javax.transaction.Status;
 import javax.transaction.SystemException;
 import javax.transaction.Transaction;
-
 import java.util.Collection;
 
 /**
--- a/core/src/main/java/org/infinispan/context/impl/LocalTxInvocationContext.java
+++ b/core/src/main/java/org/infinispan/context/impl/LocalTxInvocationContext.java
@@ -1,6 +1,5 @@
 package org.infinispan.context.impl;
 
-import org.infinispan.interceptors.SequentialInterceptorChain;
 import org.infinispan.remoting.transport.Address;
 import org.infinispan.transaction.impl.LocalTransaction;
 
@@ -20,9 +19,8 @@ import java.util.Collection;
  */
 public class LocalTxInvocationContext extends AbstractTxInvocationContext<LocalTransaction> {
 
-   public LocalTxInvocationContext(LocalTransaction localTransaction,
-                                   SequentialInterceptorChain interceptorChain) {
-      super(localTransaction, null, interceptorChain);
+   public LocalTxInvocationContext(LocalTransaction localTransaction) {
+      super(localTransaction, null);
    }
 
    @Override
--- a/core/src/main/java/org/infinispan/context/impl/NonTxInvocationContext.java
+++ b/core/src/main/java/org/infinispan/context/impl/NonTxInvocationContext.java
@@ -4,7 +4,6 @@ import org.infinispan.commons.equivalence.AnyEquivalence;
 import org.infinispan.commons.equivalence.Equivalence;
 import org.infinispan.commons.util.CollectionFactory;
 import org.infinispan.container.entries.CacheEntry;
-import org.infinispan.interceptors.SequentialInterceptorChain;
 import org.infinispan.remoting.transport.Address;
 
 import java.util.Collections;
@@ -27,15 +26,13 @@ public class NonTxInvocationContext extends AbstractInvocationContext {
    private Object lockOwner;
 
 
-   public NonTxInvocationContext(int numEntries, Address origin, Equivalence<Object> keyEq,
-                                 SequentialInterceptorChain interceptorChain) {
+   public NonTxInvocationContext(int numEntries, Address origin, Equivalence<Object> keyEq) {
       super(origin);
       lookedUpEntries = CollectionFactory.makeMap(CollectionFactory.computeCapacity(numEntries), keyEq, AnyEquivalence.getInstance());
       this.keyEq = keyEq;
    }
 
-   public NonTxInvocationContext(Address origin, Equivalence<Object> keyEq,
-                                 SequentialInterceptorChain interceptorChain) {
+   public NonTxInvocationContext(Address origin, Equivalence<Object> keyEq) {
       super(origin);
       lookedUpEntries = CollectionFactory.makeMap(INITIAL_CAPACITY, keyEq, AnyEquivalence.getInstance());
       this.keyEq = keyEq;
--- a/core/src/main/java/org/infinispan/context/impl/RemoteTxInvocationContext.java
+++ b/core/src/main/java/org/infinispan/context/impl/RemoteTxInvocationContext.java
@@ -1,6 +1,5 @@
 package org.infinispan.context.impl;
 
-import org.infinispan.interceptors.SequentialInterceptorChain;
 import org.infinispan.transaction.impl.RemoteTransaction;
 
 import javax.transaction.Transaction;
@@ -15,9 +14,8 @@ import javax.transaction.Transaction;
  */
 public class RemoteTxInvocationContext extends AbstractTxInvocationContext<RemoteTransaction> {
 
-   public RemoteTxInvocationContext(RemoteTransaction cacheTransaction,
-                                    SequentialInterceptorChain interceptorChain) {
-      super(cacheTransaction, cacheTransaction.getGlobalTransaction().getAddress(), interceptorChain);
+   public RemoteTxInvocationContext(RemoteTransaction cacheTransaction) {
+      super(cacheTransaction, cacheTransaction.getGlobalTransaction().getAddress());
    }
 
    @Override
--- a/core/src/main/java/org/infinispan/interceptors/AsyncInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/AsyncInterceptor.java
@@ -2,8 +2,8 @@ package org.infinispan.interceptors;
 
 import org.infinispan.commands.VisitableCommand;
 import org.infinispan.commons.util.Experimental;
+import org.infinispan.context.AsyncInvocationContext;
 import org.infinispan.context.InvocationContext;
-import org.infinispan.context.SequentialInvocationContext;
 
 import java.util.concurrent.CompletableFuture;
 
@@ -14,26 +14,26 @@ import java.util.concurrent.CompletableFuture;
  * @since 9.0
  */
 @Experimental
-public interface SequentialInterceptor {
+public interface AsyncInterceptor {
    /**
     * Perform some work for the command, before the command itself is executed.
     *
     * <p>Must return an instance of {@code CompletableFuture<Void>}.</p>
     *
     * <p>Can return an already-completed {@code CompletableFuture<Void>}
-    * (e.g. {@link SequentialInvocationContext#continueInvocation()}) if the interceptor is synchronous,
+    * (e.g. {@link AsyncInvocationContext#continueInvocation()}) if the interceptor is synchronous,
     * i.e. it finishes executing when {@code visitCommand} returns.</p>
     *
     * <p>The interceptor can also influence the execution of the following interceptors in the chain:</p>
     * <ul>
-    * <li>The interceptor can call {@link SequentialInvocationContext#shortCircuit(Object)} in order to skip
+    * <li>The interceptor can call {@link AsyncInvocationContext#shortCircuit(Object)} in order to skip
     * the execution of the rest of the chain (and the command itself).</li>
     * <li>The interceptor can call
-    * {@link SequentialInvocationContext#forkInvocation(VisitableCommand, ForkReturnHandler)} in order to invoke
+    * {@link AsyncInvocationContext#forkInvocation(VisitableCommand, ForkReturnHandler)} in order to invoke
     * a new command, starting with the next interceptor in the chain. The return handler then behaves as
     * another {@code visitCommand} invocation: it can allow the invocation of the original command to continue
     * with the next interceptor, short-circuit the invocation, or fork another command.</li>
-    * <li>{@link SequentialInvocationContext#forkInvocationSync(VisitableCommand)} is a synchronous
+    * <li>{@link AsyncInvocationContext#forkInvocationSync(VisitableCommand)} is a synchronous
     * alternative to {@code forkInvocation}. It is easier to use, however it is not recommended,
     * because any asynchronous work in the remaining interceptors will block the calling thread.</li>
     * </ul>
@@ -65,7 +65,7 @@ public interface SequentialInterceptor {
 
    /**
     * A return handler installed with
-    * {@link SequentialInvocationContext#forkInvocation(VisitableCommand, ForkReturnHandler)}.
+    * {@link AsyncInvocationContext#forkInvocation(VisitableCommand, ForkReturnHandler)}.
     *
     * <p>It must behave just like {@link #visitCommand(InvocationContext, VisitableCommand)}.</p>
     */
--- a/core/src/main/java/org/infinispan/interceptors/AsyncInterceptorChain.java
+++ b/core/src/main/java/org/infinispan/interceptors/AsyncInterceptorChain.java
@@ -3,13 +3,12 @@ package org.infinispan.interceptors;
 import org.infinispan.commands.VisitableCommand;
 import org.infinispan.commons.util.Experimental;
 import org.infinispan.context.InvocationContext;
-import org.infinispan.context.SequentialInvocationContext;
 
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
 /**
- * Interceptor chain using {@link SequentialInterceptor}s.
+ * Interceptor chain using {@link AsyncInterceptor}s.
  *
  * Experimental: The ability to modify the interceptors at runtime may be removed in future versions.
  *
@@ -17,11 +16,11 @@ import java.util.concurrent.CompletableFuture;
  * @since 9.0
  */
 @Experimental
-public interface SequentialInterceptorChain {
+public interface AsyncInterceptorChain {
    /**
     * @return An immutable list of the current interceptors.
     */
-   List<SequentialInterceptor> getInterceptors();
+   List<AsyncInterceptor> getInterceptors();
 
    /**
     * Inserts the given interceptor at the specified position in the chain (0 based indexing).
@@ -29,7 +28,7 @@ public interface SequentialInterceptorChain {
     * @throws IllegalArgumentException if the position is invalid (e.g. 5 and there are only 2 interceptors
     *       in the chain)
     */
-   void addInterceptor(SequentialInterceptor interceptor, int position);
+   void addInterceptor(AsyncInterceptor interceptor, int position);
 
    /**
     * Removes the interceptor at the given position.
@@ -47,23 +46,22 @@ public interface SequentialInterceptorChain {
    /**
     * Removes all the occurrences of supplied interceptor type from the chain.
     */
-   void removeInterceptor(Class<? extends SequentialInterceptor> clazz);
+   void removeInterceptor(Class<? extends AsyncInterceptor> clazz);
 
    /**
     * Adds a new interceptor in list after an interceptor of a given type.
     *
     * @return true if the interceptor was added; i.e. the {@code afterInterceptor} exists
     */
-   boolean addInterceptorAfter(SequentialInterceptor toAdd, Class<? extends
-         SequentialInterceptor> afterInterceptor);
+   boolean addInterceptorAfter(AsyncInterceptor toAdd, Class<? extends
+         AsyncInterceptor> afterInterceptor);
 
    /**
     * Adds a new interceptor in list before an interceptor of a given type.
     *
     * @return true if the interceptor was added; i.e. the {@code beforeInterceptor} exists
     */
-   boolean addInterceptorBefore(SequentialInterceptor toAdd,
-                                                Class<? extends SequentialInterceptor> beforeInterceptor);
+   boolean addInterceptorBefore(AsyncInterceptor toAdd, Class<? extends AsyncInterceptor> beforeInterceptor);
 
    /**
     * Replaces an existing interceptor of the given type in the interceptor chain with a new interceptor
@@ -73,24 +71,21 @@ public interface SequentialInterceptorChain {
     * @param toBeReplacedInterceptorType the type of interceptor that should be swapped with the new one
     * @return true if the interceptor was replaced
     */
-   boolean replaceInterceptor(SequentialInterceptor replacingInterceptor,
-                                              Class<? extends SequentialInterceptor> toBeReplacedInterceptorType);
+   boolean replaceInterceptor(AsyncInterceptor replacingInterceptor,
+         Class<? extends AsyncInterceptor> toBeReplacedInterceptorType);
 
    /**
     * Appends at the end.
     */
-   void appendInterceptor(SequentialInterceptor ci, boolean isCustom);
+   void appendInterceptor(AsyncInterceptor ci, boolean isCustom);
 
    /**
     * Walks the command through the interceptor chain. The received ctx is being passed in.
     *
     * <p>Note: Reusing the context for multiple invocations is allowed. However, the two invocations
-    * must not overlap, so calling {@code invoke(InvocationContext, VisitableCommand)} from an interceptor
-    * is not allowed.
-    * If an interceptor wants to invoke a new command and cannot use
-    * {@link SequentialInvocationContext#forkInvocation(VisitableCommand, SequentialInterceptor.ForkReturnHandler)}
-    * or {@link SequentialInvocationContext#forkInvocationSync(VisitableCommand)},
-    * it must first copy the invocation context with {@link InvocationContext#clone()}.</p>
+    * must not overlap, so calling {@code invoke(ctx, command)} from an interceptor is not allowed.
+    * If an interceptor needs to invoke a new command through the entire chain, it must first
+    * copy the invocation context with {@link InvocationContext#clone()}.</p>
     */
    Object invoke(InvocationContext ctx, VisitableCommand command);
 
@@ -102,26 +97,26 @@ public interface SequentialInterceptorChain {
    /**
     * Returns the first interceptor extending the given class, or {@code null} if there is none.
     */
-   <T extends SequentialInterceptor> T findInterceptorExtending(Class<T> interceptorClass);
+   <T extends AsyncInterceptor> T findInterceptorExtending(Class<T> interceptorClass);
 
    /**
     * Returns the first interceptor with the given class, or {@code null} if there is none.
     */
-   <T extends SequentialInterceptor> T findInterceptorWithClass(Class<T> interceptorClass);
+   <T extends AsyncInterceptor> T findInterceptorWithClass(Class<T> interceptorClass);
 
    /**
     * Checks whether the chain contains the supplied interceptor instance.
     */
-   boolean containsInstance(SequentialInterceptor interceptor);
+   boolean containsInstance(AsyncInterceptor interceptor);
 
    /**
     * Checks whether the chain contains an interceptor with the given class.
     */
-   boolean containsInterceptorType(Class<? extends SequentialInterceptor> interceptorType);
+   boolean containsInterceptorType(Class<? extends AsyncInterceptor> interceptorType);
 
    /**
     * Checks whether the chain contains an interceptor with the given class, or a subclass.
     */
-   boolean containsInterceptorType(Class<? extends SequentialInterceptor> interceptorType,
+   boolean containsInterceptorType(Class<? extends AsyncInterceptor> interceptorType,
                                                    boolean alsoMatchSubClasses);
 }
--- a/core/src/main/java/org/infinispan/interceptors/BaseCustomAsyncInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/BaseCustomAsyncInterceptor.java
@@ -7,7 +7,7 @@ import org.infinispan.factories.annotations.Stop;
 import org.infinispan.manager.EmbeddedCacheManager;
 
 /**
- * Anyone using the {@link org.infinispan.interceptors.SequentialInterceptorChain#addInterceptor(SequentialInterceptor, int)} method (or any of its
+ * Anyone using the {@link AsyncInterceptorChain#addInterceptor(AsyncInterceptor, int)} method (or any of its
  * overloaded forms) or registering custom interceptors via XML should extend this base class when creating their own 
  * custom interceptors.
  * <p />
@@ -21,7 +21,7 @@ import org.infinispan.manager.EmbeddedCacheManager;
  * @author Dan Berindei
  * @since 9.0
  */
-public class BaseCustomSequentialInterceptor extends DDSequentialInterceptor {
+public class BaseCustomAsyncInterceptor extends DDAsyncInterceptor {
    protected Cache<?, ?> cache;
    protected EmbeddedCacheManager embeddedCacheManager;
 
--- a/core/src/main/java/org/infinispan/interceptors/compat/BaseTypeConverterInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/compat/BaseTypeConverterInterceptor.java
@@ -25,7 +25,7 @@ import org.infinispan.context.Flag;
 import org.infinispan.context.InvocationContext;
 import org.infinispan.distribution.DistributionManager;
 import org.infinispan.factories.annotations.Inject;
-import org.infinispan.interceptors.DDSequentialInterceptor;
+import org.infinispan.interceptors.DDAsyncInterceptor;
 import org.infinispan.metadata.Metadata;
 import org.infinispan.stream.impl.interceptor.AbstractDelegatingEntryCacheSet;
 import org.infinispan.stream.impl.interceptor.AbstractDelegatingKeyCacheSet;
@@ -50,7 +50,7 @@ import java.util.stream.StreamSupport;
  *
  * @author Galder Zamarreño
  */
-public abstract class BaseTypeConverterInterceptor<K, V> extends DDSequentialInterceptor {
+public abstract class BaseTypeConverterInterceptor<K, V> extends DDAsyncInterceptor {
 
    private InternalEntryFactory entryFactory;
    private VersionGenerator versionGenerator;
--- a/core/src/main/java/org/infinispan/interceptors/compat/BaseTypeConverterInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/compat/BaseTypeConverterInterceptor.java
@@ -81,8 +81,12 @@ public abstract class BaseTypeConverterInterceptor<K, V> extends DDSequentialInt
          command.setKey(converter.boxKey(key));
          command.setValue(converter.boxValue(command.getValue()));
       }
-      Object ret = ctx.forkInvocationSync(command);
-      return ctx.shortCircuit(converter.unboxValue(ret));
+      return ctx.onReturn((rCtx, rCommand, rv, throwable) -> {
+         if (throwable != null)
+            throw throwable;
+
+         return CompletableFuture.completedFuture(converter.unboxValue(rv));
+      });
    }
 
    @Override
@@ -110,14 +114,16 @@ public abstract class BaseTypeConverterInterceptor<K, V> extends DDSequentialInt
       if (ctx.isOriginLocal()) {
          command.setKey(converter.boxKey(key));
       }
-      Object ret = ctx.forkInvocationSync(command);
-      if (ret != null) {
-         if (needsUnboxing(ctx)) {
-            return ctx.shortCircuit(converter.unboxValue(ret));
+      return ctx.onReturn((rCtx, rCommand, rv, throwable) -> {
+         if (throwable != null)
+            throw throwable;
+
+         if (rv == null || !needsUnboxing(rCtx)) {
+            // Preserves the original return value
+            return null;
          }
-         return ctx.shortCircuit(ret);
-      }
-      return ctx.shortCircuit(null);
+         return CompletableFuture.completedFuture(converter.unboxValue(rv));
+      });
    }
 
    @Override
@@ -128,19 +134,21 @@ public abstract class BaseTypeConverterInterceptor<K, V> extends DDSequentialInt
       if (ctx.isOriginLocal()) {
          command.setKey(converter.boxKey(key));
       }
-      Object ret = ctx.forkInvocationSync(command);
-      if (ret != null) {
-         CacheEntry entry = (CacheEntry) ret;
-         Object returnValue = entry.getValue();
-         if (needsUnboxing(ctx)) {
-            returnValue = converter.unboxValue(entry.getValue());
+      return ctx.onReturn((rCtx, rCommand, rv, throwable) -> {
+         if (throwable != null)
+            throw throwable;
+
+         if (rv == null || !needsUnboxing(rCtx)) {
+            // Preserves the original return value
+            return null;
          }
+         CacheEntry entry = (CacheEntry) rv;
+         Object returnValue = converter.unboxValue(entry.getValue());
          // Create a copy of the entry to avoid modifying the internal entry
-         return ctx.shortCircuit(entryFactory.create(
-               entry.getKey(), returnValue, entry.getMetadata(),
-               entry.getLifespan(), entry.getMaxIdle()));
-      }
-      return ctx.shortCircuit(null);
+         return CompletableFuture.completedFuture(entryFactory
+               .create(entry.getKey(), returnValue, entry.getMetadata(), entry.getLifespan(),
+                     entry.getMaxIdle()));
+      });
    }
 
    private boolean needsUnboxing(InvocationContext ctx) {
@@ -159,14 +167,16 @@ public abstract class BaseTypeConverterInterceptor<K, V> extends DDSequentialInt
          }
          command.setKeys(boxedKeys);
       }
-      Object ret = ctx.forkInvocationSync(command);
+      return ctx.onReturn((rCtx, rCommand, rv, throwable) -> {
+         if (throwable != null)
+            throw throwable;
 
-      if (ret != null && !needsUnboxing(ctx))
-         return ctx.shortCircuit(ret);
+         if (rv == null || !needsUnboxing(rCtx)) {
+            return null;
+         }
 
-      if (ret != null) {
          if (command.isReturnEntries()) {
-            Map<Object, CacheEntry> map = (Map<Object, CacheEntry>) ret;
+            Map<Object, CacheEntry> map = (Map<Object, CacheEntry>) rv;
             Map<Object, Object> unboxed = command.createMap();
             for (Entry<Object, CacheEntry> entry : map.entrySet()) {
                CacheEntry cacheEntry = entry.getValue();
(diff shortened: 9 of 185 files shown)


gustavonalle:
There was a hung process in the slave-3, I manually killed it

core/src/main/java/org/infinispan/interceptors/compat/BaseTypeConverterInterceptor.java
```diff
@@ -128,19 +138,24 @@ protected void init(InternalEntryFactory entryFactory, VersionGenerator versionG
...
+      ctx.onReturn((ctx1, command1, rv, throwable) -> {
+         if (throwable != null)
+            throw throwable;
+
+         if (rv == null) {
+            return null;
+         }
+         CacheEntry entry = (CacheEntry) rv;
          Object returnValue = entry.getValue();
          if (needsUnboxing(ctx)) {
```
pruivo: it should use the ```ctx1```

core/src/main/java/org/infinispan/interceptors/compat/BaseTypeConverterInterceptor.java
```diff
@@ -110,14 +115,19 @@ protected void init(InternalEntryFactory entryFactory, VersionGenerator versionG
...
       }
-      Object ret = ctx.forkInvocationSync(command);
-      if (ret != null) {
-         if (needsUnboxing(ctx)) {
-            return ctx.shortCircuit(converter.unboxValue(ret));
+      ctx.onReturn((ctx1, command1, rv, throwable) -> {
+         if (throwable != null)
+            throw throwable;
+
+         if (rv == null) {
```
pruivo: IMO, you can merge both ```if```s
```
if (rv == null || !needsUnboxing(ctx1)) return null;
```

core/src/main/java/org/infinispan/interceptors/compat/BaseTypeConverterInterceptor.java
```diff
@@ -128,19 +138,24 @@ protected void init(InternalEntryFactory entryFactory, VersionGenerator versionG
...
          if (needsUnboxing(ctx)) {
             returnValue = converter.unboxValue(entry.getValue());
          }
          // Create a copy of the entry to avoid modifying the internal entry
-         return ctx.shortCircuit(entryFactory.create(
-               entry.getKey(), returnValue, entry.getMetadata(),
-               entry.getLifespan(), entry.getMaxIdle()));
-      }
-      return ctx.shortCircuit(null);
+         return CompletableFuture.completedFuture(entryFactory
```
pruivo: I'm wondering if you need to always create a new entry even if no unbox is needed...

core/src/main/java/org/infinispan/interceptors/compat/BaseTypeConverterInterceptor.java
```diff
@@ -159,42 +174,47 @@ private boolean needsUnboxing(InvocationContext ctx) {
...
          command.setKeys(boxedKeys);
       }
-      Object ret = ctx.forkInvocationSync(command);
+      ctx.onReturn((ctx1, command1, rv, throwable) -> {
+         if (throwable != null)
+            throw throwable;
 
-      if (ret != null && !needsUnboxing(ctx))
-         return ctx.shortCircuit(ret);
+         if (rv == null) {
```
pruivo: here the ```if```s could be merged and  should use ```ctx1``` to check if unbox is needed.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -101,122 +101,135 @@ public void configure() {
...
+         InvocationContext ctx, boolean acquireRemoteLock, FlagAffectedCommand command, boolean isWrite)
+         throws Exception {
+      GlobalTransaction gtx = ctx.isInTxScope() ? ((TxInvocationContext) ctx).getGlobalTransaction() : null;
       ClusteredGetCommand get = cf.buildClusteredGetCommand(key, command.getFlagsBitSet(), acquireRemoteLock, gtx);
       get.setWrite(isWrite);
 
-      RpcOptionsBuilder rpcOptionsBuilder = rpcManager.getRpcOptionsBuilder(ResponseMode.WAIT_FOR_VALID_RESPONSE, DeliverOrder.NONE);
+      RpcOptionsBuilder rpcOptionsBuilder =
+            rpcManager.getRpcOptionsBuilder(ResponseMode.WAIT_FOR_VALID_RESPONSE, DeliverOrder.NONE);
       int lastTopologyId = -1;
```
pruivo: can be removed.

core/src/main/java/org/infinispan/interceptors/distribution/DistributionBulkInterceptor.java
```diff
@@ -226,17 +233,17 @@ public V setValue(V value) {
...
+      if (ctx.isInTxScope()) {
+         keySet = new TxBackingKeySet<>(Caches.getCacheWithFlags(cache, command),
+               cache.getAdvancedCache().withFlags(Flag.CACHE_MODE_LOCAL).cacheEntrySet(), command,
+               (LocalTxInvocationContext) ctx);
       } else {
-         keySet = (CacheSet<K>) ctx.forkInvocationSync(command);
+         keySet = new BackingKeySet<>(Caches.getCacheWithFlags(cache, command), cache.getAdvancedCache().withFlags(
+                 Flag.CACHE_MODE_LOCAL).cacheEntrySet(), command);
       }
       return ctx.shortCircuit(keySet);
```
pruivo: shouldn't return a completed future?

core/src/main/java/org/infinispan/interceptors/distribution/L1LastChanceInterceptor.java
```diff
@@ -76,40 +76,50 @@ public void start() {
...
-      if (shouldUpdateOnWriteCommand(command) && command.isSuccessful() &&
-            cdl.localNodeIsOwner((key = command.getKey()))) {
-         if (trace) {
-            log.trace("Sending additional invalidation for requestors if necessary.");
+      ctx.onReturn((ctx1, command1, rv, throwable) -> {
+         if (throwable != null)
+            throw throwable;
+
+         Object key;
+         if (shouldUpdateOnWriteCommand(command) && command.isSuccessful() &&
```
pruivo: is there a case where the ```command != command1``` or ```ctx != ctx1```? if not, I would remove the arguments from the return handler. otherwise, you need to update the code use the ```command1``` :)

core/src/main/java/org/infinispan/interceptors/distribution/L1LastChanceInterceptor.java
```diff
@@ -76,40 +76,50 @@ public void start() {
...
+               cdl.localNodeIsOwner((key = command.getKey()))) {
+            if (trace) {
+               log.trace("Sending additional invalidation for requestors if necessary.");
+            }
+            // Send out a last attempt L1 invalidation in case if someone cached the L1
+            // value after they already received an invalidation
+            blockOnL1FutureIfNeeded(l1Manager
```
pruivo: isn't it blocking the thread?

core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
```diff
@@ -297,31 +292,38 @@ private void abortL1UpdateOrWait(Object key) throws InterruptedException {
          }
          return ctx.continueInvocation();
       }
-      Future<?> l1InvalidationFuture = invalidateL1(ctx, command, assumeOriginKeptEntryInL1);
-      Object returnValue = ctx.forkInvocationSync(command);
-      processInvalidationResult(command, l1InvalidationFuture);
-      removeFromLocalL1(ctx, command);
-      return ctx.shortCircuit(returnValue);
+      Future<?> l1InvalidationFuture = invalidateL1InCluster(ctx, command, assumeOriginKeptEntryInL1);
+      return ctx.forkInvocation(command, (ctx1, command1, rv, throwable) -> {
```
pruivo: why ```forkInvocation()```? In this case, I was expecting a ```onReturn()``` with ```continueInvocation()```.

core/src/main/java/org/infinispan/interceptors/impl/AsyncSubCommandInvoker.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.stream.Stream;
+
+/**
+ * Invoke a sequence of sub-commands, allowing the caller to define a single
+ * {@link org.infinispan.interceptors.SequentialInterceptor.ForkReturnHandler}.
+ *
+ * @author Dan Berindei
+ * @since 9.0
+ */
+public class AsyncSubCommandInvoker<T> implements SequentialInterceptor.ForkReturnHandler {
```
pruivo: why the generic? doesn't seem to be used.

core/src/main/java/org/infinispan/interceptors/impl/BaseSequentialInvocationContext.java
```diff
@@ -73,29 +72,74 @@
...
+         Throwable throwable = extractCompletableFutureException(t);
+         return invokeReturnHandlersSync(newCommand, null, throwable);
       } finally {
+         action = INVOKE_NEXT;
          nextInterceptor = savedInterceptorNode;
+         nextReturnHandler = savedReturnHandler;
+      }
+   }
+
+   private Throwable extractCompletableFutureException(Throwable t) {
```
pruivo: wouldn't be better to add catch clauses for ```ExecutionException``` and ```CompletionException```?

core/src/main/java/org/infinispan/interceptors/impl/BaseSequentialInvocationContext.java
```diff
@@ -223,118 +257,246 @@ private void invokeNext(VisitableCommand command, Object returnValue, Throwable
...
+      }
    }
 
-   private Object doInvokeNextSync(VisitableCommand command, InterceptorListNode interceptorNode)
+   private Object invokeInterceptorsSync(VisitableCommand command, InterceptorListNode firstInterceptorNode)
          throws Throwable {
+      InterceptorListNode interceptorNode = firstInterceptorNode;
       SequentialInterceptor interceptor = interceptorNode.interceptor;
-      nextInterceptor = interceptorNode.nextNode;
+      interceptorNode = beforeVisit(command, interceptorNode, interceptor);
```
pruivo: god... what is the problem with ```forEach``` interceptor?

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
```diff
@@ -54,32 +53,39 @@ public void init(StateTransferLock stateTransferLock, Configuration configuratio
...
-         return ctx.shortCircuit(result);
+         ctx.onReturn((ctx1, command1, rv, throwable) -> {
+            if (isOwner && currentTopologyId() != commandTopologyId) {
+               throw new OutdatedTopologyException(
+                     "Cache topology changed while the command was executing: expected " +
+                           commandTopologyId + ", got " + currentTopologyId());
+            }
+            return null;
+         });
+         return ctx.continueInvocation();
```
pruivo: it isn't needed since ```onReturn()``` already returns ```CONTINUE_INVOCATION```

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
```diff
@@ -221,46 +221,18 @@ public void processEntry(MarshalledEntry marshalledEntry, AdvancedCacheLoader.Ta
...
+            throw throwable;
+
+         CacheSet<CacheEntry<K, V>> entrySet = (CacheSet<CacheEntry<K, V>>) rv;
+         if (!enabled || hasSkipLoadFlag(command)) {
+            return CompletableFuture.completedFuture(entrySet);
+         }
+         CacheSet<CacheEntry<K, V>> wrappedEntrySet = new WrappedEntrySet(command, entrySet);
+         return CompletableFuture.completedFuture(wrappedEntrySet);
+      });
+      return ctx.continueInvocation();
```
pruivo: it isn't needed since ```onReturn()``` already returns ```CONTINUE_INVOCATION```

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
```diff
@@ -288,47 +260,18 @@ public void close() {
...
+            throw throwable;
+
+         CacheSet<K> keySet = (CacheSet<K>) rv;
+         if (!enabled || hasSkipLoadFlag(command)) {
+            return CompletableFuture.completedFuture(keySet);
+         }
+         CacheSet<K> wrappedKeySet = new WrappedKeySet(command, keySet);
+         return CompletableFuture.completedFuture(wrappedKeySet);
+      });
+      return ctx.continueInvocation();
```
pruivo: it isn't needed since ```onReturn()``` already returns ```CONTINUE_INVOCATION```

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -155,14 +155,22 @@ private Transaction suspendRunningTx(TxInvocationContext ctx) throws SystemExcep
...
+         if (!isProperWriter(ctx, command, command.getKey()))
+            return null;
+
+         Object key = command.getKey();
+         boolean resp = persistenceManager.deleteFromAllStores(key, BOTH);
+         if (trace)
+            getLog().tracef("Removed entry under key %s and got response %s from CacheStore", key, resp);
+         return null;
+      });
+      return ctx.continueInvocation();
```
pruivo: it isn't needed since ```onReturn()``` already returns ```CONTINUE_INVOCATION```

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -175,43 +183,66 @@ private Transaction suspendRunningTx(TxInvocationContext ctx) throws SystemExcep
...
-      if (getStatisticsEnabled()) cacheStores.incrementAndGet();
+         Object key = command.getKey();
+         storeEntry(ctx, key, command);
+         if (getStatisticsEnabled())
+            cacheStores.incrementAndGet();
+         return null;
+      });
 
-      return ctx.shortCircuit(returnValue);
+      return ctx.continueInvocation();
```
pruivo: it isn't needed since ```onReturn()``` already returns ```CONTINUE_INVOCATION```

pruivo: I'm gonna stop spamming. there are some other places like this.

core/src/main/java/org/infinispan/interceptors/compat/BaseTypeConverterInterceptor.java
danberindei: I think i can skip it and merge the if statements, same as for `GetKeyValueCommand`.

core/src/main/java/org/infinispan/interceptors/distribution/DistributionBulkInterceptor.java
danberindei: `visitKeySetCommand` uses a trick (`cache.cacheEntrySet()`) to invoke a completely different command synchronously without `forkInvocationSync()`, and needs to skip the rest of the interceptors for the `KeySetCommand`. We'd have to return a `CompletableFuture.completedFuture(keySet)` only if we executed the other interceptors and created the key set in a `ReturnHandler`.

Personally I'd rather remove the `KeySetCommand` and to this wrapping in `CacheImpl`, but this code only ever gets executed on the originator, and we don't have/need an async version of `Cache.keySet()`, so I decided to keep this interceptor as is.

core/src/main/java/org/infinispan/interceptors/distribution/L1LastChanceInterceptor.java
danberindei: TBH I'm not 100% convinced we should keep them either... they should always be the same, they're only here to allow the reuse of `ReturnHandler` instances (e.g. `InvocationContextInterceptor.defaultReturnHandler`). 
This helps reduce the allocation rate, but as you've noticed it also makes `ReturnHandler` lambdas quite a bit uglier.

core/src/main/java/org/infinispan/interceptors/impl/NotificationInterceptor.java
```diff
@@ -18,6 +18,14 @@
  */
 public class NotificationInterceptor extends DDSequentialInterceptor {
    private CacheNotifier notifier;
+   private ReturnHandler transactionCompleteReturnHandler = (ctx1, command1, rv, throwable) -> {
+      if (throwable != null)
+         return null;
```
pruivo: what is the difference between throwing the ```throwable``` and return ```null```? They do the same IMO but in some places you throw it and in others you just return ```null```.
I'm wondering if would it be useful to create 2 new interfaces/methods. One to be invoked on a successful execution and the other in a failed execution. Just because I'm seeing a lot of this checks.

core/src/main/java/org/infinispan/interceptors/distribution/L1LastChanceInterceptor.java
danberindei: It is, this PR is still not doing anything to make the interceptor chain actually stop blocking.

core/src/main/java/org/infinispan/interceptors/impl/MarshalledValueInterceptor.java
```diff
@@ -246,52 +252,67 @@ protected void start() {
...
-         public CloseableSpliterator<K> spliterator() {
-            return new IteratorAsSpliterator.Builder<>(iterator())
-                    .setEstimateRemaining(super.spliterator().estimateSize())
-                    .setCharacteristics(Spliterator.CONCURRENT | Spliterator.DISTINCT | Spliterator.NONNULL)
-                    .get();
-         }
+                  @Override
+                  public CloseableIterator<K> iterator() {
+                     // We pass a null ctx, since this is always local invocation - if it was remote it would use
+
```
pruivo: doesn't need so much new lines :)

core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
danberindei: This is one of the subtler limitations of the async interceptors... `ReturnHandler`s are not allowed to call `forkInvocation`, only `ForkReturnHandler`s are. In other words, you cannot call `forkInvocation()` after you called `onReturn()`. And because `removeFromLocalL1()` calls `forkInvocation()`, the interceptor can't use `onReturn()`.

The `ReturnHandler`/`ForkReturnHandler` distinction is quite annoying at times, but on the positive side helps that you can call `onReturn` multiple times in the same `visit*Command` method, and it works like nested `try/finally` blocks. Also, `onReturn` is a bit cheaper :)

core/src/main/java/org/infinispan/interceptors/impl/BaseSequentialInvocationContext.java
danberindei: It would work, but then I'd need 3 `catch` blocks doing almost the same thing.

core/src/main/java/org/infinispan/interceptors/impl/BaseSequentialInvocationContext.java
danberindei: I had to manually unroll the loop to get acceptable performance in local mode... adding some comments.

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
danberindei: Yeah, there was a time when `onReturn` didn't return a `CF<Void>`, and some of the interceptors got stuck in the past...

core/src/main/java/org/infinispan/interceptors/impl/NotificationInterceptor.java
danberindei: Yep, they are exactly the same, when `throwable != null`.

I've wanted to add specialized interfaces for success/exception/finally, but things get complicated very soon if you want to avoid creating 2 lambdas for a single interceptor, so I wanted to get a better idea of how these are used. Also, I'm kind of hoping to merge `ReturnHandler` and `ForkReturnHandler` at the same time, to avoid having 8 interfaces.

core/src/main/java/org/infinispan/interceptors/distribution/L1LastChanceInterceptor.java
pruivo: I have a couple of suggestions:
* change the arguments name. ```ctx1``` is kind of weird to read. you can use a ```r``` before each argument: ```rCtx, rCommand, rValue, rThrowable``` to show that is the return.
* creating fields to store the ```ReturnHandler``` is also strange. wouldn't be better to create a private method (example ```CompletableFuture<Object> doSomethingOnReturn(...)```) and do ```ctx.onReturn(this::doSomethingOnReturn)```.

core/src/main/java/org/infinispan/interceptors/impl/CacheMgmtInterceptor.java
```diff
@@ -70,63 +70,62 @@ public void start() {
...
-         start = timeService.time();
-
-      Object retval = ctx.forkInvocationSync(command);
+      if (!statisticsEnabled || !ctx.isOriginLocal())
+         return ctx.continueInvocation();
 
-      if (statisticsEnabled && ctx.isOriginLocal()) {
+      long start = timeService.time();
+      return ctx.onReturn((ctx1, command1, rv, throwable) -> {
          long intervalMilliseconds = timeService.timeDuration(start, TimeUnit.MILLISECONDS);
```
pruivo: isn't it possible the ```ReturnHandler``` to be invoke on a different thread? I'm wondering if our stats would be less reliable now...

core/src/main/java/org/infinispan/interceptors/totalorder/TotalOrderVersionedDistributionInterceptor.java
```diff
@@ -74,9 +74,4 @@ protected void prepareOnAffectedNodes(TxInvocationContext<?> ctx, PrepareCommand
          transactionRemotelyPrepared(ctx);
       }
    }
-
-   @Override
-   protected Log getLog() {
```
pruivo: it should be there :)

core/src/main/java/org/infinispan/interceptors/xsite/OptimisticBackupInterceptor.java
```diff
@@ -19,19 +19,7 @@
 
    @Override
    public CompletableFuture<Void> visitPrepareCommand(TxInvocationContext ctx, PrepareCommand command) throws Throwable {
```
pruivo: can be removed IMO

core/src/main/java/org/infinispan/interceptors/impl/CacheMgmtInterceptor.java
danberindei: I don't see why it would be less reliable... even when calling it from the same thread, it's quite likely for the calls to happen on different CPU cores, so I'm assuming the JVM can handle it.

core/src/main/java/org/infinispan/interceptors/distribution/L1LastChanceInterceptor.java
danberindei: On Wed, May 18, 2016 at 2:05 PM, Pedro Ruivo <notifications@github.com>
wrote:

> In
> core/src/main/java/org/infinispan/interceptors/distribution/L1LastChanceInterceptor.java
> <https://github.com/infinispan/infinispan/pull/4324#discussion_r63683816>:
>
> > @@ -76,40 +76,50 @@ public void start() {
> >     }
> >
> >     public CompletableFuture<Void> visitDataWriteCommand(InvocationContext ctx, DataWriteCommand command, boolean assumeOriginKeptEntryInL1) throws Throwable {
> > -      Object returnValue = ctx.forkInvocationSync(command);
> > -      Object key;
> > -      if (shouldUpdateOnWriteCommand(command) && command.isSuccessful() &&
> > -            cdl.localNodeIsOwner((key = command.getKey()))) {
> > -         if (trace) {
> > -            log.trace("Sending additional invalidation for requestors if necessary.");
> > +      ctx.onReturn((ctx1, command1, rv, throwable) -> {
> > +         if (throwable != null)
> > +            throw throwable;
> > +
> > +         Object key;
> > +         if (shouldUpdateOnWriteCommand(command) && command.isSuccessful() &&
>
> I have a couple of suggestions:
>
>    - change the arguments name. ctx1 is kind of weird to read. you can
>    use a r before each argument: rCtx, rCommand, rValue, rThrowable to
>    show that is the return.
>    - creating fields to store the ReturnHandler is also strange. wouldn't
>    be better to create a private method (example CompletableFuture<Object>
>    doSomethingOnReturn(...)) and do
>    ctx.onReturn(this::doSomethingOnReturn).
>
>
I guess I could go with `rCtx` and `rCommand`, maybe even `rValue`, but
`rThrowable` just looks odd :)

danberindei:
Rebased @pruivo

core/src/main/java/org/infinispan/context/impl/AbstractTxInvocationContext.java
```diff
@@ -26,7 +26,7 @@
    private final T cacheTransaction;
 
    protected AbstractTxInvocationContext(T cacheTransaction, Address origin,
-                                         SequentialInterceptorChain interceptorChain) {
+                                         AsyncInterceptorChain interceptorChain) {
```
pruivo: it is not used and can be removed. the import can be removed too.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
```diff
@@ -48,11 +47,10 @@ public void setDependencies(RpcManager rpcManager,
 
    @Override
    public CompletableFuture<Void> visitRollbackCommand(TxInvocationContext ctx, RollbackCommand command) throws Throwable {
-      try {
-         return ctx.shortCircuit(ctx.forkInvocationSync(command));
-      } finally {
+      return ctx.onReturn((rCtx, rCommand, rv, throwable) -> {
```
pruivo: ```unlockAllReturnHandler``` should be used.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
```diff
@@ -66,34 +64,24 @@ public void setDependencies(RpcManager rpcManager,
...
-         return ctx.shortCircuit(ctx.forkInvocationSync(command));
-      } finally {
-         //when not invoked in an explicit tx's scope the get is non-transactional(mainly for efficiency).
-         //locks need to be released in this situation as they might have been acquired from L1.
-         if (!ctx.isInTxScope()) lockManager.unlockAll(ctx);
-      }
+      if (ctx.isInTxScope())
+         return ctx.continueInvocation();
+
+      return ctx.onReturn((rCtx, rCommand, rv, throwable) -> {
```
pruivo: ```unlockAllReturnHandler``` should be used.

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
```diff
@@ -66,34 +64,24 @@ public void setDependencies(RpcManager rpcManager,
...
+         if (throwable instanceof OutdatedTopologyException)
+            throw throwable;
 
-   protected final Object invokeNextAndCommitIf1Pc(TxInvocationContext ctx, PrepareCommand command) throws Throwable {
-      Object result = ctx.forkInvocationSync(command);
-      if (command.isOnePhaseCommit() && releaseLockOnTxCompletion(ctx)) {
-         lockManager.unlockAll(ctx);
-      }
-      return result;
+         releaseLockOnTxCompletion(((TxInvocationContext) rCtx));
```
pruivo: using the original ```ctx``` would avoid the cast :)

core/src/main/java/org/infinispan/interceptors/AsyncInterceptor.java
```diff
@@ -53,20 +54,23 @@
...
+    * TODO Have different interfaces for processing the return value and for finally-like handlers that don't have access to the return value
+    * TODO At least prohibit null return values and require an already-completed CF with a known value instead.
     */
    interface ReturnHandler {
-      CompletableFuture<Object> handle(InvocationContext ctx, VisitableCommand command, Object rv,
+      CompletableFuture<Object> handle(InvocationContext rCtx, VisitableCommand rCommand, Object rv,
             Throwable throwable) throws Throwable;
    }
 
    /*
```
pruivo: missing an ```*``` and some alignment.

core/src/main/java/org/infinispan/interceptors/AsyncInterceptorChain.java
```diff
@@ -8,24 +8,24 @@
...
  * Experimental: The ability to modify the interceptors at runtime may be removed in future versions.
  *
  * @author Dan Berindei
  * @since 9.0
  */
 @Experimental
-public interface SequentialInterceptorChain {
-   List<SequentialInterceptor> getInterceptors();
+public interface AsyncInterceptorChain {
+   List<AsyncInterceptor> getInterceptors();
```
pruivo: javadoc missing. it would be good to know if the return list is immutable or not.

core/src/main/java/org/infinispan/interceptors/AsyncInterceptorChain.java
```diff
@@ -8,24 +8,24 @@
...
 
    /**
     * Inserts the given interceptor at the specified position in the chain (o based indexing).
```
pruivo: zero based maybe :)

core/src/main/java/org/infinispan/interceptors/AsyncInterceptorChain.java
```diff
@@ -8,24 +8,24 @@
...
     *
     * @throws IllegalArgumentException if the position is invalid (e.g. 5 and there are only 2 interceptors in the
     *                                  chain)
     */
-   void addInterceptor(SequentialInterceptor interceptor, int position);
+   void addInterceptor(AsyncInterceptor interceptor, int position);
 
    /**
     * Removes the interceptor at the given postion.
```
pruivo: typo ```position```

core/src/main/java/org/infinispan/interceptors/AsyncInterceptorChain.java
```diff
@@ -43,23 +43,22 @@
    /**
     * Removes all the occurences of supplied interceptor type from the chain.
```
pruivo: typo ```occurrences```

core/src/main/java/org/infinispan/interceptors/AsyncInterceptorChain.java
```diff
@@ -43,23 +43,22 @@
...
     *
     * @return true if the interceptor was added; i.e. the afterInterceptor exists
     */
-   boolean addInterceptorAfter(SequentialInterceptor toAdd, Class<? extends
-         SequentialInterceptor> afterInterceptor);
+   boolean addInterceptorAfter(AsyncInterceptor toAdd, Class<? extends
+         AsyncInterceptor> afterInterceptor);
 
    /**
     * Adds a new interceptor in list after an interceptor of a given type.
```
pruivo: typo: after => before

core/src/main/java/org/infinispan/interceptors/AsyncInterceptorChain.java
```diff
@@ -43,23 +43,22 @@
...
     *
     * @return true if the interceptor was added; i.e. the afterInterceptor exists
```
pruivo: typo: ```beforeInterceptor```

core/src/main/java/org/infinispan/interceptors/impl/AsyncInterceptorChainImpl.java
```diff
@@ -34,21 +34,21 @@
  */
```
pruivo: typo in javadoc: ``` a chain```

core/src/main/java/org/infinispan/interceptors/impl/AsyncInterceptorChainImpl.java
```diff
@@ -34,21 +34,21 @@
...
+public class AsyncInterceptorChainImpl implements AsyncInterceptorChain {
    // Using the same list type everywhere may help with the optimization of the invocation context methods
-   private static final ImmutableListCopy<SequentialInterceptor> EMPTY_INTERCEPTORS_LIST =
-         new ImmutableListCopy<>(new SequentialInterceptor[0]);
-   private static final Log log = LogFactory.getLog(SequentialInterceptorChainImpl.class);
+   private static final ImmutableListCopy<AsyncInterceptor> EMPTY_INTERCEPTORS_LIST =
+         new ImmutableListCopy<>(new AsyncInterceptor[0]);
+   private static final Log log = LogFactory.getLog(AsyncInterceptorChainImpl.class);
 
    final ComponentMetadataRepo componentMetadataRepo;
```
pruivo: can be private

core/src/main/java/org/infinispan/interceptors/locking/AbstractTxLockingInterceptor.java
danberindei: Yep, avoiding capturing locals in the lambdas is both hard to to consistently and uglier.

core/src/main/java/org/infinispan/interceptors/AsyncInterceptor.java
pruivo: the ```*``` is still missing to make it a javadoc comments :)

core/src/main/java/org/infinispan/interceptors/impl/AsyncInterceptorChainImpl.java
```diff
@@ -125,17 +129,18 @@ public void removeInterceptor(Class<? extends SequentialInterceptor> clazz) {
       }
    }
 
-   protected boolean interceptorMatches(SequentialInterceptor interceptor,
-                                        Class<? extends SequentialInterceptor> clazz) {
-      Class<? extends SequentialInterceptor> interceptorType = interceptor.getClass();
+   protected boolean interceptorMatches(AsyncInterceptor interceptor,
```
pruivo: can be private

core/src/main/java/org/infinispan/context/AsyncInvocationContext.java
```diff
@@ -65,14 +65,14 @@
...
-   CompletableFuture<Void> forkInvocation(VisitableCommand newCommand, SequentialInterceptor.ForkReturnHandler forkReturnHandler);
+   CompletableFuture<Void> forkInvocation(VisitableCommand newCommand, AsyncInterceptor.ForkReturnHandler forkReturnHandler);
 
    /**
     * Fork the command invocation, and execute the remaining interceptors (and their return handlers)
     * synchronously.
     *
     * <p>Usually it is easier to use than
-    * {@link #forkInvocation(VisitableCommand, SequentialInterceptor.ReturnHandler)}.
+    * {@link #forkInvocation(VisitableCommand, AsyncInterceptor.ReturnHandler)}.
```
pruivo: ```ForkReturnHandler``` instead of ```ReturnHandler```

core/src/test/java/org/infinispan/context/impl/AsyncInvocationContextTest.java
```diff
@@ -115,42 +114,42 @@ public void testShortCircuitAsync() throws ExecutionException, InterruptedExcept
...
+      AsyncInterceptorChain chain = newInterceptorChain(interceptors);
       InvocationContext context = newInvocationContext(chain);
 
       Object returnValue = chain.invoke(context, command);
       assertEquals("subCommand", returnValue);
    }
 
-   protected SingleKeyNonTxInvocationContext newInvocationContext(SequentialInterceptorChain chain) {
-      // Actual implementation doesn't matter, we are only testing the BaseSequentialInvocationContext methods
+   protected SingleKeyNonTxInvocationContext newInvocationContext(AsyncInterceptorChain chain) {
```
pruivo: interceptor chain can be removed.

core/src/main/java/org/infinispan/interceptors/BaseCustomAsyncInterceptor.java
```diff
@@ -21,7 +21,7 @@
  * @author Dan Berindei
  * @since 9.0
  */
-public class BaseCustomSequentialInterceptor extends DDSequentialInterceptor {
+public class BaseCustomAsyncInterceptor extends DDAsyncInterceptor {
    protected Cache<?, ?> cache;
    protected EmbeddedCacheManager embeddedCacheManager;
```
pruivo: the ```setup``` condition is wrong. it should check if cache and the cache manager is not null but it is checking the cache twice.

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
```diff
@@ -221,46 +221,17 @@ public void processEntry(MarshalledEntry marshalledEntry, AdvancedCacheLoader.Ta
...
-               }
-            };
-      return ctx.shortCircuit(wrappedEntrySet);
+      return ctx.onReturn((rCtx, rCommand, rv, throwable) -> {
+         if (throwable != null)
+            throw throwable;
+
+         CacheSet<CacheEntry<K, V>> entrySet = (CacheSet<CacheEntry<K, V>>) rv;
+         if (!enabled || hasSkipLoadFlag(command)) {
+            return CompletableFuture.completedFuture(entrySet);
```
pruivo: ```return null``` since the return value don't change.

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
```diff
@@ -288,47 +259,17 @@ public void close() {
...
-               }
-            };
-      return ctx.shortCircuit(wrappedKeySet);
+      return ctx.onReturn((rCtx, rCommand, rv, throwable) -> {
+         if (throwable != null)
+            throw throwable;
+
+         CacheSet<K> keySet = (CacheSet<K>) rv;
+         if (!enabled || hasSkipLoadFlag(command)) {
+            return CompletableFuture.completedFuture(keySet);
```
pruivo: ```return null``` since the return value don't change.

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
pruivo: actually, you can put everything in the first ```if```, right? if throwable isn't null or not enabled or skip load flags

core/src/test/java/org/infinispan/notifications/cachelistener/CacheNotifierFilterTest.java
```diff
@@ -28,6 +28,8 @@
 import org.infinispan.test.TestingUtil;
 import org.infinispan.test.fwk.TestCacheManagerFactory;
 import org.infinispan.transaction.TransactionMode;
+import org.infinispan.util.logging.Log;
```
pruivo: there are a bunch of unused imports that can be removed.

core/src/test/java/org/infinispan/notifications/cachelistener/CacheNotifierFilterTest.java
```diff
@@ -65,12 +67,15 @@ public boolean accept(String key, String oldValue, Metadata oldMetadata, String
 
    @Listener
    private static class TestListener {
+      private static final Log log = LogFactory.getLog(TestListener.class);
       private final List<CacheEntryVisitedEvent> visitedEvents = Collections.synchronizedList(
             new ArrayList<CacheEntryVisitedEvent>());
```
pruivo: the type could be removed from the constructor.

extended-statistics/src/main/java/org/infinispan/stats/topK/CacheUsageInterceptor.java
```diff
@@ -59,31 +71,17 @@
...
-         return ctx.shortCircuit(ctx.forkInvocationSync(command));
-      } catch (WriteSkewException wse) {
-         Object key = wse.getKey();
-         if (streamSummaryContainer.isEnabled() && key != null && ctx.isOriginLocal()) {
-            streamSummaryContainer.addWriteSkewFailed(key);
-         }
-         throw wse;
+      ctx.onReturn(writeSkewReturnHandler);
+
+      if (streamSummaryContainer.isEnabled() && ctx.isOriginLocal()) {
```
pruivo: you could move the ```if``` before ```ctx.onReturn()``` and avoid invoking ```ctx.continueInvocation()```

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -274,38 +312,49 @@ private Transaction suspendRunningTx(TxInvocationContext ctx) throws SystemExcep
...
-            for (Object key : keys) {
-               CacheEntry entry = ctx.lookupEntry(key);
-               if (entry != null) {
-                  if (entry.isRemoved()) {
-                     boolean resp = persistenceManager.deleteFromAllStores(key, BOTH);
-                     if (trace) getLog().tracef("Removed entry under key %s and got response %s from CacheStore", key, resp);
-                  } else if (entry.isChanged() && isProperWriter(ctx, command, key)) {
-                     storeEntry(ctx, key, command);
-                     storedCount++;
+   public <T extends WriteCommand & ParamsCommand, K> CompletableFuture<Void> visitWriteManyCommand(
```
pruivo: can be private and the ```command``` is not needed.

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -274,38 +312,49 @@ private Transaction suspendRunningTx(TxInvocationContext ctx) throws SystemExcep
...
+            case PERSIST:
+               int storedCount = 0;
+               for (Object key : keys) {
+                  CacheEntry entry = rCtx.lookupEntry(key);
+                  if (entry != null) {
+                     if (entry.isRemoved()) {
+                        boolean resp = persistenceManager.deleteFromAllStores(key, BOTH);
+                        if (trace) getLog().tracef("Removed entry under key %s and got response %s from CacheStore", key, resp);
+                     } else {
+                        T command1 = (T) rCommand;
```
pruivo: no need to cast. use ```manyEntriesCommand```

core/src/main/java/org/infinispan/interceptors/BaseCustomAsyncInterceptor.java
danberindei: Nope, it tries to check whether the field `cache` was already initialized with a different cache instance. I think there's already enough info in the exception message below, so I won't add a comment.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
```diff
@@ -87,20 +88,65 @@
...
    private XSiteStateConsumer xSiteStateConsumer;
    private GroupManager groupManager;
+   private CacheNotifier notifier;
 
    private static final Log log = LogFactory.getLog(EntryWrappingInterceptor.class);
    private static final boolean trace = log.isTraceEnabled();
    private static final EnumSet<Flag> EVICT_FLAGS =
          EnumSet.of(Flag.SKIP_OWNERSHIP_CHECK, Flag.CACHE_MODE_LOCAL);
 
+   private final ReturnHandler dataReadReturnHandler = new ReturnHandler() {
```
pruivo: all the ReturnHandler can be converted to lambdas.

core/src/main/java/org/infinispan/interceptors/impl/InvocationContextInterceptor.java
```diff
@@ -53,6 +53,29 @@
    private static final boolean trace = log.isTraceEnabled();
    private volatile boolean shuttingDown = false;
 
+   private final ReturnHandler defaultReturnHandler = new ReturnHandler() {
```
pruivo: could be a lambda.

core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
```diff
@@ -121,27 +124,24 @@ public void start() {
...
-         returnValue = ctx.forkInvocationSync(command);
       }
-      return ctx.shortCircuit(returnValue);
+      return ctx.continueInvocation();
    }
 
-   protected Object performL1Lookup(InvocationContext ctx, boolean runInterceptorOnConflict, Object key,
+   protected CompletableFuture<Void> performL1Lookup(InvocationContext ctx, boolean runInterceptorOnConflict,
+         Object key,
                                      DataCommand command, boolean isEntry) throws Throwable {
```
pruivo: command not needed.

core/src/main/java/org/infinispan/context/impl/LocalTxInvocationContext.java
```diff
@@ -21,8 +20,8 @@
 public class LocalTxInvocationContext extends AbstractTxInvocationContext<LocalTransaction> {
 
    public LocalTxInvocationContext(LocalTransaction localTransaction,
-                                   SequentialInterceptorChain interceptorChain) {
-      super(localTransaction, null, interceptorChain);
+                                   AsyncInterceptorChain interceptorChain) {
```
pruivo: interceptor chain not needed.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -101,122 +101,134 @@ public void configure() {
...
+         newTopologyId = currentTopologyId + 1;
+         targets = new ArrayList<>(cacheTopology.getPendingCH().locateOwners(key));
+         // Remove already contacted nodes
+         targets.removeAll(cacheTopology.getReadConsistentHash().locateOwners(key));
+         if (targets.isEmpty()) {
             if (trace) {
                log.tracef("No valid values found for key '%s' (topologyId=%s).", key, currentTopologyId);
             }
-            break;
+            return CompletableFuture.completedFuture(null);
```
pruivo: maybe we can cache the null completed future in ```CompletableFutures```? it seems to be used everywhere :P

core/src/main/java/org/infinispan/context/impl/NonTxInvocationContext.java
```diff
@@ -28,14 +28,14 @@
 
 
    public NonTxInvocationContext(int numEntries, Address origin, Equivalence<Object> keyEq,
-                                 SequentialInterceptorChain interceptorChain) {
+                                 AsyncInterceptorChain interceptorChain) {
```
pruivo: not used

core/src/main/java/org/infinispan/context/impl/NonTxInvocationContext.java
```diff
@@ -28,14 +28,14 @@
...
       super(origin);
       lookedUpEntries = CollectionFactory.makeMap(CollectionFactory.computeCapacity(numEntries), keyEq, AnyEquivalence.getInstance());
       this.keyEq = keyEq;
    }
 
    public NonTxInvocationContext(Address origin, Equivalence<Object> keyEq,
-                                 SequentialInterceptorChain interceptorChain) {
+                                 AsyncInterceptorChain interceptorChain) {
```
pruivo: not used

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -44,12 +44,14 @@
  * @author Mircea Markus
  */
 public class PessimisticLockingInterceptor extends AbstractTxLockingInterceptor {
+   private static final CompletableFuture<Object> COMPLETED_FALSE = CompletableFuture.completedFuture(false);
```
pruivo: both not used

core/src/main/java/org/infinispan/transaction/impl/RemoteTransaction.java
```diff
@@ -153,4 +159,12 @@ public final TotalOrderRemoteTransactionState getTransactionState() {
          return transactionState;
       }
    }
+
+   public final CompletableFuture<Void> enterSynchronizationAsync(CompletableFuture<Void> releaseFuture) {
+      while (true) {
```
pruivo: a ```do-while``` will avoid checking 2 conditions (the while true and the CAS)

core/src/main/java/org/infinispan/context/impl/RemoteTxInvocationContext.java
```diff
@@ -16,8 +16,8 @@
 public class RemoteTxInvocationContext extends AbstractTxInvocationContext<RemoteTransaction> {
 
    public RemoteTxInvocationContext(RemoteTransaction cacheTransaction,
-                                    SequentialInterceptorChain interceptorChain) {
-      super(cacheTransaction, cacheTransaction.getGlobalTransaction().getAddress(), interceptorChain);
+                                    AsyncInterceptorChain interceptorChain) {
```
pruivo: unused

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
danberindei: I actually prefer the inner classes because I'm logging the return handler class names in `BaseAsyncInvocationContext` and it's a bit easier to find the inner class definition compared to the lambda. Of course, tooling is bound to improve this, so I'll give it a try after I'm done with the other comments (and check the performance with `::` method references, as well).

pruivo: you can still use the inner class with the lambda: 
```
private final ReturnHandler dataReadReturnHandler = (rCtx, rCommand, rv, throwable) -> {
...
}
```

pruivo:
finished second look :)

core/src/main/java/org/infinispan/transaction/impl/RemoteTransaction.java
danberindei: I'm pretty sure the JIT implements `while (true)` as an unconditional jump, but I guess `do/while` looks a bit better.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: I'm not sure they're all the same type, but I'll give it a go.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
danberindei: That's no longer an inner class :)

danberindei: I've changed the "cached" ReturnHandlers to be lambdas and they're consistently about 1% slower than anonymous inner classes in the local read-only test. The generated assembly seems to be about the same, so the difference is probably some extra indirection when a lambda is called. It would almost certainly disappear if I get the JIT to inline the `ReturnHandlers` in `invokeInterceptorsSync`, but for now I'm keeping the inner classes.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: Done

danberindei:
Updated and squashed.

pruivo:
@danberindei needs rebase :(

core/src/main/java/org/infinispan/interceptors/impl/BaseAsyncInvocationContext.java
```diff
@@ -0,0 +1,580 @@
...
+ * @author Dan Berindei
+ * @since 9.0
+ */
+public abstract class BaseAsyncInvocationContext implements InvocationContext {
+   private static final Log log = LogFactory.getLog(BaseAsyncInvocationContext.class);
+   private static final boolean trace = log.isTraceEnabled();
+   // Enable when debugging BaseAsyncInvocationContext itself
+   private static final boolean EXTRA_LOGS = SecurityActions.getBooleanProperty("org.infinispan.debug.BaseAsyncInvocationContext");
+
+   private static final CompletableFuture<Void> CONTINUE_INVOCATION = CompletableFuture.completedFuture(null);
```
pruivo: can use the cached completed future

danberindei: It's already cached :)
The reason I wanted to use a different instance here is because I may want to add a special case for interceptors that return `BaseAsyncInvocationContext.continueInvocation()` in the future.

danberindei:
@pruivo Rebased

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -233,79 +262,98 @@ private Transaction suspendRunningTx(TxInvocationContext ctx) throws SystemExcep
 
 
    private <T extends DataWriteCommand & ParamsCommand> CompletableFuture<Void> visitWriteCommand(InvocationContext ctx, T command) throws Throwable {
```
pruivo: ```command``` is not needed.

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -233,79 +262,98 @@ private Transaction suspendRunningTx(TxInvocationContext ctx) throws SystemExcep
...
-                     if (trace) getLog().tracef("Removed entry under key %s and got response %s from CacheStore", key, resp);
-                  } else if (entry.isChanged() && isProperWriter(ctx, command, key)) {
-                     storeEntry(ctx, key, command);
-                     storedCount++;
+   public CompletableFuture<Void> visitReadWriteManyEntriesCommand(InvocationContext ctx,
+         ReadWriteManyEntriesCommand command) throws Throwable {
+      return visitWriteManyCommand(ctx, command.getKeys());
+   }
+
+   private <T extends WriteCommand & ParamsCommand, K> CompletableFuture<Void> visitWriteManyCommand(
```
pruivo: removing the type ```K``' and changing the set for ```<?>``` seems more correct to me :)

pruivo:
last 2 comments and it is good to go :)

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
danberindei: I don't know if that `K` was ever used :)

danberindei:
It took me a while to notice the comments, but I've now updated it.

pruivo:

PR #4333, opened 2016-05-10
ISPN-6618 IT test for script exec over HR on secure cache

vjuranek:
https://issues.jboss.org/browse/ISPN-6618


tristantarrant:
Pushed, thanks

PR #4335, opened 2016-05-12
ISPN-6599 PutMapCommand partially applied during topology changes

danberindei:
https://issues.jboss.org/browse/ISPN-6599

The actual fix is the handling of `CacheNotFoundResponse` in `JGroupsTransport`, the rest is to fix the tests.

### DIFF
--- a/core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
+++ b/core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
@@ -47,6 +47,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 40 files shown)


danberindei:
Updated to revert the changes in `ClusterTopologyManagerImpl.CacheTopologyFilterReuser`.

gustavonalle:
testing

danberindei:
@gustavonalle Your latest log shows a NPE delivering a message, and I think it's coming from `CacheTopologyFilterReuser`:

    2016-05-13 14:24:21,379 ERROR [org.jgroups.protocols.UNICAST3] (OOB-10,server2) JGRP000039: server2: failed to deliver OOB message [dst: server2, src: server0 (4 headers), size=151 bytes, flags=OOB|DONT_BUNDLE|NO_TOTAL_ORDER]: java.lang.NullPointerException

I've added some logging in `JGroupsResponseFilterAdapter` to collect a stack trace, could you run the test again?

gustavonalle:
@danberindei compilation failure

gustavonalle:
looking

gustavonalle:
pulling

gustavonalle:
Integrated to master,  thanks @danberindei !

core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
```diff
@@ -448,9 +448,18 @@ private void recoverClusterStatus(int newViewId, final boolean isMergeView, fina
...
-            if (i == 0 || !(e.getCause() instanceof TimeoutException))
-               throw e;
-            log.debug("Timed out waiting for cluster status responses, trying again");
+            if (i != 0) {
+               if (e.getCause() instanceof TimeoutException) {
+                  log.debug("Timed out waiting for cluster status responses, trying again");
+               } else if (e.getCause() instanceof SuspectException) {
+                  if (transport.getMembers().containsAll(clusterMembers)) {
+                     log.debug("Received a CacheNotFoundResponse from one of the members, trying again");
+                     Thread.sleep(getGlobalTimeout() / CLUSTER_RECOVERY_ATTEMPTS / 2);
```
rvansa: Hi @danberindei, do you recall why you sleep here? With default settings, it's 12 seconds, and therefore, during partition, when the node installs view with unreachable members and later a view with solely itself (as it became the only node in its partition), there's first view installation + 12s delay before it becomes degraded.

danberindei: @rvansa The problem I was having is that with FORK, a node would become a member in the JGroups view before having a `CommandAwareRpcDispatcher`, and if we didn't wait at all we'd be using up all the retries in a few milliseconds (see `ConcurrentStartForkChannelTest`). Perhaps we could avoid the wait if we had an explicit "cache manager join" command to tell the coordinator which nodes are really running Infinispan.

PR #4342, opened 2016-05-13
ISPN-6629 Enable Netty Epoll support by default in the Hot Rod server

gustavonalle:
https://issues.jboss.org/browse/ISPN-6629


tristantarrant:
Pushed to master, thanks

PR #4346, opened 2016-05-16
ISPN-6570 Perform disconnect instead of quit

vjuranek:
https://issues.jboss.org/browse/ISPN-6570


tristantarrant:
Pushed to master and 8.2.x

PR #4350, opened 2016-05-16
ISPN-7284 Cluster counters

pruivo:
This is just a preview of the initial implementation about the distributed counters.

EDIT:
https://issues.jboss.org/browse/ISPN-7284
Ready for review.


distributed-counter/src/main/java/org/infinispan/counter/CounterManager.java
```diff
@@ -0,0 +1,117 @@
...
+ * <p>
+ * It is thread-safe in the way that multiples threads can retrieve/create {@link Counter}s. If it is the first time a
+ * {@link Counter} is created, concurrent threads may block until it is properly initialized.
+ *
+ * @author Pedro Ruivo
+ * @since 9.0
+ */
+public class CounterManager {
+
+   public static String COUNTER_CACHE_NAME = "__counters__";
```
tristantarrant: we need to formalize the convention for internal caches, but it should start with a triple _

distributed-counter/src/main/java/org/infinispan/counter/impl/entries/ThresholdCounterEntry.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * A {@link CounterEntry} with thresholds.
+ * <p>
+ * It ensures that the thresholds are never violated when updating its value.
+ *
+ * @author Pedro Ruivo
+ * @since 7.0
```
tristantarrant: 9.0

core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
```diff
@@ -6,6 +6,7 @@
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 import org.testng.ITestContext;
+import org.testng.AssertJUnit;
```
slaskawi: Unnecessary change

distributed-counter/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+            </resource>
+        </resources>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>generate-blueprint</id>
+                        <goals>
```
slaskawi: This is probably no longer needed.

distributed-counter/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+        <dependency>
+            <groupId>org.kohsuke.metainf-services</groupId>
+            <artifactId>metainf-services</artifactId>
+            <optional>true</optional>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <resources>
+            <resource>
```
slaskawi: This is also unnecessary

distributed-counter/src/main/java/org/infinispan/counter/Counter.java
```diff
@@ -0,0 +1,84 @@
...
+ * <p>
+ * It allows to update the counter by adding and removing. All the operations are perform asynchronously and they return
+ * a {@link CompletableFuture} when completed.
+ * <p>
+ * The implementation may support weakly consistent value via {@link #weakGetValue()}.
+ *
+ * @author Pedro Ruivo
+ * @since 9.0
+ */
+public interface Counter {
```
slaskawi: Just a silly thought.. shouldn't we use extend `AtomicLong` or expose very similar (if not the same) contract?

distributed-counter/src/main/java/org/infinispan/counter/impl/JGroupsCounter.java
```diff
@@ -0,0 +1,41 @@
+package org.infinispan.counter.impl;
+
+import org.jgroups.blocks.atomic.Counter;
+import org.jgroups.blocks.atomic.CounterService;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * // TODO: Document this
```
slaskawi: Javadoc

slaskawi:
In general looks fine to me... I'm just wondering about this `Counter` contract.. Can we make it look more similar to `AtomcLong`.

Nice work. BTW - do you have any performance or consistency data about this?

distributed-counter/src/main/java/org/infinispan/counter/Counter.java
rvansa: `AtomicLong` is class, this is iface. I see no gain in mixing those, and it's better when users are clearly aware that a counter is a distributed one rather than super-fast atomics.

distributed-counter/src/main/java/org/infinispan/counter/Counter.java
```diff
@@ -0,0 +1,84 @@
...
+   /**
+    * @return The counter name.
+    */
+   String getName();
+
+
+   /**
+    * @return The current value
+    */
+   CompletableFuture<Long> getValue();
```
rvansa: Why not just `get()`?

distributed-counter/src/main/java/org/infinispan/counter/CounterListenerCloseable.java
```diff
@@ -0,0 +1,16 @@
+package org.infinispan.counter;
+
+/**
+ * As a return of {@link Counter#addListener(CounterListener)}, it is used to un-register the {@link CounterListener}
+ * via {@link CounterListenerCloseable#close()} method.
+ *
+ * @author Pedro Ruivo
+ * @since 9.0
+ */
+public interface CounterListenerCloseable<T extends CounterListener> extends AutoCloseable {
```
rvansa: I would rather call that `CounterListenerHandle` or something like that (maybe generic `Handle` would do where `CounterListener.addListener` would return `Handle<CounterListener>`. This sounds almost like CloseableCounterListener which would suggest that it's the listener & closeable.

distributed-counter/src/main/java/org/infinispan/counter/CounterManager.java
```diff
@@ -0,0 +1,131 @@
...
+    * The most up to date value is only calculated via {@link Counter#getValue()} and it may not include in progress
+    * operations.
+    *
+    * @param name         the counter name.
+    * @param initialValue the counter initial value.
+    * @return the {@link Counter} instance.
+    */
+   public Counter getOrCreateWeakCounter(String name, long initialValue) {
+      return counters.computeIfAbsent(name, counterName -> {
+         WeakInfinispanCounter counter = new WeakInfinispanCounter(counterName, cacheManager.getCache(COUNTER_CACHE_NAME));
```
rvansa: I really hate when a method does not honor the contract on it's interface. Could we have `WeakCounter` interface where the affected methods would return `CompletableFuture<Void>`?

distributed-counter/src/main/java/org/infinispan/counter/impl/JGroupsCounter.java
rvansa: Such experiments probably don't require javadoc.

distributed-counter/src/main/java/org/infinispan/counter/CounterManager.java
```diff
@@ -0,0 +1,131 @@
...
+    * @param name         the counter name.
+    * @param initialValue the counter initial value.
+    * @return the {@link Counter} instance.
+    * @throws IllegalArgumentException if the min, max or initial value are not valid.
+    */
+   public Counter getOrCreateCounter(String name, long initialValue, long minThreshold, long maxThreshold) {
+      if (minThreshold >= initialValue || initialValue >= maxThreshold) {
+         throw new IllegalArgumentException("Initial value must be between the min and max threshold.");
+      }
+      return counters.computeIfAbsent(name, counterName -> {
```
rvansa: So if there's another counter with different thresholds, and possibly even weak one, defined, you return that instead without any warning? I would rather check that the configuration (weakness + thresholds) match to the existing one, and throw exception if there's mismatch.

distributed-counter/src/main/java/org/infinispan/counter/impl/atomic/InfinispanCounter.java
```diff
@@ -0,0 +1,164 @@
...
+      this.readWriteMap = ReadWriteMapImpl.create(functionalMap);
+      this.readOnlyMap = ReadOnlyMapImpl.create(functionalMap);
+      this.notificationManager = new NotificationManager();
+      this.weakCounter = null;
+   }
+
+   public void init(long initialValue) {
+      registerListener();
+      CounterValue initialEntry = SimpleCounterValue.createInitialEntry(initialValue);
+      CounterValue existingEntry = cache.putIfAbsent(key, initialEntry);
```
rvansa: Again, no check if this matches the configuration.

distributed-counter/src/main/java/org/infinispan/counter/impl/function/ResetFunction.java
```diff
@@ -0,0 +1,73 @@
...
+         CounterValue newEntry = currentEntry.reset();
+         if (newEntry != null) {
+            entryView.set(newEntry);
+         }
+      }
+      return null;
+   }
+
+   private static class SingletonHolder {
+      private static final ResetFunction INSTANCE = new ResetFunction();
```
rvansa: This pattern should be used when the instance is heavy and should be instantiated lazily on getInstance() call... that's a bit of overkill here, I'd move the instance to the main class.

distributed-counter/src/main/java/org/infinispan/counter/CounterEvent.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @author Pedro Ruivo
+ * @since 9.0
+ */
+public interface CounterEvent {
+
+   /**
+    * @return The counter value.
+    */
+   long getValue();
```
rvansa: Is that the before- or after- value? (the same for state) Can't we have both of them?

distributed-counter/src/main/java/org/infinispan/counter/impl/weak/WeakInfinispanCounter.java
```diff
@@ -0,0 +1,243 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
+import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
+
+/**
+ * A weak consistent {@link Counter} implementation.
+ * <p>
+ * This implementation splits the counter's state in multiple keys to be stored in the cache. When an update operation
```
rvansa: Could you always update the key that is primary-owned by this node?

rvansa: Ah, already implemented :) Worth documenting.

pom.xml
```diff
@@ -23,6 +23,7 @@
       <module>commons-test</module>
       <module>core</module>
       <module>extended-statistics</module>
+      <module>distributed-counter</module>
```
rvansa: I know we've used the term 'distributed counters', but wouldn't 'clustered counters' be more fitting?

distributed-counter/src/main/java/org/infinispan/counter/CounterManager.java
```diff
@@ -0,0 +1,131 @@
...
+   /**
+    * It returns the {@link Counter} with that specific name.
+    * <p>
+    * If the {@link Counter} does not exists, it is created. Only the first invocation will take in consideration the
+    * initial value, min and max thresholds.
+    * <p>
+    * The initial value must be between the min and max value.
+    *
+    * @param name         the counter name.
+    * @param initialValue the counter initial value.
```
rvansa: Missing thresholds javadoc. Are these inclusive or exclusive?

pruivo:
@slaskawi @rvansa thanks for your comments. I'll handle them ASAP.
I haven't decided how to configure the cache for the counters. I have a couple of alternatives and I want to hear your opinion:
* use a well know cache name and the user configure it properly 
* (similar to above) the user specifies the cache to use
* the user creates a configuration template and we create a private cache with it

note that the configuration needs to be validated (e.g. transactions are not currently supported by functional cache)

About consistency tests, I added a few to test the atomic counter. About the performance I want to adapt radargun to test it.

rvansa:
@pruivo In 99% percent of cases the user shouldn't care about the cache configuration - it's not like he would configure the eviction/expiration settings for counters. So there should be a default cache for counters, and that should be independent from the container's default cache.
Template configuration makes sense only if we were creating caches out of it - there's no point in private cache. I doubt anyone will use that cache accidentally.

slaskawi:
@pruivo How about introducing new type of cache?

pruivo:
I'm updated this branch with your comments.
@rvansa I disagree. There are a bunch of properties that need some tunning (example, numOwner, timeouts, number of keys in weak counters, persistence (to survive cluster restarts), partition handling).
@slaskawi what you have in mind?

slaskawi:
@pruivo I was thinking about introducing cache type `COUNTER` just like we have `DIST` or `REPL`... I'm not sure if this make sense...

rvansa:
@slaskawi It does not make sense to me. Maybe if you explained why you want a new cache type. From my POV cache mode says 'how are the data distributed across cluster'.
There is another way to think about it, from configuration perspective the cache mode is on the top so it tells you what options you have (how can you use it).
Anyway, counters are building blocks implemented on top of cache mode, so it's unrelated imo.

pruivo:
rebased and updated. Add documentation to user_guide section.

counter/src/main/java/org/infinispan/counter/CounterManager.java
```diff
@@ -0,0 +1,138 @@
...
+ * <p>
+ * It is thread-safe in the way that multiples threads can retrieve/create {@link Counter}s. If it is the first time a
+ * {@link Counter} is created, concurrent threads may block until it is properly initialized.
+ *
+ * @author Pedro Ruivo
+ * @since 9.0
+ */
+public class CounterManager {
+
+   public static final String COUNTER_CACHE_NAME = "__counters__";
```
tristantarrant: As we add more of these "system" caches, we should really put the names in a single place (and use the same naming convention)

counter/src/main/java/org/infinispan/counter/impl/CounterModuleLifecycle.java
```diff
@@ -0,0 +1,86 @@
...
+      addAdvancedExternalizer(externalizerMap, CounterSnapshot.EXTERNALIZER);
+      addAdvancedExternalizer(externalizerMap, AtomicCounterKey.EXTERNALIZER);
+      addAdvancedExternalizer(externalizerMap, WeakCounterKey.EXTERNALIZER);
+   }
+
+   @Override
+   public void cacheManagerStarted(GlobalComponentRegistry gcr) {
+      final EmbeddedCacheManager cacheManager = gcr.getComponent(EmbeddedCacheManager.class);
+      if (!cacheManager.getCacheConfigurationNames().contains(CounterManager.COUNTER_CACHE_NAME)) {
+         cacheManager.defineConfiguration(CounterManager.COUNTER_CACHE_NAME, createConfiguration());
```
tristantarrant: This should use the InternalCacheRegistry approach instead.

counter/src/main/java/org/infinispan/counter/CounterManager.java
```diff
@@ -0,0 +1,138 @@
...
+         }
+      }
+   }
+
+   private static <T> T validateCounter(Class<T> tClass, Object retVal) {
+      Class<?> rClass = retVal.getClass();
+      if (tClass.isAssignableFrom(rClass)) {
+         return tClass.cast(retVal);
+      }
+      throw new IllegalStateException(String.format("Unable to return counter '%s' from '%s'", tClass.getSimpleName(), rClass.getSimpleName()));
```
tristantarrant: Use the logging framework

counter/src/main/java/org/infinispan/counter/CounterManager.java
```diff
@@ -0,0 +1,138 @@
...
+    * @param name         the counter name.
+    * @param initialValue the counter initial value.
+    * @param minThreshold the minimum threshold (inclusive).
+    * @param maxThreshold the maximum threshold (inclusive).
+    * @return the {@link Counter} instance.
+    * @throws IllegalArgumentException if the min, max or initial value are not valid.
+    */
+   public Counter getOrCreateCounter(String name, long initialValue, long minThreshold, long maxThreshold) {
+      if (minThreshold > initialValue || initialValue > maxThreshold) {
+         throw new IllegalArgumentException("Initial value must be between the min and max threshold inclusive.");
```
tristantarrant: Use the logging framework

counter/src/main/java/org/infinispan/counter/impl/CounterModuleLifecycle.java
tristantarrant: It would probably make sense to introduce volatile and persistent counters. For this we should make the cache persistent (easily done via InternalCacheRegistry flags) and then use the SKIP_CACHE_STORE/LOAD accordingly

counter/src/main/java/org/infinispan/counter/impl/strong/AbstractStrongCounter.java
```diff
@@ -0,0 +1,187 @@
...
+
+   @Override
+   public CompletableFuture<Boolean> compareAndSet(long expect, long update) {
+      return readWriteMap.eval(key, new CompareAndSetFunction<>(expect, update)).thenApply(this::handleCASResult);
+   }
+
+   protected abstract Boolean handleCASResult(CounterState state);
+
+   @CacheEntryModified
+   public void updateState(CacheEntryEvent<StrongCounterKey, CounterValue> event) {
```
rvansa: I don't think it's a right thing to have a compulsory weak value on strong counters, because the listeners here bring some overhead. When the user wants a strong counter, let's give him exactly that. When he wants eventually updated counter, he should use the weak one.

counter/src/main/java/org/infinispan/counter/impl/weak/WeakCounterImpl.java
```diff
@@ -0,0 +1,309 @@
...
+ * @since 9.0
+ */
+@Listener(clustered = true, observation = Listener.Observation.POST, sync = false)
+public class WeakCounterImpl implements WeakCounter {
+
+   private static final Log log = LogFactory.getLog(WeakCounterImpl.class, Log.class);
+   private static final AtomicReferenceFieldUpdater<Entry, CounterValue> L1_UPDATER =
+         newUpdater(Entry.class, CounterValue.class, "snapshot");
+
+   private final Entry[] entries;
```
rvansa: I don't think that slicing the counter into several entries will have positive effect in the end. Right now it writes will work better under concurrent updates, but when we triangelize functional commands, too, the entry won't be locked during RPC and there won't be much need for spreading the contention.
So we could use the sliced version for a while but there should be a TODO to remove that (with perf ack) after full triangelization.

counter/src/main/java/org/infinispan/counter/impl/weak/WeakCounterImpl.java
```diff
@@ -0,0 +1,309 @@
...
+            .storage(configuration.storage()).build();
+      for (int i = 1; i < entries.length; ++i) {
+         initEntry(i, zeroConfig);
+      }
+      selector.updatePreferredKeys(cache.getDistributionManager().getWriteConsistentHash());
+   }
+
+   private void initEntry(int index, CounterConfiguration configuration) {
+      try {
+         CounterValue existing = readWriteMap.eval(entries[index].key, new InitializeCounterFunction<>(configuration))
```
rvansa: minor: you could init the entries in parallel

commons/src/main/java/org/infinispan/counter/api/CounterManager.java
```diff
@@ -0,0 +1,74 @@
...
+    * CounterManager#defineCounter(String, CounterConfiguration)} or via global configuration. This method only supports
+    * {@link CounterType#BOUNDED_STRONG} and {@link CounterType#UNBOUNDED_STRONG} counters.
+    *
+    * @param name the counter name.
+    * @return the {@link StrongCounter} instance.
+    * @throws org.infinispan.counter.exception.CounterException              if unable to retrieve the counter.
+    * @throws org.infinispan.counter.exception.CounterConfigurationException if the counter configuration is not valid
+    *                                                                        or it does not exists.
+    */
+   StrongCounter getStrongCounter(String name);
```
rvansa: You should specify if the namespace of strong counters and weak counters overlap. In `CounterManagerImpl.counters` you use `Map<String, Object>` for all counters. However, since you have distinct types of keys for weak/strong, if there's a strong key "foo" and you try to retrieve a weak key "foo" on another node, both get created without exception (if I read the code correctly). On the same node it would fail.

Maybe I missed some detail, but I've also missed a test case for this.

rvansa:
@pruivo If the weak counters weren't sliced, you could actually have one entry (underlying CounterValue) that could be used as both weak and strong counter - strong counter would use the reliable operations, and weak counter would rely on the value being updated by listeners.

commons/src/main/java/org/infinispan/counter/api/CounterManager.java
tristantarrant: I think the namespace should be unique, like we do for caches. Less confusion

commons/src/main/java/org/infinispan/counter/api/StrongCounter.java
```diff
@@ -0,0 +1,94 @@
...
+ * <p>
+ * It provides atomic updates for the counter. All the operations are perform asynchronously and they complete the
+ * {@link CompletableFuture} when completed.
+ * <p>
+ * The implementation may support weakly consistent reads via {@link #weakGetValue()}.
+ *
+ * @author Pedro Ruivo
+ * @since 9.0
+ */
+public interface StrongCounter {
```
pruivo: It can expose the configuration. I'll add it to the ```WeakCounter``` as well.

commons/src/main/java/org/infinispan/counter/api/StrongCounter.java
```diff
@@ -0,0 +1,94 @@
...
+   CompletableFuture<Long> getValue();
+
+   /**
+    * Returns the locally cached value.
+    * <p>
+    * This value may not up-to-date and it may no have the most recent updates from local and remote nodes.
+    *
+    * @return The locally cached value.
+    */
+   default long weakGetValue() {
```
pruivo: initially, my idea was to have a "L1" cache with the counter value. It uses the cluster listeners to update it so it may miss some local updates. However, it gives me more troubles and I'm no longer sure if it is useful to keep it around. Do you think I should remove it from Strong and Weak counters?

counter/src/main/java/org/infinispan/counter/configuration/CounterConfigurationParser.java
```diff
@@ -0,0 +1,184 @@
...
+      for (int i = 0; i < reader.getAttributeCount(); i++) {
+         ParseUtils.requireNoNamespaceAttribute(reader, i);
+         String value = replaceProperties(reader.getAttributeValue(i));
+         Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
+         parserCommonCounterAttributes(reader, builder, i, attribute, value);
+      }
+      while (reader.hasNext() && (reader.nextTag() != XMLStreamConstants.END_ELEMENT)) {
+         Element element = Element.forName(reader.getLocalName());
+         switch (element) {
+            case UPPER_BOUND:
```
pruivo: this was discussed in the last F2F meeting. I don't recall the reasons since we only put the outcome in the minutes :)

counter/src/main/java/org/infinispan/counter/configuration/Reliability.java
```diff
@@ -0,0 +1,23 @@
...
+    * When the cluster merges back, one partition's updates are kept the others partition's updates are lost.
+    */
+   AVAILABLE,
+   /**
+    * If the cluster splits in multiple partitions, one of the is allowed to read/update the counter and the others are
+    * only allowed to read it.
+    * <p>
+    * When the cluster merges back, the updated counter's value is sent back.
+    */
+   CONSISTENT
```
pruivo: For some reason, I though the minory partition was able to read. I'll double check and update the documentation accordingly.

counter/src/main/java/org/infinispan/counter/impl/weak/WeakCounterImpl.java
pruivo: I mostly agree. Although the contention will be reduced by the triangle, having multiple keys allows the node to pick one key in which it is the primary owner. This way, it allows us to pick the fastest update route.

commons/src/main/java/org/infinispan/counter/api/CounterManager.java
pruivo: @rvansa it will fail if a counter is defined as strong and you try to retrieve it as a WeakCounter. When a counter is retrieved from that map, I check if the type is the expected one. If not, throws an exception. 
When the map is populated (lazily), the counter is created based on the configuration stored in the configurationCache (all nodes have access to the same configuration).

counter/src/main/java/org/infinispan/counter/impl/weak/WeakCounterImpl.java
pruivo: meh... it is only invoked once per node. I don't think it would be beneficial.

counter/src/main/java/org/infinispan/counter/impl/strong/StrongCounterKey.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Set;
+
+/**
+ * The key to store in the {@link org.infinispan.Cache} used by {@link StrongCounter}.
+ *
+ * @author Pedro Ruivo
+ * @see StrongCounter
+ * @since 9.0
+ */
+public class StrongCounterKey implements CounterKey {
```
pruivo: hmmm... I don't like. It would complicate it a lot because of the marshalling, I would have to expose its internals and I would have to duplicate code.

commons/src/main/java/org/infinispan/counter/api/StrongCounter.java
rvansa: I think that it should be removed from Strong counters, and Weak counters should rely purely on those clustered listeners.

galderz:
Needs rebasing. Not much activity in last month, so I'm closing since this PR is a bit of a limbo state. Just re-open it when it's ready for new feedback or ready to be integrated.

pruivo:
bringing this back to life :)

tristantarrant:
A couple of failures

counter/src/main/java/org/infinispan/counter/impl/CounterModuleLifecycle.java
```diff
@@ -0,0 +1,361 @@
...
+/**
+ * It register a {@link CounterManagerImpl} to each {@link EmbeddedCacheManager} started and starts the cache on it.
+ *
+ * @author Pedro Ruivo
+ * @since 9.0
+ */
+@MetaInfServices(value = ModuleLifecycle.class)
+public class CounterModuleLifecycle extends AbstractModuleLifecycle {
+
+   public static final String COUNTER_CACHE_NAME = "___counters";
```
pruivo: yes all counters share the `num-onwers` and `availability`. 
About the `num-owner` I don't think a user would have different `num-owners` per counter if the counters are safe (in terms of failures).
About the `availability`, lets see how the users use it. If needed, I'll move it to be configure per counter. But, to be honest, I prefer not to create a bunch of caches for the counters if we try to support any `num-owner`/`availability` configuration...

pruivo:
@rvansa I prefer to take advantage of the lower guarantees of the weak-counter to slice it and improve its write performance.

rvansa:
@pruivo Not sure if I understand the context. I think you've responded to all my comments (haven't checked if you addressed them). I got that you prefer to keep sliced weak counters (I am ok with that), but do you still want to burden strong counters with the weak semantics (registering a listener)?

tristantarrant:
Pushed to master, thanks @pruivo

rvansa:
Sigh, so integrated with the nasty listener-burden on strong counter, without no option to turn that crap off :-/

tristantarrant:
We are still in the alpha phase. There is time to fix it.

pruivo:
@rvansa why is a listener a "weak semantic"? It is just a notification mechanism that is triggered when the counter changes. It doesn't make weaker the strong counter.

rvansa:
@pruivo It doesn't make the counter weaker, but it introduces overhead when all you need is strong semantics.

PR #4352, opened 2016-05-16
ISPN-5097 Server jgroups configuration does not specify defaults for S3

alanfx:
https://issues.jboss.org/browse/ISPN-5097

This change needs to be applied again


tristantarrant:
Rejected. Please open a new PR with the new approach

PR #4353, opened 2016-05-17
ISPN-6644 Use externalizer directly to serialize CommandInvocationId

rvansa:
https://issues.jboss.org/browse/ISPN-6644
As discussed on mailing list.

### DIFF
--- a/core/src/main/java/org/infinispan/commands/CommandInvocationId.java
+++ b/core/src/main/java/org/infinispan/commands/CommandInvocationId.java
@@ -17,10 +17,14 @@ import java.util.concurrent.atomic.AtomicLong;
  *
  * It is used to lock the key for a specific command.
  *
+ * This class is final to prevent issues as it is usually not marshalled
+ * as polymorphic object but directly using {@link #writeTo(ObjectOutput, CommandInvocationId)}
+ * and {@link #readFrom(ObjectInput)}.
+ *
  * @author Pedro Ruivo
  * @since 8.0
  */
-public class CommandInvocationId {
+public final class CommandInvocationId {
 
    private static final AtomicLong nextId = new AtomicLong(0);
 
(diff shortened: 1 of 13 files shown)


pruivo:
it looks good to me but, IMO, you can put the externalizer singleton in ```CommandInvocationId``` and make the internal class private (like I did in here: https://github.com/pruivo/infinispan/blob/t_counters/distributed-counter/src/main/java/org/infinispan/counter/impl/CounterKey.java). It would make the code smaller: ```CommandInvocationId.EXTERNALIZER.writeObject()``` vs ```CommandInvocationId.Externalizer.INSTANCE.writeObject()```

rvansa:
@pruivo I could remove the externalizer alltogether and replace it with static methods, wdyt?

pruivo:
IMO, it would be good to leave there. In case we need to send a ```CommandInvocationId``` in a list or some user adds a custom command and needs to serialize it. 
However, I don't mind to have the static methods and the externalizer can use them.

danberindei:
@pruivo It's not that hard to serialize a list with just the static methods... you only need an externalizer if the list is heterogeneous and some of the elements happen to be CommandInvocationIds - not sure that's something we want to support. Same with custom commands, any custom command that adds a `CommandInvocationId` field can serialize it using the static methods.

For me, the only question is backwards compatibility - the `org.infinispan.commands` package is part of the API, so we need to deprecate the externalizer first. Of course, nobody is ever going to see the deprecation warnings, because the externalizer isn't used directly, but at least we'll have an alternative available for some time before removing it.

pruivo:
@danberindei and what if we need to send a ```CommandInvocationId``` as a response? Are you going to have ```SuccessfulResponse``` to see what class is it? Also, back to the list, it would make you serialize the list by hand.

danberindei:
@pruivo yeah, an externalizer would be helpful if we sent the `CommandInvocationId` as a response, but how likely is that to happen? I would still deprecate it for 9.0, and remove it in 10.0 if we're still not using it by then.

pruivo:
@danberindei ok

wburns:
@pruivo, @danberindei so it sounds like we want to deprecate the externalizer and use the new static method?

pruivo:
@wburns :+1:

danberindei:
+1 @wburns

Sanne:
Nice improvements. Looking forward to these!

rvansa:
@pruivo Thanks for reminding, I though this was already integrated. Updated to the static methods.

pruivo:
testing...

pruivo:
integrated! thanks @rvansa !

PR #4356, opened 2016-05-19
ISPN-6630 and re-enable ISPN-6300 test

galderz:
https://issues.jboss.org/browse/ISPN-6630
https://issues.jboss.org/browse/ISPN-6300


scripting/src/main/java/org/infinispan/scripting/impl/DataTypedCache.java
```diff
@@ -69,22 +77,27 @@ public V get(Object key) {
 
    @Override
    public CacheSet<K> keySet() {
-      CacheStream<K> stream = getDelegate().keySet().stream().map(this::toDataType);
-      return new SetAsCacheSet<>(stream.collect(Collectors.toSet()));
+      CacheStream<K> stream = getDelegate().keySet().stream().map(
+            new ValueToTypedValueFunction<>(dataTypedCacheManager.dataType, dataTypedCacheManager.marshaller));
+      return new SetAsCacheSet<>(stream.collect(
```
wburns: I am not familiar with the script cache stuff, but this will delazify the cache contents and hold it all into memory.  Is this okay?

galderz:
Ups, I had added ISPN-6173 WIP commit by mistake, removing it from here.

scripting/src/main/java/org/infinispan/scripting/impl/DataTypedCache.java
galderz: Indeed it will bring all contents over and it's not ideal, but I could not find an easy solution to avoid that :\. I know you have done a lot of work inside the keySet...etc impls to do that, but I couldn't easily extract it out for this situation. Do you have any ideas?

It would be great if we simply had methods to expose keyStream, valueStream, entryStream and leave those methods returning `Set` instances behind :\

galderz:
Needs some rebase, working on it...

galderz:
Rebased

scripting/src/main/java/org/infinispan/scripting/impl/DataTypedCache.java
wburns: All of the examples of where we do that currently exist in the interceptor stack which rely on the cache to do alot of the other heavy lifting.

Unfortunately this means you would have to do it manually by implementing all of the various methods to do the proper conversion.

I am guessing you would also want to prevent users from doing a write to this keySet, or is that possible?

wburns:
Changes look fine, we should be able to use the cache set mapper from here https://github.com/infinispan/infinispan/pull/4369 and we won't need another commit :)

galderz:
@wburns Excellent, I'll look to integrate #4369 and then use it here :)

galderz:
PR updated, check it out @wburns 😃

core/src/test/java/org/infinispan/functional/decorators/FunctionalAdvancedCache.java
```diff
@@ -643,4 +645,186 @@ public void removeListener(Object listener) {
       }
    }
 
+   private static final class SetAsCacheSet<E> implements CacheSet<E> {
```
wburns: You should be able to use `AbstractDelegatingSet` as the base class here and then only have to implement the CacheSet specific override methods.

core/src/test/java/org/infinispan/functional/decorators/FunctionalAdvancedCache.java
```diff
@@ -643,4 +645,186 @@ public void removeListener(Object listener) {
...
+
+      @Override
+      public String toString() {
+         return "SetAsCacheSet{" +
+            "set=" + set +
+            '}';
+      }
+   }
+
+   private static class CollectionAsCacheCollection<E> implements CacheCollection<E> {
```
wburns: You should be able to use `AbstractDelegatingCollection` as the base class here and then only have to implement the CacheCollection specific override methods.

core/src/main/java/org/infinispan/util/CacheCollectionMapper.java
```diff
@@ -0,0 +1,51 @@
...
+ * a different instance with values replaced on request.  This is useful as a
+ * cache collection that is normally lazily evaluated to prevent having to
+ * pull all values into memory which can be a lot faster when checking single
+ * values and can also prevent out of memory issues.
+ *
+ * @param <E> Type of elements in collection before transformation
+ * @param <R> Type of elements in collection after transformation
+ * @since 9.0
+ */
+public class CacheCollectionMapper<E, R> extends CollectionMapper<E, R> implements CacheCollection<R> {
```
wburns: I realized I missed this one in the previous PR later last night, but I figured you had it under control :)

core/src/test/java/org/infinispan/functional/decorators/FunctionalAdvancedCache.java
wburns: Do you think we want this one as an outer class?

core/src/test/java/org/infinispan/functional/decorators/FunctionalAdvancedCache.java
```diff
@@ -643,4 +645,186 @@ public void removeListener(Object listener) {
...
+      }
+
+      @Override
+      public void clear() {
+         set.clear();
+      }
+
+      @Override
+      public CloseableSpliterator<E> spliterator() {
+         return null;
```
wburns: You can use the `Closeables` class to implement this.

wburns:
@galderz looks great! :+1: Just a couple minor cleanup comments.

galderz:
@wburns Updated

wburns:
@galderz unfortunately there are some test failures: http://ci.infinispan.org/viewLog.html?buildId=40000&buildTypeId=bt9&guest=1  Is this because now it is not lazy?  Not sure why it would try to do a remote call in a LOCAL cache..

galderz:
@wburns No, the issue is that I was also testing fixes for ISPN-6173 and disabled the tests but these tests are still not working as we know. Restoring the disabled flag on those tests.

galderz:
@wburns Updated

wburns:
@galderz sounds good :+1: ... pulling.

wburns:

PR #4359, opened 2016-05-19
ISPN-6542 Subclasses of AbstractWriteKeyCommand don't marshal the Params params field

rvansa:
* cache enum values
* replaced externalizer with static methods

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/api/functional/Param.java
+++ b/commons/src/main/java/org/infinispan/commons/api/functional/Param.java
@@ -111,6 +111,7 @@ public interface Param<P> {
       PERSIST, SKIP;
 
       public static final int ID = ParamIds.PERSISTENCE_MODE_ID;
+      private static final PersistenceMode[] CACHED_VALUES = values();
 
       @Override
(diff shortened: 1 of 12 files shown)


commons/src/main/java/org/infinispan/commons/api/functional/Param.java
```diff
@@ -128,6 +129,10 @@ public PersistenceMode get() {
       public static PersistenceMode defaultValue() {
          return PERSIST;
       }
+
+      public static PersistenceMode fromOrdinal(int ordinal) {
```
danberindei: The other enums have `valueOf(int)`.

danberindei:
Looks good, the only change I'd suggest is the `fromOrdinal` method name.

rvansa:
@danberindei ok, I've noticed just the `ordinal -> Xxx.CACHED_VALUES[ordinal]` lambdas. And `fromOrdinal` is a better name, but I'll stay consistent. Updated.

danberindei:
Integrated, thanks Radim!

PR #4361, opened 2016-05-19
ISPN-6651 Reduce WARN verbosiness due to missing netty native transport

gustavonalle:
https://issues.jboss.org/browse/ISPN-6651

Output is reduced to a single line:

```
15:37:03,377 WARN  [org.infinispan.server.core.transport.NettyTransport] (MSC service thread 1-7) ISPN005028: Native EPoll transport not available, using NIO instead: /tmp/libnetty-transport-native-epoll209796259230488777.so: Error relocating /tmp/libnetty-transport-native-epoll209796259230488777.so: __strndup: symbol not found
```

### DIFF
--- a/server/core/src/main/scala/org/infinispan/server/core/logging/JavaLog.java
+++ b/server/core/src/main/scala/org/infinispan/server/core/logging/JavaLog.java
@@ -121,6 +121,6 @@ public interface JavaLog extends org.infinispan.util.logging.Log {
    CacheConfigurationException noSniDomainConfigured();
 
    @LogMessage(level = WARN)
-   @Message(value = "EPoll based Eventloop not available, using NIO instead", id = 5028)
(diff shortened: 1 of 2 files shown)


danberindei:
Integrated, thanks Gustavo!

danberindei:
I should have waited a bit before integrating it :)
Should we log the warning if EPoll is disabled in the configuration, or if running on Windows?

gustavonalle:
@danberindei I think not, the WARN should be only present on Linuxes, I will send another PR to change it

gustavonalle:
@danberindei Actually, it does not print the warning when running on Non-Linux platforms or when native epoll is disabled via ```-Dinfinispan.server.channel.epoll=false```, since ```isEPollAvailable``` is lazily evaluated

PR #4363, opened 2016-05-20
Ispn 6547 revert previous fix and add new

slaskawi:
https://issues.jboss.org/browse/ISPN-6547

The previous fix was wrong. CDI spec says:
```
If both include filters and excludes filters are specified, only class names which match at least one include filter and don't match any exclude filter are scanned.
```

So in this case we can only play with exclusions.


danberindei:
Integrated, thanks Sebastian!

PR #4366, opened 2016-05-20
Ispn 6691 Simplify Distributed Stream local intermediate operations

wburns:
https://issues.jboss.org/browse/ISPN-6691
https://issues.jboss.org/browse/ISPN-5874

### DIFF
--- a/core/src/main/java/org/infinispan/BaseCacheStream.java
+++ b/core/src/main/java/org/infinispan/BaseCacheStream.java
@@ -1,0 +1,123 @@
+package org.infinispan;
+
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.Set;
+import java.util.Spliterator;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.stream.BaseStream;
+
+/**
+ * Interface that defines the base methods of all streams returned from a {@link Cache}.  This interface
+ * is useful to hold a reference to any of the types while still being able to invoke some methods.
+ * @author wburns
+ * @since 9.0
+ */
+public interface BaseCacheStream<T, S extends BaseStream<T, S>> extends BaseStream<T, S> {
+   /**
+    * This would disable sending requests to all other remote nodes compared to one at a time. This can reduce memory
+    * pressure on the originator node at the cost of performance.
+    * <p>Parallel distribution is enabled by default except for {@link CacheStream#iterator()} &
+    * {@link CacheStream#spliterator()}</p>
+    * @return a stream with parallel distribution disabled
+    */
+   BaseCacheStream sequentialDistribution();
+
+   /**
+    * This would enable sending requests to all other remote nodes when a terminal operator is performed.  This
+    * requires additional overhead as it must process results concurrently from various nodes, but should perform
+    * faster in the majority of cases.
+    * <p>Parallel distribution is enabled by default except for {@link CacheStream#iterator()} &
+    * {@link CacheStream#spliterator()}</p>
+    * @return a stream with parallel distribution enabled.
+    */
+   BaseCacheStream parallelDistribution();
+
+   /**
+    * Filters which entries are returned by what segment they are present in.  This method can be substantially more
+    * efficient than using a regular {@link CacheStream#filter(Predicate)} method as this can control what nodes are
+    * asked for data and what entries are read from the underlying CacheStore if present.
+    * @param segments The segments to use for this stream operation.  Any segments not in this set will be ignored.
+    * @return a stream with the segments filtered.
+    */
+   BaseCacheStream filterKeySegments(Set<Integer> segments);
+
+   /**
+    * Filters which entries are returned by only returning ones that map to the given key.  This method will
+    * be faster than a regular {@link CacheStream#filter(Predicate)} if the filter is holding references to the same
+    * keys.
+    * @param keys The keys that this stream will only operate on.
+    * @return a stream with the keys filtered.
+    */
+   BaseCacheStream filterKeys(Set<?> keys);
+
+   /**
+    * Controls how many keys are returned from a remote node when using a stream terminal operation with a distributed
+    * cache to back this stream.  This value is ignored when terminal operators that don't track keys are used.  Key
+    * tracking terminal operators are {@link CacheStream#iterator()}, {@link CacheStream#spliterator()},
+    * {@link CacheStream#forEach(Consumer)}.  Please see those methods for additional information on how this value
+    * may affect them.
+    * <p>This value may be used in the case of a a terminal operator that doesn't track keys if an intermediate
+    * operation is performed that requires bringing keys locally to do computations.  Examples of such intermediate
+    * operations are {@link CacheStream#sorted()}, {@link CacheStream#sorted(Comparator)},
+    * {@link CacheStream#distinct()}, {@link CacheStream#limit(long)}, {@link CacheStream#skip(long)}</p>
+    * <p>This value is <b>always</b> ignored when this stream is backed by a cache that is not distributed as all
+    * values are already local.</p>
+    * @param batchSize The size of each batch.  This defaults to the state transfer chunk size.
+    * @return a stream with the batch size updated
+    */
+   BaseCacheStream distributedBatchSize(int batchSize);
+
+   /**
+    * Allows registration of a segment completion listener that is notified when a segment has completed
+    * processing.  If the terminal operator has a short circuit this listener may never be called.
+    * <p>This method is designed for the sole purpose of use with the {@link CacheStream#iterator()} to allow for
+    * a user to track completion of segments as they are returned from the iterator.  Behavior of other methods
+    * is not specified.  Please see {@link CacheStream#iterator()} for more information.</p>
+    * <p>Multiple listeners may be registered upon multiple invocations of this method.  The ordering of notified
+    * listeners is not specified.</p>
+    * @param listener The listener that will be called back as segments are completed.
+    * @return a stream with the listener registered.
+    */
+   BaseCacheStream segmentCompletionListener(SegmentCompletionListener listener);
+
+   /**
+    * Disables tracking of rehash events that could occur to the underlying cache.  If a rehash event occurs while
+    * a terminal operation is being performed it is possible for some values that are in the cache to not be found.
+    * Note that you will never have an entry duplicated when rehash awareness is disabled, only lost values.
+    * <p>Most terminal operations will run faster with rehash awareness disabled even without a rehash occuring.
+    * However if a rehash occurs with this disabled be prepared to possibly receive only a subset of values.</p>
+    * @return a stream with rehash awareness disabled.
+    */
+   BaseCacheStream disableRehashAware();
+
+   /**
+    * Sets a given time to wait for a remote operation to respond by.  This timeout does nothing if the terminal
+    * operation does not go remote.
+    * <p>If a timeout does occur then a {@link java.util.concurrent.TimeoutException} is thrown from the terminal
+    * operation invoking thread or on the next call to the {@link Iterator} or {@link Spliterator}.</p>
+    * <p>Note that if a rehash occurs this timeout value is reset for the subsequent retry if rehash aware is
+    * enabled.</p>
+    * @param timeout the maximum time to wait
+    * @param unit the time unit of the timeout argument
+    * @return a stream with the timeout set
+    */
+   BaseCacheStream timeout(long timeout, TimeUnit unit);
+
+   /**
+    * Functional interface that is used as a callback when segments are completed.  Please see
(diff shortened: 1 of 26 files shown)


core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
```diff
@@ -762,73 +711,30 @@ public void accept(Iterable<E> es) {
...
+            a.segmentCompleted(segments);
+         }
+         catch (Throwable e1) {
+            try {
+               b.segmentCompleted(segments);
+            }
+            catch (Throwable e2) {
+               try {
+                  e1.addSuppressed(e2);
+               } catch (Throwable ignore) {}
```
danberindei: If there's an exception happening in `addSuppressed()`, I think it's important enough to throw that instead of the original exception.

wburns: To be honest I uses the similar logic that is present in `java.util.Streams#composeWithExceptions` since I figured they knew best what to do.

wburns: Also just to note this was copied from `DistributedCacheStream` and moved down, now that the primitive streams properly support all the methods of `CacheStreamable`

core/src/main/java/org/infinispan/CacheStreamable.java
```diff
@@ -0,0 +1,120 @@
...
+    * efficient than using a regular {@link CacheStream#filter(Predicate)} method as this can control what nodes are
+    * asked for data and what entries are read from the underlying CacheStore if present.
+    * @param segments The segments to use for this stream operation.  Any segments not in this set will be ignored.
+    * @return a stream with the segments filtered.
+    */
+   CacheStreamable filterKeySegments(Set<Integer> segments);
+
+   /**
+    * Filters which entries are returned by only returning ones that map to the given key.  This method will <b>always</b>
+    * be faster than a regular {@link CacheStream#filter(Predicate)} if any keys must be retrieved remotely or if a
```
danberindei: I wouldn't say *always*, `cache.entrySet().stream().filterKeys(cache.keySet())` is very likely to be slower regardless of clustering/persistence :)

wburns: Which case are you thinking?  Maybe I should have been more specific on the original JIRA this was added, but this was implying the filter method was holding references to the same keys.

danberindei: Yeah, the javadoc doesn't say anything like that :)

core/src/main/java/org/infinispan/CacheStreamable.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.stream.BaseStream;
+
+/**
+ * Interface that signifies which streams are able to be used with a {@link Cache}.  This method
+ * is useful to hold a reference to any of the types while still being able to invoke some methods.
+ */
+public interface CacheStreamable<T, S extends BaseStream<T, S>> extends BaseStream<T, S> {
```
danberindei: I'm not sure about the name, to me it suggests something that has a `stream()` method. Couldn't we follow the JDK's lead and name it `BaseCacheStream`?

core/src/main/java/org/infinispan/CacheStreamable.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Iterator;
+import java.util.Set;
+import java.util.Spliterator;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.stream.BaseStream;
+
+/**
+ * Interface that signifies which streams are able to be used with a {@link Cache}.  This method
```
danberindei: This looks a bit hard to read, I'm not sure how a user could use a stream with a Cache.
s/This method/This interface/

core/src/main/java/org/infinispan/stream/impl/IntermediateCacheStream.java
```diff
@@ -0,0 +1,456 @@
...
+import java.util.function.ToIntFunction;
+import java.util.function.ToLongFunction;
+import java.util.stream.Collector;
+import java.util.stream.DoubleStream;
+import java.util.stream.IntStream;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+
+/**
+ * An intermediate cache stream that the intermediate results return a reference
```
danberindei: s/that the intermediate results return a reference/returned by the intermediate operations/ ?

core/src/main/java/org/infinispan/stream/impl/IntermediateCacheStreamSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.infinispan.CacheStreamable;
+import org.infinispan.commons.util.CloseableIterator;
+import org.infinispan.stream.impl.local.AbstractLocalCacheStream;
+
+import java.util.Set;
+import java.util.stream.BaseStream;
+
+/**
+ * Stream supplier that is used when a local intermediate operation is invoked, requiring a partial remote and local
```
danberindei: s/partial/combined/ ?

core/src/main/java/org/infinispan/stream/impl/IntermediateType.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * The various intermediate types.  Note that the local intermediate operation can be a type other than the
+ * resulting stream, so we have to keep track of what it was to avoid {@link ClassCastException} issues.
+ * <p>
+ * Currently the streams generated vary based on if it wants an iterator.  This should hopefully be faster, as an
+ * iterator can require many messages, however it requires a bit more memory.  Thus terminal operations that can
```
danberindei: This should hopefully be faster: It's not clear what "this" means, or how the generated streams actually vary depending on the local intermediate operation.

core/src/main/java/org/infinispan/stream/impl/IntermediateCacheStream.java
```diff
@@ -0,0 +1,456 @@
...
+ */
+public class IntermediateCacheStream<R> implements CacheStream<R> {
+   private CacheStreamable remoteStream;
+   private final IntermediateType type;
+   private LocalCacheStream<R> localCacheStream;
```
danberindei: Should be `localStream` to match with `remoteStream`.

danberindei:
Scala compilation error:

[ERROR] /tmp/privatebuild/infinispan/server/hotrod/src/main/scala/org/infinispan/server/hotrod/iteration/IterationManager.scala:36: error: type SegmentCompletionListener is not a member of object org.infinispan.CacheStream
[INFO] class IterationSegmentsListener extends CacheStream.SegmentCompletionListener {
[INFO]                                                     ^
[ERROR] /tmp/privatebuild/infinispan/server/hotrod/src/main/scala/org/infinispan/server/hotrod/iteration/IterationManager.scala:111: error: type mismatch;
[INFO]  found   : org.infinispan.server.hotrod.iteration.IterationSegmentsListener
[INFO]  required: org.infinispan.CacheStreamable.SegmentCompletionListener
[INFO]       val iterator = filteredStream.getOrElse(stream).asInstanceOf[CacheStream[CacheEntry[AnyRef, AnyRef]]].segmentCompletionListener(segmentListener).iterator()
[INFO]                                                                                                                                       ^
[ERROR] two errors found

danberindei:
@wburns Does this compilation error in the server mean we're also breaking compatibility for the users?

core/src/main/java/org/infinispan/CacheStreamable.java
wburns: I can make it a little more clear.  The big benefit is that it will only send the keys that map to the segments on each node and also that you only have to operate upon that subset of keys instead of possibly all if there are any other intermediate operations before the filter.

core/src/main/java/org/infinispan/CacheStreamable.java
wburns: Yeah TBH I couldn't think of a good name.  Unfortunately `BaseCacheStream` doesn't fit in my opinion as it implies it is abstract to me.  Maybe `CacheStreamOperations` ? Like I said I am kinda at a loss for the name.

core/src/main/java/org/infinispan/CacheStreamable.java
wburns: K I will try to clean it up a bit.

core/src/main/java/org/infinispan/stream/impl/IntermediateCacheStream.java
wburns: Let me clean it up.

core/src/main/java/org/infinispan/stream/impl/IntermediateCacheStreamSupplier.java
wburns: Sure

core/src/main/java/org/infinispan/stream/impl/IntermediateType.java
wburns: I apologize, this was a remnant from when I was either performing an iterator operation like now or for terminal operations that require the entire stream just calling toArray.  I can remove this section.

core/src/main/java/org/infinispan/stream/impl/IntermediateCacheStream.java
wburns: Sure

wburns:
You would require a refactoring of code.  But AFAIK that is fine for 9.x.  I can for sure add a comment in the upgrading guide though.  This should be fine for backports though as well since I could have the new interface extend the old one and retain it if needed.

To be honest though users shouldn't really be using this method :-1:

core/src/main/java/org/infinispan/CacheStreamable.java
wburns: Also removed always to make it less strong :)

core/src/main/java/org/infinispan/CacheStreamable.java
wburns: Sorry I forgot that `BaseCacheStream` is an interface as well.  Poor naming imo, but I can change it to that.

danberindei:
@wburns it would be great if you could extend the old interface and preserve binary compatibility, even if it's a new major version.

wburns:
@danberindei The problem is that the interface is defined on `CacheStream` which now extends `BaseCacheStream` and it really makes sense to have the interface on the new one since it is usuable by `IntCacheStream`, `DoubleCacheStream` and `LongCacheStream.

It seems a bit odd that it would be going back to the `CacheStream` argument when using `LongCacheStream` for example.

wburns:
Updated, rebased and added upgrading comment.

core/src/main/java/org/infinispan/BaseCacheStream.java
```diff
@@ -0,0 +1,120 @@
...
+    * @param timeout the maximum time to wait
+    * @param unit the time unit of the timeout argument
+    * @return a stream with the timeout set
+    */
+   BaseCacheStream timeout(long timeout, TimeUnit unit);
+
+   /**
+    * Functional interface that is used as a callback when segments are completed.  Please see
+    * {@link BaseCacheStream#segmentCompletionListener(SegmentCompletionListener)} for more details.
+    * @since 8.0
```
danberindei: I guess it should be `@since 9.0`

core/src/main/java/org/infinispan/BaseCacheStream.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.stream.BaseStream;
+
+/**
+ * Interface that defines the base methods of all streams returned from a {@link Cache}.  This interface
+ * is useful to hold a reference to any of the types while still being able to invoke some methods.
+ */
+public interface BaseCacheStream<T, S extends BaseStream<T, S>> extends BaseStream<T, S> {
```
danberindei: Should this have a `@since` tag?

core/src/main/java/org/infinispan/BaseCacheStream.java
wburns: Technically it was an interface in 8.0 but it was moved in 9.0, haha.  I can update it.

core/src/main/java/org/infinispan/BaseCacheStream.java
wburns: Thanks, missed this one.

wburns:
Updated since tags.

danberindei:
Integrated, thanks Will!

PR #4367, opened 2016-05-23
ISPN-6572 Improve JCache Remote behaviour when Management port unavailable

galderz:
Couple of changes to improve JCache Remote behaviour:

* Added option that allows JCache remote managament access to be
  disabled. When this option is enabled, creating or removing a cache
  throws an exception indicating that the operation is not
  possible. This option is designed for situations where the server has
  pre-defined all the user caches.
* createCache calls for JCache remote should throw a CacheException when
  cache is predefined in the server when using server access where no
  management API is enabled,
  e.g. infinispan.jcache.remote.managed_access is set to false.


tristantarrant:
Pushed to master. Thanks

PR #4370, opened 2016-05-23
Properly size KnownComponentNames hash maps (avoids 8 unnecessary jav…

rhusar:
…a.util.HashMap#resize calls)

Trivial.


tristantarrant:
Pushed to master, thanks

PR #4371, opened 2016-05-25
ISPN-6702 ClientEventsOOMTest hangs the TRACE build

danberindei:
Also fixed some generics warnings in the test.

https://issues.jboss.org/browse/ISPN-6702

### DIFF
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientEventsOOMTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientEventsOOMTest.java
@@ -43,7 +43,7 @@ public class ClientEventsOOMTest extends MultiHotRodServersTest {
 
    private static BufferPoolMXBean DIRECT_POOL = getDirectMemoryPool();
 
-   private RemoteCache remoteCache;
+   private RemoteCache<Integer, byte[]> remoteCache;
 
    // There is only one Godzilla in heap, but we can use Netty's off-heap pools to multiply them and destroy the world
    private static final byte[] GODZILLA = makeGodzilla();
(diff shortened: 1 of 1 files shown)


anistor:
what was causing the failure?

anistor:
Shouldn't we backport this to 8.2.x too ?

danberindei:
I'm not sure what was causing the failure, I've been running the hotrod-client tests with the same settings (-Xmx=1500m) and I didn't get the OOME.

I changed the CI setting to 2g, which is actually the default in `parent/pom.xml`. I had some problems earlier (native OOMEs) with 2g, but back then I was also giving the parent Surefire JVM 2g, so I'm hoping it will go better this time.

Anyway, we don't have a TRACE build for 8.2.x, so I wouldn't backport it.

wburns:
We could think about making this a stress test instead as well.  @galderz WDYT?

galderz:
@anistor That was original idea by @anistor but the problem with stress tests is that they're not always run. In fact, as a middle ground, `ClientEventsOOMTest` was enhanced to take system properties for the number of entries and sleep time parameters, so that you could run it as a longer, more stressful tests if you want to do so manually.

galderz:
So, I'm against making it a stress test instead. If you want to have a stress test as well this test, that's fine by me, but this simple test which is not hugely stressful, manages to catch very subtle issues which cannot be found any other way.

wburns:
@galderz that is what I figured.  It just sucks that a test has to not run OOM to validate that it is working.  Would be nice if we could just check netty's pool allocator to verify they are returned in the properly, but that test might be a bit complex :(

anistor:
Integrated in master and 8.2.x. Thanks @danberindei !

PR #4373, opened 2016-05-25
ISPN-6706: Do not invoke CacheWriter.purge() if there is no eviction …

ksobolew:
…and the writer has preload enabled.

The thinking behind this is that when these two conditions are met, the cache
store's state is expected to be an exact copy of the transient state, and
purging the store itself is redundant: since the transient state is expected to
be the same as the persistent state, the entry will be removed from the store
IFF it is removed from the transient state.

Caveat: manual calls to evict() will break this assumption.

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
+++ b/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
@@ -13,6 +13,7 @@ import org.infinispan.container.entries.ImmortalCacheEntry;
 import org.infinispan.container.entries.InternalCacheEntry;
 import org.infinispan.context.Flag;
 import org.infinispan.context.InvocationContext;
+import org.infinispan.eviction.EvictionStrategy;
 import org.infinispan.eviction.EvictionType;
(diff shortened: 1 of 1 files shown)


wburns:
@ksobolew Unfortunately CacheWriter purge isn't related to eviction.  This is to remove entries that have expired [1].  Expiration can be configured both via xml and inserted programatically.  Due to the latter we have to always call the purge method.  It is feasible to keep state on the writer such as a boolean to trigger if any entry has ever been able to expire and it not then skip purge.

I am closing this PR, however if you wanted to possibly add the previous optimization we could field that.  Also if you wanted to suggest a change we could discuss it on the dev mailing list.  Info can be found at [2]

[1] http://infinispan.org/docs/9.0.x/user_guide/user_guide.html#_difference_between_eviction_and_expiration
[2] https://lists.jboss.org/mailman/listinfo/infinispan-dev

ksobolew:
I answered in the ISPN-6706 JIRA

wburns:
@ksobolew thanks, moved discussion to there.

wburns:
Although I do like the creativity with the optimization :+1:

PR #4374, opened 2016-05-25
ISPN-6704 Global security auth enabled required

galderz:
* When a cache has auth enabled, its global configuration also requires
  auth being enabled.


tristantarrant:
Pushed to master, thanks

PR #4376, opened 2016-05-26
ISPN-6303 Fix ScriptEngine lookup classloader

galderz:
https://issues.jboss.org/browse/ISPN-6303

* Getting ScriptEngine instances from ScriptEngineManager results in
  class loading calls for referenced classes in script. By default it
  uses the thread context classloader. This change fixes this
  classloader to the ScriptManagerImpl one.


pruivo:
looks good to me. pushing...

pruivo:
integrated! thanks @galderz !

PR #4382, opened 2016-05-27
ISPN-201: make JDBC cache store use DB transactions for commit/rollback

ryanemerson:
https://issues.jboss.org/browse/ISPN-201

### DIFF
--- a/core/src/main/java/org/infinispan/factories/InterceptorChainFactory.java
+++ b/core/src/main/java/org/infinispan/factories/InterceptorChainFactory.java
@@ -41,6 +41,7 @@ import org.infinispan.interceptors.impl.InvocationContextInterceptor;
 import org.infinispan.interceptors.impl.IsMarshallableInterceptor;
 import org.infinispan.interceptors.impl.MarshalledValueInterceptor;
 import org.infinispan.interceptors.impl.NotificationInterceptor;
+import org.infinispan.interceptors.impl.TransactionalStoreInterceptor;
 import org.infinispan.interceptors.impl.TxInterceptor;
 import org.infinispan.interceptors.impl.VersionedEntryWrappingInterceptor;
 import org.infinispan.interceptors.locking.NonTransactionalLockingInterceptor;
@@ -255,6 +256,11 @@ public class InterceptorChainFactory extends AbstractNamedCacheComponentFactory 
                interceptorChain.appendInterceptor(createInterceptor(new ClusteredCacheLoaderInterceptor(), ClusteredCacheLoaderInterceptor.class), false);
             else
                interceptorChain.appendInterceptor(createInterceptor(new CacheLoaderInterceptor(), CacheLoaderInterceptor.class), false);
+
+            boolean transactionalStore = configuration.persistence().stores().stream().anyMatch(StoreConfiguration::transactional);
+            if (transactionalStore && transactionMode.isTransactional())
+               interceptorChain.appendInterceptor(createInterceptor(new TransactionalStoreInterceptor(), TransactionalStoreInterceptor.class), false);
+
             switch (cacheMode) {
                case DIST_SYNC:
                case DIST_ASYNC:
--- a/core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
@@ -23,8 +23,6 @@ import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.container.entries.DeltaAwareCacheEntry;
 import org.infinispan.container.entries.InternalCacheEntry;
 import org.infinispan.container.entries.InternalCacheValue;
-import org.infinispan.container.versioning.EntryVersion;
-import org.infinispan.container.versioning.EntryVersionsMap;
 import org.infinispan.context.Flag;
 import org.infinispan.context.InvocationContext;
 import org.infinispan.context.impl.TxInvocationContext;
@@ -36,8 +34,7 @@ import org.infinispan.jmx.annotations.ManagedAttribute;
 import org.infinispan.jmx.annotations.ManagedOperation;
 import org.infinispan.jmx.annotations.MeasurementType;
 import org.infinispan.marshall.core.MarshalledEntryImpl;
-import org.infinispan.metadata.EmbeddedMetadata;
-import org.infinispan.metadata.Metadata;
+import org.infinispan.persistence.PersistenceUtil;
 import org.infinispan.persistence.manager.PersistenceManager;
 import org.infinispan.transaction.xa.GlobalTransaction;
 import org.infinispan.util.logging.Log;
@@ -425,7 +422,7 @@ public class CacheWriterInterceptor extends JmxStatsCommandInterceptor {
                ice = entryFactory.create(entry);
             }
             MarshalledEntryImpl marshalledEntry = new MarshalledEntryImpl(ice.getKey(), ice.getValue(), internalMetadata(ice), marshaller);
-            persistenceManager.writeToAllStores(marshalledEntry, command.hasFlag(Flag.SKIP_SHARED_CACHE_STORE) ? PRIVATE : BOTH);
+            persistenceManager.writeToAllNonTxStores(marshalledEntry, command.hasFlag(Flag.SKIP_SHARED_CACHE_STORE) ? PRIVATE : BOTH);
          }
          return null;
       }
@@ -461,9 +458,9 @@ public class CacheWriterInterceptor extends JmxStatsCommandInterceptor {
       protected Object visitSingleStore(InvocationContext ctx, FlagAffectedCommand command, Object key) throws Throwable {
          if (isProperWriter(ctx, command, key)) {
             if (generateStatistics) putCount++;
-            InternalCacheValue sv = getStoredValue(key, ctx);
+            InternalCacheValue sv = entryFactory.getValueFromCtxOrCreateNew(key, ctx);
             MarshalledEntryImpl me = new MarshalledEntryImpl(key, sv.getValue(), internalMetadata(sv), marshaller);
-            persistenceManager.writeToAllStores(me, command.hasFlag(Flag.SKIP_SHARED_CACHE_STORE) ? PRIVATE : BOTH);
+            persistenceManager.writeToAllNonTxStores(me, command.hasFlag(Flag.SKIP_SHARED_CACHE_STORE) ? PRIVATE : BOTH);
          }
          return null;
       }
@@ -488,44 +485,14 @@ public class CacheWriterInterceptor extends JmxStatsCommandInterceptor {
    }
 
    void storeEntry(InvocationContext ctx, Object key, FlagAffectedCommand command) {
-      InternalCacheValue sv = getStoredValue(key, ctx);
-      persistenceManager.writeToAllStores(new MarshalledEntryImpl(key, sv.getValue(), internalMetadata(sv), marshaller),
-                                          skipSharedStores(ctx, key, command) ? PRIVATE : BOTH);
+      InternalCacheValue sv = entryFactory.getValueFromCtxOrCreateNew(key, ctx);
+      persistenceManager.writeToAllNonTxStores(new MarshalledEntryImpl(key, sv.getValue(), internalMetadata(sv), marshaller),
+                                               skipSharedStores(ctx, key, command) ? PRIVATE : BOTH);
       if (trace) getLog().tracef("Stored entry %s under key %s", sv, key);
    }
 
    protected boolean skipSharedStores(InvocationContext ctx, Object key, FlagAffectedCommand command) {
       return !ctx.isOriginLocal() || command.hasFlag(Flag.SKIP_SHARED_CACHE_STORE);
-   }
-
-   InternalCacheValue getStoredValue(Object key, InvocationContext ctx) {
-      CacheEntry entry = ctx.lookupEntry(key);
-      if (entry instanceof InternalCacheEntry) {
-         return ((InternalCacheEntry) entry).toInternalCacheValue();
-      } else {
-         if (ctx.isInTxScope()) {
-            EntryVersionsMap updatedVersions =
-                  ((TxInvocationContext) ctx).getCacheTransaction().getUpdatedEntryVersions();
-            if (updatedVersions != null) {
-               EntryVersion version = updatedVersions.get(entry.getKey());
-               if (version != null) {
-                  Metadata metadata = entry.getMetadata();
-                  if (metadata == null) {
-                     // If no metadata passed, assumed embedded metadata
-                     metadata = new EmbeddedMetadata.Builder()
-                           .lifespan(entry.getLifespan()).maxIdle(entry.getMaxIdle())
-                           .version(version).build();
-                     return entryFactory.create(entry.getKey(), entry.getValue(), metadata).toInternalCacheValue();
-                  } else {
-                     metadata = metadata.builder().version(version).build();
-                     return entryFactory.create(entry.getKey(), entry.getValue(), metadata).toInternalCacheValue();
-                  }
-               }
-            }
-         }
-
-         return entryFactory.create(entry).toInternalCacheValue();
-      }
    }
 
    public void disableInterceptor() {
--- a/core/src/main/java/org/infinispan/interceptors/impl/TransactionalStoreInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/TransactionalStoreInterceptor.java
@@ -1,0 +1,154 @@
+package org.infinispan.interceptors.impl;
+
+import org.infinispan.atomic.impl.AtomicHashMap;
+import org.infinispan.commands.AbstractVisitor;
+import org.infinispan.commands.tx.CommitCommand;
+import org.infinispan.commands.tx.PrepareCommand;
+import org.infinispan.commands.tx.RollbackCommand;
+import org.infinispan.commands.write.ApplyDeltaCommand;
+import org.infinispan.commands.write.ClearCommand;
+import org.infinispan.commands.write.PutKeyValueCommand;
+import org.infinispan.commands.write.PutMapCommand;
+import org.infinispan.commands.write.RemoveCommand;
+import org.infinispan.commands.write.ReplaceCommand;
+import org.infinispan.commands.write.WriteCommand;
+import org.infinispan.commons.marshall.StreamingMarshaller;
+import org.infinispan.container.InternalEntryFactory;
+import org.infinispan.container.entries.CacheEntry;
+import org.infinispan.container.entries.DeltaAwareCacheEntry;
+import org.infinispan.container.entries.InternalCacheEntry;
+import org.infinispan.container.entries.InternalCacheValue;
+import org.infinispan.context.InvocationContext;
+import org.infinispan.context.impl.TxInvocationContext;
+import org.infinispan.factories.annotations.ComponentName;
+import org.infinispan.factories.annotations.Inject;
+import org.infinispan.interceptors.DDAsyncInterceptor;
+import org.infinispan.marshall.core.MarshalledEntryImpl;
+import org.infinispan.persistence.PersistenceUtil;
+import org.infinispan.persistence.manager.PersistenceManager;
+import org.infinispan.persistence.support.BatchModification;
+
+import javax.transaction.Transaction;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+
+import static org.infinispan.factories.KnownComponentNames.CACHE_MARSHALLER;
+import static org.infinispan.persistence.manager.PersistenceManager.AccessMode.SHARED;
+import org.infinispan.transaction.impl.AbstractCacheTransaction;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+
+/**
+ * An interceptor which ensures that writes to an underlying transactional store are prepared->committed/rolledback as part
+ * of the 2PC, therefore ensuring that the cache and transactional store(s) remain consistent.
+ *
+ * @author Ryan Emerson
+ * @since 9.0
+ */
+public class TransactionalStoreInterceptor extends DDAsyncInterceptor {
+
+   private PersistenceManager persistenceManager;
+   private InternalEntryFactory entryFactory;
+   private StreamingMarshaller marshaller;
+
+   @Inject
+   protected void init(PersistenceManager persistenceManager, InternalEntryFactory entryFactory,
+                       @ComponentName(CACHE_MARSHALLER) StreamingMarshaller marshaller) {
+      this.persistenceManager = persistenceManager;
+      this.entryFactory = entryFactory;
+      this.marshaller = marshaller;
+   }
+
+   @Override
+   public CompletableFuture<Void> visitPrepareCommand(TxInvocationContext ctx, PrepareCommand command) throws Throwable {
+      if (ctx.isOriginLocal()) {
+         Transaction tx = ctx.getTransaction();
+         Updater modBuilder = new Updater(ctx.getCacheTransaction().getAffectedKeys());
+         List<WriteCommand> modifications = ctx.getCacheTransaction().getAllModifications();
+         for (WriteCommand writeCommand : modifications) {
+            writeCommand.acceptVisitor(ctx, modBuilder);
+         }
+         persistenceManager.prepareAllTxStores(tx, modBuilder.modifications, SHARED);
+      }
+      return ctx.continueInvocation();
+   }
+
+   @Override
+   public CompletableFuture<Void> visitCommitCommand(TxInvocationContext ctx, CommitCommand command) throws Throwable {
+      if (ctx.isOriginLocal()) {
+         persistenceManager.commitAllTxStores(ctx.getTransaction(), SHARED);
+      }
+      return ctx.continueInvocation();
+   }
+
+   @Override
+   public CompletableFuture<Void> visitRollbackCommand(TxInvocationContext ctx, RollbackCommand command) throws Throwable {
+      if (ctx.isOriginLocal()) {
+         persistenceManager.rollbackAllTxStores(ctx.getTransaction(), SHARED);
+      }
+      return ctx.continueInvocation();
+   }
+
+   private class Updater extends AbstractVisitor {
+      private final BatchModification modifications;
+
+      Updater(Set<Object> affectedKeys) {
+         modifications = new BatchModification(affectedKeys);
+      }
+
+      @Override
+      public Object visitPutKeyValueCommand(InvocationContext ctx, PutKeyValueCommand command) throws Throwable {
+         return visitSingleStore(ctx, command.getKey());
+      }
+
+      @Override
+      public Object visitApplyDeltaCommand(InvocationContext ctx, ApplyDeltaCommand command) throws Throwable {
+         CacheEntry entry = ctx.lookupEntry(command.getKey());
+         InternalCacheEntry ice;
+         if (entry instanceof InternalCacheEntry) {
+            ice = (InternalCacheEntry) entry;
+         } else if (entry instanceof DeltaAwareCacheEntry) {
+            AtomicHashMap<?, ?> uncommittedChanges = ((DeltaAwareCacheEntry) entry).getUncommittedChages();
+            ice = entryFactory.create(entry.getKey(), uncommittedChanges, entry.getMetadata(), entry.getLifespan(), entry.getMaxIdle());
+         } else {
+            ice = entryFactory.create(entry);
+         }
+         MarshalledEntryImpl marshalledEntry = new MarshalledEntryImpl(ice.getKey(), ice.getValue(), PersistenceUtil.internalMetadata(ice), marshaller);
+         modifications.addMarshalledEntry(ice.getKey(), marshalledEntry);
+         return null;
+      }
+
+      @Override
+      public Object visitReplaceCommand(InvocationContext ctx, ReplaceCommand command) throws Throwable {
+         return visitSingleStore(ctx, command.getKey());
+      }
+
+      @Override
+      public Object visitPutMapCommand(InvocationContext ctx, PutMapCommand command) throws Throwable {
+         Map<Object, Object> map = command.getMap();
+         for (Object key : map.keySet())
+            visitSingleStore(ctx, key);
+         return null;
+      }
+
+      @Override
+      public Object visitRemoveCommand(InvocationContext ctx, RemoveCommand command) throws Throwable {
+         modifications.removeEntry(command.getKey());
+         return null;
+      }
+
+      @Override
+      public Object visitClearCommand(InvocationContext ctx, ClearCommand command) throws Throwable {
+         // This should never happen as the ClearCommand should never be included in a Tx's modification list
+         throw new UnsupportedOperationException("Clear command not supported inside a transaction");
+      }
+
+      private Object visitSingleStore(InvocationContext ctx, Object key) throws Throwable {
+         InternalCacheValue icv = entryFactory.getValueFromCtxOrCreateNew(key, ctx);
+         modifications.addMarshalledEntry(key, new MarshalledEntryImpl(key, icv.getValue(), PersistenceUtil.internalMetadata(icv), marshaller));
+         return null;
+      }
+   }
+}
--- a/core/src/main/java/org/infinispan/persistence/manager/PersistenceManager.java
+++ b/core/src/main/java/org/infinispan/persistence/manager/PersistenceManager.java
@@ -10,6 +10,10 @@ import org.infinispan.context.InvocationContext;
 import org.infinispan.filter.KeyFilter;
 import org.infinispan.persistence.spi.AdvancedCacheLoader;
 import org.infinispan.marshall.core.MarshalledEntry;
+import org.infinispan.persistence.spi.PersistenceException;
+import org.infinispan.persistence.support.BatchModification;
+
+import javax.transaction.Transaction;
 
 /**
  * Defines the logic for interacting with the chain of external storage.
@@ -55,8 +59,6 @@ public interface PersistenceManager extends Lifecycle {
    void processOnAllStores(Executor executor, KeyFilter keyFilter, AdvancedCacheLoader.CacheLoaderTask task, boolean fetchValue, boolean fetchMetadata, AccessMode mode);
 
    MarshalledEntry loadFromAllStores(Object key, InvocationContext context);
-
-   void writeToAllStores(MarshalledEntry marshalledEntry, AccessMode modes);
 
    /**
     * Returns the store one configured with fetch persistent state, or null if none exist.
@@ -109,6 +111,46 @@ public interface PersistenceManager extends Lifecycle {
 
    void setClearOnStop(boolean clearOnStop);
 
+   /**
+    * Write to all stores that are not transactional. A store is considered transactional if all of the following are true:
+    *
+    * <p><ul>
+    *    <li>The store implements {@link org.infinispan.persistence.spi.TransactionalCacheWriter}</li>
+    *    <li>The store is configured to be transactional</li>
+    *    <li>The cache's TransactionMode === TRANSACTIONAL</li>
+    * </ul></p>
+    *
+    * @param marshalledEntry the entry to be written to all non-tx stores.
+    * @param accessMode the type of access to the underlying store.
+    */
+   void writeToAllNonTxStores(MarshalledEntry marshalledEntry, AccessMode modes);
+
+   /**
+    * Perform the prepare phase of 2PC on all Tx stores.
+    *
+    * @param transaction the current transactional context.
+    * @param batchModification an object containing the write/remove operations required for this transaction.
+    * @param accessMode the type of access to the underlying store.
+    * @throws PersistenceException if an error is encountered at any of the underlying stores.
+    */
+   void prepareAllTxStores(Transaction transaction, BatchModification batchModification,
+                           AccessMode accessMode) throws PersistenceException;
+
+   /**
+    * Perform the commit operation for the provided transaction on all Tx stores.
+    *
+    * @param transaction the transactional context to be committed.
+    * @param accessMode the type of access to the underlying store.
+    */
+   void commitAllTxStores(Transaction transaction, AccessMode accessMode);
+
+   /**
+    * Perform the rollback operation for the provided transaction on all Tx stores.
+    *
+    * @param transaction the transactional context to be rolledback.
+    * @param accessMode the type of access to the underlying store.
+    */
+   void rollbackAllTxStores(Transaction transaction, AccessMode accessMode);
 }
 
 
--- a/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
+++ b/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
@@ -22,6 +22,7 @@ import org.infinispan.interceptors.AsyncInterceptor;
 import org.infinispan.interceptors.AsyncInterceptorChain;
 import org.infinispan.interceptors.impl.CacheLoaderInterceptor;
 import org.infinispan.interceptors.impl.CacheWriterInterceptor;
+import org.infinispan.interceptors.impl.TransactionalStoreInterceptor;
 import org.infinispan.marshall.core.MarshalledEntry;
 import org.infinispan.marshall.core.MarshalledEntryFactory;
 import org.infinispan.metadata.Metadata;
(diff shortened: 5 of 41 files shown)


wburns:
@ryanemerson Tbh having purge transactional shouldn't matter really.  It can do the select and if it fails to remove that is fine, that just means someone else updated it, which they can do :)

core/src/main/java/org/infinispan/interceptors/impl/TransactionalStoreInterceptor.java
```diff
@@ -0,0 +1,172 @@
...
+   @Inject
+   protected void init(TransactionalPersistenceManager persistenceManager, InternalEntryFactory entryFactory,
+                       @ComponentName(CACHE_MARSHALLER) StreamingMarshaller marshaller) {
+      this.persistenceManager = persistenceManager;
+      this.entryFactory = entryFactory;
+      this.marshaller = marshaller;
+   }
+
+   @Start(priority = 15)
+   protected void start() {
```
wburns: Remove?

core/src/main/java/org/infinispan/interceptors/impl/TransactionalStoreInterceptor.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * @author Ryan Emerson
+ */
+public class TransactionalStoreInterceptor extends DDSequentialInterceptor {
+   private TransactionalPersistenceManager persistenceManager;
+   private InternalEntryFactory entryFactory;
+   private StreamingMarshaller marshaller;
+   private boolean enabled = true;
```
wburns: This should be volatile to guarantee other threads see the update.

core/src/main/java/org/infinispan/persistence/manager/TransactionalPersistenceManager.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.infinispan.marshall.core.MarshalledEntry;
+
+import javax.transaction.Transaction;
+import java.util.List;
+
+/**
+ * @author Ryan Emerson
+ */
+public interface TransactionalPersistenceManager extends PersistenceManager {
```
wburns: I am debating maybe we should either just add these all onto the same interface or not tie the interfaces together and just have the implementer implement both.  The latter would allow us to split into 2 implementations as well if that helps anywhere, just wondering.

core/src/main/java/org/infinispan/persistence/manager/TransactionalPersistenceManager.java
```diff
@@ -0,0 +1,21 @@
+package org.infinispan.persistence.manager;
+
+import org.infinispan.marshall.core.MarshalledEntry;
+
+import javax.transaction.Transaction;
+import java.util.List;
+
+/**
+ * @author Ryan Emerson
```
wburns: Description needed.

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -74,7 +72,7 @@
    private final boolean trace = getLog().isTraceEnabled();
    PersistenceConfiguration loaderConfig = null;
    final AtomicLong cacheStores = new AtomicLong(0);
-   protected PersistenceManager persistenceManager;
+   protected TransactionalPersistenceManager persistenceManager;
```
wburns: It looks like this can just be PersistenceManager right?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -474,19 +481,44 @@ private boolean isLocalOnlyLoader(CacheLoader loader) {
...
+         storesMutex.readLock().unlock();
+      }
+   }
+
+   @Override
+   public void prepareAllTxStores(Transaction transaction, List<MarshalledEntry> modifiedEntries, List<Object> keysToRemove,
+                                  AccessMode accessMode) throws Throwable {
+      storesMutex.readLock().lock();
+      try {
+         for (TransactionalCacheWriter writer : txWriters) {
```
wburns: This seems to be inconsistent with above in `writeToAllNonTxStores`.

wburns: Actually it looks like we can use `performOnAllTxStores` here too?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -741,4 +782,39 @@ public Executor getPersistenceExecutor() {
    public StreamingMarshaller getMarshaller() {
       return m;
    }
+
+   private void removeCacheLoader(String storeType, Collection<CacheLoader> collection) {
+      Iterator<CacheLoader> clIt = collection.iterator();
```
wburns: We can actually do `removeIf`.  This way the implementation can decided best way to iterate.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -741,4 +782,39 @@ public Executor getPersistenceExecutor() {
...
+      while (clIt.hasNext()) {
+         CacheLoader w = clIt.next();
+         if (undelegate(w).getClass().getName().equals(storeType))
+            clIt.remove();
+      }
+   }
+
+   private void removeCacheWriter(String storeType, Collection<? extends CacheWriter> collection) {
+      Iterator<? extends CacheWriter> cwIt = collection.iterator();
```
wburns: Same as above.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -585,7 +621,12 @@ private void initializeWriter(StoreConfiguration cfg, CacheWriter writer, Initia
       if (writer != null) {
          if (writer instanceof DelegatingCacheWriter)
```
wburns: So does this mean that for example a singleton writer won't participate in the transaction?

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
```diff
@@ -0,0 +1,195 @@
...
+public abstract class AbstractJdbcStore<K,V> implements AdvancedLoadWriteStore<K,V>, TransactionalCacheWriter<K,V> {
+   private final Map<Transaction, Connection> transactionConnectionMap = new ConcurrentHashMap<>();
+   private AbstractJdbcStoreConfiguration configuration;
+   protected ConnectionFactory connectionFactory;
+   protected TableManager tableManager;
+   protected InitializationContext ctx;
+   protected String cacheName;
+   protected Log log;
+
+   protected abstract Log getLog();
```
wburns: These 2 can be assigned to final variables so we don't have to init them.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
```diff
@@ -0,0 +1,195 @@
...
+         log.debug("Exception while stopping", t);
+      }
+
+      try {
+         if (configuration.connectionFactory() instanceof ManagedConnectionFactory) {
+            log.tracef("Stopping mananged connection factory: %s", connectionFactory);
+            connectionFactory.stop();
+         }
+      } catch (Throwable t) {
+         if (cause == null) cause = t;
```
wburns: If it is not null we could add it as suppressed.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
```diff
@@ -0,0 +1,195 @@
...
+      Connection connection = transactionConnectionMap.get(tx);
+      if (connection == null) {
+         connection = connectionFactory.getConnection();
+         transactionConnectionMap.put(tx, connection);
+      }
+      return connection;
+   }
+
+   protected void destroyTxConnection(Transaction tx) {
+      Connection connection = transactionConnectionMap.get(tx);
```
wburns: Should be able to just do remove and use the returned value if it is not null.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
```diff
@@ -0,0 +1,195 @@
...
+         log.sqlFailureTxRollback(e);
+         throw new PersistenceException(String.format("Error during rollback of JDBC transaction (%s)", tx), e);
+      } finally {
+         destroyTxConnection(tx);
+      }
+   }
+
+   protected Connection getTxConnection(Transaction tx) {
+      Connection connection = transactionConnectionMap.get(tx);
+      if (connection == null) {
```
wburns: A lot of methods assume this won't return null, are these possible to throw NPEs ?

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
```diff
@@ -0,0 +1,195 @@
...
+      tableManager.start();
+   }
+
+   protected void waitForFutureToComplete(Future<Void> future) {
+      try {
+         future.get();
+      } catch (InterruptedException e) {
+         Thread.currentThread().interrupt();
+      } catch (ExecutionException e) {
+         log.errorExecutingParallelStoreTask(e);
```
wburns: Seems like the wrong log message?

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/binary/JdbcBinaryStore.java
```diff
@@ -357,6 +309,101 @@ public void purge(Executor threadPool, PurgeListener task) {
...
+         updateBatch.executeBatch();
+      }
+   }
+
+   private Map<Integer, Bucket> getExistingBuckets(Connection connection, List<MarshalledEntry> modifiedEntries,
+                                                   List<Object> keysToDelete) throws SQLException {
+      List<Integer> bucketIds = Stream.concat(modifiedEntries.stream().map(MarshalledEntry::getKey), keysToDelete.stream())
+            .map(this::getBuckedId)
+            .distinct()
+            .collect(Collectors.toList());
```
wburns: Maybe we can remove distinct and `toList` and just use `toSet`?

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/binary/JdbcBinaryStore.java
```diff
@@ -357,6 +309,101 @@ public void purge(Executor threadPool, PurgeListener task) {
...
+            bucket.addEntry(entryKey, entry);
+            existingBuckets.put(bucketKey, bucket);
+            newBuckets.add(bucket);
+         } else {
+            existingBucket.addEntry(entryKey, entry);
+         }
+      }
+
+      // Remove keys from bucket
+      for (Object entryKey : keysToDelete) {
```
wburns: What would happen if you have a remove and a put in that order?  Will the remove wipe out the modification?

wburns:
@ryanemerson Looks good :+1:

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
ryanemerson: Yes I will change it back. I went with Transactional, as it meant that the method call would be more explicit, i.e. ```writeToAllNonTxStores``` but I think ```writeToAllStores``` through the ```PersistenceManager``` interface should be sufficient.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
ryanemerson: Unfortunately we can't re-use ```performOnAllTxStores``` here, because ```prepareWithModifications``` throws Throwable which isn't allowed within a stream.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
ryanemerson: @wburns Good point. 

This has made me think about the compatibility of a transactional mode and a singleton cache writer in general. As it is the Tx coordinator who performs the store writers during the prepare phase of the Tx, the cluster coordinator (singleton) is not guaranteed to be involved in a Tx in distributed mode. Therefore, I think we will have to have some kind of exception/log message if a user specifies a store store to be singleton and transactional in distributed mode.  

In local and replicated mode, there is no conflict (AFAIK) between a singleton and transactional store.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
ryanemerson: @wburns I have made the change for log, however it is not possible to make this change for ```tableManager``` as it depends on a configuration object which is acquired from the InitializationContext in the init method.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
ryanemerson: @wburns If ```connection == null```, then we actively get a new one. ```connectionFactory.getConnection()``` should never return null, instead it throws a PersistenceException if it is not possible to create a connection.

core/src/main/java/org/infinispan/interceptors/impl/TransactionalStoreInterceptor.java
ryanemerson: @wburns We need the @Start annotation, but I have removed the priority assignment.

wburns: Sorry, I must be missing something, but why is it needed?

ryanemerson: @wburns actually, you're right :). I was getting ``` java.lang.NoSuchMethodException: java.lang.Object.start()``` before, but that was because of my dev setup.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
wburns: Makes sense :)

core/src/main/java/org/infinispan/persistence/spi/TransactionalCacheWriter.java
```diff
@@ -0,0 +1,27 @@
...
+
+   /**
+    * Always write modifications in prepare, as this is the only way we know the FINAL values of the entries. This is
+    * required to handle scenarios where an objects value is changed after the put command has been executed, but before
+    * commit is called on the Tx.
+    * @param transaction
+    * @param modifiedEntries
+    * @param keysToDelete
+    */
+   void prepareWithModifications(Transaction transaction, List<MarshalledEntry> modifiedEntries,
```
wburns: `throws Throwable` seems a bit broad.  Can we limit this to `PersistenceException`?

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/binary/JdbcBinaryStore.java
ryanemerson: @wburns Yes it will, I need to re-think how I pass modified entries to the store in order to take into account the order of operations.

core/src/main/java/org/infinispan/persistence/manager/TransactionalPersistenceManager.java
ryanemerson: @wburns The only problem with separating this into two distinct interfaces is that we rely on the AccessMode enum that is defined in PersitanceManager.java.  I have no objections to simply adding the methods to PersistanceManager, I just thought that the distinction could be beneficial long term, however saying that I can't think of an obvious use-case for when a Tx and non-tx persistence manager would be required

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
ryanemerson: @wburns DelegatingCacheWriters are now supported correctly. However, singleton writers are not supported as they are not compatible with shared stores, which transactional stores require.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/binary/JdbcBinaryStore.java
ryanemerson: @wburns this is resolved now by the use of ```TxModifications``` object. This uses an optimisation that ensures that only the most recent operation on a key (insert/update/delete) is passed through to the underlying stores, hence redundant operations on the stores are avoided.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
wburns: Interesting I didn't realize you couldn't have them together.  To me that makes no sense.  That means singleton stores create a single point of failure, which imo is terrible.  @danberindei does that sound correct to you?  What is the use case for singleton store then?

danberindei: @wburns I never quite figured out what what the use case for singleton stores is, but reading the javadoc of `SingletonCacheWriter` again I'm pretty sure it only makes sense with a shared store. There's no `SingletonCacheLoader` to delegate reads to the coordinator, after all...

wburns: Yes I agree; reading the SingletonCacheWriter it seems to apply it only works with shared cache stores as I would think.  Unfortunately our documentation and verification (`PersistenceConfigurationBuilder`) have it the other way.

ryanemerson:
@wburns I ended up getting rid of the separate TransactionalPersistenceManager interface as it was causing issues with the ComponentRegistry, resulting in multiple instances of PersistenceManagerImpl. The additional methods are now simply part of the PersistenceManager interface.

wburns:
@ryanemerson yeah you can workaround that using [1].  But like I said it is more what feels right.

[1] https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/factories/EmptyConstructorNamedCacheFactory.java#L135

ryanemerson:
@wburns @tristantarrant The test failures don't appear to be related to my changes. Any more thoughts? Or can we integrate the changes?

core/src/main/java/org/infinispan/factories/InterceptorChainFactory.java
```diff
@@ -255,6 +256,11 @@ public AsyncInterceptorChain buildInterceptorChain() {
                interceptorChain.appendInterceptor(createInterceptor(new ClusteredCacheLoaderInterceptor(), ClusteredCacheLoaderInterceptor.class), false);
             else
                interceptorChain.appendInterceptor(createInterceptor(new CacheLoaderInterceptor(), CacheLoaderInterceptor.class), false);
+
+            boolean transactional  = configuration.persistence().stores().stream().anyMatch(StoreConfiguration::transactional);
```
danberindei: Shouldn't we ignore transactional stores if `!transactionMode.isTransactional()`?

core/src/main/java/org/infinispan/interceptors/impl/TransactionalStoreInterceptor.java
```diff
@@ -0,0 +1,158 @@
...
+      }
+      return ctx.continueInvocation();
+   }
+
+   public void disableInterceptor() {
+      enabled = false;
+   }
+
+   private boolean isValidContext(TxInvocationContext ctx) {
+      return isStoreEnabled() && ctx.isOriginLocal() && ctx.isInTxScope();
```
danberindei: The method name is a bit misleading, the context itself can't be invalid. For the tx commands, `ctx.isInTxScope() == true`, so I'd remove that check and inline the rest of the method.

core/src/main/java/org/infinispan/interceptors/impl/TransactionalStoreInterceptor.java
```diff
@@ -0,0 +1,158 @@
...
+   @Override
+   public CompletableFuture<Void> visitRollbackCommand(TxInvocationContext ctx, RollbackCommand command) throws Throwable {
+      if (isValidContext(ctx)) {
+         persistenceManager.rollbackAllTxStores(ctx.getTransaction(), SHARED);
+      }
+      return ctx.continueInvocation();
+   }
+
+   public void disableInterceptor() {
+      enabled = false;
```
danberindei: Couldn't you remove the interceptor instead of disabling it?

core/src/main/java/org/infinispan/persistence/PersistenceUtil.java
```diff
@@ -138,4 +143,36 @@ public static MarshalledEntry loadAndCheckExpiration(PersistenceManager persiste
          return factory.create(loaded.getKey(), loaded.getValue(), (Metadata) null);
       }
    }
+
+   public static InternalCacheValue getValueFromCtxOrCreateNew(Object key, InvocationContext ctx,
+                                                               InternalEntryFactory entryFactory) {
```
danberindei: I'm not sure this logic is specific to persistence...

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -300,11 +304,11 @@ public void disableStore(String storeType) {
                result.add((T) real);
             }
          }
-         for (CacheWriter w : writers) {
+         getWritersStream().forEach(w -> {
```
danberindei: You could avoid creating some extra stream instances with a `forEachWriter` method that calls `writers.forEach()` and then `txWriters.forEach()`

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -80,6 +86,7 @@
    private TimeService timeService;
    private final List<CacheLoader> loaders = new ArrayList<>();
    private final List<CacheWriter> writers = new ArrayList<>();
+   private final List<CacheWriter> txWriters = new ArrayList<>();
```
danberindei: If the `writers` collection now only contains the non-tx writers, you should rename it to `nonTxWriters`.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -474,11 +481,28 @@ private boolean isLocalOnlyLoader(CacheLoader loader) {
 
    @Override
    public void writeToAllStores(MarshalledEntry marshalledEntry, AccessMode mode) {
+      writeToAllNonTxStores(marshalledEntry, mode);
+   }
+
+   @Override
+   public void writeToAllNonTxStores(MarshalledEntry marshalledEntry, AccessMode accessMode) {
```
danberindei: I'd reserve the `All` word for methods that work both on tx and non-tx stores.

core/src/main/java/org/infinispan/persistence/spi/TransactionalCacheWriter.java
```diff
@@ -0,0 +1,40 @@
+package org.infinispan.persistence.spi;
+
+import org.infinispan.persistence.support.BatchModification;
+
+import javax.transaction.Transaction;
+
+/**
+ * Defines the functionality of a transactional store.  This allows the implementing store to participate in the cache's 2PC
+ * protocol to ensure consistency between the underlying store and the cache.
```
danberindei: Should we have more details here about how the consistency is affected when the originator node crashes (with/without recovery)?

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
```diff
@@ -0,0 +1,204 @@
...
+      } finally {
+         destroyTxConnection(tx);
+      }
+   }
+
+   protected Connection getTxConnection(Transaction tx) {
+      Connection connection = transactionConnectionMap.get(tx);
+      if (connection == null) {
+         connection = connectionFactory.getConnection();
+         transactionConnectionMap.put(tx, connection);
```
danberindei: I think you can leak connections if you have 2 threads going through this if at the same time. You should probably use `computeIfAbsent()` instead.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
```diff
@@ -0,0 +1,204 @@
...
+      tableManager = getTableManager();
+      tableManager.setCacheName(cacheName);
+      tableManager.start();
+   }
+
+   public ConnectionFactory getConnectionFactory() {
+      return connectionFactory;
+   }
+
+   protected void waitForFutureToComplete(Future<Void> future) {
```
danberindei: I'm not sure we really need a future/executor if we're going to block the calling thread anyway...

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
danberindei: Thinking about this some more, it's kind of annoying that each store has to keep track of transactions. I'm not sure how feasible it is, but I would like it better if the `prepareWithModifications()` method returned a "store transaction" object, the persistence manager would keep track of it, and would eventually pass it to `commit().

server/integration/infinispan/src/main/resources/schema/jboss-infinispan-core_9_0.xsd
```diff
@@ -836,6 +836,11 @@
                 <xs:documentation>This setting should be set to true when multiple cache instances share the same cache store (e.g., multiple nodes in a cluster using a JDBC-based CacheStore pointing to the same, shared database.) Setting this to true avoids multiple cache instances writing the same modification multiple times. If enabled, only the node where the modification originated will write to the cache store. If disabled, each individual cache reacts to a potential remote update by storing the data to the cache store.</xs:documentation>
             </xs:annotation>
         </xs:attribute>
+        <xs:attribute name="transactional" type="xs:boolean" default="false">
+            <xs:annotation>
+                <xs:documentation>This setting should be set to true when the underlying cache store supports transactions and it is desirable for the underlying store and the cache to always remain synchronized.</xs:documentation>
```
danberindei: I think *always* is over-selling this a bit :)

core/src/main/java/org/infinispan/factories/InterceptorChainFactory.java
ryanemerson: @danberindei Good point! This also has implications for how the PersistenceManager determines which stores to write to when `!transactionMode.isTransactional()`, so I will have to make a few additional changes as well.

core/src/main/java/org/infinispan/persistence/support/BatchModification.java
```diff
@@ -0,0 +1,48 @@
...
+ * A simple wrapper class, necessary for Transactional stores, which allows MarshalledEntries and Object keys to be passed
+ * to a store implementation in order. This class also removes repeated operations on the same key in order to prevent
+ * redundant operations on the underlying store.  For example a tx, {put(1, "Test"); remove(1);}, will be simply written
+ * to the store as {remove(1);}.
+ *
+ * @author Ryan Emerson
+ */
+public class BatchModification {
+   private final Map<Object, MarshalledEntry> marshalledEntries = new HashMap<>();
+   private final Set<Object> affectedKeys = new HashSet<>();
```
ryanemerson: good idea  :+1:

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
ryanemerson: @danberindei I think such an approach could get complex fast when we consider non-jdbc stores. For example, if you have multiple store implementations that are transactional, for example JDBC based and some other Tx store, then it is necessary for multiple connection types to be managed. You could have a `StoreTransaction` interface/abstrct class with multiple implementations, however you then have the problem of how do you cater for multiple Tx stores.

Personally I think leaving the management of store transaction resources upto the store implementation allows for more flexibility, especially if people are writing their own custom stores.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
ryanemerson: @danberindei I agree, there are a few issues with the implementations of purge and process in the String based store, I simply moved the existing code and reduced some repetition with `waitForFutureToComplete`. My plan is to address these issues in a future PR as part of [1][2] as this PR already contains significant changes.

[1] https://issues.jboss.org/browse/ISPN-6843
[2]https://issues.jboss.org/browse/ISPN-6712

ryanemerson:
I have pushed the latest changes based upon @danberindei suggestions

core/src/main/java/org/infinispan/persistence/manager/PersistenceManager.java
```diff
@@ -56,8 +60,6 @@
 
    MarshalledEntry loadFromAllStores(Object key, InvocationContext context);
 
-   void writeToAllStores(MarshalledEntry marshalledEntry, AccessMode modes);
```
danberindei: Please add a `package-info.java` with a `@private` tag, to make it clear this is not part of the API and can change at any time.

core/src/main/java/org/infinispan/factories/InterceptorChainFactory.java
```diff
@@ -255,6 +256,11 @@ public AsyncInterceptorChain buildInterceptorChain() {
                interceptorChain.appendInterceptor(createInterceptor(new ClusteredCacheLoaderInterceptor(), ClusteredCacheLoaderInterceptor.class), false);
             else
                interceptorChain.appendInterceptor(createInterceptor(new CacheLoaderInterceptor(), CacheLoaderInterceptor.class), false);
+
+            boolean transactionalStore = configuration.persistence().stores().stream().anyMatch(StoreConfiguration::transactional);
+            if (transactionalStore && transactionMode.isTransactional())
+               interceptorChain.appendInterceptor(createInterceptor(new TransactionalStoreInterceptor(), TransactionalStoreInterceptor.class), false);
```
danberindei: Thinking about this some more, we should probably validate the configuration and not allow any store with `transactional=true` if the cache is not transactional and/or passivation is enabled.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -358,15 +372,17 @@ public void purgeExpired() {
...
          storesMutex.readLock().lock();
          try {
-            for (CacheWriter w : writers) {
-               if (w instanceof AdvancedCacheExpirationWriter) {
-                  ((AdvancedCacheExpirationWriter)w).purge(persistenceExecutor, advanedListener);
-               } else if (w instanceof AdvancedCacheWriter) {
-                  ((AdvancedCacheWriter)w).purge(persistenceExecutor, key -> {
+            Consumer<CacheWriter> purgeWriter = writer -> {
+               if (writer instanceof AdvancedCacheExpirationWriter) {
+                  ((AdvancedCacheExpirationWriter)writer).purge(persistenceExecutor, advanedListener);
```
danberindei: Please also fix the typo in `advanedListener`

core/src/main/java/org/infinispan/interceptors/impl/TransactionalStoreInterceptor.java
```diff
@@ -0,0 +1,151 @@
...
+
+      @Override
+      public Object visitRemoveCommand(InvocationContext ctx, RemoveCommand command) throws Throwable {
+         modifications.removeEntry(command.getKey());
+         return null;
+      }
+
+      @Override
+      public Object visitClearCommand(InvocationContext ctx, ClearCommand command) throws Throwable {
+         persistenceManager.clearAllStores(SHARED);
```
danberindei: Since clear is always non-transactional, I think you can ignore `ClearCommand` (or even throw an `UnsupportedOperationException`).

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -473,12 +492,24 @@ private boolean isLocalOnlyLoader(CacheLoader loader) {
...
 
    @Override
-   public void writeToAllStores(MarshalledEntry marshalledEntry, AccessMode mode) {
+   public void writeToAllNonTxStores(MarshalledEntry marshalledEntry, AccessMode accessMode) {
       storesMutex.readLock().lock();
       try {
-         for (CacheWriter w : writers) {
-            if (mode.canPerform(configMap.get(w))) {
-               w.write(marshalledEntry);
+         nonTxWriters.stream().filter(canPerform(accessMode)).forEach(writer -> writer.write(marshalledEntry));
```
danberindei: I like the old version better :)

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -741,4 +798,25 @@ public Executor getPersistenceExecutor() {
...
+   }
+
+   private void removeCacheWriter(String storeType, Collection<? extends CacheWriter> collection) {
+      collection.removeIf(cacheWriter -> undelegate(cacheWriter).getClass().getName().equals(storeType));
+   }
+
+   private void performOnAllTxStores(Predicate<CacheWriter> filter, Consumer<TransactionalCacheWriter> action) {
+      storesMutex.readLock().lock();
+      try {
+         txWriters.stream().map(TransactionalCacheWriter.class::cast).filter(filter).forEach(action);
```
danberindei: Couldn't you make `txWriters` a list of `TransactionalCacheWriter` and avoid the cast?

danberindei: Also, the filter predicate seems a bit excessive, since you'll always filter on the access mode. And I'm not sure escape analysis can eliminate the extra allocations...

core/src/main/java/org/infinispan/util/logging/Log.java
```diff
@@ -1426,4 +1426,13 @@ TimeoutException timeoutWaitingForView(int expectedViewId, int currentViewId,
...
    @Message(value = "The %s is no longer supported since version %s. Attribute %s on line %d will be ignored.", id = 415)
    void ignoredAttribute(String componentName, String version, String attributeName, int line);
+
+   @LogMessage(level = ERROR)
+   @Message(value = "Error executing submitted store task", id = 416)
+   void errorExecutingSubmittedStoreTask(@Cause Throwable cause);
+
+   @LogMessage(level = WARN)
+   @Message(value = "It is not possible for a store to be transactional in a non-transactional cache. " +
+         "Transactional flag ignored for store ('%s') in cache ('%s')", id = 417)
```
danberindei: No need for the brackets. And I would make this an error instead of a warning.

core/src/main/java/org/infinispan/interceptors/impl/TransactionalStoreInterceptor.java
ryanemerson: I think it's best if we just ignore the clear command, as `CacheWriterInterceptor` or `DistCacheWriterInterceptor` will still process the request.  The Cache api states "If the cache is transactional, it will not interact with the transaction.", so I think it is best if the request is still performed albeit outside of the Tx scope and we just log a warning to the user in this situation.

danberindei: This visitor isn't called when the user calls `cache.clear()` though, it's only called when the user commits a transaction. And since clear isn't interacting with transactions, a `ClearCommand` should never be added as a modification in a transaction.

ryanemerson: In the latest changes I have simply logged a debug message when the TransactionalStoreInterceptor's visitClearCommand is called (not the visitor). I don't have strong views on this though, so I can remove that if you think it is overkill.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
danberindei: On Wed, Jul 6, 2016 at 7:14 PM, Ryan Emerson <notifications@github.com>
wrote:

> In
> persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stores/AbstractJdbcStore.java
> <https://github.com/infinispan/infinispan/pull/4382#discussion_r69762003>:
>
> > +      try {
> > +         connection = getTxConnection(tx);
> > +         connection.rollback();
> > +      } catch (SQLException e) {
> > +         log.sqlFailureTxRollback(e);
> > +         throw new PersistenceException(String.format("Error during rollback of JDBC transaction (%s)", tx), e);
> > +      } finally {
> > +         destroyTxConnection(tx);
> > +      }
> > +   }
> > +
> > +   protected Connection getTxConnection(Transaction tx) {
> > +      Connection connection = transactionConnectionMap.get(tx);
> > +      if (connection == null) {
> > +         connection = connectionFactory.getConnection();
> > +         transactionConnectionMap.put(tx, connection);
>
> @danberindei <https://github.com/danberindei> I think such an approach
> could get complex fast when we consider non-jdbc stores. For example, if
> you have multiple store implementations that are transactional, for example
> JDBC based and some other Tx store, then it is necessary for multiple
> connection types to be managed. You could have a StoreTransaction
> interface/abstrct class with multiple implementations, however you then
> have the problem of how do you cater for multiple Tx stores.
>

I was thinking of keeping a list of StoreTransactions in the
CacheTransaction itself.


> Personally I think leaving the management of store transaction resources
> upto the store implementation allows for more flexibility, especially if
> people are writing their own custom stores.
>

If all stores end up implementing exactly the same thing, then flexibility
doesn't really add anything... OTOH maybe you're right, and my idea would
just make the core more complex, without actually simplifying the store.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/binary/JdbcBinaryStore.java
```diff
@@ -357,16 +308,111 @@ public void purge(Executor threadPool, PurgeListener task) {
...
+   private Set<Bucket> updateAndCreateBuckets(Collection<MarshalledEntry> modifiedEntries, Set<Object> keysToDelete,
+                                              Map<Integer, Bucket> existingBuckets) {
+      Set<Bucket> newBuckets = new HashSet<>();
+      for (MarshalledEntry entry : modifiedEntries) {
+         Integer bucketKey = getBuckedId(entry.getKey());
+         Object entryKey = entry.getKey();
+         InternalMetadata m = entry.getMetadata();
+         Bucket existingBucket = existingBuckets.get(bucketKey);
+
+         // How should we handle expired entries in a Tx?
```
danberindei: Hmmm, did you find the answer for your question, or should this be a TODO?

ryanemerson: Good spot! This should have been removed. The logic is the same as when we write for an expired entry outside of a tx, i.e. we remove the entry from the bucket if it has expired and continue to execute the rest of the Tx. I'll remove the comment and re-push

core/src/main/java/org/infinispan/interceptors/impl/TransactionalStoreInterceptor.java
danberindei: I'd remove the debug message from `TransactionalStoreInterceptor.visitClearCommand`, regular cache operations shouldn't log anything above trace unless something goes wrong. 

But if the visitor gets a `ClearCommand`, there must be a bug somewhere, so I'd throw an exception in `Updater.visitClearCommand`.

ryanemerson: Ok that makes sense, I have pushed the changes

slaskawi:
Looks good to me. @wburns @tristantarrant @danberindei do you have any comments or may I pull it?

slaskawi:
Pulling

slaskawi:
Integrated thanks @ryanemerson !

PR #4386, opened 2016-05-28
fix document, current wakeupInterval's default value is 60000.

kazuhira-r:
The default value of wakeupInterval seems to have been 5000 in Infinispan 4.x,
https://github.com/infinispan/infinispan/blob/4.2.x/core/src/main/java/org/infinispan/config/Configuration.java#L1730

I think present is 60000, but how about?
https://github.com/infinispan/infinispan/blob/9.0.0.Alpha2/core/src/main/java/org/infinispan/configuration/cache/ExpirationConfiguration.java#L16
https://github.com/infinispan/infinispan/blob/9.0.0.Alpha2/core/src/main/resources/schema/infinispan-config-9.0.xsd#L924

### DIFF
--- a/documentation/src/main/asciidoc/user_guide/eviction.adoc
+++ b/documentation/src/main/asciidoc/user_guide/eviction.adoc
@@ -96,7 +96,7 @@ Eviction is disabled by default.  If enabled (using an empty `<eviction />` elem
 
 
 * strategy: EvictionStrategy.NONE is assumed, if a strategy is not specified..
-* wakeupInterval: 5000 is used if not specified.
+* wakeupInterval: 60000 is used if not specified.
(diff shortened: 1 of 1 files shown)


danberindei:
@tristantarrant I wonder if we could link to the attribute in the generated XSD docs instead of copying the default value here.

slaskawi:
@danberindei I don't think so. Although Asciidoc supports [inserting code fragments](https://mrhaki.blogspot.com/2014/04/awesome-asciidoc-include-partial-parts.html) it requires using some special tags. I'm not sure if this is acceptable...

Anyway, the changes look fine to me, integrating

slaskawi:
Integrated, thanks @kazuhira-r !

PR #4390, opened 2016-05-31
ISPN-6739 Include generated META-INF/services in tools.jar

tristantarrant:
https://issues.jboss.org/browse/ISPN-6739


danberindei:
Integrated, thanks Tristan!

PR #4391, opened 2016-05-31
ISPN-6740 Client topologies not updated with persistent state

danberindei:
https://issues.jboss.org/browse/ISPN-6740

The HotRod client starts with topology id 0, so we initialize the
cache with topology id 1 to force a topology update on the first ping.

Please cherry-pick on 8.2.x as well.


pruivo:
pushing...

pruivo:

pruivo:
@danberindei, it needs a new PR for 8.2.x since it does not apply clearly

PR #4394, opened 2016-06-02
ISPN-6740 Client topologies not updated with persistent state

danberindei:
Backport of https://github.com/infinispan/infinispan/pull/4391 for 8.2.x
https://issues.jboss.org/browse/ISPN-6740

The HotRod client starts with topology id 0, so we initialize the
cache with topology id 1 to force a topology update on the first ping.


pruivo:

PR #4395, opened 2016-06-03
Get rid of feature and deprecation warnings when compiling Hotrod server

vjuranek:
https://issues.jboss.org/browse/ISPN-6689
https://issues.jboss.org/browse/ISPN-6754

### DIFF
--- a/server/hotrod/src/main/scala/org/infinispan/server/hotrod/AbstractTopologyAwareEncoder1x.scala
+++ b/server/hotrod/src/main/scala/org/infinispan/server/hotrod/AbstractTopologyAwareEncoder1x.scala
@@ -1,11 +1,15 @@
 package org.infinispan.server.hotrod
 
 import logging.Log
+import org.infinispan.commons.hash.Hash
+import org.infinispan.distribution.ch.impl.HashFunctionPartitioner
+import org.infinispan.distribution.group.PartitionerConsistentHash
+import org.infinispan.distribution.group.impl.GroupingPartitioner
 import org.infinispan.remoting.transport.Address
 import org.infinispan.server.core.transport.ExtendedByteBuf._
 import collection.JavaConversions._
 import org.infinispan.configuration.cache.Configuration
-import org.infinispan.distribution.ch.ConsistentHash
+import org.infinispan.distribution.ch.{KeyPartitioner, ConsistentHash}
 import io.netty.buffer.ByteBuf
 
 /**
@@ -20,8 +24,7 @@ abstract class AbstractTopologyAwareEncoder1x extends AbstractEncoder1x with Con
                                                   serverEndpointsMap: Map[Address, ServerAddress],
                                                   cfg: Configuration): AbstractHashDistAwareResponse = {
       HashDistAware11Response(lastViewId, serverEndpointsMap, cfg.clustering().hash().numOwners(),
-            DEFAULT_CONSISTENT_HASH_VERSION_1x, Integer.MAX_VALUE,
-            cfg.clustering().hash().numVirtualNodes())
+            DEFAULT_CONSISTENT_HASH_VERSION_1x, Integer.MAX_VALUE, 1)
    }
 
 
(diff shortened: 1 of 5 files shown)


server/hotrod/src/main/scala/org/infinispan/server/hotrod/AbstractTopologyAwareEncoder1x.scala
```diff
@@ -21,7 +25,7 @@ abstract class AbstractTopologyAwareEncoder1x extends AbstractEncoder1x with Con
                                                   cfg: Configuration): AbstractHashDistAwareResponse = {
       HashDistAware11Response(lastViewId, serverEndpointsMap, cfg.clustering().hash().numOwners(),
             DEFAULT_CONSISTENT_HASH_VERSION_1x, Integer.MAX_VALUE,
-            cfg.clustering().hash().numVirtualNodes())
+            cfg.clustering().hash().numSegments())
```
danberindei: `numSegments` doesn't work here, it has a completely different meaning from the now unused `numVirtualNodes`. I guess we should just remove `numVirtualNodes` in `HashConfiguration` and replace it with `1` here.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/OperationStatus.scala
```diff
@@ -33,7 +31,7 @@ object OperationStatus extends Enumeration {
 
    def withCompatibility(st: OperationStatus, isCompatibilityEnabled: Boolean): OperationStatus = {
       if (isCompatibilityEnabled) {
-         (st: @switch) match {
+         st match {
```
danberindei: Couldn't we make the `@switch` annotation work?

server/hotrod/src/main/scala/org/infinispan/server/hotrod/iteration/IterationManager.scala
```diff
@@ -17,6 +17,7 @@ import org.infinispan.server.hotrod.logging.Log
 
 import scala.collection.JavaConversions._
 import scala.collection.mutable
+import scala.language.existentials
```
danberindei: Is this necessary?

server/hotrod/src/main/scala/org/infinispan/server/hotrod/package.scala
```diff
@@ -6,6 +6,8 @@ import java.util.function.{BiConsumer => J8BiConsumer}
 
 import org.infinispan.remoting.transport.Address
 
+import scala.language.implicitConversions
```
danberindei: Is this necessary?

server/hotrod/src/main/scala/org/infinispan/server/hotrod/AbstractTopologyAwareEncoder1x.scala
vjuranek: I see - PR updated

server/hotrod/src/main/scala/org/infinispan/server/hotrod/OperationStatus.scala
vjuranek: AFAIK no if we use `enum` (which is IMHO what we want here), see e.g. [here](http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-3.html#jvms-3.10):

> The Java Virtual Machine's tableswitch and lookupswitch instructions operate only on int data. Because operations on byte, char, or short values are internally promoted to int, a switch whose expression evaluates to one of those types is compiled as though it evaluated to type int.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/iteration/IterationManager.scala
vjuranek: not necessary, but it says that existential type is used intentionally here and thus removes the feature warning during compilation.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/package.scala
vjuranek: same as above - says that `implicit` conversion functions are used intentionally

galderz:
@danberindei What changes are required here?

danberindei:
@galderz None, but I think @vjuranek can't change the label. Pulling...

danberindei:
Integrated, thanks Vojtech!

PR #4396, opened 2016-06-03
ISPN-6755: an RFC

ksobolew:
See the JIRA entry. This is what I'm going to use in the interim.

### DIFF
--- a/core/src/main/java/org/infinispan/functional/impl/MetaParams.java
+++ b/core/src/main/java/org/infinispan/functional/impl/MetaParams.java
@@ -143,15 +143,15 @@ public final class MetaParams {
     * @param metas Meta parameters to create the collection with
     * @return a collection of meta parameters without type duplicates
     */
-   static MetaParams of(MetaParam... metas) {
+   public static MetaParams of(MetaParam... metas) {
       return new MetaParams(filterDuplicates(metas));
    }
 
-   static MetaParams of(MetaParam meta) {
+   public static MetaParams of(MetaParam meta) {
       return new MetaParams(new MetaParam[]{meta});
    }
 
-   static MetaParams empty() {
+   public static MetaParams empty() {
       return new MetaParams(new MetaParam[]{});
    }
 
--- a/core/src/main/java/org/infinispan/functional/impl/MetaParamsInternalMetadata.java
+++ b/core/src/main/java/org/infinispan/functional/impl/MetaParamsInternalMetadata.java
@@ -9,7 +9,6 @@ import org.infinispan.commons.util.Util;
 import org.infinispan.container.versioning.EntryVersion;
 import org.infinispan.marshall.core.Ids;
 import org.infinispan.metadata.InternalMetadata;
-import org.infinispan.metadata.Metadata;
 
 import java.io.IOException;
 import java.io.ObjectInput;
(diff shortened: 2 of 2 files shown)


galderz:
@ksobolew Why do you need to make MetaParams publicly instantiable? This class is an internal class, kept inside the `.impl` folder, so we don't expect users to use it. Sure, the class is defined as public, but that's mostly so that we can register the externalizer which is done in another package.

Also, why do you need an InternalMetadata returned out of `from`?

galderz:
Sorry, just noticed a longer explanation is in the JIRA. Going through that.

ksobolew:
@galderz Because of CacheLoaders :) As I understand it the usual way is to take the Metadata from an entry and serialize it without any questions asked. Then the CacheLoader deserializes it and gives it back as it was. But I have several CacheLoaders that do not have any place to store the serialized form of Metadata. In fact, this was the original point of CacheLoaders (in JBCache) that they take some other data source and convert it into the internal representation - because JBCache was supposed to be a *cache* for some other resource. It works just the same if we treat Infinispan as a data store as opposed to a cache - but instead of storing the entries explicitly I let the CacheLoader create them as needed. But for that to work I need some way to communicate metadata like lifespan. And it has to be in a correct format so that the internal machinery recogizes it. Usually it's an InternalMetadataImpl; but the Functional API creates MetaParamsInternalMetadata and ignores anything that is not it (yes, there is a big TODO about compatibility with other formats). So, in conclusion, I need(ed) a way to return MetaParamsInternalMetadata from CacheLoader.load() :)

Maybe it would be beneficial to clarify the interface between Infinispan-core and CacheLoaders with respect to the Metadata object?

But, further investigation showed that I can actually use the EmbeddedMetadata. This way there's no additional Metadata object to worry about all this is moot... provided that one bug is fixed (I'll still have to report it though).

ksobolew:
@galderz Yes, there is a longer explanation in JIRA, but I wrote it up again. In part so that I can reassure myself that I still understand the issue ;) I also provided some futher thoughts about alternatives.

galderz:
@ksobolew Ok, I understand what you're trying to do and it makes sense, but as you hint at the bottom, there's no obligation for you to return a `MetaParamsInternalMetadata`. You can indeed return any Metadata implementation as long as it reflects the information that you want in it, e.g. EmbeddedMetadata. Shall we put this PR on hold until we see about that other bug that you mention?

ksobolew:
@galderz I think it could even be closed (along with the JIRA). I posted it without fully understanding what's going on. And I guess things like the interface between Infinispan-core and CacheLoaders with respect to the Metadata object should be discussed on the -devel list, not in a PR? :)

In addition to the bug I mentioned there is the fact that the Functional API is incomplete - there are some pieces mssing, like compatibility with other kinds on internal metadata (put()-with-lifespan is explicitly mentioned), or incomplete support for updates of metadata (ISPN-6756).

ksobolew:
@galderz I reported the bug as ISPN-6773

galderz:
We discuss stuff in -devel when it's a complex issue or requires an API change. So far I don't see the need to do that :)

Right, the API is not fully completed, that's why it's marked `@Experimental` 😃 

Are you working on fixes for ISPN-6733 and/or ISPN-6756?

galderz:
Oh, just saw PR for ISPN-6733! I'll have a look tomorrow

PR #4398, opened 2016-06-07
ISPN-6748: Deprecate Singleton Mode for Persistent Stores.

ryanemerson:
https://issues.jboss.org/browse/ISPN-6748

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/cache/SingletonStoreConfigurationBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/cache/SingletonStoreConfigurationBuilder.java
@@ -11,6 +11,11 @@ import org.infinispan.commons.configuration.Builder;
 import org.infinispan.commons.configuration.attributes.AttributeSet;
 import org.infinispan.configuration.global.GlobalConfiguration;
 
+/**
+ * @deprecated Singleton writers will be removed in 10.0. If it is desirable that all nodes don't write to the underlying store
+ * then a shared store should be used instead, as this only performs store writes at a key's primary owner.
+ */
+@Deprecated
 public class SingletonStoreConfigurationBuilder<S> extends AbstractStoreConfigurationChildBuilder<S> implements Builder<SingletonStoreConfiguration> {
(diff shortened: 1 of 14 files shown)


galderz:
@ryanemerson 46 new failures, they look related to the changes, can you please have a look? Thx :)

ryanemerson:
@galderz I will take another look at this when I return from PTO.

ryanemerson:
@galderz I have fixed the problems from before. The current test failures don't seem related to my changes (Hotrod).

slaskawi:
There are still some tests failing: http://ci.infinispan.org/viewLog.html?buildId=40575&buildTypeId=bt9&guest=1

Retriggering CI.

core/src/main/java/org/infinispan/configuration/cache/SingletonStoreConfigurationBuilder.java
```diff
@@ -11,6 +11,11 @@
 import org.infinispan.commons.configuration.attributes.AttributeSet;
 import org.infinispan.configuration.global.GlobalConfiguration;
 
+/**
+ * @deprecated Singleton writers will be removed in 9.1. If it is desirable that all nodes don't write to the underlying store
```
slaskawi: If we deprecate them in 9.0, we probably won't be able to remove them until 10... Am I right @tristantarrant ?

slaskawi:
The code looks fine (let's have another look at those CI failures as soon as TeamCity rebuilds this PR). The only concern that I have is removing it in `9.1`. I think we should do it in `10` (but maybe there is a good argument for that).

tristantarrant:
I agree with @slaskawi , removal won't happen until 10

ryanemerson:
@slaskawi @tristantarrant Version updated to 10.0. Can we get this in before the release tomorrow?

ryanemerson:
@slaskawi @tristantarrant The issues with the tests were resolved with a pull --rebase. This should be good to merge.

danberindei:
Integrated, thanks Ryan!

PR #4399, opened 2016-06-07
ISPN-6760 Comparing a sum of an attribute fails with ClassCastException

anistor:
https://issues.jboss.org/browse/ISPN-6760


galderz:
Integrated, thanks @anistor

PR #4404, opened 2016-06-09
ISPN-6422 Use a synchronized list to store events

tristantarrant:
https://issues.jboss.org/browse/ISPN-6422

### DIFF
--- a/server/integration/event-logger/src/main/java/org/infinispan/server/eventlogger/ServerEventLogger.java
+++ b/server/integration/event-logger/src/main/java/org/infinispan/server/eventlogger/ServerEventLogger.java
@@ -95,7 +95,7 @@ public class ServerEventLogger implements EventLogger {
 
    @Override
    public List<EventLog> getEvents(Instant start, int count, Optional<EventLogCategory> category, Optional<EventLogLevel> level) {
-      List<EventLog> events = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


anistor:
Looking...

server/integration/event-logger/src/main/java/org/infinispan/server/eventlogger/ServerEventLogger.java
```diff
@@ -95,7 +95,7 @@ TimeService getTimeService() {
 
    @Override
    public List<EventLog> getEvents(Instant start, int count, Optional<EventLogCategory> category, Optional<EventLogLevel> level) {
-      List<EventLog> events = new ArrayList<>();
+      List<EventLog> events = Collections.synchronizedList(new ArrayList<>());
       AtomicReference<Throwable> throwable = new AtomicReference<>();
       try {
          cacheManager.executor().submitConsumer(m -> {
```
anistor: The throwable is wrapped in a CacheException and thrown. The outer try-catch(Exception) will wrap it again. I think we should have a separate catch for CacheException and Exception and avoid the double wrapping.

anistor:
I made one comment, but otherwise looks good.

tristantarrant:
Done @anistor

anistor:
Integrated in master. Thanks @tristantarrant !

anistor:
Backported to 8.2.x too.

PR #4412, opened 2016-06-13
ISPN-6773: treat non-MetadataAware InternalCacheEntries as not needin…

ksobolew:
…g the entire Metadata object to be stored.

https://issues.jboss.org/browse/ISPN-6773

* Make CacheEntry *not* implement MetadataAware
  - Now only the Metadata*CacheEntry classes (like MetadataMortalCacheEntry)
    implement this interface, while non-Metadata* classes (like the plain
    MortalCacheEntry) do not.
  - The getMetadata() and setMetadata() method need to be specified in
    CacheEntry interface, as they used to be inherited from MetadataAware.
* Make InternalEntryFactoryImpl.isStoreMetadata() check the type of the
  InternalCacheEntry in addition to the Metadata object and return false if it
  does not implement MetadataAware.
* If the entry is new, pass null as the InternalCacheEntry and test only the
  Metadata object.

This way the correct code path (fot Metadata* or non-Metadata*) in
EntryWrappingInterceptor is used even if the Functional API replaces the
Metadata object inside the InternalCacheEntry.

### DIFF
--- a/core/src/main/java/org/infinispan/container/entries/metadata/MetadataAware.java
+++ b/core/src/main/java/org/infinispan/container/entries/metadata/MetadataAware.java
@@ -3,7 +3,7 @@ package org.infinispan.container.entries.metadata;
 import org.infinispan.metadata.Metadata;
 
 /**
- * Metdata aware cache entry.
+ * Marker interface for metadata aware cache entry.
  *
  * @author Galder Zamarreño
  * @since 5.3
--- a/core/src/main/java/org/infinispan/container/InternalEntryFactoryImpl.java
+++ b/core/src/main/java/org/infinispan/container/InternalEntryFactoryImpl.java
@@ -2,6 +2,7 @@ package org.infinispan.container;
 
 import org.infinispan.container.entries.*;
 import org.infinispan.container.entries.metadata.L1MetadataInternalCacheEntry;
+import org.infinispan.container.entries.metadata.MetadataAware;
 import org.infinispan.metadata.EmbeddedMetadata;
 import org.infinispan.metadata.Metadata;
 import org.infinispan.container.entries.metadata.MetadataImmortalCacheEntry;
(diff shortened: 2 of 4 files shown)


galderz:
I'm sorry but we can't accept the change as is. CacheEntry is a public API that we expose and your change could cause issues to existing applications. I agree the interface does not make as much sense as it should (see [ISPN-3460](https://issues.jboss.org/browse/ISPN-3460)) but we can just change it like that. So, I'd suggest you find an alternative way to do what you want to do. I can certainly help you find an alternative way, but you should post a test that demonstrates why you need this change.

ksobolew:
@galderz see 75d7d85f8803b7ba7e69cfef89db9dbffd6c24f2 for the test case. In my case the EmbeddedMetadata object is created by the CacheLoader, not using the put-with-metadata API, but the result is the same.

galderz:
@ksobolew I've sent a [PR](https://github.com/ksobolew/infinispan/pull/1) to your Infinispan repo to workaround this. Have a look and see what you think. The code in `InternalEntryFactoryImpl` is already making similar assumptions, so I think it should be fine.

ksobolew:
@galderz Well, it looks like a workaround all right :) But if it's OK for you then it's OK for me

galderz:
@ksobolew Seems like the PR needs rebasing, can you look into that?

ksobolew:
@galderz @slaskawi Rebased.

slaskawi:
Looking

slaskawi:
Integrated thanks!

core/src/main/java/org/infinispan/container/entries/metadata/MetadataAware.java
```diff
@@ -3,7 +3,7 @@
 import org.infinispan.metadata.Metadata;
 
 /**
- * Metdata aware cache entry.
+ * Marker interface for metadata aware cache entry.
```
danberindei: It's no longer a marker interface :)

ksobolew: @danberindei well, yeah, I planned on doing a cleanup (there's also an orphaned import somewhere there) while rebasing but forgot about it :)

PR #4413, opened 2016-06-13
ISPN-6773 (backport to 8.2.x): treat non-MetadataAware InternalCacheEntries as not needin…

ksobolew:
…g the entire Metadata object to be stored.

* Make CacheEntry *not* implement MetadataAware
  - Now only the Metadata*CacheEntry classes (like MetadataMortalCacheEntry)
    implement this interface, while non-Metadata* classes (like the plain
    MortalCacheEntry) do not.
  - The getMetadata() and setMetadata() method need to be specified in
    CacheEntry interface, as they used to be inherited from MetadataAware.
* Make InternalEntryFactoryImpl.isStoreMetadata() check the type of the
  InternalCacheEntry in addition to the Metadata object and return false if it
  does not implement MetadataAware.
* If the entry is new, pass null as the InternalCacheEntry and test only the
  Metadata object.

This way the correct code path (fot Metadata* or non-Metadata*) in
EntryWrappingInterceptor is used even if the Functional API replaces the
Metadata object inside the InternalCacheEntry.

### DIFF
--- a/core/src/main/java/org/infinispan/container/InternalEntryFactoryImpl.java
+++ b/core/src/main/java/org/infinispan/container/InternalEntryFactoryImpl.java
@@ -2,6 +2,7 @@ package org.infinispan.container;
 
 import org.infinispan.container.entries.*;
 import org.infinispan.container.entries.metadata.L1MetadataInternalCacheEntry;
+import org.infinispan.container.entries.metadata.MetadataAware;
 import org.infinispan.metadata.EmbeddedMetadata;
 import org.infinispan.metadata.Metadata;
 import org.infinispan.container.entries.metadata.MetadataImmortalCacheEntry;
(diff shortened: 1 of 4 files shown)


galderz:
Hmmm, there's 11 new failures in the [testsuite](http://ci.infinispan.org/viewLog.html?buildId=40336&buildTypeId=bt9&guest=1), also, CacheEntry is public interface that we expose via Cache, so this change would require a discussion. 

Can't you find an alternative solution? Do you have a test for this functionality that shows this problem? If you have such test I can look into it.

ksobolew:
@galderz In #4412, which is the same but based on master, all the tests pass. Regarding the API change, it looks like that CacheEntry was made to implement MetadataAware not because of the is-a relationship but to just inherit the {get,set}Metadata() methods. When we have MortalCacheEntry and MetadataMortalCacheEntry it doesn't make much sense for both of them to implement MetadataAware. That was my reasoning. And you're right, I should add a test case :)

ksobolew:
@galderz I pushed (and to #4412 as well) some test cases, of which one is relevant to this bug and the rest is convenient to have :)

galderz:
@ksobolew Ah yes, this is 8.2.x. Let me have a closer look at master and then I can comment further there.

slaskawi:
Integrated Thanks!

PR #4414, opened 2016-06-14
ISPN-6494 Re-add the DONT_BUNDLE flag

danberindei:
We're still investigating transfer-queue vs sender-sends-with-timer, and it's better to have DONT_BUNDLE for testing because transfer-queue ignores it.

Please keep the issue open.


pruivo:
testsuite looks good locally. pushing...

pruivo:

PR #4416, opened 2016-06-16
ISPN-6789 EnwtryWrappingInterceptor computes location twice on backups

danberindei:
https://issues.jboss.org/browse/ISPN-6789


pruivo:

PR #4418, opened 2016-06-17
ISPN-6790 Distribution interceptors re-compute key location after remote get

danberindei:
https://issues.jboss.org/browse/ISPN-6790


tristantarrant:
Tested it locally. Pushed to master. Thanks

danberindei:
@tristantarrant Added one more commit. The initial change was actually more expensive for replicated caches, which don't need to compute the key's hash. Or should I open a new PR?

tristantarrant:
Pushed the additional commit. Thanks

PR #4422, opened 2016-06-22
JDK9

tristantarrant:
Initial batch of changes to get most embedded stuff to work with JDK 9.
Server is still broken because of issues with Scala and WildFly

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -57,7 +57,7 @@
 
       <version.java>1.8</version.java>
 
-      <version.maven-compiler-plugin>3.1</version.maven-compiler-plugin>
+      <version.maven-compiler-plugin>3.5.1</version.maven-compiler-plugin>
 
       <version.protostream>3.0.5.Final</version.protostream>
       <version.aesh>0.66.4</version.aesh>
@@ -70,7 +70,7 @@
       <version.hibernate.search>5.6.0.Beta1</version.hibernate.search>
       <version.javax.cache>1.0.0</version.javax.cache>
       <version.cdi>1.2</version.cdi>
-      <version.jboss.marshalling>1.4.10.Final</version.jboss.marshalling>
+      <version.jboss.marshalling>1.4.11.Final</version.jboss.marshalling>
       <version.jboss.logging>3.3.0.Final</version.jboss.logging>
       <version.jgroups>3.6.9.Final</version.jgroups>
       <version.jta>1.0.1.Final</version.jta>
@@ -646,21 +646,6 @@
                                     <versionRange>[1.0-beta-1,)</versionRange>
                                     <goals>
                                        <goal>javac2</goal>
-                                    </goals>
-                                 </pluginExecutionFilter>
-                                 <action>
-                                    <ignore />
-                                 </action>
-                              </pluginExecution>
-                              <pluginExecution>
-                                 <pluginExecutionFilter>
-                                    <groupId>org.jboss.maven.plugins</groupId>
-                                    <artifactId>
-                                       maven-injection-plugin
-                                    </artifactId>
-                                    <versionRange>[1.0.2,)</versionRange>
-                                    <goals>
-                                       <goal>bytecode</goal>
                                     </goals>
                                  </pluginExecutionFilter>
                                  <action>
--- a/core/src/main/java/org/infinispan/Version.java
+++ b/core/src/main/java/org/infinispan/Version.java
@@ -3,7 +3,9 @@ package org.infinispan;
 import net.jcip.annotations.Immutable;
 
 import java.io.ByteArrayOutputStream;
+import java.io.InputStream;
 import java.util.Arrays;
+import java.util.Properties;
 
 /**
  * Contains version information about this release of Infinispan.
@@ -24,6 +26,8 @@ public class Version {
 
    private static final Version INSTANCE = new Version();
 
+   private final String version;
+   private final String codename;
    private final byte[] versionId;
    private final String moduleSlot;
    private final short versionShort;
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/infinispan/Version.java
```diff
@@ -24,25 +26,39 @@
...
    private final short versionShort;
    private final short marshallVersion;
    private final String majorMinor;
    private final String major;
 
+
    private Version() {
-      String parts[] = getParts(Injected.getVersion());
+      Properties properties = new Properties();
+      try (InputStream is = Version.class.getResourceAsStream("/META-INF/infinispan-version.properties")) {
```
Sanne: this approach would expose you to a devilish trick: in case of duplicate Infinispan jars on the classpath you might be logging version Y but executing version X. I'm not against the approach, but it's a risk to be aware of..

parent/pom.xml
```diff
@@ -1603,6 +1601,30 @@
             </plugins>
          </build>
       </profile>
+      <profile>
+         <id>jigsaw</id>
+         <activation>
+            <jdk>[1.9,)</jdk>
```
Sanne: are you sure this activates?

In my experience, I have to use "9" as format, not "1.9".

Sanne:
If you extract a clean patch which purely aims at "getting rid of maven-injection-plugin" we can start integrating the good parts.

core/src/main/java/org/infinispan/Version.java
tristantarrant: I guess the "safe" way is to ensure that we are getting the resource from the same root path as from where the Version class was loaded. Or wait for javassist to fix their issue.

parent/pom.xml
tristantarrant: 9 > 1.9 anyway :)

core/src/main/java/org/infinispan/Version.java
Sanne: BTW this is my approach using standard JDK API:
 - https://github.com/hibernate/hibernate-ogm/pull/708/commits/dc4cfeade80e16520f04d4db3f34461d0255fd82

Doesn't get you all the strange formats you want though ;) Maybe get rid of some?

bom/pom.xml
```diff
@@ -70,14 +70,14 @@
...
-      <version.jboss.marshalling>1.4.10.Final</version.jboss.marshalling>
+      <version.jboss.marshalling>1.4.11.Final</version.jboss.marshalling>
       <version.jboss.logging>3.3.0.Final</version.jboss.logging>
       <version.jgroups>3.6.9.Final</version.jgroups>
       <version.jta>1.0.1.Final</version.jta>
       <version.netty>4.1.0.CR7</version.netty>
       <version.osgi>4.3.1</version.osgi>
       <version.scala.major>2.11</version.scala.major>
-      <version.scala>${version.scala.major}.7</version.scala>
+      <version.scala>${version.scala.major}.8</version.scala>
```
tristantarrant: This should be reverted as it was a blind try and it doesn't actually work

parent/pom.xml
```diff
@@ -177,10 +177,10 @@
       <version.org.wildfly.build-tools>1.1.6.Final</version.org.wildfly.build-tools>
       <version.org.wildfly.maven-plugin>1.1.0.Alpha5</version.org.wildfly.maven-plugin>
 
-      <version.javassist>3.18.1-GA</version.javassist>
+      <version.javassist>3.20.0-GA</version.javassist>
```
tristantarrant: Again, this doesn't actually fix anything

core/src/main/java/org/infinispan/Version.java
tristantarrant: That was indeed an avenue I wanted to explore, but:
- the manifest is generated by the maven archive / bundle plugin only during the package goal (you could create a temporary manifest just for the sake of these)
- the testsuite relies on having the version available for a number of things (marshalling, configuration, etc)

Sanne: Yea I realize it's not as simple, you'll need to make some choices. The codename could be committed in the sources?

tristantarrant:
Why should I extract a clean patch just for that ?

Sanne:
> Why should I extract a clean patch just for that ?

No strong reason.. I personally just prefer to break such larger things into PRs which can be integrated independently, as clearly how to read the versions & project codenames might cause some feedback and suggestions from PR review.

Sanne:
Just stumbled upon this issue I had created some time ago:
 - https://issues.jboss.org/browse/ISPN-4149

assigned it to you now.

bom/pom.xml
```diff
@@ -654,21 +654,6 @@
                               </pluginExecution>
                               <pluginExecution>
                                  <pluginExecutionFilter>
-                                    <groupId>org.jboss.maven.plugins</groupId>
-                                    <artifactId>
-                                       maven-injection-plugin
-                                    </artifactId>
-                                    <versionRange>[1.0.2,)</versionRange>
-                                    <goals>
```
slaskawi: Why are we erasing this?

slaskawi: Please ignore me... I've just found why...

core/src/test/java/org/infinispan/marshall/multiversion/MultiPojoVersionMarshallTest.java
```diff
@@ -43,7 +43,10 @@
  * @author Galder Zamarreño
  * @since 5.0
  */
-@Test(groups = "functional", testName = "marshall.multiversion.MultiPojoVersionMarshallTest")
+/*
+ * Disabled the test since Javassist fails with JDK 9Q
+ */
+@Test(groups = "functional", testName = "marshall.multiversion.MultiPojoVersionMarshallTest", enabled = false)
```
slaskawi: Please add description which JIRA will re-enable the test.

tristantarrant:
Closing to reopen a proper PR

PR #4425, opened 2016-06-27
ISPN-3905 Murmurhash3 implementation is slow on String keys

danberindei:
https://issues.jboss.org/browse/ISPN-3905

Compute String hashes without allocation. Can probably be optimized further...

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/hash/MurmurHash3.java
+++ b/commons/src/main/java/org/infinispan/commons/hash/MurmurHash3.java
@@ -432,28 +432,28 @@ public class MurmurHash3 implements Hash {
          }
 
          if (cp <= 0x7f) {
-            addByte(state, (byte) cp, ++byteLen);
+            addByte(state, (byte) cp, byteLen++);
          } else if (cp <= 0x07ff) {
             byte b1 = (byte) (0xc0 | (0x1f & (cp >> 6)));
             byte b2 = (byte) (0x80 | (0x3f & cp));
-            addByte(state, b1, ++byteLen);
-            addByte(state, b2, ++byteLen);
+            addByte(state, b1, byteLen++);
+            addByte(state, b2, byteLen++);
          } else if (cp <= 0xffff) {
             byte b1 = (byte) (0xe0 | (0x0f & (cp >> 12)));
             byte b2 = (byte) (0x80 | (0x3f & (cp >> 6)));
             byte b3 = (byte) (0x80 | (0x3f & cp));
-            addByte(state, b1, ++byteLen);
-            addByte(state, b2, ++byteLen);
-            addByte(state, b3, ++byteLen);
+            addByte(state, b1, byteLen++);
+            addByte(state, b2, byteLen++);
+            addByte(state, b3, byteLen++);
          } else {
             byte b1 = (byte) (0xf0 | (0x07 & (cp >> 18)));
             byte b2 = (byte) (0x80 | (0x3f & (cp >> 12)));
             byte b3 = (byte) (0x80 | (0x3f & (cp >> 6)));
             byte b4 = (byte) (0x80 | (0x3f & cp));
-            addByte(state, b1, ++byteLen);
-            addByte(state, b2, ++byteLen);
-            addByte(state, b3, ++byteLen);
-            addByte(state, b4, ++byteLen);
+            addByte(state, b1, byteLen++);
+            addByte(state, b2, byteLen++);
+            addByte(state, b3, byteLen++);
+            addByte(state, b4, byteLen++);
          }
       }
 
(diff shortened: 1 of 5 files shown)


slaskawi:
Do you have any performance results @danberindei ? It would be very nice to post them here :)

danberindei:
@slaskawi The initial commit was actually slower than the allocating version, even without the extra allocation. It's faster with the second commit (MH3B):

```
1 thread
Benchmark                (initialFillRatio)  (keySize)   Mode  Cnt        Score       Error  Units
MH3Benchmark.testMH3A                   0.5         20  thrpt   60  4541775.939 ± 39804.032  ops/s
MH3Benchmark.testMH3B                   0.5         20  thrpt   60  9973051.000 ± 83607.674  ops/s
MH3Benchmark.testMH3Old                 0.5         20  thrpt   60  7263881.351 ± 74957.407  ops/s

10 threads
Benchmark                (initialFillRatio)  (keySize)   Mode  Cnt         Score        Error  Units
MH3Benchmark.testMH3A                   0.5         20  thrpt   60  17630083.421 ± 149203.536  ops/s
MH3Benchmark.testMH3B                   0.5         20  thrpt   60  43775366.597 ± 688718.248  ops/s
MH3Benchmark.testMH3Old                 0.5         20  thrpt   60  27210915.599 ± 500515.798  ops/s
```

slaskawi:
Looks good to me... It's almost 2 times faster than the baseline with 10 threads!

Let's give @rvansa 1-2 days to look into... After that I'm integrating this PR.

commons/src/main/java/org/infinispan/commons/hash/MurmurHash3.java
```diff
@@ -432,28 +432,28 @@ private long MurmurHash3_x64_64_String(String s, long seed) {
          }
 
          if (cp <= 0x7f) {
-            addByte(state, (byte) cp, ++byteLen);
+            addByte(state, (byte) cp, byteLen++);
```
rvansa: This looks like changing the result - how could the test pass (produce equal results) with the first commit only?

rvansa:
IMO this is a step forward (can't see any issue if it computes the hash correctly - , though I think that for strings, we should use plain String.hashCode(), which is already cached.

wburns:
@rvansa I think what @danberindei is trying to avoid is having to have a brand new hash implementation (different results).  This method should maintain backwards compatibility so we don't have to worry about changing the clients and they can use the existing method.  If we change the results we would have to have another hash implementation that would have to be negotiated when a client connects.

commons/src/main/java/org/infinispan/commons/hash/MurmurHash3.java
wburns: I would agree, maybe we should just squash the commits @danberindei ?

wburns:
@danberindei Nice, I am glad you took it another step forward to reduce array allocations as well :)

commons/src/main/java/org/infinispan/commons/hash/MurmurHash3.java
rvansa: No, I mean - I suppose the test passed with commit 1, and it passes with commit 2 while this seems that these should produce different hashcodes. So is there an issue with the tests or is my assumption incorrect?
In the end we could squash them, though.

pruivo:
**off-topic**
why don't use the ```hashCode()``` and remove the MurmurHash? 
For custom objects, murmur hashes the object's hash code and for arrays it hashes the content (the same as ```Arrays.hashCode()```) and for strings it contents. 
So, what are the advantage of Murmur?

danberindei:
@pruivo The default `hashCode()` implementation isn't spreading the hashes across the entire int range. Even `HashMap` needs a spread function, although it's been simplified a lot in Java 8, since they started using tree buckets. But we need a far better spreader, because of the way our consistent hashes compute a key's segment, discarding the low bits.

Tristan told me we are changing the compatibility mode implementation soon, and routing will be based only on serialized keys. So we can try to change the hash implementation for strings then, but we still don't have a good way to change the hash implementation for `byte[]`.

pruivo:
@danberindei true about the ```HashMap```, but they explain why:
```
Because the table uses power-of-two masking, sets of hashes that vary only in bits above the current mask will always collide.
```
we use the hash to find the segment of the key (by dividing by the number of segments), so I'm not sure if we have a problem or if hashing actually fix anything.

In any case, the PR looks good.

slaskawi:
Pulling

slaskawi:
Integrated thanks!

danberindei:
@pruivo In our case it's the other way around: because we use division, any hash that's below `Integer.MAX_VALUE / numSegments` maps to segment 0. If we used `String.hashCode()` directly, that would mean all short strings map to segment 0.

PR #4431, opened 2016-06-29
ISPN-6681: Replace C3P0 with HikariCP.

ryanemerson:
https://issues.jboss.org/browse/ISPN-6681

### DIFF
--- a/all/embedded/pom.xml
+++ b/all/embedded/pom.xml
@@ -308,6 +308,8 @@
                         </_exportcontents>
                         <Import-Package>
                            !com.mchange.v2.c3p0,
+                           !com.zaxxer.hikari,
+                           !com.codahale.metrics.*,
                            !com.sun.jmx.mbeanserver,
                            !infinispan.*,
                            !javax.cache.*,
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -68,6 +68,7 @@
       <version.hibernate.javax.persistence>1.0.0.Final</version.hibernate.javax.persistence>
       <version.hibernate.core>5.0.9.Final</version.hibernate.core>
       <version.hibernate.search>5.6.0.Beta1</version.hibernate.search>
+      <version.hikaricp>2.4.6</version.hikaricp>
       <version.javax.cache>1.0.0</version.javax.cache>
       <version.cdi>1.2</version.cdi>
       <version.jboss.marshalling>1.4.11.Final</version.jboss.marshalling>
@@ -79,6 +80,7 @@
       <version.scala.major>2.11</version.scala.major>
       <version.scala>${version.scala.major}.7</version.scala>
       <version.scala.xml>1.0.2</version.scala.xml>
+      <version.slf4j-jboss-logging>1.1.0.Final</version.slf4j-jboss-logging>
       <version.xstream>1.4.9</version.xstream>
       <version.c3p0_dep.mchange-commons-java>0.2.7</version.c3p0_dep.mchange-commons-java>
       <version.hibernate.entitymanager>${version.hibernate.core}</version.hibernate.entitymanager>
@@ -469,6 +471,18 @@
             <groupId>com.mchange</groupId>
             <artifactId>c3p0</artifactId>
             <version>${version.c3p0}</version>
+         </dependency>
+
+         <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <version>${version.hikaricp}</version>
+         </dependency>
+
+         <dependency>
+            <groupId>org.jboss.slf4j</groupId>
+            <artifactId>slf4j-jboss-logging</artifactId>
+            <version>${version.slf4j-jboss-logging}</version>
          </dependency>
       </dependencies>
    </dependencyManagement>
--- a/javadoc/pom.xml
+++ b/javadoc/pom.xml
@@ -234,6 +234,11 @@
                            <artifactId>c3p0</artifactId>
                            <version>0.9.5-pre8</version>
                         </additionalDependency>
+                        <additionalDependency>
+                           <groupId>com.zaxxer</groupId>
+                           <artifactId>HikariCP</artifactId>
+                           <version>${version.hikaricp}</version>
+                        </additionalDependency>
                      </additionalDependencies>
                   </configuration>
                </execution>
--- a/persistence/jdbc/pom.xml
+++ b/persistence/jdbc/pom.xml
@@ -33,6 +33,16 @@
       </dependency>
 
       <dependency>
+         <groupId>com.zaxxer</groupId>
+         <artifactId>HikariCP</artifactId>
+      </dependency>
+
+      <dependency>
+         <groupId>org.jboss.slf4j</groupId>
+         <artifactId>slf4j-jboss-logging</artifactId>
+      </dependency>
+
+      <dependency>
          <groupId>com.h2database</groupId>
          <artifactId>h2</artifactId>
          <scope>test</scope>
--- a/persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
+++ b/persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
@@ -1,12 +1,5 @@
 package org.infinispan.persistence.jdbc.connectionfactory;
 
-import java.net.URL;
-import java.sql.Connection;
-import java.sql.SQLException;
-import java.util.Properties;
-
-import org.infinispan.commons.util.FileLookup;
-import org.infinispan.commons.util.FileLookupFactory;
 import org.infinispan.persistence.jdbc.JdbcUtil;
 import org.infinispan.persistence.jdbc.configuration.ConnectionFactoryConfiguration;
 import org.infinispan.persistence.jdbc.configuration.PooledConnectionFactoryConfiguration;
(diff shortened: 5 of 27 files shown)


rvansa:
+1, I've ran some benchmarks in the past and Hikari always outperformed C3P0.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
```diff
@@ -14,86 +25,162 @@
...
+ * @author Ryan Emerson
  */
 public class PooledConnectionFactory extends ConnectionFactory {
 
-   private static final Log log = LogFactory.getLog(PooledConnectionFactory.class, Log.class);
-   private static final boolean trace = log.isTraceEnabled();
-   private ComboPooledDataSource pooledDataSource;
+   private static final Log LOG = LogFactory.getLog(PooledConnectionFactory.class, Log.class);
+   private static final boolean TRACE = LOG.isTraceEnabled();
+   private static final String FORCE_C3P0 = "c3p0.force";
```
rvansa: Wouldn't it be better to have explicit property like `infinispan.connectionpool=c3p0` (or `=hikari`)? If we think about more alternatives in the future as well...

ryanemerson: @rvansa We could go down that route, however the reason I went with "c3p0.force" is that the original intention was for c3p0 to be removed in a subsequent release. I guess this comes down to whether we want the flexibility/overhead of maintaining support for 2 (or more) connection pools in the future, or whether a single connection pool is sufficient.

tristantarrant: I agree with @ryanemerson: in the case of the PooledConnectionFactory, the actual pool should be an implementation detail, so I don't think we should expose too much about it.

tristantarrant: And for this reason I'd even drop c3p0 completely from 9.0 with no "force" flag either.

ryanemerson: @tristantarrant The reason I didn't drop c3p0 completely, is that by moving from c3p0 to Hikari, we are essentially changing the api because the two connection pools utilise different properties files. So if removed c3p0 without deprecation a user migrating from 8.2.x to 9.0 would need to transition to hikari.properties without any prior warning. Of course, if a user does not utilise a .properties file to provided advanced configuration this is a non issue.

rvansa: @ryanemerson then the PR should contain a note to the Upgrading guide.

ryanemerson: @rvansa @tristantarrant docs updated.

all/embedded/pom.xml
```diff
@@ -138,6 +138,23 @@
...
+         <artifactId>HikariCP</artifactId>
+         <optional>${uberjar.deps.optional}</optional>
+      </dependency>
+      <dependency>
+         <groupId>org.jboss.slf4j</groupId>
+         <artifactId>slf4j-jboss-logging</artifactId>
+         <optional>${uberjar.deps.optional}</optional>
+      </dependency>
+      <dependency>
+         <groupId>org.slf4j</groupId>
```
slaskawi: Is it possible to relocate all those dependencies? It is very likely that someone will have SLF4J on his classpath and there might be a version clash.

all/embedded/pom.xml
```diff
@@ -138,6 +138,23 @@
...
+         <artifactId>slf4j-api</artifactId>
+         <scope>runtime</scope>
```
slaskawi: Why the scope is runtime? I think it should be compile here.

persistence/jdbc/pom.xml
```diff
@@ -21,6 +21,27 @@
       </dependency>
 
       <dependency>
+         <groupId>com.zaxxer</groupId>
+         <artifactId>HikariCP</artifactId>
+         <optional>true</optional>
```
slaskawi: I'm not convinced that we should use optional here. If the docs says that Hikari become our default pool, why not making it mandatory here. If we did this, we wouldn't need to modify none of the integration test poms.

persistence/jdbc/pom.xml
```diff
@@ -21,6 +21,27 @@
...
+      <dependency>
+         <groupId>org.jboss.slf4j</groupId>
+         <artifactId>slf4j-jboss-logging</artifactId>
+         <scope>runtime</scope>
+         <optional>true</optional>
+      </dependency>
+
+      <dependency>
+         <groupId>org.slf4j</groupId>
+         <artifactId>slf4j-api</artifactId>
```
slaskawi: Why do we need it here? Shouldn't this be pulled by Hikari?

javadoc/pom.xml
```diff
@@ -234,6 +234,16 @@
                            <artifactId>c3p0</artifactId>
                            <version>0.9.5-pre8</version>
                         </additionalDependency>
+                        <additionalDependency>
+                           <groupId>com.zaxxer</groupId>
+                           <artifactId>HikariCP</artifactId>
+                           <version>${version.hikaricp}</version>
+                        </additionalDependency>
+                        <additionalDependency>
+                           <groupId>org.jboss.slf4j</groupId>
```
slaskawi: SLF4J API should probably be here; `slf4j-jboss-logging` is only a bridge between implementations.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
```diff
@@ -14,86 +25,162 @@
...
+   private static final String HIKARI_CONFIG = "hikaricp.configurationFile";
+   private static final String HIKARI_PROPERTIES = "hikari.properties";
+   private static final String C3P0_PROPERTIES = "c3p0.properties";
+   private static final String C3P0_CONFIG = "c3p0-config.xml";
+
+   private HikariDataSource hikari;
```
slaskawi: I see a mix of c3p0 and hikari fields here... maybe we need 2 classes instead of 1 here?

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
```diff
@@ -14,86 +25,162 @@
...
+      try {
+         MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
+         ObjectName poolName = new ObjectName("com.zaxxer.hikari:type=Pool (" + hikari.getPoolName() + ")");
+         hikariMxBean = JMX.newMXBeanProxy(mBeanServer, poolName, HikariPoolMXBean.class);
+      } catch (MalformedObjectNameException ignore) {
+         // Ignore as this should never happen
+      }
+   }
+
+   private void initC3P0DataSource(PooledConnectionFactoryConfiguration pooledConfiguration) {
```
slaskawi: Having both `initC3P0DataSource` and `initHikariDataSource` is a sign that we need 2 classes...

persistence/jdbc/src/test/java/org/infinispan/test/fwk/UnitTestDatabaseManager.java
```diff
@@ -144,7 +144,7 @@ public static void verifyConnectionLeaks(ConnectionFactory connectionFactory) {
          PooledConnectionFactory pcf = (PooledConnectionFactory) connectionFactory;
          try {
             Thread.sleep(500); // C3P0 needs a little delay before reporting the correct number of connections. Bah!
```
slaskawi: ouch... can we get rid of this sleep?

slaskawi:
Nice piece of work @ryanemerson ! There are 2 big issues from me:

1. Shouldn't we make Hikari a mandatory dependency? Especially when it becomes our default. All poms could be simplified (probably a lot!).
2. We probably need 2 implementations of `PooledConnectionFactory`. Currently it contains a mix of 2 implementations which is not nice.

persistence/jdbc/src/test/java/org/infinispan/test/fwk/UnitTestDatabaseManager.java
ryanemerson: @slaskawi This is code was included as part of ISPN-2033 and I assumed that it was a necessary evil for C3P0 to play nice. I could try and change this, however this code will be removed in a future release when C3P0 support is removed.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
ryanemerson: @slaskawi I went down the route of a single class, as C3P0 support is deprecated and will soon be removed. I am happy to re-work it to utilise two classes though

ryanemerson:
@slaskawi Thanks for the feedback, I'll look into both

persistence/jdbc/src/test/java/org/infinispan/test/fwk/UnitTestDatabaseManager.java
slaskawi: Ok, in that case let's leave it

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
slaskawi: @ryanemerson Thanks. Yes, I believe that if we had 2 separate classes, removing C3P0 would be a matter of removing one file. Right now we would need to carefully remove class internals which might not be a good idea.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
```diff
@@ -14,86 +25,162 @@
...
  * @author Tristan Tarrant
+ * @author Ryan Emerson
  */
 public class PooledConnectionFactory extends ConnectionFactory {
 
-   private static final Log log = LogFactory.getLog(PooledConnectionFactory.class, Log.class);
-   private static final boolean trace = log.isTraceEnabled();
-   private ComboPooledDataSource pooledDataSource;
+   private static final Log LOG = LogFactory.getLog(PooledConnectionFactory.class, Log.class);
+   private static final boolean TRACE = LOG.isTraceEnabled();
```
tristantarrant: We do this for trace everywhere. If you need trace, you restart.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
danberindei: @ryanemerson I would change the system property name to start with `infinispan.jdbc`. I would also rather have the location of the properties file set in the store configuration instead of a global system property -- the application could use more than one Hikari/c3p0 connection pool, and it may want different settings.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
danberindei: Except the names `log` and `trace` are usually lowercase :)

danberindei:
Test suite hangs in `org.infinispan:integrationtests-osgi` after `java.lang.NoClassDefFoundError: org/infinispan/persistence/jpa/JpaStoreTest`

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/connectionfactory/PooledConnectionFactory.java
ryanemerson: @danberindei good points, I hadn't considered multiple connection pools.

ryanemerson:
@slaskawi @danberindei Updated based upon previous feedback.

bom/pom.xml
```diff
@@ -470,6 +472,18 @@
...
+
+         <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <version>${version.hikaricp}</version>
+         </dependency>
+
+         <dependency>
+            <groupId>org.jboss.slf4j</groupId>
+            <artifactId>slf4j-jboss-logging</artifactId>
```
slaskawi: I'm not entirely sure if this should go to `bom`. Maybe `parent` would be a better candidate? 

//cc @tristantarrant

slaskawi:
Looks good to me... only one small question about this SLF4J in bom/parent.

bom/pom.xml
ryanemerson: @slaskawi I was following the lead of the C3P0 dependencies here and placed the dependencies in both bom and parent.

slaskawi: Ok, thanks for clarification.

slaskawi:
LGTM

tristantarrant:
Welcome Hikari ! Bye bye c3p0 :)
Thanks @ryanemerson

PR #4433, opened 2016-06-29
ISPN-6350 Data race in the ShardIndexManager under topology changes

gustavonalle:
https://issues.jboss.org/browse/ISPN-6350


gustavonalle:
A bit of background:

#### In Theory ####
In a stable topology, the ```AffinityIndexManager``` is responsible to apply changes to the index if it is the owner of that index shard, and since an index shard is mapped to a ispn segment, that means only data writes of primary owned keys are delegated from the ```QueryInterceptor``` to the index manager. Topology changes causes the indexes to be closed and flushed, so that new owners can take over the lucene lock.

#### In Practice ####
During topology changes there is no general consensus on who is primary owner of a segment, and propagation of ```@TopologyChangedEvent``` is not fast enough in some cases, so two nodes can actually try to apply the work to the same index and competing for its lucene lock.

#### This PR ####

This PR changes the ```AffinityIndexManager```  (renamed from ```ShardIndexManager```) so that it checks who is actually holding the lock before applying the index work locally, and use this info to forward the indexwork to the owner of the lucene lock if needed. If for some reason the lucene lock cannot be obtained, it is retried after re-checking who is the owner. The index work forwarding is done **asynchronously**, in order to avoid distributed deadlocks, meaning that during this transient state of ownership changes, for sync index backend, there could be a small delay on index visibility of the data.

core/src/main/java/org/infinispan/remoting/transport/LocalModeAddress.java
```diff
@@ -0,0 +1,22 @@
...
+   }
+
+   @Override
+   public String toString() {
+      return "Local Address";
+   }
+
+   @Override
+   public int compareTo(Address o) {
+      return 0;
```
wburns: Hrmm, maybe this return 0 only if `o == this`?  Otherwise maybe `o.compareTo(this)` to ensure symmetry ?

lucene/lucene-directory/src/main/java/org/infinispan/lucene/impl/DirectoryImplementor.java
```diff
@@ -112,6 +114,9 @@ long fileLength(final String name) {
     }
 
     IndexOutput createOutput(final String name) {
+       if(log.isDebugEnabled()) {
```
wburns: Spacing

query/src/main/java/org/infinispan/query/affinity/AffinityErrorHandler.java
```diff
@@ -0,0 +1,56 @@
...
+      super(new LogErrorHandler());
+   }
+
+   public void initialize(OperationFailedHandler operationFailedHandler) {
+      if (log.isDebugEnabled()) log.debugf("Initializing with %s", operationFailedHandler);
+      this.operationFailedHandler = operationFailedHandler;
+   }
+
+   @Override
+   protected void errorOccurred(ErrorContext context) {
```
wburns: Should this log a debug as well?  An error sounds important :)

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -0,0 +1,286 @@
...
+      try {
+         Thread.sleep(POLL_WAIT);
+      } catch (InterruptedException e) {
+         Thread.currentThread().interrupt();
+      }
+      Address lockHolder = getLockHolder(getIndexName());
+      clearIfNeeded(lockHolder);
+      if (failingOperations.stream().anyMatch(w -> w.getIdInString() != null)) {
+         logDebug("Retrying operations %s", failingOperations);
+         CompletableFuture<Void> future = CompletableFuture.supplyAsync(() -> {
```
wburns: I wonder, should this be done in a specific thread pool?  These operations can cause IO right?  Normally we shouldn't be doing that in the common fork join pool.

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -0,0 +1,286 @@
...
+   private final Lock readLock = flushLock.readLock();
+
+
+   @Override
+   public void lockObtainFailed(List<LuceneWork> failingOperations) {
+      logDebug("Operation %s failed due to lock already in used", failingOperations);
+      try {
+         Thread.sleep(POLL_WAIT);
+      } catch (InterruptedException e) {
+         Thread.currentThread().interrupt();
```
wburns: Should we try to stop processing here instead?

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -0,0 +1,286 @@
...
+            } else {
+               log.errorf(throwable, "Error reapplying operation");
+            }
+         });
+      }
+
+   }
+
+   private void clearIfNeeded(Address lockHolder) {
+      if(!rpcManager.getMembers().contains(lockHolder)) {
```
wburns: Spacing

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -0,0 +1,286 @@
...
+         final Directory directory = getDirectoryProvider().getDirectory();
+         log.warn("Forcing clear of index lock");
```
wburns: Should this be debug?  If not we need method added to Log.

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -0,0 +1,286 @@
...
+      if (lockHolder != null) {
+         InfinispanDirectoryProvider directoryProvider = (InfinispanDirectoryProvider) getDirectoryProvider();
+         int activeDeleteTasks = directoryProvider.getActiveDeleteTasks();
+         while (activeDeleteTasks > 0) {
+            try {
+               Thread.sleep(POLL_WAIT);
+               logDebug("Waiting for pending delete tasks, remaining: %s", activeDeleteTasks);
+               activeDeleteTasks = directoryProvider.getActiveDeleteTasks();
+            } catch (InterruptedException e) {
+               Thread.currentThread().interrupt();
```
wburns: This will get in an infinite loop if it ever happens :-1: 
If we want to ignore the interrupt we need to set a boolean and then reset the interrupt flag after we are outside of the while loop.

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -0,0 +1,286 @@
...
+         logDebug("Lost ownership, new owner is %s", nextOwner);
+         destination = nextOwner;
+      } else {
+         destination = (lockHolder != null && !localAddress.equals(lockHolder)) ? lockHolder : localAddress;
+      }
+      return destination;
+   }
+
+   private Map<Address, List<LuceneWork>> partitionWorkByAddress(List<LuceneWork> works) {
+      return works.stream().collect(Collectors.toMap(this::getLocation, Arrays::asList, (w1, w2) -> {
```
wburns: Hrmm will this work properly if you have key collisions?  Won't it just take the last one added?

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -0,0 +1,286 @@
...
+         w1.addAll(w2);
+         return w1;
+      }));
+   }
+
+   @Override
+   public void performOperations(List<LuceneWork> workList, IndexingMonitor monitor) {
+      if (this.cache.getCacheConfiguration().clustering().cacheMode().isClustered()) {
+         readLock.lock();
+         Address localAddress = rpcManager.getAddress();
```
wburns: Maybe just put this in the try, seems a bit cleaner with the lock then :)

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -0,0 +1,286 @@
...
+            Map<Address, List<LuceneWork>> workByAddress = partitionWorkByAddress(workList);
+            logDebug("Applying work @ %s, workMap is %s", localAddress, workByAddress);
+            List<LuceneWork> localWork = workByAddress.get(localAddress);
+            if (localWork != null && !localWork.isEmpty()) {
+               logDebug("About to apply work locally %s, hasOwnership=%s", localWork, hasOwnership);
+               super.performOperations(localWork, monitor);
+               logDebug("Work applied");
+               workByAddress.remove(localAddress);
+            }
+            workByAddress.entrySet().stream().forEach(entry -> this.sendWork(entry.getValue(), entry.getKey()));
```
wburns: Can just use forEach on the map directly.

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -0,0 +1,286 @@
...
+         super.performOperations(workList, monitor);
+      }
+   }
+
+   private void sendWork(List<LuceneWork> works, Address destination) {
+      AffinityUpdateCommand indexUpdateCommand = new AffinityUpdateCommand(ByteString.fromString(cache.getName()));
+      byte[] serializedModel = getSerializer().toSerializedModel(works);
+      indexUpdateCommand.setSerializedWorkList(serializedModel);
+      List<Address> dest = Collections.singletonList(destination);
+      log.infof("Sending work to %s to %s", works, dest);
```
wburns: Info?  If so we need Log method.

wburns: Also maybe just pass destination, message would look a little nicer imo.

query/src/main/java/org/infinispan/query/backend/SearchableCacheConfiguration.java
```diff
@@ -130,21 +141,54 @@ public boolean isIdProvidedImplicit() {
...
             key = HSEARCH_PREFIX + key.toString();
          }
          target.put(key, entry.getValue());
-         if (key.toString().endsWith(INDEX_MANAGER_IMPL_NAME) && entry.getValue().equals(ShardIndexManager.class.getName())) {
+         boolean hasAffinity = false;
+         if (key.toString().endsWith(INDEX_MANAGER_IMPL_NAME) && entry.getValue().equals(AffinityIndexManager.class.getName())) {
             target.put(key.toString().replace(INDEX_MANAGER_IMPL_NAME, SHARDING_STRATEGY), AffinityShardIdentifierProvider.class.getName());
+            hasAffinity = true;
+         }
+         if (hasAffinity) {
```
wburns: Seems this can just be an else for the previous if and remove hasAffinity?

query/src/test/java/org/infinispan/query/helper/StaticTestingErrorHandler.java
```diff
@@ -35,6 +36,9 @@ public static void assertAllGood(Cache cache) {
       SearchIntegrator searchFactory = searchManager.unwrap(SearchIntegrator.class);
       ErrorHandler errorHandler = searchFactory.getErrorHandler();
       Assert.assertNotNull(errorHandler);
+      if(errorHandler instanceof WrappingErrorHandler) {
```
wburns: Spacing

wburns:
Added some comments, unfortunately I can't comment on the functionality of the changes itself, maybe @anistor or @Sanne ?

Also I noticed quite a few classes are missing javadoc for the headers if you could add @gustavonalle

core/src/main/java/org/infinispan/remoting/transport/LocalModeAddress.java
gustavonalle: Apparently not an issue since it's used on master like that, but I'll fix it  :+1:

query/src/main/java/org/infinispan/query/affinity/AffinityErrorHandler.java
gustavonalle: this will be logged via the wrapped Hibernate Search's ```ErrorHandler```, so no point in logging it again

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
gustavonalle: I wondered the same, but since this code path can only be reached sometimes during topology changes, I opted to leave it to the fj pool. On a second though it'd be better to use the ASYNC_OPERATIONS_EXECUTOR  :+1:

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
gustavonalle: ok

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
gustavonalle: well spotted, will fix it

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
gustavonalle: If more than 1 ```LuceneWork``` is associated with the same address, the collector's merge function will add them to a list on the value

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
gustavonalle: ok

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
gustavonalle: :+1:

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
gustavonalle: This should have been debug, will fix it

lucene/lucene-directory/src/main/java/org/infinispan/lucene/impl/DirectoryImplementor.java
gustavonalle: This entire module needs re-formatting, let's not bring it in this PR to reduce noise

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
gustavonalle: But since we are here...If there's a key collision, the merge function will fail with ```UnsupportedOperationException``` because ```Arrays.asList(T...)``` produces a list that cannot be resized, will have it sorted out

query/src/main/java/org/infinispan/query/backend/SearchableCacheConfiguration.java
gustavonalle: Indeed, this was not properly cleaned

core/src/main/java/org/infinispan/remoting/transport/LocalModeAddress.java
wburns: Yeah my guess is the compareTo is probably not used on this object, but just in case :)

gustavonalle:
Rebased

galderz:
@gustavonalle The PR still shows needing changes, is that the case?

gustavonalle:
@galderz I believe I addressed all the comments so far

galderz:
Integrated, thx @gustavonalle

gustavonalle:
@galderz integrated where? Cannot see it on master

galderz:
Fixed

PR #4434, opened 2016-06-30
ISPN-6703 ClearCommand hanging in the test suite

gustavonalle:
https://issues.jboss.org/browse/ISPN-6703

This should bring back query test suite to 1min rather than the 6min currently


tristantarrant:
Pushed to master, thanks

PR #4440, opened 2016-07-01
ISPN-6217 Created Javadocs for remote client distribution

slaskawi:
https://issues.jboss.org/browse/ISPN-6217


danberindei:
There are some conflicts, I believe from Tristan's changes.

slaskawi:
Rebased

javadoc-remote/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+-&lt;script type='text/javascript'&gt;
+-var gaJsHost = (("https:" == document.location.protocol) ? "https://ssl." : "http://www.");
+-document.write(unescape("%3Cscript src='" + gaJsHost + "google-analytics.com/ga.js' type='text/javascript'%3E%3C/script%3E"));
+-&lt;/script&gt;
+-&lt;script type='text/javascript'&gt;
+-try {
+-var pageTracker = _gat._getTracker("UA-8601422-4");
+-pageTracker._trackPageview();
+-} catch(err) {}&lt;/script&gt;
+            </analytics.code>
```
tristantarrant: We need to update the analytics code to the new async stuff:
`<script>
     (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
                        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
                        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
                        })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
     
    ga('create', 'UA-8601422-1', 'auto');
    ga('send', 'pageview');
    </script>`

slaskawi:
@tristantarrant Updated. The API docs pages look good in the browser but I have no idea how to verify if they work correctly.

tristantarrant:
The release script needs to be changed to also upload the remote javadocs.
Changes will also be required on the infinispan.github.io side to link to the remote javadocs.

slaskawi:
@tristantarrant This might be a little bit tricky...

Looking at https://github.com/infinispan/infinispan/blob/master/bin/release.py#L202, I see that we upload only one Javadocs package. The [Infinispan documentation page](http://infinispan.org/documentation) has one reference to the apidocs and I think two links (as for the remote package and one for everything else) would be a bit confusing.

So my proposition is to add remote javadocs to `-all` distribution and leave the documentation page with 1 javadocs link. 

WDYT?

slaskawi:
Thumbs up means yes :) Let me refactor this...

slaskawi:
@tristantarrant Updated and rebased. 

Unfortunately I had to introduce another module called `javadocs-all` with all embedded modules and Hot Rod client. I also couldn't use transitive dependencies in `javadocs-all` since the javadoc plugin does not support it...

slaskawi:
Rebased and updated...

@tristantarrant May I ask you to take a look?

tristantarrant:
Actually @slaskawi it would be better for the "all" package to be simply named "javadoc" as before: in this way you won't have to fix the site

slaskawi:
@tristantarrant I analyzed the PR again and it looks like we need 3 JavaDoc packages:

- Embedded (only with core javadocs used in `minimal` distribution)
- Remote (for `remote` one)
- All (for `all` distribution and website)

I think it's beneficial to add those suffixes (even though it might slightly complicate packaging scripts). Does it make sense

slaskawi:
I'm turning it back to "read for review" state.

tristantarrant:
@slaskawi I think we can safely avoid deploying the javadocs artifacts to Maven: they are only needed for the distribution packages.

slaskawi:
@tristantarrant Fixed. Ready for review.

tristantarrant:
Can you please move javadoc-[all|embeded|remote] as children of the javadoc directory ?

slaskawi:
Of course... on my way...

slaskawi:
@tristantarrant Moved javadocs to their own directory.

tristantarrant:
Pushed, thanks

slaskawi:
Awesome, thanks Tristan!

PR #4446, opened 2016-07-07
ISPN-6743: Upsert does not work for MySQL

ryanemerson:
https://issues.jboss.org/browse/ISPN-6743


danberindei:
Integrated, thanks Ryan!

PR #4448, opened 2016-07-07
ISPN-6847 Kubernetes PING integration

slaskawi:
https://issues.jboss.org/browse/ISPN-6847

Library mode deployment looks like this: https://github.com/infinispan/infinispan-simple-tutorials/pull/12


all/embedded/pom.xml
```diff
@@ -325,12 +349,16 @@
...
+                           !org.jboss.dmr,
                            !org.iq80.snappy,
                            !org.jboss.marshalling.*,
                            !org.xerial.snappy,
                            !sun.misc,
                            !sun.reflect,
                            !sun.nio.ch,
+                           !__redirected,
+                           com.sun.net.httpserver;resolution:=optional,
+                           io.undertow.*;resolution:=optional,
```
slaskawi: Interesting piece for review: Use "better" http servers (like Undertow) when available.

as-modules/embedded/src/main/resources/org/jgroups/kubernetes/main/module.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<module xmlns="urn:jboss:module:1.3" name="org.jgroups.kubernetes.kubernetes">
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <artifact name="${org.jgroups.kubernetes:kubernetes}"/>
+        <artifact name="${org.jgroups.kubernetes:common}"/>
```
slaskawi: Interesting piece for review: All `KUBE_PING` bits are in 1 module. There are [plans for refactoring this protocol](https://issues.jboss.org/browse/CLOUD-662), so I would like to limit the scope of changes on our side

core/pom.xml
```diff
@@ -30,6 +30,10 @@
          <artifactId>jgroups</artifactId>
       </dependency>
 
+      <dependency>
```
slaskawi: Interesting piece for review: Adding `KUBE_PING` to core, since we would like to support it out of the box.

There are some alternatives to this approach like:
* creating another module (infinispan-cloud?) with all additional discovery modules
* add other discovery modules only to Uber Jars

documentation/src/main/asciidoc/user_guide/cloud_aws.adoc
```diff
@@ -60,5 +60,37 @@ Finally you can configure your JGroups instances to use a shared storage to exch
 ----
 
 ==== JDBC_PING
-A similar approach to S3_PING, but using a JDBC connection to a shared database. On EC2 that is quite easy using Amazon RDS. See the link:$$http://community.jboss.org/wiki/JDBCPING$$[JDBC_PING Wiki page] for details. 
+A similar approach to S3_PING, but using a JDBC connection to a shared database. On EC2 that is quite easy using Amazon RDS. See the link:$$http://community.jboss.org/wiki/JDBCPING$$[JDBC_PING Wiki page] for details.
 
+== Kubernetes and OpenShift
```
slaskawi: Interesting piece for review: Here is how to run it :)

slaskawi:
Found one failure, retriggering CI
```
TestSuite: org.infinispan.container.versioning (1)
 DistWriteSkewConsistencyTest.testValidationOnlyInPrimaryOwner 	
```

slaskawi:
The failure is not related.

core/pom.xml
gustavonalle: This shouldn't be a compile dependency, make it optional instead?

slaskawi: Yes, I've been also thinking about that. Let me remove it.

slaskawi: Ha! Found why I decided to add it to the core....

Note that default configs are placed in the `core` (I'm not sure if this is the right place for them or not). We should support Kubernetes out of the box, so I added `default-jgroups-kubernetes.xml`. Since the configuration requires `KUBE_PING` in the classpath, we need to add this dependency to core.

slaskawi:
Rebased

core/pom.xml
gustavonalle: What I meant was to keep the dependency but make it ```<optional>true</optional>```

slaskawi: Unfortunately this won't solve the problem on client side... one can add `infinispan-core` as a library, use `default-jgroups-kubernetes.xml` (it's already in core after all!) and his application will crash (because there is no `KUBE_PING` protocol on the classpath).

gustavonalle: Yes, the user needs to add explicitly the dependency, that's how optional works...
OTOH, adding it as a compile dependency will bring it to the classpath of users who don't care about kubernetes.

gustavonalle: The other better alternative than adding a compile dep to the core would, as you mentioned, a separate module like ```infinispan-kubernetes``` with the example XML and the dependencies.

slaskawi: hmmm maybe we need something better...

How about creating a separate module (`infinispan-cloud`), placing all "cloud related" configurations there and adding all discovery protocol dependencies (not only Kubernetes but probably Azure as well (in the future)).

This way users would add only one dependency to their classpath (`infinispan-cloud`) and have all necessary dependencies and configurations in place...

gustavonalle: Sounds good to me

slaskawi: Ok, I created another module called `infinispan-cloud` and moved all cloud related configs there. I also added proper information in upgrading guide.

//cc @alanfx @tristantarrant

slaskawi:
CI didn't pick up newly released artifacts (Nexus uses some sync mechanisms underneath). Retriggering

slaskawi:
As we agreed waiting till https://github.com/infinispan/infinispan/pull/4449 and https://github.com/infinispan/infinispan/pull/4468 gets in. //cc @tristantarrant @wburns

slaskawi:
Please ignore the comment above... I attached it to wrong PR ;)

galderz:
Looks good, integrated

gustavonalle:
@galderz @slaskawi The PR was integrated with 
```
<dependency>
     <groupId>org.jgroups.kubernetes</groupId>
     <artifactId>kubernetes</artifactId>
</dependency>
```

in the ```infinispan-core```, I thought we'd agreed on putting all those optional stuff in the ```cloud``` module?

gustavonalle:
@slaskawi @galderz FYI, opened https://issues.jboss.org/browse/ISPN-6937

tristantarrant:
Shouldn't this also be &lt;optional&gt; ?

gustavonalle:
@tristantarrant what exactly?

slaskawi:
Done: https://github.com/infinispan/infinispan/pull/4492

PR #4452, opened 2016-07-12
[8.2.x] ISPN-6703 ClearCommand hanging in the test suite

gustavonalle:
https://issues.jboss.org/browse/ISPN-6703

### DIFF
--- a/query/src/main/java/org/infinispan/query/indexmanager/ClusteredSwitchingBackend.java
+++ b/query/src/main/java/org/infinispan/query/indexmanager/ClusteredSwitchingBackend.java
@@ -146,13 +146,18 @@ final class ClusteredSwitchingBackend implements LazyInitializableBackend {
    }
 
    private synchronized void applyViewChangedEvent(ViewChangedEvent e) {
-      assert e != null;
-      assert e.getNewMembers().size() > 0;
-      assert e.getNewMembers().get(0) != null;
+      List<Address> newMembers = e.getNewMembers();
       if (log.isDebugEnabled()) {
-         log.debug("Notified of new View! Members: " + e.getNewMembers());
+         log.debug("Notified of new View! Members: " + newMembers);
       }
-      final Address newmaster = e.getNewMembers().get(0);
+      handleTopologyChange(newMembers);
+   }
+
+   private synchronized void handleTopologyChange(List<Address> newMembers) {
+      assert newMembers != null;
+      assert newMembers.size() > 0;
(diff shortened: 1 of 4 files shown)


query/src/main/java/org/infinispan/query/indexmanager/ClusteredSwitchingBackend.java
```diff
@@ -146,13 +146,18 @@ public synchronized void lazyInitialize() {
    }
 
    private synchronized void applyViewChangedEvent(ViewChangedEvent e) {
-      assert e != null;
-      assert e.getNewMembers().size() > 0;
-      assert e.getNewMembers().get(0) != null;
+      List<Address> newMembers = e.getNewMembers();
       if (log.isDebugEnabled()) {
-         log.debug("Notified of new View! Members: " + e.getNewMembers());
+         log.debug("Notified of new View! Members: " + newMembers);
```
slaskawi: You could use `debugf`. This way you will also get rid of `log.isDebugEnabled`.

query/src/main/java/org/infinispan/query/indexmanager/ClusteredSwitchingBackend.java
```diff
@@ -146,13 +146,18 @@ public synchronized void lazyInitialize() {
...
       }
-      final Address newmaster = e.getNewMembers().get(0);
+      handleTopologyChange(newMembers);
+   }
+
+   private synchronized void handleTopologyChange(List<Address> newMembers) {
+      assert newMembers != null;
```
slaskawi: Do we need those asserts? Testing code should to to `src/test` and nobody uses `-ea` on production. I would remove them.

gustavonalle:
@slaskawi This PR is a backport of https://github.com/infinispan/infinispan/pull/4434 so I think it'd be better to keep "as is"

slaskawi:
ooh yeah.. sorry I haven't notice it... LGTM

query/src/main/java/org/infinispan/query/indexmanager/ClusteredSwitchingBackend.java
danberindei: `isDebugEnabled()` shouldn't be necessary anyway -- if the message is logged so often that the string concatenation influences performance, it should be `trace` instead of `debug`.

danberindei:
Integrating...

danberindei:
Integrated, thanks Gustavo!

PR #4454, opened 2016-07-12
Fix release.py script

vblagoje:
Minor issue that I encountered while cutting the release


tristantarrant:
Pushed to master, thanks

PR #4456, opened 2016-07-15
ISPN-6856 Testsuite: replace inheritance with @Factory

rvansa:
My motivation is since I am adding a new cache mode that could reuse many of the tests (with minor modifications), I would duplicate the classes. Adding a one-liner into factory method, or just an annotation is better.


tristantarrant:
+1,259 −2,305 Looks particularly good !
I will squash all "Test class-bloat reduction" commits and add an extra one that fixes the CDI failures.

tristantarrant:
Pushed to master. This is my extra commit: https://github.com/infinispan/infinispan/commit/5be27582af6edc624bda3aa7f0ea13db90b0cfc1

PR #4457, opened 2016-07-15
ISPN-6859 StateTransferLock should have non-blocking API

rvansa:
https://issues.jboss.org/browse/ISPN-6859

### DIFF
--- a/core/src/main/java/org/infinispan/commands/CreateCacheCommand.java
+++ b/core/src/main/java/org/infinispan/commands/CreateCacheCommand.java
@@ -5,6 +5,7 @@ import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.infinispan.Cache;
 import org.infinispan.commands.remote.BaseRpcCommand;
@@ -17,7 +18,6 @@ import org.infinispan.topology.CacheTopology;
 import org.infinispan.util.ByteString;
 import org.infinispan.util.TimeService;
 import org.infinispan.util.concurrent.CompletableFutures;
-import org.infinispan.util.concurrent.TimeoutException;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
--- a/core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
@@ -1,23 +1,37 @@
 package org.infinispan.interceptors.impl;
 
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
 
 import org.infinispan.commands.TopologyAffectedCommand;
 import org.infinispan.commands.VisitableCommand;
 import org.infinispan.commands.remote.GetKeysInGroupCommand;
 import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.context.InvocationContext;
+import org.infinispan.distribution.ch.ConsistentHash;
 import org.infinispan.distribution.group.GroupManager;
+import org.infinispan.factories.KnownComponentNames;
+import org.infinispan.factories.annotations.ComponentName;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.interceptors.BasicInvocationStage;
 import org.infinispan.interceptors.DDAsyncInterceptor;
+import org.infinispan.interceptors.InvocationStage;
 import org.infinispan.remoting.RemoteException;
 import org.infinispan.remoting.transport.jgroups.SuspectException;
 import org.infinispan.statetransfer.OutdatedTopologyException;
 import org.infinispan.statetransfer.StateTransferLock;
 import org.infinispan.statetransfer.StateTransferManager;
 import org.infinispan.topology.CacheTopology;
+import org.infinispan.util.concurrent.CompletableFutures;
+import org.infinispan.util.concurrent.TimeoutException;
 import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
 
 /**
  * A base class for a state transfer interceptor. It contains the base code to avoid duplicating in the two current
@@ -32,59 +46,73 @@ public abstract class BaseStateTransferInterceptor extends DDAsyncInterceptor {
    private final boolean trace = getLog().isTraceEnabled();
 
    protected StateTransferManager stateTransferManager;
-   private StateTransferLock stateTransferLock;
+   protected StateTransferLock stateTransferLock;
+   protected Executor remoteExecutor;
    private GroupManager groupManager;
    private long transactionDataTimeout;
+   private ScheduledExecutorService timeoutExecutor;
 
    @Inject
    public void init(StateTransferLock stateTransferLock, Configuration configuration,
-                    StateTransferManager stateTransferManager, GroupManager groupManager) {
+                    StateTransferManager stateTransferManager, GroupManager groupManager,
+                    @ComponentName(KnownComponentNames.TIMEOUT_SCHEDULE_EXECUTOR) ScheduledExecutorService timeoutExecutor,
+                    @ComponentName(KnownComponentNames.REMOTE_COMMAND_EXECUTOR) Executor remoteExecutor) {
       this.stateTransferLock = stateTransferLock;
       this.stateTransferManager = stateTransferManager;
       this.groupManager = groupManager;
+      this.timeoutExecutor = timeoutExecutor;
+      this.remoteExecutor = remoteExecutor;
       transactionDataTimeout = configuration.clustering().remoteTimeout();
    }
 
    @Override
    public BasicInvocationStage visitGetKeysInGroupCommand(InvocationContext ctx, GetKeysInGroupCommand command) throws Throwable {
-      final String groupName = command.getGroupName();
-      final boolean isOwner = groupManager.isOwner(groupName);
       updateTopologyId(command);
-      final int commandTopologyId = command.getTopologyId();
 
       if (ctx.isOriginLocal()) {
-         return invokeNext(ctx, command).compose((stage, rCtx, rCommand, rv, t) -> {
-            boolean shouldRetry;
-            if (t != null) {
-               // Must retry if we got an OutdatedTopologyException or SuspectException
-               Throwable ce = t;
-               while (ce instanceof RemoteException) {
-                  ce = ce.getCause();
-               }
-               shouldRetry = ce instanceof OutdatedTopologyException || ce instanceof SuspectException;
-            } else {
-               // Only check the topology id if if we are an owner
-               shouldRetry = isOwner && currentTopologyId() != commandTopologyId;
-            }
-            if (shouldRetry) {
-               return retryVisitGetKeysInGroupCommand(ctx, command, commandTopologyId);
-            }
-            // No retry, either rethrow the exception or return the current result
-            if (t != null) {
-               throw t;
-            } else {
-               return returnWith(rv);
-            }
-         });
+         return invokeNext(ctx, command).compose(this::handleLocalGetKeysInGroupReturn);
       } else {
          return invokeNext(ctx, command).thenAccept((rCtx, rCommand, rv) -> {
-            if (isOwner && currentTopologyId() != commandTopologyId) {
+            GetKeysInGroupCommand cmd = (GetKeysInGroupCommand) rCommand;
+            final int commandTopologyId = cmd.getTopologyId();
+            String groupName = cmd.getGroupName();
+            if (groupManager.isOwner(groupName) && currentTopologyId() != commandTopologyId) {
                throw new OutdatedTopologyException(
                      "Cache topology changed while the command was executing: expected " +
                            commandTopologyId + ", got " + currentTopologyId());
             }
          });
       }
+   }
+
+   private BasicInvocationStage handleLocalGetKeysInGroupReturn(BasicInvocationStage stage, InvocationContext ctx,
+                                                                VisitableCommand command, Object rv, Throwable t) throws Throwable {
+      GetKeysInGroupCommand cmd = (GetKeysInGroupCommand) command;
+      final int commandTopologyId = cmd.getTopologyId();
+      boolean shouldRetry;
+      if (t != null) {
+         // Must retry if we got an OutdatedTopologyException or SuspectException
+         Throwable ce = t;
+         while (ce instanceof RemoteException) {
+            ce = ce.getCause();
+         }
+         shouldRetry = ce instanceof OutdatedTopologyException || ce instanceof SuspectException;
+      } else {
+         // Only check the topology id if if we are an owner
+         shouldRetry = groupManager.isOwner(cmd.getGroupName()) && currentTopologyId() != commandTopologyId;
+      }
+      if (shouldRetry) {
+         logRetry(cmd);
+         // We increment the topology id so that updateTopologyIdAndWaitForTransactionData waits for the next topology.
+         // Without this, we could retry the command too fast and we could get the OutdatedTopologyException again.
+         int newTopologyId = Math.max(currentTopologyId(), commandTopologyId + 1);
+         cmd.setTopologyId(newTopologyId);
+         CompletableFuture<Void> transactionDataFuture = stateTransferLock.transactionDataFuture(newTopologyId);
+         return retryWhenDone(transactionDataFuture, newTopologyId, ctx, command, null).compose(this::handleLocalGetKeysInGroupReturn);
+      } else {
+         return stage;
+      }
+
    }
 
    protected final void logRetry(VisitableCommand command) {
@@ -98,14 +126,6 @@ public abstract class BaseStateTransferInterceptor extends DDAsyncInterceptor {
       return cacheTopology == null ? -1 : cacheTopology.getTopologyId();
    }
 
-   protected final void waitForTransactionData(int topologyId) throws InterruptedException {
-      stateTransferLock.waitForTransactionData(topologyId, transactionDataTimeout, TimeUnit.MILLISECONDS);
-   }
-
-   protected final void waitForTopology(int topologyId) throws InterruptedException {
-      stateTransferLock.waitForTopology(topologyId, transactionDataTimeout, TimeUnit.MILLISECONDS);
-   }
-
    protected final void updateTopologyId(TopologyAffectedCommand command) throws InterruptedException {
       // set the topology id if it was not set before (ie. this is local command)
       // TODO Make tx commands extend FlagAffectedCommand so we can use CACHE_MODE_LOCAL in TransactionTable.cleanupStaleTransactions
@@ -117,16 +137,96 @@ public abstract class BaseStateTransferInterceptor extends DDAsyncInterceptor {
       }
    }
 
-   private BasicInvocationStage retryVisitGetKeysInGroupCommand(InvocationContext context, GetKeysInGroupCommand command,
-                                                  int commandTopologyId) throws Throwable {
-      logRetry(command);
-      // We increment the topology id so that updateTopologyIdAndWaitForTransactionData waits for the next topology.
-      // Without this, we could retry the command too fast and we could get the OutdatedTopologyException again.
-      int newTopologyId = Math.max(currentTopologyId(), commandTopologyId + 1);
-      command.setTopologyId(newTopologyId);
-      waitForTransactionData(newTopologyId);
-      return visitGetKeysInGroupCommand(context, command);
+   protected <T extends VisitableCommand> InvocationStage retryWhenDone(
+         CompletableFuture<Void> future, int topologyId, InvocationContext ctx, T command, Consumer<ConsistentHash> consistentHashConsumer) throws Throwable {
+      if (future.isDone()) {
+         getLog().tracef("Retrying command %s for topology %d", command, topologyId);
+         return invokeNext(ctx, command);
+      } else {
+         CancellableRetry<T> cancellableRetry = new CancellableRetry<>(command, topologyId, stateTransferManager, consistentHashConsumer);
+         // We have to use handleAsync and rethrow the exception in the handler, rather than
+         // thenComposeAsync(), because if `future` completes with an exception we want to continue in remoteExecutor
+         CompletableFuture<Void> retryFuture = future.handleAsync(cancellableRetry, remoteExecutor);
+         cancellableRetry.setRetryFuture(retryFuture);
+         // We want to time out the current command future, not the main topology-waiting future,
+         // but the command future can take longer time to finish.
+         ScheduledFuture<?> timeoutFuture = timeoutExecutor.schedule(cancellableRetry, transactionDataTimeout, TimeUnit.MILLISECONDS);
+         cancellableRetry.setTimeoutFuture(timeoutFuture);
+         return invokeNextAsync(ctx, command, retryFuture);
+      }
    }
 
    protected abstract Log getLog();
+
+   private static class CancellableRetry<T extends VisitableCommand> implements BiFunction<Void, Throwable, Void>, Runnable {
+      private static final AtomicReferenceFieldUpdater<CancellableRetry, Throwable> cancellableRetryUpdater
+            = AtomicReferenceFieldUpdater.newUpdater(CancellableRetry.class, Throwable.class, "cancelled");
+      private static final AtomicReferenceFieldUpdater<CancellableRetry, Object> timeoutFutureUpdater
+            = AtomicReferenceFieldUpdater.newUpdater(CancellableRetry.class, Object.class, "timeoutFuture");
+
+      private static final Log log = LogFactory.getLog(CancellableRetry.class);
+      private static final Throwable DUMMY = new Throwable("Command is retried"); // should not be ever thrown
+
+      private final T command;
+      private final int topologyId;
+      private final Consumer<ConsistentHash> consistentHashConsumer;
+      private final StateTransferManager stateTransferManager;
+      private volatile Throwable cancelled = null;
+      // retryFuture is not volatile because it is used only in the timeout handler = run()
+      // and that is scheduled after retryFuture is set
+      private CompletableFuture<Void> retryFuture;
+      // ScheduledFuture does not have any dummy implementations, so we'll use plain Object as the field
+      private volatile Object timeoutFuture;
+
+      public CancellableRetry(T command, int topologyId, StateTransferManager stateTransferManager, Consumer<ConsistentHash> consistentHashConsumer) {
+         this.command = command;
+         this.topologyId = topologyId;
+         this.consistentHashConsumer = consistentHashConsumer;
+         this.stateTransferManager = stateTransferManager;
+      }
+
+      /**
+       * This is called when the topology future completes (successfully or exceptionally)
+       */
+      @Override
+      public Void apply(Void nil, Throwable throwable) {
+         if (!timeoutFutureUpdater.compareAndSet(this, null, DUMMY)) {
+            ((ScheduledFuture) timeoutFuture).cancel(false);
+         }
+
+         if (throwable != null) {
+            throw CompletableFutures.asCompletionException(throwable);
+         }
+         if (!cancellableRetryUpdater.compareAndSet(this, null, DUMMY)) {
+            log.tracef("Not retrying command %s as it has been cancelled.", command);
+            throw CompletableFutures.asCompletionException(cancelled);
+         }
+         log.tracef("Retrying command %s for topology %d", command, topologyId);
+         if (consistentHashConsumer != null) {
+            consistentHashConsumer.accept(stateTransferManager.getCacheTopology().getReadConsistentHash());
+         }
+         return null;
+      }
+
+      /**
+       * This is called when the timeout elapses.
+       */
+      @Override
+      public void run() {
+         TimeoutException timeoutException = new TimeoutException("Timed out waiting for topology " + topologyId);
+         if (cancellableRetryUpdater.compareAndSet(this, null, timeoutException)) {
+            retryFuture.completeExceptionally(timeoutException);
+         }
+      }
+
+      void setRetryFuture(CompletableFuture<Void> retryFuture) {
+         this.retryFuture = retryFuture;
+      }
+
+      void setTimeoutFuture(ScheduledFuture<?> timeoutFuture) {
+         if (!timeoutFutureUpdater.compareAndSet(this, null, timeoutFuture)) {
+            timeoutFuture.cancel(false);
+         }
+      }
+   }
 }
--- a/core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java
+++ b/core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java
@@ -34,48 +34,84 @@ public abstract class BaseBlockingRunnable implements BlockingRunnable {
 
    @Override
    public void run() {
+      if (sync) {
+         runSync();
+      } else {
+         runAsync();
+      }
+   }
+
+   private void runSync() {
       try {
-         response = beforeInvoke();
-         if (response != null) {
-            reply.reply(response);
-            onFinally();
-            return;
-         }
-
-         CompletableFuture<Response> future = handler.invokeCommand(command);
-         if (sync) {
-            response = future.join();
-            afterInvoke();
-            reply.reply(response);
-            onFinally();
-            return;
-         }
-
-         future.whenComplete((response1, throwable) -> {
-            try {
-               if (throwable == null) {
-                  response = response1;
-                  afterInvoke();
-               } else {
-                  if (throwable instanceof CompletionException) {
-                     throwable = throwable.getCause();
-                  }
-                  afterCommandException(throwable);
-               }
-            } finally {
-               reply.reply(response);
-               onFinally();
+         CompletableFuture<Response> beforeFuture = beforeInvoke();
+         if (beforeFuture != null) {
+            response = beforeFuture.join();
+            if (response != null) {
+               return;
             }
-         });
-      } catch (Throwable t) {
-         if (t.getCause() != null && t instanceof CompletionException) {
-            t = t.getCause();
          }
-         afterCommandException(t);
-         // We didn't get a CompletableFuture from invokeCommand, so we have to send the reply here
+         CompletableFuture<Response> commandFuture = handler.invokeCommand(command);
+         response = commandFuture.join();
+         afterInvoke();
+      } catch (Throwable t) {
+         afterCommandException(unwrap(t));
+      } finally {
          reply.reply(response);
          onFinally();
       }
+   }
+
+   private void runAsync() {
+      CompletableFuture<Response> beforeFuture = beforeInvoke();
+      if (beforeFuture == null) {
+         invoke();
+      } else {
+         beforeFuture.whenComplete((rsp, throwable) -> {
+            if (rsp != null) {
+               response = rsp;
+               reply.reply(rsp);
+               onFinally();
+            } else if (throwable != null) {
+               afterCommandException(unwrap(throwable));
+               reply.reply(response);
+               onFinally();
+            } else {
+               invoke();
+            }
+         });
+      }
+   }
+
+   private void invoke() {
+      CompletableFuture<Response> commandFuture;
+      try {
+         commandFuture = handler.invokeCommand(command);
+      } catch (Throwable t) {
+         afterCommandException(unwrap(t));
+         reply.reply(response);
+         onFinally();
+         return;
+      }
+      commandFuture.whenComplete((rsp, throwable) -> {
+         try {
+            if (throwable == null) {
+               response = rsp;
+               afterInvoke();
+            } else {
+               afterCommandException(unwrap(throwable));
+            }
+         } finally {
+            reply.reply(response);
+            onFinally();
+         }
+      });
+   }
+
+   private Throwable unwrap(Throwable throwable) {
+      if (throwable instanceof CompletionException && throwable.getCause() != null) {
+         throwable = throwable.getCause();
+      }
+      return throwable;
    }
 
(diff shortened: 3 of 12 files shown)


slaskawi:
I'm not an expert in this area but I like the idea! :+1:

rvansa:
@danberindei Besides the need to rebase this, do you have any concerns?

rvansa:
@danberindei Rebased, just messed imports...

core/src/main/java/org/infinispan/commands/CreateCacheCommand.java
```diff
@@ -92,9 +94,11 @@ protected void waitForCacheToStabilize(Cache<Object, Object> cache, Configuratio
...
          long remainingTime = timeService.remainingTime(endTime, TimeUnit.NANOSECONDS);
          try {
-            stateTransferLock.waitForTopology(cacheTopology.getTopologyId() + 1, remainingTime,
-                  TimeUnit.NANOSECONDS);
-         } catch (TimeoutException ignored) {
+            CompletableFuture<Void> topologyFuture = stateTransferLock.topologyFuture(cacheTopology.getTopologyId() + 1);
+            if (topologyFuture != null) topologyFuture.get(remainingTime, TimeUnit.NANOSECONDS);
+         } catch (ExecutionException e) {
+            throw new CacheException(e.getCause());
+         } catch (java.util.concurrent.TimeoutException e) {
```
danberindei: Sever other classes have the same code, so I'd rather keep the `waitForTopology` and `waitForTransactionData` methods in `StateTransferLock` until we've made all its callers asynchronous.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
```diff
@@ -125,8 +126,40 @@ protected final void updateTopologyId(TopologyAffectedCommand command) throws In
...
+               getLog().tracef("Retrying command %s for topology %d", command, topologyId);
+               return handler.handle(ctx, command);
+            } catch (Throwable t) {
+               throw new CacheException(t);
+            }
+         });
+         // We want to time out the current command future, not the main topology-waiting future,
+         // but the command future can take longer time to finish.
+         // This scheduled future does not have to be cancelled because it won't do anything (it will
+         // sit idle in the scheduler queue).
```
danberindei: I'm not so sure about not needing to cancel these scheduled tasks. They may not use CPU by themselves, but having a larger queue of scheduled tasks makes all timer operations a bit slower.
Maybe I'm just overreacting because of the bad experience with the RPC timeout tasks, but I'd like to cancel these as soon as possible.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
```diff
@@ -125,8 +126,40 @@ protected final void updateTopologyId(TopologyAffectedCommand command) throws In
...
+               if (!future.isDone()) {
+                  retryFuture.completeExceptionally(new TimeoutException("Timed out waiting for topology " + topologyId));
+               }
+            }, transactionDataTimeout, TimeUnit.MILLISECONDS);
+         return retryFuture;
+      }
+   }
+
+   protected interface RetryHandler<T extends VisitableCommand> {
+      CompletableFuture<Void> handle(InvocationContext ctx, T command) throws Throwable;
```
danberindei: Reminds me of the AsyncInterceptor interface :)

core/src/main/java/org/infinispan/remoting/inboundhandler/DefaultTopologyRunnable.java
```diff
@@ -39,19 +38,31 @@ public boolean isReady() {
    }
 
    @Override
-   protected Response beforeInvoke() throws Exception {
+   protected InvocationStatus beforeInvoke() throws Exception {
+      // Read commands are executed directly in OOB thread pool. When the many commands are received prior
+      // to the first topology update (when the node is joining), these can wait for transaction data
+      // (even in non-tx mode) and deplete OOB TP => the node stops being responsive event to ST.
```
danberindei: Read commands should never wait, because they don't have a topology id (yet?)

danberindei: BTW, with ISPN-5469 I'm trying to remove `DefaultTopologyRunnable` altogether and wait (asynchronously) for the topology in the command itself.

core/src/main/java/org/infinispan/statetransfer/StateTransferLockImpl.java
```diff
@@ -25,14 +34,21 @@
    private static final int TOPOLOGY_ID_STOPPED = Integer.MAX_VALUE;
 
    private final ReadWriteLock ownershipLock = new ReentrantReadWriteLock();
+   private Executor remoteExecutor;
 
    private volatile int topologyId = -1;
    private final Lock topologyLock = new ReentrantLock();
-   private final Condition topologyCondition = topologyLock.newCondition();
+   private final Map<Integer, CompletableFuture<Void>> topologyFutures = new HashMap<>();
```
danberindei: I don't really want to worry about cleaning up old topology futures, I think we should be able to do it with a single volatile `CompletableFuture<Void>`.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
rvansa: Are you aware how expensive is the lookup in the queue (needed for the removal)? If the task is a O(1) retrieval + no-op, it seemed more effective to let it sit idle than try to find it. If the queue lookup is O(1), you still need to keep a reference to the scheduled future in the lambda above.

core/src/main/java/org/infinispan/commands/CreateCacheCommand.java
rvansa: Ok, I'll add default deprecated method to the iface.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
rvansa: Yes, because this was written to by aligned to that :)

core/src/main/java/org/infinispan/remoting/inboundhandler/DefaultTopologyRunnable.java
rvansa: I think I had this problem with `GetAllCommand` , but not sure atm.

core/src/main/java/org/infinispan/statetransfer/StateTransferLockImpl.java
rvansa: Hmm, that makes a case when someone wants topology > topologyId + 2 rather complicated, but I'll try to deal with it.

core/src/main/java/org/infinispan/remoting/inboundhandler/DefaultTopologyRunnable.java
danberindei: `ClusteredGetAllCommand` doesn't have a topology id either. Maybe it was one of the functional commands?

Either way, I think the proper way to fix this is to make the command's `canBlock()` return `true` if it needs to wait for a particular topology.

rvansa: I hope that `canBlock()` is about to go, with complete async processing, waiting for a topology is conceptually the same as waiting for lock.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
danberindei: Based on the comments in `ScheduledThreadPoolExecutor.DelayedWorkQueue`, all operations are `O(log n)`. Running a task is also `O(log n)`, because it pushing one task to the front and restoring the heap invariant. But the constant factor is larger, because it involves waking up the timer thread.

core/src/main/java/org/infinispan/statetransfer/StateTransferLockImpl.java
danberindei: Please wait for now, I started going it that way myself on my ISPN-5469 branch :)

rvansa: Actually it's not that complex, I can just return `topologyFuture.thenCompose(nil -> topologyFuture(expectedTopology)`.

rvansa:
Rebased and addressed comments.

core/src/main/java/org/infinispan/remoting/inboundhandler/DefaultTopologyRunnable.java
danberindei: You're right, in the bright future the command invocation will be asynchronous and we won't need `BlockingRunnable` and its subclasses at all. However, I don't expect to be able to remove `canBlock()` in version 9.0.

core/src/main/java/org/infinispan/statetransfer/StateTransferLockImpl.java
```diff
@@ -71,38 +88,38 @@ public void notifyTransactionDataReceived(int topologyId) {
...
       transactionDataTopologyId = topologyId;
+
       transactionDataLock.lock();
+      CompletableFuture<Void> oldFuture = transactionDataFuture;
       try {
-         transactionDataCondition.signalAll();
+         transactionDataFuture = new CompletableFuture<>();
       } finally {
          transactionDataLock.unlock();
+         remoteExecutor.execute(() -> oldFuture.complete(null));
```
danberindei: Haha, I wanted to publish my own version for comparison but it doesn't look that different now :)

One difference is that I removed the lock completely. I'm not sure if it's safe, I did have some new failures, but at the very least we could have a single lock for topology and for tx data. I'm also not using the remote executor here, because I want each blocked task to run on a different thread.

core/src/main/java/org/infinispan/statetransfer/StateTransferLockImpl.java
```diff
@@ -71,38 +88,38 @@ public void notifyTransactionDataReceived(int topologyId) {
...
       }
    }
 
    @Override
-   public void waitForTransactionData(int expectedTopologyId, long timeout,
-                                      TimeUnit unit) throws InterruptedException {
+   public CompletableFuture<Void> transactionDataFuture(int expectedTopologyId) {
       if (transactionDataTopologyId >= expectedTopologyId)
-         return;
+         return null;
```
danberindei: This is going to cause a NPE in CompletableFuture:969

core/src/main/java/org/infinispan/statetransfer/StateTransferLockImpl.java
rvansa: I don't think it's safe, at least here it wouldn't be. If you don't check topologyId & return the future atomically, you might return future that belongs to a topology with higher id than is the expected one.

One lock or two, little difference.

Using the remote executor rather in STI makes sense.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
```diff
@@ -125,8 +126,39 @@ protected final void updateTopologyId(TopologyAffectedCommand command) throws In
...
+               return handler.handle(ctx, command);
+            } catch (Throwable t) {
+               throw new CacheException(t);
+            }
+         });
+         // We want to time out the current command future, not the main topology-waiting future,
+         // but the command future can take longer time to finish.
+         ScheduledFuture<?> timeoutFuture = timeoutExecutor.schedule(() -> {
+            if (!future.isDone()) {
+               retryFuture.completeExceptionally(new TimeoutException("Timed out waiting for topology " + topologyId));
```
danberindei: Why not `future.completeExceptionally(...)`? That will skip the retry of the command if the timeout task runs before `future` completes, instead `retryFuture.completeExceptionally(...)` still allows the retry command to go through the invocation chain again.

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
```diff
@@ -186,20 +187,37 @@ public void init(StateTransferManager stateTransferManager) {
...
+               });
+            }
+         } else if (ce instanceof SuspectException) {
+            if (trace)
+               log.tracef("Retrying command because of suspected node, current topology is %d: %s",
+                  currentTopologyId(), rCommand);
+            // Upon SuspectException we don't wait for newer topology, as it is possible that current topology
+            // is actual but the view still contains a node that's about to leave; a broadcast to all nodes
+            // then can end with suspect exception, but we won't get any new topology. An example of this situation
+            // is when a node sends leave - topology can be installed before the new view.
```
danberindei: The opposite can be a problem as well: in a distributed cache, we don't broadcast write commands but instead send them to the owners in the current topology. If a node crashes and we get a `SuspectException`, we don't want to go in a busy loop, retrying the command and failing with a new `SuspectException` until we receive the new topology. (Not least, because using lots of CPU retrying will delay the processing of the new topology.)

We work around the problem you mention here by always using `ResponseMode.SYNCHRONOUS_IGNORE_LEAVERS` when we broadcast a cache command. That way, we can never get a `SuspectException` after we've already updated our topology.

core/src/main/java/org/infinispan/statetransfer/StateTransferLock.java
```diff
@@ -39,15 +43,35 @@
    // transaction data latch
    void notifyTransactionDataReceived(int topologyId);
 
-   void waitForTransactionData(int expectedTopologyId, long timeout, TimeUnit unit) throws InterruptedException;
+   CompletableFuture<Void> transactionDataFuture(int expectedTopologyId);
+
+   @Deprecated
+   default void waitForTransactionData(int expectedTopologyId, long timeout, TimeUnit unit) throws InterruptedException, TimeoutException {
```
danberindei: This isn't an SPI, so I'm not sure it's worth putting the implementation in the interface.

core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java
```diff
@@ -28,10 +28,18 @@ protected BaseBlockingRunnable(BasePerCacheInboundInvocationHandler handler, Cac
...
+         invocationStatus = beforeInvoke();
+         switch (invocationStatus) {
+            case CONTINUE:
+               response = handler.invokePerform(command);
+               break;
+            case CACHE_NOT_FOUND:
+               response = CacheNotFoundResponse.INSTANCE;
+               break;
+            case DEFERRED:
+               return;
```
danberindei: The remote command execution goes through this method even if the delivery order is `DeliverOrder.PER_SENDER`. But in that case, we rely on JGroups to enforce the delivery order, and JGroups instead relies on us processing the command synchronously. So you cannot continue the invocation on another thread if the command is an asynchronous write/prepare.

core/src/main/java/org/infinispan/statetransfer/StateTransferLockImpl.java
```diff
@@ -122,36 +139,36 @@ public void notifyTopologyInstalled(int topologyId) {
...
          log.tracef("Waiting for topology %d to be installed, current topology is %d", expectedTopologyId, topologyId);
       }
       topologyLock.lock();
       try {
-         long timeoutNanos = unit.toNanos(timeout);
-         while (topologyId < expectedTopologyId && timeoutNanos > 0) {
-            timeoutNanos = topologyCondition.awaitNanos(timeoutNanos);
+         if (topologyId >= expectedTopologyId) {
+            return null;
+         } else if (topologyId == expectedTopologyId - 1){
```
danberindei: Adding this check means our tests will rarely (if ever) go through the `thenCompose` branch, so you'll need a new test to exercise that branch. OTOH having to wait is already a pretty rare occurrence, so I'd rather remove this check instead.

core/src/main/java/org/infinispan/xsite/statetransfer/XSiteStateProviderImpl.java
```diff
@@ -309,6 +307,9 @@ public void run() {
          } catch (InterruptedException e) {
             error = true;
             log.unableToSendXSiteState(xSiteBackup.getSiteName(), e);
+         } catch (TimeoutException e) {
```
danberindei: It's a bit weird that we're not catching `Throwable` here...

core/src/test/java/org/infinispan/statetransfer/WriteSkewDuringStateTransferTest.java
```diff
@@ -270,10 +275,12 @@ public boolean isApplicable(InvocationContext context, VisitableCommand command)
          @Override
          public void before(InvocationContext context, VisitableCommand command, Cache cache) {
             try {
-               cache.getAdvancedCache().getComponentRegistry().getStateTransferLock().waitForTopology(currentTopology + 1,
-                                                                                                      10, TimeUnit.SECONDS);
+               CompletableFuture<Void> topologyFuture = cache.getAdvancedCache().getComponentRegistry().getStateTransferLock().topologyFuture(currentTopology + 1);
+               if (topologyFuture != null) topologyFuture.get(10, TimeUnit.SECONDS);
```
danberindei: You can revert this change.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
danberindei: Then you can also move the `timeoutFuture` before the `retryFuture`, and avoid the separate `whenComplete`.

rvansa: `future` is the one  you got from `stateTransferLock` and conceptually it should be a read-only future. You should not cancel all tasks (incl. future ones) waiting for that future just because one command has timed out.
We could get `CompletionStage` but I faintly remember that there was some issue with that - gotta check that out.

core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java
rvansa: Ack, so we have to add `forceSync` field to `DefaultTopologyRunnable`?

rvansa: @belaban Hi Bela, do you plan to add any option to JGroups 4.x that will allow the application to return a `CompletableFuture` when it is processing an ordered message, instead of forcing the thread to really process it synchronously?

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
rvansa: Okay, this was probably popping up with functional read command, where I use just the default sync options. I don't know why this was causing a broadcast, though.

rvansa: I'll add a comment about this assumption to the `SYNCHRONOUS_IGNORE_LEAVERS` option.

core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java
belaban: Not sure I follow (not familiar with this code). (Apart from the fact that I'd like Infinispan to switch to using messages and a channel instead of MessageDispatcher), processing a bunch of messages can be done asynchronously, via the async invocation interface [1]. It is the application's responsibility to enforce ordering at that point, as JGroups considers a message delivered as soon as `receive()` (or `handle()`) returned.

[1] http://www.jgroups.org/manual4/index.html#AsyncInvocation

core/src/main/java/org/infinispan/statetransfer/StateTransferLock.java
rvansa: You've asked for a temporary replacement until we fix the usages. As the method should not be in the interface either, we should not force it in the implementation.

core/src/main/java/org/infinispan/statetransfer/StateTransferLockImpl.java
rvansa: Oops, well spotted.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
danberindei: Right, I forgot that... Cancelling that future would have been ok if the timeout was handled in `StateTransferLockImpl`, because all the futures were using the same timeout, but here we don't have any guarantees that other commands that don't go through `StateTransferInterceptor` use the same timeout. 

Still, I don't think completing `retryFuture` will stop the handler from running, and that can be a problem.

core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java
rvansa: @belaban What I had in mind is preventing the duplication of ordering code (in JGroups and in application). Looking into `AsyncRequestHandler`, it would be cool if the message could be considered delivered (therefore, allowing next messages) only after `Response.send(...)` is invoked.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
rvansa: Ah, I see what you mean. I'll have to make the lambda in thenCompose an inner class, and CAS a field in there.
Or, I could use
```
CompletableFuture<Void> cancellableFuture = future.thenApply(Function.identity());
CompletableFuture<Void> retryFuture = cancellableFuture.thenCompose(...);
```
and cancel the `cancellableFuture`.

core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java
belaban: @rvansa No, that's not the way it works. The async invocation handler was specifically written to give the application a chance to reorder messages if needed. For example, a COMMIT or ROLLBACK message may get executed ahead of other messages in the (app) queue. AFAIR, the code Pedro wrote to handle messages is doing this...

rvansa: @belaban I know it's not the way it works *now*. I got that you don't want to consider this as an option - and TBH it makes more sense ordering any RPC per-key directly in Infinispan, rather than global ordering within JGroups. It's just that some parts in Infinispan still rely on ordering by JGroups. I silently hope that when Dan finishes his _opus magnum interceptorae_ the async modes will be removed, too.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
danberindei: I'm also thinking of adding a `cancel` method to `InvocationStage`, because `goAsync`/`returnWithAsync` kind of have to create a new `CompletableFuture` anyway. But until that's in master, I'd go with the inner class.

core/src/main/java/org/infinispan/statetransfer/StateTransferLock.java
danberindei: Fair enough. I was thinking that you already had the methods in `StateTransferLockImpl`, but of course the code is quite different.

core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java
danberindei: @rvansa Per-key ordering would require us to account for the fact that the primary owner can change, and a backup can receive interleaving messages from different primaries. Relying on JGroups' per-sender ordering allows us to pretend that's not a problem :)

rvansa: @danberindei How does the per-sender ordering keep us away from the problem? If it was causal ordering, I could see that, but per-sender?

I don't see a reason to use synchronous API with `DIST_ASYNC` when the asynchronous API can do the trick efficiently (without context switch and concurrency limitations through delegating the work to limited pool). On the other hand, I can see a lot of space for misuse, when coder's mind set considers only sync mode and administrator turns on `DIST_ASYNC` because it's more performant and 'seems to work'.
I know there's additional perf difference by not requiring the response, but I would rather see that as a flag. But let's keep this for infinispan-dev.

rvansa:
@danberindei Rebased & updated.

core/src/main/java/org/infinispan/commands/CreateCacheCommand.java
```diff
@@ -4,6 +4,7 @@
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
```
slaskawi: Changes in this file are not needed. Please remove them to avoid noise.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
```diff
@@ -125,9 +135,74 @@ protected final void updateTopologyId(TopologyAffectedCommand command) throws In
...
+      private static final AtomicIntegerFieldUpdater<CancellableRetry> cancellableRetryUpdater
+            = AtomicIntegerFieldUpdater.newUpdater(CancellableRetry.class, "cancelled");
+      private static final Log log = LogFactory.getLog(CancellableRetry.class);
+
+      private final T command;
+      private final int topologyId;
+      private final RetryHandler<T> handler;
+      private final InvocationContext ctx;
+      @SuppressWarnings("unused")
+      private volatile int cancelled = 0; // there's no AtomicBooleanFieldUpdater
```
slaskawi: Hoe about using `AtomicInteger`? 0 - not canceled, 1 canceled. Just like boolean...

core/src/main/java/org/infinispan/commands/CreateCacheCommand.java
rvansa: These changes are needed; see that the TimeoutException caught here is a different class.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
rvansa: Allocating another instance is what I am trying to avoid. That's what atomic updaters are good for.

core/src/main/java/org/infinispan/commands/CreateCacheCommand.java
slaskawi: ooh yes... sorry missed that.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
slaskawi: Fair enough.

slaskawi:
@danberindei Are you satisfied with this PR? It looks good to me so I would like to integrate it...

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
danberindei: Non-broadcast RPCs can also be sent as broadcasts if the cluster has more than 2 nodes and all the nodes are in the recipients list. Say if you have a distributed cache running on 2 nodes and you start a 3rd, as long as the rebalance is in progress you will have some keys with 3 owners and a write to them will be sent as a broadcast.

I don't think it's ok to retry immediately on a SuspectException, we should always wait for something. Here we can check whether `ce.getSuspect()` is still a member of the topology, and retry immediately only if it's not a member. That way, when a node crashes, we won't use 100% of a CPU (or maybe the only CPU!) retrying the same command until we receive the new topology.

Maybe we should also extend this to throw an exception if the current topology id is the same as the command's topology id AND the suspected node is not a member in the topology -- that probably means the command is replicated using `SYNCHRONOUS` when it should be using `SYNCHRONOUS_IGNORE_LEAVERS`. Or maybe I should create a separate JIRA :)

core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java
danberindei: @rvansa To answer your first question, yes, the way I did this in https://github.com/infinispan/infinispan/pull/4552 was to add a `sync` field to `BaseBlockingRunnable`: 
https://github.com/infinispan/infinispan/pull/4552/files#diff-f117a0bc3504f81ca8b7036425f73a6eR28

Now back to the more important issue of `DIST_ASYNC` :)
IMO with https://github.com/infinispan/infinispan/pull/4552, the main perf difference between `DIST_ASYNC` and `putAsync()` will be the fact that the primary releases the lock immediately after sending the command to the backups, so you can have a lot more concurrent writes to the same key. You're right though, the difference will be much smaller than it is now, so perhaps we *should* implement `DIST_ASYNC` as "replace all calls to put() with putAsync()" in 9.0. Perhaps I should start thread on the dev list :)

core/src/main/java/org/infinispan/remoting/inboundhandler/DefaultTopologyRunnable.java
danberindei: Can you change this comment to talk instead about commands that implement `TopologyAffectedCommand` and yet return `false` from `canBlock()`? Which, like I said before, I think is a bug in those commands ;)

BTW, waiting for tx data in non-tx mode means just waiting for the topology. Can you add a javadoc to `StateTransferLock.transactionDataFuture()` to mention that, instead of the comment here?

core/src/main/java/org/infinispan/remoting/inboundhandler/DefaultTopologyRunnable.java
```diff
@@ -39,19 +43,39 @@ public boolean isReady() {
...
+      CompletableFuture<Void> future;
       switch (topologyMode) {
          case WAIT_TOPOLOGY:
-            handler.getStateTransferLock().waitForTopology(waitTopology(), 1, TimeUnit.DAYS);
+            future = handler.getStateTransferLock().topologyFuture(waitTopology());
+            if (!future.isDone()) {
+               if (forceOrdering) {
+                  future.get(1, TimeUnit.DAYS);
```
danberindei: `CompletableFutures.await()`?

core/src/main/java/org/infinispan/remoting/inboundhandler/BasePerCacheInboundInvocationHandler.java
```diff
@@ -66,6 +69,7 @@ public void injectDependencies(@ComponentName(REMOTE_COMMAND_EXECUTOR) BlockingT
       this.cancellationService = cancellationService;
       this.stateTransferLock = stateTransferLock;
       this.stateTransferManager = stateTransferManager;
+      this.async = !configuration.clustering().cacheMode().isSynchronous();
```
danberindei: Even in asynchronous caches, some commands will be synchronous (and vice versa). You need to check each command's `DeliverOrder` instead.

core/src/main/java/org/infinispan/remoting/rpc/ResponseMode.java
```diff
@@ -11,6 +11,11 @@
  */
 public enum ResponseMode {
    SYNCHRONOUS,
+   /**
+    * Use this when doing a broadcast to all members to prevent SuspectExceptions from nodes that
+    * are leaving - these can be still part of the view but it is not part of the topology and
+    * the cache is not running there.
```
danberindei: This comment is a bit too narrow. First, we use `SYNCHRONOUS_IGNORE_LEAVERS` for anycast RPCs as well, not just for broadcasts. Second, we still support asymmetric clusters, and there the nodes that don't have your cache running will always return a `CacheNotFoundResponse`, which translates into a `SuspectException` with `SYNCHRONOUS`.

core/src/main/java/org/infinispan/statetransfer/StateProviderImpl.java
```diff
@@ -207,7 +211,11 @@ private CacheTopology getCacheTopology(int requestTopologyId, Address destinatio
                   "topology (%d). Waiting for topology %d to be installed locally.", isReqForTransactions ? "Transactions" : "Segments", destination,
                   requestTopologyId, currentTopologyId, requestTopologyId);
          }
-         stateTransferLock.waitForTopology(requestTopologyId, timeout, TimeUnit.MILLISECONDS);
+         try {
+            stateTransferLock.waitForTopology(requestTopologyId, timeout, TimeUnit.MILLISECONDS);
+         } catch (TimeoutException e) {
+            throw new CacheException("Failed waiting for topology " + requestTopologyId, e);
```
danberindei: Please add a log message to return an `o.i.u.c.TimeouException`, and use that everywhere `waitForTopology`/`topologyFuture` is called.

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
```diff
@@ -385,10 +398,10 @@ private boolean isTxCommandAsync(TransactionBoundaryCommand command) {
          int commandTopologyId = writeCommand.getTopologyId();
          int newTopologyId = Math.max(currentTopologyId, commandTopologyId + 1);
          writeCommand.setTopologyId(newTopologyId);
-         waitForTransactionData(newTopologyId);
-
          writeCommand.addFlag(Flag.COMMAND_RETRY);
-         return handleNonTxWriteCommand(rCtx, writeCommand);
+         // TODO: is it correct to wait for transaction data in non-tx command?
```
danberindei: Yes, it is correct, and it is equivalent to waiting for the topology.

core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java
rvansa: @danberindei Aha, I meant rather per-sender + per-key (than just per-key) ordering, but not blocking writes to keyA because update for keyB is delayed. But understood.

If releasing the lock immediately and using ordered messages is more performant, you could already adopt that for `DIST_SYNC` and use the RPC ack only to notify the originator about completion. But that could still have technical edge cases, and `DIST_ASYNC` is already expected to be less reliable.
I personally think that too much contention should not happen in practice, so optimizing that is low-priority. On the other hand, with per-sender ordering I am worried about global hiccups when a message is lost/dropped.

danberindei: > If releasing the lock immediately and using ordered messages is more performant, you could already adopt that for DIST_SYNC and use the RPC ack only to notify the originator about completion. 

Pedro's triangle work sends the ACK from the backup directly to the originator, but the primary also waits for an ACK before releasing the lock. I'm still worried about the edge cases, though :)

> I personally think that too much contention should not happen in practice, so optimizing that is low-priority.

I tend to agree, but contention might be more of a problem with the counters that Pedro is also working on. I expect a small set of counters, and a lot of writes to them.

> On the other hand, with per-sender ordering I am worried about global hiccups when a message is lost/dropped.

I'm not *too* worried. When a node receives a message out of order, it immediately sends a XMIT request to the sender, so the delay shouldn't be much more than 1 RTT.  When there's lots of allocation (or even blocking) during command processing, the receiver might take a long time to catch up, but I don't think that's a problem ATM.

rvansa:
@danberindei Addressed comments. Regarding the SuspectException, I've let it to wait all the time and throw an exception in the case you describe; hopefully I'll hit it again and will isolate a test case for that.

rvansa:
Rebased (or rather completely rewritten on the interceptor side).

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
```diff
@@ -180,30 +181,52 @@ private InvocationStage visitReadCommand(InvocationContext ctx, AbstractTopology
...
       CacheTopology beginTopology = stateTransferManager.getCacheTopology();
       consistentHashUpdater.accept(beginTopology.getReadConsistentHash());
-      updateTopologyId(command);
-      return invokeNext(ctx, command).compose((stage, rCtx, rCommand, rv, t) -> {
-         if (t == null)
-            return stage;
+      if (command.getTopologyId() == -1) {
+         command.setTopologyId(beginTopology.getTopologyId());
+      }
+      return invokeNext(ctx, command).compose((stage, rCtx, rCommand, rv, t) -> handleReadCommandReturn(stage, rCtx, rCommand, rv, t, consistentHashUpdater));
```
danberindei: Split into 2 lines?

core/src/main/java/org/infinispan/xsite/statetransfer/XSiteStateProviderImpl.java
```diff
@@ -228,7 +224,7 @@ public void run() {
                log.debugf("[X-Site State Transfer - %s] wait for min topology %s", xSiteBackup.getSiteName(), minTopologyId);
             }
 
-            waitForTopology(minTopologyId);
+            stateTransferLock.waitForTopology(minTopologyId, 1, TimeUnit.DAYS);
```
danberindei: `CompletableFutures.await(CompletableFuture)`

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
```diff
@@ -32,53 +46,37 @@
...
-               return retryVisitGetKeysInGroupCommand(ctx, command, commandTopologyId);
-            }
-            // No retry, either rethrow the exception or return the current result
-            if (t != null) {
-               throw t;
-            } else {
-               return returnWith(rv);
-            }
-         });
+         return invokeNext(ctx, command).compose(this::handleGeyKeysInGroupReturn);
```
rvansa: Why? I don't want to check that when OOTE is thrown.

rvansa:
@danberindei Updated.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
danberindei: Yeah, looking at it again it may make things more complicated. How about renaming it `handleLocalGeyKeysInGroupReturn` instead?

rvansa: Okay, now I see I have a typo in the method anyway...

danberindei: I copied it without noticing :)

rvansa:
@danberindei Updated.

danberindei:
Integrated, thanks Radim!

PR #4462, opened 2016-07-19
ISPN-6868 Expose extra configuration for the SSLContext protocol in t…

gustavonalle:
…he Hot Rod client

https://issues.jboss.org/browse/ISPN-6868

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/SslContextFactory.java
+++ b/commons/src/main/java/org/infinispan/commons/util/SslContextFactory.java
@@ -24,7 +24,7 @@ import java.security.KeyStore;
  */
 public class SslContextFactory {
    private static final Log log = LogFactory.getLog(SslContextFactory.class);
-   public static final String DEFAULT_SSL_PROTOCOL = "TLS";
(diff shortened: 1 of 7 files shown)


commons/src/main/java/org/infinispan/commons/util/SslContextFactory.java
```diff
@@ -24,12 +24,21 @@
  */
 public class SslContextFactory {
    private static final Log log = LogFactory.getLog(SslContextFactory.class);
+   public static final String DEFAULT_SSL_PROTOCOL = "TLS";
```
tristantarrant: I think it would be best to increase this to TLSv1.2

gustavonalle: Sure, I was with Java 1.6 in my mind when creating this PR  :smile:

gustavonalle:
@tristantarrant fixed

slaskawi:
The code looks good to me but could you please mention somewhere in the docs what it does and when to use it? Just a few words...

tristantarrant:
I have pushed it already, so the docs will need to be in a separate PR

PR #4463, opened 2016-07-19
ISPN-6242 Migrate to CDI 1.2 with annotation bean discovery

slaskawi:
https://issues.jboss.org/browse/ISPN-6242

### DIFF
--- a/integrationtests/as-integration-embedded/src/test/resources/beans.xml
+++ b/integrationtests/as-integration-embedded/src/test/resources/beans.xml
@@ -1,11 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<beans xmlns="http://java.sun.com/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-   xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/beans_1_0.xsd">
+<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd"
+       version="1.2" bean-discovery-mode="all">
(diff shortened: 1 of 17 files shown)


slaskawi:
Hey @mkouba ! May I ask you for review?

integrationtests/as-integration-embedded/src/test/resources/beans.xml
```diff
@@ -1,11 +1,13 @@
...
+       version="1.2" bean-discovery-mode="all">
 
-   <interceptors>
-      <class>org.infinispan.jcache.annotation.InjectedCacheResultInterceptor</class>
-      <class>org.infinispan.jcache.annotation.InjectedCachePutInterceptor</class>
-      <class>org.infinispan.jcache.annotation.InjectedCacheRemoveEntryInterceptor</class>
-      <class>org.infinispan.jcache.annotation.InjectedCacheRemoveAllInterceptor</class>
-   </interceptors>
+    <interceptors>
+        <class>org.infinispan.jcache.annotation.InjectedCacheResultInterceptor</class>
```
mkouba: Have you considered enabling these interceptors globally (with `@Priority`)? I don't know the use case but it usually makes sense and simplifies usage.

mkouba:
Looks good. But honestly I don't remember the details of `AdvancedCacheProducer`, `InfinispanExtensionEmbedded`, etc... So I cannot say anything about the "business logic".

integrationtests/as-integration-embedded/src/test/resources/beans.xml
slaskawi: Thanks for the hint! I'll experiment with it.

slaskawi: Unfortunately it looks like those Interceptors can not be enabled by default. Basically there are 2 groups of them (`injected*` and standard ones). Those injected should be used when deploying an app on Application Server and they prevent JCache from closing resources it does not own (e.g. WF `CacheManager`s).

Here is a reference to the fix: https://github.com/infinispan/infinispan/pull/2232

slaskawi:
Thanks for review @mkouba !

tristantarrant:
Pushed to master, thanks

PR #4465, opened 2016-07-21
ISPN-6882 Fix Scala warnings during REST server compilation

vjuranek:
https://issues.jboss.org/browse/ISPN-6882


slaskawi:
The same here - the code is fine... I'll assign this to @wburns

galderz:
Looks good, integrated.

PR #4469, opened 2016-07-26
ISPN-6227 - ClusteredScriptingTest.testDistributedMapReduceStreamWith…

andyuk1986:
…Flag fails on Windows

https://issues.jboss.org/browse/ISPN-6627


tristantarrant:
Pushed after a minor change to extract the expected number of words as a constant. Thanks

PR #4472, opened 2016-07-29
ISPN-6911 Add MANUAL eviction strategy

tristantarrant:
https://issues.jboss.org/browse/ISPN-6911


tristantarrant:
@rhusar see if this works for your purposes

rvansa:
Almost sounds like manual eviction with NONE set as strategy should emit a warning, to have the behviour symmetric.

rhusar:
@tristantarrant thanks Tristan, this indeed does solve the problem. 

Although feels like an overkill to introduce a new strategy only equivalent to NONE just to suppress a warning, yet I don't have other elegant solution (apart from adding another attribute and doing something along the lines EvictionConfigurationBuilder#supressPassivationWithoutEvictionWarningMessage(true));)

tristantarrant:
@rvansa I've added a warning for this now

galderz:
Looks good, integrating

galderz:
Integrated, thx @tristantarrant

PR #4478, opened 2016-08-02
ISPN-6912 clustered-indexing.xml example configuration file in server…

mgencur:
… is missing an endpoint

Indexing is useful for remote querying so the HotRod endpoint should be there.

https://issues.jboss.org/browse/ISPN-6912


galderz:
Integrated, thx @mgencur

PR #4481, opened 2016-08-03
[8.2.x] ISPN-6907: Repl/Dist cache with Transactions doesn't respect shared s…

ryanemerson:
…tore semantics


tristantarrant:
Pushed to 8.2.x. Thanks

PR #4484, opened 2016-08-03
[8.2.x] ISPN-6923 HotRod bulk commands that provide a limit need to close the

wburns:
iterator


tristantarrant:
Pushed to 8.2.x. Thanks

PR #4486, opened 2016-08-04
Fixes in RPC handling

rvansa:
https://issues.jboss.org/browse/ISPN-6925
https://issues.jboss.org/browse/ISPN-6795

The third commit does not have a JIRA as it is a fix of consequences of ISPN-6925.

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
@@ -39,6 +39,7 @@ import org.infinispan.factories.annotations.Inject;
 import org.infinispan.factories.annotations.Start;
 import org.infinispan.interceptors.impl.ClusteringInterceptor;
 import org.infinispan.interceptors.locking.ClusteringDependentLogic;
+import org.infinispan.remoting.RpcException;
 import org.infinispan.remoting.inboundhandler.DeliverOrder;
 import org.infinispan.remoting.responses.CacheNotFoundResponse;
 import org.infinispan.remoting.responses.ClusteredGetResponseValidityFilter;
@@ -217,9 +218,11 @@ public abstract class BaseDistributionInterceptor extends ClusteringInterceptor 
       ResponseFilter filter = new ClusteredGetResponseValidityFilter(targets, rpcManager.getAddress());
       RpcOptions options = rpcOptionsBuilder.responseFilter(filter).build();
       return rpcManager.invokeRemotelyAsync(targets, get, options).thenApply(responses -> {
+         boolean hasSuccesfulResponse = false;
          if (!responses.isEmpty()) {
             for (Response r : responses.values()) {
                if (r instanceof SuccessfulResponse) {
+                  hasSuccesfulResponse = true;
                   // The response value might be null.
                   SuccessfulResponse response = (SuccessfulResponse) r;
                   Object responseValue = response.getResponseValue();
@@ -239,7 +242,11 @@ public abstract class BaseDistributionInterceptor extends ClusteringInterceptor 
          if (rvrl != null) {
             rvrl.remoteValueNotFound(key);
          }
-         return null;
+         if (!hasSuccesfulResponse) {
+            throw new RpcException("No successful response: " + responses);
+         } else {
+            return null;
+         }
       });
    }
 
--- a/core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
+++ b/core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
@@ -2,6 +2,7 @@ package org.infinispan.partitionhandling.impl;
 
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
@@ -31,14 +32,16 @@ import org.infinispan.factories.annotations.Inject;
 import org.infinispan.interceptors.DDAsyncInterceptor;
 import org.infinispan.partitionhandling.AvailabilityMode;
 import org.infinispan.remoting.RpcException;
+import org.infinispan.remoting.transport.Address;
 import org.infinispan.remoting.transport.Transport;
+import org.infinispan.util.concurrent.TimeoutException;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
 public class PartitionHandlingInterceptor extends DDAsyncInterceptor {
    private static final Log log = LogFactory.getLog(PartitionHandlingInterceptor.class);
 
-   PartitionHandlingManager partitionHandlingManager;
+   private PartitionHandlingManager partitionHandlingManager;
    private Transport transport;
    private DistributionManager distributionManager;
 
@@ -129,20 +132,20 @@ public class PartitionHandlingInterceptor extends DDAsyncInterceptor {
    }
 
    private CompletableFuture<Void> handleDataReadCommand(InvocationContext ctx, DataCommand command) {
-      Object key = command.getKey();
       return ctx.onReturn((rCtx, rCommand, rv, throwable) -> {
+         DataCommand dataCommand = (DataCommand) rCommand;
          if (throwable != null) {
-            if (throwable instanceof RpcException && performPartitionCheck(rCtx, ((DataCommand) rCommand))) {
+            if (throwable instanceof RpcException && performPartitionCheck(rCtx, dataCommand)) {
                // We must have received an AvailabilityException from one of the owners.
                // There is no way to verify the cause here, but there isn't any other way to get an invalid
                // get response.
-               throw log.degradedModeKeyUnavailable(key);
+               throw log.degradedModeKeyUnavailable(dataCommand.getKey());
             } else {
                throw throwable;
             }
          }
 
-         postOperationPartitionCheck(rCtx, ((DataCommand) rCommand), key, rv);
+         postOperationPartitionCheck(rCtx, dataCommand, dataCommand.getKey());
          return null;
       });
    }
@@ -181,25 +184,13 @@ public class PartitionHandlingInterceptor extends DDAsyncInterceptor {
       }
    }
 
-   private Object postOperationPartitionCheck(InvocationContext ctx, DataCommand command, Object key, Object result) throws Throwable {
+   private void postOperationPartitionCheck(InvocationContext ctx, DataCommand command, Object key) throws Throwable {
       if (performPartitionCheck(ctx, command)) {
          // We do the availability check after the read, because the cache may have entered degraded mode
          // while we were reading from a remote node.
          partitionHandlingManager.checkRead(key);
-
-         // If all owners left and we still haven't received the availability update yet, we could return
-         // an incorrect null value. So we need a special check for null results.
-         if (result == null) {
-            // Unlike in PartitionHandlingManager.checkRead(), here we ignore the availability status
-            // and we only fail the operation if _all_ owners have left the cluster.
-            // TODO Move this to the availability strategy when implementing ISPN-4624
-            if (!InfinispanCollections.containsAny(transport.getMembers(), distributionManager.locate(key))) {
-               throw log.degradedModeKeyUnavailable(key);
-            }
-         }
       }
       // TODO We can still return a stale value if the other partition stayed active without us and we haven't entered degraded mode yet.
-      return result;
    }
 
    @Override
--- a/core/src/main/java/org/infinispan/remoting/responses/ClusteredGetResponseValidityFilter.java
+++ b/core/src/main/java/org/infinispan/remoting/responses/ClusteredGetResponseValidityFilter.java
@@ -17,12 +17,12 @@ import org.infinispan.remoting.transport.Address;
 public class ClusteredGetResponseValidityFilter implements ResponseFilter {
 
    private Collection<Address> targets;
-   private int validResponses;
+   private int acceptableResponses;
    private int missingResponses;
 
    public ClusteredGetResponseValidityFilter(Collection<Address> targets, Address self) {
       this.targets = new HashSet<Address>(targets);
-      this.validResponses = 0;
+      this.acceptableResponses = 0;
       this.missingResponses = targets.size();
       if (this.targets.contains(self)) {
          this.missingResponses--;
@@ -33,8 +33,8 @@ public class ClusteredGetResponseValidityFilter implements ResponseFilter {
    public boolean isAcceptable(Response response, Address address) {
       if (targets.contains(address)) {
          missingResponses--;
-         if (response instanceof SuccessfulResponse) {
-            validResponses++;
+         if (response instanceof SuccessfulResponse || response instanceof ExceptionResponse) {
+            acceptableResponses++;
             return true;
          }
       }
@@ -43,7 +43,7 @@ public class ClusteredGetResponseValidityFilter implements ResponseFilter {
 
    @Override
    public boolean needMoreResponses() {
-      return validResponses < 1 && missingResponses > 0;
+      return acceptableResponses < 1 && missingResponses > 0;
    }
 
 }
--- a/core/src/main/java/org/infinispan/remoting/transport/jgroups/Responses.java
+++ b/core/src/main/java/org/infinispan/remoting/transport/jgroups/Responses.java
@@ -1,0 +1,110 @@
+package org.infinispan.remoting.transport.jgroups;
+
+import org.infinispan.remoting.responses.Response;
+import org.jgroups.Address;
+import org.jgroups.util.Rsp;
+import org.jgroups.util.RspList;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+
+public class Responses implements Iterable<Rsp<Response>> {
+   public static final Responses EMPTY = new Responses(Collections.emptyList());
+   private final Address[] addresses;
+   private final AtomicReferenceArray<Rsp<Response>> responses;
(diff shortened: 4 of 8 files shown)


core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -243,7 +246,17 @@ public void configure() {
          if (rvrl != null) {
             rvrl.remoteValueNotFound(key);
          }
-         return null;
+         if (!hasSuccesfulResponse) {
+            RpcException rpcException = new RpcException("No successful response: " + responses);
```
rvansa: Actually it's a bit weird that remote node will return SuccessfulResponse{null}, I think that UnsuccessfulResponse would make more sense...

danberindei: Indeed. Actually we used to send an `UnsureResponse` back in such cases (with a check in `DistributionResponseGenerator`). But because `ClusteredGetCommand` doesn't have a topology, the check we did there was almost guaranteed to be false, so I finally removed it with ISPN-5463. 

Besides, it's weird to have a locality check in the response generator, I'd rather remove the `ResponseGenerator` and create the `Response` in the command itself. (Or maybe in the interceptors...)

galderz:
Needs rebasing

core/src/main/java/org/infinispan/remoting/responses/ClusteredGetResponseValidityFilter.java
```diff
@@ -33,7 +33,7 @@ public ClusteredGetResponseValidityFilter(Collection<Address> targets, Address s
    public boolean isAcceptable(Response response, Address address) {
       if (targets.contains(address)) {
          missingResponses--;
-         if (response instanceof SuccessfulResponse) {
+         if (response instanceof SuccessfulResponse || response instanceof ExceptionResponse) {
             validResponses++;
```
danberindei: `ExceptionResponse` is not a `ValidResponse`, so the `validResponses` field name is no longer accurate.

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
```diff
@@ -129,20 +132,31 @@ private boolean performPartitionCheck(InvocationContext ctx, LocalFlagAffectedCo
...
+            if (throwable instanceof RpcException && performPartitionCheck(rCtx, dataCommand)) {
                // We must have received an AvailabilityException from one of the owners.
                // There is no way to verify the cause here, but there isn't any other way to get an invalid
                // get response.
-               throw log.degradedModeKeyUnavailable(key);
+               throw log.degradedModeKeyUnavailable(dataCommand.getKey());
             } else {
+               // If all owners left and we still haven't received the availability update yet,
+               // we get TimeoutException from JGroupsTransport.checkRsps
+               if (throwable instanceof TimeoutException && performPartitionCheck(rCtx, dataCommand)) {
```
danberindei: Shouldn't the `TimeoutException` be eliminated by the previous commit?

If instead this is a real timeout, I don't think we should change the exception - availability exceptions should be immediate, not after a 15s timeout.

danberindei:
@rvansa Are you sure the 3rd commit is needed because of the ISPN-6925 fix and not because of the ISPN-6795 fix? To me, it seems like the ISPN-6925 fix should be sufficient.

And this PR could really use a test...

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
rvansa: Hmm, when I remove the if, the tests still work. The commits were cherry-picked from the scattered cache work, so it may have changed.

I'll remove that for now and if the problem pops in scattered cache, I'll deal with it. And you're right that the exception should be thrown immediately.

rvansa:
@danberindei Hi, I've added the tests.

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
```diff
@@ -0,0 +1,240 @@
...
+
+@Test(groups = "functional")
+@CleanupAfterMethod
+public class RemoteGetFailureTest extends MultipleCacheManagersTest {
+   private Object key;
+
+   @Override
+   protected void createCacheManagers() throws Throwable {
+      ConfigurationBuilder builder = getDefaultClusteredCacheConfig(CacheMode.DIST_SYNC);
+      // cache stop takes quite long when the view splits
```
danberindei: Could it wait for remote txs to finish, because of `cacheStopTimeout`?

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
```diff
@@ -0,0 +1,240 @@
...
+      createCluster(builder, 3);
+      waitForClusterToForm();
+      key = getKeyForCache(cache(1), cache(2));
+   }
+
+   @AfterMethod(alwaysRun = true)
+   @Override
+   protected void clearContent() throws Throwable {
+      // do not clear the caches
+      TestingUtil.killCacheManagers(false, cacheManagers.toArray(new EmbeddedCacheManager[cacheManagers.size()]));
```
danberindei: To kill caches between test methods, you can use `@CleanupAfterMethod`, or `this.cleanup = CleanupPhase.AFTER_METHOD`

core/src/main/java/org/infinispan/remoting/transport/jgroups/Responses.java
```diff
@@ -0,0 +1,102 @@
...
+   private final Address[] addresses;
+   private final AtomicReferenceArray<Rsp<Response>> responses;
+
+   public Responses(Collection<Address> addresses) {
+      int size = addresses.size();
+      this.addresses = addresses.toArray(new Address[size]);
+      this.responses = new AtomicReferenceArray<>(size);
+   }
+
+   public Responses(List<Address> dests, boolean suspected) {
```
danberindei: This would look better as a factory method instead of a constructor.

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
```diff
@@ -0,0 +1,240 @@
+package org.infinispan.distribution;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
```
danberindei: We prefer using `AssertJUnit` in TestNG tests.

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
```diff
@@ -0,0 +1,240 @@
...
+      CountDownLatch release = new CountDownLatch(1);
+      cache(1).getAdvancedCache().getAsyncInterceptorChain().addInterceptor(new DelayingInterceptor(arrival, release), 0);
+      cache(2).getAdvancedCache().getAsyncInterceptorChain().addInterceptor(new DelayingInterceptor(arrival, release), 0);
+
+      Future<?> future = fork(() -> {
+         try {
+            cache(0).get(key);
+            fail("Should have thrown");
+         } catch (RpcException e) {
+            // expected
```
danberindei: You're missing out on the wonderful `Exceptions` class, which allows you to write

      Future<?> future = fork(() -> cache(0).get(key));
      Exceptions.expectExecutionException(RpcException.class, future);

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
```diff
@@ -129,20 +132,20 @@ private boolean performPartitionCheck(InvocationContext ctx, LocalFlagAffectedCo
...
       return ctx.onReturn((rCtx, rCommand, rv, throwable) -> {
+         DataCommand dataCommand = (DataCommand) rCommand;
          if (throwable != null) {
-            if (throwable instanceof RpcException && performPartitionCheck(rCtx, ((DataCommand) rCommand))) {
+            if (throwable instanceof RpcException && performPartitionCheck(rCtx, dataCommand)) {
                // We must have received an AvailabilityException from one of the owners.
                // There is no way to verify the cause here, but there isn't any other way to get an invalid
                // get response.
-               throw log.degradedModeKeyUnavailable(key);
+               throw log.degradedModeKeyUnavailable(dataCommand.getKey());
```
danberindei: Hmmm, reading these comments, I see we do have a legitimate use case for a remote node throwing an exception during a clustered get. 

It also looks like throwing an exception on the first `AvailabilityException` will change our behaviour when one of the owners is `AVAILABLE` and the other is `DEGRADED`. I think that could happen if a partition is `DEGRADED` mode but one of the nodes can still send responses back to the other partition -- either because the split is still allowing some messages to go through, or because the partitions are merging but the availability mode hasn't been changed to `AVAILABLE` everywhere yet. I'm not yet sure whether I should worry about these...

core/src/main/java/org/infinispan/remoting/rpc/RpcManagerImpl.java
```diff
@@ -133,6 +133,7 @@ public String getPendingViewAsString() {
          return rethrowAsCacheException(e);
       }
       return invocation.handle((responseMap, throwable) -> {
+         ReplicableCommand cmd = rpc;
```
danberindei: Unused

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -243,7 +246,17 @@ public void configure() {
...
+            for (Response r : responses.values()) {
+               if (r instanceof ExceptionResponse) {
+                  rpcException.addSuppressed(((ExceptionResponse) r).getException());
+               }
+            }
+            throw rpcException;
```
danberindei: This check should be done in `JGroupsTransport`, like we call `checkRsp` for the responses of synchronous RPCs.

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
rvansa: It's non-tx cache. This really waits for ST.

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
rvansa: `@CleanupAfterMethod` is already used, this override just disables calling `cache.clear()` before `cache.stop()`. See `MultipleCacheManagersTest.clearContent(...)`, that uses `TestingUtil.killCacheManagers(true, ...)`.

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
rvansa: OK, cool.

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
rvansa: One-way communication is not really tested; it's behaves a malicious node and I am not sure if the code is bulletproof to that.
Let's deal with this rather in a separate PR, with proper tests.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: Right, upon exception, the exception is thrown directly from JGroupsTransport so this loop is unnecessary.
On the other hand, Successful/UnsuccessfulResponse should be handled at interceptor-level, so the check for hasSuccessfulReponse makes sense.

rvansa:
@danberindei Rebased & addressed.

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
danberindei: Is there any reason in particular to avoid the clear? If so, please add a comment.

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
danberindei: True, asymmetric partitioning and even the behaviour of cache operations during a merge aren't tested in depth. That's exactly why I'm hesitating so much about the `ClusteredGetValidityFilter` changes...

AFAICT with this PR a node can become `AVAILABLE`, read a key from one node, but then fail to read a remote key because the primary owner is still `DEGRADED`. Without this PR, the 2nd read would only fail if all the owners of the key are `DEGRADED`, but it would still fail in some situations, because `ClusteredGetCommand` doesn't have a topology id and so it can't block until the remote nodes become available. In the end, the difference is quite small, but this PR is supposed to "fix RPC handling", so I wouldn't include a change that makes things (ever so slightly) worse.

rvansa: The changes in this PR are removing a special check when the result is `null`, because this is a valid value; "If all owners left" (as per the comment), the command should throw an exception, but never return `null` (when a null value was not read on proper owner). So I believe that the condition below cannot be true (without catching the `RpcException` earier on).
When the check has been add, it was a non-systematic workaround that fixed the problem on a wrong place.

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
rvansa: Added a comment...

slaskawi:
LGTM, @danberindei May I integrate it?

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
danberindei: Ok, I guess that's another bug then :)

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
danberindei: Sorry, I wrote this thinking about ISPN-6795 and I without checking `ClusteredGetValidityFilter` again. Looks good, maybe just squash the commits a bit?

core/src/main/java/org/infinispan/remoting/responses/ClusteredGetResponseValidityFilter.java
```diff
@@ -33,8 +33,8 @@ public ClusteredGetResponseValidityFilter(Collection<Address> targets, Address s
    public boolean isAcceptable(Response response, Address address) {
       if (targets.contains(address)) {
          missingResponses--;
-         if (response instanceof SuccessfulResponse) {
-            validResponses++;
+         if (response instanceof SuccessfulResponse || response instanceof ExceptionResponse) {
```
rvansa: @pruivo @tristantarrant @galderz @wburns Hi, so we have a stalemate with @danberindei we'd like to decide:

When a remote get is executed (to say, 2 owners), and first owner responds with an exception, there are two things we can do:
1. ignore the exception (swallow it) and wait for the second response (possibly without exception) - that's what master does
2. report the exception immediately (suggested change)

An exception can be thrown from two reasons:
a) bug in Infinispan code
b) AvailabilityException or similar

Note: AE can be thrown on one owner and not on another under certain timing; it should not happen in steady state

Pros:
- fail fast, both during development (for a)-style exceptions), and for b) as the user code is likely to hit AE more often (so the code will deal with it properly). Swallowing exceptions hides problems.
- when all owners would throw an exception, we won't get just RpcException without any other info (until this PR, it would end up with timeout, other fixes in the PR at least throw RpcException when there's no acceptable response)
- don't do another RPC when the first one already failed - just hoping for the best when we're likely to get the same response

Cons:
- the value can be available, and could be of use to the user (rather than the exception)
- it is a change of behaviour (in a situation when the exception can happen with different timing, though)

slaskawi: Since we are talking about Get commands, I vote for option 1 (ignoring exception and waiting for second response).

tristantarrant: What about OutdatedTopologyExceptions ? (still not a steady state)

rvansa:
Added one more commit, fixing an issue when a request times out but TimeoutException is not thrown.
The test included in the PR would fail (without the fix) with RpcException.

rvansa:
Rebased to master.

core/src/main/java/org/infinispan/remoting/responses/ClusteredGetResponseValidityFilter.java
rvansa: Other opinions?

tristantarrant: Okay, my vote goes to 1. I guess we'd want the opinions of @wburns and @pruivo

rvansa: Since others' opinion seems to be `null`, I'll take the result as 1).

danberindei: @tristantarrant Turns out our current behaviour is actually 2), with the caveat that we delay throwing the exception until we get a response from all the owners. We started behaving like 1) after the staggered gets work, but because of a bug that happened only when staggered gets were actually disabled. 

Turns out supporting 1) and reporting meaningful exceptions is also pretty awkward with the `ResponseFilter` API, so I agreed with @rvansa to converge to 2) everywhere, at least until we implement our own response handling on top of `JChannel`.

danberindei:
Integrated, thanks Radim!

PR #4487, opened 2016-08-04
Further improvements in testsuite

rvansa:


tristantarrant:
Thanks @rvansa, pushed to master

core/src/test/java/org/infinispan/distribution/MagicKey.java
```diff
@@ -24,57 +33,99 @@
...
+      int segment = findSegment(ch.getNumSegments(), s -> primaryAddress.equals(ch.locatePrimaryOwnerForSegment(s)));
+      if (segment < 0) {
+         throw new IllegalStateException("Could not find any segment owned by " + primaryOwner +
+            ", primary segments: " + segments(primaryOwner));
       }
-      // we have found a hashcode that works!
-      hashcode = dummy.hashCode();
-      segment = primaryOwner.getAdvancedCache().getDistributionManager().getReadConsistentHash().getSegment(this);
+      this.segment = segment;
+      hashcode = getHashCodeForSegment(ch, segment);
```
danberindei: @rvansa When you have 256 segments (the default), it's much cheaper to find a hash code that maps to one node out of 4 (approx 1/4 of keys will match), than it is to find a hash code that maps to a certain segment (only 1/256 will match). So I'd rather keep trying to find a key with a particular primary owner, and have `findSegment` return a set of segments.

rvansa: Hmm, it could be changed to `Collection<Integer> findSegments(...)` and `getHashCodeForSegments(ConsistentHash ch, Collection<Integer> segment)` but the latter function would then return first cached hashcode mapping to one of the segments - and that would result in all keys for given owner ending up in the same segment.

danberindei: True, but we don't really use more than 10 MagicKeys per test - as I mentioned in the JIRA, when you need the hundreds of keys that would make the caching pay off, it's better to use sequential keys and to map them to nodes with a custom `KeyPartitioner`/`ConsistentHashFactory`.

PR #4491, opened 2016-08-05
ISPN-6930 ERROR logged when consuming events with SSL

gustavonalle:
https://issues.jboss.org/browse/ISPN-6930


tristantarrant:
Pushed to master, thanks

PR #4493, opened 2016-08-09
Enhancements for Functional Map API

rvansa:
Depends on #4486

### DIFF
--- a/core/src/main/java/org/infinispan/commands/CommandsFactory.java
+++ b/core/src/main/java/org/infinispan/commands/CommandsFactory.java
@@ -467,7 +467,7 @@ public interface CommandsFactory {
 
    <K, V, R> ReadOnlyKeyCommand<K, V, R> buildReadOnlyKeyCommand(K key, Function<ReadEntryView<K, V>, R> f);
 
-   <K, V, R> ReadOnlyManyCommand<K, V, R> buildReadOnlyManyCommand(Set<? extends K> keys, Function<ReadEntryView<K, V>, R> f);
+   <K, V, R> ReadOnlyManyCommand<K, V, R> buildReadOnlyManyCommand(Collection<? extends K> keys, Function<ReadEntryView<K, V>, R> f);
 
    <K, V> WriteOnlyKeyCommand<K, V> buildWriteOnlyKeyCommand(
       K key, Consumer<WriteEntryView<V>> f, Params params);
@@ -484,9 +484,9 @@ public interface CommandsFactory {
    <K, V> WriteOnlyKeyValueCommand<K, V> buildWriteOnlyKeyValueCommand(
       K key, V value, BiConsumer<V, WriteEntryView<V>> f, Params params);
 
-   <K, V> WriteOnlyManyCommand<K, V> buildWriteOnlyManyCommand(Set<? extends K> keys, Consumer<WriteEntryView<V>> f, Params params);
+   <K, V> WriteOnlyManyCommand<K, V> buildWriteOnlyManyCommand(Collection<? extends K> keys, Consumer<WriteEntryView<V>> f, Params params);
 
-   <K, V, R> ReadWriteManyCommand<K, V, R> buildReadWriteManyCommand(Set<? extends K> keys, Function<ReadWriteEntryView<K, V>, R> f, Params params);
+   <K, V, R> ReadWriteManyCommand<K, V, R> buildReadWriteManyCommand(Collection<? extends K> keys, Function<ReadWriteEntryView<K, V>, R> f, Params params);
 
    <K, V, R> ReadWriteManyEntriesCommand<K, V, R> buildReadWriteManyEntriesCommand(Map<? extends K, ? extends V> entries, BiFunction<V, ReadWriteEntryView<K, V>, R> f, Params params);
 
--- a/core/src/main/java/org/infinispan/commands/functional/ReadOnlyKeyCommand.java
+++ b/core/src/main/java/org/infinispan/commands/functional/ReadOnlyKeyCommand.java
@@ -18,6 +18,7 @@ import org.infinispan.functional.impl.EntryViews;
 
 public final class ReadOnlyKeyCommand<K, V, R> extends AbstractDataCommand implements LocalCommand {
 
+   public static final int COMMAND_ID = 62;
    private Function<ReadEntryView<K, V>, R> f;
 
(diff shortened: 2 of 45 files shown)


core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
```diff
@@ -420,7 +427,8 @@ private void wrapEntryForReplaceIfNeeded(InvocationContext ctx, ReplaceCommand c
       // locally, but if there's no need to get remote, the read-only
       // function needs to be executed, so force a non-null entry in
       // context with null content
-      if (entry == null && cdl.localNodeIsOwner(command.getKey())) {
+      // TODO: move this logic to entryFactory
+      if (entry == null && shouldRead(command.getKey())) {
          entryFactory.wrapEntryForReading(ctx, command.getKey(), NullCacheEntry.getInstance());
       }
```
rvansa: Err, I need to remove the `ctx.onReturn(commitEntriesReturnHandler);` - as entries are not fetched from remote nodes, there's nothing to commit.

rvansa:
Let me know if unwrapping the RemoteExceptions is really desired - to me it seems so at least in the functional case (but we want to be consistent) as the exceptions may signalize something, but I may be wrong.

rvansa:
I am likely about to rewrite ReadOnlyManyCommand once more as part of ISPN-6938 (it showed up that the changes span quite wider than I initially expected), but nevertheless this is still a step forward.

galderz:
@rvansa Thanks a lot for the PR! Needs rebasing but in the mean time I'm having a look :)

core/src/main/java/org/infinispan/commands/CommandsFactory.java
```diff
@@ -460,7 +460,7 @@
 
    <K, V, R> ReadOnlyKeyCommand<K, V, R> buildReadOnlyKeyCommand(K key, Function<ReadEntryView<K, V>, R> f);
 
-   <K, V, R> ReadOnlyManyCommand<K, V, R> buildReadOnlyManyCommand(Set<? extends K> keys, Function<ReadEntryView<K, V>, R> f);
+   <K, V, R> ReadOnlyManyCommand<K, V, R> buildReadOnlyManyCommand(Collection<? extends K> keys, Function<ReadEntryView<K, V>, R> f);
```
galderz: Why not a Set? I'd not expect the same key to be repeated twice in this command.

core/src/main/java/org/infinispan/commands/functional/ReadOnlyKeyCommand.java
```diff
@@ -18,6 +18,7 @@
 
 public final class ReadOnlyKeyCommand<K, V, R> extends AbstractDataCommand implements LocalCommand {
```
galderz: So, read-only functions now shipped around the cluster instead bringing data locally and doing the read locally? Nice :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -390,9 +369,9 @@ private void processCallsStaggered(ReplicableCommand command, RspFilter filter,
    }
 
    private void staggeredProcessNext(ReplicableCommand command, RspFilter filter, List<Address> dests,
-         ResponseMode mode, DeliverOrder deliverOrder, Marshaller marshaller,
-         CompletableFuture<RspList<Response>> theFuture, int destIndex, long deadline,
-         RspList<Response> rsps) {
+                                     ResponseMode mode, DeliverOrder deliverOrder, Marshaller marshaller,
```
galderz: In reformatting mood? ;)

galderz:
If the exception unwrapping is done generically, e.g. for any commands, that could be problematic if there's some reliance on the exceptions thrown. For example, I know Hot Rod server relies on certain exceptions being thrown to send different error codes to client. /cc @danberindei @pruivo 

Why do you want to unwrap them exactly?

galderz:
Some small comments, nothing major.

core/src/main/java/org/infinispan/commands/CommandsFactory.java
rvansa: Eh, I thought I put a comment somewhere, but it got lost...

When you build the collection as an user (and pass it to the functional API), it's useful to mark it as a set to prevent repetitions. However, during the internal processing you sometimes need to split the set according to keys' owners, and it's more convenient and effective to just put those subsets into `ArrayList`s.
Also the Set may use different equality/hash functions than `Object`'s default ones, so if you just create new `HashSet`s, you risk losing some keys as you add them there (if the `Object.equals()` matches for some of them.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
rvansa: Oops, I need to run the automatic formatting after rebase...

rvansa:
@galderz wrt unwrapping: The functional code can be often written in-line, and (as an user) I intuitively expect that I'll get what the lambda throws. User does not care where the data is stored, so a stack of RemoteException and similar is not very useful (maybe for debugging, but not as `try { ... } catch (NoSuchElementException e) { ... }`.
I understand that this is quite a personal preference, so if you consider this unwanted behaviour, I can drop the commit.

rvansa:
After discussion on -dev, I've decided to remove the unwrapping.

rvansa:
Reverted the unwrapping & rebased.

galderz:
Hmmm, seems like build failed?

```
[10:32:27]Step 2/5: Maven (Maven) (9m:09s)
[10:39:32][Step 2/5] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:checkstyle (checkstyle) on project infinispan-core: An error has occurred in Checkstyle report generation.
[10:41:37][Step 2/5] Step Maven (Maven) failed
```

rvansa:
@galderz Rebased and dependent PR merged - could you review pls?

galderz:
All good, integrated. Thanks a lot @rvansa for helping out :)

danberindei:
@rvansa @galderz Don't forget to look at the CI results, guys :)

https://issues.jboss.org/browse/ISPN-7050

danberindei:
@rvansa One more bug for you: https://issues.jboss.org/browse/ISPN-7063

PR #4499, opened 2016-08-10
ISPN-6679 Healthcheck API

slaskawi:
JIRA: https://issues.jboss.org/browse/ISPN-6679
Design: https://github.com/infinispan/infinispan/wiki/Health-check-API

Highlights:
- The API can be accessed using CLI/JMX/REST and also programatically.
- I decided to implement reading logs by my own (instead of using logging subsystem). The implementation is much simpler, shorter and does exactly what I need.
- The WF integration bits are based on Metrics (runtime resources which are read only). I couldn't find any advantage of using Operations for this.

### DIFF
--- a/core/src/main/java/org/infinispan/health/ClusterHealth.java
+++ b/core/src/main/java/org/infinispan/health/ClusterHealth.java
@@ -1,0 +1,29 @@
+package org.infinispan.health;
+
+import java.util.List;
+
+/**
+ * Cluster health information.
+ */
+public interface ClusterHealth {
+
+    /**
+     * Returns total cluster health.
+     */
+    HealthStatus getHealthStatus();
+
+    /**
+     * Returns the name of the cluster.
+     */
+    String getClusterName();
+
+    /**
+     * Returns the number of nodes in the cluster.
+     */
+    int getNumberOfNodes();
+
+    /**
+     * Returns node names.
+     */
+    List<String> getNodeNames();
+}
--- a/core/src/main/java/org/infinispan/health/HealthStatus.java
+++ b/core/src/main/java/org/infinispan/health/HealthStatus.java
@@ -1,0 +1,29 @@
+package org.infinispan.health;
+
+/**
+ * General Health status.
+ *
+ * @author Sebastian Łaskawiec
+ * @since 9.0
+ */
+public enum HealthStatus {
+    /**
+     * Given entity is unhealthy.
+     *
+     * <p>
+     *    An unhealthy status means that a cache is in {@link org.infinispan.partitionhandling.AvailabilityMode#DEGRADED_MODE}.
+     *    Please keep in mind that in the future additional rules might be added to reflect Unhealthy status of the cache.
+     * </p>.
+     */
+    UNHEALTHY,
+
+    /**
+     * Given entity is healthy.
+     */
+    HEALTHY,
+
+    /**
+     * Given entity is healthy but a rebalance is in progress.
+     */
+    REBALANCING
+}
--- a/core/src/main/java/org/infinispan/health/HostInfo.java
+++ b/core/src/main/java/org/infinispan/health/HostInfo.java
@@ -1,0 +1,25 @@
+package org.infinispan.health;
+
+/**
+ * Information about the host.
+ *
+ * @author Sebastian Łaskawiec
+ * @since 9.0
+ */
+public interface HostInfo {
+
+    /**
+     * Returns the number of CPUs installed in the host.
+     */
+    int getNumberOfCpus();
+
+    /**
+     * Gets total memory in KB.
+     */
+    long getTotalMemoryKb();
+
+    /**
+     * Gets free memory in KB.
+     */
+    long getFreeMemoryInKb();
+}
--- a/core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
+++ b/core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
@@ -1,0 +1,54 @@
+package org.infinispan.health.impl;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.infinispan.health.ClusterHealth;
+import org.infinispan.health.HealthStatus;
+import org.infinispan.manager.EmbeddedCacheManager;
+
+public class ClusterHealthImpl implements ClusterHealth {
+
+    private final EmbeddedCacheManager cacheManager;
+
+    public ClusterHealthImpl(EmbeddedCacheManager cacheManager) {
+        this.cacheManager = cacheManager;
+    }
+
+    @Override
+    public HealthStatus getHealthStatus() {
+        Set<HealthStatus> healthStatuses = cacheManager.getCacheNames().stream()
+                .map(cacheName -> cacheManager.getCache(cacheName))
+                .map(cache -> new CacheHealthImpl(cache))
+                .map(cacheHealth -> cacheHealth.getStatus())
+                .collect(Collectors.toSet());
+
+        if (healthStatuses.contains(HealthStatus.UNHEALTHY)) {
+            return HealthStatus.UNHEALTHY;
+        } else if (healthStatuses.contains(HealthStatus.REBALANCING)) {
+            return HealthStatus.REBALANCING;
+        }
+        return HealthStatus.HEALTHY;
+    }
+
+    @Override
+    public String getClusterName() {
+        return cacheManager.getClusterName();
+    }
+
+    @Override
+    public int getNumberOfNodes() {
+        return Optional.ofNullable(cacheManager.getTransport()).map(t -> t.getMembers().size()).orElse(1);
+    }
+
+    @Override
+    public List<String> getNodeNames() {
+        return Optional.ofNullable(cacheManager.getTransport()).map(t -> t.getMembers()).orElse(Collections.emptyList())
+              .stream()
+              .map(member -> member.toString())
+              .collect(Collectors.toList());
+    }
+}
--- a/core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
+++ b/core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
@@ -46,6 +46,10 @@ import org.infinispan.factories.KnownComponentNames;
 import org.infinispan.factories.annotations.SurvivesRestarts;
 import org.infinispan.factories.scopes.Scope;
 import org.infinispan.factories.scopes.Scopes;
+import org.infinispan.health.Health;
+import org.infinispan.health.impl.HealthImpl;
+import org.infinispan.health.impl.jmx.HealthJMXExposerImpl;
+import org.infinispan.health.jmx.HealthJMXExposer;
 import org.infinispan.jmx.CacheJmxRegistration;
 import org.infinispan.jmx.CacheManagerJmxRegistration;
 import org.infinispan.jmx.annotations.DataType;
@@ -134,6 +138,7 @@ public class DefaultCacheManager implements EmbeddedCacheManager {
    private final AuthorizationHelper authzHelper;
    private final DependencyGraph<String> cacheDependencyGraph = new DependencyGraph<>();
    private final CacheContainerStats stats;
+   private final Health health;
    private final ConfigurationManager configurationManager;
 
    @GuardedBy("this")
@@ -231,6 +236,8 @@ public class DefaultCacheManager implements EmbeddedCacheManager {
       this.globalComponentRegistry.registerComponent(cacheDependencyGraph, CACHE_DEPENDENCY_GRAPH, false);
       this.globalComponentRegistry.registerComponent(authzHelper, AuthorizationHelper.class);
       this.stats = new CacheContainerStatsImpl(this);
+      health = new HealthImpl(this);
+      globalComponentRegistry.registerComponent(new HealthJMXExposerImpl(health), HealthJMXExposer.class);
       if (start)
          start();
    }
@@ -304,6 +311,8 @@ public class DefaultCacheManager implements EmbeddedCacheManager {
          globalComponentRegistry.registerComponent(cacheDependencyGraph, CACHE_DEPENDENCY_GRAPH, false);
          authzHelper = new AuthorizationHelper(globalConfiguration.security(), AuditContext.CACHEMANAGER, globalConfiguration.globalJmxStatistics().cacheManagerName());
          stats = new CacheContainerStatsImpl(this);
+         health = new HealthImpl(this);
+         globalComponentRegistry.registerComponent(new HealthJMXExposerImpl(health), HealthJMXExposer.class);
       } catch (CacheConfigurationException ce) {
          throw ce;
       } catch (RuntimeException re) {
@@ -347,6 +356,8 @@ public class DefaultCacheManager implements EmbeddedCacheManager {
       globalComponentRegistry.registerComponent(cacheDependencyGraph, CACHE_DEPENDENCY_GRAPH, false);
       authzHelper = new AuthorizationHelper(globalConfiguration.security(), AuditContext.CACHEMANAGER, globalConfiguration.globalJmxStatistics().cacheManagerName());
       stats = new CacheContainerStatsImpl(this);
+      health = new HealthImpl(this);
+      globalComponentRegistry.registerComponent(new HealthJMXExposerImpl(health), HealthJMXExposer.class);
       if (start)
          start();
    }
@@ -987,6 +998,11 @@ public class DefaultCacheManager implements EmbeddedCacheManager {
    @Override
    public CacheContainerStats getStats() {
       return stats;
+   }
+
+   @Override
+   public Health getHealth() {
+      return health;
    }
 
    @Override
--- a/core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
+++ b/core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
@@ -12,6 +12,7 @@ import org.infinispan.factories.GlobalComponentRegistry;
 import org.infinispan.factories.annotations.SurvivesRestarts;
 import org.infinispan.factories.scopes.Scope;
 import org.infinispan.factories.scopes.Scopes;
+import org.infinispan.health.Health;
 import org.infinispan.lifecycle.ComponentStatus;
 import org.infinispan.notifications.Listenable;
 import org.infinispan.remoting.transport.Address;
@@ -319,5 +320,11 @@ public interface EmbeddedCacheManager extends CacheContainer, Listenable {
       throw new UnsupportedOperationException();
    }
 
-
+   /**
+    * Returns an entry point for a Health Check API.
+    *
+    * @since 9.0
+    * @return Health API for this {@link EmbeddedCacheManager}.
+     */
+   Health getHealth();
 }
--- a/documentation/src/main/asciidoc/server_guide/server_guide.asciidoc
+++ b/documentation/src/main/asciidoc/server_guide/server_guide.asciidoc
@@ -11,3 +11,99 @@ include::management.adoc[]
 include::configuration.adoc[]
 include::security.adoc[]
 
+== Health monitoring
+
+Infinispan server has special endpoints for monitoring cluster health. The API is exposed via:
+
+* Programmatically (using `embeddedCacheManager.getHealth()`)
+* JMX
+* CLI
+* REST (using https://docs.jboss.org/author/display/WFLY10/The+HTTP+management+API[WildFly HTTP Management API])
+
+=== Accessing Health API using JMX
+
+At first you need to connect to the Infinispan Server using JMX (use JConsole or other tool for this).
+Next, navigate to object name `jboss.datagrid-infinispan:type=CacheManager,name="clustered",component=CacheContainerHealth`.
+
+=== Accessing Health API using CLI
+
+The Health API can be also accessed using CLI interface. An example invocation and result is presented below:
+
+[source,bash]
+----
+$ ispn-cli.sh -c "/subsystem=datagrid-infinispan/cache-container=clustered/health=HEALTH:read-resource(include-runtime=true)"
+
+{
+    "outcome" => "success",
+    "result" => {
+        "cache-health" => "HEALTHY",
+        "cluster-health" => ["test"],
+        "cluster-name" => "clustered",
+        "free-memory" => 99958L,
+        "log-tail" => [
+            "2016-08-10 11:54:14,706 INFO  [org.infinispan.server.endpoint] (MSC service thread 1-5) DGENDPT10001: HotRodServer listening on 127.0.0.1:11222",
+            "2016-08-10 11:54:14,706 INFO  [org.infinispan.server.endpoint] (MSC service thread 1-1) DGENDPT10001: MemcachedServer listening on 127.0.0.1:11211",
+            "2016-08-10 11:54:14,785 INFO  [org.jboss.as.clustering.infinispan] (MSC service thread 1-6) DGISPN0001: Started ___protobuf_metadata cache from clustered container",
+            "2016-08-10 11:54:14,800 INFO  [org.jboss.as.clustering.infinispan] (MSC service thread 1-6) DGISPN0001: Started ___script_cache cache from clustered container",
+            "2016-08-10 11:54:15,159 INFO  [org.jboss.as.clustering.infinispan] (MSC service thread 1-5) DGISPN0001: Started ___hotRodTopologyCache cache from clustered container",
+            "2016-08-10 11:54:15,210 INFO  [org.infinispan.rest.NettyRestServer] (MSC service thread 1-6) ISPN012003: REST server starting, listening on 127.0.0.1:8080",
+            "2016-08-10 11:54:15,210 INFO  [org.infinispan.server.endpoint] (MSC service thread 1-6) DGENDPT10002: REST mapped to /rest",
+            "2016-08-10 11:54:15,306 INFO  [org.jboss.as] (Controller Boot Thread) WFLYSRV0060: Http management interface listening on http://127.0.0.1:9990/management",
+            "2016-08-10 11:54:15,307 INFO  [org.jboss.as] (Controller Boot Thread) WFLYSRV0051: Admin console listening on http://127.0.0.1:9990",
+            "2016-08-10 11:54:15,307 INFO  [org.jboss.as] (Controller Boot Thread) WFLYSRV0025: Infinispan Server 9.0.0-SNAPSHOT (WildFly Core 2.2.0.CR9) started in 8681ms - Started 196 of 237 services (121 services are lazy, passive or on-demand)"
+        ],
+        "number-of-cpus" => 8,
+        "number-of-nodes" => 1,
+        "total-memory" => 235520L
+    }
+}
+
+----
+
+=== Accessing Health API using REST
+
+As mentioned in the https://docs.jboss.org/author/display/WFLY10/The+HTTP+management+API[WildFly HTTP Management API] article, the HTTP API requires using proper credentials (use `add-user.sh` script for that).
+
+When credentials are set, all resources accessible by CLI can also be accessed using REST interface. An example is shown below:
+
+[source,bash]
+----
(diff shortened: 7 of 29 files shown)


documentation/src/main/asciidoc/server_guide/server_guide.asciidoc
```diff
@@ -1002,3 +1002,99 @@ remoteCacheManager = new RemoteCacheManager(clientBuilder.build());
 
 Additionally, you might also want to enable client certificate authentication (and therefore also allow the use of the EXTERNAL SASL mech to authenticate clients).
 
+== Health monitoring
+
+Hot Rod server has special endpoints for monitoring cluster health. The API is exposed via:
```
gustavonalle: s/Hot Rod server/Infinispan Server/

documentation/src/main/asciidoc/server_guide/server_guide.asciidoc
```diff
@@ -1002,3 +1002,99 @@ remoteCacheManager = new RemoteCacheManager(clientBuilder.build());
...
+
+* Programmatically (using `embeddedCacheManager.getHealth()`)
+* JMX
+* CLI
+* REST (using https://docs.jboss.org/author/display/WFLY10/The+HTTP+management+API[WildFly HTTP Management API])
+
+=== Accessing Health API using JMX
+
+At first you need to connect to the Hot Rod server using JMX (use JConsole or other tool for this).
```
gustavonalle: s/Hot Rod server/Infinispan Server/

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/HealthMetricsHandler.java
```diff
@@ -0,0 +1,217 @@
...
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
```
gustavonalle: This header seems outdated

slaskawi: All headers in this module are outdated :) But we have a checkstyle rule to fix them: https://github.com/infinispan/infinispan/blob/master/checkstyle/src/main/resources/checkstyle.xml#L132

So if you agree with me, I would like to remove all of them at once (of course if others agree with PR I'm going to propose).

documentation/src/main/asciidoc/server_guide/server_guide.asciidoc
slaskawi: Done.

documentation/src/main/asciidoc/server_guide/server_guide.asciidoc
slaskawi: Done. I think you already marked this one using the comment above.

slaskawi:
Comments addressed and rebased.

slaskawi:
Rebased

documentation/src/main/asciidoc/server_guide/server_guide.asciidoc
danberindei: Doesn't seem fixed yet.

slaskawi: Fixed now. I probably forgot to push something... my bad

slaskawi:
Fixed and rebased.

tristantarrant:
Asking for health on a standalone node via CLI fails:
15:24:58,225 ERROR [org.jboss.as.controller.management-operation] (management-handler-thread - 4) WFLYCTL0013: Operation ("read-attribute") failed - address: ([
    ("subsystem" => "datagrid-infinispan"),
    ("cache-container" => "local"),
    ("health" => "HEALTH")
]): java.lang.NullPointerException
	at org.infinispan.health.impl.ClusterHealthImpl.getNumberOfNodes(ClusterHealthImpl.java:41)
	at org.jboss.as.clustering.infinispan.subsystem.HealthMetricsHandler.executeRuntimeStep(HealthMetricsHandler.java:134)
	at org.jboss.as.controller.AbstractRuntimeOnlyHandler$1.execute(AbstractRuntimeOnlyHandler.java:53)
	at org.jboss.as.controller.AbstractOperationContext.executeStep(AbstractOperationContext.java:890)
	at org.jboss.as.controller.AbstractOperationContext.processStages(AbstractOperationContext.java:659)
	at org.jboss.as.controller.AbstractOperationContext.executeOperation(AbstractOperationContext.java:370)
	at org.jboss.as.controller.OperationContextImpl.executeOperation(OperationContextImpl.java:1329)
	at org.jboss.as.controller.ModelControllerImpl.internalExecute(ModelControllerImpl.java:400)
	at org.jboss.as.controller.ModelControllerImpl.execute(ModelControllerImpl.java:222)
	at org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler.doExecute(ModelControllerClientOperationHandler.java:208)
	at org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler.access$300(ModelControllerClientOperationHandler.java:130)
	at org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler$1$1.run(ModelControllerClientOperationHandler.java:152)
	at org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler$1$1.run(ModelControllerClientOperationHandler.java:148)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.jboss.as.controller.AccessAuditContext.doAs(AccessAuditContext.java:149)
	at org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler$1.execute(ModelControllerClientOperationHandler.java:148)
	at org.jboss.as.protocol.mgmt.AbstractMessageHandler$ManagementRequestContextImpl$1.doExecute(AbstractMessageHandler.java:363)
	at org.jboss.as.protocol.mgmt.AbstractMessageHandler$AsyncTaskRunner.run(AbstractMessageHandler.java:472)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
	at org.jboss.threads.JBossThread.run(JBossThread.java:320)

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerResource.java
```diff
@@ -372,6 +375,7 @@ public void registerChildren(ManagementResourceRegistration resourceRegistration
         super.registerChildren(resourceRegistration);
 
         // child resources
+        resourceRegistration.registerSubModel(new HealthResource(pathManager));
```
tristantarrant: This should only be registered if runtimeRegistration is true

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/HealthMetricsHandler.java
```diff
@@ -0,0 +1,217 @@
...
+                    result.set(health.getClusterHealth().getClusterName());
+                    break;
+                case NUMBER_OF_CPUS:
+                    result.set(health.getHostInfo().getNumberOfCpus());
+                    break;
+                case CLUSTER_HEALTH:
+                    result.set(health.getClusterHealth().getHealthStatus().toString());
+                    break;
+                case LOG_TAIL:
+                    File path = new File(pathManager.resolveRelativePathEntry("server.log", ServerEnvironment.SERVER_LOG_DIR));
```
tristantarrant: This would only work if the user has not modified the default name / path for the server logs.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/HealthMetricsHandler.java
```diff
@@ -0,0 +1,217 @@
...
+                    try (BufferedReader reader = new BufferedReader(new FileReader(path))) {
+                        List<String> results = new LinkedList<>();
+                        for (String line=reader.readLine(); line != null; line=reader.readLine()) {
+                            //add new lines at the end and remove the head if needed
+                            results.add(line);
+                            if(results.size() > NUMBER_OF_LINES) {
+                                results.remove(0);
+                            }
+                        }
+                        result.set(toModelNodeCollection(results));
```
tristantarrant: This is quite inefficient for large files. Looking at https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/input/ReversedLinesFileReader.java we could use that approach instead.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/HealthMetricsHandler.java
slaskawi: Yes, that is correct. That's exactly why the user needs to specify the filename when asking Logging Subsystem about the logs. I'm not sure if there is any better way to do it...

slaskawi:
I must have messed up something when rebasing. Let me have a look what happened.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerResource.java
slaskawi: Done

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/HealthMetricsHandler.java
slaskawi: I would like to avoid copy-pasting classes and we use pretty old Commons IO artifact (which doesn't have ReversedLinesFileReader). I created 2 JIRAs to address this problem: https://issues.jboss.org/browse/ISPN-7027 https://issues.jboss.org/browse/ISPN-7028

slaskawi:
@tristantarrant Thanks for the review. I addressed all the comments and rebased the code.

core/src/main/java/org/infinispan/health/ClusterHealth.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Returns the number of nodes in the cluster.
+     */
+    int getNumberOfNodes();
+
+    /**
+     * Returns node names.
+     */
+    String[] getNodeNames();
```
danberindei: -1 to returning arrays, you can't make them immutable so you always have to copy them.

core/src/main/java/org/infinispan/health/HealthFactory.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    /**
+     * Creates Health API.
+     * <p>
+     *     This method also registers the API in {@link org.infinispan.factories.GlobalComponentRegistry}.
+     * </p>
+     *
+     * @param embeddedCacheManager Cache manager to be monitored.
+     * @return Health API.
```
danberindei: I think every instance of "API" in this javadoc should be replaced with "instance".

core/src/main/java/org/infinispan/health/HealthFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public class HealthFactory {
+
+    private HealthFactory() {
+    }
+
+    /**
+     * Creates Health API.
+     * <p>
+     *     This method also registers the API in {@link org.infinispan.factories.GlobalComponentRegistry}.
```
danberindei: Does the user need to know about the GCR?

core/src/main/java/org/infinispan/health/HealthStatus.java
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * Given entity is healthy.
+     */
+    GREEN,
+
+    /**
+     * Given entity is healthy but a rebalance is in progress.
+     */
+    YELLOW
```
danberindei: I'd rather call the enum values UNHEALTHY, HEALTHY, REBALANCING, and let the UI decide what color to use.

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
```diff
@@ -230,6 +233,7 @@ public DefaultCacheManager(GlobalConfiguration globalConfiguration, Configuratio
       this.globalComponentRegistry.registerComponent(cacheDependencyGraph, CACHE_DEPENDENCY_GRAPH, false);
       this.globalComponentRegistry.registerComponent(authzHelper, AuthorizationHelper.class);
       this.stats = new CacheContainerStatsImpl(this);
+      this.health = HealthFactory.create(this);
```
danberindei: If the `Health` instance is always created from the `DefaultCacheManager` then I don't see a need for the factory.

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
```diff
@@ -319,5 +320,11 @@ default ClusterExecutor executor() {
       throw new UnsupportedOperationException();
    }
 
-
+   /**
+    * Returns an entry point for a Health Check API.
+    *
+    * @since 9.0
+    * @return Health API for this {@link EmbeddedCacheManager}.
```
danberindei: I'd leave out the `@return` tag, it just repeats the sentence above. And I'd also leave out "Returns" from the sentence above :)

core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Collections;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.infinispan.health.ClusterHealth;
+import org.infinispan.health.HealthStatus;
+import org.infinispan.manager.EmbeddedCacheManager;
+
+public class ClusterHealthImpl implements ClusterHealth {
```
danberindei: Wrong name I think, this reports the health of a single manager, and not the health of the entire cluster.

documentation/src/main/asciidoc/server_guide/server_guide.asciidoc
```diff
@@ -1002,3 +1002,99 @@ remoteCacheManager = new RemoteCacheManager(clientBuilder.build());
...
+
+[source,bash]
+----
+$ ispn-cli.sh -c "/subsystem=datagrid-infinispan/cache-container=clustered/health=HEALTH:read-resource(include-runtime=true)"
+
+{
+    "outcome" => "success",
+    "result" => {
+        "cache-health" => "GREEN",
+        "cluster-health" => ["test"],
```
danberindei: `["test"]` seems wrong here. To sum up my suggestions, I'd like to see `"manager-health" => "HEALTHY"` instead.

slaskawi: Sorry, you are absolutely correct... the output should be `"cluster-health" : ["test", "GREEN"]` since we use `RED`, `YELLOW` and `GREEN` (see the [design](https://github.com/infinispan/infinispan/wiki/Health-check-API)).

core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
slaskawi: Yes and you specify which cache manager in the CLI - e.g. `/subsystem/datagrid-infinispan/cache-container/clustered/health/HEALTH`. 

So in this context when I'm asking is my cluster healthy, I mean if all nodes in the cluster operate correctly and if cache is not in degraded mode. 

I believe the terminology is correct.

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
slaskawi: I see we have a task to add `@return` to javadocs in documentation mini-sprint. Having this in mind I would like to leave it as is.

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
slaskawi: Currently we have only one Health statistics implementation which is based on JMX. However we might add some more in the future (e.g. Kubernetes might use Prometheus). Implementing a trivial factory for this shows where to plug those. 

If you don't like, I can remove it. But I strongly believe it makes things easier to extend in the future.

core/src/main/java/org/infinispan/health/HealthStatus.java
slaskawi: I would prefer to talk about such things during design phase not the implementation. 

However if others have similar thoughts, I will change it.
@galderz @gustavonalle @tristantarrant (mentioned only folks involved in this review) - thoughts?

core/src/main/java/org/infinispan/health/HealthFactory.java
slaskawi: No, it's an implementation detail. But you're right - I'll remove it.

core/src/main/java/org/infinispan/health/HealthFactory.java
slaskawi: No, it's an implementation detail. But you're right - I'll remove it.

core/src/main/java/org/infinispan/health/HealthFactory.java
slaskawi: Done

core/src/main/java/org/infinispan/health/ClusterHealth.java
slaskawi: Fixed. Now I return lists from all those methods.

slaskawi:
@danberindei @gustavonalle @tristantarrant Updates, rebased. Ready for the next round.

core/src/main/java/org/infinispan/health/HealthStatus.java
tristantarrant: The review process is here to catch things at any time. I didn't think much about it at the time, but I do agree with @danberindei , although I'd probably change the middle state to something more specific than YELLOW but also less specific than REBALANCING (BUSY / REPAIRING, IN_MAINTENANCE ?), since we might want to indicate other situations where the cluster is best left (temporarily) alone.

gustavonalle: +1 for using names instead of colors, and REBALANCING

slaskawi: Done

slaskawi: I also updated the design.

slaskawi:
Udated, rebase and repushed...

tristantarrant:
This breaks domain mode

[Host Controller] 13:28:19,550 ERROR [org.jboss.as.controller] (Controller Boot Thread) WFLYCTL0030: No resource definition is registered for address [
[Host Controller]     ("profile" => "clustered"),
[Host Controller]     ("subsystem" => "datagrid-infinispan"),
[Host Controller]     ("cache-container" => "clustered"),
[Host Controller]     ("health" => "HEALTH")
[Host Controller] ]
[Host Controller] 13:28:19,553 FATAL [org.jboss.as.host.controller] (Controller Boot Thread) WFLYHC0034: Host Controller boot has failed in an unrecoverable manner; exiting. See previous messages for details.
[Host Controller] 13:28:19,553 FATAL [org.jboss.as.host.controller] (Controller Boot Thread) WFLYHC0178: Aborting with exit code 99

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
```diff
@@ -343,6 +343,21 @@ private void parseContainer(XMLExtendedStreamReader reader, PathAddress subsyste
...
+
+    /**
+     * A {@link HealthResource} is not populated to the XML. However the easiest way to propagate a Health node
+     * (we need to attach Health node to the Cache Container node) and attach metrics to it is to
+     * simply simulate reading it.
+     *
+     * Since we are not persisting it - there is no need to check the namespaces.
+     */
+    private void addSyntheticHealtheOperation(Map<PathAddress, ModelNode> operations, PathAddress containerAddress) {
+        PathAddress healthAddress = containerAddress.append(ModelKeys.HEALTH, ModelKeys.HEALTH_NAME);
```
tristantarrant: I think this approach is wrong. health is a runtime resource and therefore it should not be handled by the configuration model

slaskawi: Yes, I also thought this is a bit smelly but couldn't find better way to for WF to register HealthResource (technically call `HealthResource#registerAttributes`) (see https://github.com/infinispan/infinispan/pull/4499/files/fb37662ddd39d4e8f1845999e1257203ca3b7515#diff-3aa0cebf53a46eb936bd575a83d9310bR53). Do you have any other idea that I could try?

slaskawi:
@tristantarrant I will look into failures shortly but let's agree on this `addSyntheticHealtheOperation` thing first.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/HealthResource.java
```diff
@@ -0,0 +1,57 @@
...
+ * @since 9.0
+ */
+public class HealthResource extends SimpleResourceDefinition {
+
+    public static final PathElement HEALTH_PATH = PathElement.pathElement(ModelKeys.HEALTH, ModelKeys.HEALTH_NAME);
+
+    private final PathManager pathManager;
+
+    public HealthResource(PathManager pathManager) {
+        super(HEALTH_PATH,
```
tristantarrant: to make this a runtime resource, invoke the super constructor with an extra boolean set to true

slaskawi: Ha! This trick works! Let me update the tests, check if domain mode works and we are good to go.

slaskawi: Unfortunately it doesn't work :( I recompiled wrong modules... trying to figure out something

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
```diff
@@ -343,6 +343,21 @@ private void parseContainer(XMLExtendedStreamReader reader, PathAddress subsyste
                 }
             }
         }
+
+        addSyntheticHealtheOperation(operations, containerAddress);
```
rhusar: What is "Healthe"? Is that intentional?

slaskawi: hahah it's a typo :) Anyway I just removed this method... Could you please hold on with the review for ~30 mins? I'll upload new version (once I fix it :D)

rhusar: Right, this should be gone if you are going to use the runtime resource

slaskawi:
@rhusar @tristantarrant I fixed the error in domain mode (it turns out that you ALWAYS need to register submodels (I had to remove `if(runtime)resourceRegistration.registerSubModel(new HealthResource(pathManager));`)

However I didn't manage to get rid of `InfinispanSubsystemXMLReader#addSyntheticHealthOperation`. Even though the `HealthResource` is runtime (it returns `true` in `HealthResource#isRuntime`), without this synthetic operation it is not being added to the resource tree. As the result, when invoking `ispn-cli.sh` with parameters `/subsystem=datagrid-infinispan/cache-container=local:read-resource(include-aliases=true,recursive=true,include-runtime=true)` I get `... health=> undefined` back :disappointed: 

Please let me know if there is anything I can try to get rid of this nasty hack. I ran out of ideas...

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
danberindei: I'd be fine with a factory if `Health` was a component, even if it had only one implementation. But I don't like the idea of having several kind of factories in core.

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
danberindei: I hope there's no plan to copy the javadoc of all our methods into `@return` tags ;)

core/src/main/java/org/infinispan/health/HealthStatus.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * General Health status.
+ *
+ * @author Sebastian Łaskawiec
+ * @since 9.0
+ */
+public enum HealthStatus {
+    /**
+     * Given entity is unhealthy.
```
galderz: What does unhealthy mean? Could we add some more info into the javadoc?

core/src/main/java/org/infinispan/health/HostInfo.java
```diff
@@ -0,0 +1,25 @@
+package org.infinispan.health;
+
+/**
+ * Information about the host.
+ *
+ * @author Sebastian Łaskawiec
+ * @since 9.0
+ */
+public interface HostInfo {
```
galderz: Isn't this info available via JMX?

galderz: I mean, the info below is accessible via JMX mbeans provided by the JVM itself (see [here](https://docs.oracle.com/javase/8/docs/api/java/lang/management/ManagementFactory.html)). Not sure why we should provide a facade for those?

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
slaskawi: hahaha me too :)

core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
slaskawi: I'm not sure it if our responsibility to aggregate the cluster status with RPCs... Maybe a sophisticated monitoring tool (Zabbix, Hawkular, Splunk) should be used in this case?

The bottom line is that I would prefer to keep this API short for start. And if there are such needs, enhance it in the future.

core/src/main/java/org/infinispan/health/HealthStatus.java
slaskawi: Unhealthy means that some partitions are in `DEGRADED_MODE`. I added more information to the javadoc. Thanks for pointing this out.

core/src/main/java/org/infinispan/health/HostInfo.java
slaskawi: Yes, indeed. Some information are already available via JMX. However the goal is to make this API available through REST (WildFly management REST interface). This way OpenShift will be able to monitor resource utilization and check the health status.

slaskawi:
@tristantarrant @danberindei @galderz Thanks a lot for the review guys! I added all changes you requested.

@rhusar @tristantarrant It would be awesome if you found some time to help me with `addSyntheticHealthOperation`. Unfortunately none of the suggested solutions worked. I'm not sure if I'm not asking for too much but maybe if you could cherry-pick this PR on top of your repository and try to look into this a bit deeper? Maybe I missed some silly, obvious thing?

slaskawi:
Closing this one. Will add more commits and reopen.

PR #4502, opened 2016-08-10
[8.2.x] ISPN-6944 Some public query classes are missing from apidocs 

anistor:
https://issues.jboss.org/browse/ISPN-6944


danberindei:
Integrated, thanks Adrian!

PR #4508, opened 2016-08-15
Fixed lengthy test names in TC (testFoo on org.infinispan.FooTest@12345678)

rvansa:
@danberindei recently complained that recent changes in testsuite use too lengthy test names in TC. This fixes that issue by renaming the method name (adding parameters()) but while this makes even more easily distinguishable tests in IDE, it breaks the ability to navigate to the test method/re-run the test method.

A workaround (before https://youtrack.jetbrains.com/issue/IDEA-157827#comment=27-1485365 is fixed) is using a system property in CI(TC) that uses the fix, and just don't distinguish the methods on normal runs (the threads are named using test name, so it should be still distinguishable).

WDYT?

### DIFF
--- a/core/src/test/java/org/infinispan/api/mvcc/PutForExternalReadLockCleanupTest.java
+++ b/core/src/test/java/org/infinispan/api/mvcc/PutForExternalReadLockCleanupTest.java
@@ -51,7 +51,7 @@ public class PutForExternalReadLockCleanupTest extends MultipleCacheManagersTest
 
    @Override
    protected String parameters() {
-      return "{" + name + "}";
+      return "[" + name + "]";
    }
 
    public void testLockCleanupOnBackup() {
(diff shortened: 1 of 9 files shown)


danberindei:
@rvansa I found a workaround for the source navigation issue in IntelliJ: you just need to wrap the parameters in `[]` instead of `{}`. It works because the IntelliJ reporter puts the invocation count in `[]`, and so the IDE ignores everything between the square brackets.

rvansa:
@danberindei Seems this works, thanks for the tip! Rebased & replaced the brackets.

slaskawi:
+1000 for this change

danberindei:
Integrated, thanks Radim!

(I had to remove some empty lines at the end of `NamedTestMethod` because of the new checkstyle rules.)

PR #4513, opened 2016-08-17
ISPN-5683 Removed whitespaces at the end of the lines

slaskawi:
https://issues.jboss.org/browse/ISPN-5683

Please do not close the issue once this PR gets in. I'm turning on the rules gradually. 

Fixed by a simple command:
```
find . -name "*.java" -exec sed -i 's/\s*$//' {} \;
```

@rvansa @anistor @galderz @Sanne Could you please review it?

### DIFF
--- a/atomic-factory/src/main/java/org/infinispan/atomic/container/Container.java
+++ b/atomic-factory/src/main/java/org/infinispan/atomic/container/Container.java
@@ -308,7 +308,7 @@ public class Container {
       } else {
          throw new IllegalArgumentException("Unable to find constructor for " + clazz.toString() + " with " + initArgs);
       }
-      
+
    }
 
    //
--- a/atomic-factory/src/test/java/org/infinispan/atomic/Utils.java
+++ b/atomic-factory/src/test/java/org/infinispan/atomic/Utils.java
@@ -25,5 +25,5 @@ public class Utils {
       // Allow some time for all ClusteredGetCommands to finish executing
       TestingUtil.sleepThread(1000);
    }
-   
+
 }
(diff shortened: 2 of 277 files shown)


rvansa:
-1 I use git annotate (blame) to crawl through history quite often, and this introduced unwanted churr.

slaskawi:
@rvansa Hey Radim! Just to be crystal clear - the alternative is to skip this rule. 

I assume you are `-1` for submitting this change and `+1` for removing the rule for checking empty char at the end of the line?

anistor:
http://bit.ly/29oQu40  :broken_heart: :disappointed:

rvansa:
@slaskawi Yep. My priority list is : nice history > no invisible chars on the end of lines > invisible chars on the end of lines

Is it possible that such post-processing would be rather a part of a commit hook, affecting only diffs?

slaskawi:
I have an idea - would you prefer to have this one or https://github.com/infinispan/infinispan/pull/4515 ? Let the vote begin!

slaskawi:
:-1: from me

anistor:
If we give up on this then we might as well give up on the remaining ~35 rules. -1 for giving up.

anistor:
By looking at the changes I see that the majority are on empty lines or javadocs, and few on actual statements. There will be some noise in history indeed, but not where it hurts the most. I'd allow this. And would even go as far as activate the manadatory new-line at EOF rule too so we are spared from seeing these: https://github.com/infinispan/infinispan/pull/4513/files#diff-e2e6f183514f1de4a3e41498851f8a46L6

slaskawi:
@anistor has a point here. If we give up one simple rule like this one, we might give up all of them. Also if you use formatter, you should never see this violation in practice. So changing my mind, :+1: from me.

tristantarrant:
My vote is with Adrian on this one

rvansa:
Okay, I admit I haven't checked all the files - if it's mostly javadoc, it probably won't hurt.

ryanemerson:
+1 It will be worth the initial pain in the long run.

slaskawi:
Rebased

anistor:
Looking again and pulling...

anistor:
Integrated. Thanks @slaskawi !

PR #4514, opened 2016-08-17
ISPN-6954 Stabilize LiveRunningTest.liveRun

slaskawi:
https://issues.jboss.org/browse/ISPN-6954

@belaban @Sanne - Hey! I'm trying to stabilize `LiveRunningTest`. As you can see the test uses loopback and up to 5 slaves. Unfortunately I didn't manage to reproduce the failure locally but I think JGroups were operating on very low timeout values. I tried to make them a bit higher and use more defaults. Could you please have a look if proposed changes make sense? Thanks!

### DIFF
--- a/lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/LiveRunningTest.java
+++ b/lucene/directory-provider/src/test/java/org/infinispan/hibernate/search/LiveRunningTest.java
@@ -51,10 +51,10 @@ public class LiveRunningTest {
             assertViews();
          }
       } finally {
-         master.close();
          for (FullTextSessionBuilder slave : slaves) {
             slave.close();
          }
+         master.close();
       }
    }
 


belaban:
The config change doesn't really change anything :-) So +1 from me...

slaskawi:
Running some more tests: http://ci.infinispan.org/viewType.html?buildTypeId=bt9&branch_Infinispan=4514%2Fhead&tab=buildTypeStatusDiv

gustavonalle:
@slaskawi Do you have TRACE logs of the failure? Sometimes timeouts happens because something went wrong in the receiver side of the RPC

slaskawi:
@gustavonalle There's only one failure but probably not very helpful: http://ci.infinispan.org/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=40101#_focus=13093

slaskawi:
@gustavonalle @Sanne - Any other comments?

gustavonalle:
This test sometimes fails with other reasons than timeout:
```
junit.framework.AssertionFailedError: expected:<12> but was:<11> at junit.framework.Assert.fail(Assert.java:57) at junit.framework.Assert.failNotEquals(Assert.java:329)
```

```
liveRun(org.infinispan.hibernate.search.LiveRunningTest)  Time elapsed: 8.726 sec  <<< ERROR!
org.hibernate.search.exception.SearchException: Unable to reopen IndexReader
```

slaskawi:
@gustavonalle Could you please share a link to CI job?

gustavonalle:
http://ci.infinispan.org/viewLog.html?buildId=38317&tab=buildResultsDiv&buildTypeId=Infinispan_MasterHotspotJdk8&guest=1#testNameId-6124278935670372534&guest=1

gustavonalle:
To reproduce it locally, run in a loop:

```taskset -c 0,1 mvn -Dtest=LiveRunningTest test```

danberindei:
Any luck reproducing the failure @slaskawi ?

slaskawi:
Unfortunately not, but @gustavonalle is right... you can reproduce it by running `taskset -c 0,1 mvn -Dtest=LiveRunningTest test`

slaskawi:
I think I found the root cause... the order of closing nodes was wrong.. 

Now we stop all slaves and then the master. After this change, it behaves really stable on my local environment.

danberindei:
@slaskawi If it doesn't work properly that's the fault of Infinispan, not the fault of the test. Otherwise we'd have to say in the docs that we don't support stopping the master node before all the other nodes, which kind of defeats the point of using Infinispan.

slaskawi:
I guess you're right @danberindei...

@gustavonalle, @Sanne - may I ask you to also have a look into this issue? Maybe I didn't diagnose it properly (I'm unfamiliar with this code, so I can not really tell if the test or the production code needs to be fixed)?

slaskawi:
This is not a proper fix... closing.

PR #4517, opened 2016-08-17
ISPN-5833 ModuleCommandExtensions should not expose deprecated type E…

Sanne:
…xtendedModuleCommandFactory

 - https://issues.jboss.org/browse/ISPN-5833

See also pr #3742


anistor:
Looks good. Finally we can remove some rust! :)

anistor:
I see ```ExtendedModuleCommandFactory``` is still used in core/src/test/java/org/infinispan/remoting/rpc/TestModuleCommandExtensions.java. That usage should go away.

Also, do you think we could remove ExtendedModuleCommandFactory completely?  It was deprecated a major version ago.

anistor:
I've applied the suggestions myself and integrated. Thanks!

PR #4525, opened 2016-08-22
ISPN-6626 - IntegrationTest.testGet* fails on Windows

andyuk1986:
https://issues.jboss.org/browse/ISPN-6626

Adding check for "\r" which will work in case the test is run on Windows machine.

### DIFF
--- a/server/rest/src/test/scala/org/infinispan/rest/IntegrationTest.scala
+++ b/server/rest/src/test/scala/org/infinispan/rest/IntegrationTest.scala
@@ -171,8 +171,8 @@ class IntegrationTest extends RestServerTestBase {
       assertTrue(xml.contains("<key>b</key>"))
 
       val plain = getCollection("text/plain;charset=UTF-8")
-      assertTrue(plain.contains("a\n"))
-      assertTrue(plain.contains("b\n"))
+      assertTrue(plain.contains("a" + System.lineSeparator()))
+      assertTrue(plain.contains("b" + System.lineSeparator()))
 
       val json = getCollection("application/json")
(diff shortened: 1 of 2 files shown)


slaskawi:
LGTM

server/rest/src/test/scala/org/infinispan/rest/IntegrationTest.scala
```diff
@@ -171,8 +171,8 @@ class IntegrationTest extends RestServerTestBase {
       assertTrue(xml.contains("<key>b</key>"))
 
       val plain = getCollection("text/plain;charset=UTF-8")
-      assertTrue(plain.contains("a\n"))
-      assertTrue(plain.contains("b\n"))
+      assertTrue(plain.contains("a\n") || plain.contains("a\r\n"))
```
anistor: Why not use the platform line separator? ```System.getProperty("line.separator")```

anistor:
I know this does fix the test but I would rather go to the root of the issue in rest/Server.scala where the response is generated. The response is inconsistently formatted using both a hardcoded ```\n``` and also ```System.lineSeparator()``` here and there. We should just pick one and stick with it.

andyuk1986:
Thanks @anistor for comment. I completely forgot about System.lineSeparator(), thanks for the hint. 
I have changed and ran the test with lineSeparator. Now it passes. In rest/Server.scala the text/plain response is generated with lineSeparator. The only place where "\n" is used is while generating the XML response. I have changed that as well to use the System.lineSeparator(). Hope it is fine.

anistor:
Thanks @andyuk1986 ! Pulling!

PR #4526, opened 2016-08-22
ISPN-6962 OperationResponse should handle ErrorResponse

wburns:
* Default returns null instead of throwing exception


danberindei:
Integrated, thanks Will!

PR #4527, opened 2016-08-24
ISPN-6969 EmbeddedQuery uses non-serializable lambda and fails on DIST cache

anistor:
https://issues.jboss.org/browse/ISPN-6969


tristantarrant:
Pushed to master, thanks

PR #4529, opened 2016-08-25
ISPN-6970 Update to JGroups 4.0.0.Beta1

danberindei:
https://issues.jboss.org/browse/ISPN-6970


server/integration/jgroups/src/main/java/org/infinispan/server/jgroups/spi/service/ChannelConnectorBuilder.java
```diff
@@ -61,21 +61,21 @@ public ServiceName getServiceName() {
...
-    public Channel getValue() {
+    public JChannel getValue() {
         return this.channel.getValue();
     }
 
     @Override
     public void start(StartContext context) throws StartException {
-        Channel channel = this.getValue();
+        JChannel channel = this.getValue();
         if (channel instanceof JChannel) {
```
pruivo: I think this is not needed...

server/integration/jgroups/src/main/java/org/infinispan/server/jgroups/spi/service/ChannelConnectorBuilder.java
```diff
@@ -61,21 +61,21 @@ public ServiceName getServiceName() {
...
             try {
                 JmxConfigurator.registerChannel((JChannel) channel, this.server.getValue(), this.name);
```
pruivo: ... neither the cast...

server/integration/jgroups/src/main/java/org/infinispan/server/jgroups/spi/service/ChannelConnectorBuilder.java
```diff
@@ -92,7 +92,7 @@ public void start(StartContext context) throws StartException {
 
     @Override
     public void stop(StopContext context) {
-        Channel channel = this.getValue();
+        JChannel channel = this.getValue();
         channel.disconnect();
         if (channel instanceof JChannel) {
```
pruivo: ... same here...

server/integration/jgroups/src/main/java/org/infinispan/server/jgroups/spi/service/ChannelConnectorBuilder.java
```diff
@@ -92,7 +92,7 @@ public void start(StartContext context) throws StartException {
...
             try {
```
pruivo: ... and below.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/ChannelTransport.java
```diff
@@ -39,22 +38,22 @@
...
-            @Override
-            public Object objectFromBuffer(byte[] buffer, int offset, int length) throws Exception {
-                return ChannelTransport.this.factory.isUnknownForkResponse(ByteBuffer.wrap(buffer, offset, length)) ? CacheNotFoundResponse.INSTANCE : super.objectFromBuffer(buffer, offset, length);
-            }
-        };
-        this.dispatcher.setRequestMarshaller(adapter);
-        this.dispatcher.setResponseMarshaller(adapter);
+        this.dispatcher = new CommandAwareRpcDispatcher(channel, this, globalHandler, this.getTimeoutExecutor(), timeService,
+              marshaller);
+        Field bufField;
```
pruivo: hmmm... what is this suppose to do?

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -457,8 +430,36 @@ private RspListFuture processCalls(ReplicableCommand command, boolean broadcast,
...
+
+   public StreamingMarshaller getIspnMarshaller() {
+      return ispnMarshaller;
+   }
+
+   public void setIspnMarshaller(StreamingMarshaller ispnMarshaller) {
+      this.ispnMarshaller = ispnMarshaller;
    }
 
+   private class CustomRequestCorrelator extends RequestCorrelator {
```
pruivo: couldn't you update the existing CustomRequestCorrelator?

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -52,42 +55,43 @@
...
    private final TimeService timeService;
+   private StreamingMarshaller ispnMarshaller;
 
-   public CommandAwareRpcDispatcher(Channel channel, JGroupsTransport transport,
-         InboundInvocationHandler globalHandler, ScheduledExecutorService timeoutExecutor,
-         TimeService timeService) {
-      this.timeService = timeService;
-      this.server_obj = transport;
+   static {
+      ClassConfigurator.addProtocol((short) 1310, CustomRequestCorrelator.class);
```
pruivo: where did you pick the ID?

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -401,50 +379,45 @@ private void staggeredProcessNext(ReplicableCommand command, RspFilter filter, L
...
+      opts.rspFilter(filter);
 
-      GroupRequest<Response> request = cast(dests, msg, opts, false);
-      if (mode == ResponseMode.GET_NONE)
-         return null;
-
+      GroupRequest<Response> request = cast(realDest, buf, opts, false);
+      if (mode == ResponseMode.GET_NONE) return null;
       RspListFuture retval = new RspListFuture(request);
       if (request == null) {
```
pruivo: I don't think the request can be ```null```. Even if it is, the ```RspListFuture``` constructor will throw a NPE.

pruivo:
unused imports in these classes:
```
ConcurrentStartChanelLookupTest
ConcurrentStartForkChannelTest
ChannelConnectorBuilder
```

demos/distexec/src/main/release/etc/config-samples/distexec-demo/jgroups-s3_ping-aws.xml
```diff
@@ -1,18 +1,20 @@
...
    <VERIFY_SUSPECT timeout="1500"/>
-   <pbcast.NAKACK use_mcast_xmit="false" retransmit_timeout="300,600,1200,2400,4800"
+   <pbcast.NAKACK2 use_mcast_xmit="false"
                   discard_delivered_msgs="true"/>
    <UNICAST3 conn_expiry_timeout="0"/>
    <pbcast.STABLE stability_delay="1000" desired_avg_gossip="50000" max_bytes="400000"/>
    <pbcast.GMS print_local_addr="true" join_timeout="${jgroups.join_timeout:5000}" view_bundling="true"/>
-   <FC max_credits="2m" min_threshold="0.40"/>
+   <UFC max_credits="2M" min_threshold="0.40"/>
+   <MFC max_credits="2M" min_threshold="0.40"/>
```
slaskawi: It looks like this configuration uses `TCP` as transport. Having said that, `MFC` seems to be redundant.

demos/gui/src/main/release/etc/config/jgroups-relay1.xml
```diff
@@ -18,57 +18,49 @@
...
     <pbcast.STABLE stability_delay="1000" desired_avg_gossip="50000"
                    max_bytes="1000000"/>
     <pbcast.GMS print_local_addr="true" join_timeout="${jgroups.join_timeout:5000}"
                 max_bundling_time="200"
                 view_bundling="true"/>
-    <FC max_credits="2M"
-        min_threshold="0.40"/>
+    <UFC max_credits="2M"
+          min_threshold="0.40"/>
+    <MFC max_credits="2M"
```
slaskawi: Exactly opposite situation. Now we don't need `MFC`

slaskawi: Could you please look at other files for similar errors?

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
danberindei: `TEST_PING` ;)

But I think I can use a different `RequestCorrelator` constructor and avoid defining a magic id here.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
danberindei: You're right, I updated `RspListFuture` to assume the request is always non-null but I missed this check.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
danberindei: Yep, it would be better as a separate class.

demos/distexec/src/main/release/etc/config-samples/distexec-demo/jgroups-s3_ping-aws.xml
danberindei: You mean `UFC`?

demos/gui/src/main/release/etc/config/jgroups-relay1.xml
danberindei: Why wouldn't we need `MFC` here?

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/ChannelTransport.java
danberindei: It was supposed to be used by the marshaller adapter, but it's no longer necessary.

demos/distexec/src/main/release/etc/config-samples/distexec-demo/jgroups-s3_ping-aws.xml
slaskawi: No no, MFC. According to [JGroups Manual](http://www.jgroups.org/manual-3.x/html/protlist.html#d0e6322) MFC is for Multicast and UFC is for Unicast. TCP is Unicast whereas UDP is multicast.

danberindei: You can definitely send multicast messages with TCP and unicast messages with UDP.

It's true, TCP also has its own flow control, so UFC isn't really needed with TCP, but MFC is definitely needed both with TCP and with UDP.

slaskawi: As far as I know TCP is point to point and you can not send multicast. But let's ask @belaban

demos/gui/src/main/release/etc/config/jgroups-relay1.xml
belaban: If TCP is used as transport, UFC can be removed.

demos/distexec/src/main/release/etc/config-samples/distexec-demo/jgroups-s3_ping-aws.xml
danberindei: That's exactly JGroups' job, it emulates sending messages to the whole group even when the underlying transport doesn't know about multicasts. If it didn't work, we wouldn't be able to run a cluster on TCP at all :)

demos/gui/src/main/release/etc/config/jgroups-relay1.xml
danberindei: This config uses UDP, and @slaskawi was talking about removing MFC :)

slaskawi: oups my mistake... in this case the config uses UDP so we don't need UFC. Sorry my mistake...

belaban: Ah, no, with UDP you need both UFC and MFC, unless you don't want to have flow control. I suggest leave UFC/MFC in; you can still skip flow control by setting flag NO_FC _per message_. This makes sense, e.g. if it is a blocking sync RPC.

slaskawi: Wow, I didn't know that. Thanks a lot Bela for clarifying this!

server/integration/jgroups/src/main/java/org/infinispan/server/jgroups/ManagedSocketFactory.java
```diff
@@ -80,7 +80,7 @@ private Socket register(final Socket socket) {
 
     @Override
     public ServerSocket createServerSocket(String name) throws IOException {
-        return this.register(this.factory.createServerSocket(name));
+        return this.factory.createServerSocket(name);
```
danberindei: @tristantarrant @pferraro I'm not sure if this is the right way to fix this, but if we try to register an unbound socket we get an `IllegalStateException`. It worked with JGroups 3.6.x worked because `JChannel.startStack()` was overwriting the top protocol's socket factory with its own `DefaultSocketFactory`.

danberindei: To answer my own question, `ManagedSocketFactoryTest` expects the socket to be registered even if unbound. So I guess the problem is in `UnnamedBindingRegistry`, not here.

pferraro: I recall making changes to wildfly-core to address this.  Let me find the relevant upstream commit.

danberindei: Found the WildFly bug: https://issues.jboss.org/browse/WFCORE-1127
It's marked as fixed in 2.0.2, but `SocketBindingManager.UnnamedBindingRegistryImpl` still throws the `IllegalStateException` in wildfly-network 2.2.0.Final.

danberindei:
Updated to remove the `TopologyUUID` helper class.

danberindei:
Also stopped using a `ManagedSocketFactory` altogether, since that's identical to the behaviour we have with JGroups 3.6.x.

danberindei:
Fixed some integration tests that were still failing because they were checking for `TP.discard_incompatible_packets` and `TP.enable_bundling`, removed in JGroups 4.0.

danberindei:
With the default configuration, performance results with 4.0.0.Alpha1 are almost identical to 3.6.10.Final. There are small differences during the write-only stress test, where 4.0.0.Alpha1 seems to take a little more clients to break down. 

We still need to re-evaluate sender-sends-with-timer and the JGroups patches used by JDG 7.0, but that's probably better done after upgrading to 4.0.0.Beta1.

tristantarrant:
Several  failures caused by the removal of the oob thread pool in the server jgroups  subsystem

danberindei:
@tristantarrant I think I've fixed the test failures, but there are still some performance issues so I'm leaving the `Changes Required` label.

We could also decide to integrate only the 4.0.0.Alpha1 commits...

belaban:
@danberindei: @pruivo  and I investigated Pedro's performance issues (IspnPerfTest) on cluster01-04 yesterday and found out that the G1 garbage collector caused roughly a 40% slow down! Removing it improved performance.
I'm investigating why this is the case and what options can be used to tune G1.

server/integration/jgroups/src/main/java/org/infinispan/server/jgroups/subsystem/ChannelMetric.java
```diff
@@ -56,13 +57,13 @@ public ModelNode execute(JChannel channel) {
     NUM_TASKS_IN_TIMER(MetricKeys.NUM_TASKS_IN_TIMER, ModelType.INT) {
         @Override
         public ModelNode execute(JChannel channel) {
-            return new ModelNode(channel.getNumberOfTasksInTimer());
+            return new ModelNode(((TP) channel.getProtocolStack().findProtocol(TP.class)).getNumTimerTasks());
```
pruivo: you can use ```channel.getProtocolStack().getTransport()``` instead of ```findProtocol()```

server/integration/jgroups/src/main/java/org/infinispan/server/jgroups/subsystem/ChannelMetric.java
```diff
@@ -56,13 +57,13 @@ public ModelNode execute(JChannel channel) {
...
         }
     },
     NUM_TIMER_THREADS(MetricKeys.NUM_TIMER_THREADS, ModelType.INT) {
         @Override
         public ModelNode execute(JChannel channel) {
-            return new ModelNode(channel.getTimerThreads());
+            return new ModelNode(((TP) channel.getProtocolStack().findProtocol(TP.class)).getTimerThreads());
```
pruivo: you can use ```channel.getProtocolStack().getTransport()``` instead of ```findProtocol()```

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -51,42 +46,38 @@
...
-   private static final int REPLY_FLAGS_TO_CLEAR = Message.Flag.RSVP.value() | Message.Flag.SCOPED.value() |
-         Message.Flag.INTERNAL.value() | Message.Flag.NO_TOTAL_ORDER.value();
+   private static final int REPLY_FLAGS_TO_CLEAR =
+         Message.Flag.RSVP.value() | Message.Flag.INTERNAL.value() | Message.Flag.NO_TOTAL_ORDER.value();
    private static final int REPLY_FLAGS_TO_SET = NO_FC.value();
 
    private final InboundInvocationHandler handler;
    private final ScheduledExecutorService timeoutExecutor;
    private final TimeService timeService;
    private final Executor remoteExecutor;
```
pruivo: No longer needed.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -268,53 +247,45 @@ private void reply(org.jgroups.blocks.Response response, Object retVal, Replicab
...
-      msg.setBuffer(buf);
-      encodeDeliverMode(msg, deliverOrder);
+   static RequestOptions constructRequestOptions(ResponseMode mode, boolean rsvp, DeliverOrder deliverOrder,
+                                                 long timeout) {
+      RequestOptions options = new RequestOptions(mode, timeout);
+      encodeDeliverMode(options, deliverOrder);
       //some issues with the new bundler. put back the DONT_BUNDLE flag.
       if (deliverOrder == DeliverOrder.NONE || mode != ResponseMode.GET_NONE) {
-         msg.setFlag(Message.Flag.DONT_BUNDLE.value());
+         options.setFlags(Message.Flag.DONT_BUNDLE.value());
```
pruivo: should we remove the ```DONT_BUNDLE``` flag? it shown better performance without it during the F2F meeting.

core/src/main/resources/default-configs/default-jgroups-tcp.xml
```diff
@@ -10,20 +10,8 @@
         bundler_type="transfer-queue"
 
         thread_pool.min_threads="${jgroups.thread_pool.min_threads:2}"
```
pruivo: I'm wondering if the default min threads should be zero... WDYT?

core/src/test/java/org/infinispan/statetransfer/ConcurrentStartForkChannelTest.java
```diff
@@ -132,18 +133,18 @@ public Object handleUnknownForkChannel(Message message, String forkChannelId) {
 
          private Object handle(Message message) {
             short id = ClassConfigurator.getProtocolId(RequestCorrelator.class);
-            RequestCorrelator.Header header = (RequestCorrelator.Header) message.getHeader(id);
+            RequestCorrelator.Header header = message.getHeader(id);
             if (header != null) {
                log.debugf("Sending CacheNotFoundResponse reply for %s", header);
                Message response = message.makeReply().setFlag(message.getFlags())
-                     .clearFlag(Message.Flag.RSVP, Message.Flag.SCOPED);
+                     .clearFlag(Message.Flag.RSVP);
```
pruivo: should we use the ```REPLY_FLAGS_TO_CLEAR``` and ```REPLY_FLAGS_TO_SET```?

server/integration/jgroups/src/main/java/org/infinispan/server/jgroups/JChannelFactory.java
```diff
@@ -173,16 +163,16 @@ public Object handleUnknownForkChannel(Message message, String forkChannelId) {
             }
 
             private Object handle(Message message) {
-                Header header = (Header) message.getHeader(this.id);
+                Header header = message.getHeader(this.id);
                 // If this is a request expecting a response, don't leave the requester hanging - send an identifiable response on which it can filter
                 if ((header != null) && (header.type == Header.REQ) && header.rspExpected()) {
-                    Message response = message.makeReply().setFlag(message.getFlags()).clearFlag(Message.Flag.RSVP, Message.Flag.SCOPED);
+                    Message response = message.makeReply().setFlag(message.getFlags()).clearFlag(Message.Flag.RSVP);
```
pruivo: should we use the ```REPLY_FLAGS_TO_CLEAR``` and ```REPLY_FLAGS_TO_SET```?

server/integration/jgroups/src/main/java/org/infinispan/server/jgroups/JChannelFactory.java
```diff
@@ -247,12 +237,13 @@ public boolean isUnknownForkResponse(ByteBuffer buffer) {
     private void init(TP transport) {
         TransportConfiguration transportConfig = this.configuration.getTransport();
         SocketBinding binding = transportConfig.getSocketBinding();
-        if (binding != null) {
-            SocketFactory factory = transport.getSocketFactory();
-            if (!(factory instanceof ManagedSocketFactory)) {
-                transport.setSocketFactory(new ManagedSocketFactory(factory, binding.getSocketBindings()));
-            }
-        }
+        // TODO Uncomment when WFCORE-1127 is fixed and SocketBindingManager.UnnamedBindingRegistryImpl no longer throws IllegalStateException for unbound sockets
```
pruivo: is there a JIRA for this?
we can create a JIRA and link it to WFCORE-1127 to be notified when it is fixed...

core/src/main/java/org/infinispan/remoting/transport/jgroups/SingleResponseFuture.java
```diff
@@ -43,7 +47,7 @@ public void setTimeoutFuture(Future<?> timeoutFuture) {
    @Override
    public Void call() throws Exception {
       // The request timed out
-      complete(request.getResult());
+      complete(new Rsp<>());
```
pruivo: shouldn't be completed with TimeoutException?

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddressCache.java
```diff
@@ -4,6 +4,7 @@
 import org.infinispan.commons.util.concurrent.jdk8backported.EquivalentConcurrentHashMapV8;
 import org.jgroups.Address;
 import org.jgroups.util.ExtendedUUID;
+import org.jgroups.util.NameCache;
 import org.jgroups.util.UUID;
```
pruivo: Unused import :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -268,53 +247,45 @@ private void reply(org.jgroups.blocks.Response response, Object retVal, Replicab
...
-      return msg;
+      return options;
    }
 
-   Buffer marshallCall(Marshaller marshaller, ReplicableCommand command) {
-      Buffer buf;
+   Buffer marshallCall(ReplicableCommand command) {
       try {
-         buf = marshaller.objectToBuffer(command);
+         return new Buffer(ispnMarshaller.objectToByteBuffer(command));
```
pruivo: Should we use ```marshaller.objectToBuffer()``` instead? ```objectToByteBuffer()``` will always trim the ```byte[]``` and ```objectToBuffer()``` could save us some allocations. wdyt?

pruivo:
Title should be changed to "Update to JGroups 4.0.0.Beta2" :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/SingleResponseFuture.java
danberindei: A Rsp with `!isReceived() && !isSuspected()` is already interpreted as a timeout, so I thought I'd avoid having another check (and adding the exception message to the Log interface ;)). Unfortunately it's harder to do the same for `RspListFuture`, so maybe I *should* throw the exception here...

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
danberindei: The `byte[]` we create could spend a lot of time in UNICAST3/NAKACK2's retransmission tables, so I'd rather keep on trimming it when its way too large. TBH I just assumed our `MarshallerAdapter` did the same, I never thought that it might trust the size predictor 100%.

danberindei:
I thought I'd leave the upgrade to Beta2 to you, @pruivo ;)

server/integration/jgroups/src/main/java/org/infinispan/server/jgroups/JChannelFactory.java
danberindei: The JGroups subsystem doesn't have a dependency on core, so they aren't available here.

danberindei:
Instead of the Beta2 upgrade, I added another commit to add the `NO_TOTAL_ORDER` flag to all the response messages, instead of removing it.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
danberindei: I noticed `objectToBuffer` also does some trimming, but only if `estimatedSize > actualSize * 4`, so I'm switching to it.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CustomRequestCorrelator.java
```diff
@@ -1,35 +1,50 @@
...
 
    @Override
-   public void receiveSuspect(Address mbr) {
-      // Ignore suspect events, if the node actually left that will be handled by receiveView(View)
+   protected void handleResponse(Request req, Address sender, byte[] buf, int offset, int length,
+                                 boolean is_exception) {
+      Object retval;
+      try {
+         ByteArrayInputStream is = new ByteArrayInputStream(buf, offset, length);
+         retval = ispnMarshaller.objectFromInputStream(is);
```
pruivo: we should use ```objectFromByteBuffer(byte[] buf, int offset, int length)```. ```objectFromInputStream()``` will copy the input and invoke ```objectFromByteBuffer()``` with the (unnecessary) copy.

core/src/main/java/org/infinispan/manager/impl/ClusterExecutorImpl.java
```diff
@@ -72,7 +72,6 @@ public ClusterExecutorImpl(Predicate<? super Address> predicate, EmbeddedCacheMa
 
    /**
     * Returns the targets we should use for JGroups.  This excludes the local node if it is a target.
-    * @return
```
wburns: Any reason to remove this?  If anything maybe put the description as the return text :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -778,8 +766,9 @@ private boolean ignoreTimeout(ResponseFilter responseFilter) {
 
       Map<Address, Response> retval = new HashMap<>(CollectionFactory.computeCapacity(rsps.size()));
       boolean hasResponses = false;
+      Iterator<Map.Entry<Address, ReplicableCommand>> rpcCommandsIterator = rpcCommands.entrySet().iterator();
       for (Rsp<Response> rsp : rsps) {
```
wburns: The rsps List seems redundant, we can just use the futures array directly and avoid an additional allocation and copy.

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
```diff
@@ -223,7 +223,7 @@ public void waitForInitialStateTransferToComplete() throws Exception {
             initialStateTransferComplete.countDown();
          }
          if (trace) log.tracef("Waiting for initial state transfer to finish for cache %s on %s", cacheName, rpcManager.getAddress());
-         boolean success = initialStateTransferComplete.await(configuration.clustering().stateTransfer().timeout(), TimeUnit.MILLISECONDS);
+         boolean success = initialStateTransferComplete.await(100 * configuration.clustering().stateTransfer().timeout(), TimeUnit.MILLISECONDS);
```
wburns: Need to undo?  Guessing this was for testing purposes.

core/src/main/java/org/infinispan/remoting/transport/jgroups/CustomRequestCorrelator.java
danberindei: Done, I should have realized that my unmarshalling was just as bad as my marshalling :)

core/src/main/java/org/infinispan/manager/impl/ClusterExecutorImpl.java
danberindei: I'm not a big fan of duplicated text. And IntelliJ flags tags with missing text as warnings :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/CustomRequestCorrelator.java
pruivo: you forgot to remove the import of ```ByteArrayInputStream``` :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
danberindei: I made the change, but I just realized the loop is wrong. The entries in the map are not in the same order as the futures/rsps in the list.

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
danberindei: Indeed, good catch.

danberindei:
Updated.

danberindei:
Local tests look good, but CI shows the functional tests fixed by ISPN-7155 are failing so I need to have another look...

danberindei:
I had lost the ISPN-7155 fix during rebase, should be fine now.

pruivo:
@wburns any other comment or can I merge it?

wburns:
@pruivo Seems fine to me

pruivo:

PR #4530, opened 2016-08-25
Reduce dependency on JBoss Marshalling

galderz:
Grab yourself a cup of your favourite drink ☕ and then have a read :)

This is a preview of the changes coming at the marshalling layer. Eventually I'll tidy up the commits but here's a summary of the changes so that you can direct your attention to right location.

The biggest change of this PR is the switch from a single marshaller based on JBoss Marshalling to two marshallers: internal and external. Internal is the one used for primitive types, common types (String) and Infinispan Externalizers. When internal marhaller can't find how to marshall a type, it delegates the work to the external marshaller, whose default implementation uses JBoss Marshalling. This reduces our dependency on JBoss Marshalling as its only used as last resort for Serializable types for which there are no Infinispan Externalizers, or classes with JBoss Marshalling Externalizers.

A key aspect of this PR is that I've tried to work within the boundaries of existing interfaces. So there are no interface changes for externalizers nor marshaller. This work has been deferred until we study Will's requirements for off-heap.

Here are the changes in more detail:
* Internal marshaller provides `ObjectOutput`/`ObjectInput` implementations backed by a simple `byte[]` which gets expanded as needed.
* Encoding itself is currently defined in a separate stateless implementation. This separation between the marshalled state holder and the encoding was done with the idea of trying different encodings, but this separation makes some aspects awkward (e.g. encoding impl having to rely on `ObjectOutput`/`ObjectInput` implementation details). Until we have time to experiment with internal encodings, it might make sense to bring encoding and state into a single `ObjectOutput`/`ObjectInput` implementation.
* Added extra checks to `BaseStoreFunctionalTest.testPreloadStoredAsBinary` so that it verifies the preloaded store-as-binary payloads can be correctly unmarshalled.
* There are two external marshaller implementations, one that uses std Java serialization and the other that uses JBoss Marshalling. I originally started with the former but had some issues (e.g. not being able to marshall instance variables in Serializable classes that are not Serializable but for which externalizers exist; clustered query functionality relies on this). So, I decided to create one based in JBoss Marshalling, which avoids these issues and makes performance more comparable.
* JBoss Marshalling has I/O in mind and it assumes it's the sole owner of the output/input. This means that when it reads, it buffers the input contents until it can't read any more. This had issues with the shared marshalling approach, where the beginning of the `byte[]` could be unmarshalled internally, then delegate to JBMAR, and then follow with the internal marshaller. Because JBMAR's buffering, when JBMAR finished reading, the `byte[]` position would be at the very end, so logic had to be added to be able to rewind back to where JBMAR finished writing bytes. This is the problem that `PositionalBuffer.Input`/`PositionalBuffer.Output` classes resolve in a way that does not add extra state in the `ObjectOutput`/`ObjectInput` implementations.
* Each external marshaller, when writing an object, has an assert that verifies whether the class is about to marshall should be externally marshalled. Since it's an assert, it's only enabled in testing, and this has allowed me to detect types for which we didn't provide advanced externalizers. See list below for those missing externalizers that have been added.
* `CacheRpcCommandExternalizer` has been simplified to avoid wrapping `ObjectInput` in an InputStream. Simply read of the given `ObjectInput`. This class' write path has also been simplified by simply writing to the object output passed in. (@pruivo, you added these changes, so maybe you can explain why you needed this in the first place?).
* For test `CompositeKeyValueFilterConverter`, `ControlledConsistentHashFactory` and `ReplicatedControlledConsistentHashFactory` classes, @SerializeWith annotation was added so that they'd be marshalled with an externalizer. This was added when I was using the external std Java marshaller cos they were Serializable and they referenced classes that can only be marshalled with Externalizers. With the switch to JBoss Marshalling external marshaller this is not needed but makes it easier for the future if trying other external marshallers.
* Deleted `SharedStreammultimarshallertest` since the use case is no longer needed as a result of removing the separation between global and cache marshaller (ISPN-6905).
* Make `TransactionInfo` use `MarshallUtil.marshallArray` to marshall the modification array, making it easier for the internal marshaller to avoid having to support `WriteCommand[]` directly.
* Removed externalizer lifecycle checks. These were added at a time when we were having shutdown issues within the application server. Those issues were resolved and there should not really be a need for such checks any more.
* JBoss Marshalling based external marshaller uses `JBossMarshaller` class which no longer uses the previous externalizer table. Instead the external marshaller plugs a different `ObjectTable` implementation that allows JBoss Marshalling to find externalizers in the internal externalizer table.
* The internal externalizer table is a simplified version of the externalizer table which does not implement `ObjectTable` any more and removed lifecycle checks. On top of that, it identifies the different object type conglomerates which are described in InternalIds class.
* `ExternalizerTable` has been deprecated since it's no longer used for discovering externalizers. However, command factory uses it to find value matcher for lambda externalizers. So, before it can fully removed, there needs to be a way to find the value matcher without having to make the internal externalizer table a component.

Types for which missing externalizers have been added:
* `java.util.ArrayDeque`
* `java.util.Collections$EmptyList`
* `java.util.Collections$EmptyMap`
* `java.util.Collections$SingletonMap`
* `java.util.Collections$SingletonSet`
* `java.util.Collections$SynchronizedSet`
* `java.util.Collections$UnmodifiableRandomAccessList`
* `java.util.Collections$UnmodifiableSet`
* `java.util.UUID`
* `java.util.concurrent.ConcurrentHashMap`
* `org.infinispan.commons.equivalence.AnyEquivalence`
* `org.infinispan.commons.equivalence.AnyServerEquivalence`
* `org.infinispan.commons.equivalence.ByteArrayEquivalence`
* `org.infinispan.commons.util.Immutables#ImmutableSetWrapper`
* `org.infinispan.notifications.cachelistener.event.Event.Type`
* `org.infinispan.remoting.MIMECacheEntry`

TODOS:
* Fully remove `ExternalizerTable`. This is a component no longer in use and it's taking up memory, so it should be removed.
* Next up performance work will be done on this branch. Initial results on a single machine show performance improvement of 5-10% depending on the case, but this needs to be taken with a pinch of salt. More perf tests need to be run. So, next I need to profile it to see which marshalling parts are hot now and see how they can be improved. For example, for those hot externalizers, calling `OO.writeObject()` on something results in a type lookup...etc, so if the externalizer implementation knows exactly the type and which externalizer there is for that type, it could directly call it without without having to go through the externalizer table.

Finally, assigning to Pedro since he's worked on the marshalling side previously, but anyone is of course welcome to comment.


galderz:
A new commit has been added to speed up externalizer resolution. Local numbers show improved performance but these need to be further validated. These are the details of this change:

* Use IdentityIntMap instances for resolving externalizers at write
  time since weak hash maps can produce high CPU usage under heavy
  concurrency situations. This is necessary since internal externalizer
  table is used way more than old externalizer table as a result of
  dealing with more types.
* Upon read, internal externalizers are looked up via an array, so the
  externalizer ids are now resolved at startup time. This mean that the
  order in which internal externalizers are defined is fixed! New
  internal externalizers are to be defined at the bottom of the pack.
* Predefined externalizers via global configuration are now kept in
  separate collections that are sized differently. Also, since there's
  no order that can be assumed about predefined externalizers, their id
  comes from the AdvancedExternalizer implementation. As a result of
  this, an IntObjectMap instance is used for externalizer lookup when
  reading from input.
* IntObjectMap is a interface for a map whose keys are integers. This
  class, and its default IntObjectHashMap, have been brought from The
  Netty Project.

galderz:
I'm seeing an error:

```
13:55:46,872 ERROR (testng-ClientEventsOOMTest) [TestSuiteProgress] Test failed: org.infinispan.client.hotrod.event.ClientEventsOOMTest.testOOM
org.infinispan.client.hotrod.exceptions.HotRodClientException:Request for messageId=9620 returned server error (status=0x85): org.infinispan.commons.CacheException: Problems invoking command.
java.lang.OutOfMemoryError: Java heap space
```

Looking into what's causing that...

galderz:
I've investigated the error in ClientEventsOOMTest and it's indeed related. I've added a new commit to fix it:

Client events OOM owners set to 2
* The new marshaller does not have an instance cache although one can be
  provided for specific types by extending
  InstanceReusingAdvancedExternalizer. As a result of not having a
  default instance cache, when retrieving the entries to send as part of
  the listener state, half of those entries come from a different node
  when owners=1. With all these entries referencing GODZILLA byte[], the
  old JBMAR marshaller was detecting a shared instance and reusing
  it. The new marshaller does not do this by default and hence treats
  each byte[] differently and the test blows.
* This circumstance is artificial and don't think it tends to happen in
  real life, so the best option I found is to set owners=2 and avoid the
  intra-cluster marshalling issue.
* Besides, this tests objective is not to stress the intra-cluster
  marshalling layer, but how the client/server network layer handles big
  object, which it remains intact.

On top of what's in the commit message, I wanted to explain some of the other alternatives I looked into:

* A potential fix here would have been to make `ListExternalizer` and `PrimitiveExternalizer` extend InstanceReusingAdvancedExternalizer, but I didn't want to impose the extra logic into these externalizers, particularly the `PrimitiveExternalizer` without a very good reason. I came to the conclusion that this single huge byte[] shared by a 1000 entries was not really realistic. I mean, i starts as a small byte[] that the system handles well and then it's blown out of proportion. That's not what happens in reality :)
* Something I tried was to have the Godzilla wrapped around a POJO and then have a externalizer, but externalizers are not supported in client-side marshaller, so this didn't work.

So, all in all, taking in account that the test is artificial and that its purpouse is not to stress intra-cluster marshalling but netty layer to see if all those objects can be sent out, I ended up settling it by changing number of owners to 2.

@anistor (creator of test) @wburns (creator of the instance reusing externalizer) If you have other thoughts let me know.

galderz:
FYI: Although local server performance results show improvement, bigger tests in multiple nodes shows that as number of threads increases, performance decreases compared to existing marshaller.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientEventsOOMTest.java
```diff
@@ -39,7 +39,7 @@
 
    private static final int NUM_NODES = 2;
 
-   private static final int NUM_OWNERS = 1;
+   private static final int NUM_OWNERS = 2;
```
danberindei: Shouldn't we make the cache replicated instead?

core/src/main/java/org/infinispan/marshall/exts/ArrayExternalizer.java
```diff
@@ -0,0 +1,114 @@
...
+import org.infinispan.marshall.core.Ids;
+import org.jboss.marshalling.util.IdentityIntMap;
+
+/**
+ * Externalizers for diverse array types.
+ *
+ * @author Galder Zamarreño
+ * @since 6.0
+ */
+public class ArrayExternalizer extends AbstractExternalizer<Object> {
```
danberindei: Do we still need these array externalizers in the internal marshaller? IMO even `MarshallUtil.marshallArray` is too generic for our internal objects, as most of the time we should know exactly what kind of objects we are trying to marshall/unmarshall.

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
```diff
@@ -31,124 +35,118 @@
...
    private static final Log log = LogFactory.getLog(GlobalMarshaller.class);
    private final boolean trace = log.isTraceEnabled();
 
-   private JBossMarshaller defaultMarshaller;
+   GlobalComponentRegistry gcr;
+   RemoteCommandsFactory cmdFactory;
 
-   private ExternalizerTable extTable;
-   private GlobalConfiguration globalCfg;
+   InternalMarshaller internal;
```
danberindei: The `GlobalMarshaller` seems to just delegate to the `InternalMarshaller`, couldn't we implement the methods directly in `GlobalMarshaller`?

core/src/main/java/org/infinispan/marshall/core/internal/BytesEncoding.java
```diff
@@ -0,0 +1,406 @@
+package org.infinispan.marshall.core.internal;
+
+/**
+ * Basic encoding/decoding of primitives
+ */
+final class BytesEncoding implements Encoding<BytesObjectOutput, BytesObjectInput> {
```
danberindei: +1 to remove the encoding interface+class and do the encoding directly in the ByteObjectOutput/Input classes.

core/src/main/java/org/infinispan/marshall/core/internal/BytesEncoding.java
```diff
@@ -0,0 +1,406 @@
...
+
+   private void writeIntDirect(int intValue, int index, BytesObjectOutput out) {
+      byte[] buf = out.bytes; /* avoid getfield opcode */
+      buf[index] =   (byte) ((intValue >>> 24) & 0xFF);
+      buf[index+1] = (byte) ((intValue >>> 16) & 0xFF);
+      buf[index+2] = (byte) ((intValue >>>  8) & 0xFF);
+      buf[index+3] = (byte) ((intValue >>>  0) & 0xFF);
+   }
+
+   private int checkCapacity(int len, BytesObjectOutput out) {
```
danberindei: `ensureCapacity` sounds more appropriate to me.

core/src/main/java/org/infinispan/marshall/core/internal/ExternalJBossMarshaller.java
```diff
@@ -0,0 +1,226 @@
...
+   }
+
+   @Override
+   public boolean isMarshallable(Object o) throws Exception {
+      return marshaller.isMarshallable(o);
+   }
+
+   @Override
+   public ObjectOutput startObjectOutput(OutputStream os, boolean isReentrant, int estimatedSize) throws IOException {
+      throw new RuntimeException("NYI");
```
danberindei: Do you really need to implement these methods, or can you throw an `UnsupportedOperationException` instead?

core/src/main/java/org/infinispan/marshall/core/internal/ExternalJBossMarshaller.java
```diff
@@ -0,0 +1,226 @@
...
+      public int read() throws IOException {
+         try {
+            return in.readUnsignedByte();
+         } catch (ArrayIndexOutOfBoundsException e) {
+            // When JBoss Marshalling starts reading a stream, it buffers
+            // the contents, so it'll try as read as much as it can, so it
+            // can end up reading too far, in which case launder the exception
+            // into an IOException so that reading can stop without causing
+            // issues.
+            throw new EOFException();
```
danberindei: The `ByteInput` javadoc says " If no byte is available because the end of the stream has been reached, the value -1 is returned."

danberindei:
@galderz I'd rather have the externalizers in a separate pull request, and in fact I'd try to avoid as many of them as possible, and to use `MarshallUtil` instead.

core/src/main/java/org/infinispan/marshall/core/internal/ExternalJBossMarshaller.java
```diff
@@ -0,0 +1,226 @@
...
+         throw new IOException("External JBoss Marshalling requires position tracking object output");
+      }
+   }
+
+   @Override
+   public Object objectFromObjectStream(ObjectInput in) throws IOException, ClassNotFoundException {
+      if (in instanceof PositionalBuffer.Input) {
+         PositionalBuffer.Input posIn = (PositionalBuffer.Input) in;
+
+         int pos = in.readInt();
```
danberindei: I'm not sure you really need this. `limit` and `position` are `protected` in `SimpleDataInput`, so you can add a method in our `ExtendedRiverUnmarshaller` to return `position - limit` and rewind by that amount.

core/src/main/java/org/infinispan/marshall/core/internal/ExternalJBossMarshaller.java
```diff
@@ -0,0 +1,226 @@
...
+import org.infinispan.marshall.core.JBossMarshaller;
+import org.jboss.marshalling.ByteInput;
+import org.jboss.marshalling.ByteOutput;
+import org.jboss.marshalling.ObjectTable;
+import org.jboss.marshalling.Unmarshaller;
+
+final class ExternalJBossMarshaller implements StreamingMarshaller {
+
+   final MarshallableTypeHints marshallableTypeHints = new MarshallableTypeHints();
+   final JBossMarshaller marshaller;
```
danberindei: The distinction between internal and external marshallers seems a bit blurry here...

IMO the two should never mix. The user-supplied marshaller should be used for all keys and values, and never for internal objects, and the internal marshaller should not be used to write internal objects that happen to be contained in user objects. If some internal objects really need to work as part of user objects, we should make them `Serializable` and/or we should allow the user to call our internal marshaller explicitly, but the format of user objects in a cache store should never change just because we added a new internal externalizer for a JDK class.

core/src/main/java/org/infinispan/marshall/core/internal/ExternalJBossMarshaller.java
```diff
@@ -0,0 +1,226 @@
...
+         baseCfg.setObjectTable(new ObjectTable() {
+            @Override
+            public Writer getObjectWriter(Object object) throws IOException {
+               InternalExternalizerTable.MarshallableType type = externalizers.marshallable(object);
+               switch (type) {
+                  case INTERNAL:
+                  case PREDEFINED:
+                  case ANNOTATED:
+                     return (m, obj) -> {
+                        Externalizer<Object> ext = externalizers.findWriteExternalizer(obj, m);
```
danberindei: This is looking up the same key a second time in the same maps. I'd remove the `externalizers.marshallable(object)` call and call `findWriteExternalizer` directly.

core/src/main/java/org/infinispan/marshall/core/internal/ExternallyMarshallable.java
```diff
@@ -0,0 +1,170 @@
...
+            || pkg.startsWith("com.arjuna")
+            ;
+   }
+
+   private static boolean isWhiteList(String className) {
+      return className.endsWith("Exception")
+            || className.contains("$$Lambda$")
+            || className.equals("java.lang.Class")
+            || className.equals("java.util.Date") // test
+            || className.equals("org.infinispan.api.lazy.LazyCacheAPITest$CustomPojo")
```
danberindei: Couldn't we add a marker interface for the test classes, so that only the production classes have to be white-listed here?

core/src/main/java/org/infinispan/marshall/core/internal/InternalExternalizerTable.java
```diff
@@ -0,0 +1,468 @@
...
+      }
+   }
+
+   <T> Externalizer<T> findReadExternalizer(ObjectInput in) {
+      try {
+         // Check if primitive or non-primitive
+         int type = in.readUnsignedByte();
+         switch (type) {
+            case InternalIds.NULL:
+               return internalExts[0];
```
danberindei: IMO we should treat `null` specially, and an `Externalizer` should not be allowed to return `null`.

core/src/main/java/org/infinispan/marshall/core/internal/InternalExternalizerTable.java
```diff
@@ -0,0 +1,468 @@
...
+      extId = addInternalExternalizer(new TopologyAwareSyncConsistentHashFactory.Externalizer(), extId);
+      extId = addInternalExternalizer(new TransactionInfo.Externalizer(), extId);
+      extId = addInternalExternalizer(new TransientCacheEntry.Externalizer(), extId);
+      extId = addInternalExternalizer(new TransientCacheValue.Externalizer(), extId);
+      extId = addInternalExternalizer(new TransientMortalCacheEntry.Externalizer(), extId);
+      extId = addInternalExternalizer(new TransientMortalCacheValue.Externalizer(), extId);
+      extId = addInternalExternalizer(new UnsuccessfulResponse.Externalizer(), extId);
+      extId = addInternalExternalizer(new UnsureResponse.Externalizer(), extId);
+      extId = addInternalExternalizer(new UuidExternalizer(), extId);
+      extId = addInternalExternalizer(new XSiteState.XSiteStateExternalizer(), extId);
```
danberindei: I'd like to remove as many as these as possible, especially the JDK collections. We can keep them in the external `ExternalizerTable` to allow reading entries from old cache stores, but for internal classes we should be able to use `MarshallUtil`.

core/src/main/java/org/infinispan/marshall/core/internal/InternalIds.java
```diff
@@ -0,0 +1,14 @@
+package org.infinispan.marshall.core.internal;
+
+final class InternalIds {
+
+   static final int NULL                = 0x00; // null
+   static final int INTERNAL            = 0x01; // internal, including primitives
+   static final int PREDEFINED          = 0x02; // predefined foreign externalizers
+   static final int ANNOTATED           = 0x03; // annotated with @SerializeWith or @SerializeFunctionWith
+   static final int EXTERNAL            = 0x04; // external
```
danberindei: Assuming we can't eliminate these, shouldn't they be in the `MarshallableType` enum?

core/src/main/java/org/infinispan/marshall/core/internal/InternalMarshaller.java
```diff
@@ -0,0 +1,227 @@
...
+
+   final Encoding<BytesObjectOutput, BytesObjectInput> enc = new BytesEncoding();
+   final InternalExternalizerTable externalizers;
+
+   final StreamingMarshaller external;
+
+   public InternalMarshaller(GlobalComponentRegistry gcr, RemoteCommandsFactory cmdFactory) {
+      this.externalizers = new InternalExternalizerTable(enc, gcr, cmdFactory);
+      //this.external = new ExternalJavaMarshaller();
+      this.external = new ExternalJBossMarshaller(externalizers, gcr.getGlobalConfiguration());
```
danberindei: The external marshaller should definitely be configurable (unlike the internal one).

core/src/main/java/org/infinispan/marshall/core/internal/BytesObjectOutput.java
```diff
@@ -0,0 +1,131 @@
...
+   byte bytes[];
+   int pos;
+
+   public BytesObjectOutput(int size, InternalMarshaller internal) {
+      this.bytes = new byte[size];
+      this.internal = internal;
+   }
+
+   @Override
+   public void writeObject(Object obj) throws IOException {
```
danberindei: I think it would be great if we took this opportunity to add a `void writeSharedObject(Object)` method that did the work of `InstanceReusingAdvancedExternalizers`, but without using thread-locals. Or maybe `bool writeObjectReference(Object)`, so an externalizer could start with `if (writeObjectReference(obj)) return;`?

core/src/main/java/org/infinispan/marshall/core/internal/InternalMarshaller.java
```diff
@@ -0,0 +1,227 @@
...
+      // a byte[] subset within it, it's always appended with length.
+      BytesObjectInput in = BytesObjectInput.from(buf, offset, this);
+      return objectFromObjectInput(in);
+   }
+
+   @Override
+   public Object objectFromInputStream(InputStream is) throws IOException, ClassNotFoundException {
+      // This is a very limited use case, e.g. reading from a JDBC ResultSet InputStream
+      // So, this copying of the stream into a byte[] has not been problematic so far,
+      // though it's not really ideal.
```
danberindei: Hmmm, `JdbcBinaryStore` serializes an entire bucket at a time as a `Map`, and the bucket can grow quite large...

danberindei: I think you could rework the `PositionalBuffer` interfaces to be more friendly with `InputStream`'s `mark`/`reset` idiom, and you wouldn't need to copy it into a `byte[]`.

core/src/main/java/org/infinispan/marshall/core/internal/InternalMarshaller.java
```diff
@@ -0,0 +1,227 @@
...
+         bytes = new ExposedByteArrayOutputStream();
+         buf = new byte[1024];
+      }
+      int bytesRead;
+      while ((bytesRead = is.read(buf, 0, buf.length)) != -1) bytes.write(buf, 0, bytesRead);
+      return objectFromByteBuffer(bytes.getRawBuffer(), 0, bytes.size());
+   }
+
+   @Override
+   public boolean isMarshallable(Object o) throws Exception {
```
danberindei: IMO only the external marshaller should have to deal with potentially-not-marshallable objects.

core/src/main/java/org/infinispan/marshall/core/internal/InternalMarshaller.java
```diff
@@ -0,0 +1,227 @@
...
+      try {
+         return external.isMarshallable(o);
+      } catch (Exception e) {
+         throw new NotSerializableException(
+               "Object of type " + o.getClass() + " expected to be marshallable", e);
+      }
+   }
+
+   @Override
+   public BufferSizePredictor getBufferSizePredictor(Object o) {
```
danberindei: Ideally I'd like to also remove the buffer size predictor for internal objects, and instead ask the objects themselves (or their externalizers) how big a buffer they think they need. But that's probably out of scope for this PR :)

core/src/main/java/org/infinispan/marshall/core/internal/PrimitiveExternalizer.java
```diff
@@ -0,0 +1,683 @@
...
+
+   static final int BOOLEAN_ARRAY               = 0x18; // boolean[].class
+   static final int CHAR_ARRAY                  = 0x19; // ..etc..
+   static final int DOUBLE_ARRAY                = 0x1A;
+   static final int FLOAT_ARRAY                 = 0x1B;
+   static final int INT_ARRAY                   = 0x1C;
+   static final int LONG_ARRAY                  = 0x1D;
+   static final int SHORT_ARRAY                 = 0x1E;
+
+   static final int OBJECT_ARRAY                 = 0x1F; // Object[]
```
danberindei: What about `ArrayExternalizer`? TBH, I think we can remove all array externalizers and use `MarshallUtil` instead.

core/src/main/java/org/infinispan/marshall/core/internal/PrimitiveExternalizer.java
```diff
@@ -0,0 +1,683 @@
...
+         out.writeInt(len);
+         for (short v : obj) out.writeShort(v);
+      }
+   }
+
+   private void writeObjectArray(Object[] obj, ObjectOutput out) throws IOException {
+      final int len = obj.length;
+      if (len == 0) {
+         out.writeByte(ARRAY_EMPTY);
+      } else if (len <= 256) {
```
danberindei: You could use the `SMALL` and `MEDIUM` in the `if` conditions.

core/src/main/java/org/infinispan/marshall/core/internal/PrimitiveExternalizer.java
```diff
@@ -0,0 +1,683 @@
...
+         out.writeByte(ARRAY_SMALL);
+         out.writeByte(len);
```
danberindei: If you wrote `len - 1` instead, then you wouldn't need a branch to support length 256 when reading.

core/src/main/java/org/infinispan/marshall/core/internal/StreamBytesObjectOutput.java
```diff
@@ -0,0 +1,103 @@
+package org.infinispan.marshall.core.internal;
+
+import java.io.IOException;
+import java.io.ObjectOutput;
+import java.io.OutputStream;
+
+final class StreamBytesObjectOutput implements ObjectOutput {
```
danberindei: How common is it to have a `BytesObjectOutput` that's not wrapped in a `StreamBytesObjectOutput`? Memory-wise, I think it would be cheaper to have a `stream` field in every `BytesObjectOutput`.

core/src/main/java/org/infinispan/marshall/exts/CacheEventTypeExternalizer.java
```diff
@@ -0,0 +1,35 @@
...
+import org.infinispan.commons.util.Util;
+import org.infinispan.marshall.core.Ids;
+import org.infinispan.notifications.cachelistener.event.Event;
+
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.util.Set;
+
+public final class CacheEventTypeExternalizer extends AbstractExternalizer<Event.Type> {
```
danberindei: `ClusterEvent.Externalizer.readObject()` should use `MarshallUtil.marshallEnum/unmarshallEnum` instead.

core/src/main/java/org/infinispan/marshall/exts/CacheEventTypeExternalizer.java
```diff
@@ -0,0 +1,35 @@
...
+   }
+
+   @Override
+   public void writeObject(ObjectOutput output, Event.Type object) throws IOException {
+      output.writeByte(object.ordinal());
+   }
+
+   @Override
+   public Event.Type readObject(ObjectInput input) throws IOException, ClassNotFoundException {
+      return Event.Type.values()[input.readUnsignedByte()];
```
danberindei: This creates a copy of the `values()` array every time...

core/src/main/java/org/infinispan/marshall/exts/CacheRpcCommandExternalizer.java
```diff
@@ -105,14 +105,7 @@ public void writeObject(ObjectOutput output, CacheRpcCommand command) throws IOE
...
    private void marshallParameters(CacheRpcCommand cmd, StreamingMarshaller marshaller, ObjectOutput oo) throws IOException {
-      BufferSizePredictor sizePredictor = marshaller.getBufferSizePredictor(cmd);
-      int estimatedSize = sizePredictor.nextSize(cmd);
-      ObjectOutput paramsOutput = marshaller.startObjectOutput(convertObjectOutput(oo), true, estimatedSize);
-      try {
-         cmdExt.writeCommandParameters(paramsOutput, cmd);
-      } finally {
-         marshaller.finishObjectOutput(paramsOutput);
-      }
+      cmdExt.writeCommandParameters(oo, cmd);
```
danberindei: This I like :)

core/src/main/java/org/infinispan/marshall/exts/ListExternalizer.java
```diff
@@ -27,20 +28,28 @@
 
    private static final int ARRAY_LIST = 0;
    private static final int LINKED_LIST = 1;
+   private static final int EMPTY_LIST = 2;
 
-   private final IdentityIntMap<Class<?>> numbers = new IdentityIntMap<Class<?>>(2);
+   private final IdentityIntMap<Class<?>> numbers = new IdentityIntMap<>(6);
```
danberindei: I see `IdentityIntMap` rounds up the capacity to 16, but in general you should be using `CollectionFactory.computeCapacity()` to really avoid resizing.

core/src/main/java/org/infinispan/marshall/exts/ListExternalizer.java
```diff
@@ -51,6 +60,8 @@ public List readObject(ObjectInput input) throws IOException, ClassNotFoundExcep
             return MarshallUtil.unmarshallCollection(input, ArrayList::new);
          case LINKED_LIST:
             return MarshallUtil.unmarshallCollection(input, s -> new LinkedList<>());
+         case EMPTY_LIST:
+            return Collections.emptyList();
```
danberindei: Would it be really bad if we read all lists back as `ArrayList`? :)

core/src/main/java/org/infinispan/marshall/core/internal/BytesEncoding.java
```diff
@@ -0,0 +1,406 @@
...
+            int size = decodeByte(in);
+            String str = new String(in.bytes, 0, in.pos, size);
+            in.pos += size;
+            return str;
+         case 2:
+            // large string
+            return decodeStringUtf8(in);
+         case 3:
+         default:
+            throw new RuntimeException("Unkwown marker(String). mark=" + mark);
```
pruivo: typo "Unkwown"

core/src/main/java/org/infinispan/marshall/core/internal/InternalExternalizerTable.java
```diff
@@ -0,0 +1,468 @@
...
+    * This collection is sized according to an approximation of number of
+    * internal externalizers.
+    */
+   final AdvancedExternalizer[] internalExts = new AdvancedExternalizer[128];
+
+   /**
+    * Pairing between types marshalled by the externalizers predefined via
+    * global configuration and their externalizer identifiers.
+    */
+   final IdentityIntMap<Class<?>> predefExtIds = new IdentityIntMap<>(32);
```
pruivo: I have another opinion here. Since the user-defined externalizers share the same interfaces than ours internal externalizers, they could be merge in the same data structure. We would end up with a ```Map<Class, AdvancedExternalizer>``` and an ```IntObjectHashMap<AdvancedExternalizer>```.
The ```InternalIds.INTERNAL``` and use and ```InternalIds.PREDEFINED``` could be merged.

core/src/main/java/org/infinispan/marshall/core/internal/InternalExternalizerTable.java
pruivo: :+1: to treat ```null``` in the ```Marshaller```

core/src/main/java/org/infinispan/marshall/core/internal/InternalExternalizerTable.java
pruivo: I think most of our classes uses the ```MarshallUtil```, but the problem relies on the classes with ```Object``` fields that can be anything, including a JDK collection.

core/src/main/java/org/infinispan/marshall/core/internal/InternalMarshaller.java
pruivo: how would it work? the most common classes have variable size (get, put, etc..., depends on the key size). Except if we start storing the key/values as ```byte[]``` always. Then, we don't need an external marshaller :)

core/src/main/java/org/infinispan/marshall/core/internal/InternalMarshaller.java
```diff
@@ -0,0 +1,227 @@
...
+      return marshallableTypeHints.getBufferSizePredictor(o.getClass());
+   }
+
+   @Override
+   public ByteBuffer objectToBuffer(Object o) throws IOException, InterruptedException {
+      BytesObjectOutput out = writeObjectOutput(o);
+      // No triming, just take position as length
+      return new ByteBufferImpl(out.bytes, 0, out.pos);
```
pruivo: I would like to see ```out.toByteBuffer()```

core/src/main/java/org/infinispan/statetransfer/TransactionInfo.java
```diff
@@ -78,7 +79,7 @@ public Integer getId() {
       public void writeObject(ObjectOutput output, TransactionInfo object) throws IOException {
          output.writeObject(object.globalTransaction);
          output.writeInt(object.topologyId);
-         output.writeObject(object.modifications);
+         MarshallUtil.marshallArray(object.modifications, output);
          output.writeObject(object.lockedKeys);
```
pruivo: can use MarshallUtil.marshallCollection()

galderz:
Some tests failing, I'll address those today as well as comments.

core/src/main/java/org/infinispan/marshall/exts/ArrayExternalizer.java
galderz: There are times when we just don't know the type, e.g. `StreamResponseCommand.response`. This can be anything from a primitive to an array of entries, so we need some fallback for when type might be an array and we can't know for sure in advance. If we don't do anything, it'd fallback on the external marshaller.

Also, a user could store an array of a type for which there's an Externalizer. Without this support, the user would need to give us an externalizer for marshalling the array type, or it would be delegated to the external marshaller (JBoss Marshaller in this case).

core/src/main/java/org/infinispan/marshall/core/internal/ExternalJBossMarshaller.java
galderz: As explained last week, we currently have an corner case in our code base that did not allow me to have a version of an external/user marshaller that simply used standard serialization (e.g. org.infinispan.marshall.core.internal.ExternalJavaMarshaller): 

Clustered Query implementation relies on some Lucene classes are Serializable but some of their inner fields are not. We have Externalizers for these inner fields, so in the previous marshaller implementation this work fine. However, when I tried to use a ExternalJavaMarshaller, this logic broke because I didn't have a way to fit in custom serialization logic inside to be able to piggyback to the externalizers. Since that logic is still in place, I made the external jboss marshaller a bit more complex to deal with it, by being able to plug at the `ObjectTable` level and do the corresponding externalizer lookup.

Also, I think we're mixing up some concepts here:

I differentiate between internal and external marshaller. The internal marshaller deals with types we know, either hardcoded (primitives) or via pluggable externalizers. The external marshaller deals with types we don't know. These types might be user defined keys and values, but as you can see above, there might be situations where we deal with 3rd party libraries internally (not only Lucene but you could think of alternative transaction manager implementations). For 3rd party libraries we either rely on standard Java Serialization or we provide Externalizers for those types.

So, I do agree with your point above, but unless clustered query issue is solved, or the functionality is removed, I can't see how I can get around it. @Sanne did say that the clustered query functionality is a little bit flaky, so it'd be very helpful to us to get rid of it.

Any other suggestions?

gustavonalle: As discussed in the chat, I spotted ``` org.hibernate.search.query.engine.impl.LuceneHSQuery``` as one of the offending classes, being ```Serializable``` but holding internally lucene queries which we do provide externalizers. So one way of solving it for now would be to provide externalizer for those classes as well. 
Going forward, I'd not like to remove ClusteredQuery but send a string representation of the query (IQL?) over the wire to broadcast it.

core/src/main/java/org/infinispan/marshall/core/internal/BytesObjectOutput.java
galderz: That's outside of the scope of this. Such requirement is already tracked.

core/src/main/java/org/infinispan/marshall/core/internal/InternalMarshaller.java
galderz: Hmmmm, can we move this out and flesh it in a separate JIRA? PositionalBuffer is an internal implementation detail so we can adjust it at a later time. By the way, there's nothing new about this logic. This is the same logic we've been using so far.

core/src/main/java/org/infinispan/marshall/core/internal/InternalMarshaller.java
galderz: @danberindei Ideas for this has been discussed in the marshalling doc. In fact, I prefer Sanne's suggestion in that document, where sizes are adjusted to the fragments of where it's going to be stored or shipped to. So yeah, outside the scope of this.

core/src/main/java/org/infinispan/marshall/core/internal/PrimitiveExternalizer.java
galderz: This particular Object[] might need removing since it's too generic, but the array externalizers deals with situations when we don't know types and might be arrays that includes types for which we have externalizers.

core/src/main/java/org/infinispan/marshall/core/internal/StreamBytesObjectOutput.java
galderz: Not always. JGroups' RpcDispatcher.Marshaller implementation calls objectToBuffer which does not any such wrapping.

galderz:
Unless I've made a comment, I agree with all the comments made. I'm going to make a separate branch, where I can tidy up the commits and I can sort out the issues mentioned. When that's ready I'll update the PR.

galderz:
As suggested by @danberindei, I've moved all added externalizers, minor command changes to use MarshallUtil and minor test enhancements into a separate PR so that it can be more easily discussed/integrated. The PR is #4620 

The rest of the changes will be added in a separate PR that will depend on #4620.

core/src/main/java/org/infinispan/marshall/core/internal/ExternalJBossMarshaller.java
galderz: As mentioned to Gustavo, this is not massively urgent. On one side, this logic of having to find externalizers from the external marshaller is limited to this particular use case, so it does not impact the rest of code. Also, we could have use that inadvertently are relying on this kind of logic, and that's the more worrying thing. In any case, the logic will remain for the time being, at least until we have an alternative default external marshaller, at which point we might drop it.

core/src/main/java/org/infinispan/marshall/core/internal/PrimitiveExternalizer.java
galderz: Btw, we could potentially have primitive section deal with generic arrays in case they're found by chance. Currently the primitive section deals with basic Object[] and primitive types, but JBoss Marshaller code goes beyond that to deal with any type of arrays. By doing this, we would effectively get rid of ArrayExternalizer. Nice to have potentially and I'll make sure I track it down.

To reiterate, when we know we have an array, we should call to MarshallUtil, but for those cases where the type is suddenly an array and we can't guess, we need to handle it.

galderz:
Closing. Please look at #4620, #4626 and #4627 (in that order).

PR #4533, opened 2016-08-26
ISPN-5721 ISPN-6813 SNI and encryption for REST

slaskawi:
https://issues.jboss.org/browse/ISPN-6813

Highlights:
* Added encryption for REST (some parts were there but it wasn't working).
* Upgraded to the latest Netty and RESTEasy versions.
* Added `ProtocolServerConfiguration` as parent of `RestServerConfiguration`. Some parts of the configuration throw `UnsupportedOperationException` since we don't control this configuration parts.
* Extracted common parts for encryption subsystem and encryption service.

### DIFF
--- a/server/rest/src/main/java/org/infinispan/rest/configuration/RestServerConfiguration.java
+++ b/server/rest/src/main/java/org/infinispan/rest/configuration/RestServerConfiguration.java
@@ -3,34 +3,27 @@ package org.infinispan.rest.configuration;
 import java.util.Set;
 
 import org.infinispan.commons.configuration.BuiltBy;
+import org.infinispan.server.core.configuration.ProtocolServerConfiguration;
+import org.infinispan.server.core.configuration.SslConfiguration;
 
 @BuiltBy(RestServerConfigurationBuilder.class)
-public class RestServerConfiguration {
+public class RestServerConfiguration extends ProtocolServerConfiguration {
    private final ExtendedHeaders extendedHeaders;
-   private final String host;
-   private final int port;
-   private Set<String> ignoredCaches;
 
-   RestServerConfiguration(ExtendedHeaders extendedHeaders, String host, int port, Set<String> ignoredCaches) {
+   RestServerConfiguration(ExtendedHeaders extendedHeaders, String host, int port, Set<String> ignoredCaches, SslConfiguration ssl) {
+      super(null, null, host, port, -1, -1, -1, ssl, false, -1, ignoredCaches);
       this.extendedHeaders = extendedHeaders;
-      this.host = host;
-      this.port = port;
-      this.ignoredCaches = ignoredCaches;
    }
 
    public ExtendedHeaders extendedHeaders() {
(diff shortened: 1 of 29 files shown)


tristantarrant:
I've pushed Will's Scala removal from the REST module, so you need to readapt it. The difference looks very minor.

slaskawi:
Rebased

server/rest/src/main/java/org/infinispan/rest/configuration/RestServerConfiguration.java
```diff
@@ -3,34 +3,19 @@
...
-
-   public int port() {
-      return port;
-   }
-
-   public String host() {
-      return host;
-   }
-
-   public Set<String> getIgnoredCaches() {
```
danberindei: I'd rather just deprecate it for 9.0.

slaskawi: The good thing is that it's mostly used internally so we can deprecate it whenever we want...

danberindei: We can deprecate anything whenever we want, removal is the tricky part :)

WDYT @tristantarrant, since we don't support running the servers by themselves, does it mean programmatic server configuration isn't part of our API and we don't need backwards compatibility?

slaskawi: @danberindei I believe we should deprecate it for ISPN 8.2.x series and remove in a separate PR in ISPN9. 

If you agree with me I will create proper JIRAs for it.

But before that - what is the reason of deprecation ignored caches? And what functionality needs removing it?

tristantarrant: First of all we cannot deprecate something after we release a .0. 
We are also not deprecating ignored caches. Leave that method in 9.0 and mark it as deprecated.

slaskawi: I double checked the code...

`RestServerConfiguration` now extends `ProtocolServerConfiguration` which has a public method called `ignoredCaches` ;) Looks like we are good to go...

slaskawi: @tristantarrant I've just checked - this PR integrates nicely into master... there's no conflict.

tristantarrant: Yes, but getIgnoredCaches() can't go away just yet unfortunately. Leave it as a proxy to ignoreCaches()

slaskawi: Done

tristantarrant:
Finally, pushed to master. Thanks for your patience

PR #4535, opened 2016-08-26
ISPN-6979 Missing PRIMARY_OWNER indexing mode in the xsd

gustavonalle:
https://issues.jboss.org/browse/ISPN-6979


anistor:
@gustavonalle I'm wondering if it would be possible to make the Indexing enum just ON/OFF and leave the ALL/LOCAL/PRIMARY to be determined internally instead of asking the use to decide.

gustavonalle:
@anistor As part of https://issues.jboss.org/browse/ISPN-6947, this configuration will be moved to the index level rather than cache level, and the idea is to assume defaults depending on the index manager, with a possibility to override it.

anistor:
LGTM then. Pulling...

anistor:

PR #4536, opened 2016-08-26
Enumeration of Hot Rod protocol versions should be an Enum

rhusar:
Stumbled upon this while integrating into WildFly. The user-specified protocol version needs to be validated, currently there is no available list of possibly versions. Looks like the idea may have started in org.infinispan.client.hotrod.Version but was never expanded and is out of date.


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/Configuration.java
```diff
@@ -157,7 +158,7 @@ public NearCacheConfiguration nearCache() {
       return nearCache;
    }
 
-   public String protocolVersion() {
+   public ProtocolVersion protocolVersion() {
```
tristantarrant: Let's not change this signature. Maybe add a separate method

rhusar: Ok, will do, so this is considered a public API thus should not change.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/ProtocolVersion.java
```diff
@@ -0,0 +1,45 @@
+package org.infinispan.client.hotrod;
+
+import java.util.EnumSet;
+import java.util.Locale;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
```
slaskawi: Could you please add some meaningful javadoc?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/ProtocolVersion.java
```diff
@@ -0,0 +1,45 @@
...
+   ProtocolVersion(int major, int minor) {
+      version = String.format(Locale.ROOT, "%d.%d", major, minor);
+   }
+
+   @Override
+   public String toString() {
+      return version;
+   }
+
+   private static final Map<String, ProtocolVersion> versions = EnumSet.allOf(ProtocolVersion.class).stream().collect(Collectors.toMap(ProtocolVersion::toString, Function.identity()));
```
slaskawi: Could you please move it up, where the other fields are placed?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/AbstractConfigurationChildBuilder.java
```diff
@@ -105,9 +106,15 @@ public ConfigurationBuilder marshaller(Marshaller marshaller) {
       return builder.marshaller(marshaller);
    }
 
+   @Deprecated
```
slaskawi: Please add some javadoc and information which method should be used instead.

slaskawi:
Nice work! Only some picky stuff found..

wburns:
Changes seem good overall to me as well.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/ProtocolVersion.java
rhusar: Done.

tristantarrant:
I added a commit on my branch, @rhusar please integrate them here https://github.com/tristantarrant/infinispan/commit/a38430926bd67c2a0bd21d5c8172e26fdb8d8d66

rhusar:
Pulled the commit here.

tristantarrant:
Assigned a Jira, fixed one last issue, squashed and merged. Thanks

rhusar:
Thanks Tristan.

PR #4538, opened 2016-08-29
ISPN-6977 ISPN-6983 ISPN-6984 Random test failures

danberindei:
https://issues.jboss.org/browse/ISPN-6977
https://issues.jboss.org/browse/ISPN-6983
https://issues.jboss.org/browse/ISPN-6984
https://issues.jboss.org/browse/ISPN-5424

ISPN-6977 NumOwnersNodeCrashInSequenceTest random failures
ISPN-6983 ThreeNodesSplitAndMergeTest random failures
ISPN-6984 ConcurrentJoinTest random failures
ISPN-5424 Replace SemaphoreCompletionService with LimitedExecutor

### DIFF
--- a/core/src/main/java/org/infinispan/executors/LimitedExecutor.java
+++ b/core/src/main/java/org/infinispan/executors/LimitedExecutor.java
@@ -1,0 +1,179 @@
+package org.infinispan.executors;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+
+import org.infinispan.IllegalLifecycleStateException;
+import org.infinispan.util.concurrent.CompletableFutures;
+import org.infinispan.util.concurrent.WithinThreadExecutor;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+import org.jboss.logging.NDC;
+
+/**
+ * Executes tasks in the given executor, but never has more than {@code maxConcurrentTasks} tasks running at the same
+ * time.
+ *
+ * A task can finish running without allowing another task to run in its stead, with {@link #executeAsync(Supplier)}.
+ * A new task will only start after the {@code CompletableFuture} returned by the task has completed.
+ *
+ * @author Dan Berindei
+ * @since 9.0
+ */
+public class LimitedExecutor implements Executor {
+   private static final Log log = LogFactory.getLog(LimitedExecutor.class);
+
+   private final Lock lock = new ReentrantLock();
+   private final Condition condition = lock.newCondition();
+   private final String name;
+   private final Executor executor;
+   private final boolean blocking;
+   private int availablePermits;
+   private final Deque<Runnable> queue = new ArrayDeque<>();
+   private final Runner runner = new Runner();
+
+   public LimitedExecutor(String name, Executor executor, int maxConcurrentTasks) {
+      this.name = name;
+      this.executor = executor;
+      this.availablePermits = maxConcurrentTasks;
+      this.blocking = executor instanceof WithinThreadExecutor;
+   }
+
+   /**
+    * When stopping, cancel any queued tasks.
+    */
+   public void cancelQueuedTasks() {
+      lock.lock();
+      try {
+         queue.clear();
+      } finally {
+         lock.unlock();
+      }
+   }
+
+   @Override
+   public void execute(Runnable command) {
+      if (blocking) {
+         CompletableFuture<Void> f1 = new CompletableFuture<>();
+         executeInternal(() -> {
+            f1.complete(null);
+            removePermit();
+         });
+         try {
+            CompletableFutures.await(f1);
+            command.run();
+         } catch (InterruptedException ie) {
+            Thread.currentThread().interrupt();
+            throw new IllegalLifecycleStateException(ie);
+         } catch (Exception e) {
+            log.debug("Exception in blocking task", e);
+         } finally {
+            addPermit();
+         }
+         return;
+      }
+      executeInternal(command);
+   }
+
+   public void executeInternal(Runnable command) {
+      lock.lock();
+      try {
+         queue.add(command);
+      } finally {
+         lock.unlock();
+      }
+      tryExecute();
+   }
+
+   public void executeAsync(Supplier<CompletableFuture<Void>> command) {
+      execute(() -> {
+         CompletableFuture<Void> future = command.get();
+         if (!future.isDone()) {
+            removePermit();
+            future.whenComplete(runner);
+         }
+      });
+   }
+
+   private void tryExecute() {
+      boolean addRunner = false;
+      lock.lock();
+      try {
+         if (availablePermits > 0) {
+            availablePermits--;
+            addRunner = true;
+         }
+      } finally {
+         lock.unlock();
+      }
+      if (addRunner) {
+         executor.execute(runner);
+      }
+   }
+
+   private void runTasks() {
+      while (true) {
+         Runnable runnable = null;
+         lock.lock();
+         try {
+            // If the previous task was asynchronous, we can't execute a new one on the same thread
+            if (availablePermits >= 0) {
+               runnable = queue.poll();
+            }
+            if (runnable == null) {
+               availablePermits++;
+               break;
+            }
+         } finally {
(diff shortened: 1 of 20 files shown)


core/src/main/java/org/infinispan/executors/LimitedExecutor.java
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * Executes tasks in the given executor, but never has more than {@code maxConcurrentTasks} tasks running at the same
+ * time.
+ *
+ * The number of concurrent tasks can be tweaked at any time with {@link #addPermit()} and {@link #removePermit()}.
+ *
+ * @author Dan Berindei
+ * @since 9.0
+ */
+public class LimitedExecutor implements Executor {
```
slaskawi: How about using CompletionService with ThreadPoolExecutor? Because if I understand the idea of this class correctly, we want to execute X tasks concurrently and store results in a list/queue?

core/src/test/java/org/infinispan/executors/LimitedExecutorTest.java
```diff
@@ -0,0 +1,93 @@
...
+      LimitedExecutor limitedExecutor = new LimitedExecutor(NAME, executor, 1);
+
+      CompletableFuture<Void> blockingFuture = new CompletableFuture<>();
+      CompletableFuture<String> cf1 = new CompletableFuture<>();
+      limitedExecutor.executeAsync(() -> blockingFuture.thenRunAsync(() -> cf1.complete("bla")));
+
+      CompletableFuture<String> cf2 = new CompletableFuture<>();
+      limitedExecutor.execute(() -> cf2.complete("bla"));
+
+      Thread.sleep(100);
```
slaskawi: Can we get rid of those sleeps somehow?

core/src/main/java/org/infinispan/executors/LimitedExecutor.java
danberindei: No, `SemaphoreCompletionService` stores the results in a queue, but `LimitedExecutor` doesn't keep track of results.

core/src/test/java/org/infinispan/executors/LimitedExecutorTest.java
danberindei: I can change them to `sleep(10)`, but there's no deterministic way to check that `cf2` does not complete before `blockingFuture`.

rvansa: @danberindei How about incrementing shared AtomicInteger and checking that the first lambda gets 1 and second gets 2?

core/src/main/java/org/infinispan/executors/LimitedExecutor.java
slaskawi: So why can't we use a pure Executor than (if we don't about the results)?

danberindei: Because I still want to limit the number of tasks that run concurrently. A `LimitedExecutor(1)` allows me to submit 2 tasks and runs only 1 at a time, but without creating a new thread for each `LimitedExecutor` instance. `executeAsync` is also something you don't get in a regular `Executor` :)

slaskawi: Ok understood. Thanks for clarification.

slaskawi:
LGTM, @pruivo, @rvansa ?

core/src/main/java/org/infinispan/executors/LimitedExecutor.java
```diff
@@ -0,0 +1,156 @@
...
+
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+import org.jboss.logging.NDC;
+
+/**
+ * Executes tasks in the given executor, but never has more than {@code maxConcurrentTasks} tasks running at the same
+ * time.
+ *
+ * The number of concurrent tasks can be tweaked at any time with {@link #addPermit()} and {@link #removePermit()}.
```
rvansa: `addPermit` and `removePermit` are private - how can you tweak it?

core/src/main/java/org/infinispan/statetransfer/InboundTransferTask.java
```diff
@@ -48,27 +50,9 @@
...
-    */
-   private volatile boolean isCompletedSuccessfully = false;
-
-   /**
     * This latch is counted down when all segments are completely received or in case of task cancellation.
     */
-   private final CountDownLatch completionLatch = new CountDownLatch(1);
-
-   private final StateConsumerImpl stateConsumer;
+   private final CompletableFuture<Void> completionLatch = new CompletableFuture<>();
```
rvansa: update field name

core/src/main/java/org/infinispan/statetransfer/InboundTransferTask.java
```diff
@@ -34,12 +34,14 @@
    private static final boolean trace = log.isTraceEnabled();
 
    /**
-    * All access to fields {@code segments} and {@code finishedSegments} must bead done while synchronizing on {@code segments}.
+    * All access to fields {@code segments} and {@code finishedSegments} must bead done while synchronizing on {@code
```
rvansa: Next time please do all the reformatting in separate commit, for easier review.

rvansa:
Generally OK, though the formatting changes are annoying. Does checkstyle enforce them?

slaskawi:
@danberindei Once the questions are answered and @rvansa is happy, I'll integrate it.

core/src/test/java/org/infinispan/executors/LimitedExecutorTest.java
```diff
@@ -0,0 +1,93 @@
...
+      CompletableFuture<String> cf1 = new CompletableFuture<>();
+      limitedExecutor.execute(() -> {
+         awaitLatch(latch);
+         cf1.complete("bla");
+      });
+
+      CompletableFuture<String> cf2 = new CompletableFuture<>();
+      limitedExecutor.execute(() -> cf2.complete("bla"));
+
+      Thread.sleep(10);
```
slaskawi: Can we use `Thread#yield` here? Just to let JVM continue running other thread?

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -944,50 +951,44 @@ private void removeStaleData(final Set<Integer> removedSegments) throws Interrup
...
          try {
-            InvalidateCommand invalidateCmd = commandsFactory.buildInvalidateCommand(EnumUtil.bitSetOf(CACHE_MODE_LOCAL, SKIP_LOCKING), keysToRemove.toArray());
+            InvalidateCommand invalidateCmd = commandsFactory
+                  .buildInvalidateCommand(EnumUtil.bitSetOf(CACHE_MODE_LOCAL, SKIP_LOCKING), keysToRemove.toArray());
             InvocationContext ctx = icf.createNonTxInvocationContext();
             ctx.setLockOwner(invalidateCmd.getKeyLockOwner());
             interceptorChain.invoke(ctx, invalidateCmd);
 
-            if (trace) log.tracef("Removed %d keys, data container now has %d keys", keysToRemove.size(), dataContainer.size());
+            if (trace)
```
slaskawi: Some time ago I complained about those constructs (`if(trace) log.tracef(...)`). What if someone changes the logging level in the runtime (via JMX for example)? This seems to be a normal thing to do...

I know it's a common practice to use this `if(trace)` things... but on the other hand I believe it is wrong...

rvansa: I think that performance tradeoff is more desirable than ability to change runtime level to TRACE. If you could demonstrate that this does not have any perf effect on modern JVMs by a stress test, the practice could be reconsidered.

slaskawi: Ok, sounds like a plan :)

core/src/test/java/org/infinispan/executors/LimitedExecutorTest.java
danberindei: We could, but then the test may not catch all situations where `LimitedExecutor` runs the task too soon. So I'd rather keep the `sleep(10)`.

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
danberindei: @slaskawi You can try, but I doubt it -- the difference is always going to be visible, especially for fast operations like replicated mode reads.

slaskawi:
Looking

slaskawi:
The PR & master builds have lots of failures... Let's give this one a go...

PR #4539, opened 2016-08-29
fix, apidocs link

kazuhira-r:
I fixed according to the change in the form of URL of a API document.


galderz:
Integrated, thanks @kazuhira-r :)

PR #4540, opened 2016-08-30
ISPN-6989 Performance guide

slaskawi:
https://issues.jboss.org/browse/ISPN-6989

### DIFF
--- a/documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
+++ b/documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
@@ -248,3 +248,24 @@ If your strings are mostly ASCII, convert them to `UTF-8` and store them as `byt
 
 * Using `String#getBytes("UTF-8")` allows to decrease size of the object
 * Consider using G1 GC with additional JVM flag -`XX:+UseStringDeduplication`. This allows to decrease memory footprint (see http://openjdk.java.net/jeps/192[JEP 192] for details).
+
+=== Use simple cache for local caches
+
+When you don't need the full feature set of caches, you can set local cache to "simple" mode and achieve non-trivial speedup while still using Infinispan API.
+
+This is an example comparison of the difference, randomly reading/writing into cache with 2048 entries as executed on 2&times;8-core Intel(R) Xeon(R) CPU E5-2640 v3 @ 2.60GHz:
+
+.Number of operations per second (± std. dev.)
+|===
+| Cache type   | single-threaded cache.get(...) | single-threaded cache.put(...) | 32 threads cache.get(...) | 32 threads cache.put(...)
+
+| Local cache  | 14,321,510 ± 260,807 |  1,141,168 ±  6,079 |   236,644,227 ± 2,657,918 |  2,287,708 ±   100,236
+| Simple cache | 38,144,468 ± 575,420 | 11,706,053 ± 92,515 |   836,510,727 ± 3,176,794 | 47,971,836 ± 1,125,298
+| CHM          | 60,592,770 ± 924,368 | 23,533,141 ± 98,632 | 1,369,521,754 ± 4,919,753 | 75,839,121 ± 3,319,835
+|===
+
+The CHM shows comparison for ConcurrentHashMap from JSR-166 with pluggable equality/hashCode function, which is used as the underlying storage in Infinispan.
+
+Even though we use http://openjdk.java.net/projects/code-tools/jmh/[JMH] to prevent some common pitfals of microbenchmarking, consider these results only aproximative. Your mileage may vary.
+
+Please consult http://infinispan.org/docs/{infinispanversion}/user_guide/user_guide.html#simple_cache[User's Guide] for the limitations of simple cache and configuration.
--- a/documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
+++ b/documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
@@ -1,0 +1,250 @@
+= Infinispan {infinispanversion} Performance Guide
+The Infinispan community
+:toc2:
+:icons: font
+:toclevels: 3
+:numbered:
+:stem:
+
+This guide will give you information and tweaks about tuning Infinispan performance (both server and library mode).
+
+== Capacity planning
+
+Data in Infinispan is either stored as plain Java objects or in a serialized form, depending on operating mode (embedded or server) or on specific configuration options (http://infinispan.org/docs/{infinispanversion}/user_guide/user_guide.html#store_as_binary[store-as-binary]).
+Data size can be estimated using sophisticated tools like  http://openjdk.java.net/projects/code-tools/jol/[Java Object Layout]
+and the total amount of required memory can be roughly estimated using the following formulas:
+
+[stem]
+.Total Data Set in library mode
+++++
+Total Data Set = Number Of Entries * (Key Size + Value Size + 200 b (Overhead))
+++++
+
+[stem]
+.Total Data Set in server mode
+++++
+Total Data Set = Number Of Entries * (Serialized Key Size + Serialized Value Size + 200 b (Overhead))
+++++
+
+Term `overhead` is used here as an average amount of additional memory (e.g. expiration or eviction data) needed for storing an Entry in a Cache.
+
+In case of Local or Replicated mode, all data needs to fit in memory, so calculating the amount of required memory is trivial.
+
+Calculating memory requirements for Distributed mode is slightly more complicated and requires using the following:
+
+[stem]
+.Required memory in Distributed mode
+++++
+Required Memo ry = Total Data Set*(Node Failures + 2)/(Nodes - Node Failures)
+++++
+
+Where:
+
+* `Total Data Set` - Estimated size of all data
+* `Nodes` - The number of nodes in the cluster
+* `Node Failures` - Number of possible failures (also `number of owners - 1`)
+
+Calculated amount of memory should be used for setting `Xmx` and `Xms` parameters.
+
+JVM as well as Infinispan require additional memory for other tasks like searches, allocating network buffers etc.
+It is advised to allocate no more than 50% of memory with living data when using Infinispan solely as a caching data store, and
+no more than 33% of memory with living data when using Infinispan to store and analyze the data using querying, distributed execution or distributed streams.
+
+When considering large heaps, make sure there's enough CPU to perform garbage collection efficiently.
+
+== Java Virtual Machine settings
+
+Java Virtual Machine tuning might be divided into sections like memory or GC.
+Below is a list of helpful configuration parameters and a guide how to adjust them.
+
+=== Memory settings
+
+Adjusting memory size is one of the most crucial step in Infinispan tuning. The most commonly used JVM flags are:
+
+* `-Xms` - Defines the minimum heap size allowed.
+* `-Xmx` - Defines the maximum heap size allowed.
+* `-Xmn` - Defines the minimum and maximum value for the young generation.
+* `-XX:NewRatio` - Define the ratio between young and old generations. Should not be used if -Xmn is enabled.
+
+Using `Xms` equal to `Xmx` will prevent JVM from dynamically sizing memory and might decrease GC pauses caused by resizing. 
+It is a good practice to specify `Xmn` parameter. This guaranteed proper behavior during load peak (in such case Infinispan generates lots of small, short living objects).
+
+=== Garbage collection
+
+The main goal is to minimize the amount of time when JVM is paused. Having said that, CMS is a suggested GC for Infinispan applications.
+
+The most frequently used JVM flags are:
+
+* `-XX:MaxGCPauseMillis` - Sets a target for the maximum GC pause time. Should be tuned to meet the SLA.
+* `-XX:+UseConcMarkSweepGC` - Enables usage of the CMS collector.
+* `-XX:+CMSClassUnloadingEnabled` - Allows class unloading when the CMS collector is enabled.
+* `-XX:+UseParNewGC` - Utilize a parallel collector for the young generation. This parameter minimizes pausing by using multiple collection threads in parallel.
+* `-XX:+DisableExplicitGC` - Prevent explicit garbage collections.
+* `-XX:+UseG1GC` - Turn on G1 Garbage Collector.
+
+=== Other settings
+
+There are two additional parameters which are suggested to be used:
+
+* `-server` - Enables server mode for the JVM.
+* `-XX:+ UseLargePages` - Instructs the JVM to allocate memory in Large Pages. These pages must be configured at the OS level for this parameter to function successfully.
+
+=== Example configuration
+
+In most of the cases we suggest using CMS. However when using the latest JVM, G1 might perform slightly better.
+
+.32GB JVM
+----
+-server
+-Xmx32G
+-Xms32G
+-Xmn8G
+-XX:+UseLargePages
+-XX:+UseConcMarkSweepGC
+-XX:+UseParNewGC
+-XX:+DisableExplicitGC
+----
+
+.32GB JVM with G1 Garbage Collector
+----
+-server
+-Xmx32G
+-Xms32G
+-Xmn8G
+-XX:+UseG1GC
+----
+
+== Network configuration
+
+Infinispan uses TCP/IP for sending packets over the network (for both cluster communication when using TCP stack or when communication with Hot Rod clients)
+
+In order to achieve the best results, it is recommended to increase TCP send and receive window size (refer to you OS manual for instructions).
+The recommended values are:
+
+* send window size - 640 KB
+* receive window size - 25 MB
+
+== Number of threads (Server mode only)
+
+Hot Rod Server uses worker threads which are activated by a client's requests.
+It's important to match the number of worker threads to the number of concurrent client requests:
+
+[source,java]
+.Hot Rod Server worker thread pool size
+----
+<hotrod-connector socket-binding="hotrod" cache-container="local" worker-threads="200">
+   <!-- Additional configuration here -->
+</hotrod-connector>
+----
+
+== Cache Store performance
+
+In order to achieve the best performance, please follow the recommendations below when using Cache Stores:
+
+* Use async mode (write-behind) if possible
+* Prevent cache misses by preloading data
+* For JDBC Cache Store:
+** Use indexes on `id` column to prevent table scans
+** Use PRIMARY_KEY on `id` column
+** Configure batch-size, fetch-size, etc
+
+== Hints for program developers
+
+There are also several hints for developers which can be easily applied to the client application and will boost up the performance.
+
+=== Ignore return values
+
+When you're not interested in returning value of the `#put(k, v)` or `#remove(k)` method, use `Flag.IGNORE_RETURN_VALUES` flag as shown below:
+
+[source,java]
+.Using Flag.IGNORE_RETURN_VALUES
+----
+Cache noPreviousValueCache = cache.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES);
+noPreviousValueCache.put(k, v);
+----
+
+It is also possible to set this flag using ConfigurationBuilder
+
+[source,java]
+.Using ConfigurationBuilder settings
+----
+ConfigurationBuilder cb = new ConfigurationBuilder();
+cb.unsafe().unreliableReturnValues(true);
+----
+
+=== Use `Externalizer` for marshalling
+
+Infinispan uses JBoss Marshalling to transfer objects over the wire. The most efficient way to marshall user data is to provide an AdvancedExternalizer.
+This solutions prevents JBoss Marshalling from sending class name over the network and allows to save some bandwidth:
+
+[source,java]
+.User entity with Externalizer
+----
+import org.infinispan.marshall.AdvancedExternalizer;
+
+public class Book {
+
+   final String name;
+   final String author;
+
+   public Book(String name, String author) {
+      this.name = name;
+      this.author = author;
+   }
+
+   public static class BookExternalizer
+            implements AdvancedExternalizer<Book> {
+
+      @Override
+      public void writeObject(ObjectOutput output, Book book)
+            throws IOException {
+         output.writeObject(book.name);
+         output.writeObject(book.author);
+      }
+
+      @Override
+      public Person readObject(ObjectInput input)
+            throws IOException, ClassNotFoundException {
+         return new Person((String) input.readObject(), (String) input.readObject());
+      }
+
+      @Override
+      public Set<Class<? extends Book>> getTypeClasses() {
+         return Util.<Class<? extends Book>>asSet(Book.class);
+      }
+
+      @Override
+      public Integer getId() {
+         return 2345;
+      }
+   }
+}
+----
+
+The Externalizer must be registered in cache configuration. See configuration examples below:
+
+[source,java]
+.Adding Externalizer using XML
+----
+<cache-container>
+   <serialization>
+      <advanced-externalizer class="Book$BookExternalizer"/>
+   </serialization>
+</cache-container>
+----
+
+[source,java]
+.Adding Externalizer using Java
+----
+GlobalConfigurationBuilder builder = ...
+builder.serialization().addAdvancedExternalizer(new Book.BookExternalizer());
+----
+
+For more information, please refer to the link:$$../user_guide/user_guide.html#benefits_of_externalizers$$[User's Guide].
+
+=== Storing Strings efficiently
+
+If your strings are mostly ASCII, convert them to `UTF-8` and store them as `byte[]`:
+
+* Using `String#getBytes("UTF-8")` allows to decrease size of the object
+* Consider using G1 GC with additional JVM flag -`XX:+UseStringDeduplication`. This allows to decrease memory footprint (see http://openjdk.java.net/jeps/192[JEP 192] for details).
--- a/documentation/TODO.md
+++ b/documentation/TODO.md
@@ -134,6 +134,4 @@
 * Even questions that are related to product (sme-jdg, customer portal, etc) can be of use (obviously purged from any product reference)
 * Things we CANNOT do are also important, and we should point out any limitations
 
-## Performance Guide (Sebastian)
-* This should be linked in the appropriate places from the User Guide
 
--- a/documentation/TODO.md
+++ b/documentation/TODO.md
@@ -57,10 +57,6 @@
 ## Locking and concurrency (Pedro)
 * The Versioning paragraph mentions tombstones as if they exist.
 
-### Local mode cache (Radim)
-* The simple cache documentation is erroneously included in the functional map API chapter. Move it here
-* Update the information inside the performance paragraph to include results from the simple cache benchmarks
-
 ### Query + Indexing (Adrian & Gustavo)
 * Add missing parts (CQ)
 
--- a/documentation/src/main/asciidoc/faqs/faqs.asciidoc
+++ b/documentation/src/main/asciidoc/faqs/faqs.asciidoc
@@ -180,9 +180,15 @@ If, for put operations, you don't need the previous values existing in the cache
 ----
 
 Note that in this case the value returned by `cache.put()` is not reliable.
-This optimization skips a cache store read and can have very significant performance improvement effects. 
+This optimization skips a cache store read and can have very significant performance improvement effects.
+
+For more information, check out our link:$$../performance_guide/performance_guide.html$$[Performance Guide].
 
 TIP: More flags are described at link:$$../user_guide/user_guide.html#_invocation_flags$$[Per-Invocation Flags]
+
+=== How to speed up Infinispan?
+
+Have a look at our link:$$../performance_guide/performance_guide.html$$[Performance Guide].
 
 === Locking and Transaction questions
 ==== Does Infinispan support distributed eager locking?
--- a/documentation/src/main/asciidoc/index.asciidoc
+++ b/documentation/src/main/asciidoc/index.asciidoc
@@ -18,6 +18,7 @@ Infinispan.
  * A link:getting_started/getting_started.html[Getting Started Guide] (link:https://docs.jboss.org/infinispan/{infinispanversion}/pdf/getting_started.pdf[PDF])
  * A link:user_guide/user_guide.html[User Guide] (link:https://docs.jboss.org/infinispan/{infinispanversion}/pdf/user_guide.pdf[PDF])
  * A link:server_guide/server_guide.html[Server Guide] if you are running the JBoss AS-based server components (link:https://docs.jboss.org/infinispan/{infinispanversion}/pdf/server_guide.pdf[PDF])
+ * A link:performance_guide/performance_guide.html[Performance Guide] (link:https://docs.jboss.org/infinispan/{infinispanversion}/pdf/performance_guide.pdf[PDF])
  * An link:upgrading/upgrading.html[Upgrading Guide] (link:https://docs.jboss.org/infinispan/{infinispanversion}/pdf/upgrading.pdf[PDF])
  * A pretty useful link:glossary/glossary.html[Glossary] (link:https://docs.jboss.org/infinispan/{infinispanversion}/pdf/glossary.pdf[PDF])
  * A set of link:faqs/faqs.html[Frequently Asked Questions] (link:https://docs.jboss.org/infinispan/{infinispanversion}/pdf/faqs.pdf[PDF])
--- a/documentation/src/main/asciidoc/performance_guide/performance_guide-docinfo-footer.html
+++ b/documentation/src/main/asciidoc/performance_guide/performance_guide-docinfo-footer.html
@@ -1,0 +1,11 @@
+<script>
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+  ga('create', 'UA-8601422-1', 'auto');
+  ga('send', 'pageview');
+
+</script>
+
--- a/documentation/src/main/asciidoc/performance_guide/performance_guide-docinfo.html
+++ b/documentation/src/main/asciidoc/performance_guide/performance_guide-docinfo.html
@@ -1,0 +1,6 @@
+<link rel="canonical" href="http://infinispan.org/docs/stable/performance_guide/performance_guide.html">
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jstree/3.3.1/themes/default/style.min.css">
+<link rel="stylesheet" href="../css/css.css">
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jstree/3.3.1/jstree.min.js"></script>
+<script src="../js/js.js"></script>
--- a/documentation/src/main/asciidoc/user_guide/cache_api.adoc
+++ b/documentation/src/main/asciidoc/user_guide/cache_api.adoc
@@ -19,6 +19,8 @@ Attempting to perform these operations globally would have large performance imp
 
 It should be noted that using certain flags with the link:http://docs.jboss.org/infinispan/{infinispanversion}/apidocs/org/infinispan/AdvancedCache.html#withFlags-org.infinispan.context.Flag...-[withFlags] method can mitigate some of these concerns, please check each method's documentation for more details.
 
+For more performance tips, have a look at our link:$$../performance_guide/performance_guide.html$$[Performance Guide].
+
 ==== Mortal and Immortal Data
 Further to simply storing entries, Infinispan's cache API allows you to attach mortality information to data.  For example, simply using link:http://docs.oracle.com/javase/6/docs/api/java/util/Map.html#put%28K,%20V%29[put(key, value)] would create an _immortal_ entry, i.e., an entry that lives in the cache forever, until it is removed (or evicted from memory to prevent running out of memory).  If, however, you put data in the cache using link:http://docs.jboss.org/infinispan/{infinispanversion}/apidocs/org/infinispan/commons/api/BasicCache.html#put-K-V-long-java.util.concurrent.TimeUnit-[put(key, value, lifespan, timeunit)] , this creates a _mortal_ entry, i.e., an entry that has a fixed lifespan and expires after that lifespan.
 


documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+[stem]
+++++
+(Total Data Set)/(Nodes)*(1+(Node Failures)/(Nodes - Node Failures))
+++++
+
+Where:
+
+* `Total Data Set` - Estimated size of all data
+* `Nodes` - The number of nodes in the cluster
+* `Node Failures` - Number of failures possible (also number of owners)
```
rvansa: number of owners == node failures + 1

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+----
+cache.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES)
+----
+
+=== Skip cache load
+
+When you don't want to load an entry from a Cache Store, use `Flag.SKIP_CACHE_LOAD` as shown below:
+
+[source,java]
+.Flag.IGNORE_RETURN_VALUES
```
rvansa: Skip cache load, not ignore return values. However, this can break things.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+
+      @Override
+      public Integer getId() {
+         return 2345;
+      }
+   }
+}
+----
+
+The Externalizer must be registered in cache configuration.
```
rvansa: example of configuration

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done.

Yeah, I've seen the discussion on dev mailing list.

slaskawi:
Fixed comments. Ready for round #2.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,197 @@
...
+In order to achieve the best results, it is recommended to increase TCP send and receive window size (refer to you OS manual for instructions).
+The recommended values are:
+
+* send window size - 640 KB
+* receive window size - 25 MB
+
+== Number of threads (Server mode only)
+
+Hot Rod Server uses worker threads which are activated by a client's requests.
+It's important to match the number of worker threads to the number of concurrent client requests:
```
gustavonalle: The threading in Netty does not follow the thread-per-request model, and many cache operations are executed directly in the EventLoop thread and thus not via worker pool, so not sure if this applies

slaskawi: Hmmm so any suggestions how I should rephrase it?

slaskawi: [JDG7 Perf Guide](https://access.redhat.com/documentation/en-US/Red_Hat_JBoss_Data_Grid/7.0/pdf/Performance_Tuning_Guide/Red_Hat_JBoss_Data_Grid-7.0-Performance_Tuning_Guide-en-US.pdf) says this:

> Worker threads, unlike system threads, are threads activated by a client's request and do not interact
with the user. As the server is asynchronous client requests will continue to be received after this limit
is hit; instead, this number represents the number of active threads performing simultaneous
operations, typically writes.

Maybe we should say it should match the number of concurrent write requests?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,197 @@
...
+:numbered:
+:stem:
+
+This guide will give you information and tweaks about tuning Infinispan performance (both server and library mode).
+
+== Capacity planning
+
+Infinispan stores data in memory. In case of Local or Replicated mode, all data needs to fit in there,
+so calculating the amount of required memory is trivial.
+
```
gustavonalle: It'd be interesting to put some guidelines in order to estimate or calculate the amount of memory. Need to take into account the (serialized or not) key size, (serialized or not) value size, the Infinispan entry overhead (200 bytes?), and the JVM overhead (object header, alignment, etc). This can be done with estimation using tools like http://openjdk.java.net/projects/code-tools/jol/ or actually measuring the heap with a subset of data.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
gustavonalle: With the latest server changes, AFAICT, not even writes are done in the worker threads unless there is a listener attached to the cache, everything happens in the IO threads that is not configurable afaict (fixed at 2 * processors) so I'd say changing the worker threads size does not make any difference at all.
@wburns WDYT?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Added proper paragraphs about that.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+== Capacity planning
+
+Infinispan stores data in memory. Depending on operating mode data is stored in serialized form (server mode) or directly
+directly in memory (library mode). Data size can be estimated using sophisticated tools like  http://openjdk.java.net/projects/code-tools/jol/[Java Object Layout]
+and total amount of required memory can be calculated using the following formulas:
+
+[stem]
+.Total Data Set in library mode
+++++
+Total Data Set = Number Of Entries * (Key Size + Value Size + 200 b (Me tadata))
```
danberindei: s/Me tadata/Metadata/

How precise is this overhead number? I believe it actually depends a lot on what whether you are using expiration, versioning, the functional API etc.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+++++
+
+[stem]
+.Total Data Set in server mode
+++++
+Total Data Set = Number Of Entries * (Serialized Key Size + Serialized Value Size + 200 b (Me tadata))
```
danberindei: Again, versioning and compatibility mode can change the overhead.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+
+
+In case of Local or Replicated mode, all data needs to fit in memory, so calculating the amount of required memory is trivial.
+
+Calculating memory requirements for Distributed mode is slightly more complicated and requires using the following:
+
+[stem]
+.Required memory in Distributed mode
+++++
+Required Memo ry = (Total Data Set)/(Nodes)*(1+(Node Failures)/(Nodes - Node Failures))
```
danberindei: I think this should be more like `(Total Data Set) * (Number of Owners + 1) / (Minimum Number of Nodes)`.

Saying minimum number of nodes covers both when nodes crash and when the cluster is scaled down. The `+ 1` is to allow for rebalancing, as during rebalancing some segments will have an extra copy. To be fair, rebalancing also adds some extra overhead, like keeping a list of all the keys that have been modified, so it's hard to estimate accurately.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+Where:
+
+* `Total Data Set` - Estimated size of all data
+* `Nodes` - The number of nodes in the cluster
+* `Node Failures` - Number of possible failures (also `number of owners - 1`)
+
+Calculated amount of memory should be used for setting `Xmx` and `Xms` parameters.
+
+JVM as well as Infinispan require additional memory for other tasks like searches, allocating network buffers etc.
+It is advised to allocate no more than 50% of memory with living data.
```
danberindei: @alanfx I think it was even less than 50%...

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+
+When considering large heaps, make sure there's enough CPU to perform garbage collection efficiently.
```
danberindei: I wouldn't mention CPUs, memory bandwidth is usually more important for garbage collection. I'd just mention that long GC pauses can negatively impact the stability of the cluster (in addition to the latency of the application using it), so they should size their heaps accordingly.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+=== Memory settings
+
+Adjusting memory size is one of the most crucial step in Infinispan tuning. The most commonly used JVM flags are:
+
+* `-Xms` - Defines the minimum heap size allowed.
+* `-Xmx` - Defines the maximum heap size allowed.
+* `-Xmn` - Defines the minimum and maximum value for the young generation.
+* `-XX:NewRatio` - Define the ratio between young and old generations. Should not be used if -Xmn is enabled.
+
+Using `Xms` equal to `Xmx` will prevent JVM from dynamically sizing memory and might decrease GC pauses cased by resizing.
```
danberindei: s/cased/caused/, but TBH I doubt resizing causes any GC pauses... starting with a smaller heap does mean you have a smaller young generation and more old GCs, but on the other hand those old GCs might be faster because the old gen is also smaller.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+
+=== Garbage collection
+
+The main goal is to minimize the amount of time when JVM is paused. Having said that, CMS is a suggested GC for Infinispan applications.
+
+The most frequently used JVM flags are:
+
+* `XX:MaxGCPauseMillis` - Sets a target for the maximum GC pause time. Should be tuned to meet the SLA.
```
danberindei: -XX

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+* `-XX:+ UseConcMarkSweepGC` - Enables usage of the CMS collector.
+* `-XX:+ CMSClassUnloadingEnabled` - Allows class unloading when the CMS collector is enabled.
+* `-XX:+ UseParNewGC` - Utilize a parallel collector for the young generation. This parameter minimizes pausing by using multiple collection threads in parallel.
+* `-XX:+ DisableExplicitGC` - Prevent explicit garbage collections.
```
danberindei: No space after `+`

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+.32GB JVM
+----
+-server
+-Xmx32G
+-Xms32G
+-Xmn8G
+-XX:+UseLargePages
+-XX:+UseConcMarkSweepGC
+-XX:+UseParNewGC
+-XX:+DisableExplicitGC
```
danberindei: Personally I'd remove the definitions of the JVM flags, and instead just provide an example of "expected" JVM configuration.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+----
+
+== Network configuration
+
+Infinispan uses TCP/IP for sending packets over the network.
```
danberindei: Or UDP, if you configure the UDP stack ;)

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+** Use PRIMARY_KEY on `id` column
+** Configure batch-size, fetch-size, etc
+
+== Hints for program developers
+
+There are also several hints for developers which can be easily applied to the client application and will boost up the performance.
+
+=== Ignore return values
+
+When you're not interested in returning value by `#put` or `#remove` method, use `Flag.IGNORE_RETURN_VALUES` flag as shown below:
```
danberindei: s/in returning value by/in the return value of the/
Also, I'd rather use `Cache.put(k, v)` in the docs instead of `#put`.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+
+[source,java]
+.Flag.IGNORE_RETURN_VALUES
+----
+cache.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES)
```
danberindei: Looks a bit incomplete...

    Cache noPreviousValueCache = cache.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES);
    noPreviousValueCache.put(k, v);

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+----
+
+=== Skip cache load
+
+When you don't want to load an entry from a Cache Store, use `Flag.SKIP_CACHE_LOAD` as shown below:
```
danberindei: I wouldn't mention `SKIP_CACHE_LOAD` here. It's already covered by `IGNORE_RETURN_VALUES` for `put`, and for `get` you have to really know what you're doing in order to use it.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+
+[source,java]
+.Flag.SKIP_CACHE_LOAD
+----
+cache.getAdvancedCache().withFlags(Flag.SKIP_CACHE_LOAD)
+----
+
+=== Use `Externalizable` for marshalling
+
+Infinispan uses JBoss Marshalling to transfer objects over the wire. The most efficient way to marshall user data is to provide an Externalizer:
```
danberindei: I'd say more about the advantage of **`AdvancedExternalizer`** vs `Serializable`/`Externalizable`: for `Serializable`/`Externalizable`, JBoss Marshalling always sends the name of the class over the wire, but with `AdvancedExternalizer` it only sends the externalizer id. AFAIK our non-advanced `Externalizer` behaves almost exactly like `Externalizable`.

The disadvantage of `AdvancedExternalizer` is that it duplicates objects that appear twice in the object graph, and doesn't work at all the object graph has cycles -- you have to use `InstanceReusingAdvancedExternalizer` instead. Of course, it's always better if you can serialize just the root of the graph and have it write its children as primitives.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+----
+GlobalConfigurationBuilder builder = ...
+builder.serialization().addAdvancedExternalizer(new Book.BookExternalizer());
+----
+
+=== Storing Strings efficiently
+
+When the majority of stored data are Strings there are additional optimizations which can be taken into consideration:
+
+* Using `String#getBytes("UTF-8")` allows to decrease size of the object
```
danberindei: I'd rephrase this as "If your strings are mostly ASCII, convert them to `UTF-8` and store them as `byte[]`. Note that the `UTF-8` version can be bigger for non-ASCII strings."

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
alanfx: @danberindei it was 50% of the heap when only put, get, and remove are used. More like 30%-33% of the heap when analytics are being used in the cache. (i.e. querying, distributed execution, streams, etc.)

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+:icons: font
+:toclevels: 3
+:numbered:
+:stem:
+
+This guide will give you information and tweaks about tuning Infinispan performance (both server and library mode).
+
+== Capacity planning
+
+Infinispan stores data in memory. Depending on operating mode data is stored in serialized form (server mode) or directly
```
tristantarrant: "on the operating mode"

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+directly in memory (library mode). Data size can be estimated using sophisticated tools like  http://openjdk.java.net/projects/code-tools/jol/[Java Object Layout]
```
tristantarrant: "directly" repeated

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
tristantarrant: Actually this needs a little more clarification:
Data in Infinispan is either stored as plain Java objects or in a serialized form, depending on operating mode (embedded or server) or on specific configuration options (store-as-binary).

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
tristantarrant: Also you provide a way to estimate the size of a POJO using JOL, but this doesn't provide information about serialized size, which depends on the marshaller. It would be nice to have a similar tool

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: > s/Me tadata/Metadata/

This is correct :) [Asciimath](http://asciimath.org/) replaces `eta` with `η` :) Don't ask me why they implement it this way...

The overhead numbers are taken from the product documentation (see links in the JIRA). I guess it was estimated or measured.... Maybe @alanfx has some ideas how to put some meaningful information here...

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: I changed phrase:

> and total amount of required memory can be calculated using the following formulas

to 

> and total amount of required memory can be roughly estimated using the following formulas

This should emphasize that this is only an estimate.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Ok leaving 50% than

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: I disagree with you here... I would like to avoid people creating VMs (e.g. using OpenStack) with 32GB or RAM and 1 vCPU for example. In that case GC would be terribly slow. 

Also, you often can not change memory bandwidth... But on the other hand you can [measure it](https://specs.openstack.org/openstack/nova-specs/specs/liberty/approved/mem-bw.html).

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Could you please elaborate more? What do you mean by "expected" JVM configuration? Can you share an example (which I could use in this manual)?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
danberindei: I guess it makes some expressions more compact, but I'm guessing the real reason is that they don't split on word boundaries first...

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: If I see correctly, the changes you're proposing are:

- Replacing `NodeFailures` with `NumberOfOwners` (as a reminder - `NodeFailures = NumberOfOwners - 1`
- Introducing `MinimumNumberOfNodes`, which is equal to `Nodes - NodeFailures`

I have no strong opinion for the former (I'm OK with either `NumberOfOwners` and `NodeFailures`) but I'm against the latter. Often times when you create a system you're thinking about the possible failures and put VMs (or physical machines) in different racks. So I believe system architects or developers are much more familiar with the number of node failures.

You also changed the equation a little bit, so let me try to do some math on concrete numbers:
```
Total Data Set = 10 GB
Number of Owners = 2
Node Failures = Number of Owners - 1 = 2 - 1 = 1
Nodes = 10
MinumumNumberOfNodes = 10 - 1 = 9

Original formula: (10/10) * (1+(1/(10-1))) = 1 * (1+1/9) = 1.11 GB
Dan's formula: 10 * (2+1) / 9 = 3.33 GB
```

I think the amount of required memory is a bit high and I also don't understand the part with `(number of owners + 1)`.... Could you please elaborate more about this? What is the motivation behind this?

//cc @alanfx @mgencur @rvansa

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
danberindei: Most of the time, you can also split the 32GB VM into two 16GB VMs. IMO that's usually more effective for reducing the GC pause than adding CPUs.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Fixed `caused` thanks!

If I interpret the [CMS](https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/cms.html) manual correctly:

> Concurrently resize the heap and prepare the support data structures for the next collection cycle, using one processor.

That means that the additional space will be available after full collection. So I guess this paragraph is still valid. However it's a good point - "Java Performance - The Definitive Guide" suggests using `Xmn` when making both `Xms` and `Xmx` equal. I'm adding a small note on that.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Yes. Also this is the case when using client/server mode. Added note about that.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
danberindei: I meant the example you already have, but without the sections with the JVM flag definitions. I guess you could also use the  defaults from the server.

Perhaps removing all the definitions is a bit extreme, so maybe I'd keep a short explanation of the settings used in the example, but I wouldn't try to explain flags that we're not using.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: @rvansa also mentioned that it can break stuff. I agree - let's remove it.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
danberindei: "concurrently" means there's no pause

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: That's a lot of very useful information :) I'll try to summarize that and hide some of the details you mentioned (I wouldn't like to mention too much of our internal behavior since it can get out of date very quickly)

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: That would be very nice.... However I have no idea whether such tool exists... Maybe @galderz could help?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Yes but those resized heap will be available in the next collection cycle, so I interpret is after next full GC. 

Anyway - So do you propose to remove this whole fragment about `Xmx = Xms`? Or only part of it?

@alanfx - May I ask you for some input as well?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Hmmm I still do not quite understand. 

All flags have been described in previous paragraphs (like `Garbage collection` or `Memory settings`).

I'm also not sure what JVM flag definitions you have in mind and what defaults from the server (do you mean Infinispan server?).

Maybe if you could give me an example here it would be crystal clear.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
gustavonalle: Minimally you just need to:
```
GenericJBossMarshaller marshaller = new GenericJBossMarshaller();
ByteBuffer byteBuffer = marshaller.objectToBuffer( /* my custom object */);
int size = byteBuffer.getLength();
```

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: I'm personally OK with this. However since we are suggesting something else than for the produce I would like to hear @alanfx and @tristantarrant opinion.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
danberindei: Yes, you have described all the flags in the example, but also other flags, like `-XX:+CMSClassUnloadingEnabled`, and because they're not in the example it's not clear for users whether they should add it. That's why I'd like to start with our example/default JVM settings, and only explain those settings that we actually use. E.g.

    == JVM configuration

    This is an example configuration for a JVM with 32GB of heap:

        -server
        -Xmx32G

    * `-server`: select server mode, which enables the `C2` compiler. Already enabled by default on Linux.
    * `-Xmx32G`: set the maximum heap size.

    For more information, see the http://www.oracle.com/technetwork/articles/java/vmoptions-jsp-140102.html

And yes, by defaults I meant the defaults from our Infinispan Server startup scripts. Unfortunately I can't find anything that doesn't use `-Xmx512m` ATM...

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
alanfx: Actually, @mgencur has a test to check for this overhead, since it changes from release to release. I do not believe that it looks at all possible combinations of overhead, so maybe @danberindei we should discuss all of the configurations where the overhead should be checked?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
alanfx: I'd prefer if the text described both situations! "...no more than 50% of memory with living data when using Infinispan solely as a caching data store, and no more than 33% of memory with living data when using Infinispan to store and analyze the data using querying, distributed execution or distributed streams."?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
alanfx: @danberindei's math makes more sense to me. And to combine it with my recommendation above, you'd actually need 6.67GB or 9.99GB of RAM per node based the intended Infinispan usage.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,224 @@
...
+=== Memory settings
+
+Adjusting memory size is one of the most crucial step in Infinispan tuning. The most commonly used JVM flags are:
+
+* `-Xms` - Defines the minimum heap size allowed.
+* `-Xmx` - Defines the maximum heap size allowed.
+* `-Xmn` - Defines the minimum and maximum value for the young generation.
+* `-XX:NewRatio` - Define the ratio between young and old generations. Should not be used if -Xmn is enabled.
+
+Using `Xms` equal to `Xmx` will prevent JVM from dynamically sizing memory and might decrease GC pauses caused by resizing.
```
alanfx: This is the only thing I see in the Oracle docs about setting `Xms` equal to `Xmx`: 

> Setting -Xms and -Xmx to the same value increases predictability by removing the most important sizing decision from the virtual machine. However, the virtual machine is then unable to compensate if you make a poor choice.

From: https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/sizing.html#sthref22

I don't think this really implies that it improves performance significantly, just that it avoids work in JVM.

mgencur: I remember the old times when we were testing that. And using same value for -Xms and -Xmx really improved performance significantly. I don't have the numbers though.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: @danberindei @alanfx I would also mention only some average values here.... The client needs to be able to roughly estimate the amount of required memory. Calculating this on a byte level makes no sense ;)

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Sounds good to me... @tristantarrant ?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done, thanks!

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Ok got it now, thanks for being patient ;)

Maybe let's split this into two pieces - for server mode use settings similar to this and for library use this...

Let me write it in the docs. If it still won't fit your expectations we'll refactor it further...

slaskawi: Argh... It probably won't be that easy because if we squashed all Java sub-paragraphs (`Memory settings`, `Garbage collection` and `Other settings`) we will lose some background information like this:

> The main goal is to minimize the amount of time when JVM is paused. Having said that, CMS is a suggested GC for Infinispan applications.

or this:

> Using `Xms` equal to `Xmx` will prevent JVM from dynamically sizing memory and might decrease GC pauses caused by resizing. 
Also when using this combination, it is advices to specify the `Xmn` parameter as well. 

So the question is, do we still want to do it?

@alanfx - Would be great to hear your input here also.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: @alanfx @mgencur So is it OK to leave it as is?

alanfx: @slaskawi yes, I think it is ok to leave it as is

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Assuming @tristantarrant if fine with this... updating...

slaskawi: I changed the formula to: `Required Memory = Total Data Set*(Node Failures + 2)/(Nodes - Node Failures)`. As you can see I preserved node failures since I believe they are more descriptive. Please let me know if you don't like it.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: @danberindei, @alanfx Thoughts?

slaskawi:
Updated, ready for the next round.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
alanfx: @slaskawi I did comment about about setting `-Xmx` and `-Xms` to the same value. I think it is fine to keep that in the docs. I do wonder about changing to use the G1 collector by default, but that will require changing the server startup scripts as well and diverge from Wildfly. Not sure we want to do that yet, but G1 did perform better than CMS in my last round of testing.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
danberindei: @slaskawi Sure, we don't need to be too precise, but if we have a test that users can measure the exact overhead with, it would be nice to include it here.

Thinking about this some more, I'd like to change the formula to reference an "Overhead" term, and define it in the paragraph above. I'd call it "Overhead" instead of "Metadata", because a basic embedded cache without versioning and expiration doesn't have any metadata, but it still has an overhead for each entry.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
danberindei: ok

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
danberindei: ok

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
danberindei: Ok to keep it as is, and add `-XX:+UseG1GC` as an option alongside CMS.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
galderz: Yeah, I'm not sure whether providing the same amount on `Xms` and `Xmx` makes any difference these days. We should test this.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,227 @@
...
+Using `Xms` equal to `Xmx` will prevent JVM from dynamically sizing memory and might decrease GC pauses caused by resizing. 
+Also when using this combination, it is advices to specify the `Xmn` parameter as well. 
+
+=== Garbage collection
+
+The main goal is to minimize the amount of time when JVM is paused. Having said that, CMS is a suggested GC for Infinispan applications.
+
+The most frequently used JVM flags are:
+
+* `-XX:MaxGCPauseMillis` - Sets a target for the maximum GC pause time. Should be tuned to meet the SLA.
```
galderz: Does this option really work?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,227 @@
...
+* `-XX:+UseConcMarkSweepGC` - Enables usage of the CMS collector.
+* `-XX:+CMSClassUnloadingEnabled` - Allows class unloading when the CMS collector is enabled.
```
galderz: Is this option more designed for managed environments when there are deploy/undeploy cycles?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
galderz: Hmmm, I thought @wburns had changed things so that we don't execute any cache operations in the event loop any more...?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,227 @@
...
+When you're not interested in returning value of the `#put(k, v)` or `#remove(k)` method, use `Flag.IGNORE_RETURN_VALUES` flag as shown below:
+
+[source,java]
+.Flag.IGNORE_RETURN_VALUES
+----
+Cache noPreviousValueCache = cache.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES);
+noPreviousValueCache.put(k, v);
+----
+
+=== Use `Externalizable` for marshalling
```
galderz: You mean `Externalizer`?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,227 @@
...
+</cache-container>
+----
+
+[source,java]
+.Adding Externalizer using Java
+----
+GlobalConfigurationBuilder builder = ...
+builder.serialization().addAdvancedExternalizer(new Book.BookExternalizer());
+----
+
```
galderz: Maybe you could add a link to the docu on externalizers if the reader is interested in more details?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
alanfx: In the testing I did, this option did have some positive effect on the throughput, but it is a "goal" and not a "hard limit". The GC will try to meet the goal if it can, but there are no guarantees.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: @wburns Any update on this?

slaskawi: From the code I can see that worker threads are used to construct an `Executor`: https://github.com/infinispan/infinispan/blob/master/server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java#L187-L188

Then we are using this executor to process client requests: https://github.com/infinispan/infinispan/blob/966569022711401369a1a2c33704dd8e5ce2c44f/server/hotrod/src/main/java/org/infinispan/server/hotrod/transport/HotRodChannelInitializer.java#L50-L50 and https://github.com/infinispan/infinispan/blob/966569022711401369a1a2c33704dd8e5ce2c44f/server/hotrod/src/main/java/org/infinispan/server/hotrod/ContextHandler.java#L58-L58

So unless I misinterpret it - this paragraph as it is, makes perfectly sense.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Ok, I changed it to `Overhead` and added a short description.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: I added another example configuration with G1. I hope that's fine...

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: @galderz I believe it does, see: https://github.com/infinispan/infinispan/pull/4540#r79192021

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: @alanfx Maybe we should say something like this... In any case please use CMS but if your JDK is greater than version X (could you please tell me which JDK it was?), than use G1.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: I think a very good explanation is here: https://stackoverflow.com/questions/3334911/what-does-jvm-flag-cmsclassunloadingenabled-actually-do

In short - I think it should stay because of classes created on runtime using dynamic languages like Groovy. It also won't hurt any traditional app...

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Good point! Fixed, thanks!

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: I added configuration example (with G1) and a note that in the latest JVMs this GC might perform slightly better.

slaskawi:
Updated, ready for the next round.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,227 @@
...
+:icons: font
+:toclevels: 3
+:numbered:
+:stem:
+
+This guide will give you information and tweaks about tuning Infinispan performance (both server and library mode).
+
+== Capacity planning
+
+Data in Infinispan is either stored as plain Java objects or in a serialized form, depending on operating mode (embedded or server) or on specific configuration options (store-as-binary).
```
tristantarrant: Link to the user guide: ../user_guide/user_guide.html#store_as_binary

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,227 @@
...
+:numbered:
+:stem:
+
+This guide will give you information and tweaks about tuning Infinispan performance (both server and library mode).
+
+== Capacity planning
+
+Data in Infinispan is either stored as plain Java objects or in a serialized form, depending on operating mode (embedded or server) or on specific configuration options (store-as-binary). 
+Data size can be estimated using sophisticated tools like  http://openjdk.java.net/projects/code-tools/jol/[Java Object Layout]
+and total amount of required memory can be roughly estimated using the following formulas:
```
tristantarrant: and *the* total amount of memory

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,241 @@
...
+
+Adjusting memory size is one of the most crucial step in Infinispan tuning. The most commonly used JVM flags are:
+
+* `-Xms` - Defines the minimum heap size allowed.
+* `-Xmx` - Defines the maximum heap size allowed.
+* `-Xmn` - Defines the minimum and maximum value for the young generation.
+* `-XX:NewRatio` - Define the ratio between young and old generations. Should not be used if -Xmn is enabled.
+
+Using `Xms` equal to `Xmx` will prevent JVM from dynamically sizing memory and might decrease GC pauses caused by resizing. 
+Also when using this combination, it is advices to specify the `Xmn` parameter as well.
```
tristantarrant: it is **advised** to specify

Also, I'd explain why specifying Xmn is beneficial (generally because a lot of the allocations have a very short lifespan, and during high loads we generate a lot of them which, with a small NewSize would erroneously get promoted to older generations)

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
tristantarrant: I think we really need to revise our startup scripts for server, since they are sized far too large for a "starter" configuration.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,241 @@
...
+=== Ignore return values
+
+When you're not interested in returning value of the `#put(k, v)` or `#remove(k)` method, use `Flag.IGNORE_RETURN_VALUES` flag as shown below:
+
+[source,java]
+.Flag.IGNORE_RETURN_VALUES
+----
+Cache noPreviousValueCache = cache.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES);
+noPreviousValueCache.put(k, v);
+----
```
tristantarrant: Should we mention that this flag can be turned on everywhere using the **unsafe** configuration options ?

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,241 @@
...
+
+[source,java]
+.Adding Externalizer using Java
+----
+GlobalConfigurationBuilder builder = ...
+builder.serialization().addAdvancedExternalizer(new Book.BookExternalizer());
+----
+
+For more information, please refer to the http://infinispan.org/docs/dev/user_guide/user_guide.html#benefits_of_externalizers[User's Guide].
```
tristantarrant: Don't use absolute links, but ones relative to this doc, so they always reference the correct version

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
gustavonalle: There's a ```local-handler``` before that processes requests in the IO loop thread to avoid the latency penalty of the executor. I could not find the code that conditionally disable the executor based on the presence/absence of listeners, maybe this was not done in master

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
alanfx: @slaskawi I think mentioning the "latest JVMs" is a good option. I ran my tests with JVM version 1.8.0.91, but I'm not sure that's worth mentioning.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
danberindei: Actually, it should be `AdvancedExternalizer` :)

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: I replaces the last sentence with this:

> It is a good practice to specify `Xmn` parameter. This guaranteed proper behavior during load peak (in such case Infinispan generates lots of small, short living objects).

I hope this will do the trick.

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
slaskawi: Done

slaskawi:
@tristantarrant I addressed your comments and updated the PR.

Other changes introduced:
* Removed TODO entry from documentation mini sprint
* Referenced performance guide from User Guide as well as FAQ

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,250 @@
...
+----
+
+[source,java]
+.Adding Externalizer using Java
+----
+GlobalConfigurationBuilder builder = ...
+builder.serialization().addAdvancedExternalizer(new Book.BookExternalizer());
+----
+
+For more information, please refer to the http://infinispan.org/docs/{infinispanversion}/user_guide/user_guide.html#benefits_of_externalizers[User's Guide].
```
tristantarrant: Here you should use a relative link:
link:$$../user_guide/user_guide.html$$[User's Guide]

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,271 @@
...
+Term `overhead` is used here as an average amount of additional memory (e.g. expiration or eviction data) needed for storing an Entry in a Cache.
+
+In case of Local or Replicated mode, all data needs to fit in memory, so calculating the amount of required memory is trivial.
+
+Calculating memory requirements for Distributed mode is slightly more complicated and requires using the following:
+
+[stem]
+.Required memory in Distributed mode
+++++
+Required Memo ry = Total Data Set*(Node Failures + 2)/(Nodes - Node Failures)
```
tristantarrant: Extra space in the word 'Memory'

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
```diff
@@ -0,0 +1,271 @@
...
+
+Where:
+
+* `Total Data Set` - Estimated size of all data
+* `Nodes` - The number of nodes in the cluster
+* `Node Failures` - Number of possible failures (also `number of owners - 1`)
+
+Calculated amount of memory should be used for setting `Xmx` and `Xms` parameters.
+
+JVM as well as Infinispan require additional memory for other tasks like searches, allocating network buffers etc.
```
tristantarrant: Better write it as: "Infinispan will also require additional memory..."

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
tristantarrant: It looks like this is rendered using MathJax in the browser (which looks fine) but is blank in the PDF...

documentation/src/main/asciidoc/performance_guide/performance_guide.asciidoc
tristantarrant: You missed this

tristantarrant:
Pushed to master, thanks !

PR #4541, opened 2016-08-31
ISPN-4390 Realigned with ConcurrentHashMap from the JSR-166 repository

tristantarrant:
https://issues.jboss.org/browse/ISPN-4390
https://issues.jboss.org/browse/ISPN-6933

Cross-port the ConcurrentHashMap from Doug Lea's JSR-166 repo (specifically from the src/jdk8/java/util/concurrent path, revision 1.3)
Also include some necessary classes: Helper, ThreadLocalRandom
Extract the Unsafe logic to an independent class
Use the original spacing to ease future backporting
Rename the jdk8backported package to jsr166mods to avoid confusion


danberindei:
@tristantarrant GitHub won't show a diff for `EquivalentConcurrentHashMapV8.java`, says it's too large. I think we could move most of the stuff between `BOUNDED_MOD BEGIN` and `BOUNDED_MOD END` to separate files, WDYT?

We could also remove most of `ThreadLocalRandom`, since we use only the `Unsafe` bits that access the probe field in `Thread`.

And finally, CVS tracks versions only at the file level. The directory version doesn't change changes unless you add/remove a file, so you need to mention the version of each file separately instead of "src/jdk8/java/util/concurrent version 1.3".

tristantarrant:
> won't show a diff for EquivalentConcurrentHashMapV8.java, says it's too large

That is also because previous versions used our 3-space tab setting, and now I'm not reformatting the upstream code

> I think we could move most of the stuff between BOUNDED_MOD BEGIN and BOUNDED_MOD END to separate files, WDYT?

Yes, let's ask @wburns why he didn't do it last time.

> We could also remove most of ThreadLocalRandom, since we use only the Unsafe bits that access the probe field in Thread.

Yes, I wanted to change as little as possible, but in this case we need to not confuse users that this is a usable ThreadLocalRandom

> And finally, CVS tracks versions only at the file level. The directory version doesn't change changes unless you add/remove a file, so you need to mention the version of each file separately instead of "src/jdk8/java/util/concurrent version 1.3".

My comments actually refer to the original file, not the directory

tristantarrant:
@danberindei since we will be replacing the bounded stuff with caffeine soon anyway, let's not pursue further refactorings. We could also skip this for master and only keep it for 8.2.x

danberindei:
Ok @tristantarrant. @wburns do you have any more comments?

wburns:
@danberindei, @tristantarrant Doing the changes for 8.2.x sounds good to me.  Also moving the stuff to separate files would be good and help reducing noise in the future.

tristantarrant:
Closing this in favor of https://github.com/infinispan/infinispan/pull/4560 for 8.2.x

PR #4543, opened 2016-09-01
ISPN-6906 Single put key command

galderz:
This PR builds on the work on `t_6906` branch, which should be reviewed in [here](https://github.com/infinispan/infinispan/pull/4530) and adds the following:

* A single put key/value command has been added designed to reduce wrapping 
  and cost associated with a simple cache.put(k, v) call when the return value is 
  unused.
* Command is sent as is, without the need of being wrapped in a single
  RPC command.
* Command implements AdvancedExternalizer, avoiding a map lookup when
  writing the command.
* When reading the command from the binary payload, the externalizer ID
  is derived from the command id, shifting it 4 bytes. This opens the
  possibility of having externalizers for commands whose shifted id
  expands between 256 and 4095.
* This command reduces the instance variables and serialization to the
  bare minimum as long as the following conditions are met:
** Non transactions are used.
** Cache is clustered.
** Default metadata is used.
** Return value is not used.
** Key and value equivalences are default.

Some of these ideas could be applied more globally, e.g.
* Making commands implement `AdvancedExternalizer`, avoiding map lookup on write.
* Avoid wrapping commands in SingleRpcCommand, just send them as is.

Here are ideas I think we should flesh out more, but they're a bit more involved, so I'd like to wait and get a 1st round of performance feedback:
* I don't see why SinglePutKeyValueCommand should have an instance variable for the cache name. If you're sending 1000 puts for the same cache, that's 1000 instance variables with the cache name that don't add anything. IOW, the dispatcher or higher level could pass down the cache name and write it, and then delegate to the command itself.
* Again, I don't see why a command should carry the address of origin. For the origin, the address is simply used to pass that information from the IO layer to the runnable layer at which the operation is executed via the interceptor stack in the remote side. IOW, this instance variable does nothing on the sender side. Address of origin should be a parameter to BaseBlockingRunnable constructor. The only difference here is that put command would have one less instance variable in both sender and receiver, and instead that instance variable is only kept in BaseBlockingRunnable, which is run in receiver.

In essence, the command associated with a cache.put() like the one indicated above should only carry: key, value and command invocation id. All of those are specific to a particular put call and those are the only instance variables that make sense having there.

My local tests show this PR enhances performance on the existing `t_6906` branch for put only tests that I've been focusing on. `t_6906_singleput` branch shows around ~10% improvement for put only tests with custom pojos in my local server but the performance improvement needs to be validated further.


core/src/main/java/org/infinispan/commands/write/SinglePutKeyValueCommand.java
```diff
@@ -0,0 +1,307 @@
...
+ * - A replicable, non-tx put does not need to be wrapped in another command, it can be sent directly.
+ * - Single metadata is default, no metadata information needs to be serialized.
+ * - A non-conditional put always matches value, so is always successful.
+ * - If value equivalence is default, no need to serialize that info.
+ * - Only ignore return values flag set, so no need to serialize that info.
+ */
+public final class SinglePutKeyValueCommand
+   implements
+      DataWriteCommand, RemoteLockCommand, FlagAffectedCommand,
+      AdvancedExternalizer<SinglePutKeyValueCommand>,
```
rvansa: I personally find having the externalizer as inner class nicer. Let one class do the least it has to do.

core/src/main/java/org/infinispan/marshall/core/internal/InternalExternalizerTable.java
```diff
@@ -176,6 +181,10 @@ void stop() {
       if (clazz == null || primitives.getTypeClasses().contains(clazz)) {
          out.writeByte(InternalIds.PRIMITIVE);
          ext = (Externalizer<T>) primitives;
+      } else if (obj instanceof AdvancedExternalizer) {
+         ext = (Externalizer<T>) obj;
```
rvansa: You could cast it right away to AdvancedExternalizer

core/src/main/java/org/infinispan/commands/write/SinglePutKeyValueCommand.java
rvansa: Ok, now I understand why you made it directly the externalizer. Though, if you want to go this direction, I would prefer implementing just `AdvancedExternalizable` interface that would return an instance (singleton) of the externalizer.

rvansa: Especially when `getId()` is a method that can cause conflicts quite easily.

rvansa:
Would it be possible to make this BasicPutKeyValueCommand and inherit from it in PutKeyValueCommand?

galderz:
> Would it be possible to name this BasicPutKeyValueCommand and inherit from it in PutKeyValueCommand?

I wanted to avoid extending other commands because they bring a lot of baggage. SinglePutKeyValueCommand has just 4 instance variables, whereas PutKeyValueCommand has 11. Why pay the cost of 11 when I only need 4? I just need a command, that has no superclasses, which is final, and implements the necessary interfaces.

Granted, adding a new command adds extra visitor callbacks...etc, which we would avoid by extending PKVC. However, the aim of this particular branch is to see how much improvement we can get by having a dedicated command with minimal content, marshalling...etc.

Finally, I still need to add a commit to this branch that's already present in `t_6906` but which I'm waiting for confirmation that it improves things.

rvansa:
@galderz You took my question wrong; I was asking to use the inheritance in the other way - PKVC would inherit from SPKVC and add the 7 vars, SPKVC can be kept simple with those 4 vars.

PR #4546, opened 2016-09-02
[8.2.x] ISPN-6996 Update to Netty 4.1.5.Final

gustavonalle:
https://issues.jboss.org/browse/ISPN-6996

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -74,7 +74,7 @@
       <version.jboss.logging>3.3.0.Final</version.jboss.logging>
       <version.jgroups>3.6.7.Final</version.jgroups>
       <version.jta>1.0.1.Final</version.jta>
-      <version.netty>4.0.36.Final</version.netty>
+      <version.netty>4.1.5.Final</version.netty>
       <version.osgi>4.3.1</version.osgi>
       <version.rhq>4.4.0</version.rhq>
       <version.scala.major>2.11</version.scala.major>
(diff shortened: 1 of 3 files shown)


tristantarrant:
Pushed to 8.2.x, thanks

johnou:
Why not stick with Netty 4.0.x (still maintained!) in 8.2.x branch and keep Netty 4.1.x in master? Netty is used in a lot of third party libraries so updating the Netty library by a major version in an Infinispan patch version might come as an unpleasant surprise to developers.

gustavonalle:
Netty is an internal library of the Infinispan server and should be "invisible" to users, and going for 4.0 to 4.1 is updating by a minor version, not a major. Do you have any issues with the upgrade?

johnou:
oops, I had meant minor version. Regardless, it is not backwards compatible which is why I added the comment here. That said, given it is internal and I do not see it exposed in any of the infinispan pom.xmls that I use I think it's ok to leave it as is. Thanks for replying and the explanation!

rvansa:
@johnou Does Netty use a different versioning semantics? (since you say that minor versions are not backwards compatible)

gustavonalle:
This is the official statement from Netty:
```
Although we did our best to keep the backward compatibility from 4.0, 4.1 contains
multiple additions  which might not be fully backward-compatible with 4.0. 
Please make sure to recompile your application against the new version.
```

johnou:
@rvansa yes there are breaking changes moving from 4.0 to 4.1.

PR #4547, opened 2016-09-02
ISPN-6961 AbstractClusterListenerUtilTest tests fail from IDE

danberindei:
https://issues.jboss.org/browse/ISPN-6961

Remove the resetTimeServices method, it's not needed and it fails
with an IllegalArgumentException if the test didn't advance() all the
ControlledTimeService instances.

Don't log marshalling errors during shutdown.


galderz:
Integrated

PR #4548, opened 2016-09-02
ISPN-6997PessimisticTxPartitionAndMergeDuringRuntimeTest

danberindei:
.testOriginatorIsolatedPartition random failures

https://issues.jboss.org/browse/ISPN-6997

On one node, transactionTable.cleanupLeaverTransactions was being
called only before the split and after the merge, never with the list
of members during the split.

### DIFF
--- a/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
+++ b/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
@@ -111,7 +111,7 @@ public class TransactionTable implements org.infinispan.transaction.TransactionT
 
    private ConcurrentMap<Transaction, LocalTransaction> localTransactions;
    private ConcurrentMap<GlobalTransaction, LocalTransaction> globalToLocalTransactions;
-   private ConcurrentMap<GlobalTransaction, RemoteTransaction> remoteTransactions;
+   protected ConcurrentMap<GlobalTransaction, RemoteTransaction> remoteTransactions;
    private Lock minTopologyRecalculationLock;
    protected boolean clustered = false;
    protected volatile boolean running = false;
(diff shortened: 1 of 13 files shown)


danberindei:
I've looked some more into this issue and it's more complex than I thought: a remote tx command can create a new remote transaction even when the tx originator is in another partition. If the partitions merge immediately afterwards, neither the originator nor the remote node's `TransactionTable` will clean up the transaction.

core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
```diff
@@ -277,37 +277,41 @@ public int getMinTopologyId() {
    }
 
    public void cleanupLeaverTransactions(List<Address> members) {
-      // Can happen if the cache is non-transactional
-      if (remoteTransactions == null)
-         return;
+      try {
+         // Can happen if the cache is non-transactional
+         if (remoteTransactions == null)
```
slaskawi: Why did you move it inside the `try` block?

danberindei: I added the try block to make sure we don't throw any exception here that would interrupt the topology update in `StateConsumerImpl`.

slaskawi: We won't throw anything with the `return` instruction ;)

danberindei: I see what you mean now... I was confused about the "move" part, because there was no `try` block before :)

galderz:
@danberindei Shall we close this PR until you've found a better solution?

slaskawi:
One last call @danberindei :)

danberindei:
I'm giving it one more try today @slaskawi

galderz:
And that day was gone... a while back. Closing for now. Reopen when this is ready.

PR #4550, opened 2016-09-02
ISPN-6999 Reduce compatibility allocations from async interceptors

wburns:
* Only do processing for remote nodes
* Collections reduce scope to prevent lambda allocations


wburns:
Should be stable, but let tests run first.

tristantarrant:
Pushed to master. Thanks

PR #4551, opened 2016-09-07
ISPN-7003 NamedTestMethod must implement equals()

danberindei:
https://issues.jboss.org/browse/ISPN-7003


tristantarrant:
Pushed, thanks Dan

PR #4553, opened 2016-09-07
ISPN-6981 AffinityIndexManager fails to index documents in async mode

gustavonalle:
https://issues.jboss.org/browse/ISPN-6981
https://issues.jboss.org/browse/ISPN-7001

### DIFF
--- a/query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
+++ b/query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
@@ -9,6 +9,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -20,9 +21,11 @@ import javax.transaction.TransactionManager;
 import org.apache.lucene.index.IndexWriter;
 import org.apache.lucene.search.similarities.Similarity;
 import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.LockObtainFailedException;
 import org.hibernate.search.backend.IndexingMonitor;
 import org.hibernate.search.backend.LuceneWork;
 import org.hibernate.search.engine.service.spi.ServiceManager;
+import org.hibernate.search.exception.SearchException;
 import org.hibernate.search.indexes.spi.DirectoryBasedIndexManager;
 import org.hibernate.search.spi.SearchIntegrator;
 import org.hibernate.search.spi.WorkerBuildContext;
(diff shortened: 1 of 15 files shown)


gustavonalle:
Updated

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -70,37 +75,55 @@
...
          Thread.sleep(POLL_WAIT);
       } catch (InterruptedException e) {
          Thread.currentThread().interrupt();
       }
       Address lockHolder = getLockHolder(getIndexName());
       clearIfNeeded(lockHolder);
+      retryOperations(failingOperations);
+   }
+
+   private void retryOperations(List<LuceneWork> failingOperations) {
```
slaskawi: I think we should have some kind of counter here (or there is one but I can't see it).... It should retry X times and then probably give up.

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
```diff
@@ -155,26 +179,23 @@ private void handleOwnershipAcquired() {
...
-            }
-         }
-         if (wasInterrupted) {
-            Thread.currentThread().interrupt();
+      InfinispanDirectoryProvider directoryProvider = (InfinispanDirectoryProvider) getDirectoryProvider();
+      int activeDeleteTasks = directoryProvider.pendingDeleteTasks();
+      boolean wasInterrupted = false;
+      while (activeDeleteTasks > 0) {
+         try {
+            Thread.sleep(POLL_WAIT);
```
slaskawi: What will happen if there will be a deadlocked delete task? If I read the code correctly we will be spinning here forever? Is there any way to avoid this?

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
gustavonalle: The retries are already being done until it succeeds. Each ```performOperations``` calls involves a series of calls inside hibernate search. When an error occurs there, the callback method ```AffinityIndexManager#operationsFailed``` is called and then is retried again. Retries are done only for lock obtention related issues, so if for some reason the index is completely corrupt, it will not retry and log an error

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
gustavonalle: I don't think those delete tasks will deadlock (or at least was not able to reproduce it under load), they are a bunch of ```cache.remove``` being done by a separate executor.
I will cap that waiting time until replacing this executor with something better.

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
slaskawi: Ok than! Thanks for clarification!

query/src/main/java/org/infinispan/query/affinity/AffinityIndexManager.java
slaskawi: Let me add also @danberindei and @rvansa to this thread.

gustavonalle: Created https://issues.jboss.org/browse/ISPN-7025

danberindei: @slaskawi @gustavonalle In theory you could have a "deadlock" if a remote node throws an `OutdatedTopologyException` during the remove, but the local node can't install the newer topology because you're blocking it. These removals are triggered from a `@TopologyChanged` listener, which means they do block new topology updates (`LocalTopologyManagerImpl` uses a `SemaphoreCompletionService` to serialize topology updates, including the notifications). OTOH, `StateTransferInterceptor` doesn't block forever, it times out after `clustering.remoteTimeout`, so it's not a real deadlock.

gustavonalle:
Updated with a timeout when waiting for delete tasks, until ISPN-7025 is done

tristantarrant:
Nice one. Pushed to master.

PR #4554, opened 2016-09-09
Remove @Deprecated getProperties() method from RemoteCacheContainer interface.

pferraro:
Since this is a new interface, there's no reason to introduce deprecated methods.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
@@ -104,19 +104,6 @@ public class RemoteCacheManager implements RemoteCacheContainer {
    }
 
    /**
-    * Retrieves a clone of the properties currently in use.  Note that making
-    * any changes to the properties instance retrieved will not affect an
-    * already-running RemoteCacheManager.
-    *
-    * @return a clone of the properties used to configure this RemoteCacheManager
-    * @since 4.2
-    */
-   @Deprecated
-   public Properties getProperties() {
-      return configuration.properties();
-   }
-
-   /**
     * Same as {@link RemoteCacheManager(java.util.Properties)}, but it will try to lookup the config properties in the
     * classpath, in a file named <tt>hotrod-client.properties</tt>. If no properties can be found in the classpath, the
     * server tries to connect to "127.0.0.1:11222" in start.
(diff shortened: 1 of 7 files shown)


rhusar:
👍

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
```diff
@@ -103,11 +103,14 @@ public Configuration getConfiguration() {
...
+    * Retrieves a clone of the properties currently in use.  Note that making
+    * any changes to the properties instance retrieved will not affect an
+    * already-running RemoteCacheManager.
+    *
+    * @return a clone of the properties used to configure this RemoteCacheManager
     * @since 4.2
     */
-   @Override
    @Deprecated
    public Properties getProperties() {
```
slaskawi: I'm just thinking out laud - since this method was deprecated in `RemoteCacheContainer` and is deprecated now here - why don't we remove it?

rhusar: Might as well, @galderz @wburns ?

wburns: It has been deprecated for a while, so I would be fine for its removal.

pferraro: Removing this method from RemoteCacheManager can be done independently of this PR, which is only concerned with the RemoteCacheContainer interface.

slaskawi: Yes it can... but it would be like doing only haft of the work... Is it problematic to remove it here for some reason?

pferraro: This method is used by 2 unit tests in the infinispan-cdi-remote module.

tristantarrant: Any calls can be replaced with getConfiguration().properties();

slaskawi: Understood. May I ask you to create a JIRA for it? Otherwise LGTM.

tristantarrant:
Pushed to master as https://issues.jboss.org/browse/ISPN-7059

PR #4559, opened 2016-09-16
ISPN-6869 Synchronize starting and stopping DefaultCacheManager

slaskawi:
https://issues.jboss.org/browse/ISPN-6869

An approach based on 2 locks.

### DIFF
--- a/core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
+++ b/core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
@@ -20,6 +20,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.infinispan.Cache;
 import org.infinispan.IllegalLifecycleStateException;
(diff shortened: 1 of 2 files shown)


vjuranek:
@slaskawi maybe I'm missing something, but this IMHO doesn't prevent deadlock described in ISPN-6869 -  AFAIKT new cache can be still started while cache manager stop is in progress and thus result into the deadlock.

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
```diff
@@ -135,8 +135,8 @@
    private final CacheContainerStats stats;
    private final ConfigurationManager configurationManager;
 
-   @GuardedBy("this")
-   private boolean stopping;
+   private final Lock stopSemaphore = new ReentrantLock();
```
rvansa: semaphore?

slaskawi: Just a name... I can change it to lock :)

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
```diff
@@ -687,18 +693,22 @@ private void terminate(String cacheName) {
    public void stop() {
       authzHelper.checkPermission(AuthorizationPermission.LIFECYCLE);
 
-      synchronized (this) {
-         if (stopping) {
-            log.trace("Ignore call to stop as the cache manager is stopping");
-            return;
+      //if start is in progress we need to wait until it's finished
+      startSemaphore.lock();
```
slaskawi: Here is a place where the code will block waiting for start to finish

vjuranek: well, but it's a lock for staring cache manager, not a cache - sorry, still unclear to me, how it helps

slaskawi:
I was wrong here... It's a matter of stopping CacheManager and starting Caches (not CacheManager).. sorry for confusion!

PR #4563, opened 2016-09-22
ISPN-7031 Client listener failover w/ network down

galderz:
https://issues.jboss.org/browse/ISPN-7031

* Handle client listener failover situations when the network goes down
  but the server is still running. Such situations are handled by
  attempting to do a client listener failover.
* If the client listener fails over, e.g. if no servers can be reached,
  the failover is delayed until the next successful connection can be
  stablished.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/TcpTransportFactory.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/TcpTransportFactory.java
@@ -10,6 +10,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


galderz:
I couldn't think of an easy way to create a unit test for this, so if anyone has suggestions, I'm all ears.

slaskawi:
The only way I can think of is to hook into Netty's pipeline on the server side and close the connection (possibly using a custom `Handler`).

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/TcpTransportFactory.java
```diff
@@ -92,6 +95,9 @@
    private Map<byte[], Boolean> compatibilityCaches = CollectionFactory
            .makeMap(ByteArrayEquivalence.INSTANCE, AnyEquivalence.getInstance());
 
+   private final BlockingQueue<AddClientListenerOperation> disconnectedListeners =
```
slaskawi: How about using `LinkedBlockingQueue` here? It's based on linked list and we never know how many disconnected listeners we will have...

gustavonalle:
Your reviews have been transferred to another PR https://github.com/infinispan/infinispan/pull/4599

PR #4564, opened 2016-09-22
Refactor command execution patterns

rvansa:
Depends on #4493 (so you may ignore first two commits).

Most weight of this PR is in 78a1297725f8e3691b0a9934b93a6c2b7799e66f, copypasting commit description:
- EntryWrappingInterceptor has well-defined behaviour (see javadoc)
- Add VisitableCommand.executionType() that captures the routing &
  execution pattern of commands in single enum.
- Remove VisitableCommand.alwaysReadsExistingValues().
- Deprecate CacheEntry.rollback(), StateChangingEntry and its methods
  - COPIED flag in ReadCommittedEntry
- Remove oldValue from ReadCommittedEntry, add
  MVCCEntry.setValue(v, isRetry) and getValue(isRetry) with some logic
  in commands that returns correct previous value upon retry (with
  repeatable reads - without them, entry is removed from ctx before retry).
- Make ClusteredGetCommand topology-aware (add topologyId).
- Functional commands, GetAllCommand and PutMapCommand rewritten to use
  RpcManager.invokeRemotelyAsync and behave asynchronously.
- PutMapCommmand (optionally) returns List<Map.Entry> instead of Map
  to share code with functional commands (and generally be more inline).
- (Non-clustered) listeners now fire only on the owners, not on originator.
- Make DeltaAwareCacheEntry MVCC (though the plan is to remove it altogether).


rvansa:
TBH I've expected that there'll be more lines removed than added, since I've organized things... Probably too many comments and javadocs :)

Anyway, failures in `ClusteredCacheWithAffinityIndexManagerTest` are unrelated, that's ISPN-7005. The other two worry me a bit, especially the JCache one, but I can't reproduce them even when running the test in a loop.

rvansa:
Rebased, but includes #4577 and #4606 test fixes - merge those first, please.

pruivo:
Compile error:
```
[ERROR] /unsecure/pedro/work/infinispan/server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheCommands.java:[35,40] cannot find symbol
  symbol:   class ActivationInterceptor
  location: package org.infinispan.interceptors.impl
[ERROR] /unsecure/pedro/work/infinispan/server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheMetricsHandler.java:[42,40] cannot find symbol
  symbol:   class ActivationInterceptor
  location: package org.infinispan.interceptors.impl

```

core/src/main/java/org/infinispan/commands/read/AbstractLocalCommand.java
```diff
@@ -17,7 +17,7 @@
  * @author Mircea.Markus@jboss.com
  * @since 4.1
  */
-public abstract class AbstractLocalCommand extends AbstractLocalFlagAffectedCommand implements LocalCommand {
+public abstract class AbstractLocalCommand extends AbstractFlagAffectedCommand implements LocalCommand {
```
pruivo: Can you remove the static method in the end of the class?

pruivo:
For some reason, this file is created and empty:
```core/src/main/java/org/infinispan/interceptors/impl/BaseAsyncInvocationContext.java```

File: core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifierImpl.java [hunk 2]
```diff
@@ -348,7 +348,7 @@ protected final void resumeIfNeeded(Transaction transaction) {
    @Override
    public void notifyCacheEntryCreated(K key, V value, Metadata metadata, boolean pre,
                                        InvocationContext ctx, FlagAffectedCommand command) {
-      if (!cacheEntryCreatedListeners.isEmpty()) {
+      if (!cacheEntryCreatedListeners.isEmpty() && (clusteringDependentLogic == null || clusteringDependentLogic.shouldCommit(command, ctx, key, false).isLocal())) {
```
pruivo: ```clusteringDependentLogic == null``` is always ```false``` since ```clusteringDependentLogic``` is never ```null```

pruivo: same for others ```notifyCacheEntry*```

tristantarrant:
I'm afraid this needs another rebase...

rvansa:
Rebased and removed those `cdl == null` checks; Any other comments @pruivo?

pruivo:
@rvansa probably. I haven't finished the review :)

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -144,288 +141,188 @@ public final BasicInvocationStage visitClearCommand(InvocationContext ctx, Clear
...
+      List<Address> owners;
+      if (readCH.isReplicated()) {
+         if (readCH.isSegmentLocalToNode(rpcManager.getAddress(), 0)) {
             if (trace) {
-               log.tracef("No valid values found for key '%s' (topologyId=%s).", key, currentTopologyId);
+               log.tracef("Key %s is local, skipping remote get", key);
             }
             return CompletableFutures.completedNull();
          }
+         owners = readCH.getMembers();
```
pruivo: shouldn't we use ```locateOwners``` to make sure we touch the primary owner first?

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -144,288 +141,188 @@ public final BasicInvocationStage visitClearCommand(InvocationContext ctx, Clear
...
-            ResponseMode.SYNCHRONOUS_IGNORE_LEAVERS, DeliverOrder.NONE);
-      RpcOptions options = rpcOptionsBuilder.build();
-      // TODO Use multiple async invocations
-      Map<Address, Response> responses = rpcManager.invokeRemotely(commands, options);
-
-      Map<Object, InternalCacheEntry> entries = new HashMap<>();
-      for (Map.Entry<Address, Response> entry : responses.entrySet()) {
-         updateWithValues(((ClusteredGetAllCommand) commands.get(entry.getKey())).getKeys(),
-               entry.getValue(), entries);
+   protected final BasicInvocationStage handleNonTxWriteCommand(InvocationContext ctx, AbstractDataWriteCommand command)
```
pruivo: IMO, this method could be simplified to avoid so many branches.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -472,111 +369,77 @@ public BasicInvocationStage visitGetAllCommand(InvocationContext ctx, GetAllComm
...
+               if (throwable != null) {
+                  allFuture.completeExceptionally(throwable);
+               }
+               Response response = getSingleSuccessfulResponseOrFail(responseMap, allFuture);
+               if (response == null) return;
+               Object responseValue = ((SuccessfulResponse) response).getResponseValue();
+               if (responseValue instanceof InternalCacheValue[]) {
+                  InternalCacheValue[] values = (InternalCacheValue[]) responseValue;
+                  int counterValue;
+                  synchronized (allFuture) {
```
pruivo: why does it need to be synchronized?

core/src/main/java/org/infinispan/commands/remote/ClusteredGetAllCommand.java
```diff
@@ -93,29 +94,25 @@ public void init(InvocationContextFactory icf, CommandsFactory commandsFactory,
...
-               } else if (entry != null) {
-                  value = entryFactory.createValue(entry);
-               } else {
-                  value = new ImmortalCacheValue(null);
-               }
-               values.add(value);
+            CacheEntry<K, V> entry = map.get(key);
+            InternalCacheValue<V> value;
+            if (entry == null) {
+               value = null;
```
rvansa: Because we're returning `InternalCacheValue[]` in the same order as we got `List<?> keys`.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: Why should we touch primary owner first?

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: There are several remote invocations (this `invokeRemotelyAsync` is called in the for cycle), and this return handler can be invoked concurrently by many of them. As it modifies not only `allFuture.counter` (you could use AtomicInteger there) but also the context by wrapping the entries, we have to synchronize.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
```diff
@@ -203,6 +203,8 @@ public void testScenario_111() throws Exception {
     * ISPN-3315: the remote get is trigger in stable state and the reply received after the rebalance_start command.
     * However, the old owner will receive the request after the state transfer and he no longer has the key.
     */
+   // TODO: does it make sense to fix this test by waiting for the OOTE and then allowing the topology?
+   @Test(enabled = false, description = "Remote get is blocked due to origin topology < old owner topology")
```
rvansa: Originator has topology 4 and sends a remoteGet - old owner has already installed topology 5, so it throws OOTE which is propagated to originator. When it sees OOTE, it waits for next topology (5) but that never comes because the test blocks topology updates on originator. Therefore, the test fails with timeout.

core/src/main/java/org/infinispan/commands/remote/ClusteredGetCommand.java
```diff
@@ -113,12 +114,14 @@ public byte getCommandId() {
    @Override
    public void writeTo(ObjectOutput output) throws IOException {
       output.writeObject(key);
+      output.writeInt(topologyId);
```
pruivo: I disagree. we do the same trick for the cache name.

core/src/main/java/org/infinispan/distribution/DistributionInfo.java
```diff
@@ -0,0 +1,66 @@
+package org.infinispan.distribution;
+
+import java.util.List;
+
+import org.infinispan.distribution.ch.ConsistentHash;
+import org.infinispan.remoting.transport.Address;
+
+public class DistributionInfo {
```
pruivo: at least, it can be simplified. You just need the list of owners (first owner in the list is the primary) and an enum to say if it is the primary/backup/non-owner :)
It would allow to u￼ se a switch statement instead of using a bunch of if...
this is my opinion of course :)

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
pruivo: first, shouldn't we try to fetch the most up-to-date value?
second, isn't the goal of the messy staggered gets to fetch the most up-to-date value from the primary before trying the backups?

pruivo:
@rvansa needs rebase again

core/src/main/java/org/infinispan/distribution/DistributionInfo.java
rvansa: There is only one place with a switch-like check. I wanted to compute as little as possible, though I understand your desire; I can replace `isPrimary()` and `isBackup()` with an enum...

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: The values should be mostly consistent across owners, I don't think it makes sense to optimize for the rare situation when it does not break correctness. The `locateOwners` allocates another array (and does not cache it).
The goal (or purpose) of staggered gets is to not cause more churn on network and processing than necessary, fetching most up-to-date values is not relevant there.

Nevertheless, this code should be encapsuleted somewhere else (something like that `DistributionInfo`), checking if the cache is replicated does not really belong here.

rvansa:
Wrt CI failures: InfinispanNodeFailureTest fails due to ISPN-7140, I can't reproduce the partitionhandling failures and the query test just didn't make it in time (random failure).

core/src/main/java/org/infinispan/container/EntryFactoryImpl.java
```diff
@@ -35,58 +35,54 @@
...
+            }
             ctx.putLookedUpEntry(key, cacheEntry);
          }
       }
 
       if (trace) {
          log.tracef("Wrap %s for read. Entry=%s", toStr(key), cacheEntry);
       }
-      return cacheEntry;
+      return;
```
pruivo: it can be removed.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
pruivo: I was wrong :)

core/src/main/java/org/infinispan/commands/write/RemoveCommand.java
```diff
@@ -72,17 +72,17 @@ public Object perform(InvocationContext ctx) throws Throwable {
...
          nonExistent = true;
-         if (valueMatcher.matches(e, value, null, valueEquivalence)) {
-            if (e != null) {
-               e.setChanged(true);
-               e.setRemoved(true);
-               e.setCreated(false);
-               if (this instanceof EvictCommand) {
-                  e.setEvicted(true);
-               }
+         if (valueMatcher.matches(prevValue, value, null, valueEquivalence)) {
```
pruivo: ```prevValue``` can be replaced by ```null```

pruivo:
Minor "fixes":
```NullCacheEntry.clone()``` should return ```INSTANCE```
```ReadOnlyKeyCommand.perform()``` checking ```entry==null``` twice
```CacheImpl.asyncSkipsThread(EnumSet<Flag> flags, K key)``` can be removed
```GetCacheEntryCommand``` ```remotelyFetchedValue``` field can be removed
14 unused imports

**Comment by rvansa** [comment]
@pruivo Addressed the comments (left unused imports in files I haven't touched), but I've added one more commit there - I've figured out that adding version read while wrapping is not the best idea as you have to rewrite it later on if persistence is used. I had some trouble with that in functional tests. So please review the change as well.

**Comment by danberindei** [comment]
@rvansa There seem to be more test failures with the new commit.

**Comment by rvansa** [comment]
@danberindei I haven't been lucky reproducing the failure in `FunctionalNonTxJoinerBecomingBackupOwnerTest.testBackupOwnerJoiningDuringPutIfAbsent` locally :-/ And `SiteManualSwitchTest.testManualClusterSwitch` does not seem related.

File: core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java [hunk 12]
```diff
@@ -508,47 +584,79 @@ protected void commitSingleEntry(CacheEntry entry, Metadata metadata, FlagAffect
...
                if (previousEntry != null) {
                   previousValue = previousEntry.getValue();
                   previousMetadata = previousEntry.getMetadata();
                }
-               commitManager.commit(entry, metadata, trackFlag, l1Invalidation);
-               if (!isForeignOwned) {
-                  notifyCommitEntry(created, removed, expired, entry, ctx, command, previousValue, previousMetadata);
+               if (isL1Write && previousEntry != null && !previousEntry.isL1Entry()) {
+                  // don't overwrite non-L1 entry with L1 (e.g. when originator == backup
+                  // and therefore we have two contexts on one node)
```
danberindei: If we have regular writes while we're not an owner, that means `CACHE_MODE_LOCAL` was used, and I think it's safe to say the combination of `CACHE_MODE_LOCAL` and L1 leads to undefined results. Also L1 writes don't acquire a lock, you'd need to do this check in `DataContainer.compute()`.

OTOH I think L1 writes are actually committed in `org.infinispan.interceptors.distribution.L1WriteSynchronizer#runL1UpdateIfPossible()`, so the `COMMIT_NON_LOCAL` branch is only used for L1 invalidations.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
```diff
@@ -508,47 +584,79 @@ protected void commitSingleEntry(CacheEntry entry, Metadata metadata, FlagAffect
...
+         boolean transactional = ctx.isInTxScope() && (command == null || !command.hasFlag(Flag.PUT_FOR_EXTERNAL_READ));
+         // When a command is local-mode, it does not get written by replicating origin -> primary -> backup but
+         // when origin == backup it's written right from the original context
+         // ClearCommand is also broadcast to all nodes from originator, and on originator it should remove entries
+         // for which this node is backup owner even though it did not get the removal from primary owner.
+         if (transactional || !ctx.isOriginLocal() || (command != null &&
+               (command.hasFlag(Flag.CACHE_MODE_LOCAL) || command instanceof ClearCommand))) {
+            // During ST, entries whose ownership is lost are invalidated by InvalidateCommand
+            // without any flags, and at that point we're no longer owners - the only information is that
+            // the origin is not local and the entry is removed.
```
danberindei: `InvalidateCommand` is never replicated, so it doesn't really need `CACHE_MODE_LOCAL`, but we can add it if it simplifies things.

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
```diff
@@ -143,6 +146,22 @@ private BasicInvocationStage handleDataReadCommand(InvocationContext ctx, DataCo
                // get response.
                throw log.degradedModeKeyUnavailable(dataCommand.getKey());
             } else {
+               // If all owners left and we still haven't received the availability update yet,
+               // we get OutdatedTopologyException from BaseDistributionInterceptor.retrieveFromProperSource
+               if (t instanceof OutdatedTopologyException && performPartitionCheck(rCtx, dataCommand)) {
+                  // Unlike in PartitionHandlingManager.checkRead(), here we ignore the availability status
+                  // and we only fail the operation if _all_ owners have left the cluster.
+                  // TODO Move this to the availability strategy when implementing ISPN-4624
```
danberindei: ISPN-4624 is about how a cache changes its availability mode, not about how individual operations react to a particular availability mode. I'm not sure we want to make that customizable.

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
```diff
@@ -380,13 +386,19 @@ private BasicInvocationStage handleNonTxWriteCommand(InvocationContext ctx, Writ
          }
          if (!(ce instanceof OutdatedTopologyException) && !(ce instanceof SuspectException))
             throw t;
+         WriteCommand writeCommand = (WriteCommand) rCommand;
+         if (writeCommand.hasFlag(Flag.ZERO_LOCK_ACQUISITION_TIMEOUT)) {
+            // the command is expected to not wait
```
danberindei: Actually, the command is expected not to wait for another operation modifying the same lock, the flag javadoc doesn't say anything about other kinds of waiting.

core/src/main/java/org/infinispan/transaction/xa/CacheTransaction.java
```diff
@@ -130,7 +130,8 @@ default void putLookedUpRemoteVersion(Object key, EntryVersion version) {}
     * <p/>
     * Note: used in Repeatable Read + Write Skew + Clustering + Versioning.
     */
-   void replaceVersionRead(Object key, EntryVersion version);
+   @Deprecated
+   default void replaceVersionRead(Object key, EntryVersion version) { addVersionRead(key, version);}
```
danberindei: I wouldn't put the whole method on one line, not even if it's deprecated :)
Please also add a `@deprecated Since 9.0` in the javadoc, it'll make it easier to decide when to remove it later.

core/src/test/java/org/infinispan/api/flags/FlagsEnabledTest.java
```diff
@@ -82,7 +82,7 @@ public void testWithFlagsSemantics() {
 
       MagicKey nonLocalKey = new MagicKey("nonLocal", cache2);
       cache1.put(nonLocalKey, "value");
-      // Write skew check needs the previous version on the originator AND on the primary owner
+      // In transactional mode the value is loaded locally and later correct version is verified by WSC
```
danberindei: TBH I found the old comment clearer :)

core/src/test/java/org/infinispan/api/flags/FlagsEnabledTest.java
```diff
@@ -93,6 +93,8 @@ public void testWithFlagsSemantics() {
             cache1LocalOnly.withFlags(SKIP_CACHE_LOAD);
       MagicKey localKey2 = new MagicKey("local2", cache1);
       cache1SkipRemoteAndStores.put(localKey2, "value");
+      // local-mode operation does not get into PrepareCommand as modification and WSC is not executed
+      // (though, it's committed as it's in context and it's stored as it is retured by localTransaction.getAllModifications()
```
danberindei: How about "CACHE_MODE_LOCAL operation is not replicated with the PrepareCommand, but the entry is updated on the origin"?

core/src/test/java/org/infinispan/api/mvcc/LockTestBase.java
```diff
@@ -225,10 +225,13 @@ public void testWriteDoesntBlockRead() throws Exception {
...
-      if (repeatableRead)
-         assert null == cache.get("k2") : "Should have repeatable read";
+      String value = cache.get("k2");
+      if (repeatableRead) {
+         assert null == value : "Should have repeatable read";
+      }
       else
-         assertEquals("Read committed should see committed changes", "v2", cache.get("k2"));
+         // no guarantees with read committed
+         assertTrue(null == value || "v2".equals(value));
```
danberindei: Actually we do guarantee that the tx sees values committed by other txs if the cache is local and without a store.

core/src/test/java/org/infinispan/distribution/groups/TransactionalGetGroupKeysTest.java
```diff
@@ -140,8 +140,7 @@ public void testGetGroupsWithConcurrentReplace() throws Exception {
 
       testCache.primaryOwner.put(key(1), value(-1));
 
-      if ((factory == TestCacheFactory.PRIMARY_OWNER || factory == TestCacheFactory.BACKUP_OWNER) &&
-            isolationLevel == IsolationLevel.READ_COMMITTED) {
+      if (isolationLevel == IsolationLevel.READ_COMMITTED) {
```
danberindei: Does this mean we re-read the value from remote nodes on every read in a read-committed tx?

core/src/test/java/org/infinispan/functional/FunctionalCachestoreReadOnlyTest.java
```diff
@@ -21,8 +21,18 @@
...
 @Test(groups = "functional", testName = "functional.FunctionalCachestoreReadOnlyTest")
 public class FunctionalCachestoreReadOnlyTest extends AbstractFunctionalReadOnlyOpTest {
-   public FunctionalCachestoreReadOnlyTest() {
-      isPersistenceEnabled = true;
+
+   @Override
+   public Object[] factory() {
+      return new Object[] {
+         new FunctionalCachestoreReadOnlyTest().persistence(true).passivation(false),
+         new FunctionalCachestoreReadOnlyTest().persistence(true).passivation(true),
```
danberindei: `persistence` could still be set to `true` in the constructor...

core/src/test/java/org/infinispan/notifications/cachelistener/cluster/AbstractClusterListenerNonTxTest.java
```diff
@@ -62,22 +62,40 @@ public String call() throws Exception {
...
 
-      // We should have received 2 events
-      assertEquals(clusterListener.events.size(), 2);
-      CacheEntryEvent<Object, String> event = clusterListener.events.get(0);
-      assertEquals(event.getType(), Event.Type.CACHE_ENTRY_CREATED);
-      CacheEntryCreatedEvent<Object, String> createEvent = (CacheEntryCreatedEvent<Object, String>)event;
-      assertFalse(createEvent.isCommandRetried());
-      assertEquals(createEvent.getKey(), key);
-      assertEquals(createEvent.getValue(), FIRST_VALUE);
+      // We should have received 2 or 3 events - see below
```
danberindei: Just move that comment here :)

core/src/test/java/org/infinispan/persistence/ClusteredTxConditionalCommandPassivationTest.java
```diff
@@ -24,19 +25,24 @@ public ClusteredTxConditionalCommandPassivationTest() {
...
-               assertEquals("non owner load", 0, cacheHelper.loads(Ownership.NON_OWNER));
-            } else {
+            if (skipLoad) {
                assertLoad(cacheHelper, 0, 0, 0);
+            } else {
+               // The entry is loaded into DC upon the initial value retrieval (ClusteredGetCommand).
+               // It gets loaded all the time on primary, but if the response to the retrieval does not
+               // come soon enough, staggered logic sends second retrieval to backup owner, so it's possible
+               // that both owners load once.
+               assertEquals("primary owner load", 1, cacheHelper.loads(Ownership.PRIMARY));
```
danberindei: In theory we could do this check before the primary cache loader managed to load the entry and increment its load count.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
```diff
@@ -200,118 +200,6 @@ public void testScenario_111() throws Exception {
    }
 
    /**
-    * ISPN-3315: the remote get is trigger in stable state and the reply received after the rebalance_start command.
-    * However, the old owner will receive the request after the state transfer and he no longer has the key.
```
danberindei: I think you need to update the table at the top of the class as well. Actually I'd rather keep retrying without waiting for a new topology, just doing it from the `StateTransferInterceptor` instead of the `XDistributionInterceptor`.

tree/src/test/java/org/infinispan/api/tree/NodeMoveAPIPessimisticTest.java
```diff
@@ -22,4 +22,11 @@ protected ConfigurationBuilder createConfigurationBuilder() {
             .transaction().lockingMode(LockingMode.PESSIMISTIC);
       return cb;
    }
+
+   @Test(enabled = false, description = "The test does not work on pessimistic RR caches, " +
+         "because moveCtoB reads in the root node into context. Then, the move operation does not check " +
+         "if the entry was not changed after lock (no pessimistic WSC) and the move can be executed.")
```
danberindei: Did this work before?

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -1178,6 +1178,10 @@ final void putAll(Map<? extends K, ? extends V> map, Metadata metadata, long exp
    private void putAllInternal(Map<? extends K, ? extends V> map, Metadata metadata, long explicitFlags, InvocationContext ctx) {
       InfinispanCollections.assertNotNullEntries(map, "map");
       Metadata merged = applyDefaultMetadata(metadata);
+      // Vanilla PutMapCommand returns previous values; add IGNORE_RETURN_VALUES as the API will drop the return value
+      // interceptors are free to clear this flag if appropriate (since interceptors are the only consumers of the
+      // return value)
```
rvansa: And that's why we set the flag by default, and interceptor request them by clearing the flag. I could introduce another flag `FORCE_RETURN_VALUE` but this would pollute the code without any gain.

core/src/main/java/org/infinispan/commands/VisitableCommand.java
```diff
@@ -48,15 +48,23 @@
    boolean ignoreCommandOnStatus(ComponentStatus status);
 
    /**
-    * @return {@code true} if the command needs to read the previous values of the keys it acts on.
+    * @return Nodes on which the command needs to read the previous values of the keys it acts on,
+    * or <code>null</code> if an invocation does not make sense.
```
rvansa: I don't see anything outdated - `loadType()` returns `PRIMARY`, `OWNERS`, or `DONT_LOAD` (that means empty set of nodes).

core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyCommand.java
```diff
@@ -18,6 +18,8 @@
 import org.infinispan.functional.impl.EntryViews;
 import org.infinispan.functional.impl.Params;
 
+// TODO: the command does not carry previous values to backup, so it can cause
+// the values on primary and backup owners to diverge in case of topology change
```
rvansa: No, I haven't touched that; `ReadWriteKeyValueCommand` is the only command that sends the previous value - it was simply not implemented in the others.

core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyValueCommand.java
```diff
@@ -103,7 +103,9 @@ public Object perform(InvocationContext ctx) throws Throwable {
       // If the value has been update while on the retry, use the newer value.
       // Also take into account that the value might have been removed.
       // TODO: Configure equivalence function
-      if (valueUnchanged(e, prevValue, value) || valueRemoved(e, prevValue)) {
+      // TODO: this won't work properly until we store if the command was executed or not...
+      Object oldPrevValue = e.getValue(hasFlag(Flag.COMMAND_RETRY));
```
rvansa: With `ReadCommittedEntry` (as so far the func commands are non-tx only) this will always return the current value. As the entry is removed from context in `EntryWrappingInterceptor.removeFromContextOnRetry` before wrapping the value again on a new command execution, it returns always the value currently in DC, and that's the best you can get.
I admit that the `getValue(retry)` may be unnecessary (I could try to remove that logic and run testsuite again), this code went through several iterations and given that it's broken by design from the beginning...

core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyValueCommand.java
```diff
@@ -103,7 +103,9 @@ public Object perform(InvocationContext ctx) throws Throwable {
...
+      if (valueUnchanged(oldPrevValue, prevValue, value) || valueRemoved(oldPrevValue, prevValue)) {
```
rvansa: Good catch, though I haven't really changed that part :)

core/src/main/java/org/infinispan/commands/read/KeySetCommand.java
```diff
@@ -48,8 +48,8 @@ public Object acceptVisitor(InvocationContext ctx, Visitor visitor) throws Throw
    }
 
    @Override
-   public boolean readsExistingValues() {
-      return false;
+   public LoadType loadType() {
+      return null;
```
rvansa: I would say `UNDEF` - yep, I could add that. Since you didn't like the `OTHER` option in the past, I've rather used null :)

core/src/main/java/org/infinispan/commands/remote/ClusteredGetCommand.java
```diff
@@ -78,24 +79,23 @@ public void initialize(InvocationContextFactory icf, CommandsFactory commandsFac
    public CompletableFuture<Object> invokeAsync() throws Throwable {
       // make sure the get command doesn't perform a remote call
       // as our caller is already calling the ClusteredGetCommand on all the relevant nodes
-      long flagBitSet = EnumUtil.bitSetOf(Flag.SKIP_REMOTE_LOOKUP, Flag.CACHE_MODE_LOCAL);
+      // CACHE_MODE_LOCAL is not used as it can be used when we want to ignore the ownership with respect to reads
```
rvansa: The flag forces the operation to load non-owned entries, even if it is null. When a ClusteredGetCommand comes to a node that is no longer owner of given entry, but the topology id matches, `UnsuccessfulResponse.INSTANCE` is returned instead. I can't recall when this happens, but I think I've seen that in some stress test.

core/src/main/java/org/infinispan/commands/remote/ClusteredGetCommand.java
```diff
@@ -78,24 +79,23 @@ public void initialize(InvocationContextFactory icf, CommandsFactory commandsFac
...
-            MVCCEntry mvccEntry = (MVCCEntry) cacheEntry;
+            MVCCEntry mvccEntry = (MVCCEntry) rv;
             return entryFactory.createValue(mvccEntry);
-         } else {
-            InternalCacheEntry internalCacheEntry = (InternalCacheEntry) cacheEntry;
+         } else if (rv instanceof InternalCacheEntry) {
+            InternalCacheEntry internalCacheEntry = (InternalCacheEntry) rv;
             return internalCacheEntry.toInternalCacheValue();
+         } else {
+            return rv;
```
rvansa: `UnsuccessfulResponse.INSTANCE`

core/src/main/java/org/infinispan/commands/write/PutMapCommand.java
```diff
@@ -251,8 +261,8 @@ public boolean ignoreCommandOnStatus(ComponentStatus status) {
    }
 
    @Override
-   public boolean readsExistingValues() {
-      return false;
+   public LoadType loadType() {
+      return hasFlag(Flag.IGNORE_RETURN_VALUES) ? LoadType.DONT_LOAD : LoadType.PRIMARY;
```
rvansa: Interceptors work on commands and context. Do you want a special flag in context that will force loading return values, and all the interceptors will have to check that, too? I don't like that idea at all.
Seems that Index.ALL could really require that on all owners, which is pretty annoying and not limited to PutMapCommand. And it should not work now either if persistence is involved.

core/src/main/java/org/infinispan/container/EntryFactoryImpl.java
```diff
@@ -226,21 +217,22 @@ private CacheEntry getFromContext(InvocationContext ctx, Object key) {
       return cacheEntry;
    }
 
-   private InternalCacheEntry getFromContainer(Object key, boolean ignoreOwnership, boolean writeOperation) {
-      final boolean isLocal = distributionManager == null || distributionManager.getLocality(key).isLocal();
-      if (isLocal || ignoreOwnership) {
+   private CacheEntry getFromContainer(Object key, boolean shouldLoad, boolean writeOperation) {
```
rvansa: `shouldLoad == isOwner || hasFlag(SKIP_OWNERSHIP_CHECK)`

core/src/main/java/org/infinispan/container/entries/ClusteredRepeatableReadEntry.java
```diff
@@ -71,6 +71,25 @@ public boolean performWriteSkewCheck(DataContainer container, PersistenceManager
       return InequalVersionComparisonResult.EQUAL == result;
    }
 
+   private EntryVersion getCurrentEntryVersion(DataContainer container, PersistenceManager persistenceManager, TxInvocationContext ctx, VersionGenerator versionGenerator, TimeService timeService) {
+      EntryVersion prevVersion;// on origin, the version seen is acquired without the lock, so we have to retrieve it again
+      // TODO: persistence should be more orthogonal to any entry type - this should be handled in interceptor
```
rvansa: The second command says that access to persistence belongs to interceptors, as these modify the state of context, commands and other components. Methods withing `CacheEntry` should access only the entry itself.

core/src/main/java/org/infinispan/container/entries/DeltaAwareCacheEntry.java
```diff
@@ -146,16 +148,26 @@ public final K getKey() {
...
+      return uncommittedChanges;
    }
 
    @Override
    public final DeltaAware setValue(DeltaAware value) {
-      DeltaAware oldValue = this.value;
-      this.value = (DeltaAware) value;
+      DeltaAware oldValue = getValue();
+      this.value = value;
+      // TODO: this won't work without the copy for normal deltas
```
rvansa: I think that the problem is that when the value is not AtomicHashMap, we don't copy it and modify the original DeltaAware value.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -144,288 +141,181 @@ public final BasicInvocationStage visitClearCommand(InvocationContext ctx, Clear
...
-   private CompletableFuture<InternalCacheEntry> doRetrieveFromProperSource(Object key, InternalCacheEntry value,
-         int lastTopologyId, InvocationContext ctx, FlagAffectedCommand command, boolean isWrite) {
-      if (value != null)
-         return CompletableFuture.completedFuture(value);
-
-      final CacheTopology cacheTopology = stateTransferManager.getCacheTopology();
-      final int currentTopologyId = cacheTopology.getTopologyId();
+   protected final CompletableFuture<Void> remoteGet(InvocationContext ctx, AbstractDataCommand command,
+                                                     boolean isWrite) throws Exception {
+      final CacheTopology cacheTopology = checkTopologyId(command);
```
rvansa: We should follow simple principles, and use them everywhere - reasoning about operations is already too complex. Slowing down operations during rebalance is acceptable.

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -106,32 +93,11 @@ public BasicInvocationStage visitPutKeyValueCommand(InvocationContext ctx, PutKe
...
-   private InvocationStage visitGetCommand(InvocationContext ctx, AbstractDataCommand command)
-         throws Throwable {
-      Object key = command.getKey();
-      CacheEntry entry = ctx.lookupEntry(key);
-      // If the cache entry has the value lock flag set, skip the remote get.
-      if (ctx.isOriginLocal() && valueIsMissing(entry)) {
-         if (readNeedsRemoteValue(ctx, command)) {
-            remoteGet(ctx, key, false, command);
-         }
+      // remote commands are replayed in EWI
```
rvansa: Sorry, missing context...?

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
```diff
@@ -65,17 +67,54 @@
...
+ * wrapped is a prerequisite for the command to run and therefore commit the entry, but it's not up to this interceptor
+ * to make sure the entry is always wrapped.
+ * <p>
+ * If the entry should be read locally but it's not found in DC, the entry will be wrapped (either as NullCacheEntry
+ * for reads or other appropriate type for writes). Such entry must then return <code>false</code> on
+ * {@link CacheEntry#skipLookup()} as it's value is unsure (subsequent interceptors can retrieve the new value from the
+ * cache store or remote node and call {@link EntryFactory#wrapExternalEntry} to update the context and make the entry valid).
+ * <p>
+ * With repeatable reads, the value that is context must not be overwritten by value out of the transaction (only commands
+ * in this transaction can change the context entry. That's why we need to call {@link CacheEntry#setSkipLookup} after
```
rvansa: Read-committed permits seeing new values, so we don't bother. The call is noop with non-RR entries anyway.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
```diff
@@ -651,40 +653,57 @@ private BasicInvocationStage setSkipRemoteGetsAndInvokeNextForDataCommand(Invoca
...
    // This visitor replays the entry wrapping during remote prepare.
    // Remote writes never request the previous value from a different node,
    // so it should be safe to keep this synchronous.
+   //
+   // We ignore the ownership and wrap the entry. We have already checked writeCH for ownership
+   // - we could also check readCH, and always wrap NullCacheEntry if we cannot read but since
+   // the loaded entry is only used for listeners, it wouldn't do much good (we cannot reliably
+   // fire listeners without remote lookup).
+   // TODO: loaded entry is also used for WSC on primary and for delta/functional writes. If the node
+   // has not got the data yet, it has to request them explicitly!
```
rvansa: WSC will work as long as primary = read owner. If backup != read owner, atomic hashmaps don't work correctly, because the needed value is not fetched from remote node. In order to fix that, we would have to execute the Visitor non-synchronously and let it do a remoteGet in TxDistributionInterceptor.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
```diff
@@ -410,14 +443,60 @@ public void init(StateTransferLock stateTransferLock) {
...
+               }
+            }
          } finally {
             stateTransferLock.releaseSharedTopologyLock();
          }
       }
 
+      public Commit shouldCommit(FlagAffectedCommand command, InvocationContext ctx, Object key, boolean removed) {
+         // The logic here should be in sync with the logic in DistributionLogic but without the L1 stuff.
+         // Look for comments there
```
rvansa: Oops, outdated, difference is only in `commitSingleEntry`. I should remove one of the methods.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
rvansa: `CACHE_MODE_LOCAL` javadoc explicitly says that write goes to L1. But the issue here is different, and I think it was related to the code where I've assumed that puts retrieve the value ahead. There an originator == backup would fetch the remote value into context, execute write on primary and get the value updated, and then it attempted to commit the value it has in context as L1 entry. But that shouldn't be necessary now (I hope) and is prone to the absence of locking anyway.

Pch, I see that `runL1UpdateIfPossible` writes the L1 entry without DC.compute. I wonder if anyone has ran resilience tests with L1 recently...

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
rvansa: Ok, it could work with `CACHE_MODE_LOCAL | SKIP_CACHE_LISTENERS`

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
rvansa: Then we should update the javadoc - the purpose is really "Do it now or forget it"

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
danberindei: I agree a new flag wouldn't make sense, I was actually thinking of a boolean in the command itself, like `isForwarded`. My reasoning was that it would be easier to add something that's specifically internal-only than fixing all the places in the code that ignore `IGNORE_RETURN_VALUES` for `PutMapCommand`. E.g. should `BaseTypeConverterInterceptor` unwrap the results?

core/src/main/java/org/infinispan/commands/VisitableCommand.java
danberindei: Oops, I read "nodes" and I thought the javadoc came from a method returning a collection of addresses or something like that :)
When the invocation doesn't make sense, shouldn't the method throw an `IllegalStateException`?

core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyCommand.java
danberindei: Ok

core/src/main/java/org/infinispan/commands/read/KeySetCommand.java
danberindei: I had forgotten that you already have `DONT_LOAD`, and I thought that's what `null` was supposed to mean.

core/src/main/java/org/infinispan/commands/remote/ClusteredGetCommand.java
danberindei: I'm not sure how the topology id could match and yet the node would not be the owner... could it happen if you receive a `ClusteredGetCommand` with topology `x` while you're still at topology `x-1`?

rvansa: No, the command is delayed until you receive given topology... it's probably a leftover, I'll cut that off, then.

core/src/test/java/org/infinispan/api/mvcc/LockTestBase.java
rvansa: Where do we commit to that?

core/src/main/java/org/infinispan/commands/remote/ClusteredGetCommand.java
danberindei: I actually thought a lot about returning actual `Response` instances from `VisitableCommand.perform()`, but I didn't like the idea that every interceptor would have to know about it in order to ignore it...

core/src/test/java/org/infinispan/persistence/ClusteredTxConditionalCommandPassivationTest.java
rvansa: Duh? The assert is called when the command finishes...

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
rvansa: Is there a reason for topology being installed on one node later than on the others?

core/src/main/java/org/infinispan/commands/write/PutMapCommand.java
danberindei: Well, `QueryInterceptor` doesn't really care about the return value, it needs the previous value -- it's just that with the "classic" single-key cache operations, the return value was enough to determine the previous value. If a functional operation replaces "a" with "b" and returns "c", we must remove "a" from the index, add "b" instead, and ignore "c".

We sometimes need to force a remote invocation to return the previous value to the originator for indexing, but I'm hoping we can separate that from the command's return value, so we can implement it the same way for each command instead of having ever more elaborate conventions for the return values.

core/src/main/java/org/infinispan/container/EntryFactoryImpl.java
danberindei: Then maybe you need to change the callers as well? All of them use a variable called `isOwner`.

core/src/main/java/org/infinispan/commands/remote/ClusteredGetCommand.java
rvansa: Since commands can return aribitrary value and sometimes you need to return failure (though different than an exception - e.g. including value), the interceptors have to anticipate this. Or, we could really use just exceptions, though I am not in favour of that - one of the changes I haven't squeezed into the PR is replacing `WriteCommand.updateStatusFromRemoteResponse` with `UnsuccessfulResponse` carrying value.

core/src/main/java/org/infinispan/container/entries/ClusteredRepeatableReadEntry.java
danberindei: I agree `CacheEntry` implementations in general shouldn't know about the persistence, but after a while I've come to see `MVCCEntry` implementations more as part of `EntryWrappingInterceptor` :)

core/src/main/java/org/infinispan/container/entries/DeltaAwareCacheEntry.java
```diff
@@ -168,9 +180,7 @@ public boolean isNull() {
    public final void commit(final DataContainer<K, DeltaAware> container, final Metadata metadata) {
       //If possible, we now ensure copy-on-write semantics. This way, it can ensure the correct transaction isolation.
       //note: we want to merge/copy to/from the data container value.
-      container.compute(key, new DataContainer.ComputeAction<K, DeltaAware>() {
-         @Override
-         public InternalCacheEntry<K, DeltaAware> compute(K key, InternalCacheEntry<K, DeltaAware> oldEntry, InternalEntryFactory factory) {
+      PersistenceUtil.loadAndComputeInDataContainer(container, persistenceManager, key, ctx, timeService, (key, oldEntry, factory) -> {
```
danberindei: I was thinking that in non-tx mode too, the wrapping happens after the locking, so you can't have other writes modifying the entry. Of course, I had forgotten this is for `applyDelta()`, which works with partial locks.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: IMO the fact that reads are never blocked by state transfer is a very nice feature of Infinispan and I wouldn't give it up just because the code looks a bit simpler without it...

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
danberindei: You're only calling `wrapMissingWrittenKeysAsNull` if `ctx.isOriginLocal()`, so I understood the comment to mean you don't want to wrap missing keys when replaying the write command during remote prepare. I'm not sure why, though.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
danberindei: Ok, I just took the comment to mean that with read-committed it's ok to re-fetch the value from a remote node.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
```diff
@@ -281,120 +320,74 @@ public final BasicInvocationStage visitClearCommand(InvocationContext ctx, Clear
...
-
-         return true;
-      } else if (command.hasFlag(Flag.CACHE_MODE_LOCAL)) {
+   private void removeFromContextOnRetry(InvocationContext ctx, AbstractDataWriteCommand command) {
+      // When originator is a backup and it becomes primary (and we retry the command), the context already
+      // contains the value before the command started to be executed. However, another modification could happen
+      // after this node became an owner, so we have to force a reload.
+      // With repeatable reads, we cannot just remove the entry from context; instead of we will rely
+      // on the write skew check to do the reload & comparison in the end.
+      // TODO: what about pessimistic locking?
```
danberindei: > Eh, replace value in DC from the context?

I was thinking of replacing the value in the `MVCCEntry` with the previous value from the same `MVCCEntry`.

> I am not breaking RR for pessimistic tx, but a concurrent modification may be missed, therefore the conditional operation won't behave correctly. Though, if we're already acquired the lock, there shouldn't be a concurrent op. The TODO was more about reasoning about that :)
> And yes, WSC will be thrown only in case of concurrent modification.

Then I guess the comment is no longer needed? It makes it look like the implementation is incomplete.

> Are there any tests that simulate retries on tx cache?

Check out `TxReplay*Test`. Of course, they don't test retries of tx write commands before the commit.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
danberindei: I don't think we have to do both at once, we can do the remoteGet in TxDistributionInterceptor for correctness now, and fix performance later.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
danberindei: > CACHE_MODE_LOCAL javadoc explicitly says that write goes to L1. 

Hmm, I never realized that `CACHE_MODE_LOCAL` writes are supposed to go into L1... I think maybe the idea was that `putForExternalRead()` should never go remotely, but it's weird because the node isn't going to be registered as an L1 requestor for that key, so `L1ManagerImpl.flushCache()` isn't going to send any invalidation.

> Pch, I see that runL1UpdateIfPossible writes the L1 entry without DC.compute. I wonder if anyone has ran resilience tests with L1 recently...

Synchronizers have their own per-key locking, I think `DC.compute` isn't needed as long as you don't have `CACHE_MODE_LOCAL` writes to L1.

> But the issue here is different, and I think it was related to the code where I've assumed that puts retrieve the value ahead. There an originator == backup would fetch the remote value into context, execute write on primary and get the value updated, and then it attempted to commit the value it has in context as L1 entry. But that shouldn't be necessary now (I hope) and is prone to the absence of locking anyway.

Yeah, I think now the only way to write to L1 with proper locking is through an `L1Synchronizer`.

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
danberindei: You mean that's what `putForExternalRead()` needs, or do you have other use cases in mind?

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
rvansa: I meant that EWI.EWV already makes sure that all entries are wrapped, therefore there's no need to check that again here.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
danberindei: GC would be the most common one. But when the cluster is under a lot of load, some nodes may also discard the `CacheTopologyControlCommand` message and have to wait for a retransmission.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: Okay, so we have a two situations when a node blocks:
* sender has lower topology than the receiver: Currently receiver throws `OTE` - we could replace `OTE` with a `OutdatedTopologyResponse` that will carry newer routing information.
* receiver has lower topology than the sender: Receiver gives the sender the entry if it owns it, if it does not it blocks on state transfer lock (in STI, probably) until it gets the command topology. If it is not a read owner in the command topology, it returns `OTR` with owners according to the current topology.

If sender receives `OTR`, it routes the command according to this topology (as it might be higher than its current topology). Not sure if the `WAIT_FOR_FIRST_RESPONSE` should return `OTR` immediately or wait until succesful response/all nodes return `OTR`.

WDYT?

core/src/test/java/org/infinispan/distribution/groups/TransactionalGetGroupKeysTest.java
rvansa: This is specific to GetKeysInGroupCommand: since we don't know the finite set of keys in given group, we have to always send the command to remote nodes to get the whole group. The response will come with the modified value and since we're RC, the value in context gets overwritten.
I had to run the test to realize that, so I'll add a comment there.

rvansa: Note that if we did regular get, it would be satisfied from local context and there wouldn't be any remote operation.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: Replying with an `OutdatedTopologyResponse` instead of the exception sounds like a good thing, if only to minimize the overhead of all those stack traces. But I still don't see a good reason for the reads to block at all... 

Say the sender has topology TS and the receiver has topology TR. 

If TS < TR and the receiver is a read owner in TR, the receiver knows it has the proper value, so it can return it. If it's only a write owner or not an owner at all, it can return an OTR. But if the sender receives an OTR and it still has topology TS, it knows who to ask for the value immediately: the write owners in TS. The coordinator can't finish a rebalance without installing the rebalance-in-progress topology on all the members, so the sender must install a newer rebalance-in-progress topology before the receiver can see a topology where the write owners in TS to all become non-owners.

If TS > TR, the receiver can only be a read owner for more keys in topology TS if it was a write owner of those keys in topology TR and it had already sent the REBALANCE_CONFIRM(TR) command to the coordinator before the sender installed topology TR. But that means the receiver already has the requested key, so blocking in STI to receive topology TS delays the read but doesn't make it any safer.

For non-staggered gets, when the response filter is called with an OTR it could compute what the new recipients would be (either from the OTR or from the sender's current topology) and decide whether to stop the current request based on how much of an overlap there is. But for staggered gets I think the penalty for ending the request early and starting a new one is minimal, so I'd always cancel the request when receiving the first OTR.

core/src/main/java/org/infinispan/commands/remote/ClusteredGetCommand.java
danberindei: I don't think they had to anticipate this before, e.q. `QueryInterceptor` doesn't check `command.isSuccessful()` for `PutKeyValueCommand` and just assumes the result is the previous value. We can definitely change them, but we'd have to add some tests as well.

core/src/test/java/org/infinispan/distribution/groups/TransactionalGetGroupKeysTest.java
rvansa: I got lost in the conditionals :) The key *is* in context after we call `getGroup()` the first time. On the remote node, it always iterates whole DC and finds all keys that are in given group, and sends this result back. In RC the context is overwritten, in RR it is not.
It's possible that in the past, the context was not overwritten neither in RC and therefore the test had to contain that condition - I haven't checked that.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: > See how complex is the reasoning who must finish what before something else? And it makes that algorithm quite fragile.

True, this depends very closely on the way a rebalance is started/finished. But I'd argue you need to learn how rebalancing works anyway in order to test all the possible cases. Testing all the possible ways in which in which write commands can be retried because of a topology change is pretty painful, too, I'm not sure we are testing that with all the possible interceptors even now. And we still need new tests for write commands that are retried because of a remote get, especially in a transaction.

> sending readCH in OTR is not that expensive.

It can be expensive if the cluster is big, and that's exactly when we're most likely to get a delay in the topology installation.

> If TS > TR, we deal with the case that the owner lost some keys in TR that were in TS. 

You mean the owner gained some keys in TS that were not in TR, right?

> Rather than block and wait if it eventually gets the keys back, it could just return OTR without any CH, because it knows that its own CH is already outdated. If you say that the sender will always get some successful case, we can leave it up to that because it's easy to detect and throw e.g. RpcException if the assumption changes.

I'm saying the sender will always get some successful case *if* we allow the receiver to return an entry for which it is a write-owner, knowing that the new topology in which it is a read-owner was already sent out by the coordinator. If the receiver returns an OTR whenever it's not a read-owner, the sender is no longer guaranteed to get a valid response and it can't wait for a newer topology either (because it already has the newer topology).

core/src/test/java/org/infinispan/distribution/groups/TransactionalGetGroupKeysTest.java
danberindei: Ok, now I think I really got it :)

I got confused by the `GetKeysInGroupCommand` name, I thought it returns just the keys and the originator then has to invoke a `GetAllCommand` to get all the values (that are not already in the context). But since we're including those values in the response anyway, it makes sense to update the context entries with them.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: > But I'd argue you need to learn how rebalancing works anyway in order to test all the possible cases.
You need to know how a rebalance happens, but you don't want to make assumptions which write owners will become read owners. We have to isolate concepts as much as possible.

> You mean the owner gained some keys in TS that were not in TR, right?
Yes, sorry. TR has less keys than TS but the time flows the other way.

> receiver to return an entry for which it is a write-owner, knowing that the new topology in which it is a read-owner was already sent out by the coordinator
Okay, I see that you can deduce that it already got all entries and confirmed it with current state transfer, but this is a complex and fragile assumption based on the state transfer implementation, not on actual consistent hashes. That's why I wanted to block there for the moment until the new topology comes and respond only afterward, as it does not know for sure (= from CH) that it can do that. So, if we send OTR (or rather UnsuccessfulResponse as the sender topology is not outdated), the sender has to retry.

danberindei: > You need to know how a rebalance happens, but you don't want to make assumptions which write owners will become read owners. We have to isolate concepts as much as possible.

In general I'm trying to choose the simpler algorithm too, but in this case I'm not at all convinced it's worth losing the "reads never block" bullet point. I think you should put this out on the mailing list, so more people can weigh in.

> So, if we send OTR (or rather UnsuccessfulResponse as the sender topology is not outdated), the sender has to retry.

If the choice is between waiting on the receiver and retrying from the sender with the same topology id, I'd rather wait on the receiver.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
danberindei: I've created https://issues.jboss.org/browse/ISPN-7165 to remove this feature from `CACHE__MODE_LOCAL`.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -472,111 +362,77 @@ public BasicInvocationStage visitGetAllCommand(InvocationContext ctx, GetAllComm
...
-         }
-         if (topologyChanged) {
-            throw new OutdatedTopologyException("Cache topology changed while the command was executing: expected " +
-                        commandTopologyId + ", got " + currentTopologyId);
-         }
+         CacheTopology cacheTopology = checkTopologyId(command);
+         ConsistentHash ch = cacheTopology.getReadConsistentHash();
 
          // At this point, we know that an entry located on this node that exists in the data container/store
          // must also exist in the context.
```
rvansa: Why? The comment says: `key in readCH(localAddress) && key in DC => key in ctx`

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
rvansa: I think that in Hibernate ORM 2LC on dist cache PFERs run remotely, too. Not sure if these are sync/async, but PFERs should not imply CACHE_MODE_LOCAL.
What's the purpose of zero lock timeout rather than fail-fast?

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: I thought at this point the key should be in the context even if it's not in DC, as long as it's owned by the local node in the read CH?

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
danberindei: Yes, it means fail-fast, but only for locking :)

When it comes to RPCs, you already have 2 options to avoid blocking: `CACHE_MODE_LOCAL` or `FORCE_ASYNCHRONOUS`. (I see `putForExternalRead` uses the latter.) I don't think we need another option to block for RPCs, but only if there's no topology change...

rvansa: State transfer lock is a lock as well :)

Anyway, I don't agree, but I can remove that change since it's not a necessary part of the PR. You can have the dangling threads/futures waiting for the topology and then failing if you like them so much.

danberindei: You could also say that JGroups also uses locking under the hood (e.g. in UFC/MFC), so you shouldn't do any RPCs with `ZERO_LOCK_ACQUISITION_TIMEOUT`. But obviously we want to to RPCs for `putForExternalRead`, so we have to limit the semantics of ZLAT at some point.

rvansa:
@danberindei Rebased & added couple of commits (starting with ISPN-7029 Fixed all scenarios in RemoteGetDuringStateTransferTest), these should be probably all squashed if you find this ready to be integrated.

tree/src/test/java/org/infinispan/api/tree/NodeMoveAPIPessimisticTest.java
rvansa: The test was passing by coincidence IIRC, but generally, the `move` did not work in all cases.

core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyCommand.java
galderz: @rvansa That's right. It was implemented partially because the solution is not optimal.

core/src/main/java/org/infinispan/container/EntryFactory.java
```diff
@@ -1,68 +1,110 @@
...
+ * <h3>Expected wrapping patterns</h3>
+ *
+ * {@link EntryWrappingInterceptor} checks {@link CacheTopology#getReadConsistentHash()} and if this node is an owner
+ * of the key, it loads the entry from {@link DataContainer}. On the way back through interceptor stack, the entry is
+ * committed from {@link EntryWrappingInterceptor} through {@link ClusteringDependentLogic} which checks
+ * {@link CacheTopology#getWriteConsistentHash()}.
+ * <p>
+ * Entry being wrapped is a prerequisite for the command to run and therefore commit the entry, but it's not up to
+ * {@link EntryWrappingInterceptor} to make sure the entry is always wrapped - the invariant is
+ *       (can read locally) => (entry is wrapped).
```
rvansa: I meant this as invariant for interceptors below, but you're right, I am not clear in my use of fancy words :) Anyway, the other direction does not hold, entry may be wrapped because:
a) it is an L1 entry (by EWI)
b) the command is about to write the entry (by distribution interceptor) because this node is owner according to writeCH (but not readCH)
c) this is a transactional cache (by dist. int. after remoteGet)

core/src/main/java/org/infinispan/container/EntryFactory.java
```diff
@@ -1,68 +1,110 @@
...
+ * locally at all.
+ * <p>
+ * If the entry should be read locally but it's not found in DC, the entry will be wrapped (either as
+ * {@link NullCacheEntry} for reads or other appropriate type for writes). Such entry returns <code>false</code> on
+ * {@link CacheEntry#skipLookup()} as it's value is unsure (subsequent interceptors can retrieve the new value from the
+ * cache store or remote node and call {@link EntryFactory#wrapExternalEntry} to update the context.
+ * <p>
+ * With repeatable reads, the value that is context must not be overwritten by value out of the transaction (only commands
+ * in this transaction can change the context entry. That's why {@link EntryWrappingInterceptor} calls
+ * {@link CacheEntry#setSkipLookup} from the return handler for every command.
```
rvansa: What would you add?

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
```diff
@@ -192,17 +186,26 @@ private InvocationStage visitReadCommand(InvocationContext ctx, AbstractTopology
...
          AbstractTopologyAffectedCommand cmd = (AbstractTopologyAffectedCommand) rCommand;
-         int newTopologyId = Math.max(currentTopologyId(), cmd.getTopologyId() + 1);
+         int desiredTopologyId = cmd.getTopologyId() + 1;
+         if (ce instanceof OutdatedTopologyException) {
+            OutdatedTopologyException ote = (OutdatedTopologyException) ce;
+            if (ote.desiredTopologyId >= 0) {
+               desiredTopologyId = ote.desiredTopologyId;
+            }
+         }
+         int newTopologyId = Math.max(currentTopologyId, desiredTopologyId);
```
rvansa: Good catch. Thinking about names again, I'll probably rename that to `requestedTopologyId` - we shouldn't be too passionate about topologies :)

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -144,288 +144,181 @@ public final BasicInvocationStage visitClearCommand(InvocationContext ctx, Clear
...
-      Map<Address, Response> responses = rpcManager.invokeRemotely(commands, options);
-
-      Map<Object, InternalCacheEntry> entries = new HashMap<>();
-      for (Map.Entry<Address, Response> entry : responses.entrySet()) {
-         updateWithValues(((ClusteredGetAllCommand) commands.get(entry.getKey())).getKeys(),
-               entry.getValue(), entries);
+      if (entry == null) {
+         DistributionInfo info = new DistributionInfo(key, checkTopologyId(command).getWriteConsistentHash(), rpcManager.getAddress());
+         boolean load = shouldLoad(ctx, command, info);
+         if (info.isPrimary()) {
```
rvansa: It shouldn't, though I was about to handle any case correctly. I'll throw IllegalStateException instead.

core/src/main/java/org/infinispan/commands/write/PutKeyValueCommand.java
```diff
@@ -96,13 +97,17 @@ public Object perform(InvocationContext ctx) throws Throwable {
...
-      Object entryValue = e.getValue();
-      if (!valueMatcher.matches(e, null, value, valueEquivalence)) {
+      Object prevValue = e.getValue();
+      if (!valueMatcher.matches(prevValue, null, value, valueEquivalence)) {
          successful = false;
-         return entryValue;
+         return prevValue;
+      } else if (hasFlag(Flag.COMMAND_RETRY) && putIfAbsent) {
+         // TODO: hack branch!
+         e.setValue(null);
```
rvansa: Hmm, I should set `e.setCreated(true);` here as well.

danberindei: It seems wrong for the listeners to get `previous value == null` when `putIfAbsent()` is retried and `previous value == new value` when a regular `put()` is retried in otherwise identical conditions.

Even if the previous value is now correct for `putIfAbsent()`, the listener can't assume all writes will be `putIfAbsent()`, so it should still check for `event.isCommandRetried()` and not trust the previous value when it's `true`. So I'm not sure this really helps the user or just adds another special case...

rvansa: All right, it's half-way fix, and it does not break the testsuite, so removing that branch...

rvansa:
@danberindei addressed comments, squashed and rebased on top of #4457 - please merge that one first.

tristantarrant:
Needs rebase again

rvansa:
Just some imports, rebased.

rvansa:
@danberindei `GetAllCommandTest` is failing because of https://issues.jboss.org/browse/ISPN-7229 , I'll implement that, issue a separate PR and rebase this on top of that.

slaskawi:
@pruivo @danberindei @rvansa Hey Guys! Any update with this one?

rvansa:
@slaskawi currently blocked by ISPN-7229 so waiting for https://github.com/infinispan/infinispan/pull/4681

rvansa:
Rebased, though I am ok with #4669 going in first.

rvansa:
Rebased on top of #4669 
I've also reverted return value for PutMapCommand from `List<Map.Entry>` to `Map`, as it would require more changes in triangle impl (those will be needed anyway for functional commands triangelization, but that can wait).

rvansa:
Rebased with minor changes in TriangleDistributionInterceptor.

danberindei:
@rvansa I don't see new test failures, but somehow the test suite (with trace logging enabled) takes 5 mins more than master on my machine (which is already 3 mins more than the 9.5 mins I got with the triangle commit).

rvansa:
@danberindei On my machine without the tracing the build + testsuite for infinispan-core took 5:09 on master and 5:11 on this PR. With trace I would probably run out of space on /tmp if I tried to log the whole testsuite.

danberindei:
@rvansa `log4j2-trace.xml` uses a `CompressedFile` appender, and the compressed log is around 400mb for core.

danberindei:
I got a lot of variation running the test suite on the same commit, so I integrated this and I'll investigate the tests duration separately. Thanks Radim!

PR #4565, opened 2016-09-23
ISPN-6806 State transfer exception serializing L1InternalCacheEntry

danberindei:
https://issues.jboss.org/browse/ISPN-6806

Avoid sending L1 entries via state transfer.

L1 entries will still survive on the new primary owner when the last
owner leaves/crashes, but they won't be transferred to another node,
and they still expire, so they will eventually be lost.


core/src/test/java/org/infinispan/distribution/rehash/RehashWithL1Test.java
```diff
@@ -76,4 +95,14 @@ public Void call() throws Exception {
       assertEquals(0, cache(2).size());
    }
 
+   private static class MyBaseControlledConsistentHashFactory extends BaseControlledConsistentHashFactory {
```
rvansa: I would call this LastMemberIsOwnerCHF to describe the intent.

tristantarrant:
Pushed to master

PR #4566, opened 2016-09-23
ISPN-6806 State transfer exception serializing L1InternalCacheEntry

danberindei:
https://issues.jboss.org/browse/ISPN-6806

Avoid sending L1 entries via state transfer.

L1 entries will still survive on the new primary owner when the last
owner leaves/crashes, but they won't be transferred to another node,
and they still expire, so they will eventually be lost.


tristantarrant:
Pushed to 8.2.x

tristantarrant:
Also pushed to 8.1.x

PR #4569, opened 2016-09-26
ISPN-7036 Resolve classes properly when Thread context classloader is null

tsegismont:
If TCCL is null, fallback to ISPN classes classloader


slaskawi:
LGTM

//cc @tristantarrant @danberindei

tristantarrant:
We should probably always use the classloader which has loaded the Infinispan classes and ignore TCCL. Once upon a time the intention was to remove the ability for the user to specify a custom classloader. @galderz @pferraro  wdyt ?

tsegismont:
If the fix is fine with you, I can create a PR for the 8.2.x branch

2016-09-27 9:09 GMT+02:00 Tristan Tarrant <notifications@github.com>:

tristantarrant:
Thanks @tsegismont, pushed to both master and 8.2.x

PR #4574, opened 2016-09-27
Potential null pointer dereference. Minor refactoring.

rvprg:
Added a null check and moved the check from the getter for operationName inside the constructor. Now operationName takes methodName in case when mo is null.


tristantarrant:
Thanks,  pushed to both master and 8.2.x

PR #4577, opened 2016-09-29
ISPN-7063 FunctionalCachestoreTest and FunctionalInMemoryTest random …

rvansa:
https://issues.jboss.org/browse/ISPN-7063

PR #4578, opened 2016-09-30
Caffeine and converter wrapped cache changes

wburns:
https://issues.jboss.org/browse/ISPN-6998


rvansa:
@wburns I've fixed those testsuite failures in https://github.com/infinispan/infinispan/pull/4577

wburns:
@rvansa Thank you very much!  Saved me a bunch of time.  I was trying to figure out how those were failing randomly, but didn't have much luck.

rvansa:
@wburns actually it was @danberindei who realized what the problem is, I am just the coding ape...

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
```diff
@@ -100,30 +113,60 @@ protected DefaultDataContainer(int concurrencyLevel, long thresholdSize,
...
-      EntrySizeCalculator<K, InternalCacheEntry<K, V>> sizeCalculator =
-            thresholdPolicy == EvictionType.MEMORY ? new CacheEntrySizeCalculator<>(
-                    new MarshalledValueEntrySizeCalculator()) : null;
+            throw new UnsupportedOperationException("Policy not supported: " + thresholdPolicy);
+      }
+      entries = new CaffeineConcurrentMap<>(applyListener(caffeine, evictionListener).build());
+   }
+
+   private Caffeine<K, InternalCacheEntry<K, V>> applyListener(Caffeine<K, InternalCacheEntry<K, V>> caffeine, DefaultEvictionListener listener) {
+      return caffeine.executor(new WithinThreadExecutor()).removalListener((k, v, c) -> {
```
ben-manes: Can you use the default executor (FJP)? This is penalizing the caller, which might be small, but could be deferred to minimize response times.

If its just for unit tests, you could provide a same-thread executor only in that context.

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
```diff
@@ -136,13 +179,17 @@ protected DefaultDataContainer(int concurrencyLevel, long thresholdSize,
...
-            throw new IllegalArgumentException("No such eviction thread policy " + strategy);
+            throw new IllegalArgumentException("No such eviction thread policy " + policy);
       }
 
       EntrySizeCalculator<K, InternalCacheEntry<K, V>> calc = new CacheEntrySizeCalculator<>(sizeCalculator);
 
-      entries = new BoundedEquivalentConcurrentHashMapV8<>(thresholdSize, Eviction.LRU, evictionListener, keyEquivalence,
-              AnyEquivalence.getInstance(), calc);
+      com.github.benmanes.caffeine.cache.Cache<K, InternalCacheEntry<K, V>> cache = applyListener(Caffeine.newBuilder()
+            .<K, InternalCacheEntry<K, V>>weigher((k, v) -> (int) calc.calculateSize(k, v))
```
ben-manes:
```java
.weigher((K k, InternalCacheEntry<K, V> v) -> (int) calc.calculateSize(k, v))
```

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
wburns: There is an issue I believe with Passivation when an entry is evicted before it has been moved to the back end store.  If we use a different thread there is a timing issue where the object is neither in memory or in the store.  I think I might be able to use a writer only since that is synchronous by default iirc, but I believe that also has an issue.

I can try running the tests without it, but I am not confident it will behave as we need.

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
wburns: Thanks :+1:

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
ben-manes: The writer is synchronous and performed in the `compute` block that is mutating the entry. So it will be run while the entry is still visible in the hash map. The removal listener is run after its been removed from the hash map, so it does not block map operations. If asynchronous then its preferred.

If you can use the default executor and the writer, then that's probably best. If the listener calls are expensive you might be able to hand-off synchronously and process asynchronously (e.g. via a write-behind buffer like in the RxJava example). I tried to keep costs well balanced so while a same-thread executor should be fine, I do prefer if possible hiding penalties from the caller.

wburns:
Should be ready as long as tests still pass :) Haven't checked it in a while.

ben-manes:
fyi, released `2.3.4`. Please upgrade prior to merging.

wburns:
Updated.

wburns:
Looks like some total order write skew are failing intermittently..

ben-manes:
Doubtfully related, but FYI: 
The cache does have some randomness to protect against HashDoS style attacks. That means tests can't rely on predictable order of eviction.

wburns:
The test failures are also in master, if I backport just the test - so it seems like a latent bug with passivation.

wburns:
Actually the randomness is what is causing the issue, but that itself isn't the issue.  It has uncovered a bug in write skew check with a cache store.  The bug resets the version of an entry.  And the test prior always evicted the same entry thus the versions matched, but when each node evicts a different entry the versions don't match anymore.

wburns:
Finally a run passed :) http://ci.infinispan.org/viewLog.html?buildId=44231&tab=buildResultsDiv&buildTypeId=bt9

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/near/EvictInvalidatedNearCacheTest.java
```diff
@@ -37,7 +37,7 @@ public void testEvictAfterReachingMax() {
       assertClient.get(3, "v1").expectNearGetNull(3).expectNearPutIfAbsent(3, "v1");
       assertClient.get(2, "v1").expectNearGetValue(2, "v1");
       assertClient.get(3, "v1").expectNearGetValue(3, "v1");
-      assertClient.get(1, "v1").expectNearGetNull(1).expectNearPutIfAbsent(1, "v1");
+//      assertClient.get(1, "v1").expectNearGetNull(1).expectNearPutIfAbsent(1, "v1");
```
tristantarrant: Unless there is a good reason to keep this commented (with a Jira to potentially uncomment it), I'd remove it altogether

server/integration/testsuite/src/test/java/org/infinispan/server/test/cs/jdbc/SingleNodeJdbcStoreIT.java
```diff
@@ -414,9 +422,9 @@ public boolean isSatisfied() throws Exception {
     public String getStoredKey(RemoteCache cache, String key) throws IOException, InterruptedException {
         // 1. marshall the key
         // 2. encode it with base64 (that's what DefaultTwoWayKey2StringMapper does)
-        // 3. prefix it with 9 (again, done by DefaultTwoWayKey2StringMapper to mark the key as byte array type)
+        // 3. prefix it with 8 (again, done by DefaultTwoWayKey2StringMapper to mark the key as wrapped byte array type)
         // 4. prefix it with UTF-16 BOM (that is what DefaultTwoWayKey2StringMapper does for non string values)
-        return '\uFEFF' + "9" + Base64.encodeBytes(cache.getRemoteCacheManager().getMarshaller().objectToByteBuffer(key));
+        return '\uFEFF' + "8" + Base64.encodeBytes(cache.getRemoteCacheManager().getMarshaller().objectToByteBuffer(key));
```
tristantarrant: Will it be possible to load data created by a previous version ?

object-filter/src/main/java/org/infinispan/objectfilter/impl/predicateindex/ProtobufMatcherEvalContext.java
```diff
@@ -32,7 +33,12 @@ public ProtobufMatcherEvalContext(Object userContext, Object eventType, Object i
       super(userContext, eventType, instance);
       this.serializationContext = serializationContext;
       try {
-         ProtobufParser.INSTANCE.parse(this, wrappedMessageDescriptor, (byte[]) getInstance());
+         Object inst = getInstance();
+         if (inst instanceof WrappedByteArray) {
```
tristantarrant: since you know the classes, maybe inst.getClass()==WrappedByteArray.class (or even inst.getClass().isArray() would be even cheaper)

commons/src/main/java/org/infinispan/commons/util/CollectionFactory.java
```diff
@@ -101,8 +103,7 @@ public static int computeCapacity(int expectedSize, float loadFactor) {
    }
 
    public static <K, V> ConcurrentMap<K, V> makeBoundedConcurrentMap(int maxSize) {
-      return new BoundedEquivalentConcurrentHashMapV8<>(maxSize,
-            AnyEquivalence.getInstance(), AnyEquivalence.getInstance());
+      return new CaffeineConcurrentMap<>(Caffeine.newBuilder().maximumSize(maxSize).build());
```
tristantarrant: This PR doesn't actually remove the BECHMV8 classes, but they probably should go.
Also we need to document this change in both the user guide and in the upgrading doc

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/near/EvictInvalidatedNearCacheTest.java
wburns: Sure, yeah it is because the eviction is not deterministic anymore.

server/integration/testsuite/src/test/java/org/infinispan/server/test/cs/jdbc/SingleNodeJdbcStoreIT.java
wburns: Let me look into it again.

commons/src/main/java/org/infinispan/commons/util/CollectionFactory.java
wburns: I have another PR for that, already has 27K deletions, didn't want to make to completely overload this one.  I will be submitting it to Preview today.

server/integration/testsuite/src/test/java/org/infinispan/server/test/cs/jdbc/SingleNodeJdbcStoreIT.java
wburns: Hrmm I am not exactly sure the different ways users would load the data, but the Key2StringMapper can still read the value out.

However if you just upgraded the server only I don't think you could load the previous table as it is stored as a byte[].  I can unwrap the WBA and store it as a byte[] still for the JDBC classes Although this will be an issue for all of the Cache Stores not just JDBC.  WDYT?

commons/src/main/java/org/infinispan/commons/util/CollectionFactory.java
wburns: Created #4657

server/integration/testsuite/src/test/java/org/infinispan/server/test/cs/jdbc/SingleNodeJdbcStoreIT.java
tristantarrant: Can you create a Jira for this task ?

wburns:
This can be integrated in with #4657.

server/integration/testsuite/src/test/java/org/infinispan/server/test/cs/jdbc/SingleNodeJdbcStoreIT.java
wburns: Sure

wburns: Created https://issues.jboss.org/browse/ISPN-7223

core/src/main/java/org/infinispan/util/function/RemovableFunction.java
```diff
@@ -0,0 +1,14 @@
+package org.infinispan.util.function;
+
+import java.util.function.Function;
+
+/**
+ * Placeholder interface used to describe a function that can be used as is for removal.  This is useful for
+ * allowing an iterator to use removal.
+ * <p>
+ * Users shouldn't ever need to use this interface directly
```
danberindei: Could users implement it too? What are the requirements for a mapping function to be `RemovableFunction`?

commons/src/main/java/org/infinispan/commons/util/SpliteratorMapper.java
```diff
@@ -44,7 +44,13 @@ public long estimateSize() {
 
    @Override
    public int characteristics() {
-      return spliterator.characteristics();
+      int characteristics = spliterator.characteristics();
+      if (mapper instanceof DistinctFunction) {
```
danberindei: I'd rather call this `InjectiveFunction`, `DistinctFunction` makes me think that I should have a counter and generate a new value every time the function is called :)

core/src/test/java/org/infinispan/eviction/impl/BaseEvictionFunctionalTest.java
```diff
@@ -60,17 +61,9 @@ public void testSimpleEvictionMaxEntries() throws Exception {
...
-         expectedEvictions = (int) Math.ceil(CACHE_SIZE + (CACHE_SIZE * .05));
-      } else {
-         // Otherwise is LRU and that will evict on each write
-         expectedEvictions = CACHE_SIZE * 2;
-      }
-      assertEquals("eviction events count should be same with case size: " + evictionListener.getEvictedEvents(),
-            expectedEvictions, evictionListener.getEvictedEvents().size());
+      // We don't know for sure how many will be evicted due to randomness, but we know they MUST evict
+      // at least a size worth since we are writing more than double
+      assertTrue(evictionListener.evictedEntries.size() > CACHE_SIZE);
```
danberindei: Does the randomness mean that sometimes an existing key is kept in preference to an inserted one?

server/integration/testsuite/src/test/java/org/infinispan/server/test/cs/jdbc/SingleNodeJdbcStoreIT.java
danberindei: @wburns Does the format of the values change too, if using `storeAsBinary`?

commons/src/main/java/org/infinispan/util/AbstractDelegatingMap.java
```diff
@@ -9,6 +9,11 @@
    protected abstract Map<K, V> delegate();
 
    @Override
+   public V putIfAbsent(K key, V value) {
+      return delegate().putIfAbsent(key, value);
```
wburns: These were defined on AbstractDelegatingConcurrentMap, let me move them down :)

core/src/main/java/org/infinispan/cache/impl/AbstractDelegatingAdvancedCache.java
```diff
@@ -39,6 +44,7 @@
  * @author Tristan Tarrant
  * @see org.infinispan.cache.impl.AbstractDelegatingCache
  */
+@MBean(objectName = CacheImpl.OBJECT_NAME, description = "Component that represents an individual cache instance.")
 public class AbstractDelegatingAdvancedCache<K, V> extends AbstractDelegatingCache<K, V> implements AdvancedCache<K, V> {
```
wburns: I agree, that would make it cleaner.  Let me create another JIRA for that.

core/src/main/java/org/infinispan/cache/impl/AbstractDelegatingCache.java
```diff
@@ -73,7 +80,24 @@ public String getName() {
...
 
+   @ManagedAttribute(
+         description = "Returns the cache name",
+         displayName = "Cache name",
+         dataType = DataType.TRAIT,
+         displayType = DisplayType.SUMMARY
+   )
+   public String getCacheName() {
+      String name = getName().equals(BasicCacheContainer.DEFAULT_CACHE_NAME) ? "Default Cache" : getName();
+      return name + "(" + getCacheConfiguration().clustering().cacheMode().toString().toLowerCase() + ")";
```
wburns: :)

core/src/test/java/org/infinispan/eviction/impl/BaseEvictionFunctionalTest.java
ben-manes: The cache uses an admission window to monitor the new arrivals. When an insert pushes an entry out of this window into the next generation, that "candidate" entry and the eviction policy's victim are evaluated. Which ever has the largest frequency is retained. There are a few unpredictable aspects to this,
 - The frequencies are stored in a popularity sketch so collisions can raise the estimate
 - A "warm" candidate may be retained instead of a "hot" victim, randomly, to protect from HashDoS attacks.
 - A large "weighted" entry might cause the admission window to flush, but then itself not pass the filter and be removed. Imagine a cache size of 2MB with 100kb entries, and then the insertion of a 1.5MB entry. If that large entry has no historic frequency, it might be rejected when it forces itself to be promoted.

core/src/main/java/org/infinispan/cache/impl/CompatibilityAdvancedCache.java
```diff
@@ -0,0 +1,136 @@
...
+            } catch (Exception e) {
+               throw new CacheException("Unable to unmarshall return value");
+            }
+         }
+
+         if (target instanceof byte[]) {
+            // Try standard deserialization
+            try {
+               ObjectInputStream ois = new ObjectInputStream(new ByteArrayInputStream((byte[]) target));
+               return ois.readObject();
```
wburns: This is the same as before https://github.com/infinispan/infinispan/blob/f9945e7707ca107c9ce0bfaf8808103685300d36/core/src/main/java/org/infinispan/interceptors/compat/TypeConverterInterceptor.java#L93

core/src/main/java/org/infinispan/interceptors/impl/CompatibilityInterceptor.java
```diff
@@ -0,0 +1,43 @@
...
+      return super.visitReplaceCommand(ctx, command);
+   }
+
+   protected void addVersionIfNeeded(MetadataAwareCommand cmd) {
+      Metadata metadata = cmd.getMetadata();
+      if (metadata.version() == null) {
+         Metadata newMetadata = metadata.builder()
+               .version(versionGenerator.generateNew())
+               .build();
+         cmd.setMetadata(newMetadata);
```
wburns: I wish I knew, this is another thing where it was like this way before, heh

https://github.com/infinispan/infinispan/blob/f9945e7707ca107c9ce0bfaf8808103685300d36/core/src/main/java/org/infinispan/interceptors/compat/BaseTypeConverterInterceptor.java#L195

I can log a JIRA though so someone can look into it further.

commons/src/main/java/org/infinispan/commons/util/SpliteratorMapper.java
wburns: Sure sounds good.

wburns:
@danberindei I will add the updates to the off heap PR #4673

server/integration/testsuite/src/test/java/org/infinispan/server/test/cs/jdbc/SingleNodeJdbcStoreIT.java
danberindei: I was thinking about `storeAsBinary` in 9.0.0.Alpha1 vs 9.0.0.Beta1. I suspect the values do change, because we're including the byte array length with the encoding scheme used by `MarshallUtil.marshallInt`, whereas Alpha1 is probably just passing the byte array to the store.

core/src/main/java/org/infinispan/util/function/RemovableFunction.java
danberindei: Yeah, it doesn't make a lot of sense for the users to implement it.

The naming bothers me a bit, but I can't think of a better one... So instead, I'm wondering if you can just assume that in a compatibility/storeAsBinary cache the first mapping will be a conversion? It's not like an iterator without the conversion function could support `remove()` anyway, right?

Off-topic, what's the use case for going from a stream to a mutable iterator? I noticed the JDK streams don't support removal.

core/src/test/java/org/infinispan/eviction/impl/BaseEvictionFunctionalTest.java
danberindei: Thanks for the details @ben-manes. I knew there's some randomness as to which element is evicted, I was just surprised that the entry being inserted is also considered for eviction.

I was also wondering whether it's ok not to call the eviction listener for the inserted entry in this case.

ben-manes: It won't be unless weights are bigger than a tolerance. It would be so too in your older version if the weight of one entry exceeded the maximum capacity. That could either flush the entire cache or only reject the entry, the latter being what Caffeine does. So it should have always been possible, just rare in both implementations. I think had LIRS been weighted it would have had the same issue, as the HIR region is similar to our admission window.

danberindei: @ben-manes This test doesn't use weights at all, so that's not an issue.

The explanation was much simpler than what I was imagining: even if the test never accesses the keys after insertion, sometimes an older key will be retained instead of a newer one. Since the test inserts `2 * CACHE_SIZE` entries and then re-inserts the first `CACHE_SIZE` entries, the re-insert loop sometimes finds the entry already in the cache and doesn't have to evict anything. We could just change the test to always insert new keys in the cache, and the number of evictions would be guaranteed to be `2 * CACHE_SIZE`.

danberindei:
@wburns this failure sounds like could be related:

    testRemoveFailWriteSkewWithPassivation(org.infinispan.container.versioning.DistWriteSkewTest)  Time elapsed: 0.129 sec  <<< FAILURE!
    java.lang.AssertionError: The key was not evicted after 10 inserts
	at org.testng.AssertJUnit.assertTrue(AssertJUnit.java:24)
	at org.infinispan.container.versioning.AbstractClusteredWriteSkewTest.doTestWriteSkewWithPassivation(AbstractClusteredWriteSkewTest.java:316)
	at org.infinispan.container.versioning.AbstractClusteredWriteSkewTest.testRemoveFailWriteSkewWithPassivation(AbstractClusteredWriteSkewTest.java:152)

PR #4579, opened 2016-09-30
ISPN-7067 Cache.evict() sometimes removes instead of passivating

danberindei:
https://issues.jboss.org/browse/ISPN-7067


tristantarrant:
Pushed to master. Thanks

PR #4581, opened 2016-09-30
[8.2.x] ISPN-7064 RPC to leaver times out instead of finishing immediately

danberindei:
https://issues.jboss.org/browse/ISPN-7064

* Set the GroupRequest listener before the request is sent.
* Send the state transfer cancel commands asynchronously.


slaskawi:
Integrated, thanks Dan!

PR #4582, opened 2016-09-30
Necessary null checks added because WeakReference.get() may return null.

rvprg:
Some weak references were used without checking for null.


tristantarrant:
Pushed to master and 8.2.x, thanks

PR #4590, opened 2016-10-04
stream documentation

wburns:
Add stream documentation regarding serialization
Also removed html which was accidentally committed.


tristantarrant:
Pushed to master. Thanks

PR #4592, opened 2016-10-05
[8.1.x] ISPN-6857 OutdatedTopologyException in clustered invalidation cache

panossot:
https://issues.jboss.org/browse/ISPN-6857

### DIFF
--- a/core/src/main/java/org/infinispan/util/logging/Log.java
+++ b/core/src/main/java/org/infinispan/util/logging/Log.java
@@ -1416,5 +1416,8 @@ public interface Log extends BasicLogger {
    @LogMessage(level = WARN)
    @Message(value = "Received unsolicited state from node %s for segment %d of cache %s", id = 396)
    void ignoringUnsolicitedState(Address node, int segment, String cacheName);
+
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/infinispan/util/logging/Log.java
```diff
@@ -1416,5 +1416,8 @@ void preparedTxAlreadyExists(RecoveryAwareTransaction previous,
    @LogMessage(level = WARN)
    @Message(value = "Received unsolicited state from node %s for segment %d of cache %s", id = 396)
    void ignoringUnsolicitedState(Address node, int segment, String cacheName);
+
+   @Message(value = "Cannot enable '%s' in invalidation caches!", id = 397)
```
tristantarrant: You cannot introduce a logging ID which conflicts with one used by future versions. You should use id 420.

tristantarrant:
Also please amend the commit log to include the Jira.

core/src/main/java/org/infinispan/util/logging/Log.java
panossot: It should be ok now.

tristantarrant:
Pushed to 8.1.x
Thanks

PR #4594, opened 2016-10-05
ISPN-7078 Improve the clustering section of the user guide

danberindei:
https://issues.jboss.org/browse/ISPN-7078


documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -258,111 +355,96 @@ the initial nodes do not appear within the specified timeout, the cache manager
...
-Based on a supplied cluster
-link:http://docs.jboss.org/infinispan/{infinispanversion}/apidocs/org/infinispan/remoting/transport/Address.html[address]
-identifying the node, the service returns a key that will be hashed to that particular
-node.
+In a distributed cache, a key is allocated to a list of nodes with an opaque algorithm.
+There is no easy way to reverse the computation and generate a key that maps to a
+particular node.
+However, we can generate a sequence of (pseudo-)random keys, see what their primary
+owner is, and hand them out to the application when it needs a key mapping to a
+particular node.
```
tristantarrant: Maybe we should mention the AffinityPartitioner here ? /cc @gustavonalle

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -449,79 +519,81 @@ Or, if you are using XML:
...
+----
+public interface Grouper<T> {
+    String computeGroup(T key, String group);
+
+    Class<T> getKeyType();
+}
+----
 
-Let's take a look at an example of a _Grouper_:
+If multiple `Grouper` s are configured for the same key type, all of them will be called,
```
tristantarrant: The separate 's' makes it look weird... Maybe use "Grouper classes" instead ?

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -545,121 +617,118 @@ Or, if you are using XML:
...
- * Note that, concurrent addition perform by other transactions/threads to the group may not be removed.
- *
- * @param groupName the group name.
- */
-void removeGroup(String groupName);
-----
+`getGroup(groupName)`::
+  Retrieves all keys in the cache that belong to a group.
+`removeGroup(groupName)`::
+  Removes all the keys in the cache that belong to a group.
```
tristantarrant: When describing an API, pointing to the relevant JavaDocs is a good ida.

tristantarrant:
Good job on using SVG for the diagrams. Now they are also editable.

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1,211 +1,308 @@
...
+Instead, when a key is written to, the cache only aims to remove data that may be stale
+from other nodes.
+This cache mode only makes sense if you have another, permanent store for your data such
+as a database and are only using Infinispan as an optimization in a read-heavy system,
+to prevent hitting the database for every read.
+If a cache is configured for invalidation, every time data is changed in a cache, other
+caches in the cluster receive a message informing them that their data is now stale and
+should be removed from memory and from any local store.
+
+.Invalidation mode
```
slaskawi: I've already seen a question about clearing invalidation cache... May I ask you to say a few words how to do it?

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1,211 +1,308 @@
...
+  It always assigns a key to the same node in every cache as long as the cluster is
+  symmetric (i.e. all caches run on all nodes).
+  It does have some weaknesses: the load distribution is a bit uneven, and it also moves
+  more segments than strictly necessary on a join or leave.
+TopologyAwareSyncConsistentHashFactory::
+  Selected automatically when <<ServerHinting, Server Hinting>> is enabled.
+  Similar to _SyncConsistentHashFactory_, but adapted for
+  <<ServerHinting, Server Hinting>>.
+DefaultConsistentHashFactory::
+  The old default.
```
slaskawi: The doesn't sound good... Maybe something like - "using prior to Infinispan X" ?

documentation/src/main/asciidoc/user_guide/clustering.adoc
danberindei: The `AffinityPartitioner` works more like grouping, because it can only guarantee that 2 keys are mapped to the same owners and not that a key is mapped to a particular node.

documentation/src/main/asciidoc/user_guide/clustering.adoc
danberindei: I believe you just need to call `cache.clear()`, like in any other mode, so it's not worth mentioning it here. If `cache.clear()` doesn't work, then perhaps we have a bug.

danberindei:
Updated.

tristantarrant:
Merged, thanks!

PR #4595, opened 2016-10-05
ISPN-6045: Fix transaction aware iterator remove

lamperi:
Fixed the problem where TransactionAwareKeyCloseableIterator did not set
the previousValue field, which made any calls to remove() throw NPE.
Additionally, the iterator now throws IllegalStateException if next()
has not been called, or remove() has been already called for the current
item as Iterator interface specifies.

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
@@ -595,7 +595,11 @@ public class TxInterceptor<K, V> extends DDAsyncInterceptor implements JmxStatis
 
       @Override
       public void remove() {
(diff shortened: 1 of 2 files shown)


wburns:
@lamperi Thanks!  Not sure how I didn't notice the JIRA originally.  Unfortunately the code at [1] had it correct and wasn't copied to the tx part :(  Thanks again for noticing.

[1] https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/stream/impl/RemovableIterator.java

wburns:
Pulling...

wburns:
Integrated... thanks @lamperi !

lamperi:
Thanks for the quick merge 👍

PR #4597, opened 2016-10-05
Documentation: Moved CLI chapt under Management Tool

rigazilla:
Moved CLI under management tool as include file. Adjusted sequence level


tristantarrant:
Pushed, thanks

PR #4598, opened 2016-10-06
Documentation: Lock and Transactions reviewed

pruivo:
Reviewed the locking and transaction documentation.
Didn't redone the images :( yet.


tristantarrant:
forceCommit.png should be cropped to remove the window decorations.

documentation/src/main/asciidoc/user_guide/locking.adoc
```diff
@@ -29,112 +36,101 @@ NOTE: If the operation is conditional and it fails on the primary owner, it is n
...
+Similar to the way the JDK's `ConcurrentHashMap` allocates locks, this allows for a highly scalable, fixed-overhead locking mechanism in exchange for potentially unrelated entries being blocked by the same lock.
 
-The alternative is to disable lock striping - which would mean a _new_ lock is created per entry.  This approach _may_ give you greater concurrent throughput, but it will be at the cost of additional memory usage, garbage collection churn, etc. 
+The alternative is to disable lock striping - which would mean a _new_ lock is created per entry.
+This approach _may_ give you greater concurrent throughput, but it will be at the cost of additional memory usage, garbage collection churn, etc.
 
 .Default lock striping settings
-IMPORTANT: From Infinispan 5.0, lock striping is disabled by default, due to potential deadlocks that can happen if locks for different keys end up in the same lock stripe. Previously, in Infinispan 4.x lock striping used to be enabled by default.
+IMPORTANT: From Infinispan 5.0, lock striping is disabled by default, due to potential deadlocks that can happen if locks for different keys end up in the same lock stripe.
```
rvansa: Can those deadlocks happen in non-tx mode, too?

documentation/src/main/asciidoc/user_guide/locking.adoc
```diff
@@ -29,112 +36,101 @@ NOTE: If the operation is conditional and it fails on the primary owner, it is n
...
+If another transaction, `TX2`, is started on `B` (or any other node) and `TX2` tries to lock `K` then it will fail with a timeout as the lock is already held by `TX1`.
+The reason for this is the that the lock for a key `K` is always, deterministically, acquired on the same node of the cluster, regardless of where the transaction originates.
 
-----
+===  Data Versioning
+Infinispan supports two forms of data versioning: simple and external.
+The simple versioning is used in transactional caches for write skew check.
+Check section <<tx:write-skew>> section for detail about it.
+
+The external versioning is used to encapsulate an external source of data versioning within Infinispan, such as when using Infinispan with Hibernate which in turn gets its data version information directly from a database.
```
rvansa: So how does Infinispan use that info? How do you configure it?

documentation/src/main/asciidoc/user_guide/locking.adoc
```diff
@@ -10,9 +10,16 @@ MVCC offers many advantages over coarse-grained Java synchronization and even JD
 === Locking implementation details
 Infinispan's MVCC implementation makes use of minimal locks and synchronizations, leaning heavily towards lock-free techniques such as link:http://en.wikipedia.org/wiki/Compare-and-swap[compare-and-swap] and lock-free data structures wherever possible, which helps optimize for multi-CPU and multi-core environments. 
 
-In particular, Infinispan's MVCC implementation is heavily optimized for readers.  Reader threads do not acquire explicit locks for entries, and instead directly read the entry in question.
+In particular, Infinispan's MVCC implementation is heavily optimized for readers.
```
pruivo: IMO, yes. We copy the value to the invocation context and the operation works on that copy.

documentation/src/main/asciidoc/user_guide/locking.adoc
pruivo: I'll remove the sentence. 
The deadlock can happen in non-tx cache but it is not caused by the lock striping.

documentation/src/main/asciidoc/user_guide/locking.adoc
pruivo: it doesn't use it. 
it only says that the user can store its own version (via Metadata). from that, it can install custom interceptors to manage it if needed.

pruivo:
updated!

tristantarrant:
Pushed to master. Thanks @pruivo

PR #4600, opened 2016-10-07
ISPN-7080 NPE in CacheNotifierImpl by LIRS eviction listener

tedwon:
https://issues.jboss.org/browse/ISPN-7080

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8.java
+++ b/commons/src/main/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8.java
@@ -1140,7 +1140,6 @@ public class BoundedEquivalentConcurrentHashMapV8<K,V> extends AbstractMap<K,V>
                   // Non resident was already evicted and now it is no longer in the
                   // queue or the stack so it is effectively gone - however we want to
                   // remove the now null node
-                  removedLIR.setState(Recency.EVICTING);
                   Collection<LIRSNode<K, V>> nodesToEvict = nodesToEvictTL.get();
                   if (nodesToEvict == null) {
                      nodesToEvict = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/infinispan/eviction/impl/BaseEvictionFunctionalTest.java
```diff
@@ -44,6 +48,13 @@ public void testSimpleEvictionMaxEntries() throws Exception {
       }
       Thread.sleep(1000); // sleep long enough to allow the thread to wake-up
       assert CACHE_SIZE >= cache.size() : "cache size too big: " + cache.size();
+      assert CACHE_SIZE == evictionListener.getEvictedEvents() : "eviction events count should be same with case size: " + evictionListener.getEvictedEvents();
+
+      for (int i = 0; i < CACHE_SIZE; i++) {
+         cache.put("key-" + (i + 1), "value-" + (i + 1), 1, TimeUnit.MINUTES);
+      }
+      Thread.sleep(1000); // sleep long enough to allow the thread to wake-up
+      assert CACHE_SIZE >= cache.size() : "cache size too big: " + cache.size();
```
danberindei: I'd use `AbstractInfinispanTest.eventually()` to wait for a shorter time when the reaper finishes removing everything in 100ms.

wburns: Actually these tests are a bit out dated, eviction should happen on the same thread.  So a sleep shouldn't be required.  Although I am not sure why this is also testing expiration at the same time.

commons/src/main/java/org/infinispan/commons/util/concurrent/jdk8backported/BoundedEquivalentConcurrentHashMapV8.java
```diff
@@ -1140,7 +1140,6 @@ void demoteLowestLIR() {
                   // Non resident was already evicted and now it is no longer in the
                   // queue or the stack so it is effectively gone - however we want to
                   // remove the now null node
-                  removedLIR.setState(Recency.EVICTING);
```
wburns: Unfortunately this fix is not quite correct.  This will still cause an additional eviction event to be generated.  You can see it by adding an additional line to the test to verify the correct number of events are generated.

We instead need to lock the segment and the node and remove the element if it is still HIR_NONRESIDENT (someone else could have updated) and not generate an event.  The event was generated when it went from HIR_RESIDENT to HIR_NONRESIDENT.

wburns:
Superseded by #4639

PR #4603, opened 2016-10-07
Turned DuplicatedDomainsCdiIT back on

slaskawi:
This is a trivial fix but I also upgrade shrinkwrap here. Please let CI run it at least twice before integrating.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -160,7 +160,7 @@
       <version.org.picketbox>4.0.20.Final</version.org.picketbox>
       <version.postgresql.driver>9.3-1101-jdbc41</version.postgresql.driver>
       <version.resteasy>3.1.0.Beta1</version.resteasy>
-      <version.shrinkwrapResolver>2.1.0</version.shrinkwrapResolver>
+      <version.shrinkwrapResolver>2.2.4</version.shrinkwrapResolver>
       <version.slf4j>1.7.21</version.slf4j>
       <version.slf4j-jboss-logging>1.1.0.Final</version.slf4j-jboss-logging>
       <version.snappy>1.0.5</version.snappy>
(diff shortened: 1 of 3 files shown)


tristantarrant:
This will need to be rebased on top of the new combined modules

tristantarrant:
Please rebase

parent/pom.xml
```diff
@@ -493,6 +493,20 @@
...
          </dependency>
          <dependency>
+            <groupId>org.jboss.shrinkwrap.resolver</groupId>
+            <artifactId>shrinkwrap-resolver-bom</artifactId>
+            <version>${version.shrinkwrapResolver}</version>
+            <type>pom</type>
+            <scope>import</scope>
+         </dependency>
+         <!-- This is a work around for https://issues.jboss.org/browse/SHRINKRES-238 -->
+         <!-- After fixing the issue above, the dependency below can be safely removed -->
```
danberindei: How about the `import` dependency above? Would you still need it if shrinkwrap resolved transitive dependencies properly?

slaskawi: The problem is that we import `wildfly-core-parent` that has a dependency to `maven-aether-provider`. Here's the relevant part of the output of executing `mvn -X dependency:tree`:

```
org.apache.maven:maven-aether-provider:jar:3.2.1:test (version managed from 3.2.5 by org.wildfly.core:wildfly-core-parent:2.2.0.Final)
```

So Maven figures out that `maven-aether-provider` provided by WF Core is [closer](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Transitive_Dependencies) than the one provided by `shrinkwrap-resolver-bom`. The only way to override this (at least to my knowledge) is to add a specific version to `dependencyManagement`. 

Sucks.. I know..

slaskawi: By Uber Jars I assume you mean `shrinkwrap-resolver-bom`...

It is the recommended way of using the resolvers: https://github.com/shrinkwrap/resolver
> Obtaining ShrinkWrap Resolvers for use in your system can be done in a single pass by declaring a dependency upon the depchain module in a Maven pom.xml:
```
<dependencies>
    ...
    <dependency>
      <groupId>org.jboss.shrinkwrap.resolver</groupId>
      <artifactId>shrinkwrap-resolver-depchain</artifactId>
      <version>${version.shrinkwrap.resolvers}</version>
      <scope>test</scope>
      <type>pom</type>
    </dependency>
    ...
</dependencies>
```

However it seems you are correct - `shrinkwrap-resolver-impl-maven` can be removed. However I still couldn't get rid of `maven-aether-provider`.

danberindei: By uber-jar I meant "org.infinispan:infinispan-cdi-embedded" in the test -- I mistook it for one of our uber-jars, because that's what I thought SHRINKRES-238 was about. Sorry about that!

I was also puzzled by `<scope>import</scope>`, I don't think I ever saw this being used before. I see in your link that was the recommended way of adding the dependency with Arquillian prior to 1.1.0.Final, but I guess it's fine to keep it that way.

danberindei:
Integrated, thanks Sebastian!

PR #4606, opened 2016-10-10
ISPN-7097 Fix random failures in FlagsReplicationTest

rvansa:
https://issues.jboss.org/browse/ISPN-7097

Make sure that the CommitThread runs the transaction started in LockingThread.

### DIFF
--- a/core/src/test/java/org/infinispan/replication/FlagsReplicationTest.java
+++ b/core/src/test/java/org/infinispan/replication/FlagsReplicationTest.java
@@ -3,11 +3,7 @@ package org.infinispan.replication;
 import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertFalse;
 import static org.testng.AssertJUnit.assertNull;
-import static org.testng.AssertJUnit.assertTrue;
-
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


danberindei:
@rvansa I'd like a JIRA, for one I'd like to see an explanation why this doesn't fail every single time :)

danberindei:
Looking at the code, I'm not sure why the test is using separate threads at all. `haveSecondaryThreadTakeLock` could just as well start a tx, acquire the lock, suspend the tx and return it, with `haveSecondaryThreadReleaseLock` resuming the tx and committing it.

rvansa:
@danberindei I think that it does not fail all the time because threads are recycled and there is a good chance that the FJP will assign the operation the same thread. If it does not, there won't be any transaction to commit.
TBH I forgot if I've managed to trace the actual behaviour, I have it in my commit pool for quite some time :) But changing that to single-threaded test makes sense, I'll do that.

rvansa:
@danberindei here you come....

danberindei:
@rvansa That's why it's good to create a JIRA, it means you have it written down and you don't have to remember everything. E.g. `AbstractInfinispanTest.fork()` actually uses a `ThreadPoolExecutor`, not FJP :)

danberindei:
Integrated, thanks Radim!

PR #4607, opened 2016-10-10
ISPN-7090 Asynchronous invocation doesn't work with multiple futures

danberindei:
https://issues.jboss.org/browse/ISPN-7090


slaskawi:
Thanks Dan!

PR #4610, opened 2016-10-10
[8.2.x] ISPN-7074 Provide a copier for indexed entities

gustavonalle:


gustavonalle:
The failures were already there, integrating

PR #4619, opened 2016-10-17
ISPN-7121 Remove Expired doesn't work with no auto commit in a clustered

wburns:
cache

* Make sure to start transactions when required

https://issues.jboss.org/browse/ISPN-7121


tristantarrant:
Pushed to 8.2.x, thanks !

PR #4620, opened 2016-10-18
ISPN-6906 (1/3) Minor marshalling adjustments

galderz:
- Add AnyServerEquivalence externalizer
- Fixed VersionAwaremarshallerTest tests so that they verify that the
  data read is the expected one.
- Add externalizers for Equivalence implementations.
- Make TransactionInfo use MarshallUtil.marshallArray to marshall the
  modification array, making it easier for the internal marshaller to
  avoid having to support WriteCommand[] directly.
- Make ClusterEvent use marshallEnum/unmarshallEnum for event's type
  enumeration instance.
~~- Add @SerializeWith() annotation was added along with a simple
  externalizer for those classes that would have delegated into standard
  Java serialization. This happened for CompositeKeyValueFilterConverter,
  ControlledConsistentHashFactory and ReplicatedControlledConsistentHashFactory.~~
- Added missing externalizers for Java's UUID, ArrayDeque,
  ImmutableSetWrapper, java.util.Collections$UnmodifiableRandomAccessList,
  java.util.Collections$EmptyList, java.util.Collections$SingletonMap,
  java.util.Collections$EmptyMap, java.util.Collections$SingletonSet,
  java.util.Collections$SynchronizedSet and
  java.util.Collections$UnmodifiableSet
- Added java.util.UUID advanced externalizer.
- A test has been added to main marshalling test to verify @SeralizeWith
  annotated instances are marshalled and unmarshalled correctly.
- Add ConcurrentHashMap to MapExternalizer.
- Added MIMECacheentry externalizer.
- Added extra checks to
  BaseStoreFunctionalTest.testPreloadStoredAsBinary
  so that it verifies the preloaded store-as-binary payloads can
  be correctly unmarshalled.


pruivo:
@galderz are the 3 PRs open going to replace the big one (#4530)?

commons/src/main/java/org/infinispan/commons/marshall/exts/EquivalenceExternalizer.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.ObjectOutput;
+import java.util.Set;
+
+public final class EquivalenceExternalizer extends AbstractExternalizer<Equivalence> {
+
+   private static final int BYTE_ARRAY_EQ    = 0x00;
+   private static final int ANY_EQ           = 0x01;
+   private static final int ANY_SERVER_EQ    = 0x02;
+
+   private final IdentityIntMap<Class<?>> subIds = new IdentityIntMap<>(2);
```
pruivo: initial value should be 3

core/src/test/java/org/infinispan/filter/CompositeKeyValueFilterConverter.java
```diff
@@ -14,6 +19,7 @@
  * @author wburns
  * @since 7.0
  */
+@SerializeWith(CompositeKeyValueFilterConverter.Ext.class)
```
pruivo: why using the annotation instead of creating an ```AdvancedExternalizer``` like in the other classes?

core/src/main/java/org/infinispan/notifications/cachelistener/cluster/ClusterEvent.java
```diff
@@ -175,7 +176,7 @@ public void writeObject(ObjectOutput output, ClusterEvent object) throws IOExcep
       public ClusterEvent readObject(ObjectInput input) throws IOException, ClassNotFoundException {
 
          return new ClusterEvent(input.readObject(), input.readObject(), input.readObject(),
-                                 (Metadata)input.readObject(),(Type)input.readObject(),
+                                 (Metadata)input.readObject(),(Type) MarshallUtil.unmarshallEnum(input, Type::valueOf),
```
pruivo: unneeded cast :)

core/src/main/java/org/infinispan/remoting/MIMECacheEntry.java
```diff
@@ -47,4 +55,35 @@ public boolean equals(Object o) {
...
+      @Override
+      public Set<Class<? extends MIMECacheEntry>> getTypeClasses() {
+         return Util.<Class<? extends MIMECacheEntry>>asSet(MIMECacheEntry.class);
+      }
+
+      @Override
+      public void writeObject(ObjectOutput out, MIMECacheEntry obj) throws IOException {
+         out.writeUTF(obj.contentType);
+         out.writeInt(obj.data.length);
+         out.write(obj.data);
```
pruivo: use ```MarshallUtil.marshallByteArray(obj.data, out);```

core/src/main/java/org/infinispan/marshall/exts/QueueExternalizers.java
```diff
@@ -0,0 +1,54 @@
...
+import org.jboss.marshalling.util.IdentityIntMap;
+
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.util.ArrayDeque;
+import java.util.Queue;
+import java.util.Set;
+
+public final class QueueExternalizers extends AbstractExternalizer<Queue> {
```
pruivo: this looks a lot like ```ListExternalizer```. Don't you want to put the queues in ```ListExternalizer```?
If not, change the name to ```QueueExternalizer``` (without the ```s``` in the end)

core/src/test/java/org/infinispan/util/ReplicatedControlledConsistentHashFactory.java
```diff
@@ -15,6 +22,7 @@
  * @author Dan Berindei
  * @since 7.0
  */
+@SerializeWith(ReplicatedControlledConsistentHashFactory.Ext.class)
```
pruivo: why the annotation?

core/src/test/java/org/infinispan/util/ReplicatedControlledConsistentHashFactory.java
```diff
@@ -85,4 +98,22 @@ public ReplicatedConsistentHash union(ReplicatedConsistentHash ch1, ReplicatedCo
    public void setMembersToUse(List<Address> membersToUse) {
       this.membersToUse = membersToUse;
    }
+
+   public static final class Ext implements Externalizer<ReplicatedControlledConsistentHashFactory> {
+
+      @Override
+      public void writeObject(ObjectOutput output, ReplicatedControlledConsistentHashFactory object) throws IOException {
+         MarshallUtil.marshallCollection(object.membersToUse, output);
+         output.writeObject(object.primaryOwnerIndices);
```
pruivo: well, you can add a new method to ```MarshallUtil``` to (un)marshall the int array.

core/src/main/java/org/infinispan/statetransfer/TransactionInfo.java
```diff
@@ -78,17 +80,17 @@ public Integer getId() {
...
 
       @Override
       @SuppressWarnings("unchecked")
       public TransactionInfo readObject(ObjectInput input) throws IOException, ClassNotFoundException {
          GlobalTransaction globalTransaction = (GlobalTransaction) input.readObject();
          int topologyId = input.readInt();
-         WriteCommand[] modifications = (WriteCommand[]) input.readObject();
-         Set<Object> lockedKeys = (Set<Object>) input.readObject();
+         WriteCommand[] modifications = MarshallUtil.unmarshallArray(input, WriteCommand[]::new);
+         Set<Object> lockedKeys = (Set<Object>) MarshallUtil.unmarshallCollection(input, HashSet::new);
```
pruivo: unneeded cast

core/src/test/java/org/infinispan/util/ReplicatedControlledConsistentHashFactory.java
danberindei: +1 to have an explicit method for array of ints

core/src/test/java/org/infinispan/filter/CompositeKeyValueFilterConverter.java
galderz: > why using the annotation instead of creating an AdvancedExternalizer like in the other classes?

Probably cos it wasn't a class that's needed by core per se. I mean, the user can use it for doing dealing with composite keys but it felt like an edge case rather than a common case. Then again, the rest of of helper classes, e.g. `KeyFilterAsKeyValueFilter` have advanced externalizers pre-registered, so I'll change that accordingly.

galderz: > Also, shouldn't the externalizer class be called Externalizer?

Are we in the mood of nitpicking? ;)

Since you are asking, having named the externalizer `Externalizer`, I would have had to use the fully qualified name in `implements org.infinispan.commons.marshall.Externalizer`. So instead of doing that, I simply called it `Ext` which is the prefix of `Externalizer`.

Hope that clears up this confusion ;)

galderz: > why using the annotation instead of creating an AdvancedExternalizer like in the other classes?

Actually, it's a test class so it's fine as is :)

core/src/main/java/org/infinispan/remoting/MIMECacheEntry.java
galderz: Why specifically? That method is calling this signature:

`write(byte b[])`

And that signature is just fine as is, it won't require any type lookups.

core/src/test/java/org/infinispan/filter/CompositeKeyValueFilterConverter.java
galderz: In fact, this change is not required, or at least not in its current form.

I originally put it in when trying to make sure no Infinispan classes were being serialized with the assumption that they extended Serializable. Hence I added the annotation based externalizer for this test class. However, once #4627 is in, it will probably just extend `ExternalPojo` and it won't need the annotation.

So, I'll go ahead and remove this change.

galderz:
> I was hoping this PR included throwing an exception when one of the tests relied on a Serializeable class

^ Not sure if it's clear, but in order to avoid noise the work for ISPN-6906 was split into 3 PRs. It's the 3rd PR (#4627) that contains this logic.

> I would have liked to play more with using MarshallUtil methods for collections instead of adding new externalizers :)

It's my fault that new externalizers were added. This was the result of adding logic to make sure none of the classes we used relied on implementing Serializable. This logic comes in inside the 3rd PR (#4627).

We should certainly use, were it makes sense, MarshallUtil for collection related marshalling, but that's an extra task on top of the work here.

galderz:
All sorted. Couple of comments need clarification.

commons/src/main/java/org/infinispan/commons/util/Immutables.java
```diff
@@ -545,6 +545,30 @@ public ImmutableEntrySetWrapper(Set<? extends Entry<? extends K, ? extends V>> s
...
+
+      @Override
+      public Integer getId() {
+         return Ids.IMMUTABLE_SET;
+      }
+
+      @Override
+      @SuppressWarnings("unchecked")
+      public Set<Class<? extends Set>> getTypeClasses() {
+         return Util.<Class<? extends Set>>asSet(ImmutableSetWrapper.class);
```
galderz: Makes sense but, as mentioned below, I'd rather split this out of ISPN-6906.

commons/src/main/java/org/infinispan/commons/marshall/Ids.java
```diff
@@ -25,4 +25,8 @@
    int LAMBDA_WITH_METAS = 160;
 
    int IMMUTABLE_ENTRY = 167;
+
+   int IMMUTABLE_SET = 173;
+
+   int EQUIVALENCE = 174;
```
galderz: Or when Will removes equivalence classes he can remove this ;)

galderz:
Oh, missed some final comments from Dan. I've commented on those now and sorted the Map externalizer comment.

core/src/test/java/org/infinispan/filter/CompositeKeyValueFilterConverter.java
pruivo: ops, I didn't notice it was a test class :)

core/src/main/java/org/infinispan/marshall/exts/QueueExternalizers.java
pruivo: yes :+1:

danberindei: I see there's a `CollectionExternalizer` already, but it only deals with `ReadOnlySegmentAwareCollection`.

core/src/main/java/org/infinispan/remoting/MIMECacheEntry.java
galderz: I can certainly use `marshallByteArray()` but I'd argue that the `marshallInt` call inside is doing extra work that's not required specifically for byte[]. E.g. there's no need to check for any negative numbers.

On top of that, since you're talking about duplication, what does `MarshallUtil.marshallInt` do differently to `UnsignedNumeric.writeUnsignedInt`? Is the difference dealing with negative numbers or is there more to it?

galderz: I think calling `obj.write(byte[])` in this case is just fine. When the new marshaller is in place (PR 2/3), we can discuss further how to come up with a single method to write byte arrays. At the end of the day, `obj.write(byte[])` should call up to that agreed implementation, whether that is `MarshallUtil.marshallByteArray` or any other impl.

core/src/main/java/org/infinispan/marshall/exts/QueueExternalizers.java
galderz: We spotted Dan, I'll group them all together inside CollectionExternalizer.

galderz:
I've added a new commit with a regrouping of collection externalizers into `CollectionExternalizer`. That should be it.

core/src/main/java/org/infinispan/remoting/MIMECacheEntry.java
pruivo: the main difference is the ```MarshallUtil.marshallByteArray``` checks and handles ```null``` while ```obj.write(byte[])``` throws NPE

core/src/main/java/org/infinispan/marshall/exts/CollectionExternalizer.java
```diff
@@ -10,29 +10,97 @@
...
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
```
pruivo: unused import

core/src/main/java/org/infinispan/marshall/exts/CollectionExternalizer.java
```diff
@@ -10,29 +10,97 @@
...
    }
 
    @Override
    public Collection readObject(ObjectInput input) throws IOException, ClassNotFoundException {
       int magicNumber = input.readUnsignedByte();
       switch (magicNumber) {
+         case ARRAY_LIST:
+            return MarshallUtil.unmarshallCollection(input, ArrayList::new);
+         case LINKED_LIST:
+            return MarshallUtil.unmarshallCollection(input, s -> new LinkedList<>());
```
pruivo: you can use ```unmarshallCollectionUnbounded()``` when the collection constructor does not need the ```size```

core/src/main/java/org/infinispan/marshall/core/Ids.java
```diff
@@ -180,4 +180,12 @@
 
    int READ_ONLY_SNAPSHOT_VIEW = 169;
    int NO_VALUE_READ_ONLY_VIEW = 170;
+
+   int MIME_CACHE_ENTRY = 171;
+   int UUID = 172;
+
+   // 173-174 used in commons
+   int QUEUE = 175;
```
pruivo: unused. can be removed.

pruivo:
@galderz, maybe you can squash the commits in the end?

galderz:
@pruivo Fixed and squashed

pruivo:
@galderz I'm going to wait for the CI to run , but you have one related test failing:
```
JBossMarshallerTest.testInternalDuplicateExternalizerId
```

galderz:
Checking...

galderz:
Fixed and pushed.

pruivo:
I triggered another run in CI to see if it works with JGroups 4.0.0.Beta1 :)

galderz:
@pruivo It looks good, can we integrate this?

pruivo:
integrated! thanks @galderz !

PR #4621, opened 2016-10-18
Default XSD/doc values are now extracted from source.

ryanemerson:
This consists of a java program `DefaultAttributeExtractor.java`, which outputs the default values for configuration resources found on the classpath to either a propteries file or asciidoc attributes file. The process of extracting default key/values combos are provided in the specified implementation of the DefaultsResolver interface. 

Values extracted to a defaults.proprties file, on relevant modules, are then used to apply a maven-resource filter to *.xsd files where ${...} placeholders are replaced with the relevant default values.  

For asciidoctor integration, the default values are exported to a .asciidoctor file where they are defined as attributes. To use these values you simply include the defaults.asciidoctor file and call the attribute where required, e.g. for the simple cache default can be called via `{Configuration.simpleCache}`.

@tristantarrant I have updated all of the relevant *.xsd files (AFAIK) to use default placeholders, however the asciidoc files remain the same as many of the defaults are referenced in sentences etc and need more attention.

### DIFF
--- a/persistence/soft-index/pom.xml
+++ b/persistence/soft-index/pom.xml
@@ -15,10 +15,6 @@
 
    <dependencies>
       <dependency>
-         <groupId>${project.groupId}</groupId>
-         <artifactId>infinispan-core</artifactId>
-      </dependency>
-      <dependency>
          <groupId>org.kohsuke.metainf-services</groupId>
          <artifactId>metainf-services</artifactId>
          <optional>true</optional>
@@ -80,6 +76,18 @@
                   </Include-Resource>
                </instructions>
             </configuration>
+         </plugin>
+         <plugin>
+            <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-defaults-maven-plugin</artifactId>
+            <executions>
+               <execution>
+                  <id>extract-defaults</id>
+                  <goals>
+                     <goal>extract-defaults</goal>
+                  </goals>
+               </execution>
+            </executions>
          </plugin>
       </plugins>
    </build>
--- a/plugins/infinispan-defaults-maven-plugin/pom.xml
+++ b/plugins/infinispan-defaults-maven-plugin/pom.xml
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.infinispan</groupId>
+        <artifactId>infinispan-parent</artifactId>
+        <version>9.0.0-SNAPSHOT</version>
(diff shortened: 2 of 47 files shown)


galderz:
Needs rebasing

persistence/soft-index/pom.xml
```diff
@@ -81,6 +81,58 @@
...
+            </configuration>
+            <executions>
+               <execution>
+                  <id>defaults</id>
+                  <phase>process-classes</phase>
+                  <goals>
+                     <goal>java</goal>
+                  </goals>
+                  <configuration>
+                     <mainClass>org.infinispan.commons.configuration.defaults.DefaultAttributeExtractor</mainClass>
```
slaskawi: Maybe we should create a Maven plugin for it? This parts is being copied all over the place.

commons/src/main/java/org/infinispan/commons/configuration/defaults/InfinispanDefaultsResolver.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.infinispan.commons.configuration.attributes.Attribute;
+import org.infinispan.commons.configuration.attributes.AttributeDefinition;
+import org.infinispan.commons.configuration.attributes.AttributeSet;
+
+/**
+ * @author Ryan Emerson
```
slaskawi: Could you please remove those empty javadocs? Or even better - put something meaningful here?

persistence/soft-index/pom.xml
ryanemerson: @slaskawi Yeah I did think that a Maven plugin would maybe better once I started pasting the same stuff into the pom. I've never developed a plugin before, so I'll look into what that entails etc

ryanemerson:
@slaskawi As suggested I have refactored the PR as a maven plugin, which has reduced the verbosity of the pom's somewhat, and I have improved the javadocs.

plugins/infinispan-defaults-maven-plugin/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-plugin-plugin</artifactId>
+                <version>3.5</version>
+                <configuration>
+                    <skipErrorNoDescriptorsFound>true</skipErrorNoDescriptorsFound>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.infinispan</groupId>
```
slaskawi: This is probably a mistake - the plugin uses itself?

ryanemerson: @slaskawi Good spot :) I've pushed the changes

galderz:
Anything else to do here?

ryanemerson:
@galderz Not as far as I'm concerned

slaskawi:
Hey @ryanemerson Looks good to me. Could you please rebase it. Once you and TeamCity are done on this PR, I'll integrate it.

ryanemerson:
@slaskawi rebased, thanks!

slaskawi:
Thanks @ryanemerson !

PR #4623, opened 2016-10-19
ISPN-7126 InfinispanIndexManager hangs under load

gustavonalle:
https://issues.jboss.org/browse/ISPN-7126

Reproducer only. 
This test hangs on master and eventually times out, but works as expected before https://github.com/infinispan/infinispan/commit/ad0b1cc84778074769adf9312042fad74169bf2d

### DIFF
--- a/core/src/test/java/org/infinispan/test/fwk/TestCacheManagerFactory.java
+++ b/core/src/test/java/org/infinispan/test/fwk/TestCacheManagerFactory.java
@@ -343,7 +343,7 @@ public class TestCacheManagerFactory {
       amendTransport(cfg, new TransportFlags());
    }
 
-   private static void amendTransport(GlobalConfigurationBuilder builder, TransportFlags flags) {
+   public static void amendTransport(GlobalConfigurationBuilder builder, TransportFlags flags) {
       org.infinispan.configuration.global.GlobalConfiguration gc = builder.build();
       if (gc.transport().transport() != null) { //this is local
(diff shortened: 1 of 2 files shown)


danberindei:
@gustavonalle Ok, looks like the problem is that the `QueryInterceptor.visitPutKeyValueCommand` `thenAccept` lambda does a blocking `Cache.putIfAbsent` (indirectly). 

Before ISPN-5469, this would block a remote-executor thread. After ISPN-5469, this would block the OOB thread that received the response from the primary owner. That means we need less remote-executor threads, but we need more OOB threads. I only tested with 10 threads, but that got rid of the deadlock for me.

Bela and Pedro have been discussing merging the JGroups (ex-OOB) and remote-executor thread pools with JGroups 4.0 anyway, so I think just increasing the OOB thread pool size is acceptable for now. Ideally though, indexing would also be asynchronous and return a `CompletableFuture`.

gustavonalle:
@danberindei Thanks. I tried it yesterday and it solved for me as well. I was in doubt if that was the expected behaviour after 5469 and a bit reluctant to increase the OOB thread pool, but will do that for now.

gustavonalle:
@danberindei wrt index being asynchronous, I will see if the putIfAbsent during lucene lock acquiring can be asynchronous, this would avoid blocking in the OOB altogether (will block somewhere else)

danberindei:
@gustavonalle One step at a time :)

PR #4624, opened 2016-10-19
ISPN-7128 JMX attribute configurationAsProperties cannot be read on s…

jmarkos:
…ome caches

This fix is definitely hacky, but at least fast to come up with.
Other option would be to fix the problematic InterceptorConfiguration.interceptor method, but I don't know if we want to spend time fixing deprecated methods...


tristantarrant:
Pushed !

PR #4626, opened 2016-10-20
ISPN-6903 (2/3) New marshaller (handled feedback)

galderz:
Once #4620 has been integrated we can remove some noise from this PR.

This PR focuses on simplifying the ideas introduced in #4530 and handle the feedback provided. To be more precise, this is what has been done:
- Make `ClientEventsOOMTest` use replicated cache.
- Merge BytesEncoding logic BytesObject\* classes.
- Get rid of `InternalMarshaller` class and move code to `GlobalMarshaller`.
- Rename from `checkCapacity` to `ensureCapacity` in `BytesObjectOutput`.
- Return -1 (instead of throwing `EOFException`) in `JBossByteInput.read()` implementation when it reads to far from the backing array.
- `ClusterEvent.Externalizer` now uses MarshallUtil.marshallEnum/unmarshallEnum instead.
- `TransactionInfo` now calls `MarshallUtil.marshallCollection()`
-  Move internal/\* classes to marshaller/core and use package private access for better encapsulation.
- Separate external and internal externalizer tables.
- Provided a dedicated, simplified, implementation of Class to AdvancedExternalizer lookup map.
- Class to AdvancedExternalizer map has the capability to produce a reverse map which maps from AdvancedExternalizer ID to AdvancedExternalizer (for reads).
- Reverse map implementation varies depending on whether max AdvancedExternalizer ID is known (e.g. internal externalizers) or not (e.g. external externalizers). The former case is optimised to be a simple array implementation, whereas the latter hashes IDs to deal with potential large IDs since we can't make assumptions.
- Deal with null in GlobalMarshaller.
- Deal with primitives independently (e.g. not via Externalizer) so that you don't have to do weird things in the JBoss Marshaller external marshaller.
- Dealing with primitives independently reduces number of look ups from 2 to 1.
- External marshaller is now configurable.
- Remove ExternalizerTable.
- SMALL and MEDIUM constants used in `Primitives` class to deal with feedback [here](https://github.com/infinispan/infinispan/pull/4530#r78183198).
- `writeObjectArray` and similar methods now writes `len - 1` instead, so that you don't need a branch when creating new primitive arrays.
- Rewinding in ExternalJBossMarshaller now achieved without the need of extra interfaces and using information that can be extracted out of `RiverUnmarshaller`.
- Bunch up externalizer IDs, remove unused IDs and removed unused externalizers to help with better sizing decisions of Class to Externalizer map.


galderz:
Of course, once everyone's happy with this, I'll squash commits involving new marshaller.

galderz:
Needs rebasing, I'll sort that out today

galderz:
Replaced by #4648

PR #4630, opened 2016-10-24
ISPN-7127 Broken InboundTransferTask may lose segments on restart

danberindei:
https://issues.jboss.org/browse/ISPN-7127
- Only retry an inbound transfer when we receive a new cache topology.
- Move the segment list to the end of log messages

### DIFF
--- a/core/src/main/java/org/infinispan/statetransfer/InboundTransferTask.java
+++ b/core/src/main/java/org/infinispan/statetransfer/InboundTransferTask.java
@@ -118,7 +118,7 @@ public class InboundTransferTask {
             return completionFuture;
          }
          if (trace) {
-            log.tracef("Requesting segments %s of cache %s from node %s", segmentsCopy, cacheName, source);
+            log.tracef("Requesting state from node %s for segments %s", source, segmentsCopy);
          }
          // start transfer of cache entries
          try {
@@ -127,17 +127,17 @@ public class InboundTransferTask {
             Response response = responses.get(source);
             if (response instanceof SuccessfulResponse) {
                if (trace) {
-                  log.tracef("Successfully requested segments %s of cache %s from node %s", segmentsCopy, cacheName, source);
+                  log.tracef("Successfully requested state from node %s for segments %s", source, segmentsCopy);
                }
                return completionFuture;
             } else {
                Exception e = response instanceof ExceptionResponse ?
                      ((ExceptionResponse) response).getException() : new CacheException(String.valueOf(response));
-               log.failedToRequestSegments(segmentsCopy, cacheName, source, e);
+               log.failedToRequestSegments(cacheName, source, segmentsCopy, e);
                completionFuture.completeExceptionally(e);
             }
          } catch (Exception e) {
-            log.failedToRequestSegments(segmentsCopy, cacheName, source, e);
+            log.failedToRequestSegments(cacheName, source, segmentsCopy, e);
             completionFuture.completeExceptionally(e);
          }
       }
@@ -155,7 +155,7 @@ public class InboundTransferTask {
       }
 
       if (trace) {
-         log.tracef("Cancelling inbound state transfer of segments %s of cache %s", cancelledSegments, cacheName);
+         log.tracef("Partially cancelling inbound state transfer from node %s, segments %s", source, cancelledSegments);
       }
 
       synchronized (segments) {
@@ -184,7 +184,7 @@ public class InboundTransferTask {
 
          Set<Integer> segmentsCopy = getSegments();
          if (trace) {
-            log.tracef("Cancelling inbound state transfer of segments %s of cache %s", segmentsCopy, cacheName);
+            log.tracef("Cancelling inbound state transfer from %s with segments %s", source, segmentsCopy);
          }
 
          sendCancelCommand(segmentsCopy);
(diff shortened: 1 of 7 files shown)


rvansa:
No test case? :-/

danberindei:
@rvansa Unfortunately I haven't been able to figure out how the random failures in `StateTransferRestartTest` actually happened, and my attempts to add strategic delays in `StateConsumerImpl` didn't make the failures easier to reproduce. I also looked at your `ConcurrentOverlappingLeaveTest` log, but I didn't see any failure there...

rvansa:
@danberindei IIRC in the ConcurrentOverlappingLeaveTest a segment was not transfered to a node since the transfer has been canceled (and new one did not start). But fair enough, let's hope that this PR actually fixes something... I hope I'll find some time to check the test log again, add more logging and catch it in pyjamas...

danberindei:
@rvansa I had added a new test while working on a more involved solution, but I decided not to add it to the PR because it didn't fail often enough. Weird enough, I tried it a few times on master today and it failed every time, so I've added it.

core/src/test/java/org/infinispan/statetransfer/StateTransferRestart2Test.java
```diff
@@ -0,0 +1,154 @@
...
+ * @author Dan Berindei
+ * @since 5.2
+ */
+@Test(groups = "functional", testName = "statetransfer.StateTransferRestartTest")
+@CleanupAfterMethod
+public class StateTransferRestart2Test extends MultipleCacheManagersTest {
+
+   private ConfigurationBuilder cfgBuilder;
+
+   private void waitForStateTransfer(Cache... caches) throws InterruptedException {
```
slaskawi: Would it be possible to use `org.infinispan.test.TestingUtil#waitForRehashToComplete(org.infinispan.Cache...)` here?

core/src/test/java/org/infinispan/statetransfer/StateTransferRestart2Test.java
```diff
@@ -0,0 +1,154 @@
...
+      log.info("get c2");
+      final Cache<Object, Object> c2 = cache(2);
+
+      log.info("waiting for cluster { c0, c2 }");
+      TestingUtil.blockUntilViewsChanged(10000, 2, c0, c2);
+
+      log.infof("c0 entrySet size before : %d", c0.entrySet().size());
+      log.infof("c2 entrySet size before : %d", c2.entrySet().size());
+
+      eventually(new Condition() {
```
slaskawi: This could be a lambda. It would be shorter.

core/src/main/java/org/infinispan/statetransfer/InboundTransferTask.java
```diff
@@ -118,7 +118,7 @@ public Address getSource() {
             return completionFuture;
          }
          if (trace) {
-            log.tracef("Requesting segments %s of cache %s from node %s", segmentsCopy, cacheName, source);
+            log.tracef("Requesting state from node %s for segments %s", source, segmentsCopy);
```
wburns: Sorry I might be missing something, but why take out the cacheName?  Seems we would lose some information in the log messages.  Or is it just information overload? :)

core/src/main/java/org/infinispan/statetransfer/OutboundTransferTask.java
```diff
@@ -182,13 +182,13 @@ public void processEntry(MarshalledEntry me, AdvancedCacheLoader.TaskContext tas
       } catch (Throwable t) {
          // ignore eventual exceptions caused by cancellation (have InterruptedException as the root cause)
          if (isCancelled()) {
-            log.debugf("Transfer of segments %s of cache %s to node %s cancelled", segments, cacheName, destination);
+            log.tracef("Ignoring error in already cancelled transfer to node %s, segments %s", destination, segments);
```
wburns: Looks like we want a `if (trace)` around this?

core/src/test/java/org/infinispan/statetransfer/StateTransferRestart2Test.java
```diff
@@ -0,0 +1,154 @@
...
+         for (StateTransferManager aStm : stm) {
+            if (aStm.isStateTransferInProgress()) {
+               inProgress = true;
+               break;
+            }
+         }
+         if (!inProgress) {
+            break;
+         }
+         wait(100);
```
wburns: It doesn't seem like we are holding the lock on `this` when calling wait, which is required as part of the spec.

core/src/test/java/org/infinispan/statetransfer/StateTransferRestart2Test.java
danberindei: I think so, I copied it from `StateTrasnsferRestartTest` without paying too much attention.

core/src/main/java/org/infinispan/statetransfer/InboundTransferTask.java
danberindei: I think the cache name was already in the NDC. Also, none of the other messages contain the cache name.

core/src/test/java/org/infinispan/statetransfer/StateTransferRestart2Test.java
danberindei: I guess it never actually got to this point :) I'm removing this method and using `TestingUtil.waitForRehashToComplete()` instead.

galderz:
No @danberindei replies since October to @wburns' and @slaskawi's comments. Closing for now. @danberindei reopen when you have time to reply to them.

danberindei:
Oops, can't reopen after you force-push...

PR #4631, opened 2016-10-25
ISPN-7129 Duplicate cache names shouldn't be allowed

jmarkos:
https://issues.jboss.org/browse/ISPN-7129

Removed the ParserOverrideTest after discussion with @galderz, since we have templates now.


danberindei:
@jmarkos @galderz I think as long as the functionality of extending the default cache still works, `ParserOverrideTest` can stay. Not sure how the template overrides are tested anyway...

I'd also like to see a test that actually has a duplicate cache name and throws the exception.

jmarkos:
@danberindei the test cannot stay if my changes are to be accepted, because it fails with them (as it has multiple caches with the same name). So if we want to continue allowing overriding a cache configuration with another cache with the same name (which I think is more often a mistake on the users part than an intention), this PR and the jira can be closed.

danberindei:
@jmarkos Sorry, I didn't mean it should stay exactly the same, although TBH I was hoping some of it was dedicated to testing how the inheritance of configuration from the default cache to the named caches used to work. I see that's not the case, and in fact it doesn't work any more :)

I think there would still be some value in updating this test to check extending a template works and that extending another cache doesn't, unless there's another test for that already. WDYT?

jmarkos:
Named caches still inherit the configuration of the default cache, with this configuration
```
    <cache-container name="c1" default-cache="mydefault">
        <local-cache name="mydefault">
            <eviction max-entries="100" />
        </local-cache>
        <local-cache name="inherit"/>
    </cache-container>
```
the 'inherit' cache will also have max-entries=100.

This
```
        <local-cache name="mydefault">
            <eviction max-entries="100" />
        </local-cache>
        <local-cache name="inherit" configuration="mydefault"/>
```
throws "Exception in thread "main" org.infinispan.commons.CacheConfigurationException: ISPN000375: Cannot use configuration 'mydefault' as a template" but I haven't found a test for it.

The only test for the templates I found (in core) is this one https://github.com/infinispan/infinispan/blob/master/core/src/test/java/org/infinispan/configuration/parsing/UnifiedXmlFileParsingTest.java.

I can add a test that verifies you cannot define duplicate cache names, would that be enough?

rvansa:
So what's the current/desired status? I think that cache `xyz` should not inherit anything from the default cache, as this is rather cryptic; if user wants default cache and custom cache inheriting similar configuration, he should use
```
<cache-container name="c1" default-cache="mydefault">
    <local-cache-template name="common"><!-- whatever --></local-cache-template>
    <local-cache name="mydefault" configuration="common"/>
    <local-cache name="inherit" configuration="common"> <!-- add something --></local-cache>
</cache-container>
```

jmarkos:
I think inheriting from the default cache was a concept that was useful before templates, now it seems like a relict.

galderz:
I think default cache inheritance should go. 

@jmarkos, since there might be different opinions, it might be good if a dev thread is started to finalise the decision?

jmarkos:
Rebased and added a test (to XmlFileParsingTest, not sure if that's the right place).

galderz:
Needs rebasing. Also, this PR depends on #4744.

galderz:
Still needs rebasing. Closing for now. @jmarkos please reopen when you've got around to sorting this out. Thanks for the contribution!

jmarkos:
I was waiting for https://github.com/infinispan/infinispan/pull/4744 to get merged, since it looked like there could be conflicts. I'll look at this asap.

PR #4633, opened 2016-10-26
ISPN-7141 LimitedExecutorTest.testConcurrencyLimit random failures

danberindei:
https://issues.jboss.org/browse/ISPN-7141

### DIFF
--- a/core/src/test/java/org/infinispan/executors/LimitedExecutorTest.java
+++ b/core/src/test/java/org/infinispan/executors/LimitedExecutorTest.java
@@ -1,18 +1,15 @@
 package org.infinispan.executors;
 
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertFalse;
 
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
 
 import org.infinispan.test.AbstractInfinispanTest;
-import org.infinispan.test.TestException;
 import org.infinispan.util.concurrent.WithinThreadExecutor;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.Test;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/infinispan/executors/LimitedExecutorTest.java
```diff
@@ -57,11 +57,13 @@ public void testConcurrencyLimit() throws Exception {
       limitedExecutor.execute(() -> cf2.complete("bla"));
 
       Thread.sleep(10);
+      assertFalse(cf1.isDone());
       assertFalse(cf2.isDone());
 
       latch.countDown();
       assertEquals("bla", cf1.get(10, SECONDS));
       assertEquals("bla", cf2.get(10, SECONDS));
+      eventuallyEquals(0, executor::getActiveCount);
```
rvansa: Do you still need the two threads with this `eventuallyEquals`? I would rather move that on the beginning of the test method, as this line may not be reached upon test failure (and you don't want to fail the next test method).

danberindei: Yes, I realized I actually needed 2 threads for `testConcurrencyLimit()`.
But I added another `eventuallyEquals()` call at the beginning of the test methods that use the executor.

rvansa: Okay, so the 2 threads are there to allow the possibility that the tasks will run in parallel, right? But what do you actually test here? The test confirms that eventually both tasks will be executed, but there's nothing testing the level of parallelism.

danberindei: I've added javadocs to the test methods. `testConcurrencyLimit()` uses regular `execute()`, so it needs 2 threads to check that one task really blocks the other. This test method doesn't need 2 threads, because it uses `executeAsync()`, and the 1st task blocks the executor without using any threads.

rvansa: I would still add a check into the second task verifying that the first task has finished; otherwise you could replace LimitedExecutor with some executor that starts executing the second task a bit later (than the 10 ms) and it would still succeed.

danberindei: I guess having written `LimitedExecutor` myself, I didn't think it could ever delay a task without a good reason :)
Anyway, I've updated the test to better deal with the possibility.

rvansa:
The tests would work with single `blocker` as well, but I think this is good to be integrated as-is.

wburns:
Pulling..

wburns:

PR #4634, opened 2016-10-26
ISPN-7142 Query DSL 'like' operator should consider \ as escape chara…

anistor:
…cter

* The escape char is considered to be '\\' by default and you cannot specify another escape char for now.
* rename RegexCondition to LikeCondition

Jira: https://issues.jboss.org/browse/ISPN-7142


gustavonalle:
merged, thanks @anistor !

PR #4638, opened 2016-10-27
ISPN-7155 Failures in CI from functional tests

rvansa:
Maybe this could be the difference!


danberindei:
@rvansa Yep, this seems to do the trick! I'm pretty sure I'd have integrated it by now if it was in a PR by itself ;)

rvansa:
@danberindei Updated with comments.

danberindei:
Integrated, thanks Radim!

PR #4641, opened 2016-10-31
ISPN-7154 Add test to assert RPCs for the AffinityIndexManager

gustavonalle:
https://issues.jboss.org/browse/ISPN-7154

### DIFF
--- a/query/src/test/java/org/infinispan/query/affinity/RpcCollector.java
+++ b/query/src/test/java/org/infinispan/query/affinity/RpcCollector.java
@@ -1,0 +1,40 @@
+package org.infinispan.query.affinity;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import org.infinispan.commands.remote.BaseRpcCommand;
+import org.infinispan.statetransfer.StateRequestCommand;
+import org.infinispan.statetransfer.StateResponseCommand;
+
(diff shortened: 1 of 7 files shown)


query/src/test/java/org/infinispan/query/affinity/RpcCollector.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.infinispan.statetransfer.StateRequestCommand;
+import org.infinispan.statetransfer.StateResponseCommand;
+
+class RpcCollector {
+
+   private List<RpcDetail> rpcCollection = new ArrayList<>();
+
+   private Predicate<RpcDetail> stateTransferPredicate = rpc ->
+         Arrays.asList(StateRequestCommand.class, StateResponseCommand.class).contains(rpc.getCommand().getClass());
```
anistor: Doesn't this build a new list on each invocation? While this looks very functionalicious and escape analysis can save the day I would still think of writing this as plain old ```rpc.getCommand().getClass() == X.class or rpc.getCommand().getClass() == Y.class```.

anistor:
I raised a single potential issue.  Otherwise, LGTM.

query/src/test/java/org/infinispan/query/affinity/RpcCollector.java
gustavonalle: functionalicious? :smile:

It will allocate a list every time the predicate is evaluated, let me sort it out

gustavonalle:
Updated

anistor:

PR #4642, opened 2016-11-02
ISPN-7161 RemoteGetFailureTest.testOneOwnerSuspectedNoFilter failures

danberindei:
https://issues.jboss.org/browse/ISPN-7161

* Add delay after view installation
* Change asserts to include the response in the failure messages

### DIFF
--- a/core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
+++ b/core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
@@ -3,7 +3,6 @@ package org.infinispan.distribution;
 import static org.infinispan.test.Exceptions.expectException;
 import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertFalse;
-import static org.testng.AssertJUnit.assertNotNull;
 import static org.testng.AssertJUnit.assertTrue;
 
 import java.lang.reflect.Field;
@@ -25,6 +24,7 @@ import org.infinispan.commons.CacheException;
 import org.infinispan.commons.equivalence.AnyEquivalence;
 import org.infinispan.configuration.cache.CacheMode;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.container.entries.ImmortalCacheValue;
 import org.infinispan.context.InvocationContext;
 import org.infinispan.interceptors.BasicInvocationStage;
 import org.infinispan.interceptors.DDAsyncInterceptor;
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
```diff
@@ -198,30 +200,34 @@ public void testOneOwnerSuspectedNoFilter(Method m) throws ExecutionException, I
...
-      CompletableFuture<Map<org.infinispan.remoting.transport.Address, Response>> future = cache(0).getAdvancedCache().getRpcManager().invokeRemotelyAsync(owners, clusteredGet, rpcOptions);
+      RpcManager rpcManager = cache(0).getAdvancedCache().getRpcManager();
+      CompletableFuture<Map<Address, Response>> future = rpcManager.invokeRemotelyAsync(owners, clusteredGet, rpcOptions);
 
       assertTrue(arrival.await(10, TimeUnit.SECONDS));
 
       installNewView(cache(0), cache(0), cache(1));
 
+      // RequestCorrelator processes the view asynchronously, so we need to wait a bit for node 2 to be suspected
+      Thread.sleep(100);
```
rvansa: Can't we add some listener and countdown here?

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
```diff
@@ -198,30 +200,34 @@ public void testOneOwnerSuspectedNoFilter(Method m) throws ExecutionException, I
...
       release1.countDown();
-      Map<org.infinispan.remoting.transport.Address, Response> responses = future.get();
+      Map<Address, Response> responses = future.get();
       long requestCompleted = System.nanoTime();
-      assertTrue(responses.get(address1).isSuccessful());
-      assertNotNull(((SuccessfulResponse) responses.get(address1)).getResponseValue());
-      assertTrue(responses.get(address2) instanceof CacheNotFoundResponse);
-      assertTrue(TimeUnit.NANOSECONDS.toSeconds(requestCompleted - requestAllowed) < timeout);
+
+      assertEquals(0, TimeUnit.NANOSECONDS.toSeconds(requestCompleted - requestAllowed), timeout);
```
rvansa: Is this assertEquals(double expected, double actual, double delta)? A bit confusing. Then, the comparison there is `Math.abs(expected - actual) <= delta` while we want strict comparison: `actual < delta` (`expected` is 0)

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
danberindei: I thought about subclassing `CommandAwareRpcDispatcher` and `CustomRequestCorrelator` to make `receiveView` synchronous, but I decided that it wasn't worth the effort.

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
danberindei: This seemed the simplest way to log the actual duration after a failure, but I agree it's a bit confusing.

slaskawi:
Integrated thanks!

PR #4645, opened 2016-11-08
ISPN-7175 Do not require a trust store for the HotRod server

tristantarrant:
https://issues.jboss.org/browse/ISPN-7175

- enhance the ssl test to try more combinations with client certs
- enhance the security documentation for server/clients


galderz:
Looks good, integrated.

PR #4648, opened 2016-11-09
ISPN-6906 (2/3) New marshaller

galderz:
Once #4620 has been integrated we can remove some noise from this PR.

This PR focuses on simplifying the ideas introduced in #4530 and handle the feedback provided. To be more precise, this is what has been done:
- Make `ClientEventsOOMTest` use replicated cache.
- Merge BytesEncoding logic BytesObject\* classes.
- Get rid of `InternalMarshaller` class and move code to `GlobalMarshaller`.
- Rename from `checkCapacity` to `ensureCapacity` in `BytesObjectOutput`.
- Return -1 (instead of throwing `EOFException`) in `JBossByteInput.read()` implementation when it reads to far from the backing array.
- `ClusterEvent.Externalizer` now uses MarshallUtil.marshallEnum/unmarshallEnum instead.
- `TransactionInfo` now calls `MarshallUtil.marshallCollection()`
-  Move internal/\* classes to marshaller/core and use package private access for better encapsulation.
- Separate external and internal externalizer tables.
- Provided a dedicated, simplified, implementation of Class to AdvancedExternalizer lookup map.
- Class to AdvancedExternalizer map has the capability to produce a reverse map which maps from AdvancedExternalizer ID to AdvancedExternalizer (for reads).
- Reverse map implementation varies depending on whether max AdvancedExternalizer ID is known (e.g. internal externalizers) or not (e.g. external externalizers). The former case is optimised to be a simple array implementation, whereas the latter hashes IDs to deal with potential large IDs since we can't make assumptions.
- Deal with null in GlobalMarshaller.
- Deal with primitives independently (e.g. not via Externalizer) so that you don't have to do weird things in the JBoss Marshaller external marshaller.
- Dealing with primitives independently reduces number of look ups from 2 to 1.
- External marshaller is now configurable.
- Remove ExternalizerTable.
- SMALL and MEDIUM constants used in `Primitives` class to deal with feedback [here](https://github.com/infinispan/infinispan/pull/4530#r78183198).
- `writeObjectArray` and similar methods now writes `len - 1` instead, so that you don't need a branch when creating new primitive arrays.
- Rewinding in ExternalJBossMarshaller now achieved without the need of extra interfaces and using information that can be extracted out of `RiverUnmarshaller`.
- Bunch up externalizer IDs, remove unused IDs and removed unused externalizers to help with better sizing decisions of Class to Externalizer map.


galderz:
This has been rebased with latest master and has PR 1/3 in it. Ready for review.

pruivo:
@galderz something lost in the rebase. I have a lot of tests failing with:
```
Caused by: java.lang.AssertionError: Check support for: class org.infinispan.util.ControlledConsistentHashFactory
	at org.infinispan.marshall.core.ExternalJBossMarshaller.objectToObjectStream(ExternalJBossMarshaller.java:29)
	at org.infinispan.marshall.core.GlobalMarshaller.writeUnknown(GlobalMarshaller.java:379)
```

galderz:
@pruivo Should be fixed now

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
```diff
@@ -31,124 +39,413 @@
...
  */
 @Scope(Scopes.GLOBAL)
-public class GlobalMarshaller extends AbstractMarshaller implements StreamingMarshaller {
+public class GlobalMarshaller implements StreamingMarshaller {
 
    private static final Log log = LogFactory.getLog(GlobalMarshaller.class);
    private final boolean trace = log.isTraceEnabled();
 
-   private JBossMarshaller defaultMarshaller;
+   final MarshallableTypeHints marshallableTypeHints = new MarshallableTypeHints();
```
ryanemerson: Should we make this private?

core/src/main/java/org/infinispan/marshall/core/ExternalJBossMarshaller.java
```diff
@@ -0,0 +1,180 @@
...
+import org.infinispan.commons.marshall.MarshallableTypeHints;
+import org.infinispan.commons.marshall.StreamingMarshaller;
+import org.infinispan.configuration.global.GlobalConfiguration;
+import org.jboss.marshalling.ByteInput;
+import org.jboss.marshalling.ByteOutput;
+
+final class ExternalJBossMarshaller implements StreamingMarshaller {
+
+   final MarshallableTypeHints marshallableTypeHints = new MarshallableTypeHints();
+   final JBossMarshaller marshaller;
```
ryanemerson: Should we make these private?

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
```diff
@@ -31,124 +39,413 @@
...
+      if (internalExt != null)
+         return (out, object) -> {
+            writeInternalClean(object, internalExt, out);
+         };
+
+
+      AdvancedExternalizer externalExt = externalExts.get(clazz);
+      if (externalExt != null)
+         return (out, object) -> {
+            writeExternalClean(object, externalExt, out);
```
ryanemerson: Can be reduced to a single line lambda

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
```diff
@@ -31,124 +39,413 @@
...
+   }
+
+   static final int ID_NULL                        = 0x00;
+   static final int ID_PRIMITIVE                   = 0x01;
+   static final int ID_INTERNAL                    = 0x02;
+   static final int ID_EXTERNAL                    = 0x03;
+   static final int ID_ANNOTATED                   = 0x04;
+   static final int ID_UNKNOWN                     = 0x05;
+
+   static final int NOT_FOUND                      = -1;
```
ryanemerson: Minor, but would it be better to declare these at the top of the class so that they are more visible?

core/src/main/java/org/infinispan/marshall/core/Primitives.java
```diff
@@ -0,0 +1,638 @@
...
+   }
+
+   private static byte[] readFully(byte[] arr, BytesObjectInput in) {
+      in.readFully(arr);
+      return arr;
+   }
+
+   private static boolean[] readBooleanArray(BytesObjectInput in) throws IOException {
+      byte type = in.readByte();
+      int len;
```
ryanemerson: Unused variable

core/src/main/java/org/infinispan/marshall/exts/CacheRpcCommandExternalizer.java
```diff
@@ -105,14 +105,7 @@ public void writeObject(ObjectOutput output, CacheRpcCommand command) throws IOE
...
    private void marshallParameters(CacheRpcCommand cmd, StreamingMarshaller marshaller, ObjectOutput oo) throws IOException {
-      BufferSizePredictor sizePredictor = marshaller.getBufferSizePredictor(cmd);
-      int estimatedSize = sizePredictor.nextSize(cmd);
-      ObjectOutput paramsOutput = marshaller.startObjectOutput(convertObjectOutput(oo), true, estimatedSize);
-      try {
-         cmdExt.writeCommandParameters(paramsOutput, cmd);
-      } finally {
-         marshaller.finishObjectOutput(paramsOutput);
-      }
+      cmdExt.writeCommandParameters(oo, cmd);
```
ryanemerson: Unused imports for BufferSizePredictor, DelegationObjectInput and Inputstream

core/src/main/java/org/infinispan/marshall/core/ExternalExternalizers.java
```diff
@@ -0,0 +1,97 @@
...
+final class ExternalExternalizers {
+
+   private static final Log log = LogFactory.getLog(ExternalExternalizers.class);
+
+   private ExternalExternalizers() {
+   }
+
+   static ClassToExternalizerMap load(GlobalConfiguration globalCfg) {
+      // TODO Add initial value and load factor
+      ClassToExternalizerMap exts = new ClassToExternalizerMap(4, 0x0.6p0f);
```
pruivo: don't you wanna put the load factor in a more readable way? :)

core/src/main/java/org/infinispan/marshall/core/ExternalJBossMarshaller.java
```diff
@@ -0,0 +1,165 @@
...
+import org.infinispan.commons.io.ByteBuffer;
+import org.infinispan.commons.marshall.BufferSizePredictor;
+import org.infinispan.commons.marshall.MarshallableTypeHints;
+import org.infinispan.commons.marshall.StreamingMarshaller;
+import org.infinispan.commons.marshall.jboss.ExtendedRiverUnmarshaller;
+import org.infinispan.configuration.global.GlobalConfiguration;
+import org.jboss.marshalling.ByteInput;
+import org.jboss.marshalling.ByteOutput;
+
+final class ExternalJBossMarshaller implements StreamingMarshaller {
```
pruivo: what about creating a new interface for the external marshaller? most of the method of ```StreamingMarshaller``` are not used.

core/src/main/java/org/infinispan/marshall/core/Ids.java
```diff
@@ -9,178 +9,7 @@
...
-
-   int READ_ONLY_SNAPSHOT_VIEW = 169;
-   int NO_VALUE_READ_ONLY_VIEW = 170;
-
-   int MIME_CACHE_ENTRY = 171;
-   int UUID = 172;
-
-   // 173-174 used in commons
-   int ARRAYS = 175;
+   // All identifiers now defined in commons for a more complete view.
```
pruivo: why not remove the class then? :)

pruivo:
I'm wondering if we can take this change to update the ```StreamingMarshaller``` and ```Marshaller``` interfaces. They have a bunch of methods that doesn't seem to fit in our current uses cases.

galderz:
> I'm wondering if we can take this change to update the StreamingMarshaller and Marshaller interfaces. They have a bunch of methods that doesn't seem to fit in our current uses cases.

^ That's a huge impact since we expose these classes many places, even to the users, since they're pluggable. So, -1 from me, let's stick to the original plan. API changes can always come later, with the implementation doing what we want it to do.

core/src/main/java/org/infinispan/marshall/core/BytesObjectInput.java
```diff
@@ -0,0 +1,250 @@
...
+import java.io.ObjectInput;
+
+final class BytesObjectInput implements ObjectInput, PositionalBuffer.Input {
+
+   final byte bytes[];
+
+   final GlobalMarshaller marshaller;
+
+   int pos;
+   int offset; // needed for external JBoss Marshalling, to be able to rewind correctly when the bytes are prepended :(
```
galderz: This class is package default and it's `final`, so `protected` adds nothing here. If anything `private` but I didn't feel the need for it.

core/src/main/java/org/infinispan/marshall/core/ExternalJBossMarshaller.java
galderz: Hmmm, not sure it adds much. It's another package private, final class.

core/src/main/java/org/infinispan/marshall/core/ClassToExternalizerMap.java
```diff
@@ -0,0 +1,447 @@
...
+         if (k == null) {
+            // not found
+            return null;
+         }
+         hc = (hc + 1) & mask;
+      }
+   }
+
+   /**
+    * Put a value into the map.  Any previous mapping is discarded silently.
```
galderz: Hmmm, could be. I've separate this to [ISPN-7210](https://issues.jboss.org/browse/ISPN-7210).

core/src/main/java/org/infinispan/marshall/core/ExternalJBossMarshaller.java
galderz: Not in the scope of this. In fact, external marshaller as concept is likely to go away and be replaced by a user marshaller (see marshalling doc).

core/src/main/java/org/infinispan/marshall/core/Ids.java
galderz: It would created a lot of noise in this PR, since you'd have to change all references to this class. Can be done at a later stage when cleaning things up. I can probably do it in PR 3/3 where I'm adding ExternalPojo marker to mark classes that are expected to be externally marshallable (see Dan's comment in previous PR).

galderz:
I've added a commit with minor fixes based on your reviews. 

Unless I've commented, I've applied the changes as suggested.

pruivo:
just some minor comments/cleanups:
unused imports in CacheRpcCommandExternalizer:
* import java.io.OutputStream;
* import org.infinispan.commons.marshall.DelegatingObjectOutput;

unused field in CacheRpcCommandExternalizer:
* private final StreamingMarshaller globalMarshaller;

GlobalMarshaller.objectFromInputStream():
* Do you mind to create a JIRA to remove the copy and use the InputStream directly (create StreamByteObjectInput?)

Isn't the String[] missing in the Primitives?

galderz:
> GlobalMarshaller.objectFromInputStream():
> Do you mind to create a JIRA to remove the copy and use the InputStream directly (create StreamByteObjectInput?)

^ [ISPN-7212](https://issues.jboss.org/browse/ISPN-7212)

> Isn't the String[] missing in the Primitives?

^ [ISPN-7213](https://issues.jboss.org/browse/ISPN-7213)

pruivo:
integrated! thanks @galderz

PR #4649, opened 2016-11-09
ISPN-7181 Include RocksDB JNI into ISPN embedded uber jar

vjuranek:
https://issues.jboss.org/browse/ISPN-7181

### DIFF
--- a/all/embedded/pom.xml
+++ b/all/embedded/pom.xml
@@ -190,7 +190,6 @@
                      <artifactSet>
                         <excludes>
                            <exclude>org.infinispan:infinispan-embedded</exclude>
-                           <exclude>org.rocksdb:rocksdbjni:jar:</exclude>
                            <exclude>org.apache.logging.log4j:log4j-core:jar:</exclude>
                            <exclude>org.apache.logging.log4j:log4j-api:jar:</exclude>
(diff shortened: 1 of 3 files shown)


tristantarrant:
the uberjar is currently full of unneeded dependencies. Would like to see that cleaned up first

vjuranek:
Sorry for late reply @tristantarrant . I looked on it little bit and honestly I'm not sure what should be contained in given uber-jar. I there any list or some description, what should be there (I'm afraid test suite coverage is not sufficient to use it as a hint ... in sense that if test suite is passing it doesn't imply that the changes are ok). Thanks!

slaskawi:
@tristantarrant @vjuranek Hey Guys! Could you please tell me what is the expected result with this PR? The JIRA mentions adding RockDB to the Uber jar and from the commit I see it's being removed. So I'm a bit confused. Could you please give me a hand with this?

vjuranek:
@slaskawi it's removed from exclusion, so effectively it's added :-) Removed is from integration tests where it was added just to make them passing

slaskawi:
ooh sorry, I'm blind... yes, now it makes sense. @tristantarrant I'm ok with those changes. How about you?

slaskawi:
Integrated thanks!

PR #4650, opened 2016-11-09
Continue to provide BaseQuery.getJPAQuery() method for backward compat

anistor:
* we provide this method temporarily to avoid sudden compatibility issues with existing clients
* this is deprecated and will be removed in Infinispan 10.0


tristantarrant:
merged to master. Thanks

PR #4652, opened 2016-11-09
Backports for 5.2.x (BZ1378875 and BZ1378877)

pruivo:
https://bugzilla.redhat.com/show_bug.cgi?id=1378875
https://bugzilla.redhat.com/show_bug.cgi?id=1378877

### DIFF
--- a/core/src/main/java/org/infinispan/transaction/TransactionTable.java
+++ b/core/src/main/java/org/infinispan/transaction/TransactionTable.java
@@ -23,7 +23,23 @@
 
 package org.infinispan.transaction;
 
-import net.jcip.annotations.GuardedBy;
+import static org.infinispan.util.Util.currentMillisFromNanotime;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+import javax.transaction.Transaction;
+import javax.transaction.TransactionSynchronizationRegistry;
+
 import org.infinispan.Cache;
 import org.infinispan.CacheException;
 import org.infinispan.commands.CommandsFactory;
@@ -33,6 +49,8 @@ import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.configuration.cache.Configurations;
 import org.infinispan.context.InvocationContextContainer;
 import org.infinispan.context.impl.TxInvocationContext;
+import org.infinispan.factories.KnownComponentNames;
+import org.infinispan.factories.annotations.ComponentName;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.factories.annotations.Start;
 import org.infinispan.factories.annotations.Stop;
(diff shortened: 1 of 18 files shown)


core/src/main/java/org/infinispan/transaction/TransactionTable.java
```diff
@@ -532,40 +524,82 @@ private void shutDownGracefully() {
...
+        if (completedTransactionsInfo != null && !completedTransactionsInfo.isEmpty()) {
+            try {
+                log.tracef("About to cleanup completed transaction. Initial size is %d", completedTransactionsInfo.size());
+                long beginning = System.nanoTime();
+                long minCompleteTimestamp = System.nanoTime()
+                        - TimeUnit.MILLISECONDS.toNanos(configuration.transaction().completedTxTimeout());
+                int removedEntries = 0;
+
+                // this iterator is weekly consistent and will never throw ConcurrentModificationException
+                Iterator<Map.Entry<Address, CompletedTransactionsPerNode>> nodeIterator = completedTransactionsInfo.entrySet()
```
rvansa: You could just iterate `values()`

core/src/main/java/org/infinispan/transaction/TransactionTable.java
```diff
@@ -532,40 +524,82 @@ private void shutDownGracefully() {
...
+                            removedEntries++;
+                        }
+                    }
+                }
             long duration = System.nanoTime() - beginning;
 
             log.tracef("Finished cleaning up completed transactions. %d transactions were removed, total duration was %d millis, " +
                   "current number of completed transactions is %d", removedEntries, TimeUnit.NANOSECONDS.toMillis(duration),
-                  completedTransactions.size());
+                  completedTransactionsInfo.size());
```
rvansa: I think this is wrong, `completedTransactionsInfo.size()` is just number of nodes, not transactions. The same with the trace log on the beginning.

core/src/main/java/org/infinispan/transaction/TransactionTable.java
```diff
@@ -532,40 +524,82 @@ private void shutDownGracefully() {
     * on a remote node. This might cause leaks, e.g. if the transaction is prepared, committed and prepared again.
     * Once marked as completed (because of commit or rollback) any further prepare received on that transaction are discarded.
     */
-   public void markTransactionCompleted(GlobalTransaction globalTx) {
-      completedTransactions.put(globalTx, System.nanoTime());
-   }
+    public void markTransactionCompleted(GlobalTransaction globalTx) {
+        if (completedTransactionsInfo != null) {
```
pruivo: no, why?

core/src/main/java/org/infinispan/transaction/TransactionTable.java
```diff
@@ -532,40 +524,82 @@ private void shutDownGracefully() {
...
-            long beginning = System.nanoTime();
-            while (iterator.hasNext()) {
-               Map.Entry<GlobalTransaction, Long> e = iterator.next();
-               long ageNanos = System.nanoTime() - e.getValue();
-               if (TimeUnit.NANOSECONDS.toMillis(ageNanos) >= timeout) {
-                  iterator.remove();
-                  removedEntries++;
-               }
-            }
+        if (completedTransactionsInfo != null && !completedTransactionsInfo.isEmpty()) {
```
pruivo: can you explain what you mean?

core/src/main/java/org/infinispan/transaction/TransactionTable.java
rvansa: after it is started, the `completedTransactionsInfo` can't be null - why the check then?

core/src/main/java/org/infinispan/transaction/TransactionTable.java
rvansa:
```
if (foo) {
   indented stuff
}
```
vs.
```
if (!foo) return;
stuff not indented;
```
Makes the code a bit more readable, imo

core/src/main/java/org/infinispan/transaction/TransactionTable.java
pruivo: ```completedTransactionsInfo``` is only set for clustered caches. It is ```null``` for local transactional caches.

danberindei:
@pruivo @rvansa I actually changed it to 2 maps soon after: https://github.com/infinispan/infinispan/commit/8728568b6642d3a9e7177c84a4d3ad515d10e36e
https://issues.jboss.org/browse/ISPN-4293

I think that commit should be included in the backport.

pruivo:
@danberindei done!

rvansa:
OK, all my comments are addressed.

danberindei:
Integrated, thanks Pedro!

PR #4654, opened 2016-11-10
ISPN-4772 Increase the lock acquisition timeout in core tests

danberindei:
https://issues.jboss.org/browse/ISPN-4772

Even when the test expects a failure, it usually also has some
operations that must succeed, so using a timeout of only 100ms
can lead to random failures.


core/src/test/java/org/infinispan/atomic/AtomicHashMapPessimisticConcurrencyTest.java
```diff
@@ -42,7 +43,7 @@
    @Override
    protected EmbeddedCacheManager createCacheManager() throws Exception {
       ConfigurationBuilder builder = TestCacheManagerFactory.getDefaultCacheConfiguration(true);
-      builder.locking().lockAcquisitionTimeout(1000);
+      builder.locking().lockAcquisitionTimeout(TestingUtil.shortTimeoutMillis());
```
pruivo: well, here it is actually reducing the timeout instead of increasing it :)

danberindei: Well, I didn't find any reason why it should be higher here :)

slaskawi:
Allrighy.. let's give this a try on TC :D

slaskawi:
Integrated, thanks!

PR #4656, opened 2016-11-10
ISPN-7184 Server startup can fail after the upgrade to JGroups 4

danberindei:
https://issues.jboss.org/browse/ISPN-7184

Re-add support for empty messages as a signal for missing fork channels
both in embedded mode and in server mode.

The server uses an empty message to signal that the fork channel with
that name is not running on a remote node.
ISPN-6970 made some changes to response unmarshalling and broke that.


jmarkos:
FYI: I tested this and it seems to work.

wburns:
Hrmm looks like build hung.  Have you seen an issue with FullSyncUseSynchronizationTotalOrderTest before @danberindei ?

https://gist.github.com/wburns/6a83f5ee9e88f40afc2cd876b89b9528

danberindei:
@wburns Yes, I even have a fix :)
https://github.com/infinispan/infinispan/pull/4663

wburns:

PR #4660, opened 2016-11-11
[8.2.x] Backports for ISPN-6792, ISPN-7067, ISPN-7088, ISPN-6445, ISPN-7155

danberindei:
https://issues.jboss.org/browse/ISPN-7101

I've also included ISPN-6445, because I was getting random failures in `PreloadingWithWriteBehindTest` locally.


rhusar:
Thanks Dan, I have added ISPN-6445 to the list then too.

There are about 3 more issues that are on the Jira but not included here, will there be part 2 or what's the plan?

tristantarrant:
Gustavo and I are taking care of the rest

rhusar:
Great, thanks for the update.

PR #4663, opened 2016-11-14
ISPN-7172 Total order caches can hang during join

danberindei:
https://issues.jboss.org/browse/ISPN-7172

* Fix WithinThreadExecutor handling in LimitedExecutor
* Remove LimitedExecutor permit before putting the LocalCacheStatus
  in the runningCaches map.

### DIFF
--- a/core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
+++ b/core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
@@ -130,15 +130,17 @@ public class LocalTopologyManagerImpl implements LocalTopologyManager, GlobalSta
       // For Total Order caches, we must not move the topology updates to another thread
       ExecutorService topologyUpdatesExecutor = joinInfo.isTotalOrder() ? withinThreadExecutor : asyncTransportExecutor;
       LocalCacheStatus cacheStatus = new LocalCacheStatus(cacheName, joinInfo, stm, phm, topologyUpdatesExecutor);
+
+      // Pretend the join is using up a thread from the topology updates executor.
+      // This ensures that the initial topology and the GET_CACHE_LISTENERS request will happen on this thread,
+      // and other topology updates are only handled after we complete joinFuture.
+      CompletableFuture<Void> joinFuture = new CompletableFuture<>();
+      cacheStatus.getTopologyUpdatesExecutor().executeAsync(() -> joinFuture);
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
```diff
@@ -130,15 +130,17 @@ public CacheTopology join(String cacheName, CacheJoinInfo joinInfo, CacheTopolog
...
 
-      long timeout = joinInfo.getTimeout();
-      long endTime = timeService.expectedEndTime(timeout, TimeUnit.MILLISECONDS);
       // Pretend the join is using up a thread from the topology updates executor.
       // This ensures that the initial topology and the GET_CACHE_LISTENERS request will happen on this thread,
-      // and other topology updates are only handled after we call backgroundTaskFinished(null)
+      // and other topology updates are only handled after we complete joinFuture.
       CompletableFuture<Void> joinFuture = new CompletableFuture<>();
       cacheStatus.getTopologyUpdatesExecutor().executeAsync(() -> joinFuture);
+
```
pruivo: is this change really needed?

danberindei: I'm not sure if it's strictly necessary for correctness at this point or it just makes tests more predictable, but I think it's safer to implement it as promised in the comments above.

pruivo: I don't see how initializing the  ```runningCaches``` before or after the sending the ```joinFuture``` to the executor could affect anything... 
My point is: I'm not seeing any difference in the logic executed. So, what side effect am I missing?

danberindei: Submitting `joinFuture` will prevent the executor from running any other task until `joinFuture` is completed. If the cache exists in `runningCaches` and it's `LimitedExecutor` has a free spot, it will process topology updates, and those have a chance of doing "stuff" before we have properly joined.
I think the initial stuff I was worried about was just blocking for a new view, which could overwhelm the OOB thread pool given enough caches.

pruivo: makes sense.

pruivo:

PR #4668, opened 2016-11-15
Refactor query api

anistor:
* https://issues.jboss.org/browse/ISPN-2826
  CacheQuery and ResultIterator now have a type parameter. If the query targets multiple types it is the user's responsibility to choose an actual type which is a common super type of all targeted types or CCE will occur during fetching. If no such suitable type exists then the user can specify Object or just declare it raw.
As an exception, ```CacheQuery.projection``` method does not use the type parameter of the class and instead it returns ```CacheQuery<Object[]>```.

* https://issues.jboss.org/browse/ISPN-7205
The query DSL was modified to avoid the need to call ```toBuilder()```. 

* PLEASE pay attention to backward compat. All API changes are intended to be backward source-compatible (possibly binary too, but binary is not an absolute must).

### DIFF
--- a/query/src/main/java/org/infinispan/query/CacheQuery.java
+++ b/query/src/main/java/org/infinispan/query/CacheQuery.java
@@ -20,14 +20,14 @@ import org.hibernate.search.query.engine.spi.FacetManager;
  * @author Marko Luksa
  * @see org.infinispan.query.impl.SearchManagerImpl#getQuery(org.apache.lucene.search.Query, Class...)
  */
-public interface CacheQuery extends Iterable<Object> {
+public interface CacheQuery<E> extends Iterable<E> {
 
    /**
     * Returns the results of a search as a list.
     *
     * @return list of objects that were found from the search.
     */
-   List<Object> list();
+   List<E> list();
 
    /**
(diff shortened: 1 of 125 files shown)


query/src/main/java/org/infinispan/query/CacheQuery.java
```diff
@@ -103,9 +103,9 @@
     * If the projected field is not a projectable field, null is returned in the object[]
     *
     * @param fields the projected field names
-    * @return {@code this}  to allow for method chaining
+    * @return {@code this} to allow for method chaining, but the type parameter now becomes {@code Object[]}
     */
-   CacheQuery projection(String... fields);
+   CacheQuery<Object[]> projection(String... fields);
```
gustavonalle: This is the only think that may break compatibility, but I prefer this version with ```Object[]```

anistor: Yes, I like that better too.

anistor:
I've made some more trivial updates to this PR. It is good to go now from my side.

gustavonalle:
merged, thanks!

PR #4669, opened 2016-11-15

pruivo:
Preview of Triangle code.

~~Missing~~ TODO list:
* Functional API => it is not demanded so far. It will be updated later.
* Replicated caches (it will still uses the old approach) => no performance issues. It will continue working as before.

~~Known issues~~:
* ~~4 or 5 tests randomly failing related to topology changes/state transfer~~
* ~~Code name is changed to help identify in benchmarks (undo when integrate)~~

Changelog:
* Fixed OpDistBackupFailure2Test (xsite)
* Fixed OperationsDuringStateTransferTest
* Should be as stable as master.

### DIFF
--- a/core/src/main/java/org/infinispan/commands/write/DataWriteCommand.java
+++ b/core/src/main/java/org/infinispan/commands/write/DataWriteCommand.java
@@ -1,5 +1,6 @@
 package org.infinispan.commands.write;
 
+import org.infinispan.commands.CommandInvocationId;
 import org.infinispan.commands.DataCommand;
 
 /**
@@ -9,4 +10,32 @@ import org.infinispan.commands.DataCommand;
  * @since 4.0
  */
 public interface DataWriteCommand extends WriteCommand, DataCommand {
+
+   /**
+    * @return the {@link CommandInvocationId} associated to the command.
+    */
+   CommandInvocationId getCommandInvocationId();
+
+   /**
+    * Initializes the {@link BackupWriteRcpCommand} to send the update to backup owner of a key.
+    * <p>
+    * This method will be invoked in the primary owner only.
+    *
+    * @param command the {@link BackupWriteRcpCommand} to initialize.
+    */
+   default void initBackupWriteRcpCommand(BackupWriteRcpCommand command) {
+      throw new UnsupportedOperationException();
+   }
+
+   /**
+    * Initializes the primary owner acknowledge with the return value, the {@link CommandInvocationId} and the topology
+    * id.
+    *
+    * @param command          the {@link PrimaryAckCommand} to initialize.
+    * @param localReturnValue the local return value.
+    */
+   default void initPrimaryAck(PrimaryAckCommand command, Object localReturnValue) {
+      throw new UnsupportedOperationException();
+   }
+
 }
--- a/core/src/main/java/org/infinispan/commands/write/PrimaryAckCommand.java
+++ b/core/src/main/java/org/infinispan/commands/write/PrimaryAckCommand.java
@@ -1,0 +1,161 @@
+package org.infinispan.commands.write;
+
+import org.infinispan.commands.CommandInvocationId;
+import org.infinispan.commands.remote.BaseRpcCommand;
+import org.infinispan.commons.marshall.MarshallUtil;
+import org.infinispan.util.ByteString;
+import org.infinispan.util.concurrent.CommandAckCollector;
+import org.infinispan.util.concurrent.CompletableFutures;
+
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * A command that represents an acknowledge sent by the primary owner to the originator.
+ * <p>
+ * The acknowledge signals a successful or unsuccessful execution of the operation and it contains the return value of
+ * the operation.
+ *
+ * @author Pedro Ruivo
+ * @since 9.0
+ */
+public class PrimaryAckCommand extends BaseRpcCommand {
+
+   public static final byte COMMAND_ID = 30;
+   private static final Type[] CACHED_TYPE = Type.values();
+   private CommandInvocationId commandInvocationId;
+   private Object returnValue;
+   private Type type;
+   private CommandAckCollector commandAckCollector;
+   private int topologyId;
+
+   public PrimaryAckCommand() {
+      super(null);
+   }
+
+   public PrimaryAckCommand(ByteString cacheName) {
+      super(cacheName);
+   }
+
+   private static Type valueOf(int index) {
+      return CACHED_TYPE[index];
+   }
+
+   public void initCommandInvocationIdAndTopologyId(CommandInvocationId id, int topologyId) {
+      this.commandInvocationId = id;
+      this.topologyId = topologyId;
+   }
+
+   @Override
+   public CompletableFuture<Object> invokeAsync() throws Throwable {
+      switch (type) {
+         case SUCCESS_WITH_BOOL_RETURN_VALUE:
+         case SUCCESS_WITH_RETURN_VALUE:
+         case SUCCESS_WITHOUT_RETURN_VALUE:
+            commandAckCollector.primaryAck(commandInvocationId, returnValue, true, getOrigin(), topologyId);
+            break;
+         case UNSUCCESSFUL_WITH_BOOL_RETURN_VALUE:
+         case UNSUCCESSFUL_WITH_RETURN_VALUE:
+         case UNSUCCESSFUL_WITHOUT_RETURN_VALUE:
+            commandAckCollector.primaryAck(commandInvocationId, returnValue, false, getOrigin(), topologyId);
+            break;
+      }
+      return CompletableFutures.completedNull();
+   }
+
+   @Override
+   public byte getCommandId() {
+      return COMMAND_ID;
+   }
+
+   @Override
+   public boolean isReturnValueExpected() {
+      return false;
+   }
+
+   @Override
+   public void writeTo(ObjectOutput output) throws IOException {
+      output.writeInt(topologyId);
+      CommandInvocationId.writeTo(output, commandInvocationId);
+      MarshallUtil.marshallEnum(type, output);
+      switch (type) {
+         case SUCCESS_WITH_RETURN_VALUE:
+         case UNSUCCESSFUL_WITH_RETURN_VALUE:
+            output.writeObject(returnValue);
+            break;
+         default:
+      }
+   }
+
+   @Override
+   public void readFrom(ObjectInput input) throws IOException, ClassNotFoundException {
+      topologyId = input.readInt();
+      commandInvocationId = CommandInvocationId.readFrom(input);
+      type = MarshallUtil.unmarshallEnum(input, PrimaryAckCommand::valueOf);
+      assert type != null;
+      switch (type) {
+         case SUCCESS_WITH_RETURN_VALUE:
+         case UNSUCCESSFUL_WITH_RETURN_VALUE:
+            returnValue = input.readObject();
+            break;
+         case SUCCESS_WITH_BOOL_RETURN_VALUE:
+            returnValue = true;
+            break;
+         case UNSUCCESSFUL_WITH_BOOL_RETURN_VALUE:
+            returnValue = false;
+            break;
+         default:
+      }
+   }
+
+   public void initWithReturnValue(boolean success, Object returnValue) {
+      this.returnValue = returnValue;
+      if (success) {
+         type = Type.SUCCESS_WITH_RETURN_VALUE;
+      } else {
+         type = Type.UNSUCCESSFUL_WITH_RETURN_VALUE;
+      }
+   }
+
+   public void initWithBoolReturnValue(boolean success) {
+      this.returnValue = success;
+      if (success) {
+         type = Type.SUCCESS_WITH_BOOL_RETURN_VALUE;
+      } else {
+         type = Type.UNSUCCESSFUL_WITH_BOOL_RETURN_VALUE;
+      }
+   }
+
+   public void initWithoutReturnValue(boolean success) {
+      if (success) {
+         type = Type.SUCCESS_WITHOUT_RETURN_VALUE;
+      } else {
+         type = Type.UNSUCCESSFUL_WITHOUT_RETURN_VALUE;
+      }
+   }
+
+   public void setCommandAckCollector(CommandAckCollector commandAckCollector) {
+      this.commandAckCollector = commandAckCollector;
+   }
+
+   @Override
+   public String toString() {
+      return "PrimaryAckCommand{" +
+            "commandInvocationId=" + commandInvocationId +
+            ", returnValue=" + returnValue +
+            ", type=" + type +
+            ", topologyId=" + topologyId +
+            '}';
+   }
+
+   private enum Type {
+      SUCCESS_WITH_RETURN_VALUE,
+      SUCCESS_WITH_BOOL_RETURN_VALUE,
+      SUCCESS_WITHOUT_RETURN_VALUE,
+      UNSUCCESSFUL_WITH_RETURN_VALUE,
+      UNSUCCESSFUL_WITH_BOOL_RETURN_VALUE,
+      UNSUCCESSFUL_WITHOUT_RETURN_VALUE
+   }
+}
--- a/core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
@@ -4,6 +4,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
@@ -39,6 +40,7 @@ import org.infinispan.interceptors.impl.BaseRpcInterceptor;
 import org.infinispan.interceptors.impl.MultiSubCommandInvoker;
 import org.infinispan.interceptors.locking.ClusteringDependentLogic;
 import org.infinispan.statetransfer.StateTransferLock;
+import org.infinispan.util.concurrent.CommandAckCollector;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
@@ -60,6 +62,7 @@ public class L1NonTxInterceptor extends BaseRpcInterceptor {
    protected DataContainer dataContainer;
    protected Configuration config;
    protected StateTransferLock stateTransferLock;
+   private CommandAckCollector commandAckCollector;
 
    private long l1Lifespan;
    private long replicationTimeout;
@@ -92,7 +95,7 @@ public class L1NonTxInterceptor extends BaseRpcInterceptor {
    @Inject
    public void init(L1Manager l1Manager, ClusteringDependentLogic cdl, EntryFactory entryFactory,
                     DataContainer dataContainer, Configuration config, StateTransferLock stateTransferLock,
-                    CommandsFactory commandsFactory) {
+                    CommandsFactory commandsFactory, CommandAckCollector commandAckCollector) {
       this.l1Manager = l1Manager;
       this.cdl = cdl;
       this.entryFactory = entryFactory;
@@ -100,6 +103,7 @@ public class L1NonTxInterceptor extends BaseRpcInterceptor {
       this.config = config;
       this.stateTransferLock = stateTransferLock;
       this.commandsFactory = commandsFactory;
+      this.commandAckCollector = commandAckCollector;
    }
 
    @Start
@@ -143,8 +147,8 @@ public class L1NonTxInterceptor extends BaseRpcInterceptor {
       return invokeNext(ctx, command);
    }
 
-   protected BasicInvocationStage performL1Lookup(InvocationContext ctx, VisitableCommand command,
-         boolean runInterceptorOnConflict, Object key, boolean isEntry) throws Throwable {
+   private BasicInvocationStage performL1Lookup(InvocationContext ctx, VisitableCommand command,
+                                                boolean runInterceptorOnConflict, Object key, boolean isEntry) throws Throwable {
       // Most times the putIfAbsent will be successful, so not doing a get first
       L1WriteSynchronizer l1WriteSync = new L1WriteSynchronizer(dataContainer, l1Lifespan, stateTransferLock,
                                                                 cdl);
@@ -286,7 +290,7 @@ public class L1NonTxInterceptor extends BaseRpcInterceptor {
    }
 
    private BasicInvocationStage handleDataWriteCommand(InvocationContext ctx, DataWriteCommand command,
-         boolean assumeOriginKeptEntryInL1) throws Throwable {
+         boolean assumeOriginKeptEntryInL1) {
       if (command.hasFlag(Flag.CACHE_MODE_LOCAL)) {
          if (trace) {
             log.tracef("local mode forced, suppressing L1 calls.");
@@ -301,11 +305,15 @@ public class L1NonTxInterceptor extends BaseRpcInterceptor {
       });
    }
 
-   private BasicInvocationStage removeFromLocalL1(InvocationContext ctx, DataWriteCommand command, Object returnValue)
-         throws Throwable {
+   private BasicInvocationStage removeFromLocalL1(InvocationContext ctx, DataWriteCommand command, Object returnValue) {
       if (ctx.isOriginLocal() && !cdl.localNodeIsOwner(command.getKey())) {
+         CompletableFuture<?> pendingAcks = commandAckCollector.getCollectorCompletableFuture(command.getCommandInvocationId(), false);
          VisitableCommand removeFromL1Command = removeFromL1Command(ctx, command.getKey());
-         return invokeNext(ctx, removeFromL1Command).thenApply((rCtx, rCommand, rv) -> returnValue);
+         if (pendingAcks == null) {
+            return invokeNext(ctx, removeFromL1Command).thenApply((rCtx, rCommand, rv) -> returnValue);
+         } else {
+            return invokeNextAsync(ctx, removeFromL1Command, pendingAcks).thenApply((rCtx, rCommand, rv) -> returnValue);
+         }
       } else if (trace) {
          log.trace("Allowing entry to commit as local node is owner");
       }
--- a/core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
+++ b/core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
@@ -48,6 +48,26 @@ public interface RpcManager {
    Map<Address, Response> invokeRemotely(Map<Address, ReplicableCommand> rpcs, RpcOptions options);
 
    /**
+    * Asynchronously sends the {@link ReplicableCommand} to the destination using the specified {@link DeliverOrder}.
+    *
+    * @param destination  the destination's {@link Address}.
+    * @param command      the {@link ReplicableCommand} to send.
+    * @param deliverOrder the {@link DeliverOrder} to use.
+    */
+   void sendTo(Address destination, ReplicableCommand command, DeliverOrder deliverOrder);
+
+   /**
+    * Asynchronously sends the {@link ReplicableCommand} to the set of destination using the specified {@link
+    * DeliverOrder}.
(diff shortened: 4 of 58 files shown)


core/src/main/java/org/infinispan/commands/Visitor.java
```diff
@@ -57,6 +58,8 @@
 
    Object visitApplyDeltaCommand(InvocationContext ctx, ApplyDeltaCommand command) throws Throwable;
 
+   Object visitBackupWriteCommand(InvocationContext ctx, BackupWriteCommand command) throws Throwable;
```
pruivo: why? the other commands aren't ```default``` and its default implementation is in ```AbstractVisitor```

core/src/main/java/org/infinispan/commands/write/BackupPutMapAckCommand.java
```diff
@@ -0,0 +1,97 @@
...
+      super(cacheName);
+      this.commandInvocationId = commandInvocationId;
+      this.segments = segments;
+      this.topologyId = topologyId;
+   }
+
+   @Override
+   public Object invoke() throws Throwable {
+      commandAckCollector.putMapBackupAck(commandInvocationId, getOrigin(), segments, topologyId);
+      return null;
```
pruivo: actually, ```perform()```, ```invoke()``` and ```invokeAsync()``` have a default implementation. It was kind of confusing which one to implement :)

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -0,0 +1,512 @@
...
+      if (collector != null) {
+         try {
+            if (trace) {
+               log.tracef("[Collector#%s] Waiting for acks.", id);
+            }
+            return awaitFuture(collector.getFuture(), timeout, timeUnit);
+         } catch (CompletionException e) {
+            throw CompletableFutures.extractException(e);
+         } finally {
+            collectorMap.remove(id);
```
pruivo: I don't see the advantage of it. I need to create a collector every time the command is retried. So, I need to remove it when the command finishes or when the command needs to retry.

core/src/main/java/org/infinispan/statetransfer/OutdatedTopologyException.java
```diff
@@ -5,19 +5,44 @@
...
+ * <p>
  * This can happen for non-tx caches, if the primary owner doesn't respond (either because it left the
  * cluster or because this particular cache is no longer running).
  *
  * @author Dan Berindei
  * @since 6.0
  */
 public class OutdatedTopologyException extends CacheException {
+
+   //TODO? should we create a special externalizer to return the cached exception? Is it ever worth to send the stack trace/message when originated remotely?
```
pruivo: ok

core/src/main/java/org/infinispan/commands/write/DataWriteCommand.java
```diff
@@ -9,4 +10,31 @@
...
+      throw new UnsupportedOperationException();
+   }
+
+   /**
+    * Initializes the primary owner acknowledges with the return value and if it is successful or not.
+    *
+    * @param command     the {@link PrimaryAckCommand} to initialize.
+    * @param returnValue the local return value.
+    */
+   default void initPrimaryAck(PrimaryAckCommand command, Object returnValue) {
```
pruivo: hmmm... I think I can create the primary ack here. I'll change

core/src/main/java/org/infinispan/commands/write/PrimaryAckCommand.java
```diff
@@ -0,0 +1,172 @@
...
+         type = Type.UNSUCCESSFUL_WITH_RETURN_VALUE;
+      }
+   }
+
+   public void initWithBoolReturnValue(boolean success) {
+      this.returnValue = success;
+      if (success) {
+         type = Type.SUCCESS_WITH_BOOL_RETURN_VALUE;
+      } else {
+         type = Type.UNSUCCESSFUL_WITH_BOOL_RETURN_VALUE;
```
pruivo: don't get what you mean. you don't like the enum and you'd like to  have a ```successful``` field instead. is that it?

core/src/main/java/org/infinispan/interceptors/impl/DistCacheWriterInterceptor.java
```diff
@@ -87,6 +88,25 @@ public BasicInvocationStage visitPutKeyValueCommand(InvocationContext ctx, PutKe
    }
 
    @Override
+   public Object visitBackupWriteCommand(InvocationContext ctx, BackupWriteCommand command) throws Throwable {
+      return invokeNext(ctx, command).thenApply((rCtx, rCommand, rv) -> {
+         BackupWriteCommand backupWriteCommand = (BackupWriteCommand) rCommand;
+         if (!isStoreEnabled(backupWriteCommand))
+            return rv;
+
+         //no need to check for proper write. we are remote, in a backup owner.
```
pruivo: I'll check for the topology id before updating the persistence.

core/src/main/java/org/infinispan/commands/write/DataWriteCommand.java
pruivo: actually I think it is a bad idea. the primary ack command needs the cache name and it is not available in the data write commands.

pruivo:
rebased and updated

core/src/main/java/org/infinispan/commands/Visitor.java
danberindei: Fair enough.

core/src/main/java/org/infinispan/commands/write/PrimaryAckCommand.java
danberindei: No, I was thinking the boolean return value doesn't add anything because you're already using a different `Type` for each possible value.

core/src/main/java/org/infinispan/interceptors/impl/DistCacheWriterInterceptor.java
danberindei: I think that would only work with `acquireSharedTopologyLock` (sorry, that's the one I meant, not the exclusive one). Without it you'd still have a chance of a topology change between the topology check and the actual write. If that happens, we could be writing an entry that we don't own any more after the stale entry invalidation. The command would be retried because EWI would notice the topology change during commit, but the entry would stay in the store.

While writing this I realized shared stores are even weirder, because we kind of assume exclusive access to each key from the primary owner, yet there can be multiple nodes thinking they're the primary owner at the same time. E.g. with a JDBC store that doesn't support upsert, I think we could search for the key, find 0 rows and decide to do an insert, and then fail it because the other primary owner already did the insert. There was a question on SO about primary key violation errors with a shared store and numOwners > 1, I wonder if their test involved something like this...

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
danberindei: I was thinking about changing the async operations to not block any threads on the originator.

danberindei:
@pruivo maybe you should change this to `Ready for review` and split the replicated mode and functional ops into separate PRs?

core/src/main/java/org/infinispan/interceptors/impl/DistCacheWriterInterceptor.java
pruivo: that doesn't seem a problem with the Triangle but with our code in general. I'll open a JIRA and fix it later.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: I've changed to use ```returnWithAsyn()```. let me know if it is ok.

core/src/main/java/org/infinispan/interceptors/impl/DistCacheWriterInterceptor.java
pruivo: created: https://issues.jboss.org/browse/ISPN-7239

pruivo:
rebased and updated. 
Updated the description as well and changed the state to Ready :)

core/src/main/java/org/infinispan/interceptors/impl/DistCacheWriterInterceptor.java
danberindei: Ok

core/src/main/java/org/infinispan/remoting/transport/Transport.java
```diff
@@ -55,6 +55,38 @@
...
+    * Asynchronously sends the {@link ReplicableCommand} to the set of destination using the specified {@link
+    * DeliverOrder}.
+    *
+    * @param rpcCommand   the {@link ReplicableCommand} to send.
+    * @param deliverOrder the {@link DeliverOrder} to use.
+    * @param destinations the collection of destination's {@link Address}. If {@code null}, it sends to all the members
+    *                     in the cluster.
+    * @throws Exception if there was problem sending the request.
+    */
+   void sendTo(ReplicableCommand rpcCommand, DeliverOrder deliverOrder, Collection<Address> destinations) throws Exception;
```
rvansa: Order of arguments

core/src/main/java/org/infinispan/remoting/transport/Transport.java
```diff
@@ -55,6 +55,38 @@
...
+
+   /**
+    * Asynchronously sends the {@link ReplicableCommand} to all members in the cluster using the specified {@link
+    * DeliverOrder}.
+    *
+    * @param rpcCommand   the {@link ReplicableCommand} to send.
+    * @param deliverOrder the {@link DeliverOrder} to use.
+    * @throws Exception if there was problem sending the request.
+    */
+   void sendToAll(ReplicableCommand rpcCommand, DeliverOrder deliverOrder) throws Exception;
```
rvansa: Is this necessary? The other methods are fine with `destionations == null` meaning "all"

core/src/main/java/org/infinispan/statetransfer/OutdatedTopologyException.java
rvansa: Since the ST isn't helpful, we shouldn't record it at all (the Exception(msg, cause, false, false) ctor)

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -0,0 +1,482 @@
...
+         } else if (owners.retainAll(members) && owners.isEmpty()) {
+            if (trace) {
+               log.tracef("[Collector#%s] Some backups left the cluster.", id);
+            }
+            markReady();
+         }
+      }
+   }
+
+   private static class PutMapCollector implements Collector<Map<Object, Object>> {
```
rvansa: This should rather be `MultiKeyCollector` (or some better name) to accomodate functional commands as well. At least I hope it will work for func commands, too.
And also rename the methods above.

core/src/test/java/org/infinispan/api/NonTxCacheTest.java
```diff
@@ -58,7 +56,7 @@ public void testRemove() throws Exception {
 
    @Override
    protected void createCacheManagers() throws Throwable {
-      ConfigurationBuilder builder = getDefaultClusteredCacheConfig(cacheMode, false);
+      ConfigurationBuilder builder = getDefaultClusteredCacheConfig(CacheMode.REPL_SYNC, false);
```
rvansa: Have you removed testing distributed case intentionally?

core/src/test/java/org/infinispan/distribution/DistSyncL1FuncTest.java
```diff
@@ -3,45 +3,41 @@
...
 import org.infinispan.test.TestingUtil;
 import org.infinispan.tx.dld.ControlledRpcManager;
-import org.infinispan.util.concurrent.IsolationLevel;
+import org.infinispan.util.concurrent.CommandAckCollector;
 import org.testng.annotations.Test;
 
 @Test(groups = {"functional", "smoke"}, testName = "distribution.DistSyncL1FuncTest")
 public class DistSyncL1FuncTest extends BaseDistSyncL1Test {
 
-   @Override
```
rvansa: Again, removing test cases?

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -0,0 +1,482 @@
...
+                  id, from, topologyId, this.topologyId, segments);
+         }
+         if (this.topologyId != topologyId) {
+            return;
+         }
+         ack(backups, from, segments);
+      }
+
+      private void ack(Map<Address, Collection<Integer>> map, Address from, Collection<Integer> segments) {
+         Collection<Integer> pendingSegments = map.getOrDefault(from, Collections.emptyList());
```
rvansa: When does a node send just some segments (and not all of them)? (of course primary can first send primary ack for some segments and then backup ack for those it is a backup, but I am asking about multiple acks from different senders).

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
```diff
@@ -48,6 +48,26 @@
    Map<Address, Response> invokeRemotely(Map<Address, ReplicableCommand> rpcs, RpcOptions options);
 
    /**
+    * Asynchronously sends the {@link ReplicableCommand} to the destination using the specified {@link DeliverOrder}.
+    *
+    * @param destination  the destination's {@link Address}.
+    * @param command      the {@link ReplicableCommand} to send.
+    * @param deliverOrder the {@link DeliverOrder} to use.
+    */
+   void sendTo(Address destination, ReplicableCommand command, DeliverOrder deliverOrder);
```
pruivo: remove ```ResponseMode.ASYNCHRONOUS```? I don't think so. 
The ```invokeRemotely()``` are still useful for replicated caches and tx caches. But, IMO, it doesn't make sense to use them in triangle because we know that all messages are sent asynchronously. It is a overkill to go through the ```invokeRemotely()``` logic. 
We can't do the same for replicated/tx caches the configuration can change the message sending.

core/src/main/java/org/infinispan/remoting/transport/Transport.java
pruivo: hmmm. yes I think it makes sense to remove it.

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
```diff
@@ -48,6 +48,26 @@
...
+   /**
+    * Asynchronously sends the {@link ReplicableCommand} to the set of destination using the specified {@link
+    * DeliverOrder}.
+    *
+    * @param command      the {@link ReplicableCommand} to send.
+    * @param deliverOrder the {@link DeliverOrder} to use.
+    * @param destinations the collection of destination's {@link Address}. If {@code null}, it sends to all the members
+    *                     in the cluster.
+    */
+   void sendTo(ReplicableCommand command, DeliverOrder deliverOrder, Collection<Address> destinations);
```
pruivo: I only changed the order to remove the warning when you invoke with destination ```null``` (it doesn't know which method to invoke)

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: ok

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: the segments in the backups can have different primary owners. So, the backup will receive a different command from them and it only acks a subset of its own segments.

core/src/test/java/org/infinispan/api/NonTxCacheTest.java
pruivo: yes, it is testing an issue with the old algorithm. Triangle doesn't suffer the same issue and only the replicated cache does.

core/src/test/java/org/infinispan/distribution/DistSyncL1FuncTest.java
pruivo: yes. the isolation level doesn't affect non-transactional caches. It doesn't make sense to test the same thing 3 times...

core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
```diff
@@ -301,11 +317,15 @@ private BasicInvocationStage handleDataWriteCommand(InvocationContext ctx, DataW
       });
    }
 
-   private BasicInvocationStage removeFromLocalL1(InvocationContext ctx, DataWriteCommand command, Object returnValue)
-         throws Throwable {
+   private BasicInvocationStage removeFromLocalL1(InvocationContext ctx, DataWriteCommand command, Object returnValue) {
       if (ctx.isOriginLocal() && !cdl.localNodeIsOwner(command.getKey())) {
+         CompletableFuture<?> pendingAcks = commandAckCollector.getCollectorCompletableFuture(command.getCommandInvocationId(), false);
```
danberindei: Shouldn't `TriangleInterceptor`'s `returnWithAsync` only complete after we got all the acks?

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
danberindei: Speaking of the `invokeRemotely()` logic, have you thought whether we should change the RPC stats to consider the duration of the RPC from the moment we sent the command to the primary and until we got the ack from the last backup? The same for the replication timeout, should it apply just to the interaction with the primary or should it include the time spent waiting for backup acks?

core/src/test/java/org/infinispan/api/NonTxCacheTest.java
danberindei: Maybe you could change the test name to make it more specific to the bug it's testing?

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
rvansa: I see. In that case I would rather consider `sendTo` and `sendToMany` with the same order of arguments.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: yes, I think you're right. we only receive one ack from each primary

core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
pruivo: isn't that the goal of ```returnWithAsync```? wait for the ```CompletablyFuture``` I provide and then continue to execute the rest of the return handlers?

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
pruivo: About the stats, I haven't think about it. Do you think it makes sense? IMO, not really. It should be similar than the average write time.
About the replication timeout, it is used to wait for backups acks. I think it is more inline with the previous algorithm...

pruivo:
rebased.

rvansa:
@pruivo One last question: I don't really like the behaviour-switch in BackupWriteCommand, and the asymmetry with PutMapCommand which is forwarded as-is. Is the reason for a separate command just not marshalling the unneeded bits. or is the another problem this solves?

core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
danberindei: I thought you had already waited for the same `CompletableFuture` in `TriangleInterceptor`, but I see now I was confusing `TriangleInterceptor` and `TriangleDistributionInterceptor`. WDYT about renaming `TriangleInterceptor` to `TriangleAckInterceptor`?

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
danberindei: I was asking because I'm not convinced the replication time is relevant with the master algorithm either. (We need a name for it, maybe "sync forwarding"?) 

E.g. the RPCs from the primary to the backups are double-counted, because they're recorded explicitly on the primary and they're also recorded as part of the RPC from the originator to the primary. So I'm thinking about removing the RpcManagerImpl statistics altogether WDYT?

core/src/main/java/org/infinispan/commands/write/BackupWriteCommand.java
```diff
@@ -0,0 +1,297 @@
...
+   }
+
+   @Override
+   public boolean ignoreCommandOnStatus(ComponentStatus status) {
+      return false;
+   }
+
+   @Override
+   public boolean readsExistingValues() {
+      return false;
```
rvansa: What if the command carries DELTA_WRITE? Then it should read previous value...

danberindei: +1, although I'm not sure we have proper tests for non-tx DeltaAware writes yet.

danberindei:
@rvansa Yeah, the switch in `BackupWriteCommand` doesn't look pretty. @pruivo Maybe we can copy the entry flags from the primary's context entry and only have a special case for `DELTA_WRITE`?

core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
pruivo: the TriangleInterceptor is placed before the L1 and the L1 needs to send the invalidation after the operation is completed. 
I can change the name if you like, but I think you got confused by something else, no?

core/src/main/java/org/infinispan/commands/write/BackupWriteCommand.java
pruivo: I'm not sure if it is really needed. The DeltaAwareCacheEntry.commit() will invoke the DataContainer.compute() and will merge under the most recent value.

pruivo:
the BackupWriteCommand exists for the following reasons:
* doesn't need to send garbage
* code is simplified since we know that is remote and it is backup (topology changes will throw exception)

I'm ok to create a backup version of PutMap if you want.

the switch is needed because I need a way to distinguish between a remove and an remove expire. Adding the DELTA to the enum seem a good idea to me.

core/src/main/java/org/infinispan/commands/write/BackupWriteCommand.java
rvansa: OK, forget about it... It does not break any tests and I'll change the method soon anyway.
It's just about the method semantics, and when should the value from persistence be loaded.
:skull: to DeltaAware...

pruivo: ah ah. I'll change :)

currently,I have 3 things I hate: the delta and the PutMap (multi key commands in general), the pre events :)

rvansa:
@pruivo I just think that it does not fit into the big picture well, but it would require major rework to unify everything. It can simplify *your* code == Triangle*Interceptors, but the other interceptors must deal with another method *and* the possibility that PKVC is executed on backup node.
Btw., for marshalling you could follow the pattern in `ClusteredGetCommand` (which is not too different from `SingleRpcCommand(GKVC)` anymore). A flag that would alter the marshalling would help reduce the garbage well enough, too.

pruivo:
@rvansa I agree that it is a little bit out of the box with the remaining code. IMO, this is the way to go. Currently, we have to many condition in the commands. For example, the PKVC you have to check if it is condition, if it is successful, if it is delta, if it is loca/remote, if it is a owner or not... and so won. 

Can you explain your idea about the marshalling? I don't mind to make it a ```CacheRpcCommand``` but the ```invoke()``` method will be invoked when the command is delivered and in the interceptor chain.

rvansa:
@pruivo True, but then you have a lot of boilerplate in the interceptors. But we have to prohibit uses of PKVC as backup command (eventually, not right now). Would be great if you could start a wiki page outlining the "future big picture" then (CC me then pls).

The idea was to send `BackupWriteCommand` that would be non-`VisitableCommand` over the wire, and then in `invoke` create PKVC/whatever and pass that down the stack. There's one unnecessary allocation but that shouldn't be too troublesome since it's short-lived (unmarshall -> invoke).
But if you want this command in interceptor chain, it's void.

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
danberindei: I've created https://issues.jboss.org/browse/ISPN-7272

core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
danberindei: I guess the `TriangleInterceptor` name made me think there's only one interceptor, and I forgot that (and why) you need a different one to receive the acks. But I guess you're right, what I really needed was a comment in the code explaining why we need to wait for the acks here ;)

pruivo: remember in the "sync forward" algorithm, when the distribution interceptor returns, all the backups have updated its data container. Now, it has only a CompletableFuture. All tasks needed after the updates need to wait the future. If no interceptor has waited for it, it is the TriangleInterceptor job to do it. 
I'll change the name to TirangleAckInterceptor

pruivo:
@rvansa If I remove the BackupWriteCommand, will I get this integrated?

rvansa:
@pruivo I don't insist on removing that command at this point, I am fine with integrating this as-is. But since you've started to break certain concepts, you should have a plan how to complete that, because current state is not ideal.

danberindei:
-1 to remove BackupWriteCommand at this point. I also want to move towards unifying put/remove/etc., but @rvansa is right that we need a plan.

core/src/main/java/org/infinispan/interceptors/distribution/L1NonTxInterceptor.java
danberindei: I'm not sure L1 really needs to wait for the value to be updated on all the backups before doing the invalidation. Each owner maintains its own list of requestors and does its own invalidations, so the invalidation from `L1LastChanceInterceptor` (after `EntryWrappingInterceptor` committed the entry) should be enough.

danberindei: Oops, L1NonTxInterceptor actually sends the L1 invalidations before invokeNext, after waiting for the acks, because we're on the originator :)

With the current code we could still remove the local L1 entry and request it again from another command. Even if we get the old value from the backup, the backup write will send an L1 invalidation to when it actually applies the value. However, that's only because L1NonTxInterceptor uses `ctx.getOrigin()`, which is the primary, instead of the actual originator of the write. So it's probably better to keep this wait.

pruivo:
@rvansa @danberindei updated.
to avoid entropy, I've removed BackupWriteCommand and created BackWriteRpcCommand (doesn't wrap the commands in a SingleRpcCommand).

pruivo:
@danberindei @rvansa yes, I've a plan to update the commands. 
The BackupWriteCommand will be handled by non-tx caches in the backup owner. Similar command would be created for the PutMapCommand.  The functional API doesn't need any change.
About the tx caches, the write commands array would be removed and replaced by an UpdateEntry (or similar). One per key (no duplication, if a key is written multiple times, only the last one sticks) with all the info needed (previous version is write skew enabled). This won't be a command.

rvansa:
@pruivo Not the functional API, but if the backup func write will be handled by the BackupWriteCommand, there will be another field for the function and enum value, along with updating all the switches that use this enum. And the counterpart for PutMapCommand will also need to accomodate (Read)WriteMany(Entries)Commands.

The discrepancy I see in BWC is that on origin this is processed as the PKVC, forwarded to primary as PKVC but then to backup as BWC with a big enum. You also don't like checking `ctx.isOriginLocal()`, so I thought the "big plan" would include `PrimaryWriteCommand` with all the flavours in enum, and the PKVC form would persist only on the originator. And then I'd expect some resolution for the primary == originator case.

danberindei:
Integrated, thanks Pedro!

pruivo:
@rvansa I dont see why we need to change the functional API. You only have 2 "things" invoked. the key and the function. And both needs to be handled by the backup/primary owners.

the "normal" operations, the backup owners only have 2 possible operations: write or remove. The backups don't need to know if it is a put or a replace, conditional or not.

PR #4671, opened 2016-11-18
ISPN-6906 Add ExternalPojo marker

galderz:
* Marker interface used to signal test classes that are supposed to be
  externally marshalled.

### DIFF
--- a/core/src/main/java/org/infinispan/marshall/core/ExternallyMarshallable.java
+++ b/core/src/main/java/org/infinispan/marshall/core/ExternallyMarshallable.java
@@ -2,18 +2,18 @@ package org.infinispan.marshall.core;
 
 import java.io.Serializable;
 
-// Temporary class, should become active by default when testing
 final class ExternallyMarshallable {
 
    private ExternallyMarshallable() {
       // Static class
    }
 
-   public static boolean isAllowed(Object obj) {
+   static boolean isAllowed(Object obj) {
       String pkg = obj.getClass().getPackage().getName();
       boolean isBlackList =
             obj instanceof Serializable
             && isMarshallablePackage(pkg)
+            && !(obj instanceof ExternalPojo)
             && !isWhiteList(obj.getClass().getName());
       return !isBlackList;
    }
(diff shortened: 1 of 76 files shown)


pruivo:
is this exception supposed to printed? 
```
Caused by: java.lang.AssertionError: Check support for: class org.infinispan.tx.exception.ExceptionInCommandTest$MyDelta
```

core/src/main/java/org/infinispan/marshall/core/ExternallyMarshallable.java
```diff
@@ -2,18 +2,18 @@
...
    }
 
-   public static boolean isAllowed(Object obj) {
+   static boolean isAllowed(Object obj) {
       String pkg = obj.getClass().getPackage().getName();
       boolean isBlackList =
             obj instanceof Serializable
             && isMarshallablePackage(pkg)
+            && !(obj instanceof ExternalPojo)
             && !isWhiteList(obj.getClass().getName());
```
danberindei: I'm not sure how this is supposed to work for user classes, as users can't modify the whitelist and they can't always modify their classes to extend `ExternalPojo` either. They could write+configure externalizers, but do we want to force them to write externalizers for all their classes?

My thinking was that we'd only have the blacklist active during the tests, so we'd only need `ExternalPojo` for our internal classes.

pruivo: hmmm... what is the problem that ```ExternalPojo``` tries to solve?

pruivo:
the ```HotRodDistributedEventsTest``` failures seems related to missing ```ExternalPojo``` in some test classes.

rvansa:
Why do you use a marker interface instead of annotation? I thought marker interfaces are Java 1.4 thing.

core/src/main/java/org/infinispan/marshall/core/ExternallyMarshallable.java
galderz: > I'm not sure how this is supposed to work for user classes

^ That was never the objective of this, but only for test classes and the check only happens when assertions are enabled.

galderz: > hmmm... what is the problem that ExternalPojo tries to solve?

Detect Infinispan classes for which we should have externalizers. If we forget that a class has a externalizer, and it uses standard Serialization, this check will catch it.

If we really expect it to be Serializable, e.g. testing Serializable classes, then the class can be made to extend ExternalPojo interface as marker.

galderz:
> the HotRodDistributedEventsTest failures seems related to missing ExternalPojo in some test classes.

Ok, I'll address that.

> Why do you use a marker interface instead of annotation? I thought marker interfaces are Java 1.4 thing.

Only used for testing, either method should work just fine.

galderz:
Fixed `ExceptionInCommandTest` and `HotRodDistributedEventsTest`

galderz:
Rebasing...

galderz:
Rebased

core/src/main/java/org/infinispan/marshall/core/ExternallyMarshallable.java
danberindei: @galderz I was thinking about the `ExternalPojo` javadoc, which says `One such example are pojos used in testing, but not limited to,` and I wasn't sure what the actual limits are ;)

danberindei:
Integrated, thanks Galder!

PR #4657, opened 2016-11-10
Ispn 7177 Remove Outdated Eviction related classes

wburns:
This builds off of #4578 and cannot be integrated until after it.

https://issues.jboss.org/browse/ISPN-7177

PR #4672, opened 2016-11-18
p may become null and this should be checked

rvprg:


wburns:
This change is not needed with #4657 as this class will be completely removed.

rvprg:
Perfect! Thank you.

PR #4676, opened 2016-11-21
ISPN-6906 Fix offset bug when backtracking

galderz:


core/src/main/java/org/infinispan/marshall/core/BytesObjectInput.java
```diff
@@ -68,7 +68,7 @@ public long skip(long n) {
       } else {
          int idx = Math.min(bytes.length, pos);
          long skip = idx + n;
-         pos = (int) (skip + offset);
+         pos = (int) skip;
```
pruivo: shouldn't be the ```max()``` between ```offset``` and ```skip``` to avoid the ```pos``` to be previous than ```offset```?

galderz:
@pruivo Fixed

pruivo:
integrated! thanks @galderz !

core/src/main/java/org/infinispan/marshall/core/BytesObjectInput.java
```diff
@@ -68,7 +68,8 @@ public long skip(long n) {
       } else {
          int idx = Math.min(bytes.length, pos);
          long skip = idx + n;
-         pos = (int) (skip + offset);
+         // Calculate max to avoid skipping before offset
+         pos = (int) Math.max(skip, offset);
```
danberindei: For symmetry with the positive skip case, perhaps this would have been nicer:

         long skip = bytes.length - pos;
         if (skip > n)
            skip = n;

         pos += skip;
         return skip;

PR #4677, opened 2016-11-21
ISPN-6817 Remove multi pojo version test

galderz:


pruivo:
integrated! thanks @galderz

PR #4679, opened 2016-11-23
Ispn 7160/ispn 7158/spring test failures

slaskawi:
https://issues.jboss.org/browse/ISPN-7160
https://issues.jboss.org/browse/ISPN-7158

Good to be back :D

### DIFF
--- a/spring/spring4/spring4-common/src/test/java/org/infinispan/spring/session/InfinispanApplicationPublishedBridgeTCK.java
+++ b/spring/spring4/spring4-common/src/test/java/org/infinispan/spring/session/InfinispanApplicationPublishedBridgeTCK.java
@@ -10,6 +10,7 @@ import java.util.stream.Stream;
 
 import org.infinispan.spring.provider.SpringCache;
 import org.infinispan.spring.session.util.EventsWaiter;
+import org.infinispan.test.AbstractInfinispanTest;
 import org.springframework.context.ApplicationEvent;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.session.MapSession;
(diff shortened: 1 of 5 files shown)


spring/spring4/spring4-common/src/test/java/org/infinispan/spring/session/InfinispanApplicationPublishedBridgeTCK.java
```diff
@@ -64,6 +65,11 @@ public void testEventBridge() throws Exception {
 //      EventsWaiter.assertNumberOfEvents(() -> eventsCollector.getEvents(), SessionDestroyedEvent.class, 2, 10, TimeUnit.SECONDS);
    }
 
+   private void sleepOneSecond() {
+      long oneSecondSleep = System.currentTimeMillis() + 1000;
+      eventually(() -> System.currentTimeMillis() > oneSecondSleep);
```
danberindei: How is this better than a regular `sleep`?

danberindei: Oops, I should have read the bug description :)

Still, wouldn't it be possible to use a `TimeService`?

slaskawi: I tried to look into `TimeService` as you suggested but I couldn't find any nice way to hook in.

I could construct a loop using `isTimeExpired` but the end result would be the same. But maybe I missed something...

danberindei:
Integrated, thanks Sebastian!

PR #4680, opened 2016-11-23
ISPN-7227 Configure Hibernate Search to use JGroups4 config

vjuranek:
https://issues.jboss.org/browse/ISPN-7227


slaskawi:
Integrated thanks!

PR #4681, opened 2016-11-24
ISPN-7229 Marshalling arrays throws exception

rvansa:
* Replace ArrayExternalizers with array support in GlobalMarshaller
* If all the elements of array use the same externalizer, pack it only once
* Use ClassIdentifiers to efficiently encode array component type
* Change medium array boundary to cover maximum range

https://issues.jboss.org/browse/ISPN-7229


core/src/main/java/org/infinispan/marshall/core/InternalClassIds.java
```diff
@@ -0,0 +1,21 @@
+package org.infinispan.marshall.core;
+
+/**
+ * Identifiers for marshalling class name itself; this comes handy when marshalling arrays
+ * and the array component type is interface (and therefore it does not have its AdvancedExternalizer).
+ *
+ * This identifiers don't clash with {@link org.infinispan.commons.marshall.Ids} and therefore it can use the same range.
+ */
+public interface InternalClassIds {
```
galderz: I find the name confusing, shouldn't it be better `ClassNameIds`? Or if only using it for arrays, `ArrayClassIds`?

Why an interface? And why public? A default package final class with no constructor would be much better option for defining constants only used in `marshall.core` package.

galderz:
Any quantifiable improvements of this new array encoding?

core/src/main/java/org/infinispan/marshall/core/ClassIdentifiers.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.infinispan.configuration.global.GlobalConfiguration;
+import org.infinispan.container.entries.InternalCacheValue;
+
+public class ClassIdentifiers implements InternalClassIds {
```
galderz: Should be package private at least. The only public class in this package is `GlobalMarshaller`.

core/src/main/java/org/infinispan/marshall/core/ClassIdentifiers.java
```diff
@@ -0,0 +1,71 @@
...
+   private final static ClassIdentifiers INSTANCE = new ClassIdentifiers();
```
galderz: This is very dangerous. You have a static reference to a class that contains a reference to a collection whose values are classes. You're essentially creating a classloader leak in a managed environment.

Given the dangers, I'd avoid the singleton static class here. 

Also, if you instantiate it in GlobalMarshaller, make sure you clear it when it stops.

core/src/main/java/org/infinispan/marshall/core/ClassIdentifiers.java
```diff
@@ -0,0 +1,71 @@
...
+
+   // the hashmap is not changed after static ctor, therefore concurrent access is safe
+   private final Map<Class<?>, Integer> classToId = new HashMap<>();
```
galderz: `HashMap` sucks for this. We gave much efficient ways to have Class to Integer mappings, see what was done in `ClassToExternalizerMap`.

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
```diff
@@ -63,6 +102,7 @@
    private IdToExternalizerMap reverseInternalExts;
    private ClassToExternalizerMap externalExts;
    private IdToExternalizerMap reverseExternalExts;
+   private ClassIdentifiers classIdentifiers;
```
galderz: Needs to be cleared on stop().

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
```diff
@@ -370,8 +414,189 @@ private void writeNonNullableObject(Object obj, BytesObjectOutput out) throws IO
       }
    }
 
+   private void writeArray(Class<?> clazz, Object array, BytesObjectOutput out) throws IOException {
+      out.writeByte(ID_ARRAY);
+      Class<?> componentType = clazz.getComponentType();
+
+      int length = Array.getLength(array);
```
galderz: How expensive are these `java.lang.reflect.Array` calls?

core/src/main/java/org/infinispan/marshall/core/Primitives.java
```diff
@@ -177,13 +179,13 @@ private static void writeByteArray(byte[] obj, BytesObjectOutput out) {
...
+      } else if (len <= SMALL_ARRAY_MAX) {
          out.writeByte(ID_ARRAY_SMALL);
-         out.writeByte(len - 1);
+         out.writeByte(len - SMALL_ARRAY_MIN);
          out.write(obj, 0, len);
-      } else if (len <= ID_MEDIUM) {
+      } else if (len <= MEDIUM_ARRAY_MAX) {
          out.writeByte(ID_ARRAY_MEDIUM);
-         out.writeShort(len - 1);
+         out.writeShort(len - MEDIUM_ARRAY_MIN);
```
galderz: Why change this? Is there a real difference between writing a smaller or larger short? This is what `writeShort` does:

```
      int newcount = ensureCapacity(2);
      final int s = pos;
      bytes[s] = (byte) (v >> 8);
      bytes[s+1] = (byte) v;
      pos = newcount;
```

core/src/test/java/org/infinispan/marshall/VersionAwareMarshallerTest.java
```diff
@@ -659,6 +712,75 @@ public void readExternal(ObjectInput in) throws IOException, ClassNotFoundExcept
...
+   public static class PojoWithMultiExternalizer extends Pojo {
+
+      public PojoWithMultiExternalizer(int i, boolean b) {
+         super(i, b);
+      }
+
+      public static class Externalizer implements AdvancedExternalizer<Object> {
+         @Override
+         public Set<Class<?>> getTypeClasses() {
+            return Util.asSet(PojoWithMultiExternalizer.class, Thread.class);
```
galderz: `Thread.class`?

I think what you should be testing here is that a normal `Externalizer` (not advanced one, but annotated) work as expected.

core/src/main/java/org/infinispan/marshall/core/InternalClassIds.java
rvansa: `ClassNameIds` sounds fine. It is used for array component types, but it does not host array classes.

Right, it should not be public class. But interface allows me to omit the `static final`.

core/src/main/java/org/infinispan/marshall/core/ClassIdentifiers.java
rvansa: Yes, `HashMap` sucks, we need a hashmap with perfect hashing function. Though, with all due respect, the implementation in `ClassToExternalizerMap` sucks, too. Solving collisions by just going on the next spot in the table generates a sequences of often colliding elements. You even don't count how many collisions there are (to resize when you reach a threshold).
Seems I'll have to implement a perfect hashing table to comfort you here...

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
rvansa: Haven't micro-benchmarked that, and it's native code. I guess that compilers can optimize this pretty well.

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
```diff
@@ -370,8 +414,189 @@ private void writeNonNullableObject(Object obj, BytesObjectOutput out) throws IO
...
+         } else if (length <= Primitives.MEDIUM_ARRAY_MAX) {
+            flags = FLAG_ARRAY_MEDIUM;
+         } else {
+            flags = FLAG_ARRAY_LARGE;
+         }
+         Object firstElement = Array.get(array, 0);
+         if (firstElement != null) {
+            elementType = firstElement.getClass();
+         }
+         for (int i = 1; i < length; ++i) {
```
rvansa: I would assume that just checking all the elements is pretty cheap, compared to everything else that's going on there. However, assuming that and backing off is an interesting strategy, I'll think about that. Not sure if that's worth the effort, though.

Ad creating the array 'ourselves' - this is a generic marshalling code, I don't see how would you pass that hint to marshaller from the place where you create the array.

core/src/main/java/org/infinispan/marshall/core/Primitives.java
rvansa: The change is that you cover wider range of array lengths with just two-byte array length. The same reason as you've offsetted the small array by 1, you can offset medium array by 256 because you will never write a medium-size array with length < 256.

rvansa:
@galderz
> Any quantifiable improvements of this new array encoding?

My use case (and the place where I've spotted the NPE) was being able to marshall `InternalCacheValue[]` with instances with AdvancedExternalizer efficiently. Currently only certain array types were handled, and the externalizer would be looked up & encoded again and again. The least I have shortened the encoding of the externalizer identifiers.

rvansa:
And thanks for the pointers to class leaks, I haven't realized that.

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
rvansa: @galderz I was thinking about optimistically assuming that all instances are of the same type, and the most compact implementation (I can think of) would require one byte (1-0xFF) telling how many instances of single type there. The byte would be only reserved ahead, and then written once we:
a) reach end of array
b) reach different class instance
c) write 255 elements.
If we find out that 256th element is of the same type, we would set the single-type-length byte to 0 and reserve another byte to continue.
When hitting different class instance, I would back off to writing all the other elements without any compression, because this algorithm would add one byte per element for alternating sequences (sucha as when adding key on index `2*i` and value on `2*i + 1`) could be common.

The downside is one byte more per message, and increased complexity. Do you think it's worth it?

core/src/main/java/org/infinispan/marshall/core/ClassIdentifiers.java
galderz: Btw, the biggest danger here is if the collection keeps track of user classes. If they're JDK or Infinispan classes it should be fine, but to be on the safe side, I'd avoid any static strong references to classes.

core/src/main/java/org/infinispan/marshall/core/ClassIdentifiers.java
galderz: > Though, with all due respect, the implementation in ClassToExternalizerMap sucks, too. Solving collisions by just going on the next spot in the table generates a sequences of often colliding elements. You even don't count how many collisions there are (to resize when you reach a threshold).

To most important use of `ClassToExternalizerMap` is for internal externalizers, which we know exactly how many they are. I've sized that collection with the right size and load factor to avoid collisions as much as possible, so although you highlight a problem, in practical terms I don't expect it to happen.

> Seems I'll have to implement a perfect hashing table to comfort you here...

I didn't ask for perfect hashing table :). I'm just asking that given the elements added to this map and its use of, we can have something a bit more lightweight than a `HashMap`.

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
galderz: I'm not sure compaction should really be our goal. Compaction helps reduce payload sizes but if the code becomes to complex for the CPUs to be able to predict what will happen and get ahead of it, or for us to understand the code, I'm not sure about its merits, particularly if there's no hard data to back up one strategy versus the other.

danberindei: @rvansa -100 to add RLE encoding for this. In fact I would avoid any special encoding of the element type for arrays, just to make the code simpler. I don't expect 1 byte per element to matter in the grand scheme of things...

core/src/main/java/org/infinispan/marshall/core/Primitives.java
galderz: Oh right, the range has changed. I hadn't noticed that.

galderz:
Needs rebasing

galderz:
> And thanks for the pointers to class leaks, I haven't realized that.

I might have been to strong in my statement. The danger is particularly present when user classes are involved. Your current code might be fine, but I'd avoid the static part.

galderz:
> The least I have shortened the encoding of the externalizer identifiers.

^ Can you give us some numbers if possible?

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
rvansa: Direct CPU effect (+less externalizer lookups, -more complex branching)  is debatable, bandwidth-wise effect (less bytes) is obvious, though maybe neglible. Do you require me to write a JMH benchmark (where I could provide too favourable datasets), or shall we bring another opinion to this discussion?

rvansa:
>>  The least I have shortened the encoding of the externalizer identifiers.

> ^ Can you give us some numbers if possible?

You mean pre-PR and after? pre-PR the custom arrays could not be serialized at all. For Object[] with 100 elements containing PojoWithExternalizer, pre-PR is:
all-null: 104 bytes
all-set: 1004 bytes
99 set: 995 bytes
98 set: 986 bytes
With PR, this is (as expected):
all-null: 4 bytes
all-set: 509 bytes
99 set: 995 bytes
98 set: 986 bytes
Note that this uses external externalizer (4 bytes), not internal (1 byte).

rvansa:
@galderz Rebased && updated - mainly the classloader potential leaks, and replaced HashMap with perfect hash map implementation. I considered replacing ClassToExternalizerMap with it, too, or extracting the map functions in that class with LinearProbingHashMap (that would extend ArrayMap), but there are two differences that you could object:
a) you compare classes with `==` while this is not suitable for general purpose implementation
b) general purpose map uses `Object[]` array, because Java does not permit generic `new K[capacity]` and therefore the code needs to contains some casts.
Therefore, I'll leave replacing linear probing hashmap at your (later) discretion.

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
rvansa: @danberindei Ah, I actually haven't noticed that it was your comment, you both have yellowish icons :) Anyway, I am also against RLE. And regarding testability, I think that `testArrays` now covers most of the cases (for short arrays).

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
danberindei: Agree, it's definitely not going through reflection, just an `instanceof` check.

galderz:
Integrated, thx @rvansa

PR #4687, opened 2016-11-29
ISPN-7238 Reduce number of CompletableFuture allocations

danberindei:
https://issues.jboss.org/browse/ISPN-7238

I ran a test doing distributed reads using JMH, and I'm seeing 2990 bytes/op with this branch compared to 3198 bytes/op with master.


core/src/main/java/org/infinispan/interceptors/impl/AsyncInvocationStage.java
```diff
@@ -124,39 +117,146 @@ public InvocationStage toInvocationStage(InvocationContext newCtx, VisitableComm
...
+    */
+   protected boolean addHandler(InvocationComposeHandler composeHandler) {
+      synchronized (this) {
+         if (frozen) {
+            return false;
+         }
+         handlers[tail & mask] = composeHandler;
+         if (((tail + 1 - head) & mask) == 0) {
+            frozen = true;
+            return false;
```
rvansa: You've managed to add the handler, why are you returning `false`?

core/src/main/java/org/infinispan/interceptors/impl/AsyncInvocationStage.java
```diff
@@ -61,20 +52,30 @@ public Object get() throws Throwable {
 
    @Override
    public boolean isDone() {
+      lockHandlers();
```
rvansa: I wonder if you should lock the handlers also with `toCompletableFuture()`, but given it's unsafe to assume `toCompletableFuture().isDone()` => `this.isDone()`, it's probably not necessary.

danberindei: Yeah, it's not really necessary, because we can't protect from the user calling `s1 = stage.compose(...); s2 = stage.compose(...);` and having s2's handler receive the return value from s1's handler anyway. I first intended freezing to throw exceptions when compose was called again, but then I changed it to be used mostly instead of resizing the deque.

core/src/main/java/org/infinispan/interceptors/impl/AsyncInvocationStage.java
danberindei: Because the add didn't really succeed, as the deque supported only `mask` elements. I've fixed that, and it now supports `HANDLERS_DEQUE_CAPACITY` elements.

danberindei:
Updated with more comments @rvansa

core/src/main/java/org/infinispan/commands/ReplicableCommand.java
```diff
@@ -19,34 +19,46 @@
...
+    * The default implementation and {@link #perform(InvocationContext)} will be removed in future versions.
+    * </p>
+    *
+    * @since 9.0
     */
-   @Deprecated
-   default Object perform(InvocationContext ignored) throws Throwable {
-      return invoke();
+   default CompletableFuture<Object> invokeAsync() throws Throwable {
+      return CompletableFuture.completedFuture(perform(null));
```
rvansa: And btw, you shouldn't  change the order if you don't have to.

danberindei: I changed the order to put `invokeAsync` first, because that's the one I want commands to actually implement :)

I don't think I can preserve compatibility with existing custom commands and also prepare for removing `perform(InvocationContext)` without making every method `default`.

slaskawi:

PR #4692, opened 2016-11-30
ISPN-7072 Make Hibernate optional dependency

slaskawi:
https://issues.jboss.org/browse/ISPN-7072

### DIFF
--- a/all/embedded/pom.xml
+++ b/all/embedded/pom.xml
@@ -361,7 +361,7 @@
                            *
                         </Import-Package>
                         <DynamicImport-Package>*</DynamicImport-Package>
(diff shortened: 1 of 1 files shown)


Sanne:
why? The Jpa CacheStore requires it.  It would be better to make the JPA CacheStore optional, at least the error messages are understandable?

slaskawi:
Yes I also think making JPA Cache Store optional would be a better option.

Sanne:
N.B. I had to fix integration tests which had started failing with a very unfriendly error message as Hibernate was missing, while the JPA Cachestore was being started. (By "undriendly" I mean it was hard to guess what the problem was, I had to debug it to figure out which nested exception was being swallowed)

PR #4705, opened 2016-12-05
ISPN-7265 NullPointerException when preload data

pruivo:
https://issues.jboss.org/browse/ISPN-7265


tristantarrant:
Pushed to both master and 8.2.x

PR #4707, opened 2016-12-06
ISPN-7289 Stop applying the DONT_BUNDLE flag

danberindei:
https://issues.jboss.org/browse/ISPN-7289

I couldn't decide whether this should have the `Trivial` or the `Performance ACK required` label set, so I added none of them ;)

### DIFF
--- a/core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
+++ b/core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
@@ -248,10 +248,6 @@ public class CommandAwareRpcDispatcher extends MessageDispatcher {
                                                  long timeout) {
       RequestOptions options = new RequestOptions(mode, timeout);
       encodeDeliverMode(options, deliverOrder);
-      //some issues with the new bundler. put back the DONT_BUNDLE flag.
-      if (deliverOrder == DeliverOrder.NONE || mode != ResponseMode.GET_NONE) {
-         options.setFlags(Message.Flag.DONT_BUNDLE.value());
-      }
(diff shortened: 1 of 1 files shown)


slaskawi:
I think it should have both :D

rvansa:
I think we should wait with integrating until we explain the trouble with integrated triangle performance.

danberindei:
@rvansa I'm not sure if it's really a problem with the triangle, the results with TCP seemed to be the same as before, so it's more likely a problem with UDP + bundling. Anyway, at least on my machine, removing DONT_BUNDLE seems beneficial both with TCP and with UDP:

|Branch|JGroups config|Ops/s|
|---------|-------------------|--------|
|master|UDP default|26839|
|master|TCP default|30397|
|ISPN-7289|UDP default|31696|
|ISPN-7289|TCP default|37142|

rvansa:
@danberindei I don't say this causes any problems, though TBH I am lost if the DONT_BUNDLE is ignored or half-ignored in some version of Infinispan/JGroups. I just say that it will be easier to detect potential impact of this PR across all configurations in proper environment if we delay integration until the triangle is resolved.

danberindei:
@rvansa I'd call it "half-ignored": the network packets look the same w/ or w/out DONT_BUNDLE, but the dispatching of batches on the receiver is more efficient w/out DONT_BUNDLE. 

And since the triangle algorithm generates a lot more batches than the sync forwarding (aka before triangle), I don't think it's fair to try to improve the triangle performance without considering DONT_BUNDLE.

rvansa:
@danberindei Okay then - let's integrate this and see... Maybe @jmarkos could check it out isolated before it hits the fan :)

PR #4709, opened 2016-12-07
[8.2.x] ISPN-3391 Upgrade to JBoss Marshalling 2.0.0.Beta3

pferraro:
https://issues.jboss.org/browse/ISPN-3391

See https://issues.jboss.org/browse/WFLY-7746 for rationale.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -70,7 +70,7 @@
       <version.hibernate.search>5.6.0.Beta1</version.hibernate.search>
       <version.javax.cache>1.0.0</version.javax.cache>
       <version.cdi>1.0-SP4</version.cdi>
-      <version.jboss.marshalling>1.4.10.Final</version.jboss.marshalling>
+      <version.jboss.marshalling>2.0.0.Beta3</version.jboss.marshalling>
       <version.jboss.logging>3.3.0.Final</version.jboss.logging>
       <version.jgroups>3.6.7.Final</version.jgroups>
       <version.jta>1.0.1.Final</version.jta>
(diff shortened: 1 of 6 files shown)


galderz:
Looks good, same comment about the commit message.

pferraro:
@galderz Done.

gustavonalle:
@galderz @pferraro  This caused 80+ failures in the server test suite http://ci.infinispan.org/viewLog.html?buildId=46282&tab=buildResultsDiv&buildTypeId=Infinispan_82xHotspotJdk8&guest=1

gustavonalle:
Looks like the build is not available anymore, but the failures can be seen on 8.2.x, test is ```org.infinispan.server.test.client.hotrod.HotRodRemoteCacheIT```

galderz:
After doing some debugging I've noticed that server is built with JBoss Marshalling jar 2.0.0.Beta3, but the River marshaller jar still points to 1.4.10. That's probably the cause, but I'm unsure in which of all the pom(s) the change should be applied. I've pinged @tristantarrant.

gustavonalle:
@galderz You can try creating a module for the river on ```https://github.com/infinispan/infinispan/tree/8.2.x/server/integration/feature-pack/src/main/resources/modules/system/layers/base/org/jboss/marshalling```. Weird that ```master``` works with 1.4.1 though.

galderz:
I've just created [ISPN-7652](https://issues.jboss.org/browse/ISPN-7652). I think I've fixed the issue in 8.2.x branch, but I need to see if I can add a test that causes master to fail too.

PR #4710, opened 2016-12-08
Tests for ISPN-7300 and ISPN-7301 and extending query testsuite

Holmistr:
Extending query testsuite and tests for:
https://issues.jboss.org/browse/ISPN-7300
https://issues.jboss.org/browse/ISPN-7301

Tests testFullTextTermDoesntOccur and testFullTextTermRightOperandAnalyzed should be failing because of the above mentioned issues.


anistor:
I've disabled the failing test2 and integrated it. Thanks @Holmistr !

PR #4711, opened 2016-12-08
ISPN-7283 Make cluster and client listeners more fine grained

galderz:
https://issues.jboss.org/browse/ISPN-7283

This is an unpolished preview PR, please bear that in mind when reviewing.

To understand this initial solution, it's important to understand that I started by thinking this was just a problem of server/hotrod, where we were not being as fine grained as possible WRT client interests. So, I started by adding those listener interests to the Hot Rod protocol with a single byte, so that requires a HR protocol change. Then in server, I went for the tedious option of having multiple listener impls depending on which annotation combo I was interested. I knew this wouldn't scale very well if more listener events were added in the future, but I just wanted to get something going...

Then, I realised that this was enough because the cluster listener is installed around, RemoteClusterListener, listens for everything. So, not only the changes in server/hotrod were needed, but core needed changing too.

So, then I started looking into how I could have listener class but have a way to define which, of all the annotations present in that class, I was interested in registering the listener for. This led to the creation of `addFilteredListener` method. I think this is the best concept going forward, rather than having to create N possible listener classes depending on which annotations we're interested in. This of course means that a lot of the new code in server/hotrod can probably go if I can simply call `addFilteredListener` from there.

So, if everyone agrees with my approach, a key aspect here is defining where and how this method to add filtered listener should look like. Right now it looks like this:

```
   /**
    * Add a listener limiting cache entry specific callbacks only to those
    * annotations that are present in the annotation filter collection.
    */
   <C> void addFilteredListener(Object listener,
         CacheEventFilter<? super K, ? super V> filter,
         CacheEventConverter<? super K, ? super V, C> converter,
         Set<Class<? extends Annotation>> filterAnnotations);
```

So, the API is slightly clunky right now because those filter annotations can only filter `CacheEntryCreated`, `CacheEntryModified`, `CacheEntryExpired` and `CacheEntryRemoved` annotations. So, what this means if a listener has methods with all these four annotations, and I pass a filterAnnotation containing `CacheEntryCreated` and `CacheEntryModified`, then it will only listen for those. However, a key aspect here is that if the listener had other non cache-entry specific callbacks, e.g. `ViewChanged` or `TransactionCompleted`, I wanted to live those still there (this was actually needed to get `RemoteClusterListener` to work properly).

If anyone has other ideas of how this API should look like I'd love to hear. Also, to be able to easily consume this API from server/hotrod, it might be better if it lived in `FilteringListenable`. Would users find this API useful? Otherwise, `CacheNotifier` instance can always be retrieved via `AdvancedCache.getComponentRegistry(...)`.

Finally, fixes in `CacheNotifierImpl` and `AbstractListenerImpl` contain a lot of code duplication which would need to be sorted out. The code is a bit messy in these classes and in the end, having tried multiple things and debugged odd behaviours, I found the safest thing to do but was to copy/paste some methods and adjust them. Not great :|

### DIFF
--- a/core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifier.java
+++ b/core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifier.java
@@ -10,11 +10,15 @@ import org.infinispan.factories.scopes.Scopes;
 import org.infinispan.metadata.Metadata;
 import org.infinispan.notifications.ClassLoaderAwareFilteringListenable;
 import org.infinispan.notifications.ClassLoaderAwareListenable;
+import org.infinispan.notifications.cachelistener.filter.CacheEventConverter;
+import org.infinispan.notifications.cachelistener.filter.CacheEventFilter;
 import org.infinispan.partitionhandling.AvailabilityMode;
 import org.infinispan.topology.CacheTopology;
 import org.infinispan.transaction.xa.GlobalTransaction;
 
+import java.lang.annotation.Annotation;
 import java.util.Collection;
+import java.util.Set;
 
 /**
  * Public interface with all allowed notifications.
@@ -110,4 +114,14 @@ public interface CacheNotifier<K, V> extends ClassLoaderAwareFilteringListenable
     * @param typeConverter the converter instance; can be {@code null}
     */
    void setTypeConverter(TypeConverter typeConverter);
+
+   /**
+    * Add a listener limiting cache entry specific callbacks only to those
+    * annotations that are present in the annotation filter collection.
+    */
+   <C> void addFilteredListener(Object listener,
+         CacheEventFilter<? super K, ? super V> filter,
+         CacheEventConverter<? super K, ? super V, C> converter,
+         Set<Class<? extends Annotation>> filterAnnotations);
+
 }
--- a/core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifierImpl.java
+++ b/core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifierImpl.java
@@ -44,6 +44,7 @@ import org.infinispan.notifications.impl.ListenerInvocation;
 import org.infinispan.partitionhandling.AvailabilityMode;
 import org.infinispan.remoting.transport.Address;
 import org.infinispan.remoting.transport.jgroups.SuspectException;
+import org.infinispan.security.Security;
 import org.infinispan.topology.CacheTopology;
 import org.infinispan.transaction.xa.GlobalTransaction;
 import org.infinispan.util.logging.Log;
@@ -54,6 +55,9 @@ import javax.transaction.Transaction;
 import javax.transaction.TransactionManager;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.*;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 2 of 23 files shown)


galderz:
Btw, I'm only submitting 8.2.x for now until we have agreed on all the points mentioned below. Once the solution is agreed I'll port it over to master.

core/src/main/java/org/infinispan/notifications/impl/AbstractListenerImpl.java
```diff
@@ -207,6 +215,10 @@ protected boolean validateAndAddListenerInvocations(Object listener, AbstractInv
                   builder.setMethod(m);
                   builder.setAnnotation(annotationClass);
                   L invocation = builder.build();
+
+                  if (log.isTraceEnabled())
```
wburns: Might as well use `trace`

core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifier.java
```diff
@@ -110,4 +114,14 @@ void notifyCacheEntryPassivated(K key, V value, boolean pre,
     * @param typeConverter the converter instance; can be {@code null}
     */
    void setTypeConverter(TypeConverter typeConverter);
+
+   /**
+    * Add a listener limiting cache entry specific callbacks only to those
+    * annotations that are present in the annotation filter collection.
+    */
+   <C> void addFilteredListener(Object listener,
```
wburns: Agreed, this definitely needs to move to the `FilteringListenable` interface

core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifierImpl.java
```diff
@@ -940,6 +955,252 @@ public void addListener(Object listener, KeyFilter<? super K> filter) {
       addListener(listener, filter, converter, null);
    }
 
+   @Override
+   public <C> void addFilteredListener(Object listener, CacheEventFilter<? super K, ? super V> filter,
```
wburns: Yeah I think we can condense this down to be a bit to not be duplicated.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientEventSenders.scala
```diff
@@ -0,0 +1,392 @@
...
+        val delegate = new StatelessClientEventSender(ch, listenerId, version, eventType)
+        createStatelessCompatibilityClientEventSender(delegate, HotRodTypeConverter(compatibility.marshaller()), listenerInterests)
+      case (true, true) =>
+        val delegate = new StatelessClientEventSender(ch, listenerId, version, eventType)
+        createStatefulCompatibilityClientEventSender(delegate, HotRodTypeConverter(compatibility.marshaller()), listenerInterests)
+    }
+  }
+
+  def createStatelessClientEventSender(ch: Channel, listenerId: Bytes, version: Byte, eventType: ClientEventType, listenerInterests: Byte) = {
+    listenerInterests match {
```
wburns: Is this still required to have the various `*Sender` classes?  Can't we just have 1 class with all of the remote notification annotations then apply the filter?

galderz:
> We still would want to clean up some of the duplication. And it seems like the server changes can be reduced significantly as they don't take into account the new Set argument

^ See description, I already explained that there's some duplication due to the way I went around to fixing it. It should indeed be sorted out.

server/hotrod/src/main/scala/org/infinispan/server/hotrod/ClientEventSenders.scala
galderz: That's the plan, see description which describes that this is lingering code from the initial solution. I should indeed go away once filter has been applied.

galderz:
I'll go ahead and sort out missing pieces.

galderz:
I've polished this PR and submitted #4721, so closing this one.

PR #4712, opened 2016-12-08
ISPN-1947 ISPN-7299 Fixed staging plugin

slaskawi:
https://issues.jboss.org/browse/ISPN-1947
https://issues.jboss.org/browse/ISPN-7299

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -95,18 +95,22 @@
       <version.hibernate_dep.javassist>3.20.0-GA</version.hibernate_dep.javassist>
 
       <uberjar.deps.optional>true</uberjar.deps.optional>
+
+      <jboss.releases.repo.id>jboss-releases-repository</jboss.releases.repo.id>
+      <jboss.releases.nexus.url>https://repository.jboss.org/nexus</jboss.releases.nexus.url>
+      <jboss.releases.repo.url>${jboss.releases.nexus.url}/service/local/staging/deploy/maven2/</jboss.releases.repo.url>
+      <jboss.snapshots.repo.id>jboss-snapshots-repository</jboss.snapshots.repo.id>
+      <jboss.snapshots.repo.url>https://repository.jboss.org/nexus/content/repositories/snapshots/</jboss.snapshots.repo.url>
    </properties>
 
    <distributionManagement>
       <repository>
-         <id>jboss-releases-repository</id>
-         <name>JBoss Release Repository</name>
-         <url>https://repository.jboss.org/nexus/service/local/staging/deploy/maven2/</url>
+         <id>${jboss.releases.repo.id}</id>
+         <url>${jboss.releases.repo.url}</url>
       </repository>
       <snapshotRepository>
-         <id>jboss-snapshots-repository</id>
-         <name>JBoss Snapshot Repository</name>
-         <url>https://repository.jboss.org/nexus/content/repositories/snapshots/</url>
+         <id>${jboss.snapshots.repo.id}</id>
+         <url>${jboss.snapshots.repo.url}</url>
       </snapshotRepository>
    </distributionManagement>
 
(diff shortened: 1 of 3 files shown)


gustavonalle:
Can we maybe test that the plugin is really working or do we need to wait for the next release to find out? 
Maybe use a nexus docker image and do a release against it.

slaskawi:
I actually tested it with server build: https://origin-repository.jboss.org/nexus/content/repositories/public-jboss/org/infinispan/server/infinispan-server-build/9.0.0.Beta1/

It seems to work fine. The biggest problem with testing this thing is that Nexus provides staging repositories in Enterprise version (not the Community). I think I've already used all my emails for 30-day
testing licenses :D Let me ask them if I can request another one or if I can create a testing repo in our Nexus...

slaskawi:
I did some testing with Nexus staging instance. It seems to work fine. 

But the final test would be during the release.

bom/pom.xml
```diff
@@ -536,6 +545,18 @@
...
       </pluginManagement>
+      <plugins>
+         <plugin>
+            <groupId>org.sonatype.plugins</groupId>
+            <artifactId>nexus-staging-maven-plugin</artifactId>
+            <extensions>true</extensions>
+            <configuration>
+               <nexusUrl>${jboss.releases.nexus.url}</nexusUrl>
+               <serverId>${jboss.releases.repo.id}</serverId>
+               <autoReleaseAfterClose>true</autoReleaseAfterClose>
```
danberindei: Doesn't seem like this really belongs in a BOM that is imported by other projects...

slaskawi: Well yes and no...

If you import this BOM, you won't trigger this plugin on your project (note it's in `/build/plugins`). The problem is when you use the BOM as your parent, than yes - the you will run the staging plugin when you invoke `mvn deploy`. But please note that today we have the same problem - if you use this BOM as a parent of your project you will use JBoss Repositories from `/distributionManagement` since they are also inherited. 

So yes, it's wrong but our current approach is also wrong.

Now diving into more technical things - this plugin needs to hook into `default-deploy` execution. This is done by `<extensions>true</extensions>` in the configuration. Now if we don't like this approach, we will need to disable the default Deploy plugin (adding an execution with goal `none`) and invoke deployment separately using command line. Something like this: `mvn install && mvn nexus-staging-maven-plugin:deploy -D...<all required parameters to get this thing work>`. This probably needs some experiments since I've never done it before.

As I wrote in my email on internal mailing list - I believe we should use Nexus Staging and automate the release process as a whole. That's a much better approach than automatically closing the staging repository here.

slaskawi:
I've just tried to remove the Staging plugin from BOM and Server BOM and invoke:

```
mvn clean install && mvn org.sonatype.plugins:nexus-staging-maven-plugin:1.6.7:deploy -DnexusUrl='https://repository.jboss.org/nexus' -DserverId='jboss-releases-repository' -DautoReleaseAfterClose=true -DskipLocalStaging=true
```

Unfortunately I hit this: http://stackoverflow.com/a/38148135

So polluting BOMs with Staging is the only way that worked for me...

bom/pom.xml
tristantarrant: the infinispan-bom should be user-consumable, while infinispan-parent is our actual parent, where we define plugin configs, lifecycle, etc. If polluting the bom is the only way (I hope not) then everything should be relegated in a release-specific profile (e.g. distribution)

danberindei: @slaskawi I wouldn't really want to invoke `nexus-staging-maven-plugin:deploy` separately, my proposal would be to make all our modules inherit from `infinispan-parent` and only import `wildfly-parent` instead of inheriting from it. I'm not sure how easy that would be, but I'm thinking we can always copy the stuff we need from `wildfly-parent` into our own POMs...

slaskawi: Yes, but than `infinispan-bom` would inherit from parent.... ouch!

slaskawi:
Moved everything into `distribution` profile as @tristantarrant requested.

bom/pom.xml
danberindei: I forgot that `infinispan-bom` needs to have the staging plugin too... then there's really no way to avoid putting the staging plugin configuration in there.

slaskawi:
Together with @tristantarrant we agreed that closing staging automatically is not the best approach. What we would like to to instead is to release our stuff, run automated tests on staging repository and close it when they are OK.

Closing this one in favor of https://github.com/infinispan/infinispan/pull/4729

Sanne:
I've commented on https://issues.jboss.org/browse/ISPN-1947 . Seems we're not on the same page on the functionality provided by Nexus.

Please reconsider?

slaskawi:
Yes we are. The thing is that Infinispan will not always be the last released component. Having said that it is not wise to close the staging repo in poms. A better idea is to do it in TeamCity when triggering release job (via Nexus REST API).

Sanne:
ah, ic. Good idea to not do it then!

PR #4713, opened 2016-12-08
ISPN-7297 java.lang.IllegalStateException on TriangleAckInterceptor.java

pruivo:
Caused by a miss configuration in RequestOptions

https://issues.jboss.org/browse/ISPN-7297

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
@@ -189,7 +189,8 @@ public class TriangleDistributionInterceptor extends NonTxDistributionIntercepto
          return invokeNext(context, command);
       }
       final CacheTopology topology = checkTopologyId(command);
-      DistributionInfo distributionInfo = new DistributionInfo(command.getKey(), topology.getWriteConsistentHash(), rpcManager.getAddress());
+      DistributionInfo distributionInfo = new DistributionInfo(command.getKey(), topology.getWriteConsistentHash(),
+            rpcManager.getAddress());
 
       switch (distributionInfo.ownership()) {
          case PRIMARY:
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
```diff
@@ -232,24 +237,33 @@ private BasicInvocationStage primaryOwnerWrite(InvocationContext context, DataWr
...
-            if (rCtx.isOriginLocal() && (isSynchronous(dwCommand) || dwCommand.isReturnValueExpected())) {
-               commandAckCollector.create(id, rv, distributionInfo.owners(), dwCommand.getTopologyId());
-            }
-            if (trace) {
-               log.tracef("Command %s send to backup owner %s.", dwCommand.getCommandInvocationId(), backupOwners);
-            }
+         //protect with the shared lock.
+         //it was possible the topology change before the collector was created. the collector would be initializer
+         //with an old members that would never send the ack back resulting in timeout.
+         stateTransferLock.acquireSharedTopologyLock();
```
danberindei: I would prefer checking the topology again after the collector was created, and retry if it changed, to avoid the contention on the number of readers in the shared lock.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -777,7 +777,7 @@ private void sendToAll(ReplicableCommand rpcCommand, DeliverOrder deliverOrder)
          AnycastAddress anycastAddress = new AnycastAddress();
          dispatcher.sendMessage(anycastAddress, buffer, options);
       } else {
-         dispatcher.castMessage(null, buffer, options);
+         dispatcher.castMessage(null, buffer, options.anycasting(false));
```
danberindei: Looks superfluous. the default is already `false`. It actually made me think maybe we're setting it to `true` earlier, and that we shouldn't do that :)

core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
pruivo: on other hand, you are creating a Collector to be removed later. I'm not sure if it has any impact in performance

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
pruivo: I know. I just want to be sure in case if one of us in the future change the default for some reason.

core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
pruivo: so, should I change it?

pruivo:
@danberindei updated

danberindei:
Integrated, thanks Pedro!

PR #4714, opened 2016-12-08

pruivo:
* Cancel timeout task when the operation is completed.
* Reduce the number of allocated completable future.

Linked to triangle JIRA to keep track of the fixes.
https://issues.jboss.org/browse/ISPN-6919

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/TriangleAckInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/TriangleAckInterceptor.java
@@ -1,12 +1,8 @@
 package org.infinispan.interceptors;
 
 import java.util.Collection;
-import java.util.Iterator;
-import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


danberindei:
Similar to `RspListFuture`, we could put the `timeoutFuture` in the collector itself, and the collector can implement `BiConsumer` and `Callable` so that it doesn't create other lambda instances. (`SingleRspFuture` doesn't actually implement `BiConsumer`, but it should.)

I also thinking about starting the timeout task when the RPC is sent, not when we get back to the `TriangleAckInterceptor`, because there could be a potentially lengthy store write in between them.

Still, the fix looks good, so I'm integrating it.

pruivo:
@danberindei are you integrate it or should I perform the change? :)

pruivo:
ah, I've noticed that it is already in master. closing the PR.

danberindei:
Sorry, forgot to close it. I already have a PR open to reduce the number of allocated CompletableFutures and I need to change `SingleResponseFuture` too, so I can make the changes there.

pruivo:
@danberindei, if you want I've been profiling with the change:
https://github.com/infinispan/infinispan/compare/master...pruivo:t_triangle_perf

I can make a PR or you can fetch the diff and apply in your branch. what do you prefer?

PR #4715, opened 2016-12-08
ISPN-7301 Right operand in term query is not being analyzed

anistor:
https://issues.jboss.org/browse/ISPN-7301


tristantarrant:
Merged to master, thanks

PR #4716, opened 2016-12-09
Add quickstarts classes to whitelist of ExternallyMarshalable

Holmistr:
We have Selenium tests for quickstarts, which are started via maven-surefire-plugin, which enables assertions. Due to this, the assert fails on simple quickstart classes. In order not to complicate quickstarts with ExternalPojo or anything other configuration, I'm adding this package to whitelist.


galderz:
Thanks @Holmistr

PR #4722, opened 2016-12-13

pruivo:
Make collector implement Callable and BiConsumer to avoid creating lambdas

https://issues.jboss.org/browse/ISPN-6919


danberindei:
Integrated, thanks Pedro!

PR #4725, opened 2016-12-13
ISPN-7309 Exception cause might be null

galderz:
https://issues.jboss.org/browse/ISPN-7309


wburns:
Integrated into master, Thanks @galderz !

PR #4728, opened 2016-12-14
ISPN-7283 Finer grained remote event listeners

galderz:
https://issues.jboss.org/browse/ISPN-7283

* Cluster listener installed for remote listener is now more fine
  grained, only adding listener callbacks for those events that the
  client is interested in.
* To find out which callbacks the client is interested in, the Hot Rod
  protocol has been changed to add a byte indicating the event types
  interested in.
* Added addFilteredListener which allows the caller to decide which
  cache-entry specific events the listener is interested in.

### DIFF
--- a/server/hotrod/src/main/java/org/infinispan/server/hotrod/Decoder2x.java
+++ b/server/hotrod/src/main/java/org/infinispan/server/hotrod/Decoder2x.java
@@ -405,12 +405,20 @@ class Decoder2x implements VersionedDecoder {
                requestCtx.setUseRawData(useRawData);
 
                buffer.markReaderIndex();
-               out.add(hrCtx);
 
                return requestCtx;
             }).isPresent()) {
                return;
             }
+            if (Constants.isVersionPost25(header.version)) {
+               int listenerInterests = ExtendedByteBufJava.readMaybeVInt(buffer);
+               if (listenerInterests == Integer.MIN_VALUE)
+                  return;
+
+               requestCtx.setListenerInterests(listenerInterests);
+               buffer.markReaderIndex();
+            }
+            out.add(hrCtx);
             break;
          case REMOVE_CLIENT_LISTENER:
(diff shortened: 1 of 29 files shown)


server/hotrod/src/main/java/org/infinispan/server/hotrod/Decoder2x.java
```diff
@@ -411,6 +411,13 @@ public void customReadKey(HotRodHeader header, ByteBuf buffer, CacheDecodeContex
             }).isPresent()) {
                return;
             }
+            if (Constants.isVersionPost25(header.version)) {
+               int listenerInterests = ExtendedByteBufJava.readMaybeVInt(buffer);
+               if (listenerInterests == Integer.MIN_VALUE)
+                  return;
+
+               requestCtx.setListenerInterests(listenerInterests);
+            }
```
wburns: We need to move `out.add(hrCtx);` from line 408 to after this if block so that if we don't have all bytes it works properly.  We also would need to mark the reader index before that.

I believe the tests should have caught this unless they aren't using the `SingleByteFrameDecoderChannelInitializer` correctly.

galderz: Hmmmmm, not sure exactly how you expect the code to look. It's quite confusing the way it currently looks. Can you paste the final code somewhere or send me a PR?

galderz: Is it like [this](https://gist.github.com/galderz/ee907261b68fdebe452b193e075e942f)? 

Also, I didn't see any tests failing.

wburns: Seems good to me.  Yeah like I mentioned it seems like the tests aren't using it properly atm.  Either that or we would have gotten failures on another call.

galderz:
@wburns Should be fixed now

wburns:
Looks good to me, pulling...

wburns:

PR #4732, opened 2016-12-16
ISPN-6802 Micro-optimizations for read operations

danberindei:
https://issues.jboss.org/browse/ISPN-6802

* Don't try to commit context entries for read commands
* Don't try to release locks for read commands
* Remove ClearCommand class check
* Avoid duplicate thread-local lookup for tx reads without explicit transaction
* Use SingleKeyNonTxInvocationContext for remote gets even in tx caches
* Cache transactional attribute as field in CacheImpl
* Cache configuration attributes in interceptors
* Simplify SingleKeyNonTxInvocationContext entry lookup
* JGroupsTransport should compute the address eagerly
* IsMarshallableInterceptor shouldn't do anything without an asynchronous store


tristantarrant:
Pushed to master with a tiny fix in HotRodMarshallingTest

rvansa:
@danberindei Hi Dan, where are the L1 entries committed to DC, then?

PR #4734, opened 2016-12-16
ISPN-6803 Pre-compute bitsets for each flag

danberindei:
https://issues.jboss.org/browse/ISPN-6803


rvansa:
Have you checked generated assembly what the original `hasFlag(Flag.SOME)` really did? I would assume that in C2 this is already inlined to a constant.

danberindei:
@rvansa Yes, this came from looking at the generated assembly (and one of Shipilev's blog posts). C2 can only replace `static final` fields with a constant, not regular `final` fields (like the ordinal in an enum).

danberindei:
Rebased + updated to replace some usages of `EnumUtil.hasEnum()` and `EnumUtil.setEnum()` with their `FlagBitSets`-based alternatives.

tristantarrant:
In it goes. Thanks

PR #4674, opened 2016-11-21
ISPN-449 Hot Rod streaming ops

tristantarrant:
https://issues.jboss.org/browse/ISPN-449

This implements put/putifabsent/get/replace ops which instead of taking a value return an OutputStream/InputStream
This PR also includes some refactoring to the server which makes the code much less horrible in a few places. I also got 3%-5% perf improvements in some runs compared to master on my local machine, but I haven't been able to get consistent enough numbers, so take this with a grain of salt.

PR #4735, opened 2016-12-16
Functional tests for Streaming API

zhostasa:
* two new containers in arquillian configuration
 * new cache for default-dist.xml config
 * HotRodRemoteStreamingIT test class

Set of tests for #4674


tristantarrant:
Pushed to master, thanks

PR #4737, opened 2016-12-20
Resolve SQL Server Errors

ryanemerson:
https://issues.jboss.org/browse/ISPN-7292
https://issues.jboss.org/browse/ISPN-3618


tristantarrant:
@jmarkos can you verify that this works ?
I guess we could install SQL Server on the CI machines, since there is now a version available for Linux: https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-setup-red-hat

ryanemerson:
@tristantarrant The SQL server version available for linux will be good for checking upsert and other queries, however we won't be able to use it to check ISPN-3618 as that is only applicable to SQL Server pre 2016.  For testing pre 2016 versions I had to setup a Windows VM.

tristantarrant:
Pushed to master, thanks

PR #4741, opened 2016-12-21
ISPN-7322 Improve triangle algorithm: ordering by segment

pruivo:
https://issues.jboss.org/browse/ISPN-7322

Send the backup updates unordered (OOB) and the ordering is done at Infinispan level.

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/marshall/MarshallUtil.java
+++ b/commons/src/main/java/org/infinispan/commons/marshall/MarshallUtil.java
@@ -1,5 +1,7 @@
 package org.infinispan.commons.marshall;
 
+import static org.infinispan.commons.util.CollectionFactory.computeCapacity;
+
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
@@ -63,9 +65,55 @@ public class MarshallUtil {
       if (size == NULL_VALUE) {
          return null;
       }
-      final T map = Objects.requireNonNull(builder, "MapBuilder must be non-null").build(size);
+      final T map = Objects.requireNonNull(builder, "MapBuilder must be non-null").build(computeCapacity(size));
       for (int i = 0; i < size; i++) //noinspection unchecked
          map.put((K) in.readObject(), (V) in.readObject());
+      return map;
+   }
+
+   /**
+    * Marshall the {@code map} to the {@code ObjectOutput}.
+    * <p>
+    * {@code null} maps are supported.
+    *
+    * @param map {@link Map} to marshall.
+    * @param out {@link ObjectOutput} to write. It must be non-null.
+    * @param <K> Key type of the map.
+    * @param <V> Value type of the map.
+    * @param <T> Type of the {@link Map}.
+    * @throws IOException If any of the usual Input/Output related exceptions occur.
+    */
+   public static <K, V, T extends Map<K, V>> void marshallMap(T map, ElementWriter<K> keyWriter, ElementWriter<V> valueWrite, ObjectOutput out) throws IOException {
+      final int mapSize = map == null ? NULL_VALUE : map.size();
+      marshallSize(out, mapSize);
+      if (mapSize <= 0) return;
+
+      for (Map.Entry<K, V> me : map.entrySet()) {
+         keyWriter.writeTo(out, me.getKey());
+         valueWrite.writeTo(out, me.getValue());
+      }
+   }
+
+   /**
+    * Unmarshall the {@link Map}.
+    * <p>
+    * If the marshalled map is {@link null}, then the {@link MapBuilder} is not invoked.
+    *
+    * @param in      {@link ObjectInput} to read.
+    * @param builder {@link MapBuilder} to create the concrete {@link Map} implementation.
+    * @return The populated {@link Map} created by the {@link MapBuilder} or {@code null}.
+    * @throws IOException            If any of the usual Input/Output related exceptions occur.
+    * @throws ClassNotFoundException If the class of a serialized object cannot be found.
+    * @see #marshallMap(Map, ElementWriter, ElementWriter, ObjectOutput)
+    */
+   public static <K, V, T extends Map<K, V>> T unmarshallMap(ObjectInput in, ElementReader<K> keyReader, ElementReader<V> valueReader, MapBuilder<K, V, T> builder) throws IOException, ClassNotFoundException {
+      final int size = unmarshallSize(in);
+      if (size == NULL_VALUE) {
+         return null;
+      }
+      final T map = Objects.requireNonNull(builder, "MapBuilder must be non-null").build(computeCapacity(size));
+      for (int i = 0; i < size; i++) //noinspection unchecked
+         map.put(keyReader.readFrom(in), valueReader.readFrom(in));
       return map;
    }
 
@@ -301,7 +349,7 @@ public class MarshallUtil {
    }
 
    /**
-    * Same as {@link #unmarshallArray(ObjectInput, ArrayBuilder)} but specialzed for byte array.
+    * Same as {@link #unmarshallArray(ObjectInput, ArrayBuilder)} but specialized for byte array.
     * <p>
     * No {@link ArrayBuilder} is necessary.
     *
--- a/core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
+++ b/core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
@@ -1,0 +1,149 @@
+package org.infinispan.distribution;
+
+import java.util.Iterator;
+import java.util.Map;
+
+import org.infinispan.statetransfer.OutdatedTopologyException;
+import org.infinispan.topology.CacheTopology;
+
+import net.jcip.annotations.GuardedBy;
+
+/**
+ * It manages the order of updates from the primary owner to backup owner.
+ * <p>
+ * It depends on the cache topology id. The primary owner assigns the sequence number to the backup command and then
+ * sends it to the backup owner. In the backup owner, the command awaits until it is its turn to be executed.
+ * <p>
+ * If the command topology id does not match, it throws an {@link OutdatedTopologyException}.
+ * <p>
+ * The sequence order starts with 1 and it is per segment based. This allows segments to be updated concurrently.
+ *
+ * @author Pedro Ruivo
+ * @since 9.0
+ */
+public class TriangleOrderManager {
+
+   private final TriangleSequencer[] sequencers;
+   private volatile CacheTopology currentCacheTopology;
+
+   public TriangleOrderManager(int segments) {
+      TriangleSequencer[] triangleSequencers = new TriangleSequencer[segments];
+      for (int i = 0; i < segments; ++i) {
+         triangleSequencers[i] = new TriangleSequencer();
+      }
+      sequencers = triangleSequencers;
+   }
+
+   public long next(int segmentId, final int commandTopologyId) {
+      checkTopologyId(commandTopologyId);
+      try {
+         return getNext(segmentId, commandTopologyId);
+      } finally {
+         //check if topology didn't change in the meanwhile
+         checkTopologyId(commandTopologyId);
+      }
+   }
+
+   public void next(Map<Integer, Long> segmentsIds, final int commandTopologyId) {
+      checkTopologyId(commandTopologyId);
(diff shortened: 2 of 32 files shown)


Holmistr:
run performance tests please

Holmistr:
@pruivo please ignore the previous comment. I've set up triggering performance tests via GitHub comment and I would like to use your PR as a first "real" run. I will send an email about it on mailing list if it successfully passes.

Holmistr:
run performance tests please

pruivo:
@Holmistr cool! :) 
can I trust the results post by the "bot"?

Holmistr:
@pruivo I'm just writing an email to infinispan mailing list, give me 30 minutes, everything should be explained there :)

core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
```diff
@@ -0,0 +1,187 @@
...
+         } else {
+            //this topology is higher than the command topology id.
+            //another topology was installed. this command will fail with OutdatedTopologyException.
+            throw OutdatedTopologyException.getCachedInstance();
+         }
+      }
+
+      private void setTopologyId(int topologyId) {
+         this.topologyId = topologyId;
+         this.sequenceNumber = 1;
```
danberindei: Needs to be synchronized.

core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
```diff
@@ -0,0 +1,187 @@
...
+   }
+
+   public void removeOldSegments(Collection<Integer> ownedSegments) {
+      if (ownedSegments.isEmpty()) {
+         //no segments?
+         receiverSegmentMap.clear();
+         senderSegmentMap.clear();
+      } else {
+         receiverSegmentMap.keySet().retainAll(ownedSegments);
+         senderSegmentMap.keySet().retainAll(ownedSegments);
```
danberindei: A new command with that segment will recreate the `ReceiverSequencer` anyway, even if it has a lower topology id, so I'm not sure there's any point in removing segments not owned.

Same for `SenderSequencer`, I'd rather just increment the topology id and reset the sequence number.

core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
```diff
@@ -0,0 +1,187 @@
...
+      @GuardedBy("this")
+      private long nextSequenceNumber = 1;
+
+      private ReceiverSequencer(int topologyId) {
+         setTopologyId(topologyId);
+      }
+
+      private synchronized void deliver(int commandTopologyId, long sequenceNumber) {
+         if (topologyId == commandTopologyId && nextSequenceNumber == sequenceNumber) {
+            nextSequenceNumber++;
```
danberindei: I feel like there should be a way to trigger the next command directly here, instead of having to call `checkForReadyTasks()` after, because that method has to iterate through all the tasks in the queue and it's never going to run more than 1 task per segment.

OTOH it's probably good that `checkForReadyTasks()` can pick a command for any segment, and one segment's  commands can't monopolize a remote executor thread forever, so I'm not pushing for a change yet.

core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
```diff
@@ -0,0 +1,187 @@
...
+         }
+      }
+
+      private synchronized boolean isNext(int commandTopologyId, long sequenceNumber) {
+         if (topologyId == commandTopologyId) {
+            return nextSequenceNumber == sequenceNumber;
+         } else if (topologyId < commandTopologyId) {
+            //update topology. this command will be the first
+            setTopologyId(commandTopologyId);
+            return nextSequenceNumber == sequenceNumber;
```
danberindei: This branch doesn't seem right. We already check that the command's topology id is the same as the current topology id in `next()` (a bit excessively, I would say), and I assume we also rely on `CheckTopologyAction` to run the command only after the new topology was installed. So I'd rather throw an exception if we got at this point with `topologyId < commandTopologyId`.

core/src/main/java/org/infinispan/remoting/inboundhandler/NonTotalOrderTxPerCacheInboundInvocationHandler.java
```diff
@@ -125,17 +123,18 @@ public boolean isReady() {
       }
    }
 
-   private ReadyAction createReadyAction(int topologyId, TransactionalRemoteLockCommand command) {
-      if (command.hasSkipLocking()) {
+   private <T extends TransactionalRemoteLockCommand & ReplicableCommand> ReadyAction createReadyAction(int topologyId,
+         T replicableCommand) {
```
danberindei: You could also change `RemoteLockCommand` to extend `ReplicableCommand`...

core/src/main/java/org/infinispan/remoting/inboundhandler/action/ActionState.java
```diff
@@ -14,20 +14,20 @@
...
       this.command = command;
       this.commandTopologyId = commandTopologyId;
       this.timeout = timeout;
    }
 
-   public final RemoteLockCommand getCommand() {
-      return command;
+   public final <T extends ReplicableCommand> T getCommand() {
+      //noinspection unchecked
+      return (T) command;
```
danberindei: Same here, you wouldn't need the generic parameter if `RemoteLockCommand` extended `ReplicableCommand`.

core/src/main/java/org/infinispan/remoting/inboundhandler/action/TriangleOrderAction.java
```diff
@@ -0,0 +1,41 @@
+package org.infinispan.remoting.inboundhandler.action;
+
+import org.infinispan.distribution.TriangleOrderManager;
+import org.infinispan.util.concurrent.BlockingTaskAwareExecutorService;
+
+/**
+ * An {@link Action} that checks if the command is the next to be executed.
+ * <p>
+ * This action is used by the triangle to order updates from the primary owner to the backup owner.
```
danberindei: triangle... algorithm?

core/src/test/java/org/infinispan/distribution/TriangleOrderManagerTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+   public void testInvalidTopologyId() {
+      TriangleOrderManager triangleOrderManager = new TriangleOrderManager();
+      triangleOrderManager.updateCacheTopology(mockCacheTopology(1));
+
+      try {
+         triangleOrderManager.next(0, 0);
+         fail("Exception expected!");
+      } catch (OutdatedTopologyException e) {
+         log.debug("exception", e);
```
danberindei: I suggest using `Exceptions.expectException()`.

core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
```diff
@@ -0,0 +1,187 @@
...
+   private final Map<Integer, SenderSequencer> senderSegmentMap;
+   private final Map<Integer, ReceiverSequencer> receiverSegmentMap;
+   private volatile CacheTopology currentCacheTopology;
+
+   public TriangleOrderManager() {
+      senderSegmentMap = new ConcurrentHashMap<>();
+      receiverSegmentMap = new ConcurrentHashMap<>();
+   }
+
+   public int calculateSegmentId(Object key) {
```
danberindei: I'd rather not expose this in the `TriangleOrderManager`, as both callers already have access to `ClusterDependentLogic`.

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -300,6 +304,7 @@ public void onTopologyUpdate(final CacheTopology cacheTopology, final boolean is
       // Ensures writes to the data container use the right consistent hash
       // No need for a try/finally block, since it's just an assignment
       stateTransferLock.acquireExclusiveTopologyLock();
+      triangleOrderManager.updateCacheTopology(cacheTopology);
```
danberindei: What happens with commands with the old topology id that were queued at this point, if they manage to run before `this.cacheTopology = cacheTopology`? Granted, that window is likely to be very very short, but in theory we could be executing some commands in the wrong order.

core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
pruivo: ok. I'll double check

core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
pruivo: just keeping it simple. like you said, multiple commands can finish at the same time or close to each other and a single iteration can pick them all and start handling them.

core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
pruivo: no. the ReceiverSequencer isn't updating it's topology when the topology change. it updates when it receives a command with higher topology. it is what this branch is doing (the receiver's topology is less than the command topology)

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
pruivo: the cache topology is only used to assign the sequence number to the backup commands. It doesn't affect the deliver order.

pruivo: in more detail, the ```isNext``` is only invoked when the topology id is higher or the same as the command topology id. This is ensured by the ReadAction mechanism that waits for the topology before performing the next checks.

commons/src/main/java/org/infinispan/commons/marshall/MarshallUtil.java
```diff
@@ -63,13 +65,59 @@
...
+    * {@code null} maps are supported.
+    *
+    * @param map {@link Map} to marshall.
+    * @param out {@link ObjectOutput} to write. It must be non-null.
+    * @param <K> Key type of the map.
+    * @param <V> Value type of the map.
+    * @param <T> Type of the {@link Map}.
+    * @throws IOException If any of the usual Input/Output related exceptions occur.
+    */
+   public static <K, V, T extends Map<K, V>> void marshallMap(T map, ElementWriter<K> keyWriter, ElementWriter<V> valueWrite, ObjectOutput out) throws IOException {
```
pruivo: well, it has the same order as the other ```marshallMap```

danberindei: True, you'd have to change all the collection marshall methods, so let's leave it as is.

core/src/main/java/org/infinispan/distribution/TriangleOrderManager.java
danberindei: Sorry, I think I confused a bit `TOM.setCacheTopology()` with `TOM.removeOldSegments()` and thought it also updated the topology id on each sender/receiver.

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
danberindei: You're right, having the ReceiverSequencer topology id updated only after it sees the first command with the new topology means SCI's cacheTopology was already updated and any commands with the old topology will throw an OutdatedTopologyException. Please add some comments to help me remember this the next time I read it :)

pruivo:
@danberindei updated! if it is ok for you, trigger another perf run to see if I didn't mess up anything :P

danberindei:
n performance tests please

danberindei:
run performance tests please

Holmistr:
run performance tests please

Holmistr:
@pruivo @danberindei I was doing some changes yesterday and I accidentally broke uploading the link back on the PR, however, the tests run successfully, report generation was also successful. 

Link to the results is [here](http://perfrepo.mw.lab.eng.bos.redhat.com/reports/tablecomparisonreport/4711).

pruivo:
@Holmistr I was wondering what happened :) 
@danberindei the results looks good.

danberindei:
Integrated, thanks Pedro!

PR #4742, opened 2016-12-21
ISPN-7323 Reduce number of lambda allocations

danberindei:
https://issues.jboss.org/browse/ISPN-7323

Never use the this::method form, as it creates a new lambda instance
every time.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ClientListenerNotifier.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ClientListenerNotifier.java
@@ -21,6 +21,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 
 import org.infinispan.client.hotrod.annotation.ClientCacheEntryCreated;
 import org.infinispan.client.hotrod.annotation.ClientCacheEntryExpired;
@@ -61,6 +62,8 @@ public class ClientListenerNotifier {
    private final Codec codec;
    private final Marshaller marshaller;
    private final TransportFactory transportFactory;
+
+   private final Consumer<WrappedByteArray> failoverClientListener = this::failoverClientListener;
 
    protected ClientListenerNotifier(
          ExecutorService executor, Codec codec,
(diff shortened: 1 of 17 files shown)


tristantarrant:
Pushed to master

wburns:
Wow didn't realize this::method created a new instance each time :(  Surprised the JVM wouldn't cache that.

rvansa:
@wburns Caching that automatically would unexpectedly increase object size. I hope that static method references are fine, though.

wburns:
@rvansa Agreed, too bad it wouldn't recognize lambdas that are reused a lot and cache them specifically then :)  Although that is a bit more complicated.

rvansa:
@wburns Changing object layout in runtime would likely cause deoptimization of methods even in different classes, and possibly complicate native field accessors as well, so it's likely not worth it.

galderz:
AFAIK, those lambdas that don't capture external variables are the ones that can be optimized/cached by the JVM.

rvansa:
@galderz It captures `this`, as an external variable. According to http://stackoverflow.com/questions/23983832/is-method-reference-caching-a-good-idea-in-java-8
> The current JRE (April 2015, Oracle JDK 8_40) does not cache lambda instances having a state

That means that such optimization is not implemented specifically for lambdas. However once the call-site has been linked, it's subject to optimizations incl. escape analysis and inlining, so the allocation *might* be optimized out. I would say that the only way to test that is warm up the code properly (make sure that it went through C2) and then look on TLAB allocations for given lambda in JFR.

danberindei:
@wburns @rvansa I think it would be pretty much impossible to optimize for this case in the JIT, especially when you consider that we have different interceptor instance for each cache, and an interceptor instance will create a different kind of lambda in each of its visit methods. AFAIK the JIT only keeps track of what classes are used in a call site, not the actual instances.

danberindei:
@galderz @rvansa is right, the allocation could in theory be elided through escape analysis, but that won't work in our case because the lambda is potentially stored in a `CompletableFuture`.

Off-topic, I set up my IntelliJ to underline captured variables, but it still doesn't show calls to non-static methods that implicitly capture `this`. So I think extracting the lambda code in a separate method is a great idea, because it "proves" that you're not capturing anything except `this`.

galderz:
> @galderz It captures this, as an external variable.

Right, but only for non-static lambdas, right? In general I try to make lambdas as swappable by static methods as possible, since those guarantee there's no instance specific state being captured. You could always capture static state though.

danberindei:
@galderz That's not usually possible in the interceptors, because almost every lambda needs to access one or more injected components.

PR #4743, opened 2016-12-22
ISPN-7324 DDAsyncInterceptor indirection slows down replicated reads

danberindei:
https://issues.jboss.org/browse/ISPN-7324

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/DDAsyncInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/DDAsyncInterceptor.java
@@ -46,6 +46,10 @@ public abstract class DDAsyncInterceptor extends BaseAsyncInterceptor implements
    @SuppressWarnings("unchecked")
    @Override
    public final BasicInvocationStage visitCommand(InvocationContext ctx, VisitableCommand command) throws Throwable {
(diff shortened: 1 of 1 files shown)


rvansa:
+0.1
I've heard `instanceof` checks could be tricky, because the type from the class hierarchy is cached. Do you know about some good read on this topic?

danberindei:
I'm using https://shipilev.net/blog/2015/black-magic-method-dispatch/ as a reference. One of the experiments he tries is to "peel off the hottest type with manual checks" to improve performance for megamorphic calls.

rvansa:
That's about megamorphic calls, but manual `instanceof` improves perf by caching the hierarchy, and that's where my knowledge is below freezing point. See https://issues.apache.org/jira/secure/attachment/12833228/reduce_instanceof_typechecking.pdf

danberindei:
@rvansa Yeah, that link doesn't sound very encouraging... the theory is that C2 will have better information on the `else` branch, e.g. when you have 66% reads, 33% writes, 1% "other". 

And it actually worked beautifully for reads without transactions and with implicit transactions. But it actually made things much worse with explicit transactions, because there we have 33.3% read, 33.3% prepare, 33.3% commit, so I'm closing the PR.

PR #4744, opened 2016-12-22
ISPN-7066 Drop default cache inheritance

tristantarrant:
- Make the default cache just a normal named cache
- Make the default-cache attribute on the schema optional
- Create index caches if not configured by the user with sensible defaults

https://issues.jboss.org/browse/ISPN-7066
https://issues.jboss.org/browse/ISPN-7350


tristantarrant:
I am aware of the many failures in the query test suite, which make assumptions about the default cache. I will try to fix them as soon as I can.
In the meantime, please review the core changes.

galderz:
Looks good to me

ryanemerson:
LGTM as well. Nitpick, [unused import](https://github.com/tristantarrant/infinispan/blob/8d45c0030160663dd96f45d89f3bc8b2cd3823f7/core/src/main/java/org/infinispan/configuration/ConfigurationManager.java#L3)

tristantarrant:
@Sanne @gustavonalle I have fixed a direct problem with TypeConverters not being registered in the Component Registry, but now I have hit another snag:
many tests in the query testsuite rely on the fact that the index caches (metadata, chunks, locks) are implicitly defined. In the past this meant that they inherited the default configuration, but since this no longer happens,  it breaks things in new and creative ways. When asking for an undefined caches should we: 
- break with a cacheconfigurationexception (my preference)
- return a default cache configuration (new ConfigurationBuilder().build()) (not what users expect)
- use the default cache as a template of sorts (horrid)
My approach for fixing the tests is to actually define sensible configurations for the 3 index caches.
Please let me know your thoughts

rvansa:
I also support the exception solution. I've misconfigured (due to a typo) the query caches several times in the past and there's little way to deduce what's the problem from the error message.

tristantarrant:
Throwing an exception now causes > 800 failures on core since many tests just assume they can getCache(...) without defining a configuration. I am going ahead and fixing all these anyway, since IMHO assuming that a random configuration is present is wrong anyway.
If we want to support the old (broken) behaviour, we can add a flag to enable inheritance from the default cache configuration.

gustavonalle:
@tristantarrant Can't we just implicitly define them with a sensible default, plus print a WARN "LuceneData, LuceneMetadata and LuceneLock implicitly created, be careful because Infinispan inheritance from the default cache has been removed".

Those sensible defaults would be:
* Indexed.NONE
* If the user cache is LOCAL, all 3 caches should be local
* If the user cache is clustered, the lock/metadata caches should be REPL and LuceneData should 
be DIST

gustavonalle:
> we can add a flag to enable inheritance from 

Oh no, another knob!  :smile:

knobs

Sanne:
I sympathise with Radim's reasoning - I also had trouble in the past caused by typos - so I'd like we aimed for exceptions in the long term but I dislike abrupt changes.
+1 for a warning for the time being.

That said, the 3 caches _LuceneData_, _LuceneMetadata_ and _LuceneLock_ are default names which are defining caches with very specific usage, we could provide a default configuration which is much better than the standard default.

The DirectoryProvider for Hibernate Search starts its own CacheManager and bundles a default configuration which makes sense for this use case - but that benefit is limited for Hibernate Search users, leaving Infinispan Query users with an inferior experience.

I dislike the idea of treating these cache names in a special way, as I guess the whole point of this work was to get rid of "secret keywords", so proposal:

Shall we patch the Lucene Directory code to probe for the valid configuration to be provided for these caches, or failing this to define one programmatically using better defaults?

gustavonalle:
> Shall we patch the Lucene Directory code to probe for the valid configuration to be provided for these caches, or failing this to define one programmatically using better defaults?

+1, that was my proposal as well.

Sanne:
> +1, that was my proposal as well.

Ok cool. I wasn't sure how you meant to do it, but I guess we're on the same page then.

galderz:
Needs rebasing

tristantarrant:
Updated:
- I believe I've fixed all tests that were failing because of ISPN-7066
- I've added ISPN-7350 which creates default index caches if they are missing

This PR is still going for the "hard" approach: i.e. it throws an exception when attempting to getCache() for a non-existing named configuration. This can be easily changed for a softer, WARNing-only approach.

core/src/test/java/org/infinispan/jmx/CacheMBeanTest.java
```diff
@@ -81,7 +82,8 @@ public void testManagerStopRemovesCacheMBean(Method m) throws Exception {
       ObjectName defaultOn = getCacheObjectName(otherJmxDomain);
       ObjectName galderOn = getCacheObjectName(otherJmxDomain, "galder(local)");
       ObjectName managerON = getCacheManagerObjectName(otherJmxDomain);
-      CacheContainer otherContainer = TestCacheManagerFactory.createCacheManagerEnforceJmxDomain(otherJmxDomain);
+      EmbeddedCacheManager otherContainer = TestCacheManagerFactory.createCacheManagerEnforceJmxDomain(otherJmxDomain);
+      otherContainer.defineConfiguration("galder", new ConfigurationBuilder().build());
```
wburns: Glad we are defining Galder properly now! 👍

core/src/test/java/org/infinispan/jmx/CacheMBeanTest.java
```diff
@@ -119,16 +121,6 @@ public void testDuplicateJmxDomainOnlyCacheExposesJmxStatistics() throws Excepti
       }
    }
 
-   public void testMalformedCacheName(Method m) throws Exception {
```
wburns: Do we have a test that verifies the exception is thrown in a case like this?

core/src/test/java/org/infinispan/test/MultiCacheManagerCallable.java
```diff
@@ -20,4 +21,10 @@ public void call() throws Exception {
       // No-op
    }
 
+   protected void defineConfigurationOnAllManagers(String cacheName, ConfigurationBuilder b) {
```
wburns: Doesn't look like this is used at all.

core/src/test/java/org/infinispan/jmx/CacheMBeanTest.java
tristantarrant: TBH that test is just wrong: it does not use JMX to interact with the cache and it is just attempting to retrieve a cache from a container using a weird name... and it passes because it throws no exception since it creates the cache on the fly...

wburns: Yeah I didn't mean the test itself was good.  But rather that we have a test that an exception is indeed thrown when we try to retrieve a cache that isn't defined.

tristantarrant:
After fixing all tests and rebasing, I've reenabled the implicit creation of caches using the default cache configuration. Let's see what CI says.

commons/src/main/java/org/infinispan/commons/api/BasicCacheContainer.java
```diff
@@ -14,6 +14,7 @@
  * @since 4.0
  */
 public interface BasicCacheContainer extends Lifecycle {
```
tristantarrant: Done

galderz:
Thanks @tristantarrant for this! Ready to go in

PR #4746, opened 2016-12-29
ISPN-6802 Remove remaining equivalence references

danberindei:
The invocation context implementations and the transaction
implementations still have a reference to AnyEquivalence.


galderz:
Looks good, integrating.

PR #4747, opened 2017-01-06
Experiments with the async interceptors API to minimize allocations for replicated reads

danberindei:


core/src/main/java/org/infinispan/interceptors/InvocationStage.java
```diff
@@ -12,74 +23,248 @@
...
+    */
+   <P1, P2> InvocationStage thenAccept(P1 p1, P2 p2, TriConsumer<P1, P2, Object> action);
+
+   /**
+    * Execute a callback after the stage has finished successfully.
+    * <p>
+    * The callback can return any {@link InvocationStage}, e.g. using
+    * {@link BaseAsyncInterceptor#invokeNext(InvocationContext, VisitableCommand)} to invoke a new command.
+    */
+   <P1, P2> InvocationStage thenCompose(P1 p1, P2 p2, TriFunction<P1, P2, Object, InvocationStage> function);
```
pruivo: what is the goal of all theses P1 and P2?
If it is to capture the context and command, doesn't make more sense to use InvocationContext and VisitableCommand?
If it is to capture any variable, only 2 parameters isn't enough.

core/src/test/java/org/infinispan/replication/ReplicatedAPITest.java
```diff
@@ -57,7 +57,9 @@ public void put() {
       cache1.put("key", "value");
       waitForRpc(cache2);
 
+      tm(cache1).begin();
```
pruivo: why?

core/src/main/java/org/infinispan/util/function/TetraConsumer.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unlike most other functional interfaces, {@code TriConsumer} is expected
+ * to operate via side-effects.
+ *
+ * <p>This is a <a href="package-summary.html">functional interface</a>
+ * whose functional method is {@link #accept(Object, Object, Object, Object)}.
+ *
+ * @param <T> the type of the first argument to the operation
+ * @param <U> the type of the second argument to the operation
+ * @param <V> the type of the third argument to the operation
+ *
```
pruivo: ```W``` is missing

core/src/main/java/org/infinispan/util/function/TetraConsumer.java
```diff
@@ -0,0 +1,31 @@
+package org.infinispan.util.function;
+
+
+/**
+ * Represents an operation that accepts three input arguments and returns no
+ * result. This is the four-arity specialization of {@link java.util.function.Consumer}.
+ * Unlike most other functional interfaces, {@code TriConsumer} is expected
```
pruivo: ```TriConsumer``` => ```TetraConsumer```

core/src/main/java/org/infinispan/interceptors/impl/ClusteringInterceptor.java
```diff
@@ -17,17 +17,13 @@
 
    protected CommandsFactory cf;
    protected EntryFactory entryFactory;
-   protected LockManager lockManager;
```
pruivo: don't forget to remove the imports

core/src/test/java/org/infinispan/functional/AbstractFunctionalOpTest.java
```diff
@@ -283,9 +283,9 @@ public static VisitableCommand getCurrent() {
       }
 
       @Override
-      protected BasicInvocationStage handleDefault(InvocationContext ctx, VisitableCommand command) throws Throwable {
+      protected InvocationStage handleDefault(InvocationContext ctx, VisitableCommand command) throws Throwable {
          current.set(command);
-         return invokeNext(ctx, command).handle((rCtx, rCommand, rv, t) -> current.remove());
+         return invokeNext(ctx, command).whenComplete(ctx, command, (rCtx, rCommand, rv, t) -> current.remove());
```
danberindei: I'll have to ask @rvansa about that...

core/src/main/java/org/infinispan/interceptors/impl/AsyncResult.java
```diff
@@ -0,0 +1,213 @@
+package org.infinispan.interceptors.impl;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Predicate;
+
+import org.jgroups.annotations.GuardedBy;
```
danberindei: Yep.

core/src/test/java/org/infinispan/interceptors/impl/AsyncResultTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * @author Dan Berindei
+ * @since 9.0
+ */
+@Test(groups = "unit", testName = "interceptors.AsyncResultTest")
+public class AsyncResultTest extends AbstractInfinispanTest {
+   private static final AsyncResult.Invoker0 INVOKER0 = (callback, returnValue, throwable) -> null;
```
danberindei: There are only 2 cases that really matter here, whether `head < tail` or `head >= tail`.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
```diff
@@ -54,7 +50,10 @@
 
    private long transactionDataTimeout;
 
-   private final InvocationComposeHandler handleLocalGetKeysInGroupReturn = this::handleLocalGetKeysInGroupReturn;
+   private final TetraFunction<InvocationContext, VisitableCommand, Object, Throwable, InvocationStage>
+         handleLocalGetKeysInGroupReturn1 = this::handleLocalGetKeysInGroupReturn;
+   private final TetraFunction<InvocationContext, VisitableCommand, Object, Throwable, InvocationStage>
```
danberindei: I'm not sure, I must have used the "extracted field" refactoring too much :)

core/src/main/java/org/infinispan/interceptors/InvocationStage.java
danberindei: Yes, the goal is to capture any variable, and I want to add overloads with 3 parameters too. 4 might be excessive, although I did see at least one callback that could use them.

My initial version (in branch `t_async_api_change_experiments`) was just adding the `ctx` and `command` parameters to all the composition methods, but I thought if we're repeating `ctx, command` so much we should at least get some benefits by being able to use other variables. TBH it's nice, even when the parameters are the context and the command, to be able to use the proper type without declaring a new command variable in the lambda.

core/src/test/java/org/infinispan/replication/ReplicatedAPITest.java
danberindei: I was just abusing the test to debug explicit transactions with just one read operation - I should create a new test method instead.

core/src/test/java/org/infinispan/functional/AbstractFunctionalOpTest.java
rvansa: That's just a `finally` handler, I think that previously the `handle()` method was called incorrectly (since it accepted a consumer, while `CF.handle()` accepts a function), so it should be correct here.

danberindei: @rvansa I think Pedro meant that the callback is going to be called on a different thread, so it's going to remove the wrong thread-local. We could use `try { invokeNext(ctx, command).get(); } finally { current.remove() }` to make the removal more reliable, but that still won't guarantee that the lambda that reads the thread-local is invoked on the same thread.

galderz:
Needs rebasing

danberindei:
Closing this, all those `thenApply()` overloads are just too much :)

PR #4749, opened 2017-01-09
Minor documentation fixes

tristantarrant:
- drop the equivalence chapter
- include the functional API under the cache API
- add new line to cloud chapter to ensure client/server is not adopted there
- fix section numbering in server guide


galderz:
Looks good :)

PR #4755, opened 2017-01-11
ISPN-7286: Cache stores should declare whether they can be shared.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7286

I have created a generic `@Store` annotation which currently only specifies whether a store can be shared or not (default=true), however in the future we can additional store characteristics here to help with config validation. If a store doesn't have a `@Store` annotation, e.g. third party stores, then the additional validation is simply ignored.

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/persistence/Store.java
+++ b/commons/src/main/java/org/infinispan/commons/persistence/Store.java
@@ -1,0 +1,21 @@
+package org.infinispan.commons.persistence;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Store. An annotation for identifying a persistent store and explicitly stating some of its characteristics.
+ *
+ * @author Ryan Emerson
+ * @since 9.0
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface Store {
+   /**
+    * Whether the store can be shared amongst nodes in a distributed/replicated cache
+    */
+   boolean shared() default false;
+}
--- a/core/src/main/java/org/infinispan/configuration/cache/AbstractStoreConfigurationBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/cache/AbstractStoreConfigurationBuilder.java
@@ -14,6 +14,7 @@ import java.util.Properties;
 import org.infinispan.commons.CacheConfigurationException;
 import org.infinispan.commons.configuration.Builder;
 import org.infinispan.commons.configuration.attributes.AttributeSet;
+import org.infinispan.commons.persistence.Store;
 import org.infinispan.commons.util.ReflectionUtil;
 import org.infinispan.commons.util.TypedProperties;
 import org.infinispan.configuration.global.GlobalConfiguration;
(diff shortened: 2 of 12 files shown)


commons/src/main/java/org/infinispan/commons/persistence/Store.java
```diff
@@ -0,0 +1,21 @@
...
+ * Store. An annotation for identifying a persistent store and explicitly stating some of its characteristics.
+ *
+ * @author Ryan Emerson
+ * @since 9.0
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface Store {
+   /**
+    * Whether the store can be shared amongst nodes in a distributed cache
```
wburns: Technically this would affect replicated caches as well.  Maybe just put 'in a clustered cache'

commons/src/main/java/org/infinispan/commons/persistence/Store.java
```diff
@@ -0,0 +1,21 @@
...
+    */
+   boolean shared() default true;
```
wburns: I kinda wonder if we want to put it as false by default.  Being shared has restrictions so I would think a user would have to understand those to mark it as shared.

core/src/main/java/org/infinispan/configuration/cache/AbstractStoreConfigurationBuilder.java
```diff
@@ -188,6 +189,15 @@ public void validate() {
       boolean preload = attributes.attribute(PRELOAD).get();
       boolean transactional = attributes.attribute(TRANSACTIONAL).get();
       ConfigurationBuilder builder = getBuilder();
+
+      Class storeClazz = attributes.getKlass();
+      if (storeClazz != null && storeClazz.isAnnotationPresent(Store.class)) {
```
wburns: Thinking maybe if the annotation isn't present and shared is enabled that we log a WARN message saying that we can't verify if the cache store can be shared or not - as in it should be marked with @Store annotation.

ryanemerson: @wburns I agree, however I think that we can make this more generic as I envisage that long term we can use the `@Store` attribute to validate other store properties. So I have made it so that if the `@Store` annotation is missing a generic warning is issued stating that it is not possible for all validation to occur when said annotation is missing.

ryanemerson:
@wburns Updated based on your feedback. Once we have integrated, I'll send PRs to our cache stores in `/infinispan`

galderz:
Hmmmm, I'm not sure I understand the need for this annotation:

If you want to develop a cache store plugin, you have to implement `CacheWriter` interface, and you need to come up with configuration class for it. Why can't this shared attribute be part of the configuration of a cache store itself? This could be something that's not configurable by the user but it's hardcoded in the configuration, achieving same effect as annotation.

ryanemerson:
@galderz I think annotations on the store implementation is more appropriate, than a field in the associated store configuration, as a store's ability to be shared is a property of the store itself not a configuration option.  Furthermore, the `AbstractStoreConfiguration` already has a `shared` field so introducing another attribute, e.g. `isShareable` could cause confusion.

We could add a method(s) to the `CacheWriter` interface, such as is `boolean isWriterShareable();`, and use a default method so we don't break existing stores that are out of our control. However, such an approach will be more verbose than using an annotation, especially if/when we need to validate additional store properties in the future.

galderz:
> I think annotations on the store implementation are more appropriate than a field in the associated store configuration, as a store's ability to be shared is a property of the store itself not a configuration option. Furthermore, the AbstractStoreConfiguration already has a shared field so introducing another attribute, e.g. isShareable could cause confusion.

I agree that this is a property of the store itself as opposed to a configuration thing, but I'm a little worried that about the amount of things a developer has to do: come up with a CacheWriter impl, create configuration for it, and add annotation to the impl.

> We could add a method(s) to the CacheWriter interface, such as is boolean isWriterShareable();, and use a default method so we don't break existing stores that are out of our control. However, such an approach will be more verbose than using an annotation, especially if/when we need to validate additional store properties in the future.

I agree, this is a more verbose option. I don't want to pollute interfaces with characteristics details.

Let's go with your approach. Although it adds one more thing it's probably the right thing to do.

ryanemerson:
@galderz Ok great. It's worth noting that the `@Store` attribute is optional, so if a user forgets to add it then all they get is a warning. Also, my plan is to update/maintain [infinispan-cachestore-archetype](https://github.com/infinispan/infinispan-cachestore-archetype) to help users out when creating custom stores.

galderz:
@ryanemerson It's in now, remember to add documentation for this!

PR #4756, opened 2017-01-12
ISPN-7296 Bounded Off Heap

wburns:
https://issues.jboss.org/browse/ISPN-7296

This uses a simple LRU bounded cache which is applied to all entries in a given cache.  There is a lru lock that is required to be acquired whenever reading or writing to the lru based attributes or nodes off heap.

The eviction is done by acquiring the said lru lock to test the size and then if it is deemed to needed removal it read the head node of the lru stack and then try to acquire the write lock for the segment of that key.  If it can it will then remove it otherwise it must release the lru lock and try again.  This could be costly if there are a lot of failures to acquire the write lock.  Unfortunately due to deadlocking issues lruLock must be obtained after the write lock of a given segment (thus why a try lock must be used)

### DIFF
--- a/core/src/test/java/org/infinispan/container/offheap/OffHeapMultiNodeTest.java
+++ b/core/src/test/java/org/infinispan/container/offheap/OffHeapMultiNodeTest.java
@@ -59,7 +59,7 @@ public class OffHeapMultiNodeTest extends MultipleCacheManagersTest {
 
       assertNull(map.put(key, value));
 
-      for (int i = 0; i < 100; ++i) {
+      for (int i = 0; i < 50; ++i) {
          map.put(randomBytes(KEY_SIZE), value);
       }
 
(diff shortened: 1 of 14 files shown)


wburns:
Ran with JMH a bit and the performance for the write only seems at a manageable level.

Ran test with eviction of COUNT for object, binary and off heap storage with a size of 500

Average of 3 runs each

Object: 33996
Binary: 36115.5
OffHeap: 32108

For some reason binary was quite fast in my tests.  But as you can see off heap was the slowest but only ~11% slower than the fastest.

core/src/test/java/org/infinispan/container/offheap/OffHeapMultiNodeTest.java
```diff
@@ -59,7 +59,7 @@ public void testPutRemovePut() {
 
       assertNull(map.put(key, value));
 
-      for (int i = 0; i < 100; ++i) {
+      for (int i = 0; i < 0; ++i) {
```
wburns: Weird thought I reverted that :)

wburns:
Fixed test counter not being reverted :)

tristantarrant:
Pushed to master, thanks @wburns this is awesome work !

PR #4757, opened 2017-01-12
StringBasedStoreMultinodeIT Test Failures caused by Eviction config.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7352

Trivial, but let's make sure it actually fixes the issue on TC :)


wburns:
ah thanks.  Thought I got all of these but I guess not :)

wburns:

PR #4758, opened 2017-01-12
ISPN-5729 Remote query - org/apache/lucene/uninverting/UninvertingRea…

anistor:
…der missing on "order by" query

https://issues.jboss.org/browse/ISPN-5729


server/integration/feature-pack/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
```diff
@@ -8,6 +8,7 @@
         <module name="javax.transaction.api" />
         <module name="org.hibernate.commons-annotations" />
         <module name="org.apache.lucene" export="true" />
+        <module name="org.apache.lucene.internal" export="true" />
```
gustavonalle: This sounds a bit strange, an "internal" that is exported? Maybe we should move whatever need to be exported to the "public" module?

gustavonalle: Also, I suppose this change would need to go to the modules we distribute as well https://github.com/infinispan/infinispan/tree/master/wildfly-modules
In that case, we simply overlay whatever the hibernate search module zip gives us, probably we should tell them to export the missing classes...

integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/client/BaseHotRodQueryIT.java
```diff
@@ -68,4 +68,31 @@ public void testRemoteQuery() throws Exception {
 
       rcm.stop();
    }
+
+   @Test
+   public void testUninverting() throws Exception {
```
gustavonalle: It's maybe not trivial to figure out what this test is doing, maybe put a comment like "Using a sort on a field that does not contain DocValues so Lucene is forced to uninvert it" to the posterity?

anistor:
Fair points. I'll address them once I get at least current version of this PR to run successfully in CI....

server/integration/feature-pack/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
gustavonalle: @anistor answering your question from the chat, the lucene-internal module was originally created as [part of ](https://github.com/hibernate/hibernate-search/commit/f9f547a5cdd877ab447a5bdcc391c5acec96ca5a) https://hibernate.atlassian.net/browse/HSEARCH-1849 and judging by the contents, the goal is to host older codecs and compatibility with older Lucene versions.

So maybe we could follow the wirings done in hibernate search that it'd work. Check [here](https://github.com/hibernate/hibernate-search/blob/master/modules/src/main/modules/search/engine/module.xml), [here](https://github.com/hibernate/hibernate-search/blob/master/modules/src/main/modules/lucene/module.xml) and [here](https://github.com/hibernate/hibernate-search/blob/master/modules/src/main/modules/lucene_internal/module.xml)

WRT ```wildfly-modules```, this not required since we simply explode hibernate search modules in the our zip. Probably we just need to test that the Uninverter works.

galderz:
@anistor Have you addressed the questions?

galderz:
As far as CI, `RemoteQueryIspnDirIT` is failing, but it's a flaky one.

anistor:
Only minor comments addressed. Still need to cleanup de module.xml. I'm on it.

server/integration/feature-pack/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
anistor: It seems exporting "org.apache.lucene.internal" was not need. I've added it as a regular dependency and it's enough to fix the problem. 

No need to do anything for the wildfly-modules; they work as they are because the hibernate search modules do not have this issue. 

I think it should be all right now.

PR #4767, opened 2017-01-16
ISPN-7358 Fix Hot Rod server request pipelining

galderz:
* Header parameters might be read twice unless there's a check to see if
  the header value has already been assigned.
* Message id comparison to see if there were not enough bytes should be
  done against Long.MIN_VALUE and not Integer.MIN_VALUE because message
  id is a long.
* Added a basic pipelining test that uses a simplified Hot Rod client
  for sending many operations in one go.

### DIFF
--- a/server/hotrod/src/main/java/org/infinispan/server/hotrod/Decoder2x.java
+++ b/server/hotrod/src/main/java/org/infinispan/server/hotrod/Decoder2x.java
@@ -62,7 +62,13 @@ class Decoder2x implements VersionedDecoder {
             return false;
          }
          byte streamOp = buffer.readByte();
+         if (CacheDecodeContext.isTrace)
+            log.tracef("Header op: %d", streamOp);
+
          int length = ExtendedByteBufJava.readMaybeVInt(buffer);
+         if (CacheDecodeContext.isTrace)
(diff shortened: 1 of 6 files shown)


server/hotrod/src/main/java/org/infinispan/server/hotrod/Decoder2x.java
```diff
@@ -79,21 +85,42 @@ public boolean readHeader(ByteBuf buffer, byte version, long messageId, HotRodHe
          }
          buffer.markReaderIndex();
       }
-      int flag = ExtendedByteBufJava.readMaybeVInt(buffer);
-      if (flag == Integer.MIN_VALUE) {
-         return false;
+
+      if (header.flag == null) {
```
wburns: If you want to use the if statements to check if it is null, then you have to mark the read buffer at the end of the if block.  Otherwise a retry could get messed up.  This is true for all 3.

tristantarrant:
With the change in the hot rod header from native fields to objects I wonder if performance is affected

wburns:
I will be fixing this in another PR as the single frame decoder wasn't be applied properly causing tests to not catch replay bugs.

PR #4770, opened 2017-01-17
ISPN-7361 WrappedByteArray needs toString() impl

galderz:


tristantarrant:
Pushed to master, thanks

PR #4772, opened 2017-01-17
ISPN-7366 and ISPN-7369: Trace information + PutMapCommand bug fix.

pruivo:
https://issues.jboss.org/browse/ISPN-7366
https://issues.jboss.org/browse/ISPN-7369


galderz:
Looking into this... I'll re-run some tests locally first.

galderz:
I can confirm the tests pass :)

galderz:
In the JIRA it says:

> When numOnwer > 2...

But the configuration I'm using in the JS client is numOwners=2...

Anyway, the fix looks good to me, but I'm slightly worried there's no unit tests inside core to replicate this issue? @pruivo can we do something about that?

pruivo:
@galderz I was wrong about "numOwners > 2". 
the ```PutMapCommandNonTxTest``` catches this bug. I don't think it will need a new test.

PR #4773, opened 2017-01-17
[do not integrate] Nano experiments

pruivo:


pruivo:
run performance tests please

core/src/main/java/org/infinispan/container/EntryFactoryImpl.java
```diff
@@ -209,6 +209,18 @@ private CacheEntry getFromContainer(Object key, boolean isOwner, boolean writeOp
...
 
+   private CacheEntry getFromContainerForRead(Object key, boolean isOwner) {
+      InternalCacheEntry ice = container.get(key);
+      if (trace) {
+         log.tracef("Retrieved from container %s", ice);
+      }
+      if (isOwner) {
+         return ice == null ? NullCacheEntry.getInstance() : ice;
+      } else {
+         return ice == null || !ice.isL1Entry() ? null : ice;
```
pruivo: I could have read the code wrongly but this method is only invoked when it ```isOwner==true`` or ``isL1Enabled==true``(the condition is checked in ```wrapEntryForReading()```. In any case, the entry should be valid

Holmistr:
@pruivo Due to changes yesterday, the PerfAck job wasn't triggered. I'm making some enhancements today, I will make sure to re-trigger it when done.

pruivo:
Although the changes affects all configuration, the configuration that may benefit the most is ```library+replicated+reads+no-tx```

core/src/main/java/org/infinispan/container/EntryFactoryImpl.java
danberindei: Right, I missed that we ignore non-L1 entries here. Then the only difference from getFromContainer() is that this method is skipping the if (writeOperation) check in innerGetFromContainer()?

danberindei:
@pruivo There are lots of test failures.

core/src/main/java/org/infinispan/container/EntryFactoryImpl.java
pruivo: yes

Holmistr:
run performance tests please
library repl reads

Holmistr:
@pruivo PerfAck again needs to be aborted because of higher priority testing.

Holmistr:
run performance tests please
library repl reads

PR #4774, opened 2017-01-18
ISPN-7367 CommandAckCollector Improvements

pruivo:
https://issues.jboss.org/browse/ISPN-7367

### DIFF
--- a/core/src/main/java/org/infinispan/commands/ReplicableCommand.java
+++ b/core/src/main/java/org/infinispan/commands/ReplicableCommand.java
@@ -7,6 +7,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 
 import org.infinispan.context.InvocationContext;
+import org.infinispan.remoting.transport.Address;
 import org.infinispan.util.concurrent.CompletableFutures;
 
 /**
@@ -20,7 +21,7 @@ import org.infinispan.util.concurrent.CompletableFutures;
 public interface ReplicableCommand {
    /**
     * Invoke the command asynchronously.
-    *
+    * <p>
     * <p>This method replaces {@link #perform(InvocationContext)} for remote execution.
     * The default implementation and {@link #perform(InvocationContext)} will be removed in future versions.
     * </p>
@@ -33,7 +34,7 @@ public interface ReplicableCommand {
 
    /**
     * Invoke the command synchronously.
-    *
+    * <p>
     * <p>This method is optional. Unless your command never blocks, please implement {@link #invokeAsync()} instead.</p>
     *
     * @since 9.0
@@ -69,8 +70,10 @@ public interface ReplicableCommand {
    byte getCommandId();
 
    /**
-    * If true, a return value will be provided when performed remotely.  Otherwise, a remote {@link org.infinispan.remoting.responses.ResponseGenerator}
-    * may choose to simply return null to save on marshalling costs.
+    * If true, a return value will be provided when performed remotely.  Otherwise, a remote {@link
+    * org.infinispan.remoting.responses.ResponseGenerator} may choose to simply return null to save on marshalling
+    * costs.
+    *
     * @return true or false
     */
    boolean isReturnValueExpected();
@@ -83,7 +86,7 @@ public interface ReplicableCommand {
     * retransmissions. So, the commands that can block (waiting for some state, acquiring locks, etc.) should return
     * true.
     *
-    * @return  {@code true} if the command can block/wait, {@code false} otherwise
+    * @return {@code true} if the command can block/wait, {@code false} otherwise
     */
    boolean canBlock();
 
@@ -103,4 +106,15 @@ public interface ReplicableCommand {
     * @throws ClassNotFoundException if it tries to load an undefined class.
     */
    void readFrom(ObjectInput input) throws IOException, ClassNotFoundException;
+
+   /**
+    * Sets the sender's {@link Address}.
+    * <p>
+    * By default, it doesn't set anything. Implement this method if the sender's {@link Address} is needed.
+    *
+    * @param origin the sender's {@link Address}
+    */
+   default void setOrigin(Address origin) {
+      //no-op by default
+   }
 }
--- a/core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
@@ -170,7 +170,8 @@ public class TriangleDistributionInterceptor extends NonTxDistributionIntercepto
          backupPutMapRcpCommand.setMap(map);
          backupPutMapRcpCommand.setSequence(sequence);
          if (trace) {
-            logCommandSequence(command.getCommandInvocationId(), segmentId, sequence);
+            log.tracef("Command %s got sequence %s for segment %s", command.getCommandInvocationId(), segmentId,
+                  sequence);
          }
          rpcManager.sendToMany(owners.subList(1, size), backupPutMapRcpCommand, DeliverOrder.NONE);
       }
@@ -188,14 +189,14 @@ public class TriangleDistributionInterceptor extends NonTxDistributionIntercepto
 
       if (sync) {
          commandAckCollector
-               .createMultiKeyCollector(command.getCommandInvocationId(), filter.primaries.keySet(), filter.backups,
-                     command.getTopologyId());
+               .createMultiKeyCollector(command.getCommandInvocationId().getId(), filter.primaries.keySet(),
+                     filter.backups, command.getTopologyId());
          final Map<Object, Object> localEntries = filter.primaries.remove(localAddress);
          forwardToPrimaryOwners(command, filter);
          if (localEntries != null) {
             sendToBackups(command, localEntries, consistentHash);
             CompletableFuture<?> remoteGet = checkRemoteGetIfNeeded(ctx, command, localEntries.keySet(), consistentHash,
-                                                                loadType == PRIMARY || loadType == OWNER);
+                  loadType == PRIMARY || loadType == OWNER);
             return makeStage(asyncInvokeNext(ctx, command, remoteGet))
                   .andFinally(ctx, command, onPutMapWithLocalEntries);
          }
@@ -213,20 +214,23 @@ public class TriangleDistributionInterceptor extends NonTxDistributionIntercepto
       return invokeNext(ctx, command);
    }
 
-   private void afterPutMapCommand(InvocationContext ctx, VisitableCommand rCommand, Object rv, Throwable t) {
+   private void afterPutMapCommand(@SuppressWarnings("unused") InvocationContext ignored, VisitableCommand rCommand,
+         Object rv, Throwable t) {
       PutMapCommand cmd = (PutMapCommand) rCommand;
       if (t != null) {
-         commandAckCollector.completeExceptionally(cmd.getCommandInvocationId(), t, cmd.getTopologyId());
+         commandAckCollector.completeExceptionally(cmd.getCommandInvocationId().getId(), t, cmd.getTopologyId());
       } else {
          //noinspection unchecked
-         commandAckCollector.multiKeyPrimaryAck(cmd.getCommandInvocationId(), localAddress,
-                                                (Map<Object, Object>) rv, cmd.getTopologyId());
+         commandAckCollector.multiKeyPrimaryAck(cmd.getCommandInvocationId().getId(), localAddress,
+               (Map<Object, Object>) rv, cmd.getTopologyId());
       }
    }
 
-   private Object onPutMapNoLocalEntriesException(InvocationContext ctx, VisitableCommand rCommand, Throwable t) throws Throwable {
+   private Object onPutMapNoLocalEntriesException(@SuppressWarnings("unused") InvocationContext ignored,
+         VisitableCommand rCommand, Throwable t)
+         throws Throwable {
       PutMapCommand cmd = (PutMapCommand) rCommand;
-      commandAckCollector.completeExceptionally(cmd.getCommandInvocationId(), t, cmd.getTopologyId());
+      commandAckCollector.completeExceptionally(cmd.getCommandInvocationId().getId(), t, cmd.getTopologyId());
       assert t != null;
       throw t;
    }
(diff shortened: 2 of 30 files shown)


pruivo:
PerfAck: library + dist + writes + no-tx

pruivo:
run performance tests please
library dist writes

galderz:
@pruivo Results are in but they seem worse for 100 threads?

pruivo:
run performance tests please
library dist writes

jmarkos:
run performance tests please
library dist writes

pruivo:
personal note: results above used commit f5b4b6b

galderz:
Whoa!! Huge increase in perf!

galderz:
TX writes 100 threads still lower, more work to come?

pruivo:
@galderz . short answer, yes. 
I'm not trusting 100% with the results. it is too much based on the changes. Also, the snapshot didn't pick today's commit and I want to compare with no-bundler.
Transactional caches is not the goal of the PR and they shouldn't be affected.

galderz:
@pruivo So, do you want to tidy up and include in Beta2? Or wait?

pruivo:
@galderz you can release Beta2 without it :) I want to make sure the results are real or not, and I need Dan to review it as well.

galderz:
Fair point

galderz:
Removed milestone

pruivo:
run performance tests please
library dist writes

pruivo:
run performance tests please
library dist writes

pruivo:
squashed and rebased.

pruivo:
run performance tests please
library dist writes

danberindei:
@pruivo So this PR is still a bit slower than master?

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -115,12 +121,12 @@ public void createMultiKeyCollector(CommandInvocationId id, Collection<Address>
...
     * Acknowledges a {@link org.infinispan.commands.write.PutMapCommand} completion in the primary owner.
     *
-    * @param id          the {@link CommandInvocationId}.
+    * @param id          the id from {@link CommandInvocationId#getId()}.
     * @param from        the primary owner.
     * @param returnValue the return value.
     * @param topologyId  the topology id.
     */
-   public void multiKeyPrimaryAck(CommandInvocationId id, Address from, Map<Object, Object> returnValue,
+   public void multiKeyPrimaryAck(long id, Address from, Map<Object, Object> returnValue,
```
danberindei: Shouldn't this `id` be boxed as well?

Actually, I think I'd prefer keeping an unboxed id in the collectors, if only for the extra object header + reference.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -338,27 +340,27 @@ synchronized final void completeExceptionally(Throwable throwable, int topologyI
...
-      private SingleKeyCollector(CommandInvocationId id, Object returnValue, Collection<Address> owners,
-            int topologyId) {
-         super(id, topologyId);
+      private SingleKeyCollector(Long id, Object returnValue, Collection<Address> owners,
+            int topologyId, long timeoutNanos) {
+         super(id, topologyId, timeoutNanos);
          this.returnValue = returnValue;
          this.primaryOwner = owners.iterator().next();
-         Collection<Address> tmpOwners = new HashSet<>(owners);
+         HashSet<Address> tmpOwners = new HashSet<>(owners);
```
danberindei: I think it would be cheaper to receive the list of backups as a parameter, especially after https://github.com/infinispan/infinispan/pull/4804 goes in.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -257,29 +261,27 @@ public void onMembersChange(Collection<Address> members) {
...
-      return log.timeoutWaitingForAcks(Util.prettyPrintTime(timeoutNanoSeconds, TimeUnit.NANOSECONDS), id);
-   }
-
    private abstract class Collector<T> implements Callable<Void>, BiConsumer<T, Throwable> {
 
-      final CommandInvocationId id;
+      final Long id;
       final CompletableFuture<T> future;
       final int topologyId;
+      private final long timeoutNanoSeconds;
```
danberindei: Couldn't we create the timeout task immediately when the collector is created and avoid storing the timeout here?

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -338,27 +340,27 @@ synchronized final void completeExceptionally(Throwable throwable, int topologyI
 
    private class SingleKeyCollector extends Collector<Object> {
       @GuardedBy("this")
-      private final Collection<Address> owners;
+      private final HashSet<Address> owners;
```
danberindei: I'm hoping the JIT can optimize `Set` just as well here...

core/src/main/java/org/infinispan/commands/ReplicableCommand.java
```diff
@@ -103,4 +106,15 @@ default Object perform(InvocationContext ctx) throws Throwable {
...
    void readFrom(ObjectInput input) throws IOException, ClassNotFoundException;
+
+   /**
+    * Sets the sender's {@link Address}.
+    * <p>
+    * By default, it doesn't set anything. Implement this method if the sender's {@link Address} is needed.
+    *
+    * @param origin the sender's {@link Address}
+    */
+   default void setOrigin(Address origin) {
```
pruivo: maybe next time? 
The ```CacheRpcCommand``` has a ```set/getOrigin``` but the ```ReplicableCommand``` not for some reason.

core/src/main/java/org/infinispan/remoting/inboundhandler/GlobalInboundInvocationHandler.java
```diff
@@ -135,54 +126,61 @@ private void initializeCacheRpcCommand(CacheRpcCommand command, ComponentRegistr
...
-               if (throwable instanceof InterruptedException || throwable instanceof IllegalLifecycleStateException) {
-                  log.shutdownHandlingCommand(command);
-                  reply.reply(shuttingDownResponse());
-               } else {
-                  log.exceptionHandlingCommand(command, throwable);
-                  reply.reply(exceptionHandlingCommand(throwable));
-               }
+   private void invokeReplicableCommand(ReplicableCommand command, Reply reply, boolean preserveOrder)
+         throws Throwable {
+      globalComponentRegistry.wireDependencies(command);
```
pruivo: we can't. The ```ReplicableCommand``` inject their dependencies via ```@Inject``` annotation. like ```CacheTopologyControlCommand```

pruivo: I probably should remove the ack collector from the marshaller to make it stateless.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: ok.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: I guess

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: unfortunately, it doesn't. and unfortunately it doesn't help. I can revert the changes if you want.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: what is your idea? 
I still have to copy it to remove the members as the ack is received.

pruivo:
@danberindei updated!

core/src/main/java/org/infinispan/commands/ReplicableCommand.java
danberindei: Ok, let's leave it for a future PR.

core/src/main/java/org/infinispan/remoting/inboundhandler/GlobalInboundInvocationHandler.java
danberindei: `@Inject` in `CacheTopologyControlCommand` is convenient, but not essential, so we can do things manually if it shaves 0.1% off CPU time. Another PR, though :)

I'm fine with the ack collector in the externalizer, I definitely wouldn't want to use reflection to set it on the ack commands.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
danberindei: This would remove 8 bytes from each collector, you should be more excited about it ;)

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
danberindei: I was just curious to see if you saw a difference in perf, I'm fine with `HashSet`.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: I would be more excited if removed 8 bytes from the network :)

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
danberindei: If you start with the backups then you don't have to add + remove the primary to/from the set.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: I saw differences in jitwatch :) it inlines the method but I didn't see any different in numbers.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
pruivo: it makes sense :) I'll change

pruivo:
@danberindei updated again :)

pruivo:
run performance tests please
library dist writes

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -431,25 +431,34 @@ void doCompleteExceptionally(Throwable throwable) {
...
          future.complete(returnValue);
       }
+
+      private void traceReady() {
+         log.tracef("[Collector#%s] Ready! Return value=%ss.", id, returnValue);
+      }
+
+      private void traceBackupAck(int topologyId, Address from) {
+         log.tracef("[Collector#%s] Backup ACK. Address=%s, TopologyId=%s (expected=%s)",
+               id, from, topologyId, this.topologyId);
```
danberindei: Here too, `if (trace)` could be in the method.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -43,36 +43,43 @@
...
 
+   private static TimeoutException createTimeoutException(long timeoutNanoSeconds, long id) {
+      return log.timeoutWaitingForAcks(Util.prettyPrintTime(timeoutNanoSeconds, TimeUnit.NANOSECONDS), id);
+   }
+
    @Inject
    public void inject(
-         @ComponentName(value = KnownComponentNames.TIMEOUT_SCHEDULE_EXECUTOR) ScheduledExecutorService timeoutExecutor,
-         Configuration configuration) {
+         @ComponentName(value = KnownComponentNames.TIMEOUT_SCHEDULE_EXECUTOR) ScheduledExecutorService timeoutExecutor) {
```
danberindei: Don't need `value =`

core/src/main/java/org/infinispan/util/logging/Log.java
```diff
@@ -1473,7 +1473,7 @@ TimeoutException timeoutWaitingForView(int expectedViewId, int currentViewId,
    //removed unused message (id=426)
 
    @Message(value = "Timeout after %s waiting for acks. Id=%s", id = 427)
-   TimeoutException timeoutWaitingForAcks(String timeout, CommandInvocationId id);
+   TimeoutException timeoutWaitingForAcks(String timeout, Long id);
```
danberindei: The id could be a primitive `long`.

pruivo:
@danberindei updated!

danberindei:
CI build hanged in `InitialClusterSizeTest`, but I don't think it's related:
https://issues.jboss.org/browse/ISPN-7439

danberindei:
The AffinityTest failure seems specific to this branch:
http://ci.infinispan.org/project.html?projectId=Infinispan&testNameId=8733223218815095024&tab=testDetails

pruivo:
finally understood the issue. The CommandAckCollector cannot be a global component, it must be a per-cache component since it needs the cache topology updates to keeps its members updated.

pruivo:
run performance tests please
library dist writes

core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
```diff
@@ -91,12 +91,11 @@
    private static final Log log = LogFactory.getLog(TriangleDistributionInterceptor.class);
    private static final boolean trace = log.isTraceEnabled();
    private CommandAckCollector commandAckCollector;
+   private final InvocationExceptionFunction onPutMapNoLocalEntriesException = this::onPutMapNoLocalEntriesException;
```
danberindei: I think it's better to group the lambdas at the end then mixing them with the injected components.

pruivo: no idea why the reformat put the field there... but it does it consistently

pruivo:
updated

danberindei:
Integrated, thanks Pedro!

PR #4776, opened 2017-01-19
ISPN-7371: infinispan-defaults - Rocksdb jar not found for leveldb.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7371

When running `mvn install` we need to get classes from the jar, but when running `mvn test` the jar is not present, instead the classes are already on the classpath. Therefore, we log a missing jar instead of throwing an exception as we know the classes will be retrieved from the classpath.


tristantarrant:
Merged, thanks

PR #4778, opened 2017-01-19
ISPN-7356 Random TimeoutExceptions on query suite

gustavonalle:
https://issues.jboss.org/browse/ISPN-7356

### DIFF
--- a/core/src/test/resources/stacks/tcp.xml
+++ b/core/src/test/resources/stacks/tcp.xml
@@ -8,10 +8,8 @@
          port_range="30"
          recv_buf_size="20m"
          send_buf_size="640k"
-         max_bundle_size="31k"
          enable_diagnostics="false"
-         tcp_nodelay="true"
-         bundler_type="transfer-queue"
+         bundler_type="no-bundler"
 
          thread_naming_pattern="pl"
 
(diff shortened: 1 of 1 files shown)


gustavonalle:
After a couple of CI runs, the errors are not happening anymore, so good to go!

Sanne:
do you have an explanation?

gustavonalle:
@Sanne No. I noticed the failures appeared after we stopped applying the DONT_BUNDLE flag on core, so I basically got rid of bundling altogether on tests.

Sanne:
thanks. Looks like there's a problem with the bundlers then? Or at least how Infinispan is using them? @belaban @danberindei

gustavonalle:
in the JIRA there is a TRACE for jgroups + infinispan in one of the failures, in case someone wants to take a deeper look.

anistor:
I've merged it. Could not stand those failures anymore :). Thanks @gustavonalle !

Sanne:
thanks @anistor I was about to do the same :)

Still let's not sweep mistakes under the carpet. @danberindei @belaban we might have a problem with the bundler implementation?

Since it's causing a timeout in a not so stressing test (it's not a benchmark) I think we might be on a bug r rather than a mere performance inefficiency.

gustavonalle:
@Sanne is right, this is far from being stressful: it uses small cluester (mostly with 3 nodes) with very few entries (< 10). Those caches use indexes stored on extra caches, and it relies (mostly) on SYNC RPCs and custom commands.

PR #4784, opened 2017-01-22
ISPN-7377 RPCs to non-members throw spurious TimeoutException

danberindei:
… members

https://issues.jboss.org/browse/ISPN-7377

MessageDispatcher.cast() removes missing members from the destinations list

### DIFF
--- a/core/src/main/java/org/infinispan/remoting/transport/jgroups/Responses.java
+++ b/core/src/main/java/org/infinispan/remoting/transport/jgroups/Responses.java
@@ -24,31 +24,20 @@ import net.jcip.annotations.ThreadSafe;
 @ThreadSafe
 public class Responses implements Iterable<Map.Entry<Address, Rsp<Response>>> {
    public static final Responses EMPTY = new Responses(Collections.emptyList());
+   public static final Rsp<Response> SUSPECTED_RSP = new Rsp<>();
+
    private final Address[] addresses;
    private final AtomicReferenceArray<Rsp<Response>> responses;
    private volatile boolean timedOut = false;
+
+   static {
+      SUSPECTED_RSP.setSuspected();
+   }
 
    public Responses(Collection<Address> addresses) {
       int size = addresses.size();
       this.addresses = addresses.toArray(new Address[size]);
       this.responses = new AtomicReferenceArray<>(size);
-   }
-
-   /**
-    * Constructs a new instance with responses from all recipients marked
-    * as suspected.
-    *
-    * @param dests Collection of recipients.
-    * @return
-    */
-   public static Responses suspected(Collection<Address> dests) {
-      Responses responses = new Responses(dests);
-      for (int i = 0; i < responses.addresses.length; ++i) {
-         Rsp<Response> rsp = new Rsp<>();
-         rsp.setSuspected();
-         responses.responses.set(i, rsp);
-      }
-      return responses;
    }
 
    /**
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/infinispan/remoting/transport/jgroups/Responses.java
```diff
@@ -58,19 +64,38 @@ public static Responses suspected(Collection<Address> dests) {
...
-      this.responses = new AtomicReferenceArray<>(size);
-      int i = 0;
-      for (Map.Entry<Address, Rsp<Response>> e : results.entrySet()) {
-         addresses[i] = e.getKey();
-         Rsp<Response> rsp = e.getValue();
-         // keep non-received response as null
-         if (rsp.wasReceived() || rsp.wasSuspected() || rsp.wasUnreachable()) {
-            responses.set(i, rsp);
+   public Responses(Collection<Address> dests, RspList<Response> results) {
+      if (dests != null && dests.size() < results.size()) {
```
rvansa: I don't really understand when this happens: is `dest` strict subset of addresses in `results`? Does this mean that there is a result from node we have not sent the message to? On the other hand, it seems that an address from `dests` could be missing in `results` so I don't understand why are you comparing the size of the sets this way...

rvansa:
Is it possible to add an unit test?

core/src/main/java/org/infinispan/remoting/transport/jgroups/Responses.java
danberindei: Yeah, the condition should have been the other way around, and that's why the fix didn't really work.

danberindei:
I fixed the comparison and added a test.

rvansa:
@danberindei Okay, looks better :) However, I would rather like to see this as a fix to JGroups than a workaround like this. Or, even better, we could get rid of the `RspList` completely, as this is quite heavy allocation-wise (it's a HashMap), pushing `Responses` to JGroups. There are quite a few translations from JGroups APIs to different sorts of maps for each request.

belaban:
I've been thinking about changing this to use a more reactive pattern like onNext() for each response, terminated by onCompletion() / onError().
However, this is an API change and I don't want to introduce this so late in the game. Besides, Infinispan will move from MessageDispatcher to JChannel and therefore this is (or will be) moot.

rvansa:
@danberindei How do we remember to remove this burden once Infinispan stops using MD, then?

danberindei:
@rvansa If we stop using MD then we automatically stop using RspList as well, we'll keep only the `HashMap` in `Responses` :)

I've already started working on this, and for now I'm still using `Rsp`, but I completely replaced `RspList` with `Responses`. Before I open the PR I intend to remove the usage of `Rsp` as well. I have also considered changing the API, but I'm not convinced about using lambdas for this.

@belaban Still, @rvansa is right, it would be better if we didn't need this workaround, and `MessageDispatcher.cast()` created a `Rsp` instance for each recipient when the collection of addresses is not `null`.

belaban:
Not sure I follow... why do you still need Rsp or Responses (from JGroups?) if you use the channel directly?

danberindei:
@belaban `Responses` is our own class, but for some reason @rvansa wants to push it to JGroups ;)
I'm still using `Rsp` at the current stage of the migration to JChannel, but I'm removing those usages soon.

rvansa:
@danberindei I don't want to push it into JGroups if that wouldn't be of use; I want to avoid RspList -> Responses transformations.

galderz:
So, what's the conclusion to this?

danberindei:
I think it's ready to go in, and we'll resume the `Responses` discussion when I finish https://issues.jboss.org/browse/ISPN-6971 :)

slaskawi:
Integrated thanks!

PR #4785, opened 2017-01-23
ISPN-7363 Fix Runnable inject for disexec.

SetoKaiba:
Test case for it is added as well.

### DIFF
--- a/cdi/embedded/src/main/java/org/infinispan/cdi/embedded/DelegatingDistributedTaskLifecycle.java
+++ b/cdi/embedded/src/main/java/org/infinispan/cdi/embedded/DelegatingDistributedTaskLifecycle.java
@@ -9,7 +9,7 @@ import javax.enterprise.inject.spi.InjectionTarget;
 
 import org.infinispan.Cache;
 import org.infinispan.cdi.common.util.CDIHelper;
-import org.infinispan.distexec.DefaultExecutorService;
+import org.infinispan.distexec.RunnableAdapter;
 import org.infinispan.distexec.spi.DistributedTaskLifecycle;
 
 public class DelegatingDistributedTaskLifecycle implements DistributedTaskLifecycle {
@@ -49,8 +49,8 @@ public class DelegatingDistributedTaskLifecycle implements DistributedTaskLifecy
             return;
          ContextInputCache.set(inputDataCache);
           preInject(bm, task);
-          if(task instanceof DefaultExecutorService.RunnableAdapter) {
-             preInject(bm, ((DefaultExecutorService.RunnableAdapter) task).getTask());
+          if(task instanceof RunnableAdapter) {
+             preInject(bm, ((RunnableAdapter) task).getTask());
           }
       }
 
@@ -61,9 +61,9 @@ public class DelegatingDistributedTaskLifecycle implements DistributedTaskLifecy
             if (bm == null)
                return;
             postInject(bm, task);
-            if(task instanceof DefaultExecutorService.RunnableAdapter)
+            if(task instanceof RunnableAdapter)
             {
-               postInject(bm, ((DefaultExecutorService.RunnableAdapter) task).getTask());
+               postInject(bm, ((RunnableAdapter) task).getTask());
             }
          } finally {
             ContextInputCache.clean();
--- a/cdi/embedded/src/main/java/org/infinispan/cdi/embedded/DelegatingDistributedTaskLifecycle.java
+++ b/cdi/embedded/src/main/java/org/infinispan/cdi/embedded/DelegatingDistributedTaskLifecycle.java
@@ -9,6 +9,7 @@ import javax.enterprise.inject.spi.InjectionTarget;
 
 import org.infinispan.Cache;
 import org.infinispan.cdi.common.util.CDIHelper;
+import org.infinispan.distexec.DefaultExecutorService;
 import org.infinispan.distexec.spi.DistributedTaskLifecycle;
 
 public class DelegatingDistributedTaskLifecycle implements DistributedTaskLifecycle {
@@ -47,12 +48,10 @@ public class DelegatingDistributedTaskLifecycle implements DistributedTaskLifecy
          if (bm == null)
             return;
          ContextInputCache.set(inputDataCache);
-         Class<Callable<T>> clazz = (Class<Callable<T>>) task.getClass();
-         AnnotatedType<Callable<T>> type = bm.createAnnotatedType(clazz);
-         InjectionTarget<Callable<T>> it = bm.createInjectionTarget(type);
-         CreationalContext<Callable<T>> ctx = bm.createCreationalContext(null);
-         it.inject(task, ctx);
-         it.postConstruct(task);
+          preInject(bm, task);
+          if(task instanceof DefaultExecutorService.RunnableAdapter) {
+             preInject(bm, ((DefaultExecutorService.RunnableAdapter) task).getTask());
+          }
       }
 
       @Override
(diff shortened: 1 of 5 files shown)


SetoKaiba:
The DistributedTaskLifecycleService methods are all intended to work with Callable. And Callable is run in the cluster instead of their own node. Then DistributedExecuteCommand have to be modified as well because it's intended to work with Callable as well. If so, then a lot of code need to be modified. And also to mention, the injecting occurs in the running node instead of the submit node. It's the minimum code I to modify for fix it I think. That's why I implement like that.
If you don't want to add any reference to RunnableAdapter in the DistributedExecuteCommand, DistributedTaskLifeCycleService, DelegatingDistributedTaskLifecycle or DistributedTaskLifeCycleService. Then you have to add Runnable field in DistributedExecuteCommand. And add Runnable methods to DistributedTaskLifecycle, DelegatingDistributedTaskLifecycle and DistributedTaskLifeCycleService.

slaskawi:
Adding @wburns and @tristantarrant to the loop...

SetoKaiba:
I don't like to change any accessing level as well. That's why I messaged slaskawi via IRC and said it may be not easy to fix.
The code can be implemented beautifully with a lot of code modified. 
I was originally ask for CDI support for ClusterExecutor instead of DistributedExecutorService. But after I write some test. I found that Runnable for DES is not working as well.
Wanna see a more beautiful fix for this as well.

slaskawi:
I think I understand fully the problem now...

If we submit a `Callable` instance with `@Inject @Input Cache c;` field, then CDI is able to find it (using field scanning) and do proper injection ([here a proper code snippet](https://github.com/infinispan/infinispan/blob/8.2.x/cdi/embedded/src/main/java/org/infinispan/cdi/DelegatingDistributedTaskLifecycle.java#L53)). 

However, if we submit a `Runnable`, the `DefaultExecutorService` will wrap it up into a `RunnableAdapter` ([see here](https://github.com/infinispan/infinispan/blob/8.2.x/core/src/main/java/org/infinispan/distexec/DefaultExecutorService.java#L389)). The problem is that we lose the information about the fields. The `DelegatingDistributedTaskLifecycle` class just receives a `RunnableAdapter` with no fields (or with no fields annotated using `@Inject @Input`) and CDI can not help us any way.

I believe we have at least a couple of choices here:

- Implement `Runnable` support directly into `DefaultExecutorService`. No wrapping, we need pure support. This will solve all the issues but possibly will cause a lot of refactoring, just as @SetoKaiba mentioned.
- We can allow tight coupling between CDI Extension (`DelegatingDistributedTaskLifecycle`) and `DefaultExecutorService`. This effectively means accepting @SetoKaiba 's PR as is.
- We can ban submitting `Runnables`. This way the user will have to wrap it to `Callable` himself. This actually makes a lot of sense, since you normally expect a result when submitting anything.
- We can remove the `RunnableAdapter` and use Lambdas to wrap `Runnable`s (the code will look like this: `(Callable<Void>) () -> { task.run(); return null; }`). Then in CDI extension I can scan all fields (together with all superclasses) and inject necessary caches. Note, with this approach we don't entirely lose information about the fields.

I would vote for the 3rd option (but first I need to write a quick and dirty POC to check if this can be done this way).

Any thoughts? @vblagoje @wburns

SetoKaiba:
Hi. I just read your comment. Maybe the 4th won't work as well. Since the Lambda is with no difference between a RunnableAdapter. Because the Callable/Runnable submitted by user is not a managed bean. Then it won't inject fields for it. That's why we need DelegatingDistributedTaskLifecycle. If you're using lambda, the task is still not a managed bean. The fields in task won't be injected as well.
Ban the Runnable would be an option. Because Callable<Void> can achieve the same thing. Anyway, without result is understandable, because the task maybe used to make changes to data in the cache, not for a result. Because the cache is distributed. The result is actually there for nodes.

slaskawi:
Yes, I've just noticed you are right @SetoKaiba. Wrapping into Lambda won't help us. That means that option no 4 is off the table.

In that case I would vote for option no 2 but with small changes. I would advise extracting `RunnableAdapter` to a separate class (IMO this will look much better than relying on static inner class) and I would also like to add a comment that any `RunnableAdapter` refactoring might break CDI.

SetoKaiba:
Made some changes based on slaskai's commit.

cdi/embedded/src/main/java/org/infinispan/cdi/embedded/DelegatingDistributedTaskLifecycle.java
```diff
@@ -61,14 +60,31 @@ public DelegatingDistributedTaskLifecycle() {
...
             if (bm == null)
                return;
-            Class<Callable<T>> clazz = (Class<Callable<T>>) task.getClass();
-            AnnotatedType<Callable<T>> type = bm.createAnnotatedType(clazz);
-            InjectionTarget<Callable<T>> it = bm.createInjectionTarget(type);
-            it.preDestroy(task);
-            it.dispose(task);
+            postInject(bm, task);
+            if(task instanceof DefaultExecutorService.RunnableAdapter)
+            {
```
slaskawi: Could you please use proper [formatter](https://github.com/infinispan/infinispan/tree/master/ide-settings)?

core/src/test/java/org/infinispan/distexec/LocalDistributedExecutorTest.java
```diff
@@ -90,6 +90,18 @@ public void basicInvocation(Callable<Integer> call) throws Exception {
...
+   /**
+    * Helper public method (used by CDI module), disabled as some IDEs invoke it as a test method
+    *
+    * @param call
+    * @throws Exception
+    */
+   @Test(enabled = false) // Disable explicitly to avoid TestNG thinking this is a test!!
+   public void basicInvocation(Runnable call) throws Exception {
+      DistributedExecutorService des = createDES(getCache());
+      des.submit(call);
```
slaskawi: I believe, we should call get on the future:

```
des.submit(call).get()
```

SetoKaiba:
Formatted code and add the call of future.get();

galderz:
We're having some issues with the build completing, but in the tests run so far, some timeouts are appearing around `DistributedExecutorCDITest` and other `DistributedExecutorTest` classes. @SetoKaiba can you confirm whether these failures are related to your changes?

SetoKaiba:
@slaskawi Could you please confirm the issue? I can't still make infinispan test case running on my project. I can't start the test. But it still uses the old classes from the repo instead of the local files.

SetoKaiba:
@galderz My test case can be run in my environment now. I don't know why it uses the local changes now. But I can see the test is run successfully without any exception or timeout.

SetoKaiba:
@galderz Please verify this patch. I run the test successfully with the patch. No timeout.
image

slaskawi:
Looking...

slaskawi:
Squashed as `d5b8f888ceae78390205746413d011103c74e765` and integrated into `master` branch.

Thanks a lot @SetoKaiba!

danberindei:
@slaskawi @SetoKaiba could this be causing `DistributedExecutionCompletionTest.testBasicInvocationWithRunnable` to hang? E.g. http://ci.infinispan.org/viewLog.html?buildId=49851&tab=buildResultsDiv&buildTypeId=bt9

pruivo:
@danberindei yes it is. I'm testing a fix and I'll do a PR soon.

pruivo:
Fix here: https://github.com/infinispan/infinispan/pull/4857

slaskawi:
@danberindei Hanging tests should be fixed by https://github.com/infinispan/infinispan/pull/4858

PR #4793, opened 2017-01-24
ISPN-7351 Additional testcase for @Named qualifiers

slaskawi:
https://issues.jboss.org/browse/ISPN-7351

The implementation seems to be fine, adding only test cases.

### DIFF
--- a/cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/cachemanager/programmatic/Config.java
+++ b/cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/cachemanager/programmatic/Config.java
@@ -3,6 +3,7 @@ package org.infinispan.cdi.embedded.test.cachemanager.programmatic;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Disposes;
 import javax.enterprise.inject.Produces;
+import javax.inject.Named;
 
 import org.infinispan.cdi.embedded.ConfigureCache;
 import org.infinispan.configuration.cache.Configuration;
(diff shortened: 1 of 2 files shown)


SetoKaiba:
@Named("large")
   @ConfigureCache("large")
   @Produces
   public Configuration largeConfiguration = new ConfigurationBuilder().memory().size(3).build();

   @Named("small")
   @ConfigureCache("large")
   @Produces
   public Configuration largeConfiguration = new ConfigurationBuilder().memory().size(3).build();

And then you will see the exception.

SetoKaiba:
@Named qualifier is not treated as normal qualifier. Then the @Default will be automatically added. Then it will cause the exception.

slaskawi:
Interesting! Let me check it!

cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/cachemanager/programmatic/Config.java
```diff
@@ -18,6 +19,29 @@
...
+   public Configuration largeConfiguration = new ConfigurationBuilder().memory().size(10).build();
+
+   /**
+    * The same as the above. The intention here is to check whether we can use 2 Configurations with <code>@Named</code>
+    * annotations.
+    */
+   @Named("super-large")
+   @ConfigureCache("super-large")
+   @Produces
+   public Configuration superLargeConfiguration = new ConfigurationBuilder().memory().size(20).build();
```
slaskawi: I added another Configuration which uses `@Named` annotation. It seems fine on my machine, but let's check if it's green on [TeamCity](http://ci.infinispan.org).

SetoKaiba: The exception only exists in 8.x. You're right. It's working correctly with 9.x.

tristantarrant:
Merged, thanks

PR #4795, opened 2017-01-24
ISPN-7076 Cache clear doesn't work when passivation is enabled

wburns:
* Added interceptor to ensure clear is propagated

https://issues.jboss.org/browse/ISPN-7076


wburns:
@ryanemerson I have a feeling that it is just a flaky test, but I am looking into it

wburns:
@ryanemerson Yeah I can confirm that it is not related.  That is a different issue, basically  Caffeine sometimes doesn't evict an entry.  The test(s) still need to be tweaked.

ryanemerson:
@wburns Integrated, thanks!

PR #4797, opened 2017-01-24
ISPN-7296 Add eviction for off-heap

wburns:
* Fixed lru count not including hashCode
* Moved StorageType to public package
* Updated schema for some additional documentation
* Reduced load for test so it doesn't fail with tracing

https://issues.jboss.org/browse/ISPN-7296


tristantarrant:
Looks good. Pushed to master

PR #4800, opened 2017-01-25
ISPN-7396 Improve default bundle size and fragment size

danberindei:
https://issues.jboss.org/browse/ISPN-7396

### DIFF
--- a/cloud/src/main/resources/default-configs/default-jgroups-ec2.xml
+++ b/cloud/src/main/resources/default-configs/default-jgroups-ec2.xml
@@ -54,5 +54,4 @@
    <MFC max_credits="2m" 
         min_threshold="0.40"
    />
-   <FRAG2 />
 </config>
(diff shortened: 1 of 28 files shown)


tristantarrant:
CI says: java.lang.IllegalArgumentException: frag_size (8500) has to be < TP.max_bundle_size (8000)

danberindei:
@tristantarrant Yeah, I thought it was the other way around... I wish they could be the same :)

danberindei:
run performance tests please
library

tristantarrant:
It feels like a mistake: why would I ever want a bundle bigger than the fragment size ?

tristantarrant:
Pushed to master

danberindei:
@tristantarrant I think it's because the bundle size accounts for the size of the headers, and FRAG2 doesn't, but I'm not sure.

PR #4801, opened 2017-01-25
ISPN-7397: Replace use of shared SingleFileStore with DummyInMemoryStore

ryanemerson:
https://issues.jboss.org/browse/ISPN-7397


ryanemerson:
@danberindei Basic javadoc created for DummyInMemoryStore as requested. I also noticed that the debug and failKey options were never actually used by the store or any tests (in a meaningul way), so I have removed these options from the configuration and builder.

danberindei:
Integrated, thanks Ryan!

PR #4802, opened 2017-01-25
ISPN-7398 Fix AbstractProtocolServerTest.testStartingWithoutTransport

tristantarrant:
https://issues.jboss.org/browse/ISPN-7398

### DIFF
--- a/server/core/src/test/java/org/infinispan/server/core/AbstractProtocolServerTest.java
+++ b/server/core/src/test/java/org/infinispan/server/core/AbstractProtocolServerTest.java
@@ -1,5 +1,7 @@
 package org.infinispan.server.core;
 
+import org.infinispan.configuration.cache.Configuration;
+import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.manager.DefaultCacheManager;
 import org.infinispan.server.core.configuration.MockServerConfigurationBuilder;
 import org.infinispan.server.core.configuration.ProtocolServerConfiguration;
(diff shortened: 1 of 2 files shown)


server/core/src/test/java/org/infinispan/server/core/AbstractProtocolServerTest.java
```diff
@@ -50,7 +52,7 @@ public void testStartingWithoutTransport() {
       b.startTransport(false);
       AbstractProtocolServer server = new MockProtocolServer();
       try {
-         server.start(b.build(), new DefaultCacheManager());
+         server.start(b.build(), new DefaultCacheManager(new ConfigurationBuilder().build()));
```
slaskawi: Why do we need this change? Are we deprecating the default constructor?

wburns:
Looking...

wburns:
I think we also need to clean up the `getCache(String, String)` method a bit

Maybe we should have it do

`if (DEFAULT_CACHE_NAME.equals(cacheName)) {
         log.deprecatedDefaultCache();
         return getCache();
      }
`

This way we also get the default cache exception thrown.

server/core/src/test/java/org/infinispan/server/core/AbstractProtocolServerTest.java
tristantarrant: No, but the default constructor no longer provides a default cache, unless you pass it in.
The default constructor alone will give you a cache manager with no caches. You can use defineConfiguration() to provide some

tristantarrant:
Hm, that would create an infinite loop, but yes we need a null handler there too. Let me fix it.

tristantarrant:
@wburns fixed

wburns:
Ah yes in the case when default cache name is set.  I was trying to trust
the of block instead of copying, but I guess copying is required :+1:

wburns:
Pulling...

wburns:

PR #4804, opened 2017-01-25
ISPN-7400 Cache segment ownership information in DistributionManager

danberindei:
https://issues.jboss.org/browse/ISPN-7400
https://issues.jboss.org/browse/ISPN-7553

### DIFF
--- a/core/src/main/java/org/infinispan/commands/write/InvalidateL1Command.java
+++ b/core/src/main/java/org/infinispan/commands/write/InvalidateL1Command.java
@@ -13,7 +13,6 @@ import org.infinispan.container.DataContainer;
 import org.infinispan.container.entries.InternalCacheEntry;
 import org.infinispan.container.entries.MVCCEntry;
 import org.infinispan.context.InvocationContext;
-import org.infinispan.distribution.DataLocality;
 import org.infinispan.distribution.DistributionManager;
 import org.infinispan.notifications.cachelistener.CacheNotifier;
 import org.infinispan.remoting.transport.Address;
@@ -77,8 +76,8 @@ public class InvalidateL1Command extends InvalidateCommand {
       for (Object k : getKeys()) {
          InternalCacheEntry ice = dataContainer.get(k);
          if (ice != null) {
-            DataLocality locality = dm.getLocality(k);
-            if (!locality.isLocal()) {
+            boolean isLocal = dm.getCacheTopology().isWriteOwner(k);
+            if (!isLocal) {
                if (trace) log.tracef("Invalidating key %s.", k);
                MVCCEntry e = (MVCCEntry) ctx.lookupEntry(k);
                notify(ctx, e, true);
--- a/core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
+++ b/core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
@@ -1,0 +1,189 @@
+package org.infinispan.distribution;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.infinispan.commons.hash.MurmurHash3;
+import org.infinispan.commons.util.Immutables;
+import org.infinispan.commons.util.SmallIntSet;
+import org.infinispan.configuration.cache.CacheMode;
+import org.infinispan.distribution.ch.ConsistentHash;
+import org.infinispan.distribution.ch.KeyPartitioner;
+import org.infinispan.distribution.ch.impl.ReplicatedConsistentHash;
+import org.infinispan.remoting.transport.Address;
+import org.infinispan.topology.CacheTopology;
+
+/**
+ * Extends {@link CacheTopology} with information about keys owned by the local node.
+ *
+ * @author Dan Berindei
+ * @since 9.0
+ */
+public class LocalizedCacheTopology extends CacheTopology {
+
+   private final Address localAddress;
+   private final KeyPartitioner keyPartitioner;
+   private final boolean isDistributed;
+   private final boolean allLocal;
+   private final boolean isSegmented;
+   private final int numSegments;
+   private final int numOwners;
+   private final DistributionInfo[] distributionInfos;
+
+   public static LocalizedCacheTopology makeSingletonTopology(CacheMode cacheMode, Address localAddress) {
+      List<Address> members = Collections.singletonList(localAddress);
+      ConsistentHash ch = new ReplicatedConsistentHash(MurmurHash3.getInstance(), members, new int[]{0});
+      CacheTopology cacheTopology = new CacheTopology(0, 0, ch, null, members, null);
+      return new LocalizedCacheTopology(cacheMode, cacheTopology, key -> 0, localAddress);
+   }
+
+   public LocalizedCacheTopology(CacheMode cacheMode, CacheTopology cacheTopology, KeyPartitioner keyPartitioner,
+                                 Address localAddress) {
+      super(cacheTopology.getTopologyId(), cacheTopology.getRebalanceId(), cacheTopology.getCurrentCH(),
+            cacheTopology.getPendingCH(), cacheTopology.getUnionCH(), cacheTopology.getActualMembers(),
+            cacheTopology.getMembersPersistentUUIDs());
+
+      ConsistentHash readCH = getReadConsistentHash();
+      ConsistentHash writeCH = getWriteConsistentHash();
+
+      this.localAddress = localAddress;
+      this.keyPartitioner = keyPartitioner;
+      this.isDistributed = cacheMode.isDistributed();
+      boolean isReplicated = cacheMode.isReplicated();
+      this.isSegmented = isDistributed || isReplicated;
+      this.numSegments = readCH.getNumSegments();
+      this.numOwners = readCH.getNumOwners();
+
+      if (isDistributed) {
+         this.distributionInfos = new DistributionInfo[numSegments];
+         for (int segmentId = 0; segmentId < numSegments; segmentId++) {
+            Address primary = readCH.locatePrimaryOwnerForSegment(segmentId);
+            List<Address> readOwners = readCH.locateOwnersForSegment(segmentId);
+            List<Address> writeOwners = writeCH.locateOwnersForSegment(segmentId);
+            Collection<Address> writeBackups = writeOwners.subList(1, writeOwners.size());
+            this.distributionInfos[segmentId] =
+                  new DistributionInfo(segmentId, primary, readOwners, writeOwners, writeBackups, localAddress);
+         }
+         this.allLocal = false;
+      } else if (isReplicated) {
+         // Writes must be broadcast to the entire cluster
+         Map<Address, List<Address>> readOwnersMap = new HashMap<>();
+         Map<Address, List<Address>> writeOwnersMap = new HashMap<>();
+         this.distributionInfos = new DistributionInfo[numSegments];
+         for (int segmentId = 0; segmentId < numSegments; segmentId++) {
+            int segmentCopy = segmentId;
+            Address primary = readCH.locatePrimaryOwnerForSegment(segmentId);
+            List<Address> readOwners = readOwnersMap.computeIfAbsent(primary, p ->
+                  Immutables.immutableListCopy(readCH.locateOwnersForSegment(segmentCopy)));
+            List<Address> writeOwners = writeOwnersMap.computeIfAbsent(primary, p ->
+                  Immutables.immutableListCopy(writeCH.locateOwnersForSegment(segmentCopy)));
+            List<Address> writeBackups = writeOwners.subList(1, writeOwners.size());
+            this.distributionInfos[segmentId] =
+                  new DistributionInfo(segmentId, primary, readOwners, writeOwners, writeBackups, localAddress);
+         }
+         this.allLocal = readOwnersMap.containsKey(localAddress);
+      } else { // Invalidation/Local
+         assert cacheMode.isInvalidation() || cacheMode == CacheMode.LOCAL;
+         // Reads and writes are local, only the invalidation is replicated
+         List<Address> owners = Collections.singletonList(localAddress);
+         List<Address> writeBackups = Collections.emptyList();
+         this.distributionInfos = new DistributionInfo[]{
+               new DistributionInfo(0, localAddress, owners, owners, writeBackups, localAddress)
+         };
+         this.allLocal = true;
+      }
+   }
+
+   /**
+    * @return {@code true} iff key {@code key} can be read without going remote.
+    */
+   public boolean isReadOwner(Object key) {
+      if (allLocal)
+         return true;
+
+      int segmentId = keyPartitioner.getSegment(key);
+      return distributionInfos[segmentId].isReadOwner();
+   }
+
+
+   /**
+    * @return {@code true} iff writing a value for key {@code key} will update it on the local node.
+    */
+   public boolean isWriteOwner(Object key) {
+      if (allLocal)
+         return true;
+
+      int segmentId = keyPartitioner.getSegment(key);
+      return distributionInfos[segmentId].isWriteOwner();
+   }
+
+   /**
+    * @return The consistent hash segment of key {@code key}
+    */
+   public int getSegment(Object key) {
+      return keyPartitioner.getSegment(key);
+   }
+
+   /**
+    * @return Information about the ownership of segment {@code segment}, including the primary owner.
+    */
+   public DistributionInfo getDistributionForSegment(int segmentId) {
+      return distributionInfos[segmentId];
+   }
+
+   /**
+    * @return Information about the ownership of key {@code key}, including the primary owner.
+    */
+   public DistributionInfo getDistribution(Object key) {
+      int segmentId = isSegmented ? keyPartitioner.getSegment(key) : 0;
+      return distributionInfos[segmentId];
+   }
+
+   /**
+    * @return An unordered collection with the write owners of {@code key}.
+    */
+   public Collection<Address> getWriteOwners(Object key) {
+      int segmentId = isDistributed ? keyPartitioner.getSegment(key) : 0;
+      return distributionInfos[segmentId].writeOwners();
+   }
+
+   /**
+    * @return An unordered collection with the write owners of {@code keys}.
+    */
+   public Collection<Address> getWriteOwners(Collection<?> keys) {
+      if (keys.isEmpty()) {
+         return Collections.emptySet();
+      }
+      if (isDistributed) {
+         if (keys.size() == 1) {
+            Object singleKey = keys.iterator().next();
+            return getDistribution(singleKey).writeOwners();
+         } else {
+            SmallIntSet segments = new SmallIntSet(numSegments);
+            // Expecting some overlap between keys
+            Set<Address> owners = new HashSet<>(2 * numOwners);
+            for (Object key : keys) {
+               int segment = keyPartitioner.getSegment(key);
+               if (segments.add(segment)) {
+                  owners.addAll(getDistributionForSegment(segment).writeOwners());
+               }
+            }
+            return owners;
+         }
+      } else {
+         return getDistributionForSegment(0).writeOwners();
+      }
+   }
+
+   /**
+    * @return The address of the local node.
+    */
+   public Address getLocalAddress() {
+      return localAddress;
+   }
+}
--- a/core/src/main/java/org/infinispan/factories/DistributionManagerFactory.java
+++ b/core/src/main/java/org/infinispan/factories/DistributionManagerFactory.java
@@ -9,7 +9,7 @@ public class DistributionManagerFactory extends AbstractNamedCacheComponentFacto
    @Override
    @SuppressWarnings("unchecked")
    public <T> T construct(Class<T> componentType) {
-      if (configuration.clustering().cacheMode().isDistributed() || configuration.clustering().cacheMode().isReplicated())
+      if (configuration.clustering().cacheMode().isClustered())
          return (T) new DistributionManagerImpl();
       else
          return null;
--- a/core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
@@ -42,15 +42,14 @@ import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.context.impl.TxInvocationContext;
 import org.infinispan.distribution.DistributionInfo;
 import org.infinispan.distribution.DistributionManager;
-import org.infinispan.distribution.Ownership;
+import org.infinispan.distribution.LocalizedCacheTopology;
 import org.infinispan.distribution.RemoteValueRetrievedListener;
 import org.infinispan.distribution.ch.ConsistentHash;
-import org.infinispan.distribution.group.GroupManager;
+import org.infinispan.distribution.ch.KeyPartitioner;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.factories.annotations.Start;
 import org.infinispan.interceptors.InvocationSuccessFunction;
 import org.infinispan.interceptors.impl.ClusteringInterceptor;
-import org.infinispan.interceptors.locking.ClusteringDependentLogic;
 import org.infinispan.remoting.RemoteException;
 import org.infinispan.remoting.RpcException;
 import org.infinispan.remoting.responses.CacheNotFoundResponse;
@@ -82,10 +81,11 @@ public abstract class BaseDistributionInterceptor extends ClusteringInterceptor 
    private static final boolean trace = log.isTraceEnabled();
 
    protected DistributionManager dm;
-   protected ClusteringDependentLogic cdl;
    protected RemoteValueRetrievedListener rvrl;
+   protected KeyPartitioner keyPartitioner;
+
    protected boolean isL1Enabled;
-   private GroupManager groupManager;
+   protected boolean isReplicated;
 
    private final ReadOnlyManyHelper readOnlyManyHelper = new ReadOnlyManyHelper();
    private final InvocationSuccessFunction primaryReturnHandler = this::primaryReturnHandler;
@@ -96,12 +96,11 @@ public abstract class BaseDistributionInterceptor extends ClusteringInterceptor 
    }
 
    @Inject
-   public void injectDependencies(DistributionManager distributionManager, ClusteringDependentLogic cdl,
-         RemoteValueRetrievedListener rvrl, GroupManager groupManager) {
+   public void injectDependencies(DistributionManager distributionManager, RemoteValueRetrievedListener rvrl,
+                                  KeyPartitioner keyPartitioner) {
       this.dm = distributionManager;
-      this.cdl = cdl;
       this.rvrl = rvrl;
-      this.groupManager = groupManager;
+      this.keyPartitioner = keyPartitioner;
    }
 
 
@@ -109,6 +108,7 @@ public abstract class BaseDistributionInterceptor extends ClusteringInterceptor 
    public void configure() {
       // Can't rely on the super injectConfiguration() to be called before our injectDependencies() method2
       isL1Enabled = cacheConfiguration.clustering().l1().enabled();
+      isReplicated = cacheConfiguration.clustering().cacheMode().isReplicated();
    }
 
    @Override
@@ -119,8 +119,9 @@ public abstract class BaseDistributionInterceptor extends ClusteringInterceptor 
          //don't go remote if we are an owner.
          return invokeNext(ctx, command);
       }
+      Address primaryOwner = dm.getCacheTopology().getDistribution(groupName).primary();
       CompletableFuture<Map<Address, Response>> future = rpcManager.invokeRemotelyAsync(
-            Collections.singleton(groupManager.getPrimaryOwner(groupName)), command, defaultSyncOptions);
+            Collections.singleton(primaryOwner), command, defaultSyncOptions);
       return asyncInvokeNext(ctx, command, future.thenAccept(responses -> {
          if (!responses.isEmpty()) {
             Response response = responses.values().iterator().next();
@@ -147,14 +148,13 @@ public abstract class BaseDistributionInterceptor extends ClusteringInterceptor 
 
    protected <C extends FlagAffectedCommand & TopologyAffectedCommand> CompletableFuture<Void> remoteGet(
          InvocationContext ctx, C command, Object key, boolean isWrite) {
-      CacheTopology cacheTopology = checkTopologyId(command);
+      LocalizedCacheTopology cacheTopology = checkTopologyId(command);
       int topologyId = cacheTopology.getTopologyId();
-      ConsistentHash readCH = cacheTopology.getReadConsistentHash();
 
-      DistributionInfo info = new DistributionInfo(key, readCH, rpcManager.getAddress());
-      if (info.ownership() != Ownership.NON_OWNER) {
+      DistributionInfo info = cacheTopology.getDistribution(key);
+      if (info.isReadOwner()) {
          if (trace) {
-            log.tracef("Key %s is local, skipping remote get. Command topology is %d, current topology is %d",
+            log.tracef("Key %s became local after wrapping, retrying command. Command topology is %d, current topology is %d",
                   key, command.getTopologyId(), topologyId);
          }
          // The topology has changed between EWI and BDI, let's retry
@@ -165,14 +165,14 @@ public abstract class BaseDistributionInterceptor extends ClusteringInterceptor 
       }
       if (trace) {
          log.tracef("Perform remote get for key %s. currentTopologyId=%s, owners=%s",
-            key, topologyId, info.owners());
+            key, topologyId, info.readOwners());
       }
 
       ClusteredGetCommand getCommand = cf.buildClusteredGetCommand(key, command.getFlagsBitSet());
       getCommand.setTopologyId(topologyId);
       getCommand.setWrite(isWrite);
 
-      return rpcManager.invokeRemotelyAsync(info.owners(), getCommand, staggeredOptions).thenAccept(responses -> {
+      return rpcManager.invokeRemotelyAsync(info.readOwners(), getCommand, staggeredOptions).thenAccept(responses -> {
          for (Response r : responses.values()) {
             if (r instanceof SuccessfulResponse) {
                SuccessfulResponse response = (SuccessfulResponse) r;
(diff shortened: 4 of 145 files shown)


danberindei:
run performance tests please
library

danberindei:
This needs more work, I missed some of the places still using the ConsistentHash methods, and when I changed them I realized computing the hash of the key in a replicated cache is actually more expensive than checking if the local node is a member of the CH.

danberindei:
@rvansa I don't think it's excessive, considering that the only computation that happens in the `getDistributionInfo()` call is the mapping of the key to the segment. OTOH, when the cache is replicated, we could tell if the key is local without computing the segment, but we need the segment to distinguish between `Ownership.PRIMARY` and `Ownership.BACKUP` (both for read and for write).

rvansa:
@danberindei `ReplicatedConsistentHash.locateOwnersForSegment` allocates a list and does some iterations and comparisons - that's non-trivial. See https://github.com/danberindei/infinispan/blob/af33faa1d6a97d43ed1fba36c5bf5a83e6446477/core/src/main/java/org/infinispan/distribution/ch/impl/ReplicatedConsistentHash.java#L127

Is there a place where you use single DistributionInfo instance for accessing both readCH and writeCH?

danberindei:
@rvansa Yeah, but the point of this PR is to do that computation only once per topology update.

rvansa:
@danberindei Oh, I haven't realized that part - then it's fine ofc :)

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
```diff
@@ -476,15 +480,28 @@ public int getSegmentForKey(Object key) {
...
+      }
+
+      @Override
+      public boolean localNodeIsOwner(Object key) {
+         return distributionManager.getDistributionInfo(key).write().isOwner();
       }
 
       @Override
+      public boolean localNodeIsPrimaryOwner(Object key) {
+         return distributionManager.getDistributionInfo(key).read() == Ownership.PRIMARY;
```
rvansa: why read? DM.getLocality used to use writeCH

core/src/main/java/org/infinispan/util/IntSet.java
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * Represent a set of integers (e.g. segments) as a {@code BitSet}.
+ *
+ * @author Dan Berindei
+ * @since 9.0
+ */
+public class IntSet implements Set<Integer> {
+   BitSet segments;
+
+   public IntSet() {
```
rvansa: I would remove this ctor to prevent accidental use with too big ranges - user should be aware of the max.

core/src/main/java/org/infinispan/util/IntSet.java
```diff
@@ -0,0 +1,196 @@
...
+import java.util.PrimitiveIterator;
+import java.util.Set;
+
+/**
+ * Represent a set of integers (e.g. segments) as a {@code BitSet}.
+ *
+ * @author Dan Berindei
+ * @since 9.0
+ */
+public class IntSet implements Set<Integer> {
```
rvansa: Is it necessary to extend `Set<Integer>` here? I would rather provide primitive methods if such use is possible, and if we want compatibility with `Set<Integer>`, warn through `java.util.Tripwire`

core/src/main/java/org/infinispan/statetransfer/OutboundTransferTask.java
```diff
@@ -85,7 +86,7 @@
 
    private InternalEntryFactory entryFactory;
 
-   public OutboundTransferTask(Address destination, Set<Integer> segments, int stateTransferChunkSize,
+   public OutboundTransferTask(Address destination, Set<Integer> segments, int chunkSize,
```
rvansa: Enforce `IntSet` here?

core/src/main/java/org/infinispan/util/IntSet.java
rvansa: Eh, haven't realized that Tripwire is not a public class...

core/src/main/java/org/infinispan/commands/write/InvalidateL1Command.java
```diff
@@ -104,8 +104,8 @@ public boolean shouldInvoke(InvocationContext ctx) {
       for (Object k : getKeys()) {
          // If any key in the set of keys to invalidate is not local, or we are uncertain due to a rehash, then we
          // process this command.
-         DataLocality locality = dm.getLocality(k);
-         if (!locality.isLocal() || locality.isUncertain()) return true;
+         if (!dm.getDistributionInfo(k).read().isOwner())
```
danberindei: True, going through all the interceptors isn't that much extra work, and removing the check here would allow us to remove `VisitableCommand.shouldInvoke()` as well. I've created [ISPN-7431](https://issues.jboss.org/browse/ISPN-7431) for it.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
danberindei: The primary must be the same everywhere. I've changed this a bit, making `isPrimary()` independent of reads and writes.

core/src/main/java/org/infinispan/util/IntSet.java
danberindei: I wanted to be able to introduce it gradually, and extending `Set<Integer>` helps with that. Otherwise I could have used `BitSet` directly... although TBH I want the "compact" `toString()` as well :)

core/src/main/java/org/infinispan/statetransfer/OutboundTransferTask.java
danberindei: It's going into a `CopyOnWriteArraySet` anyway, so it doesn't really matter where the boxing happens.

core/src/main/java/org/infinispan/util/IntSet.java
danberindei: It does save us from having to marshall both the size and the capacity.

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
rvansa: > `making isPrimary() independent of reads and writes`

Have you changed that recently? In fact this was true even before... but is there any comment preferably in `CacheTopology` stating this?

core/src/main/java/org/infinispan/util/IntSet.java
rvansa: If that means that you'll remove `IntSet` in the next JIRA, use `BitSet` right away. Otherwise the class will stay rotting in the codebase.

danberindei: I said I could have used `BitSet`, not that I wanted to :)

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
```diff
@@ -0,0 +1,190 @@
...
+    */
+   public DistributionInfo getDistributionForSegment(int segmentId) {
+      return distributionInfos[segmentId];
+   }
+
+   /**
+    * @return Information about the ownership of key {@code key}, including the primary owner.
+    */
+   public DistributionInfo getDistribution(Object key) {
+      int segmentId = (isDistributed || isReplicated) ? keyPartitioner.getSegment(key) : 0;
```
rvansa: What about `isSegmented` instead?

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
```diff
@@ -0,0 +1,190 @@
...
+      return distributionInfos[segmentId];
+   }
+
+   /**
+    * @return An unordered collection with the write owners of {@code key}, or {@code null} if the cache is replicated.
+    */
+   public Collection<Address> getWriteTargets(Object key) {
```
rvansa: this method should belong to `DistributionInterceptor`

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
```diff
@@ -0,0 +1,190 @@
...
+import org.infinispan.remoting.transport.Address;
+import org.infinispan.topology.CacheTopology;
+
+/**
+ * Extends {@link CacheTopology} with information about keys owned by the local node.
+ *
+ * @author Dan Berindei
+ * @since 9.0
+ */
+public class LocalizedCacheTopology extends CacheTopology {
```
rvansa: I think you're tying the LocalizedCacheTopology to a cacheMode too much; `CacheTopology` should be kept at the owners-level abstraction.
I'll try to comment where should particular methods belong.
Please think about scattered cache mode as the close-future newcomer; that should modify as few places as possible.

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
```diff
@@ -0,0 +1,190 @@
...
+      ConsistentHash writeCH = getWriteConsistentHash();
+
+      this.localAddress = localAddress;
+      this.keyPartitioner = keyPartitioner;
+      this.isDistributed = cacheMode.isDistributed();
+      this.isReplicated = cacheMode.isReplicated();
+      this.numSegments = readCH.getNumSegments();
+      this.numOwners = readCH.getNumOwners();
+
+      if (isDistributed) {
```
rvansa: I would move these cases to `ClusteringDependentLogic`s

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -253,16 +253,7 @@ private boolean shouldLoad(InvocationContext ctx, AbstractDataWriteCommand comma
...
-                  case PRIMARY:
-                     return true;
-                  case BACKUP:
-                     return !ctx.isOriginLocal();
-                  case NON_OWNER:
-                     return false;
-                  default:
-                     throw new IllegalStateException();
-               }
+               return info.isPrimary() || info.isWriteOwner() && !ctx.isOriginLocal();
```
rvansa: Please use parentheses when combining `||` and `&&` to improve readability

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -50,6 +50,8 @@
 import org.infinispan.context.InvocationContext;
 import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.context.impl.TxInvocationContext;
+import org.infinispan.distribution.DistributionInfo;
```
rvansa: unused imports?

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
```diff
@@ -94,29 +87,52 @@ public boolean isLocal() {
...
 
+   /**
+    * @return information about the location of keys.
+    */
+   LocalizedCacheTopology getCacheTopology();
+
+   /**
+    * @deprecated Since 9.0, please use {@code getCacheTopology().isWriteOwner(key)} instead.
+    */
+   @Deprecated
```
rvansa: Make those methods default?

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
```diff
@@ -286,8 +292,8 @@ public CacheTopology getCacheTopology() {
       }
 
       if (cmdTopologyId < localTopologyId) {
-         ConsistentHash writeCh = cacheTopology.getWriteConsistentHash();
-         Set<Address> newTargets = new HashSet<>(writeCh.locateAllOwners(affectedKeys));
+         Collection<Address> writeTargets = cacheTopology.getWriteTargets(affectedKeys);
```
rvansa: `LCT.getWriteOwners`, never returning `null` would be better. Some people consider `null` an anachronism :)

core/src/main/java/org/infinispan/util/concurrent/locks/LockUtil.java
```diff
@@ -27,18 +28,14 @@ public static Ownership getLockOwnership(Object key, ClusteringDependentLogic cl
...
             key;
-      if (clusteringDependentLogic.localNodeIsPrimaryOwner(keyToCheck)) {
-         return Ownership.PRIMARY;
-      } else if (clusteringDependentLogic.localNodeIsOwner(keyToCheck)) {
-         return Ownership.BACKUP;
-      } else {
-         return Ownership.NON_OWNER;
-      }
+      DistributionInfo distributionInfo = clusteringDependentLogic.getCacheTopology().getDistribution(keyToCheck);
+      return distributionInfo.isPrimary() ? Ownership.PRIMARY :
```
rvansa: I would add `DistributionInfo.getOwnership()` directly there

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
danberindei: The reason I pushed so much in `LocalizedCacheTopology` is that I want to try replacing the `topologyId` in `TopologyAffectedCommand` with a `LocalizedCacheTopology` instance. That way, we'd know that all the interceptors use the same ownership information, and we'd only need one topology check after the remote get for reads and before the commit for writes.

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
danberindei: This happens when the topology changes, and I wanted to avoid having `StateConsumerImpl` depend on `ClusteringDependentLogic`. Besides, I have this feeling that `ClusteringDependentLogic` is doing way too much, I'd love to use `LocalizedCacheTopology` (or maybe some subclasses) for all the ownership stuff and to leave `ClusteringDependentLogic` dealing only with the commit logic (at which point we could also rename it to something more specific).

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: I don't know about you, but for me mapping `||` to `+` and `&&` to `*` is so ingrained that I find parentheses annoying in expressions like this.

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
danberindei: I wanted to keep it here so that it's available for `ClusteringDependentLogic.getOwners()`. But in the meantime I decided that `ClusteringDependentLogic` was only accessible via injection or `AdvancedCache.getComponentRegistry()`, so I can remove methods without breaking backwards compatibility.

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
danberindei: Some people who use other languages? ;)

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
danberindei: All implementations extend `AbstractClusteringDependentLogic`, so I thought I'd keep all implementations there, but sure.

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
danberindei: I'd keep it here because the collection overload is used from `TransactionTable`. But I'm thinking about separating the `null` stuff into a separate method, something like `writeNeedsBroadcast()`.

core/src/main/java/org/infinispan/util/concurrent/locks/LockUtil.java
danberindei: I was hoping to avoid the discussion of read vs write primary by having a single `isPrimary()` in there :)

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
rvansa: That plan's perfectly fine; the cache mode does not change, so the switch should be in the user of the routing information. `LocalizedCacheTopology` should provide only information "who's read/write" owner, an that should be the line. Otherwise, you will end up with a second `ClusteringDependentLogic`, just using if-else instead of polymorphism.

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
rvansa: While I agree that CDL should go on a diet, this is the point which should contain "general things that are based on cache mode". The "writeNeedsBroadcast" shouldn't be there, nor in LocalizedCacheTopology either; it should be in the distribution interceptor. But localization of topology based on cache mode sounds reasonable.
Should the commit-related stuff really depend that much on cache mode? And should it be in CDL at all, rather than in EWI or CommitManager?

rvansa: Hmm, looking on the TransactionTable and other places, it seems that writeNeedsBroadcast should go to CDL, as there are more places that mess with remoting, not just distribution interceptor.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: I'll try to remember that mnemonic

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
danberindei: I only meant I changed `DistributionInfo` so that it doesn't have 2 `Ownership` methods, each possibly being `PRIMARY`.

core/src/main/java/org/infinispan/distribution/LocalizedCacheTopology.java
danberindei: My contention is that we shouldn't have a place for "general things that are based on the cache mode" at all: each "thing" should have its own place. For now, I'm just trying to move the location stuff from CDL to `DistributionManager`/`LocalizedCacheTopology`.

This PR has gotten pretty unwieldy as it is, so I'd like to keep `getWriteTargets()` with the existing semantic of `CDL.getOwners()` and explore a better solution later.

As for the commit stuff, `CommitManager` seems way too tied up with state transfer... although the commit logic in the current CDL also has to know about state transfer, so maybe it's not such a stretch moving it there. I totally agree that commit stuff should be mostly independent to the cache mode, but I didn't get to look into that at all.

rvansa: `getWriteTargets()` has nothing to do with ownership, but with replication (mcasts/broadcasts).
`LocalizedCacheTopology` should describe ownership, not cache modes.

Since replication does not have single "its own place", either we duplicate code a bit based on cache mode in each place or we can put that in one place, but I think CDL is better place for that.

rvansa:
I think that we've reached a stalemate with Dan about the function of `LocalizedCacheTopology`. I prefer to introduce new interfaces as clear, and if we need to introduce something that does not have clear home, I would rather put that into CDL which we now needs cleanup.

Nevertheless, generally this PR is an improvement.

If @pruivo or @wburns feel content with the parts I've objected to, please go forward and integrate.

danberindei:
@rvansa I gave in and I changed `getWriteTargets()` to never return `null`, with the callers checking if the cache is replicated in order to perform a broadcast themselves. I just need to fix some tests before pushing...

danberindei:
Updated!

danberindei:
run performance tests please
library repl

tristantarrant:
Please rebase

danberindei:
Any other comments guys?

pruivo:
one failure caused by the PR:
```
shouldAvoidRpcsDuringIndexing(org.infinispan.query.affinity.AffinityRpcTest)  Time elapsed: 0.631 sec  <<< FAILURE!
java.lang.NullPointerException
	at org.infinispan.query.affinity.RpcDetail.isRpcToItself(RpcDetail.java:32)

```

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -305,18 +292,18 @@ private Object primaryReturnHandler(InvocationContext ctx, VisitableCommand visi
...
          return localResult;
       }
-      ConsistentHash writeCH = checkTopologyId(command).getWriteConsistentHash();
-      List<Address> recipients = writeCH.isReplicated() ? null : writeCH.locateOwners(command.getKey());
 
       // Cache the matcher and reset it if we get OOTE (or any other exception) from backup
       ValueMatcher originalMatcher = command.getValueMatcher();
       // Ignore the previous value on the backup owners
       command.setValueMatcher(ValueMatcher.MATCH_ALWAYS);
+      Collection<Address> recipients = isReplicated ? null : owners;
```
danberindei: Asymmetric caches is exactly why I wanted to use `null` here, to keep the RPC using broadcast while a node joins instead of switching between broadcast and unicast. For performance it probably doesn't make a big difference, but the UNICAST3/NAKACK2 switch also hurts consistency in async non-tx caches.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -328,7 +315,7 @@ private Object primaryReturnHandler(InvocationContext ctx, VisitableCommand visi
       }));
    }
 
-   private RpcOptions determineRpcOptionsForBackupReplication(RpcManager rpc, boolean isSync, List<Address> recipients) {
+   private RpcOptions determineRpcOptionsForBackupReplication(RpcManager rpc, boolean isSync, Collection<Address> recipients) {
```
danberindei: AFAICT `defaultSyncOptions` uses `ResponseMode.SYNCHRONOUS`, not `ResponseMode.SYNCHRONOUS_IGNORE_LEAVERS`.

rvansa: So  does rpc.getDefaultRpcOptions(true);
And you can cache rpcs options for `S_I_L`, too.

danberindei: @rvansa very true, but I not really related to this PR...

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -77,7 +77,7 @@ public void inject(
     * @param owners     the owners of the key. It assumes the first element as primary owner.
     * @param topologyId the current topology id.
     */
-   public void create(long id, Collection<Address> owners, int topologyId) {
+   public void create(long id, List<Address> owners, int topologyId) {
```
danberindei: If it's a set, you don't have any ordering guarantees so you can't assume the first element is the primary.

core/src/main/java/org/infinispan/factories/DistributionManagerFactory.java
```diff
@@ -9,7 +9,7 @@
    @Override
    @SuppressWarnings("unchecked")
    public <T> T construct(Class<T> componentType) {
-      if (configuration.clustering().cacheMode().isDistributed() || configuration.clustering().cacheMode().isReplicated())
+      if (configuration.clustering().cacheMode().isClustered())
```
danberindei: My final goal is to use `DistributionManager` for all ownership related stuff, regardless of clustering mode (although local caches may remain without one), so I can remove `ClusteringDependentLogic` completely.

core/src/main/java/org/infinispan/statetransfer/InboundTransferTask.java
```diff
@@ -214,8 +215,8 @@ public void onStateReceived(int segmentId, boolean isLastChunk) {
       if (!isCancelled && isLastChunk) {
          boolean isCompleted = false;
          synchronized (segments) {
-            if (segments.contains(segmentId)) {
-               finishedSegments.add(segmentId);
+            if (segments.contains((Object) segmentId)) {
```
danberindei: No idea, I think I wrote the initial code using `IntSet` half a year ago :)

commons/src/main/java/org/infinispan/commons/util/IntSet.java
```diff
@@ -0,0 +1,270 @@
...
+         set.set(i);
+      }
+      return set;
+   }
+
+   public IntSet() {
+      bitSet = new BitSet();
+   }
+
+   public IntSet(int initialCapacity) {
```
danberindei: I find it useful, the bitset will only re-allocate if we guessed the capacity wrong, unlike `HashSet<>(initialCapacity)`.

commons/src/main/java/org/infinispan/commons/util/IntSet.java
```diff
@@ -0,0 +1,270 @@
...
+         } else {
+            sb.append(runStart).append('-').append(i);
+         }
+      }
+      sb.append('}');
+      return sb.toString();
+   }
+
+   public static void writeTo(ObjectOutput output, IntSet set) throws IOException {
+      output.writeInt(set.size());
```
danberindei: The size can't be bigger than `Integer.MAX_VALUE`.

core/src/main/java/org/infinispan/statetransfer/OutboundTransferTask.java
```diff
@@ -285,8 +286,8 @@ public String toString() {
       return "OutboundTransferTask{" +
             "topologyId=" + topologyId +
             ", destination=" + destination +
-            ", segments=" + segments +
-            ", stateTransferChunkSize=" + stateTransferChunkSize +
+            ", segments=" + new IntSet(segments) +
```
danberindei: Because `IntSet` orders the elements and shortens ranges, e.g. {1, 2, 3, 5} is written as {1-3, 5}.

commons/src/main/java/org/infinispan/commons/util/IntSet.java
wburns: but it is always positive :)

core/src/main/java/org/infinispan/util/RangeSet.java
```diff
@@ -0,0 +1,154 @@
...
+      }
+
+      @Override
+      public boolean hasNext() {
+         return next < size;
+      }
+
+      @Override
+      public Integer next() {
+         return next++;
```
danberindei: I guess I was relying on the user calling `hasNext()`, adding a check here too.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
wburns: Yeah half the time I always have to go to https://docs.oracle.com/javase/tutorial/java/nutsandbolts/operators.html

commons/src/main/java/org/infinispan/commons/util/IntSet.java
pruivo: not really. if you do  the following code, it will have to allocate:
```
IntSet set = new IntSet(1);
set.add(1024);
```

core/src/main/java/org/infinispan/statetransfer/OutboundTransferTask.java
pruivo: well, I guess it would be better to add another ```toStr``` to ```org.infinispan.commons.util.Util``` for the integer set than pay the cost of creating an ```IntSet``` and iterate over the ```Set``` twice...

pruivo: hmmm... on other hand, you will need to order the ```Set```. I'm not sure if it can be done.

commons/src/main/java/org/infinispan/commons/util/IntSet.java
```diff
@@ -0,0 +1,270 @@
...
+      for (int element : set) {
```
wburns: Yes for the case when you have few elements, but if you have more elements it should be much smaller. Tradeoffs :(

How often do we generate an IntSet of only 1 element though?

So for 1024 size if you have < 32 elements it would be better to print out int positions.

But for any more it would be the inverse. It looks like IntSet is used for both primary and backup owners sets. In the primary owner case it would be 1024 / numNodes which would require numNodes to be > 32 to have less than 32 segments.  And in the backup owner case it is similar and gets worse if backup owners increases.

danberindei: Yes, it looks like it should be more efficient.

core/src/main/java/org/infinispan/cache/impl/SimpleCacheImpl.java
```diff
@@ -455,7 +455,7 @@ public void clear() {
       boolean hasListeners = this.hasListeners;
       ArrayList<InternalCacheEntry<K, V>> copyEntries = null;
       if (hasListeners) {
-         copyEntries = new ArrayList<>(dataContainer.size());
+         copyEntries = new ArrayList<>(dataContainer.sizeIncludingExpired());
```
danberindei: I opened https://github.com/infinispan/infinispan/pull/4927

commons/src/main/java/org/infinispan/commons/util/IntSet.java
danberindei: Sorry, I was thinking `UnsignedNumeric` is about handling the sign properly for values > `Integer.MAX_VALUE`. I take it you just meant I could be using fewer bytes, @pruivo?

commons/src/main/java/org/infinispan/commons/util/IntSet.java
danberindei: I see what you mean now, I was thinking about capacity as the available range but that's obviously not what it means for the other collections. I'm changing the parameter name to `initialRange`.

commons/src/main/java/org/infinispan/commons/util/IntSet.java
danberindei: I had missed your comment @pruivo... the use case for `IntSet` is definitely small ranges based at `0`, so perhaps I should rename it to `SmallIntSet`.

core/src/main/java/org/infinispan/statetransfer/OutboundTransferTask.java
danberindei: Yeah, it would need some temporary storage anyway.

commons/src/main/java/org/infinispan/commons/util/IntSet.java
pruivo: :+1:

commons/src/main/java/org/infinispan/commons/util/IntSet.java
pruivo: yes.

commons/src/main/java/org/infinispan/commons/util/IntSet.java
pruivo: do you think that it is worth to keep it?

danberindei:
Updated. The `AffinityRpcTest` failure appeared because the PR fixed ISPN-7553, and `AffinityRpcTest` was relying on it by using a replicated cache with numOwners = 1.

commons/src/main/java/org/infinispan/commons/util/IntSet.java
wburns: I like that as well 👍

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
```diff
@@ -57,12 +54,9 @@
  *
  * @author Mircea Markus
  * @author Pedro Ruivo
- * @deprecated Since 9.0, no longer public API.
  */
-@Deprecated
```
pruivo: bump

core/src/main/java/org/infinispan/statetransfer/CommitManager.java
```diff
@@ -26,7 +26,7 @@
  * @author Pedro Ruivo
  * @since 7.0
  */
-public class CommitManager {
+public class  CommitManager {
```
pruivo: you miss this one

core/src/test/java/org/infinispan/replication/SyncReplTest.java
```diff
@@ -8,6 +8,7 @@
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
```
pruivo: miss this one

core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
danberindei: `org.infinispan.interceptors.locking` already has a `@private` tag.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
```diff
@@ -92,6 +92,7 @@
    private DataContainer<Object, Object> dataContainer;
    protected ClusteringDependentLogic cdl;
    private VersionGenerator versionGenerator;
+   private DistributionManager distributionManager;
```
danberindei: I'm now using this instead of `cdl`.

pruivo:
no more comments from me.
however, on test failure is caused by the PR:
```
org.infinispan.commons.marshall.NotSerializableException: org.infinispan.remoting.transport.LocalModeAddress
```

danberindei:
@pruivo Looks like I was too aggressive caching the `LocalLogic` topology in a static field. X-site backup works even with local caches, and they need to replicate the local node's address inside a `CommandInvocationId`, so when the manager is clustered we need to use the real address even in local caches.

pruivo:
@danberindei needs rebase

danberindei:
Rebased.

pruivo:
@danberindei a lot of failures in CI

danberindei:
@pruivo Radim's fault, no doubt :)

danberindei:
Not this time! The tests got broken before the rebase, because some tests were instantiating `LocalLogic` without injecting dependencies. They should be fixed now.

pruivo:
looks good. waiting for CI.

pruivo:
@danberindei needs rebase.

danberindei:
Rebased.

gustavonalle:
@danberindei there are some test failures, could you check?

pruivo:
@danberindei you added ```TriangleAckInterceptor```

danberindei:
@gustavonalle They don't seem related.
@pruivo Fixed, I didn't realize I compilation could work with empty files :)

pruivo:
@danberindei unused imports: ```ConditionalOperationPrimaryOwnerFailTest```

danberindei:
Updated @pruivo!

PR #4806, opened 2017-01-26
ISPN-7404: Unresolved directive error in user guide.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7404


tristantarrant:
Pushed to master, thanks

PR #4808, opened 2017-01-27
ISPN-7407 Increase max_credits to 8m

danberindei:
https://issues.jboss.org/browse/ISPN-7407

### DIFF
--- a/core/src/main/resources/default-configs/default-jgroups-udp.xml
+++ b/core/src/main/resources/default-configs/default-jgroups-udp.xml
@@ -49,7 +49,7 @@
    <pbcast.GMS print_local_addr="false"
                join_timeout="${jgroups.join_timeout:5000}"
    />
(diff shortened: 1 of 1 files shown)


danberindei:
run performance tests please
library dist

pruivo:
@danberindei should we integrate this?

rvansa:
@pruivo Previews shouldn't be integrated until author changes the label.

While the effect of UFC is interesting, I don't think that we should keep increasing the default anytime we find out that it causes hiccups in some scenario.

pruivo:
@rvansa I know. it is just has been 4 days without activity :)

danberindei:
@pruivo To be fair, 2 of those days were in the weekend :)

I'm closing it, setting `max_credits` to 8m seems too much like disabling FC and that makes me nervous. Still, it's good to know that it does improve non-tx reads if we run out of options :)

PR #4809, opened 2017-01-27
Comment out unused log messages

rhusar:
Cleans up the current logging mess of roughly 100 unused log messages. Among other things like unnecessary String translations it reduces the core jar distribution size by 0.5% :-)

I do prefer (and others elsewhere too) to keep the messages in commented out so that they can be looked up for historical reasons (that's how i stumbled upon this actually) and mainly, so that messages IDs are less likely to be accidentally reused.


pruivo:
@rhusar first, thanks. second, can you create a JIRA and update the commit message? :)

galderz:
Added JIRA ID and integrated it, thx @rhusar

rhusar:
Thanks @galderz ; sorry I was on PTO last week and didn't get to it this week.

PR #4811, opened 2017-01-30
ISPN-7066 Fix for ExampleConfigsIT#testXsiteConfig failures

tristantarrant:
Do not attempt to use the default cache for xsite backups if one hasn't been configured


pruivo:
integrated! thanks @tristantarrant

PR #4814, opened 2017-01-31
ISPN-7423 Restructure client/server documentation

galderz:
* Restructure the server documentation so that it can be more easily
accessed via the table of contents.


tristantarrant:
Great.  Merged !

PR #4819, opened 2017-02-01
ISPN-7427 Print CLI stats by navigating the async interceptors

tristantarrant:
https://issues.jboss.org/browse/ISPN-7427


ryanemerson:
@tristantarrant Integrated, thanks

PR #4827, opened 2017-02-02
ISPN-5021 Nodes that finish the rebalance later can see outdated values

rvansa:
https://issues.jboss.org/browse/ISPN-5021

### DIFF
--- a/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
+++ b/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
@@ -393,7 +393,7 @@ public class StateConsumerImpl implements StateConsumer {
          int rebalanceTopologyId = stateTransferTopologyId.get();
          if (trace) log.tracef("Topology update processed, stateTransferTopologyId = %d, startRebalance = %s, pending CH = %s",
                (Object)rebalanceTopologyId, startRebalance, cacheTopology.getPendingCH());
-         if (rebalanceTopologyId != NO_REBALANCE_IN_PROGRESS && !startRebalance) {
+         if (rebalanceTopologyId != NO_REBALANCE_IN_PROGRESS && !startRebalance && cacheTopology.getPhase() != CacheTopology.Phase.READ_OLD_WRITE_ALL) {
             // we have received a topology update without a pending CH, signalling the end of the rebalance
             boolean changed = stateTransferTopologyId.compareAndSet(rebalanceTopologyId, NO_REBALANCE_IN_PROGRESS);
             if (changed) {
--- a/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
+++ b/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
@@ -67,6 +67,7 @@ import org.infinispan.remoting.rpc.RpcOptions;
 import org.infinispan.remoting.transport.Address;
 import org.infinispan.remoting.transport.jgroups.SuspectException;
 import org.infinispan.topology.CacheTopology;
+import org.infinispan.topology.LocalTopologyManager;
 import org.infinispan.transaction.impl.RemoteTransaction;
 import org.infinispan.transaction.impl.TransactionTable;
 import org.infinispan.transaction.totalorder.TotalOrderLatch;
@@ -99,6 +100,7 @@ public class StateConsumerImpl implements StateConsumer {
 
    private Cache cache;
    private StateTransferManager stateTransferManager;
+   private LocalTopologyManager localTopologyManager;
    private String cacheName;
    private Configuration configuration;
    private RpcManager rpcManager;
@@ -184,6 +186,7 @@ public class StateConsumerImpl implements StateConsumer {
    public void init(Cache cache,
                     @ComponentName(STATE_TRANSFER_EXECUTOR) ExecutorService stateTransferExecutor,
                     StateTransferManager stateTransferManager,
+                    LocalTopologyManager localTopologyManager,
                     AsyncInterceptorChain interceptorChain,
                     InvocationContextFactory icf,
                     Configuration configuration,
@@ -204,6 +207,7 @@ public class StateConsumerImpl implements StateConsumer {
       this.cacheName = cache.getName();
       this.stateTransferExecutor = stateTransferExecutor;
       this.stateTransferManager = stateTransferManager;
+      this.localTopologyManager = localTopologyManager;
       this.interceptorChain = interceptorChain;
       this.icf = icf;
       this.configuration = configuration;
@@ -389,24 +393,21 @@ public class StateConsumerImpl implements StateConsumer {
          int rebalanceTopologyId = stateTransferTopologyId.get();
          if (trace) log.tracef("Topology update processed, stateTransferTopologyId = %d, startRebalance = %s, pending CH = %s",
                (Object)rebalanceTopologyId, startRebalance, cacheTopology.getPendingCH());
-         if (rebalanceTopologyId != NO_REBALANCE_IN_PROGRESS) {
-            // there was a rebalance in progress
-            if (!startRebalance && cacheTopology.getPendingCH() == null) {
-               // we have received a topology update without a pending CH, signalling the end of the rebalance
-               boolean changed = stateTransferTopologyId.compareAndSet(rebalanceTopologyId, NO_REBALANCE_IN_PROGRESS);
-               if (changed) {
-                  stopApplyingState();
+         if (rebalanceTopologyId != NO_REBALANCE_IN_PROGRESS && !startRebalance) {
+            // we have received a topology update without a pending CH, signalling the end of the rebalance
+            boolean changed = stateTransferTopologyId.compareAndSet(rebalanceTopologyId, NO_REBALANCE_IN_PROGRESS);
+            if (changed) {
+               stopApplyingState();
 
-                  // if the coordinator changed, we might get two concurrent topology updates,
-                  // but we only want to notify the @DataRehashed listeners once
-                  cacheNotifier.notifyDataRehashed(previousReadCh, cacheTopology.getCurrentCH(), previousWriteCh,
-                        cacheTopology.getTopologyId(), false);
-                  if (trace) {
-                     log.tracef("Unlock State Transfer in Progress for topology ID %s", cacheTopology.getTopologyId());
-                  }
-                  if (isTotalOrder) {
-                     totalOrderManager.notifyStateTransferEnd();
-                  }
+               // if the coordinator changed, we might get two concurrent topology updates,
+               // but we only want to notify the @DataRehashed listeners once
+               cacheNotifier.notifyDataRehashed(previousReadCh, cacheTopology.getPendingCH(), previousWriteCh,
+                     cacheTopology.getTopologyId(), false);
+               if (trace) {
+                  log.tracef("Unlock State Transfer in Progress for topology ID %s", cacheTopology.getTopologyId());
+               }
+               if (isTotalOrder) {
+                  totalOrderManager.notifyStateTransferEnd();
                }
             }
          }
@@ -424,11 +425,24 @@ public class StateConsumerImpl implements StateConsumer {
          // Remove the transactions whose originators have left the cache.
          // Need to do it now, after we have applied any transactions from other nodes,
          // and after notifyTransactionDataReceived - otherwise the RollbackCommands would block.
-         if (transactionTable != null) {
-            transactionTable.cleanupLeaverTransactions(rpcManager.getTransport().getMembers());
+         try {
+            if (transactionTable != null) {
+               transactionTable.cleanupLeaverTransactions(rpcManager.getTransport().getMembers());
+            }
+         } catch (Exception e) {
+            // Do not fail state transfer when the cleanup fails. See ISPN-7437 for details.
+            log.transactionCleanupError(e);
          }
 
          commandAckCollector.onMembersChange(newWriteCh.getMembers());
+
+         // The rebalance (READ_OLD_WRITE_ALL) is confirmed through notifyEndOfRebalanceIfNeeded
+         // and STABLE does not have to be confirmed at all
+         switch (cacheTopology.getPhase()) {
+            case READ_ALL_WRITE_ALL:
+            case READ_NEW_WRITE_ALL:
+               localTopologyManager.confirmTopology(cacheName, cacheTopology.getTopologyId(), cacheTopology.getRebalanceId(), null);
+         }
 
          // Any data for segments we do not own should be removed from data container and cache store
          // We need to discard data from all segments we don't own, not just those we previously owned,
@@ -437,7 +451,7 @@ public class StateConsumerImpl implements StateConsumer {
          Set<Integer> removedSegments;
          boolean wasMember =
                previousWriteCh != null && previousWriteCh.getMembers().contains(rpcManager.getAddress());
-         if (isMember || wasMember) {
+         if ((isMember || wasMember) && cacheTopology.getPhase() == CacheTopology.Phase.STABLE) {
             removedSegments = new HashSet<>(newWriteCh.getNumSegments());
             for (int i = 0; i < newWriteCh.getNumSegments(); i++) {
                removedSegments.add(i);
--- a/core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
+++ b/core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
@@ -190,14 +190,11 @@ public class StateTransferInterceptor extends BaseStateTransferInterceptor {
    }
 
    private <C extends VisitableCommand & TopologyAffectedCommand & FlagAffectedCommand> Object handleReadCommand(InvocationContext ctx, C command) throws Throwable {
-      return isLocalOnly(command) ? invokeNext(ctx, command) :
-            updateAndInvokeNextRead(ctx, command);
-   }
-
-   private <C extends VisitableCommand & TopologyAffectedCommand> Object updateAndInvokeNextRead(InvocationContext ctx, C command)
-         throws InterruptedException {
+      if (isLocalOnly(command)) {
+         return invokeNext(ctx, command);
+      }
       updateTopologyId(command);
-      return invokeNextAndHandle(ctx, command,handleReadCommandReturn);
+      return invokeNextAndHandle(ctx, command, handleReadCommandReturn);
    }
 
    private Object handleReadCommandReturn(InvocationContext rCtx, VisitableCommand rCommand, Object rv, Throwable t)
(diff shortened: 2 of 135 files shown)


rvansa:
One thing I find a bit fishy is the `AllOwnersLeftException`; this is in place because of `PartitionHandlingInterceptor`, as we have to propagate the information that all the owners left from current partition. Without partition handling, we could simply wrap null. W/ PH, the options were either throwing an exception or returning a special value that PHI would detect. FTR I don't consider a special value `null` distinctive enough.
In my opinion, a better variant could be installing a hook into distribution interceptor, that would deal with the situation directly there, but that would be out of scope of the current PR.

pruivo:
ok... lets try to reduce the size of this PR.
@rvansa can you create a single PR to handle all the renames? It will be a simple PR to handle and it would reduce this one significantly :)

```
TestingUtil.waitForRehashToComplete => TestingUtil.waitForStateTransferToComplete
LocalTopologyManager.confirmRebalance => LocalTopologyManager.confirmTopology (don't forget AbstractControlledLocalTopologyManager.beforeConfirmRebalance)
ClusterTopologyManager.handleRebalanceCompleted => ClusterTopologyManager.handleTopologyConfirm
CacheTopologyControlCommand.Type.REBALANCE_CONFIRM => CacheTopologyControlCommand.Type.TOPOLOGY_CONFIRM
log.rebalanceError => log.topologyChangeError
CLUSTER.rebalanceCompleted => CLUSTER.topologyChangeCompleted
cacheStatus.doConfirmRebalance => cacheStatus.doTopologyConfirm
RebalanceConfirmationCollector => TopologyConfirmationCollector
RebalanceConfirmationCollector.confirmRebalance => TopologyConfirmationCollector.confirmTopologyUpdate
```

tristantarrant:
Needs rebase as well

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
```diff
@@ -54,40 +63,88 @@
...
-
-   The originator will retry the remote get if node 1 processes the request in T2.
-   If node 1 processes the request in T0 or T1, it still has the entry, and the originator doesn't have to retry.
+   T0   initial topology
+   T1   rehash started
+   T2   rehash finished but state transfer not complete
+   T3   read new, write all topology
+   T4   state transfer completed
 
    | sc     | osc | first request | process request 1 | receive response 1 | retry | process request 2 | receive response 2 |
```
danberindei: You can remove the `osc` column, unless you want to update it to reference the scenario number before your changes.

Also, I suggest splitting the `process request 1` column into `primary request 1` and `backup request 1` instead of using the `1:` and `2:` prefixes.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
```diff
@@ -54,40 +63,88 @@
...
-   | 121_12 |     | T1            | 1:T2              | T1                 | Y     | 2:T1              | T2                 |
-   | [3]    | 6.1 | T1            | 1:T2              | T1                 | Y     | 2:T2              | T1                 |
-   | 121_22 |     | T1            | 1:T2              | T1                 | Y     | 2:T2              | T2                 |
-   | [4]    | 7   | T1            | 1:T2              | T2                 | Y     | 2:T1              | T1                 |
-   | 122_12 | 7   | T1            | 1:T2              | T2                 | Y     | 2:T1              | T2                 |
-   | 122_22 | 7.1 | T1            | 1:T2              | T2                 | Y     | 2:T2              | T2                 |
-
-   *) There will be two retries: first retry will return UnsuccessfulResponse as node 2 is not read owner in T1
-      and node 0 will have to update to T2, the second retry will require update to T2 in node 2 before responding
+   | 010    | 1   | T0            | 1:T1              | T0                 | N1    |                   |                    |
```
danberindei: I think `N1` should be `T1`.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
```diff
@@ -54,40 +63,88 @@
...
+   | [4]    |     | T1            | 1:T4              | T0/T1/T2           |       |                   |                    |
+   | [2]    |     | T1            | 1:T4              | T3/T4              | Y     | 2:T0/T1/T2        |                    |
+   | [4]    |     | T1            | 1:T4              | T3/T4              | Y     | 2:T3/T4           | T0/T1/T2           |
+   | [1]    |     | T1            | 1:T4              | T3/T4              | Y     | 2:T3/T4           | T3/T4              |
+   | [2]    |     | T2            | 1:T0              |                    |       |                   |                    |
+   | [4]    |     | T2            | 1: *, 2:  *       | T0/T1              |       |                   |                    |
+   | 2112   | 8.1 | T2            | 1:T1, 2: T1       | T2/T3/T4           | N1    |                   |                    |
+   | 2122   | 8.2 | T2            | 1:T1, 2: T2       | T2/T3/T4           | N1+2  |                   |                    |
+   | 2132   | 8.3 | T2            | 1:T1, 2: T3/T4    | T2/T3/T4           | N1+2  |                   |                    |
```
danberindei: `N1 + 2`?

Also, I'd put an `_` before the final `2`, so that there's always an underscore between the first request and the second request.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
```diff
@@ -54,40 +63,88 @@
...
+   | [2]    |     | T4            | 2: T0/T1/T2       |                    |       |                   |                    |
+   | [4]    | 12  | T4            | 2:T3              | T0/T1/T2/T3        | N2    |                   |                    |
+   | 434    | 12  | T4            | 2:T3              | T4                 | N2    |                   |                    |
+
+   *) The retry will go to both node 1 and 2 but 1 in T3 will respond with UnsureResponse
    [1] too similar to the previous scenario
-   [2] impossible because we can't have T0 on one node and T2 on another node at the same time
-   [3] impossible, retry is executed only after we allow the topology update
+   [2] impossible because we topologies can't differ by more than 1 at the same time
    [4] impossible, first response was received in later topology than second response
```
danberindei: Since the table is much bigger than before, maybe we should remove the `[4]` rows altogether?

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
```diff
@@ -120,11 +177,16 @@ public void testScenario_010() throws Exception {
...
+      cache(0).getAdvancedCache().getAsyncInterceptorChain()
+            .addInterceptorAfter(new AssertNoRetryInterceptor(), StateTransferInterceptor.class);
+
       //remote get is processed in current topology id.
       Future<Object> remoteGetFuture = remoteGet(cache(0), key);
       rpcManager0.waitForCommandToBlock();
 
-      NewNode joiner = addNode();
+      NewNode joiner = addNode(null, cb -> cb.customInterceptors().addInterceptor()
+         .position(InterceptorConfiguration.Position.FIRST).interceptor(new FailReadsInterceptor()));
```
danberindei: I'm not sure this test should test what happens when a read fails on a node that should have the data based on its current topology, even if (I'm assuming) this lets you simplify the test somewhat.

core/src/test/java/org/infinispan/test/TestingUtil.java
```diff
@@ -222,9 +222,14 @@ public static Object extractField(Class type, Object target, String fieldName) {
...
 
-   public static void waitForRehashToComplete(Cache... caches) {
-      final int REHASH_TIMEOUT_SECONDS = 60; //Needs to be rather large to prevent sporadic failures on CI
-      final long giveup = System.nanoTime() + TimeUnit.SECONDS.toNanos(REHASH_TIMEOUT_SECONDS);
+   /**
+    * Waits until pendingCH() is null on all caches, currentCH.getMembers() contains all caches provided as the param
+    * and all segments have numOwners owners.
+    * @param caches
+    */
+   public static void waitForStateTransferToComplete(Cache... caches) {
```
danberindei: I suggest `waitForStableCache()` instead.

danberindei:
@pruivo @slaskawi Doesn't really make sense to me to move all the renames to a separate PR, since the names are tied to the new rebalance phases. Maybe just leave out `TestingUtil.waitForRehashToComplete()`.

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -424,20 +425,33 @@ public void onTopologyUpdate(final CacheTopology cacheTopology, final boolean is
...
+            }
+         } catch (Exception e) {
+            // Do not fail state transfer when the cleanup fails. See ISPN-7437 for details.
+            log.transactionCleanupError(e);
          }
 
          commandAckCollector.onMembersChange(newWriteCh.getMembers());
 
+         // The rebalance (READ_OLD_WRITE_ALL) is confirmed through notifyEndOfRebalanceIfNeeded
+         // and STABLE does not have to be confirmed at all
```
rvansa: I was considering using a `invokeRemotelyAsync`, but as the topology confirmation is sent from different parts, that would mean creating an uncompleted CF, registering it for given topology somewhere (`LocalCacheStatus`?) and something similar on the coordinator side. I've left that path after a while of refactoring.

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
```diff
@@ -357,7 +361,7 @@ private Object handleTxWriteReturn(InvocationContext rCtx, VisitableCommand rCom
          throws Throwable {
       int retryTopologyId = -1;
       WriteCommand writeCommand = (WriteCommand) rCommand;
-      if (t instanceof OutdatedTopologyException) {
+      if (t instanceof OutdatedTopologyException || t instanceof AllOwnersLostException) {
```
rvansa: You mean in JGroupsTransport? There it's a different level of abstraction... I would decide probably after your glorious withdrawal from `MessageDispatcher`.

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
```diff
@@ -416,7 +420,7 @@ private Object handleNonTxWriteReturn(InvocationContext rCtx,
       while (ce instanceof RemoteException) {
          ce = ce.getCause();
       }
-      if (!(ce instanceof OutdatedTopologyException) && !(ce instanceof SuspectException))
+      if (!(ce instanceof OutdatedTopologyException) && !(ce instanceof SuspectException) && !(ce instanceof AllOwnersLostException))
```
rvansa: Yes, e.g. when a backup needs previous value, tries to fetch that from remote nodes and finds out that all are dead...

core/src/main/java/org/infinispan/topology/CacheTopology.java
```diff
@@ -223,4 +249,42 @@ public Integer getId() {
...
    }
+
+   /**
+    * Phase of topology change during rebalance. Using four phases guarantees these properties:
+    *
+    * 1. T(x+1).writeCH contains all nodes from Tx.readCH (this is the requirement for ISPN-5021)
+    * 2. Tx.readCH and T(x+1).readCH has non-empty subset of nodes (that will allow no blocking for read commands
+    *    && reading only entries node owns according to readCH)
+    *
+    * Old entries should be wiped out only after coming to the {@link #STABLE} phase.
```
rvansa: Hmm, when node A is in READ_NEW_WRITE_ALL, it's possible that node B is in READ_ALL_WRITE_ALL. Therefore, B could send a read request to A (and other nodes) - A will respond with UnsureResponse anyway. So yes, we could already drop the data at that point.

rvansa: However, if you drop old entries at the beginning of READ_NEW_WRITE_ALL phase, some writes will store data in DC, and you do not wipe these. Actually, when a node is in this phase, it does not need to execute the writes locally, but it has to send writes to all *other* nodes according to writeCH. We could optimize that out in `LocalizedCacheTopology` by removing self as a write owner, and then we could move the clear to the point when we receive RNWA.

core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
```diff
@@ -307,13 +307,21 @@ private boolean doHandleTopologyUpdate(String cacheName, CacheTopology cacheTopo
 
          ConsistentHash unionCH = null;
          if (cacheTopology.getPendingCH() != null) {
-            unionCH = cacheStatus.getJoinInfo().getConsistentHashFactory().union(cacheTopology.getCurrentCH(),
-                  cacheTopology.getPendingCH());
+            if (cacheTopology.getPhase() == CacheTopology.Phase.READ_NEW_WRITE_ALL) {
+               // When removing members from topology, we have to make sure that the unionCH has
+               // owners from pendingCH (which is used as the readCH in this phase) before
+               // owners from currentCH, as primary owners must be always readable.
```
rvansa: Will change the comment, but I won't extend the scope of the PR unless necessary.

core/src/test/java/org/infinispan/test/TestingUtil.java
rvansa: That's not much more clear. In #4862 I use `waitForRebalanceToComplete`. Could be `waitUntilRebalanceCompletes`, if you find that better.

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
danberindei: Yeah, `LocalCacheStatus` would be the place to do it, because it now has the `LimitedExecutorService` to synchronize the execution of the topology updates, and `future.whenComplete()` should be able to replace the executor. But you were probably right to abandon that approach here :)

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
danberindei: I was thinking of doing it in the remote get `ResponseFilter`, but you're probably right, the current filter interface isn't a good fit.

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
danberindei: Ok

core/src/main/java/org/infinispan/topology/CacheTopology.java
danberindei: No, I don't think the added complexity is worth it, as the only real benefit would be in the tests that check the data container directly.

core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
danberindei: Sure

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -139,153 +140,143 @@ public AvailabilityMode getAvailabilityMode() {
    }
 
    @Override
-   public void updateAvailabilityMode(List<Address> actualMembers, AvailabilityMode newAvailabilityMode,
+   public synchronized void updateAvailabilityMode(List<Address> actualMembers, AvailabilityMode newAvailabilityMode,
```
rvansa: I really prefer concise code, when one keyword can replace two extra lines and indentation level.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
rvansa: Will remove `osc`, but won't add another columns; the prefixes show clearly which node does that.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
rvansa: Nope, N1 means: we won't do a retry because we got response from node 1.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
rvansa: N1+2: we got successful responses from both node 1 and 2.
There's no second request, that's the point of the missing underscore.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
rvansa: I'd keep it to make sure we have all the combinations covered (at least by reasoning that it's not possible).

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
danberindei: Ok, please add a comment.

danberindei: Actually, if there's no second request, what does the final 2 mean? Most of the other scenario ids only have 3 digits.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
danberindei: Ok

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
rvansa: It's not the final 2, it's the second number which got duplicated. If you read the topology references as written on the line in the table, you'll see those numbers:
2132: Originator in topology *2* sent request to both two nodes, node 1 receiving that in topology *1* and node 2 receiving that in topology *3*. The response i received still in topology *2*.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
danberindei: IMO it's not really an assertion because the exception gets ignored if there's a successful response from another node, so I would have preferred an out-of-band way to assert there was no request to this node. (e.g. with an `assertNoHits` method on the interceptor).

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
danberindei: Sorry, I guess I got confused by my old codes then: looking at the table now, I have no idea why I didn't make all the codes at least 4 digits, or why I didn't write which node owns the key in the initial topology and in the final one :)

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
rvansa: Added explicit assertion.

PR #4832, opened 2017-02-03
removed fix for the java 6 bug (id 6427854) fixed in java 7

diuis:
I removed the fix for the Java 6 bug (id 6427854, http://bugs.java.com/view_bug.do?bug_id=6427854).
It was fixed in 2011 (with Java7 b08).
Infinispan is baselined on Java 8.


tristantarrant:
Thanks !

PR #4837, opened 2017-02-08
ISPN-6038 Asymmetric remote caches not supported

galderz:
* Asymmetric clusters where some nodes might have some caches and others
  don't is not really a supported scenarion.
* Ever since ping on startup was enabled by default, there is the
  possibility that under such scenarios, if the target for where the
  ping will be directed is a node that has not a cache, the returned
  cache will be null.


tristantarrant:
the server testsuite also has some asymmetric tests too: server/integration/testsuite/src/test/java/org/infinispan/server/test/asymmetric/AsymmetricClusterIT.java

galderz:
Sorted @tristantarrant

tristantarrant:
Some more places:
https://github.com/infinispan/infinispan/blob/master/server/integration/testsuite/build-testsuite.xml#L188
https://github.com/infinispan/infinispan/blob/master/server/integration/testsuite/src/test/resources/config/infinispan/asymmetric-1.xml
https://github.com/infinispan/infinispan/blob/master/server/integration/testsuite/src/test/resources/config/infinispan/asymmetric-2.xml

galderz:
@tristantarrant Fixed

PR #4842, opened 2017-02-09
ISPN-7454 Always set trimStacktrace to false

galderz:
https://issues.jboss.org/browse/ISPN-7454

* Without this specific setting, exception causes for test failures are
  hidden from the report.


tristantarrant:
Can't this be set in the parent POM declaration instead of specifying per-module ?

galderz:
I'll see if that's possible.

PR #4843, opened 2017-02-09
Fix Lucene Directory integration tests and improve the modules

Sanne:
I've added a README file in the root of the integration tests, as some changes had been made which make me suspect people didn't understand the purpose of these tests.

Please do not override all components to "align" them blindly among all components: the point of the modules - and hence the modular tests - is not only to verify a sound structure of the module, but also to verify cross-compatibility among the different versions which we expect to be valid combinations in the practical world.

 - https://issues.jboss.org/browse/ISPN-7455
 - https://issues.jboss.org/browse/ISPN-7458

### DIFF
--- a/integrationtests/as-lucene-directory/READ.ME
+++ b/integrationtests/as-lucene-directory/READ.ME
@@ -1,0 +1,47 @@
+##                                                                           ##
+#  Integration tests for the Infinispan Lucene Directory modules for WildFly  #
+##                                                                           ##
+
+The `Infinispan Lucene Directory` is used in a variety of contexts,
+in combination with different versions of other frameworks.
+
+Some of these integration tests will look very similar to each other,
+but pay attention to the actual versions / module slots being used
+which will likely differ.
+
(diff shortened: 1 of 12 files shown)


tristantarrant:
CI reports failures with these:
 query.DSLQueryIT  	
query.ElasticsearchIndexManagerIT  	
query.InfinispanQueryIT

Sanne:
thanks @tristantarrant , I'll check those out. I really need a faster machine :)

Sanne:
The tests are failing (rightfully) because after my changes the "${lucene.slot}" property of the infinispan query module is no longer being filtered correctly.

I can't see how that relates with my changes, however this works fine on master.

Sanne:
I pushed a fix and rebased. Should all work now.

I didn't fully understand the strange behaviour of ANT in this case, but my last commit contains a workaround which seems effective.

gustavonalle:
Thanks, let's wait for CI

tristantarrant:
Merged, thanks

PR #4844, opened 2017-02-09
ISPN-7454 Always set trimStacktrace to false

galderz:
https://issues.jboss.org/browse/ISPN-7454

* Without this specific setting, exception causes for test failures are
hidden from the report.


wburns:
Wow you broke github... Infinity files changed :)

tristantarrant:
Verified locally. Merged. Thanks

PR #4845, opened 2017-02-09

alanfx:
JUnit reports

* Disabled the XML report in the parent POM
* Changed the XML file prefix in PolarionJUnitXMLReporter from
"POLARION-" to "TEST-"

### DIFF
--- a/commons-test/src/main/java/org/infinispan/commons/test/PolarionJUnitXMLReporter.java
+++ b/commons-test/src/main/java/org/infinispan/commons/test/PolarionJUnitXMLReporter.java
@@ -1,5 +1,6 @@
 package org.infinispan.commons.test;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -16,7 +17,6 @@ import org.testng.ITestContext;
 import org.testng.ITestNGMethod;
 import org.testng.ITestResult;
 import org.testng.annotations.Test;
-import org.testng.collections.Lists;
 import org.testng.collections.Maps;
 import org.testng.internal.IResultListener2;
 import org.testng.internal.Utils;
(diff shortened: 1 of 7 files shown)


alanfx:
@danberindei can you take a look at this? I figured out how to disable the creation of the Surefire XML results files, and the text summary results files can be disabled by adding `<useFile>false</useFile>` to the parent POM as well. Thanks

danberindei:
@alanfx Looks like nobody uses the text reports, please disable them too.

alanfx:
@danberindei updated!

danberindei:
@alanfx replacing the Surefire XML report with yours seems to have impacted the number of tests in CI: 14945 (this PR) vs 17709 (master)

http://ci.infinispan.org/viewLog.html?buildId=49855
http://ci.infinispan.org/viewLog.html?buildId=48199

You can download a CSV of the test names as parsed by TeamCity from the tests tab and compare them to see which ones are missing, then compare the generated XML from your reporter and the default one.

Also, I like how the suite name is now the module name, but there are still > 2k tests with `<no suite>`. Maybe we can set a proper suite for those too?

alanfx:
@danberindei ok, I got it figured out. Disabling the XML file in the parent/pom.xml was also preventing the XML file from being generated for modules that use JUnit. I had to enable the XML results file in those modules to fix the issue, and now the number of tests executed on this branch is higher due to the changes to show each of the test invocations for tests that use a data provider. Let me know what you think.

alanfx:
@danberindei updated with the changes we discussed. Tests that are invoked multiple times will report a single `<testcase>` element if all invocations pass, or will report all failures for the test.

danberindei:
Integrated, thanks Alan!

alanfx:
@danberindei I don't see these changes in Infinispan master. Can you check again? Thanks

danberindei:
Sorry @alanfx, I had pushed it to my fork by mistake... it's in now.

alanfx:
No worries @danberindei, and thanks for sticking with me on this!

PR #4851, opened 2017-02-10
ISPN-6942 Listener.observation is not documented

wburns:
https://issues.jboss.org/browse/ISPN-6942

### DIFF
--- a/core/src/main/java/org/infinispan/notifications/Listener.java
+++ b/core/src/main/java/org/infinispan/notifications/Listener.java
@@ -241,22 +241,44 @@ public @interface Listener {
     **/
    boolean includeCurrentState() default false;
 
+   /**
+    * Returns the type of observation level this listener defines.
+    * @return the observation level of this listener
+    * @see Observation
+    * @since 7.2
+    */
    Observation observation() default Observation.BOTH;
 
 
+   /**
+    * Enumeration that defines when a listener event can be observed. A listener can receive an event before and/or
+    * after an operation has occurred.  If the user wishes to be notified just before the operation completes
+    * the listener should observe using {@link Observation#PRE} level.  If the user wishes to be notified only
+    * after the operation completes the listener should observe using {@link Observation#POST} level.  If the user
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/notifications/Listener.java
```diff
@@ -5,6 +5,8 @@
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import org.infinispan.notifications.cachelistener.event.Event;
```
pruivo: unsed import

core/src/main/java/org/infinispan/notifications/Listener.java
```diff
@@ -241,22 +243,46 @@
     **/
    boolean includeCurrentState() default false;
 
+   /**
+    * Returns the type of observation level this listener defines.  A listener can define an observation level to reduce
```
pruivo: instead of repeating the javadoc, what about:
```
Returns the type of observation level this listener defines.
@see Observation
```

core/src/main/java/org/infinispan/notifications/Listener.java
wburns: It is used for the javadoc

core/src/main/java/org/infinispan/notifications/Listener.java
wburns: Sure

core/src/main/java/org/infinispan/notifications/Listener.java
wburns: I can remove that though as well.

pruivo: I didn't see any reference to ```Event``` in your javadoc...

wburns:
Updated

core/src/main/java/org/infinispan/notifications/Listener.java
wburns: It was in the old one, I had removed it

pruivo:
integrated! thanks @wburns

PR #4853, opened 2017-02-12
ISPN-7447 fix for DefaultExecutorService without keys specified.

SetoKaiba:
ISPN-7447 fix for DefaultExecutorService without keys specified.
And also make the with keys specified work like others. Clone the task.getCallable for target equals me.

I don't know what is the clone for. But the Callable is cloned in other submit DistributedTask<T> methods. So I added it as well. Maybe anyone can tell whether the clone is required? I just added it like other submit DistributedTask<T> methods.
For no keys specified, it's submitted to local. And it's reuse the submit(Address, DistributedTask<T>). So the features like failover in DistributedTask will work.

### DIFF
--- a/core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
+++ b/core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
@@ -3,6 +3,7 @@ package org.infinispan.distexec;
 import java.io.Serializable;
 import java.util.List;
 import java.util.Random;
+import java.util.Scanner;
 import java.util.concurrent.*;
 
 import org.infinispan.Cache;
@@ -65,11 +66,11 @@ public class DistributedExecutorFailoverTest extends MultipleCacheManagersTest {
          future.whenComplete((aVoid,throwable) -> {
             Executors.newSingleThreadExecutor().submit((Callable<Void>)() -> {
                 future.get();
-
-                AssertJUnit.assertEquals(sameNodeTaskFailoverPolicy.failoverCount, 2);
                 return null;
             });
          });
+         Thread.sleep(1000);
+         AssertJUnit.assertEquals(sameNodeTaskFailoverPolicy.failoverCount, 2);
       } catch (Exception ex) {
          AssertJUnit.fail("Task did not failover properly " + ex);
       } finally {
--- a/core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
+++ b/core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
@@ -3,8 +3,11 @@ package org.infinispan.distexec;
 import java.io.Serializable;
 import java.util.List;
 import java.util.Random;
-import java.util.Scanner;
-import java.util.concurrent.*;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 import org.infinispan.Cache;
 import org.infinispan.configuration.cache.CacheMode;
@@ -43,7 +46,7 @@ public class DistributedExecutorFailoverTest extends MultipleCacheManagersTest {
    }
 
    public void testBasicLocalDistributedCallable() throws Exception {
-       long taskTimeout = TimeUnit.SECONDS.toMillis(15);
+      long taskTimeout = TimeUnit.SECONDS.toMillis(15);
       EmbeddedCacheManager cacheManager1 = manager(0);
       final EmbeddedCacheManager cacheManager2 = manager(1);
       final EmbeddedCacheManager cacheManager3 = manager(2);
@@ -58,17 +61,24 @@ public class DistributedExecutorFailoverTest extends MultipleCacheManagersTest {
 
          SameNodeTaskFailoverPolicy sameNodeTaskFailoverPolicy = new SameNodeTaskFailoverPolicy(3);
          DistributedTaskBuilder<Void> builder = des
-                 .createDistributedTaskBuilder(new TestCallable())
-                 .failoverPolicy(sameNodeTaskFailoverPolicy)
-                 .timeout(taskTimeout, TimeUnit.MILLISECONDS);
+               .createDistributedTaskBuilder(new TestCallable())
+               .failoverPolicy(sameNodeTaskFailoverPolicy)
+               .timeout(taskTimeout, TimeUnit.MILLISECONDS);
 
          CompletableFuture<Void> future = des.submit(builder.build());
-         try {future.get(10, TimeUnit.SECONDS);} catch(Exception e) {}
+         TriggerFailover(future);
          eventuallyEquals(3, () -> sameNodeTaskFailoverPolicy.failoverCount);
       } catch (Exception ex) {
          AssertJUnit.fail("Task did not failover properly " + ex);
       } finally {
          des.shutdown();
+      }
+   }
+
+   private void TriggerFailover(CompletableFuture<Void> future) {
+      try {
+         future.get(10, TimeUnit.SECONDS);
+      } catch (Exception e) {
       }
    }
 
@@ -163,7 +173,7 @@ public class DistributedExecutorFailoverTest extends MultipleCacheManagersTest {
 
    static class SameNodeTaskFailoverPolicy implements DistributedTaskFailoverPolicy {
       private final int maxFailoverCount;
-      private int failoverCount = 0;
+      private volatile int failoverCount = 0;
 
       public SameNodeTaskFailoverPolicy(int maxFailoverCount) {
          super();
--- a/core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
+++ b/core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
@@ -63,12 +63,7 @@ public class DistributedExecutorFailoverTest extends MultipleCacheManagersTest {
                  .timeout(taskTimeout, TimeUnit.MILLISECONDS);
 
          CompletableFuture<Void> future = des.submit(builder.build());
-         future.whenComplete((aVoid,throwable) -> {
-            Executors.newSingleThreadExecutor().submit((Callable<Void>)() -> {
-                future.get();
-                return null;
-            });
-         });
+         try {future.get(10, TimeUnit.SECONDS);} catch(Exception e) {}
          eventuallyEquals(3, () -> sameNodeTaskFailoverPolicy.failoverCount);
       } catch (Exception ex) {
          AssertJUnit.fail("Task did not failover properly " + ex);
--- a/core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
+++ b/core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
@@ -3,10 +3,7 @@ package org.infinispan.distexec;
 import java.io.Serializable;
 import java.util.List;
 import java.util.Random;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 
 import org.infinispan.Cache;
 import org.infinispan.configuration.cache.CacheMode;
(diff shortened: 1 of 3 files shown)


vblagoje:
@SetoKaiba There is a need for cloning; I wrote this code appx. four-five years ago and I remember the need for cloning, but I did not write the comment why! Also, I could not read your repo where you have put the tests. See how we wrote other DES unit tests and maybe you can adapt your tests to our framework. Perhaps you can add your tests to this PR as well? :-) Thoughts @wburns

SetoKaiba:
@vblagoje But the original implementation doesn't clone. I added it because I take other methods as reference and find others are all with cloning for local.

My repo is deleted because I have found out the bugs.  It's not test case, but for reproducing.
I'll add a TestCase for submit method without keys specified, which is fixed in the patch. But since the original DistributedExecutorFailoverTest is with cluster. I have to shutdown others before this test.

And the future.get is not working only occurs in 8.2.x. But it's difficult for me to fix it. Because it use a delegate NotifyingFutureImpl for LocalDisitributedTaskPart. So I migrated to 9.x.

SetoKaiba:
@vblagoje test case is added.

danberindei:
@SetoKaiba @vblagoje The cloning is needed because the task may have non-`transient` fields that change during execution. If the task starts executing on the local node and marshalling tries to write those fields from another thread, it could cause hard-to-track errors.

I find the case where there are no input keys a bit weird. Normally we execute the task on one of the primary owners of the input keys, but if there are no keys, there are no owners, so IMO we should not execute the task at all.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
```diff
@@ -44,6 +41,42 @@ protected String cacheName() {
...
+                 .createDistributedTaskBuilder(new TestCallable())
+                 .failoverPolicy(sameNodeTaskFailoverPolicy)
+                 .timeout(taskTimeout, TimeUnit.MILLISECONDS);
+
+         CompletableFuture<Void> future = des.submit(builder.build());
+         future.whenComplete((aVoid,throwable) -> {
+            Executors.newSingleThreadExecutor().submit((Callable<Void>)() -> {
+                future.get();
+
+                AssertJUnit.assertEquals(sameNodeTaskFailoverPolicy.failoverCount, 2);
```
danberindei: This needs to happen on the main thread, or TestNG won't report it.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
```diff
@@ -44,6 +41,42 @@ protected String cacheName() {
...
+         SameNodeTaskFailoverPolicy sameNodeTaskFailoverPolicy = new SameNodeTaskFailoverPolicy(2);
+         DistributedTaskBuilder<Void> builder = des
+                 .createDistributedTaskBuilder(new TestCallable())
+                 .failoverPolicy(sameNodeTaskFailoverPolicy)
+                 .timeout(taskTimeout, TimeUnit.MILLISECONDS);
+
+         CompletableFuture<Void> future = des.submit(builder.build());
+         future.whenComplete((aVoid,throwable) -> {
+            Executors.newSingleThreadExecutor().submit((Callable<Void>)() -> {
+                future.get();
```
danberindei: This should be on the main thread, and it should have a timeout, e.g. `future.get(10, SECONDS)`.

SetoKaiba: How can I assert it in the non-main thread? future.whenComplete and open a new thread to trigger the failover is part of test. That's what I need.
Let me explain why this is necessary. The task maybe used to changes the data in the cache. It doesn't need a return value at all. But no future.get, the failover won't be triggered.

SetoKaiba:
And also, no input keys and input cache is normal for my case. Just use it as ClusterExecutor. Use CDI to inject the cache for data modification. And without input keys, it's run locally.

SetoKaiba:
Ok. I know how change it.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
```diff
@@ -3,10 +3,8 @@
 import java.io.Serializable;
 import java.util.List;
 import java.util.Random;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
+import java.util.Scanner;
+import java.util.concurrent.*;
```
slaskawi: Could you please remove star imports? Also, please make sure you're using the latest formatter: https://github.com/infinispan/infinispan/tree/master/ide-settings

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
```diff
@@ -44,6 +42,36 @@ protected String cacheName() {
       return "DistributedExecutorFailoverTest";
    }
 
+   public void testBasicLocalDistributedCallable() throws Exception {
+       long taskTimeout = TimeUnit.SECONDS.toMillis(15);
```
slaskawi: Just a picky one - please remove the space at the beginning.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
```diff
@@ -44,6 +42,36 @@ protected String cacheName() {
...
+         des = new DefaultExecutorService(cache1);
+
+         SameNodeTaskFailoverPolicy sameNodeTaskFailoverPolicy = new SameNodeTaskFailoverPolicy(3);
+         DistributedTaskBuilder<Void> builder = des
+                 .createDistributedTaskBuilder(new TestCallable())
+                 .failoverPolicy(sameNodeTaskFailoverPolicy)
+                 .timeout(taskTimeout, TimeUnit.MILLISECONDS);
+
+         CompletableFuture<Void> future = des.submit(builder.build());
+         try {future.get(10, TimeUnit.SECONDS);} catch(Exception e) {}
```
slaskawi: Could you please extract it to some other method? This way we would get rid of this nasty formatting here.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
```diff
@@ -44,6 +42,36 @@ protected String cacheName() {
...
+
+      try {
+         // initiate task from cache1 and execute on same node
+         des = new DefaultExecutorService(cache1);
+
+         SameNodeTaskFailoverPolicy sameNodeTaskFailoverPolicy = new SameNodeTaskFailoverPolicy(3);
+         DistributedTaskBuilder<Void> builder = des
+                 .createDistributedTaskBuilder(new TestCallable())
+                 .failoverPolicy(sameNodeTaskFailoverPolicy)
+                 .timeout(taskTimeout, TimeUnit.MILLISECONDS);
```
slaskawi: You could probably use `.timeout(15, TimeUnit.SECONDS)` and remove the `taskTimeout` variable.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
```diff
@@ -44,6 +42,36 @@ protected String cacheName() {
...
+
+         CompletableFuture<Void> future = des.submit(builder.build());
+         try {future.get(10, TimeUnit.SECONDS);} catch(Exception e) {}
+         eventuallyEquals(3, () -> sameNodeTaskFailoverPolicy.failoverCount);
+      } catch (Exception ex) {
+         AssertJUnit.fail("Task did not failover properly " + ex);
```
slaskawi: Just rethrow exception from here and TestNG will interpret it correctly.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
```diff
@@ -124,4 +160,25 @@ public int maxFailoverAttempts() {
          return maxFailoverCount;
       }
    }
+
+   static class SameNodeTaskFailoverPolicy implements DistributedTaskFailoverPolicy {
+      private final int maxFailoverCount;
+      private int failoverCount = 0;
+
+      public SameNodeTaskFailoverPolicy(int maxFailoverCount) {
+         super();
```
slaskawi: `super()` seems redundant.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
```diff
@@ -124,4 +160,25 @@ public int maxFailoverAttempts() {
          return maxFailoverCount;
       }
    }
+
+   static class SameNodeTaskFailoverPolicy implements DistributedTaskFailoverPolicy {
+      private final int maxFailoverCount;
```
slaskawi: I think those counters will be updated/read by different thread. Am I right? If so, please make them `volatille`.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
SetoKaiba: I worked with multiple projects and forgot to switch to infinispan code style.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
SetoKaiba: Use the formatter will fix this as well.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
SetoKaiba: No. It's the same with the original test. So I think it should keep the same way to code.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
SetoKaiba: No. It's the same with the original test. So I think it should keep the same way to code.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
SetoKaiba: Ok.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
SetoKaiba: No. It's the same with the original failover policy. So I think it should keep the same way to code.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
SetoKaiba: Okay. Extracted.

SetoKaiba:
@vblagoje @danberindei @slaskawi According to the result from CI, the problem is caused by cloning. But cloning exists in the methods below. 

- submit(Address target, DistributedTask<T> task)
- submit(DistributedTask<T> task, K... input)
- submitEverywhere(DistributedTask<T> task)

So I added cloning to the submit(DistributedTask<T> task, K... input) as well.

But it seems that the cloning will check something. But adding ExternalPojo to every Callable should not be required. The problem is caused by the below isAllowed call.
`assert ExternallyMarshallable.isAllowed(obj) : "Check support for: " + obj.getClass();`

SetoKaiba:
@slaskawi Added ExternalPojo to Callable test cases which is required. And also remove the kill of other managers. The test suite should be okay now.

core/src/test/java/org/infinispan/distexec/DistributedExecutorFailoverTest.java
wburns: Also unfortunately ++ is not atomic (which means you could lose an upate)  You would need to use an AtomicInteger instead which allows for `incrementAndGet`

SetoKaiba: Okay. Fixed. Since slaskawi mis-referred the line. The code isn't shown as outdated.
And also I'd like to know whether infinispan with an atomic integer cluster wide?
Hazelcast is with one. But it seems I can't find it in infinispan.

Before development of my project, I take a look at all the IMDG as well. And I chose infinispan in the end. Because it's more familiar with my previous using stuffs from JBoss, like Weld, WildFly, Netty, Resteasy, Hibernate and so on.

tristantarrant: Counters are going to come in Infinispan 9.1. https://github.com/infinispan/infinispan/pull/4350

danberindei:
Integrated, thanks @SetoKaiba !

I added another commit to make use of our `TestException` and `Exceptions.expectExecutionException()`.

PR #4854, opened 2017-02-13
[perf] client/server experiments

pruivo:
compute hash code lazily instead of when unmarshalling key/values

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
+++ b/commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
@@ -16,11 +16,10 @@ import org.infinispan.commons.util.Util;
  */
 public class WrappedByteArray implements WrappedBytes {
    private final byte[] bytes;
-   private transient int hashCode;
+   private transient int hashCode = 0;
 
    public WrappedByteArray(byte[] bytes) {
       this.bytes = bytes;
-      this.hashCode = Arrays.hashCode(bytes);
    }
 
    public WrappedByteArray(byte[] bytes, int hashCode) {
@@ -74,7 +73,7 @@ public class WrappedByteArray implements WrappedBytes {
 
    @Override
    public int hashCode() {
-      return hashCode;
+      return hashCode == 0 ? (hashCode = Arrays.hashCode(bytes)) : hashCode;
    }
 
    @Override
--- a/core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
@@ -91,12 +91,11 @@ public class TriangleDistributionInterceptor extends NonTxDistributionIntercepto
    private static final Log log = LogFactory.getLog(TriangleDistributionInterceptor.class);
    private static final boolean trace = log.isTraceEnabled();
    private CommandAckCollector commandAckCollector;
+   private final InvocationExceptionFunction onPutMapNoLocalEntriesException = this::onPutMapNoLocalEntriesException;
    private CommandsFactory commandsFactory;
    private TriangleOrderManager triangleOrderManager;
    private Address localAddress;
-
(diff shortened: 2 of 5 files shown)


pruivo:
run performance tests please
cs dist

pruivo:
run performance tests please
cs dist write
library dist write

core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
```diff
@@ -324,37 +356,36 @@ private Object primaryOwnerWrite(InvocationContext context, DataWriteCommand com
...
    }
 
-   private void logCommandSequence(CommandInvocationId id, int segment, long sequence) {
-         log.tracef("Command %s got sequence %s for segment %s", id, sequence, segment);
+   private void sendToBackup(DataWriteCommand command, Collection<Address> backups, int segmentId) {
+      long sequenceNumber = triangleOrderManager.next(segmentId, command.getTopologyId());
+      BackupWriteRcpCommand backupWriteRcpCommand = commandsFactory.buildBackupWriteRcpCommand(command);
+      backupWriteRcpCommand.setSequence(sequenceNumber);
+      if (trace) {
+         logCommandSequence(command.getCommandInvocationId(), segmentId, sequenceNumber, backups);
```
danberindei: What does this pattern buy us? Doesn't the JIT eliminate everything inside `if (trace)` anyway?

core/src/main/java/org/infinispan/util/ByteString.java
```diff
@@ -39,7 +38,7 @@ public static ByteString emptyString() {
 
    @Override
    public int hashCode() {
-      return hash;
+      return hash == 0 ? (hash = Arrays.hashCode(b)) : hash;
```
danberindei: Do we ever create a `ByteString` without looking it up in a map of caches?

commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
```diff
@@ -74,7 +73,7 @@ public boolean equalsWrappedBytes(WrappedBytes other) {
 
    @Override
    public int hashCode() {
-      return hashCode;
+      return hashCode == 0 ? (hashCode = Arrays.hashCode(bytes)) : hashCode;
```
danberindei: Maybe we shouldn't cache the hash code at all? `MurmurHash3.hash(Object)` doesn't use it anyway...

core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
pruivo: > What does this pattern buy us?

headaches...

core/src/main/java/org/infinispan/util/ByteString.java
pruivo: no. the first thing we do when we create it, is to put it in a map or check it in a map :P

commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
pruivo: AFAIK, it is only used to put keys in DataContainer

core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
danberindei: :D

commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
danberindei: I guess it's used for reading as well, so a write operation would benefit from the cached hash code.

OTOH with compressed oops removing the field will remove 8 bytes from `WrappedByteArray`, and keys aren't supposed to be that long anyway, so I would still remove the field.

PR #4855, opened 2017-02-13
ISPN-6176 Rename DummyTransactionManager to EmbeddedTransactionManager

pruivo:
* All dummy* related classes deprecated
* New embedded* related classes added
* Testsuite is using the new classes

https://issues.jboss.org/browse/ISPN-6176

I would like an opinion on
* ```DummyXid``` and ```DummyNoXaXid``` has been kept. should it be changed to a new name?
* There is a ```useXaXid``` boolean that is set when recovery is in use. this boolean choose between the creation of ```DummyNoXaXid```  or ```DummyXid```. IMO, the field should be removed and we should create only ```DummyXid```. Also, I can remove ```RecoveryEmbeddedTransactionManagerLookup```.

### DIFF
--- a/documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
+++ b/documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
@@ -9,6 +9,16 @@ This guide walks you through the process of upgrading Infinispan.
 
 == Upgrading from 8.x to 9.0
 
+=== Deprecated all the dummy related transaction classes.
+The following classes has been deprecated and they will be removed in the future:
+
+* `DummyBaseTransactionManager`: replaced by `EmbeddedBasedTransactionManager`;
+* `DummyNoXaXid` and `DummyXid`: replaced by `EmbeddedXid`;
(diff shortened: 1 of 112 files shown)


tristantarrant:
I think we should rename DummyXid to EmbbededXid
Also, ok to remove DymmyNoXaXid and the useXaXid field and RecoveryEmbeddedTransactionManagerLookup

pruivo:
updated

documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
```diff
@@ -9,6 +9,16 @@ This guide walks you through the process of upgrading Infinispan.
 
 == Upgrading from 8.x to 9.0
 
+=== Deprecated all the dummy related transaction classes.
+The following classes has been deprecated and they will be removed in the future:
```
tristantarrant: s/has/have/

tristantarrant:
Pushed to master, thanks

PR #4856, opened 2017-02-13
ISPN-7464 Make caffeine optional dependency for infinispan core

wburns:
* Make caffeine optional dependency for maven
* Make caffeine internal dependency for as modules

https://issues.jboss.org/browse/ISPN-7464

### DIFF
--- a/wildfly-modules/src/main/resources/org/infinispan/commons/main/module.xml
+++ b/wildfly-modules/src/main/resources/org/infinispan/commons/main/module.xml
@@ -6,7 +6,7 @@
     <dependencies>
         <module name="javax.api"/>
         <module name="javax.transaction.api"/>
-        <module name="com.github.ben-manes.caffeine" slot="${slot}" export="true"/>
+        <module name="com.github.ben-manes.caffeine" slot="${slot}"/>
         <module name="org.jboss.logging"/>
         <module name="org.jboss.marshalling" slot="${slot}" services="import"/>
         <module name="sun.jdk"/>
--- a/commons/src/main/resources/features.xml
+++ b/commons/src/main/resources/features.xml
@@ -5,5 +5,6 @@
       <bundle>mvn:org.jboss.logging/jboss-logging/${version.jboss.logging}</bundle>
       <bundle>mvn:org.jboss.marshalling/jboss-marshalling-osgi/${version.jboss.marshalling}</bundle>
       <bundle>mvn:org.apache.logging.log4j/log4j-api/2.0</bundle>
+      <bundle>mvn:com.github.ben-manes.caffeine/caffeine/${version.caffeine}</bundle>
    </feature>
 </features>
--- a/wildfly-modules/src/main/resources/org/infinispan/client/hotrod/main/module.xml
+++ b/wildfly-modules/src/main/resources/org/infinispan/client/hotrod/main/module.xml
@@ -8,6 +8,7 @@
         <module name="javax.api"/>
         <module name="org.apache.commons.pool"/>
         <module name="org.infinispan.commons" slot="${slot}"/>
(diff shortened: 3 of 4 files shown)


wburns:
Wait to make sure build all passes so far

wburns:
Updated commit to point to correct JIRA.  Branch unfortunately is still old, left alone to keep PR as is.

commons/pom.xml
```diff
@@ -17,6 +17,7 @@
       <dependency>
          <groupId>com.github.ben-manes.caffeine</groupId>
          <artifactId>caffeine</artifactId>
+         <optional>true</optional>
```
Sanne: Can you clarify this. Would `infinispan-commons` work fine when caffeine is missing?

wburns: It would work, except for the method CollectionFactory#makeBoundedConcurrentMap

wildfly-modules/src/main/resources/org/infinispan/commons/main/module.xml
```diff
@@ -6,7 +6,7 @@
     <dependencies>
         <module name="javax.api"/>
         <module name="javax.transaction.api"/>
-        <module name="com.github.ben-manes.caffeine" slot="${slot}" export="true"/>
+        <module name="com.github.ben-manes.caffeine"/>
```
Sanne: thanks, that is the line I was wondering about when I asked on IRC if that was necessary. Apparently not :)

commons/pom.xml
wburns: Although I am not sure if the CollectionFactory class itself would work since it has an import for Caffeine.

wildfly-modules/src/main/resources/org/infinispan/commons/main/module.xml
wburns: Yeah you can chalk that up to me not knowing what to do with wildfly-modules :)

client/hotrod-client/pom.xml
```diff
@@ -25,6 +25,11 @@
       </dependency>
 
       <dependency>
+         <groupId>com.github.ben-manes.caffeine</groupId>
+         <artifactId>caffeine</artifactId>
+      </dependency>
```
Sanne: why does the client need caffeine?

wburns: The Near cache in the client is a bounded concurrent map.

commons/pom.xml
Sanne: In other words: it would be great if infinispan-commons would avoid having this dependency, then we'd have reassurances from the compiler.

Sanne: ic. Are there many use cases to have `infinispan-commons`, yet no have caffeine then?
If all (or most) uses will need this, one might as well keep it as a non-optional dependency.

wildfly-modules/src/main/resources/org/infinispan/commons/main/module.xml
Sanne: as discussed on IRC: the slot qualifier needs to stay.

wburns:
Updated to leave maven pom alone.  Also made sure slot is defined in case of different versions of Caffeine are deployed together

wburns:
I will also remove the modules from core and client to see if they pass tests without.  Before I had to export them, maybe they are working without now.

wburns:
Seems the other imports are required, adding them back in.

Sanne:
Last one looks good. I'll wait CI to have all results, as I only run a specific subset of tests.

Sanne:
"All checks have failed" ..
I'll leave this to someone who knows what to look for. +1 to merge!

tristantarrant:
Pushed to master

PR #4858, opened 2017-02-14
Stabilized DistributedExecutorCDITest

slaskawi:
https://issues.jboss.org/browse/ISPN-7474

### DIFF
--- a/cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/distexec/DistributedExecutorCDITest.java
+++ b/cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/distexec/DistributedExecutorCDITest.java
@@ -89,8 +89,7 @@ public class DistributedExecutorCDITest extends MultipleCacheManagersArquillianT
       }
    }
 
-   static class ImpliedInputCacheCallable implements Callable<Integer>, Serializable {
-
+   static class ImpliedInputCacheCallable implements Callable<Integer>, Serializable, ExternalPojo {
 
       /** The serialVersionUID */
       private static final long serialVersionUID = 5770069398989111268L;
@@ -108,8 +107,7 @@ public class DistributedExecutorCDITest extends MultipleCacheManagersArquillianT
       }
    }
 
-   static class ImpliedInputCacheRunnable implements Runnable, Serializable {
-
+   static class ImpliedInputCacheRunnable implements Runnable, Serializable, ExternalPojo {
 
       /** The serialVersionUID */
       private static final long serialVersionUID = 5770069398989111268L;


cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/distexec/DistributedExecutorCDITest.java
```diff
@@ -89,11 +83,7 @@ public void run() {
       }
    }
 
-   static class ImpliedInputCacheCallable implements Callable<Integer>, Serializable {
-
-
-      /** The serialVersionUID */
-      private static final long serialVersionUID = 5770069398989111268L;
+   static class ImpliedInputCacheCallable implements Callable<Integer>, Serializable, ExternalPojo {
```
slaskawi: Adding `ExternalPojo` was crucial here. But I'm not sure why to be honest, @galderz any opinions?

I can see it's a consequence of this commit https://github.com/infinispan/infinispan/commit/82d268bdea0b2bb3bef4eed1a8bda4ceeaf91409. So I'm wondering what is the impact for the end user? If one creates a new `Callable` and submits it to our `DistributedExecutorService`, does it mean that the submitted class needs to implement `ExternalPojo` as well?

danberindei: @slaskawi There should be no consequence for user classes, `ExternalPojo` is only needed for our test classes (see `ExternallyMarshallable.isAllowed()` and `ExternallyMarshallable.isMarshallablePackage()`).

cdi/embedded/src/main/java/org/infinispan/cdi/embedded/DelegatingDistributedTaskLifecycle.java
```diff
@@ -44,13 +45,19 @@ public DelegatingDistributedTaskLifecycle() {
...
+            BeanManager bm = CDIHelper.getBeanManager();
+            if (bm == null)
+               return;
+            ContextInputCache.set(inputDataCache);
+            preInject(bm, task);
+            if (task instanceof RunnableAdapter) {
+               preInject(bm, ((RunnableAdapter) task).getTask());
+            }
+         } catch (Exception e) {
+            throw new IllegalStateException("Injcting input caches failed. This may result in NullPointerExceptions in client code", e);
```
danberindei: Typo: Injcting

Also, I think continuing to run the task after throwing the `IllegalStateException` would be a bug, as the NPE in the user code would hide the ISE.

cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/distexec/DistributedExecutorCDITest.java
```diff
@@ -62,9 +62,6 @@ public void testInvocationUsingImpliedInputCacheRunnable() throws Exception {
 
    static class SimpleCallable implements Callable<Integer>, Serializable, ExternalPojo {
 
-      /** The serialVersionUID */
-      private static final long serialVersionUID = -8589149500259272402L;
```
danberindei: Why remove the id if the class is still `Serializable`?

cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/distexec/DistributedExecutorCDITest.java
slaskawi: Thanks @danberindei for clarifying this!

cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/distexec/DistributedExecutorCDITest.java
slaskawi: The compiler will generate one for us. Plus the code is shorter.

cdi/embedded/src/main/java/org/infinispan/cdi/embedded/DelegatingDistributedTaskLifecycle.java
slaskawi: Granted. Let me revert this change.

cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/distexec/DistributedExecutorCDITest.java
danberindei: AFAIK the compiler doesn't generate it automatically, it's computed on the fly:
stackoverflow.com/questions/29081134/why-compiler-do-not-set-serialversionuid-automatically

But the real reason to keep it is that IntelliJ will show an annoying warning if the class is `Serializable` and doesn't have a `serialVersionUID` :)

danberindei:
@slaskawi No JIRA?

slaskawi:
@danberindei I though this will be a quick one. Creating JIRA.

cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/distexec/DistributedExecutorCDITest.java
slaskawi: There's a tiny little option to disable the warning you know :D

slaskawi: You actually intrigued me with this... and I think the SO comment is not quite accurate...

So the `Serializable` class [Javadoc](http://docs.oracle.com/javase/7/docs/api/java/io/Serializable.html) states:

> If a serializable class does not explicitly declare a serialVersionUID, then the serialization runtime will calculate a default serialVersionUID value for that class based on various aspects of the class, as described in the Java(TM) Object Serialization Specification. However, it is strongly recommended that all serializable classes explicitly declare serialVersionUID values, since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations, and can thus result in unexpected InvalidClassExceptions during deserialization. Therefore, to guarantee a consistent serialVersionUID value across different java compiler implementations, a serializable class must declare an explicit serialVersionUID value.

Now the [Java Serialization Specification](http://docs.oracle.com/javase/8/docs/platform/serialization/spec/class.html#a4100) states:

> The serialVersionUID is computed using the signature of a stream of bytes that reflect the class definition. The National Institute of Standards and Technology (NIST) Secure Hash Algorithm (SHA-1) is used to compute a signature for the stream. The first two 32-bit quantities are used to form a 64-bit hash. A java.lang.DataOutputStream is used to convert primitive data types to a sequence of bytes. The values input to the stream are defined by the Java Virtual Machine (VM) specification for classes. Class modifiers may include the ACC_PUBLIC, ACC_FINAL, ACC_INTERFACE, and ACC_ABSTRACT flags; other flags are ignored and do not affect serialVersionUID computation. Similarly, for field modifiers, only the ACC_PUBLIC, ACC_PRIVATE, ACC_PROTECTED, ACC_STATIC, ACC_FINAL, ACC_VOLATILE, and ACC_TRANSIENT flags are used when computing serialVersionUID values. For constructor and method modifiers, only the ACC_PUBLIC, ACC_PRIVATE, ACC_PROTECTED, ACC_STATIC, ACC_FINAL, ACC_SYNCHRONIZED, ACC_NATIVE, ACC_ABSTRACT and ACC_STRICT flags are used. Names and descriptors are written in the format used by the java.io.DataOutputStream.writeUTF method.

So yes, the `serialVersionUID` is computed in the runtime but based on "the signature of a stream of bytes that reflect the class definition", so based on compiler's output. I believe that's why they say "it is strongly recommended that all serializable classes explicitly declare serialVersionUID values, since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations". 

And finally, [this question](http://stackoverflow.com/questions/29081134/why-compiler-do-not-set-serialversionuid-automatically) doesn't make sense I think (assuming that there's no bug in JVM in serialization runtime of course!).

tristantarrant:
CI says /opt/buildAgent/work/e7918bceb48c007e/cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/distexec/DistributedExecutorCDITest.java:[114,33] variable serialVersionUID is already defined in class org.infinispan.cdi.embedded.test.distexec.DistributedExecutorCDITest

slaskawi:
@tristantarrant Fixed

cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/distexec/DistributedExecutorCDITest.java
danberindei: True, the question is kind of pointless, because the OP is worried about compatibility of his serialized class across different JDK versions/implementations, and that's what the JLS is for.

I definitely wouldn't think twice about integrating a PR that added a `Serializable` class without a `serialVersionUID`. But since it was already here, I don't think there's anything to gain from removing it, either.

slaskawi:
The failure is gone. The code is approved. Merging...

PR #4859, opened 2017-02-14
[perf] wrapped bytes experiement

pruivo:
I've removed the hashcode field. I'm wondering if we should have 2 implementation: one for keys (with the field pre-computed) and another for values (without the field). @danberindei  what do you think?

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
+++ b/commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
@@ -16,7 +16,7 @@ import org.infinispan.commons.util.Util;
  */
 public class WrappedByteArray implements WrappedBytes {
    private final byte[] bytes;
-   private transient int hashCode;
+   private final int hashCode;
 
    public WrappedByteArray(byte[] bytes) {
       this.bytes = bytes;
@@ -100,12 +100,12 @@ public class WrappedByteArray implements WrappedBytes {
       @Override
       public void writeObject(ObjectOutput output, WrappedByteArray object) throws IOException {
          MarshallUtil.marshallByteArray(object.bytes, output);
+         output.writeInt(object.hashCode);
       }
 
       @Override
       public WrappedByteArray readObject(ObjectInput input) throws IOException, ClassNotFoundException {
-         byte[] bytes = MarshallUtil.unmarshallByteArray(input);
-         return new WrappedByteArray(bytes);
+         return new WrappedByteArray(MarshallUtil.unmarshallByteArray(input), input.readInt());
       }
    }
 }
--- a/commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
+++ b/commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
@@ -16,9 +16,16 @@ import org.infinispan.commons.util.Util;
  */
 public class WrappedByteArray implements WrappedBytes {
    private final byte[] bytes;
+   private transient int hashCode;
 
    public WrappedByteArray(byte[] bytes) {
       this.bytes = bytes;
+      this.hashCode = Arrays.hashCode(bytes);
+   }
+
+   public WrappedByteArray(byte[] bytes, int hashCode) {
+      this.bytes = bytes;
+      this.hashCode = hashCode;
    }
 
    @Override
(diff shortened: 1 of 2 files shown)


pruivo:
run performance tests please
library
cs

Sanne:
> I'm wondering if we should have 2 implementation: one for keys (with the field pre-computed) and another for values (without the field). 

Interesting idea!
Remember to check with [jol](http://openjdk.java.net/projects/code-tools/jol/) If making such a change actually wins you any space. As far as I remember, this change should save you 8 bytes per instance!
The drawback would be with polymorhic calls cost, but I assume you don't really need a shared interface?

Sanne:
@pruivo BTW such a change would get my +1 just becasue it makes sense, independently from it being confirmed by Ci performance tests.
If the performance tests don't approve, there's something fishy to look at why it wouldn't improve.

(less allocations is always better, but it might throw off some other tuning point out of balance)

Sanne:
some more context about JOL: http://lists.jboss.org/pipermail/infinispan-dev/2013-December/014311.html

pruivo:
well, it already implements a shared interface ```public class WrappedByteArray implements WrappedBytes``` but most of our code uses the class directly. 
But I'm not sure if it is worth. The hash code should be used mostly one time when put/retrieve from ```DataContainer```. Our consistent hash uses the bytes directly.

danberindei:
@pruivo I don't think I'd want a different class for the keys and the values.

commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
```diff
@@ -63,25 +56,19 @@ public boolean equals(Object o) {
...
    public boolean equalsWrappedBytes(WrappedBytes other) {
-      int length = getLength();
-      if (other.getLength() != length) return false;
-      if (other.hashCode() != hashCode()) return false;
-      for (int i = 0; i < length; ++i) {
-         if (getByte(i) != other.getByte(i)) return false;
-      }
-      return true;
+      //Array.equals already checks the length and hashcode is lazy now
+      return Arrays.equals(bytes, other.getBytes());
```
wburns: Unfortunately we can't do this.  There is no guarantee that getBytes will return a byte[].  For example when using off heap this will always be null.

commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
```diff
@@ -16,16 +16,9 @@
  */
 public class WrappedByteArray implements WrappedBytes {
    private final byte[] bytes;
-   private transient int hashCode;
```
wburns: What was the motivation here?  Just to save 8 bytes per instance?  The hashCode will be called on every node at least once since it has to be actually put into the container.  Is this to avoid calling hashCode for values?

Sadly for off heap the hashCode is always done even for values before hand and caching it can save a lot of time, since it is required for values to be computed.

wburns: Although this still would slow down off heap quite a bit since the hashCode would have to be computed again multiple times.

wburns:
The change to not have the hashCode could save at most 8 bytes.  It would average out to 4 bytes.  The problem is the byte[] is not a predefined size so it might be less than the alignment size the and the int might fix perfectly in the unused alignment space.

Sanne:
Hu? Come confusion here. It's not about averages, and doesn't depend on the actual byte array size. It will save 8 bytes for each instance.

commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
wburns: Actually I take that back, this shouldn't be changed.  The purpose of this method is to allow for something without a backing array to test equality.  This will be useful when off heap comes back with an instance.  I would say either leave it alone, throw an exception or remove this method.

wburns:
Yes sorry, scatter brained atm, the padding is there no matter what other values are present.

wburns:
`
But I'm not sure if it is worth. The hash code should be used mostly one time when put/retrieve from DataContainer. Our consistent hash uses the bytes directly.
`
Yeah unfortunately due to the client it has to :(

wburns:
I would say for now we can just remove the field and not worry about it.  The equalsWrappingBytes method should just be removed for now as well.  It can be reevaluated again if we ever link off heap to other parts of the system

pruivo:
run performance tests please
cs

commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
pruivo: I can leave it alone but I don't see how to test for equality without the array...

pruivo:
@Sanne @wburns I've checked with jol and it can save 8 or 16 bytes in a 64 bits jvm (depends of the configuration)

pruivo:
trying another approach: sending the computed hash code over the network

pruivo:
run performance tests please
cs

Sanne:
> trying another approach: sending the computed hash code over the network

why ? Is it to avoid re-computing, or to address some issue?

pruivo:
@Sanne performance is much worst

Sanne:
> performance is much worst

I'm not surprised. We still don't have a system which gets you a magic number to let you know if the change is good or bad ;)

It doesn't mean it's a bad idea though, nor that you should abandon it. You need to figure out why the number changed this way.

wburns:
> performance is much worst
Is worse than what?  Than master?  If so that sounds very fishy.  I don't see the completed run report, is that where you are basing that from?

jmarkos:
run performance tests please
cs

pruivo:
run performance tests please
cs

Holmistr:
@pruivo C/S tests are broken right now because of enhancements being applied (selecting different settings.xml). I will retrigger it once fixed.

jmarkos:
run performance tests please
cs

Holmistr:
run performance tests please
cs

PR #4861, opened 2017-02-15
ISPN-5021 Rename some methods

rvansa:
Although this change doesn't make much sense alone, I was asked to file it as a PR separate from #4827 .

No functional changes so far.


rvansa:
Dan does not find the split to two PRs necessary (as do I).

PR #4862, opened 2017-02-15
ISPN-5021 Nodes that finish the rebalance later can see outdated values

rvansa:
Replacement for #4827 , depends on #4861


rvansa:
As per http://lists.jboss.org/pipermail/infinispan-dev/2017-February/017165.html I've decided to use the term "rebalance" for the whole operation that involves topology changes and data transfer.

rvansa:
@danberindei Hi, rebased and addressed those comments I haven't replied to in #4827 
I haven't removed REBALANCE_START, maybe in another PR.

rvansa:
After few hours running the failing tests in loop, I was not able to reproduce the failure :-/ (haven't checked the CDI one, though)

tristantarrant:
And now this needs a rebase...

rvansa:
Rebased.

rvansa:
Oops, still have some trouble with one of the tests, due to increasing staggered delay. I'll see if I'll rather tweak the test or the code...

danberindei:
@rvansa I also see some new test failures in the hotrod-client module:
http://ci.infinispan.org/viewLog.html?buildId=50556&buildTypeId=bt9

rvansa:
@danberindei Okay, solved the problem, though it's a bit hackish.

rvansa:
Ah, those infinispan-server-hotrod failures are in places which expect topology with certain number. I wonder why this didn't fail before. I'll fix them tomorrow.

pferraro:
@rvansa Can we additionally backport this to 8.2.x?

rvansa:
@pferraro I don't see any technical reason why not. I would proceed cautiously, though, because I could have introduced another bug, so it would be better to run resilience stress tests on master. While discovering a regression soon after major release is sort of acceptable, adding that in n-th bugfix would be unfortunate.

rvansa:
It seems this does not get to 9.0.0 :( Scheduling for 9.1

danberindei:
Needs rebase. Also last build had some new failures:

http://ci.infinispan.org/viewLog.html?buildId=51438&tab=buildResultsDiv&buildTypeId=bt9

rvansa:
Rebased, there will be one failure in AtomicMapStateTransferTest, this is handled in #4947

rvansa:
Rebased. I believe that the test failure in `ReplCommandRetryTest` is JGroups-related, composing a testcase for Bela.

@danberindei Could you please set up some date when you'll have a chance to do the review? Rebasing is not fun.

rvansa:
@danberindei Monday is a good day for a review.

rvansa:
Hi @pruivo, would you be so kind and give me a review of this PR? I'd like to get this in in early 9.1 stage to have some time to resolve any potential issues. And I already made scattered cache (scheduled for 9.1) dependent on this, so let's work on making it within the timeframe.

rvansa:
For the record, I believe that the issue causing failure in `ReplCommandRetryTest` is https://issues.jboss.org/browse/JGRP-2162

danberindei:
Ok @rvansa, you almost distracted me again with the reference to JGRP-2162, but now I'm really reviewing this! :D

core/src/main/java/org/infinispan/topology/CacheTopology.java
```diff
@@ -123,6 +126,22 @@ else if (currentCH != null)
...
+         case READ_OLD_WRITE_ALL:
+            assert pendingCH != null;
+            assert unionCH != null;
+            return currentCH;
+         case READ_ALL_WRITE_ALL:
+            assert pendingCH != null;
+            return unionCH;
+         case READ_NEW_WRITE_ALL:
+            assert unionCH != null;
+            return pendingCH;
```
danberindei: I think this would be a good time to rename `currentCH` to `oldCH` and `pendingCH` to `newCH`. But maybe in a separate PR :)

core/src/main/java/org/infinispan/util/logging/events/Messages.java
```diff
@@ -29,8 +29,8 @@
    @Message(value = "Started local rebalance", id = 100002)
    String rebalanceStarted();
 
-   @Message(value = "Finished local rebalance", id = 100003)
-   String rebalanceCompleted();
+   @Message(value = "Finished local rebalance phase", id = 100003)
+   String rebalancePhaseConfirmed();
```
danberindei: It's weird that none of these messages have parameters... I'd change it to "Node %s finished receiving state for rebalance %d" (with CacheTopologyControlCommand.rebalanceId), and the one before to "Started rebalance %d" (it seems to be used for the REBALANCE_START broadcast, so it's not local).

core/src/test/java/org/infinispan/distribution/rehash/NonTxPrimaryOwnerBecomingNonOwnerTest.java
```diff
@@ -124,14 +124,9 @@ private void doTest(final TestWriteOperation op) throws Exception {
...
-         @Override
-         public boolean isSatisfied() throws Exception {
-            return cache0.getRpcManager().getMembers().size() == 3 &&
-                  cache1.getRpcManager().getMembers().size() == 3 &&
-                  cache2.getRpcManager().getMembers().size() == 3;
-         }
-      });
+      eventually(() -> cache0.getRpcManager().getMembers().size() == 3 &&
+            cache1.getRpcManager().getMembers().size() == 3 &&
+            cache2.getRpcManager().getMembers().size() == 3);
```
danberindei: Could be replaced with 3 `eventuallyEquals(3, () -> cacheX.getRpcManager().getMembers().size())` calls.

core/src/test/java/org/infinispan/distribution/rehash/NonTxPrimaryOwnerBecomingNonOwnerTest.java
```diff
@@ -146,46 +141,38 @@ public boolean isSatisfied() throws Exception {
...
       checkPoint.trigger("allow_topology_" + postJoinTopologyId + "_on_" + address(0));
-      eventually(new Condition() {
-         @Override
-         public boolean isSatisfied() throws Exception {
-            CacheTopology cacheTopology = cache0.getComponentRegistry().getStateTransferManager().getCacheTopology();
-            return cacheTopology.getTopologyId() == postJoinTopologyId;
-         }
-      });
+      eventually(() -> cache0.getComponentRegistry().getStateTransferManager()
+            .getCacheTopology().getTopologyId() == postJoinTopologyId);
```
danberindei: `eventuallyEquals` would work here, too.

core/src/test/java/org/infinispan/statetransfer/RemoteGetDuringStateTransferTest.java
```diff
@@ -54,40 +63,88 @@
...
 
-   T = initial topology, T1  = rebalance started but we have no data yet, T1.5 rebalance started and we have data, T2  = rebalance finished
-
-   The originator will retry the remote get if node 1 processes the request in T2.
-   If node 1 processes the request in T0 or T1, it still has the entry, and the originator doesn't have to retry.
+   T0   initial topology
+   T1   state transfer started
+   T2   state transfer finished but rebalance not complete
+   T3   read new, write all topology
+   T4   rebalance completed
```
danberindei: I had a few comments on the old PR here that you didn't handle, e.g. https://github.com/infinispan/infinispan/pull/4827#discussion-diff-101270697R70

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -350,132 +336,160 @@ public void doConfirmRebalance(Address member, int receivedTopologyId) throws Ex
...
          }
-         CacheTopology newTopology = new CacheTopology(topologyId + 1, rebalanceId, newCurrentCH, newPendingCH, actualMembers, persistentUUIDManager.mapAddresses(actualMembers));
-         setCurrentTopology(newTopology);
-
-         clusterTopologyManager.broadcastTopologyUpdate(cacheName, newTopology, availabilityMode,
-               isTotalOrder(), isDistributed());
       }
+      // Losing members during state transfer could lead to a state where we have more than two topologies
+      // concurrently in the cluster. We need to make sure that all the topologies are compatible (properties set
+      // in CacheTopology docs hold) - we just remove lost members.
```
danberindei: I think we need a new phase for the topology that has removed one member before starting the rebalance, otherwise there will be a mismatch between the meaning of `Phase.STABLE` and a stable topology as used by partition handling.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -192,15 +193,24 @@ public final Object visitClearCommand(InvocationContext ctx, ClearCommand comman
...
+      if (responses.values().stream().anyMatch(UnsureResponse.class::isInstance)) {
+         // We got only unsure responses, as all nodes that were read-owners at the time when we've sent
+         // the request have progressed to newer topology. However we are guaranteed to have progressed
+         // to a topology at most one older, and can immediately retry.
+         return OutdatedTopologyException.getCachedInstance();
+      } else {
+         // Another instance when we don't get any successful response is when all owners are lost. We'll handle
+         // this later in StateTransferInterceptor, as we have to signal this to PartitionHandlingInterceptor
+         // if that's present.
+         return AllOwnersLostException.INSTANCE;
```
rvansa: If there was an exception, it would be thrown somewhere from JGroupsTransport, this code won't handle any ExceptionResponse instances.

I don't see why we need another delegation method (I would rather expose `OTE.INSTANCE` directly), but I'll comply in this PR and correct it later.

WRT PartitionHandlingInterceptor - checking "Why the hell did I get this exception" sounds prone to races to me; I prefer to signal the condition directly. We could give `AOLE` and `OTE` a common ancestor, if we want to spare some characters in `STI`.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
rvansa: Are you talking just about nomenclature mismatch, or actual bug?

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: > If there was an exception, it would be thrown somewhere from JGroupsTransport, this code won't handle any ExceptionResponse instances.

Fair enough, but it's not very clear from the code.

> [...] I would rather expose OTE.INSTANCE directly [...]

I'm also ok with removing the delegation method from OTE, I just like code that does the same thing to look the same :)

> [...] checking "Why the hell did I get this exception" sounds prone to races to me [...]

I think PHI is still prone to races, if the partition check succeeds it will throw the `AOLE` back to the user instead of an `AvailabilityException`. And we don't want the user to see `AOLE` any more then we want them to see `OutdatedTopologyException`... even with partition handling disabled, the user should see a `null` value instead of an exception.

IMO `OTE` shouldn't be so surprising in a degraded cache, because read requests are sent based on the (outdated, sort of) stable topology's CH. I agree `AOLE` makes things a bit clearer, but ignoring the exception name, the implementation should be the same.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
danberindei: Just a nomenclature mismatch at this point, but a bug is sure to follow when there's a naming confusion :)

core/src/main/java/org/infinispan/topology/CacheTopologyControlCommand.java
```diff
@@ -36,8 +36,8 @@
       JOIN,
       // A member is signaling that it wants to leave the cluster.
       LEAVE,
-      // A member is confirming that it finished the rebalance operation.
-      REBALANCE_CONFIRM,
+      // A member is confirming that it has finished the topology change during rebalance.
+      REBALANCE_PHASE_CONFIRM,
```
rvansa: The confirmation is sent after each phase, not just after state transfer. You have defined the terms in this way:

> state transfer is the actual transfer of the data, and the rebalance is the whole process of adding
a "pending" consistent hash, starting the state transfer, confirming the transfer of state, and finally replacing the "current" consistent hash with the pending one.

So we're confirming one of the phases of rebalance, not just state transfer.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: Both AOLE and OTE will be always caught by STI, if these aren't, it's a bug in STI.

I think I had a case in `NumOwnersNodeCrashInSequenceTest` where I couldn't handle OTEs in PHI, and I had to create AOLEs - I wouldn't bother with another exception type just to make that more readable for you :)

danberindei: I'd assume you want to make it readable for yourself in the future, too :)

core/src/main/java/org/infinispan/topology/CacheTopologyControlCommand.java
danberindei: Ok, I missed that you're sending a confirmation for each phase. Can you update the comment to mention this?

rvansa: I think the comment  describes exactly that: finishing topology change = change from topology with one id to second id. I could be more verbose, saying that this doesn't occur when the target topology is STABLE one.

danberindei: 1. To me, a topology change is just installing a new topology, it doesn't include requesting or receiving new state.
2. Saying "*the* topology change during rebalance" implies that only one topology change is being confirmed per rebalance.

rvansa: 1. You don't request/receive new state, but IIRC you block sending another topology until you receive the confirmation of installment from all live nodes. The member confirms that exactly what you describe happened.
2. ok, I admit I use the first article that comes to my mind if at all :)

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -398,27 +402,25 @@ public void onTopologyUpdate(final CacheTopology cacheTopology, final boolean is
...
-                     log.tracef("Unlock State Transfer in Progress for topology ID %s", cacheTopology.getTopologyId());
-                  }
-                  if (isTotalOrder) {
-                     totalOrderManager.notifyStateTransferEnd();
-                  }
+               (Object)stateTransferTopologyId, startRebalance, cacheTopology.getPendingCH());
+         if (stateTransferTopologyId != NO_STATE_TRANSFER_IN_PROGRESS && !startRebalance && cacheTopology.getPhase() != CacheTopology.Phase.READ_OLD_WRITE_ALL) {
+            // we have received a topology update that does not start a state transfer, and neither is a members update
+            // due to crashed member during state transfer
+            boolean changed = this.stateTransferTopologyId.compareAndSet(stateTransferTopologyId, NO_STATE_TRANSFER_IN_PROGRESS);
```
rvansa: Maybe in another PR, if you don't mind.

rvansa:
Addressed with :heart:

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
danberindei: Sure.

core/src/main/java/org/infinispan/interceptors/impl/BaseRpcInterceptor.java
```diff
@@ -65,20 +78,25 @@ public void init() {
 
    private void initRpcOptions() {
       // This is a simplified state-less version of ClusteredGetResponseValidityFilter
```
danberindei: The comment is still here :)

danberindei:
Integrated, thanks Radim!

rhusar:
Awesome! @rvansa as I mentioned earlier, can we now get this backported to 8.2.x stream so we can run our tests and see if we are affected?

rhusar:
@rvansa Any chance of producing a backport PR for 8.2?

PR #4863, opened 2017-02-15
[perf] Another TriangleOrderManager

pruivo:
another triangle order manager that doesn't depend on the number of segment but relies on regular (FIFO messages)

pros: doesn't depends on the number of segments
cons: relies on FIFO messages
not sure if pros or cons: less threads (1 deliver thread per primary owner)


pruivo:
run performance tests please
library dist writes

pruivo:
run performance tests please
[UDP 1k]
library dist writes

Holmistr:
@pruivo The last results are still with 10kb entries, as it says in the description of the report. The reason is that the syntax is [UDP 1kb] instean of [UDP 1k] (missing 'b'). I'm retriggering it.

Holmistr:
run performance tests please
[UDP 1kb]
library dist writes

pruivo:
ops my bad @Holmistr. at least the results are consistent :)

pruivo:
run performance tests please
library

PR #4864, opened 2017-02-16
[ISPN-7480] JDBC cache store doesn't work on Sybase

Holmistr:
See: https://issues.jboss.org/browse/ISPN-7480

The logger was probably a copy&paste error, hence correcting it together.


tristantarrant:
Merged to master

PR #4868, opened 2017-02-20
DMR operation register-proto-schemas fails with NPE if the proto file has syntax errors

danberindei:
https://issues.jboss.org/browse/ISPN-7405

Preload is also problematic, both when the file has syntax/semantic errors and when it doesn't (because the {{.errors}} file is locked either way).


core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -62,6 +62,7 @@
 import org.infinispan.remoting.rpc.ResponseMode;
 import org.infinispan.remoting.rpc.RpcOptions;
 import org.infinispan.remoting.transport.Address;
+import org.infinispan.remoting.transport.jgroups.SuspectException;
```
anistor: unused import?

anistor:
Looks good to me. I'll merge it after I see the CI results.

anistor:

danberindei:
Thanks for fixing the imports, Adrian ;)

PR #4869, opened 2017-02-20
ISPN-7496 Ignore Leavers when processing StateTransfer

slaskawi:
https://issues.jboss.org/browse/ISPN-7496

### DIFF
--- a/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
+++ b/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
@@ -26,7 +26,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.locks.LockSupport;
 
 import javax.transaction.SystemException;
 import javax.transaction.TransactionManager;
@@ -165,7 +164,8 @@ public class StateConsumerImpl implements StateConsumer {
 
    private volatile boolean ownsData = false;
 
-   private RpcOptions rpcOptions;
+   private RpcOptions synchronousRpcOptions;
+   private RpcOptions synchronousIgnoreLeaversRpcOptions;
 
    public StateConsumerImpl() {
    }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -652,7 +651,7 @@ public void start() {
       isFetchEnabled = (mode.isDistributed() || mode.isReplicated()) &&
               (configuration.clustering().stateTransfer().fetchInMemoryState() || configuration.persistence().fetchPersistentState());
       //rpc options does not changes in runtime. we can use always the same instance.
-      rpcOptions = rpcManager.getRpcOptionsBuilder(ResponseMode.SYNCHRONOUS)
+      rpcOptions = rpcManager.getRpcOptionsBuilder(ResponseMode.SYNCHRONOUS_IGNORE_LEAVERS)
```
slaskawi: Those options are used to obtain `ClusterListeners`. This operation is safe for sure (since they [fail without any exception](https://github.com/infinispan/infinispan/blob/c6b1c4dbca3eb65f70358d3f2e91b8f550dd5592/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java#L831-L831)). 

However it's also used for [transactions](https://github.com/infinispan/infinispan/blob/c6b1c4dbca3eb65f70358d3f2e91b8f550dd5592/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java#L848-L848). I would still say it's safe since after all we already lost a node...

rvansa: Maybe you could rename that field to `syncIgnoreLeavers` to make other potential uses aware.

slaskawi: Good idea!

slaskawi:
I decided to split obtaining transactions (with `SYNCHRONOUS` option) and listeners (with `SYNCHRONOUS_IGNORE_LEAVERS`) to be sure that I won't do any harm.

I also tried to write a test but it seems pointless since in case of failure, listeners just emit a warning and move on.

danberindei:
@wburns could you have a look?

wburns:
@danberindei seems fine to me.

pruivo:
all tests ok locally, but the core testsuite looks slower. pushing...

pruivo:
integrated! thanks @slaskawi

PR #4871, opened 2017-02-21
ISPN-7487 Fixed NPE in ClusterCacheStatus.java:147

slaskawi:
https://issues.jboss.org/browse/ISPN-7487

### DIFF
--- a/core/src/main/java/org/infinispan/partitionhandling/impl/PreferConsistencyStrategy.java
+++ b/core/src/main/java/org/infinispan/partitionhandling/impl/PreferConsistencyStrategy.java
@@ -179,10 +179,8 @@ public class PreferConsistencyStrategy implements AvailabilityStrategy {
          actualMembers.retainAll(mergedTopology.getMembers());
          mergedAvailabilityMode = AvailabilityMode.DEGRADED_MODE;
       } else {
-         log.debugf("No current topology, recovered only joiners for cache %s", context.getCacheName());
-         mergedTopology = null;
-         actualMembers = null;
-         mergedAvailabilityMode = AvailabilityMode.AVAILABLE;
+         log.debugf("No current topology, recovered only joiners for cache %s. Skipping availability update.", context.getCacheName());
+         return;
       }
 
       // Increment the topology id so that it's bigger than any topology that might have been sent by the old
--- a/core/src/test/java/org/infinispan/partitionhandling/PreferConsistencyStrategyTest.java
+++ b/core/src/test/java/org/infinispan/partitionhandling/PreferConsistencyStrategyTest.java
@@ -1,0 +1,52 @@
+package org.infinispan.partitionhandling;
+
+import java.util.Collections;
+import java.util.Optional;
+
+import org.infinispan.partitionhandling.impl.AvailabilityStrategyContext;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/infinispan/partitionhandling/impl/PreferConsistencyStrategy.java
```diff
@@ -179,10 +179,8 @@ public void onPartitionMerge(AvailabilityStrategyContext context,
          actualMembers.retainAll(mergedTopology.getMembers());
          mergedAvailabilityMode = AvailabilityMode.DEGRADED_MODE;
       } else {
-         log.debugf("No current topology, recovered only joiners for cache %s", context.getCacheName());
-         mergedTopology = null;
-         actualMembers = null;
-         mergedAvailabilityMode = AvailabilityMode.AVAILABLE;
+         log.debugf("No current topology, recovered only joiners for cache %s. Skipping availability update.", context.getCacheName());
+         return;
```
danberindei: I'm not sure about this change, I'm thinking maybe the coordinator already has a non-null topology in the `ClusterCacheStatus` and in that case, it should be replaced with a `null` topology.

Also, the test passes even without this change.

pruivo:
@slaskawi is this needed for CR2? I'm going to start the release in 30 min

slaskawi:
@pruivo @danberindei Unfortunately I need it for CR2 :(

I will integrate it as is and create a JIRA to get back to this: https://issues.jboss.org/browse/ISPN-7525

core/src/main/java/org/infinispan/partitionhandling/impl/PreferConsistencyStrategy.java
slaskawi: I've been playing with the code a little bit....

When I reverted this change (and allowed `mergedTopology` to be `null` and `actualMembers` to be an empty list) the `ClusterStatus` created [initial topology](https://github.com/infinispan/infinispan/blob/7f991051539e0d0006ef1839c32df3fcfd5c229b/core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java#L422). I assume this is not correct since we are updating [because of merge](https://github.com/infinispan/infinispan/blob/995683a3d6d27e665c186e895bdf71aa01cbb29a/core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java#L499-L499). 

So unless I'm mistaken, the proposed fix seems alright. And of course, thank you for noting that the test was wrong.

danberindei: The `doMergePartitions` method name is misleading, it could be just that the old coordinator crashed and the new coordinator just needs to recover one partition. 

I think the idea was to create an initial topology if the cache has no "full" members yet, but there is one node who sent a join request to the old coordinator and didn't get a response with the initial cache topology before sending its status response. The joiner should be retrying to join with the new coordinator though, so I'm not sure it's worth keeping track of these joiners during the recovery process.

The case I was thinking about before is when there is a merge, but a node also left the cluster in the same view. If that node was the last member of one cache, we should make sure that cache's topology is set to `null`. Unfortunately you're right, we were trying to create an initial topology here too, and that probably caused a `NPE` sooner or later...

PR #4880, opened 2017-02-22
Experiment: make remote gets only go to the primary owner

danberindei:


danberindei:
run performance tests please
library dist

PR #4884, opened 2017-02-22
ISPN-7507 Deprecate async options for tx cache

pruivo:
https://issues.jboss.org/browse/ISPN-7507

remove the options and simplify the condition for sync/async mode in transactional caches. Also, solves some IDE warning.


core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
```diff
@@ -319,6 +324,12 @@ public void validate() {
       if (!attributes.attribute(NOTIFICATIONS).get()) {
          log.transactionNotificationsDisabled();
       }
+      if (attributes.attribute(TRANSACTION_MODE).get() == TransactionMode.TRANSACTIONAL && !cacheMode.isSynchronous()) {
+         //change to sync or throw exception?
+         CacheMode syncCacheMod = cacheMode.toSync();
+         log.unsupportedAsyncCacheMode(cacheMode, syncCacheMod);
```
tristantarrant: I would be even more agressive: throw an exception. What do others think ?

ryanemerson: +1 for throwing an exception. As a user I would prefer to fail-fast and explicitly.

pruivo:
@tristantarrant @ryanemerson updated!

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -509,11 +474,14 @@ protected ReadOnlyKeyCommand remoteReadOnlyCommand(InvocationContext ctx, ReadOn
       return new TxReadOnlyKeyCommand(command, getMutationsOnKey((TxInvocationContext) ctx, command.getKey()));
    }
 
+   protected RpcOptions createRpcOptions() {
+      return rpcManager.getRpcOptionsBuilder(ResponseMode.SYNCHRONOUS_IGNORE_LEAVERS, DeliverOrder.NONE).build();
```
ryanemerson: Wouldn't it be better to store this as a protected variable in start() as the returned value is constant?

pruivo: no. you can change the replication timeout at runtime and you can't cache it. I tried and some tests failed.

rvansa: Eh, {{RpcOptions}} are cached on a handful of places - OK, I am the villain who does that, but noone complained during review so far. Does it make sense to change that timeout? The builder + the RpcOptions is created thousand times per second, there should be at least an event to refresh such cached objects.

pruivo: doing a search, only the ```BaseRpcInterceptor``` is caching and it shouldn't (```staggeredOptions``` and ```defaultSyncOptions```). 
Async rpc options doesn't matter since the timeout is ignored.

pruivo: anyway, I prefer to remove the possibility to change configuration at runtime.

tristantarrant: The configuration Attributes support AttributeListeners so that you could hook into that

rvansa: Cool, I'll try to create a fix.

rvansa: #4897

pruivo: do you want me to change this as well? (based on your changes)

pruivo:
rebased! @tristantarrant can you take a look?

tristantarrant:
Merged, thanks @pruivo

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ClientConnectionPoolingTest.java
```diff
@@ -178,29 +172,19 @@ public void testMaxActiveReached() throws Exception {
...
-               return 2 == connectionPool.getNumActive(hrServ1Addr) &&
-                     2 == connectionPool.getNumActive(hrServ2Addr) &&
-                     0 == connectionPool.getNumIdle(hrServ1Addr) &&
-                     0 == connectionPool.getNumIdle(hrServ2Addr);
-            }
-         });
+         eventually(() -> 2 == connectionPool.getNumActive(hrServ1Addr) &&
+               2 == connectionPool.getNumActive(hrServ2Addr) &&
+               0 == connectionPool.getNumIdle(hrServ1Addr) &&
+               0 == connectionPool.getNumIdle(hrServ2Addr));
```
danberindei: Note for the future: you can use `eventuallyEquals(2, () -> connectionPool.getNumActive(hrServ1Addr))` for a nicer exception message.

PR #4885, opened 2017-02-22
ISPN-7512 Add FRAG3 back to TCP configuration

pruivo:
https://issues.jboss.org/browse/ISPN-7512

### DIFF
--- a/server/integration/jgroups/src/main/resources/subsystem-templates/infinispan-jgroups.xml
+++ b/server/integration/jgroups/src/main/resources/subsystem-templates/infinispan-jgroups.xml
@@ -39,6 +39,9 @@
                 <protocol type="pbcast.STABLE" />
                 <protocol type="pbcast.GMS" />
                 <protocol type="MFC" />
+                <protocol type="FRAG3">
+                    <property name="frag_size">200k</property>
+                </protocol>
                 <?AUTH?>
             </stack>
             <stack name="tcp-gossip">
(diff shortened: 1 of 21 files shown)


alanfx:
👍 💯 🥇 from me!

wburns:
👍 Seems good to me, but I would wait to see what @danberindei says.  I don't understand how it would work with a bigger message (larger than credit size) without FRAG.

slaskawi:
@pruivo Just a silly question, shouldn't we run perf tests here?

danberindei:
@slaskawi When the difference is between working and not-working, I'm not sure perf tests are relevant... either way, our perf tests use 10k values, so they would never be fragmented (the default `frag_size` is 60k).

I'd make `frag_size` even bigger, say 200k, to make fragmentation even less likely. For reference, `max_credits * min_threshold` is 800k in our default configs.

pruivo:
updated the ```frag_size``` except for server configuration that uses ```jgroups-defaults.xml``` values.

tristantarrant:
Aha, server does not have a way to distinguish between defaults for TCP and for UDP.

server/integration/jgroups/src/main/resources/subsystem-templates/infinispan-jgroups.xml
```diff
@@ -39,6 +39,7 @@
                 <protocol type="pbcast.STABLE" />
                 <protocol type="pbcast.GMS" />
                 <protocol type="MFC" />
+                <protocol type="FRAG3" />
```
tristantarrant: Add the explicit frag size for TCP to override the UDP default.

server/integration/jgroups/src/main/resources/subsystem-templates/infinispan-jgroups.xml
```diff
@@ -57,6 +58,7 @@
...
```
tristantarrant: Same here

pruivo:
@tristantarrant added the explicit ```frag_size``` to that 2 places

tristantarrant:
Merged, thanks

PR #4886, opened 2017-02-22
ISPN-7513 AbstractClusteredWriteSkewTest can fail randomly still

wburns:
* Tries inserting different keys if first set fails

https://issues.jboss.org/browse/ISPN-7513

### DIFF
--- a/core/src/test/java/org/infinispan/container/versioning/AbstractClusteredWriteSkewTest.java
+++ b/core/src/test/java/org/infinispan/container/versioning/AbstractClusteredWriteSkewTest.java
@@ -301,19 +301,22 @@ public abstract class AbstractClusteredWriteSkewTest extends MultipleCacheManage
       }
 
       DataContainer dataContainer = TestingUtil.extractComponent(primaryOwner, DataContainer.class);
-      MagicKey[] keys = new MagicKey[MAX_ENTRIES];
+      int retryCount = 5;
       int insertCount = 10;
-      for (int i = 0; i < insertCount; ++i) {
-         for (int j = 0; j < MAX_ENTRIES; ++j) {
-            MagicKey tempKey = keys[j];
-            if (tempKey == null) {
-               tempKey = new MagicKey("other-key-" + j, primaryOwner);
-               keys[j] = tempKey;
+      for (int retry = 0; retry < 5; ++retry) {
+         MagicKey[] keys = new MagicKey[MAX_ENTRIES];
+         for (int i = 0; i < insertCount; ++i) {
+            for (int j = 0; j < MAX_ENTRIES; ++j) {
+               MagicKey tempKey = keys[j];
+               if (tempKey == null) {
+                  tempKey = new MagicKey("other-key-" + j, primaryOwner);
+                  keys[j] = tempKey;
+               }
+               primaryOwner.put(tempKey, "value");
             }
-            primaryOwner.put(tempKey, "value");
          }
       }
-      assertTrue("The key was not evicted after " + insertCount + " inserts", dataContainer.peek(key) == null);
+      assertTrue("The key was not evicted after " + insertCount + " inserts - retried " + retryCount, dataContainer.peek(key) == null);
 
       log.debugf("It is going to try to commit the suspended transaction");
       try {


core/src/test/java/org/infinispan/container/versioning/AbstractClusteredWriteSkewTest.java
```diff
@@ -301,19 +301,22 @@ private void doTestWriteSkewWithPassivation(boolean executeOnPrimaryOwner, Opera
...
-      MagicKey[] keys = new MagicKey[MAX_ENTRIES];
+      int retryCount = 5;
       int insertCount = 10;
-      for (int i = 0; i < insertCount; ++i) {
-         for (int j = 0; j < MAX_ENTRIES; ++j) {
-            MagicKey tempKey = keys[j];
-            if (tempKey == null) {
-               tempKey = new MagicKey("other-key-" + j, primaryOwner);
-               keys[j] = tempKey;
+      for (int retry = 0; retry < 5; ++retry) {
```
danberindei: Is there anything here that could stop the loop after the first iteration?

wburns: That was the goal, unfortunately there is no 'peak' equivalent for caffeine.  So this could cause it to fail possibly.  That was my initial goal actually (hence retryCount variable name).  It should be named something else now :P

rvansa: Is there a difference to `insertCount = 50`?

wburns: Actually yes there is.  Sadly `insertCount = 50` could still fail.  The biggest difference is that the array is reinitialized causing new keys to be generated https://github.com/infinispan/infinispan/pull/4886/files/1d8d89a0c952c1f7bf2ade50ce18921a08cd4c90#diff-2936358487ff1033e57010e8fff8ca62R307 . This forces caffeine in my test to always evict the desired key after it completes when I tested over 100K times.

danberindei: Ok, my question was related to the variable name and the commit comment... I really couldn't tell from the code what's failing so as to be retried :)

Could we not use something like `cache.evict(key)` to force the key to be passivated? IMO only tests that exercise the data container directly should care about how many inserts it takes for the data container to passivate an entry.

wburns: I originally had done dataContainer.remove (which should be akin to evict) and I was getting write skew errors for some reason (tbh I didn't look into it too closely).

danberindei: I wouldn't expect `dataContainer.remove(key)` to passivate the entry, but to remove it without writing it to disk.

wburns: Yes sorry forgot it was running a test with passivation :)

wburns: Will try evict and create a new PR if needed.

ben-manes: Sorry, I'm confused. By peek would getIfPresent suffice? If it's randomness, here's an ugly [hack](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/test/java/com/github/benmanes/caffeine/cache/RandomSeedEnforcer.java) that I use in my tests.

ben-manes: asMap().containksKey(key) is probably what you want for peeking, since that doesn't effect the eviction policy.

wburns: What we had before was a method we called `peek` that would return the value from the container without updating the underlying eviction structures.  This is nice to have since we may end up reading from the container more than once per command and we don't want that to increase the number of hits for a given key.

So if getIfPresent doesn't update internal eviction structure, we could definitely use that!

wburns: @ben-manes We also want to retrieve the value as well though

ben-manes: Oh, thanks. This tends to make sense for testing but not real world code. So I'm hesitant to introduce it as not carrying its weight.

wburns: Yeah tbh the more I think about it we would probably really want this (for non test code). The reason being is that when we do a put we first do a read before hand, so we can possibly do some verification checks of version and other things before actually doing the real put. This would lead to entries being promoted much more quickly in the eviction structure since there would be multiple hits.

ben-manes: But since the cache might evict any entry, that seems prone for race conditions.

wburns: Internally you may get something else on the return from put, but we don't use that, so it wouldn't affect us.  But I understand if you wouldn't want a feature like peek.

danberindei: @ben-manes yes, the version check is only usable with eviction if the cache has a cache store, which would keep the entries after they were evicted by the data container.

ben-manes: Does it make more sense to have the store inclusive, so a peek can go directly there?

ben-manes: There is also the map compute methods if you want to do read, verify, write.

wburns:
Created #4894 as well to do evict.

ben-manes:
Please upgrade to 2.4. There was a rare race condition that I overlooked, but the Solr guys helped track it down.

I also put a [slide deck](https://docs.google.com/presentation/d/1NlDxyXsUG1qlVHMl4vsUUBQfAJ2c2NsFPNPr2qymIBs/edit) together. Of interest might be the idea of using a timer wheel for per-entry expiration.

wburns:
@ben-manes Sounds good.  I will check it out, thanks!

PR #4890, opened 2017-02-23
ISPN-7433 Lower default xmit_interval to 100ms

danberindei:
This time with additional changes to work around JGRP-2159

https://issues.jboss.org/browse/ISPN-7433

### DIFF
--- a/core/src/main/java/org/infinispan/remoting/transport/Transport.java
+++ b/core/src/main/java/org/infinispan/remoting/transport/Transport.java
@@ -146,6 +146,15 @@ public interface Transport extends Lifecycle {
     */
    int getViewId();
 
+   /**
+    * @return A {@link CompletableFuture} that completes when the transport has installed the expected view.
+    */
+   CompletableFuture<Void> withView(int expectedViewId);
+
+   /**
+    * @deprecated Since 9.0, please use {@link #withView(int)} instead.
+    */
+   @Deprecated
    void waitForView(int viewId) throws InterruptedException;
 
    Log getLog();
--- a/core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
+++ b/core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
@@ -6,9 +6,11 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.infinispan.Version;
 import org.infinispan.commands.ReplicableCommand;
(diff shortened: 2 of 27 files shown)


danberindei:
Updated to replace the `use_delta_views="false"` workaround with `install_view_locally_first="true"`

core/src/main/java/org/infinispan/remoting/transport/Transport.java
```diff
@@ -146,6 +146,15 @@
     */
    int getViewId();
 
+   /**
+    * @return A {@link CompletableFuture} that completes when the transport has installed the expected view.
+    */
+   CompletableFuture<Void> viewFuture(int expectedViewId);
```
tristantarrant: Not sure I like the shunting of the "type" into the method name. Just view() would be enough.

rvansa: I don't think so. While `viewFuture` might not be the best one, `view()` seems like returning a list of members (or similar structure), while this is `CF<Void>`. If you want to avoid "future", it could be something like `whenView(...)` or `withView(...)` which can be used in fluent syntax: `whenView(...).thenApply(...)` (ok, that's not really Shakespeare's style...)

danberindei: @tristantarrant @rvansa +1 for `withView()`.

tristantarrant: that's much better.

alanfx: What about `viewView()`? :-)

danberindei: Renamed to `withView()`, sorry @alanfx ;)

alanfx: 🥇 It was worth a shot, @danberindei!

core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
```diff
@@ -520,8 +529,15 @@ public boolean isTotalOrderCache(String cacheName) {
...
    }
 
-   private void waitForView(int viewId) throws InterruptedException {
-      transport.waitForView(viewId);
+   private void waitForView(int viewId, long timeout, TimeUnit timeUnit) throws InterruptedException {
+      try {
+         transport.withView(viewId).get(timeout, timeUnit);
+      } catch (ExecutionException e) {
+         // The view future should never complete with an exception
+         log.errorf(e, "Unexpected exception waiting for view %d, current view is %d", viewId, transport.getViewId());
```
pruivo: shouldn't have a method in ```Log``` interface?

core/src/test/java/org/infinispan/remoting/transport/ControlledTransport.java
```diff
@@ -5,6 +5,7 @@
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
```
pruivo: unused import

pruivo:
@danberindei needs rebase.

danberindei:
@pruivo Updated

pruivo:
@danberindei one more comment:
```tcp_mping/tcp1.xml``` and ```tcp_mping/tcp2.xml``` are missing ```install_view_locally_first=true```

pruivo:
@danberindei then, you can squash the JGRP-2159 commits :)

danberindei:
Ok, XMLs updated and commits squashed.

I added one more change to `LocalTopologyManagerImpl`, as I noticed one more case where the joiner was blocking forever because of https://issues.jboss.org/browse/ISPN-7554

pruivo:

PR #4891, opened 2017-02-23
ISPN-7519 NPE and Deadlock during server start

gustavonalle:
https://issues.jboss.org/browse/ISPN-7519

Cannot reliably reproduce the issue, it may be fixed by the Netty upgrade [according to Norman Maurer](https://github.com/netty/netty/issues/6443#issuecomment-281997046)


pruivo:
looks good. pushing

pruivo:
integrated! thanks @gustavonalle

PR #4892, opened 2017-02-23
ISPN-7512 Add FRAG3 back to TCP configuration

pruivo:
* use default frag size

https://issues.jboss.org/browse/ISPN-7512


danberindei:
Integrated, thanks Pedro!

PR #4897, opened 2017-02-23
ISPN-7523 Update cached RpcOptions when timeout changes

rvansa:
https://issues.jboss.org/browse/ISPN-7523

Not sure if the `volatile` is really needed - we could probably live a moment before the update is propagated to the other threads. But volatile reads should be cheap.

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/cache/ClusteringConfiguration.java
+++ b/core/src/main/java/org/infinispan/configuration/cache/ClusteringConfiguration.java
@@ -19,7 +19,7 @@ public class ClusteringConfiguration {
          AttributeDefinition.builder("remoteTimeout", TimeUnit.SECONDS.toMillis(15)).build();
 
    static AttributeSet attributeDefinitionSet() {
-      return new AttributeSet(ClusteringConfiguration.class, CACHE_MODE);
+      return new AttributeSet(ClusteringConfiguration.class, CACHE_MODE, REMOTE_TIMEOUT);
    }
 
    private final Attribute<CacheMode> cacheMode;
--- a/core/src/main/java/org/infinispan/interceptors/impl/BaseRpcInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/BaseRpcInterceptor.java
@@ -8,6 +8,7 @@ import java.util.concurrent.CompletableFuture;
 
 import org.infinispan.commands.FlagAffectedCommand;
 import org.infinispan.commands.tx.PrepareCommand;
+import org.infinispan.configuration.cache.ClusteringConfiguration;
 import org.infinispan.context.Flag;
 import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.context.impl.LocalTxInvocationContext;
@@ -41,8 +42,8 @@ public abstract class BaseRpcInterceptor extends DDAsyncInterceptor {
    protected RpcManager rpcManager;
 
    protected boolean defaultSynchronous;
-   protected RpcOptions staggeredOptions;
-   protected RpcOptions defaultSyncOptions;
+   protected volatile RpcOptions staggeredOptions;
+   protected volatile RpcOptions defaultSyncOptions;
    protected RpcOptions defaultAsyncOptions;
 
    protected abstract Log getLog();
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/infinispan/interceptors/impl/BaseRpcInterceptor.java
```diff
@@ -44,9 +45,9 @@
 
    protected boolean defaultSynchronous;
    private boolean syncCommitPhase;
-   protected RpcOptions staggeredOptions;
-   protected RpcOptions defaultSyncOptions;
-   protected RpcOptions defaultAsyncOptions;
+   protected volatile RpcOptions staggeredOptions;
+   protected volatile RpcOptions defaultSyncOptions;
+   protected volatile RpcOptions defaultAsyncOptions;
```
pruivo: async options doesn't care about timeouts. can be left as it is

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -651,13 +652,19 @@ public void start() {
       CacheMode mode = configuration.clustering().cacheMode();
       isFetchEnabled = (mode.isDistributed() || mode.isReplicated()) &&
               (configuration.clustering().stateTransfer().fetchInMemoryState() || configuration.persistence().fetchPersistentState());
-      //rpc options does not changes in runtime. we can use always the same instance.
-      rpcOptions = rpcManager.getRpcOptionsBuilder(ResponseMode.SYNCHRONOUS)
-            .timeout(timeout, TimeUnit.MILLISECONDS).build();
```
pruivo: this is not the remote timeout. it is the state transfer timeout.

rvansa:
Thanks, that's for doing a fix in a hurry...

pruivo:
@rvansa needs rebase

core/src/main/java/org/infinispan/configuration/cache/ClusteringConfiguration.java
```diff
@@ -19,7 +19,7 @@
          AttributeDefinition.builder("remoteTimeout", TimeUnit.SECONDS.toMillis(15)).build();
 
    static AttributeSet attributeDefinitionSet() {
-      return new AttributeSet(ClusteringConfiguration.class, CACHE_MODE);
+      return new AttributeSet(ClusteringConfiguration.class, CACHE_MODE, REMOTE_TIMEOUT);
```
pruivo: this is confusing. REMOTE_TIMEOUT exists in the SyncConfiguration. I'm not sure which one is really used or if it is really a good idea to add it here...

rvansa: Uh, do I need to register the listener on a deprecated class? @tristantarrant please counsel...

rvansa:
Rebased.

core/src/main/java/org/infinispan/configuration/cache/ClusteringConfiguration.java
tristantarrant: The REMOTE_TIMEOUT attribute in SyncConfiguration is actually a ref to the ClusteringConfiguration one, so @rvansa is right

pruivo: @tristantarrant and the attribute REMOTE_TIMEOUT needs to be added to the Clustering attribute set?

tristantarrant: Since it is ClusteringConfiguration which owns the attribute, it makes sense for it to exist in that attribute set rather than in the deprecated one. AttributeSets matter when copying a configuration and when computing equality, otherwise it doesn't matter much.

pruivo:
looks good. testing it locally.

pruivo:
well, it is fine locally... but CI should finish soon

pruivo:
@rvansa can you look at ```
AtomicMapAPITest.testConcurrentTx```? it is the first time it failed but I don't think it is related.

rvansa:
@pruivo Haven't seen it failing either, but given that it fails just after 500 ms, I guess any hiccup (GC is my favorite accused) could cause this.

pruivo:
integrated! thanks @rvansa !

PR #4899, opened 2017-02-23
ISPN-7522 LocalMode[No]PassivationTest fails randomly

wburns:
* Fixed bug where cache loader can miss entry with bulk op

https://issues.jboss.org/browse/ISPN-7522

The actual fix is trivial, but understanding what it is doing requires a little more knowledge.  Basically persistence thread is constantly feeding entries and then once it is complete it calls the close method.  However the user thread is constantly pulling from the queue being fed to it.

There was a small window where the queue could be empty for the user thread (so it checks to see if the supplier was closed).  If the persistence thread puts stuff into a queue and closes the supplier during this time the user thread won't see those last entries.

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/util/PersistenceManagerCloseableSupplier.java
+++ b/core/src/main/java/org/infinispan/persistence/util/PersistenceManagerCloseableSupplier.java
@@ -42,7 +42,7 @@ public class PersistenceManagerCloseableSupplier<K, V> implements CloseableSuppl
    private final Condition closeCondition = closeLock.newCondition();
 
    private boolean closed = false;
-   private AtomicReference<AdvancedCacheLoader.CacheLoaderTask<K, V>> taskRef = new AtomicReference<>();
+   private final AtomicReference<AdvancedCacheLoader.CacheLoaderTask<K, V>> taskRef = new AtomicReference<>();
 
    public PersistenceManagerCloseableSupplier(Executor executor, PersistenceManager manager,
                                               InternalEntryFactory factory, KeyFilter<K> filter, long timeout,
@@ -124,6 +124,11 @@ public class PersistenceManagerCloseableSupplier<K, V> implements CloseableSuppl
          closeLock.lock();
          try {
             if (closed) {
+               // If is possible that this supplier was closed right after we polled the queue. In that case they may
+               // have also inserted a value. We need to double check that the queue is REALLY empty after being closed.
+               if ((entry = queue.poll()) != null) {
+                  break;
+               }
                break;
             }
             long targetTime = System.nanoTime() + unit.toNanos(timeout);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/infinispan/persistence/util/PersistenceManagerCloseableSupplier.java
```diff
@@ -124,6 +124,11 @@ public void processEntry(MarshalledEntry<K, V> marshalledEntry, AdvancedCacheLoa
          closeLock.lock();
          try {
             if (closed) {
+               // If is possible that this supplier was closed right after we polled the queue. In that case they may
+               // have also inserted a value. We need to double check that the queue is REALLY empty after being closed.
+               if (queue.peek() != null) {
```
pruivo: could you just do a poll and, if the entry is not null, return it (instead of continue?)

wburns: Hrmm yeah that should be fine.

core/src/test/java/org/infinispan/persistence/LocalModePassivationTest.java
```diff
@@ -59,31 +55,25 @@ protected void configureConfiguration(ConfigurationBuilder cb) {
...
    @Override
    protected EmbeddedCacheManager createCacheManager() throws Exception {
-      cacheStoreDir = new File(TestingUtil.tmpDirectory(this.getClass()));
-      Util.recursiveFileRemove(cacheStoreDir);
-
       ConfigurationBuilder builder = getDefaultClusteredCacheConfig(CacheMode.LOCAL, true, true);
       builder.transaction().transactionMode(TransactionMode.TRANSACTIONAL).lockingMode(LockingMode.PESSIMISTIC)
             .transactionManagerLookup(new EmbeddedTransactionManagerLookup())
-            .eviction().maxEntries(150).strategy(EvictionStrategy.LIRS)
+            .memory().storageType(StorageType.BINARY).size(150)
```
danberindei: Is the BINARY storage type significant?

core/src/main/java/org/infinispan/persistence/util/PersistenceManagerCloseableSupplier.java
wburns: On second thought this could possibly swallow an interrupt, since we don't reset the interrupt state.  I can fix but it will take longer.

wburns: Nvm I can do a break instead of return and it should work.

core/src/test/java/org/infinispan/persistence/LocalModePassivationTest.java
wburns: That is just what the test was doing before :)  All I did was update it to use the new element.

wburns:
Updated

pruivo:
pushing...

pruivo:
integrated! thanks @wburns

PR #4902, opened 2017-02-24
ISPN-7530 InfinispanStoreRocksDBIT leaves files in the current directory

danberindei:
https://issues.jboss.org/browse/ISPN-7530

### DIFF
--- a/integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/InfinispanStoreRocksDBIT.java
+++ b/integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/InfinispanStoreRocksDBIT.java
@@ -3,8 +3,11 @@ package org.infinispan.test.integration.as;
 import static java.io.File.separator;
 import static org.junit.Assert.assertEquals;
 
+import java.io.File;
+
 import org.infinispan.Cache;
 import org.infinispan.Version;
+import org.infinispan.commons.util.Util;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.configuration.global.GlobalConfigurationBuilder;
 import org.infinispan.manager.DefaultCacheManager;
@@ -19,6 +22,8 @@ import org.jboss.shrinkwrap.api.asset.StringAsset;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.jboss.shrinkwrap.descriptor.api.Descriptors;
 import org.jboss.shrinkwrap.descriptor.api.spec.se.manifest.ManifestDescriptor;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
@@ -30,6 +35,8 @@ import org.junit.runner.RunWith;
  */
 @RunWith(Arquillian.class)
 public class InfinispanStoreRocksDBIT {
+   private static String dataDir = tmpDirectory(InfinispanStoreRocksDBIT.class) + File.separator + "rocksdb-testcache";
+   private static String expiredDir = tmpDirectory(InfinispanStoreRocksDBIT.class) + File.separator + "rocksdb-expiredtestcache";
 
    @Deployment
    public static Archive<?> deployment() {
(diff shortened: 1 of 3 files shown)


integrationtests/wildfly-modules/src/test/resources/arquillian.xml
```diff
@@ -11,10 +11,13 @@
             <configuration>
                 <property name="jbossHome">${basedir}/target/node1/wildfly-${version.org.wildfly}</property>
                 <property name="javaVmArguments">-Djboss.socket.binding.port-offset=100 -Xmx512m -Djava.net.preferIPv4Stack=true
-                    -Djgroups.bind_addr=127.0.0.1 -Djgroups.join_timeout=2000
+                     -Djgroups.bind_addr=127.0.0.1 -Djgroups.join_timeout=2000
                 </property>
                 <!-- To debug the Arquillian managed application server: -->
-                <!-- property name="javaVmArguments">-Xrunjdwp:transport=dt_socket,address=8000,server=y,suspend=y -Xmx512m -Dorg.jboss.remoting-jmx.timeout=300  -Djava.net.preferIPv4Stack=true -Djgroups.bind_addr=127.0.0.1</property-->
+                <!--<property name="javaVmArguments">-Xrunjdwp:transport=dt_socket,address=8000,server=n,suspend=y -Dorg.jboss.remoting-jmx.timeout=300-->
+                    <!-- -Djboss.socket.binding.port-offset=100 -Xmx512m -Djava.net.preferIPv4Stack=true-->
```
slaskawi: Changes in this file could be probably removed.

integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/InfinispanStoreRocksDBIT.java
```diff
@@ -55,9 +69,12 @@ public void testCacheManager() {
...
       gcb.globalJmxStatistics().allowDuplicateDomains(true);
 
+      String tmpDirectory = tmpDirectory(this.getClass());
       ConfigurationBuilder builder = new ConfigurationBuilder();
-      builder.persistence().addStore(RocksDBStoreConfigurationBuilder.class)
-            .location(tmpDirectory(this.getClass()));
+      builder.persistence()
+             .addStore(RocksDBStoreConfigurationBuilder.class)
+             .location(tmpDirectory)
+             .expiredLocation(tmpDirectory + "-expired");
```
slaskawi: Good one! :)

integrationtests/wildfly-modules/src/test/resources/arquillian.xml
danberindei: I really want to keep this in because the existing commented `javaVmArguments` were incomplete, and it was frustrating when I enabled debugging and the test wouldn't run.

danberindei:
Updated to fix `RocksDBCacheStoreIT`, I missed the fact that `dataDir` had to be in sync with the XML configuration in `rocksdb-(dist|repl|local).xml`.

integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/InfinispanStoreRocksDBIT.java
pruivo: I'm confused... shouldn't we use ```expireDir``` instead?

integrationtests/wildfly-modules/src/test/resources/arquillian.xml
slaskawi: In that case, could you please add a comment with those switches at the beginning of the file? Let's now throw rubbish all around :D

integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/InfinispanStoreRocksDBIT.java
slaskawi: I think RocksDB needs both those directories (but I might be wrong). Maybe @tristantarrant would know.

pruivo: ok  but here you are using
```
tmpDirectory(this.getClass()) + "-expired"
```
and in the beginning you are creating
```
private static String expiredDir = tmpDirectory(InfinispanStoreRocksDBIT.class) + File.separator + "rocksdb-expiredtestcache";
```

integrationtests/wildfly-modules/src/test/resources/arquillian.xml
danberindei: @slaskawi I could remove the comment altogether, but I would definitely not leave the comment as it was before...

integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/InfinispanStoreRocksDBIT.java
danberindei: Fixed @pruivo

pruivo:
@danberindei and @slaskawi :+1: looks good to me.

PR #4903, opened 2017-02-25
[perf] removing triangle ack interceptor

pruivo:
another approach:
* triangle ack interceptor is removed.
* acks are sent in the per-cache inbound handler (similar to replies). this is really needed because if an exception occurs before the triangle ack interceptor, it wouldn't be sent back to originator
* primary ack is removed. the command is forward to the primary with a sync rpc.
* collector waits on triangle distribution interceptor.


pruivo:
ps. the code is not completed

pruivo:
run performance tests please
library

pruivo:
run performance tests please
cs

pruivo:
run performance tests please
cs

pruivo:
run performance tests please
cs dist writes

pruivo:
run performance tests please
library

pruivo:
run performance tests please
cs

PR #4905, opened 2017-02-27
ISPN-7526 Write skew check throws even if the previous value was not read

rvansa:
https://issues.jboss.org/browse/ISPN-7526

### DIFF
--- a/core/src/main/java/org/infinispan/container/entries/ClusteredRepeatableReadEntry.java
+++ b/core/src/main/java/org/infinispan/container/entries/ClusteredRepeatableReadEntry.java
@@ -2,8 +2,6 @@ package org.infinispan.container.entries;
 
 import static org.infinispan.commons.util.Util.toStr;
 
-import java.util.concurrent.atomic.AtomicReference;
-
 import org.infinispan.container.DataContainer;
 import org.infinispan.container.entries.versioned.Versioned;
 import org.infinispan.container.versioning.EntryVersion;
@@ -27,7 +25,6 @@ public class ClusteredRepeatableReadEntry extends RepeatableReadEntry implements
 
    private static final Log log = LogFactory.getLog(ClusteredRepeatableReadEntry.class);
    private static final boolean trace = log.isTraceEnabled();
-   private static final AtomicReference<Boolean> ignored = new AtomicReference<>();
 
    public ClusteredRepeatableReadEntry(Object key, Object value, Metadata metadata) {
       super(key, value, metadata);
@@ -82,7 +79,7 @@ public class ClusteredRepeatableReadEntry extends RepeatableReadEntry implements
       EntryVersion prevVersion;// on origin, the version seen is acquired without the lock, so we have to retrieve it again
       // TODO: persistence should be more orthogonal to any entry type - this should be handled in interceptor
       InternalCacheEntry ice = PersistenceUtil.loadAndStoreInDataContainer(container, persistenceManager, getKey(),
-            ctx, timeService, ignored);
+            ctx, timeService, null);
       if (ice == null) {
          if (trace) {
             log.tracef("No entry for key %s found in data container", toStr(key));
--- a/core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
@@ -30,7 +30,6 @@ import org.infinispan.commands.write.PutKeyValueCommand;
 import org.infinispan.commands.write.PutMapCommand;
 import org.infinispan.commands.write.RemoveCommand;
 import org.infinispan.commands.write.ReplaceCommand;
-import org.infinispan.container.DataContainer;
 import org.infinispan.context.InvocationContext;
 import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.factories.annotations.Inject;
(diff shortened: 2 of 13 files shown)


core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
```diff
@@ -50,7 +50,6 @@
    private final boolean trace = getLog().isTraceEnabled();
 
    protected LockManager lockManager;
-   protected DataContainer<Object, Object> dataContainer;
```
pruivo: remove unused import

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -64,6 +64,7 @@
 import org.infinispan.remoting.transport.Address;
 import org.infinispan.statetransfer.OutdatedTopologyException;
 import org.infinispan.topology.CacheTopology;
+import org.infinispan.transaction.impl.AbstractCacheTransaction;
```
pruivo: another unused import

core/src/main/java/org/infinispan/container/entries/ClusteredRepeatableReadEntry.java
```diff
@@ -82,7 +79,7 @@ private EntryVersion getCurrentEntryVersion(DataContainer container, Persistence
       EntryVersion prevVersion;// on origin, the version seen is acquired without the lock, so we have to retrieve it again
       // TODO: persistence should be more orthogonal to any entry type - this should be handled in interceptor
       InternalCacheEntry ice = PersistenceUtil.loadAndStoreInDataContainer(container, persistenceManager, getKey(),
-            ctx, timeService, ignored);
+            ctx, timeService, null);
```
pruivo: I'm not sure about this. wouldn't be cheaper to have the ```AtomicReference``` (```volatile``` write) instead of changing the ```loadAndStoreInDataContainer``` to check for ```null```?

core/src/main/java/org/infinispan/interceptors/totalorder/TotalOrderInterceptor.java
```diff
@@ -4,18 +4,30 @@
 
 import java.util.Collection;
 
+import org.infinispan.commands.DataCommand;
+import org.infinispan.commands.VisitableCommand;
 import org.infinispan.commands.control.LockControlCommand;
+import org.infinispan.commands.read.GetAllCommand;
+import org.infinispan.commands.read.GetCacheEntryCommand;
+import org.infinispan.commands.read.GetKeyValueCommand;
+import org.infinispan.commands.remote.GetKeysInGroupCommand;
```
pruivo: unused import

core/src/main/java/org/infinispan/interceptors/totalorder/TotalOrderInterceptor.java
```diff
@@ -222,4 +241,80 @@ private void simulateLocking(TxInvocationContext context, PrepareCommand command
          }
       }
    }
+
+   // Total order does not use OptimisticLockingInterceptor, therefore we can't record the key being read in there
```
pruivo: :-1: 
I think this code and the code from ```OptimisticLockingInterceptor``` should be moved to the ```VersionedEntryWrappingInterceptor```.

core/src/main/java/org/infinispan/container/entries/ClusteredRepeatableReadEntry.java
rvansa: I haven't benchmarked it, but volatile write to single location would limit scalability. Volatile write requires to acquire exclusive ownership of the cache line, and I smell contention here. Theoretically, the check could be slower in an uncontended case (though I doubt so), but once we start hitting the reference with many threads...

core/src/main/java/org/infinispan/interceptors/totalorder/TotalOrderInterceptor.java
rvansa: I agree, this was merely a temporary solution. I am reverting the changes in TO interceptor in #4908. If you review that one, I could stash those changes.

pruivo:
@rvansa I took a look at the other PR. I guess I can close this one?

rvansa:
@pruivo Okay, closing.

core/src/main/java/org/infinispan/container/entries/ClusteredRepeatableReadEntry.java
danberindei: +1 to remove the `AtomicReference` here.

I'd even resurrect the `isLoaded()` flag in `InternalCacheEntry` and use it instead, removing the `AtomicReference` parameter completely. The `isLoaded.get() == Boolean.FALSE` case seems equivalent to `entry == null`, so we don't really need 3 possibilities.

pruivo: you have 3 cases:
* entry was found (hit)
* entry wasn't found (miss)
* it didn't try to fetch from persistence (ignore cache loader)

danberindei: @pruivo AFAICT it only ignores the store if the entry exists in the data container, flags are only checked in the interceptor. So these cases could map to:

1. entry != null && entry.isLoaded()
2. entry == null
3. entry != null && !entry.isLoaded()

PR #4906, opened 2017-02-27
ISPN-7505 Non-functional read in transaction with modifications returns old value 

rvansa:
https://issues.jboss.org/browse/ISPN-7505

Depens on #4905

### DIFF
--- a/core/src/main/java/org/infinispan/commands/functional/TxReadOnlyManyCommand.java
+++ b/core/src/main/java/org/infinispan/commands/functional/TxReadOnlyManyCommand.java
@@ -12,8 +12,7 @@ import java.util.Iterator;
 import java.util.List;
 
 import org.infinispan.commons.api.functional.EntryView;
-import org.infinispan.commons.marshall.MarshallUtil;
-import org.infinispan.container.entries.CacheEntry;
+import org.infinispan.container.entries.MVCCEntry;
 import org.infinispan.context.InvocationContext;
 import org.infinispan.functional.impl.EntryViews;
 
@@ -98,7 +97,7 @@ public class TxReadOnlyManyCommand<K, V, R> extends ReadOnlyManyCommand<K, V, R>
       Iterator<List<Mutation<K, V, ?>>> mutIt = mutations.iterator();
       for (K k : keys) {
          List<Mutation<K, V, ?>> mutations = mutIt.next();
-         CacheEntry<K, V> entry = lookupCacheEntry(ctx, k);
+         MVCCEntry<K, V> entry = (MVCCEntry<K, V>) lookupCacheEntry(ctx, k);
          EntryView.ReadEntryView<K, V> ro;
          Object ret = null;
          if (mutations.isEmpty()) {
@@ -107,6 +106,7 @@ public class TxReadOnlyManyCommand<K, V, R> extends ReadOnlyManyCommand<K, V, R>
             EntryView.ReadWriteEntryView rw = EntryViews.readWrite(entry);
             for (Mutation<K, V, ?> mutation : mutations) {
                ret = mutation.apply(rw);
+               entry.updatePreviousValue();
             }
             ro = rw;
          }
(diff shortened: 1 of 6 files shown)


pruivo:
@rvansa needs rebase

rvansa:
@pruivo rebased.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -387,8 +387,13 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
       }
 
+      CompletableFuture<Void> allFuture = remoteGetAll(ctx, command, requestedKeys);
+      return asyncInvokeNext(ctx, command, allFuture);
+   }
+
+   protected CompletableFuture<Void> remoteGetAll(InvocationContext ctx, GetAllCommand command, Map<Address, List<Object>> requestedKeys) {
       GlobalTransaction gtx = ctx.isInTxScope() ? ((TxInvocationContext) ctx).getGlobalTransaction() : null;
-      CompletableFutureWithCounter allFuture = new CompletableFutureWithCounter(requestedKeys.size());
+      BaseDistributionInterceptor.CompletableFutureWithCounter allFuture = new BaseDistributionInterceptor.CompletableFutureWithCounter(requestedKeys.size());
```
pruivo: I'm pretty sure this isn't needed.

core/src/main/java/org/infinispan/commands/functional/TxReadOnlyManyCommand.java
```diff
@@ -107,6 +106,7 @@ public Object perform(InvocationContext ctx) throws Throwable {
             EntryView.ReadWriteEntryView rw = EntryViews.readWrite(entry);
             for (Mutation<K, V, ?> mutation : mutations) {
                ret = mutation.apply(rw);
+               entry.updatePreviousValue();
```
pruivo: I'm not sure why do you need to invoke this method in each iteration. can you elaborate? (same for the other places in TxDistributionInterceptor)

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -45,13 +47,17 @@
 import org.infinispan.commands.write.ReplaceCommand;
 import org.infinispan.commands.write.ValueMatcher;
 import org.infinispan.commands.write.WriteCommand;
+import org.infinispan.commons.api.functional.EntryView;
 import org.infinispan.container.entries.CacheEntry;
+import org.infinispan.container.entries.MVCCEntry;
+import org.infinispan.container.entries.ReadCommittedEntry;
```
pruivo: unused import

core/src/main/java/org/infinispan/commands/functional/TxReadOnlyManyCommand.java
rvansa: During command execution with RR entries, the 'actual value' is taken from the `oldValue` - see `RepeatableReadEntry.setValue()`, used e.g. in `ReplaceCommand.perform()` or `PKVC.performPut:271`. Some places don't use that, e.g. the functional commands, but these don't deal with retries properly anyway. Actually `RemoveCommand` does not use it either - this could be a bug.
So, we have to finalize the application of mutation by calling `updatePreviousValue()` in order to  get 'correct' previous value returned by `setValue()`in next mutation. This is usually done by EWI after each command, but here we have to call it manually, since the mutations don't go through the stack.

rvansa:
Sorry for the compilation errors. And please let CI confirm the changes before integrating.

core/src/main/java/org/infinispan/commands/functional/TxReadOnlyManyCommand.java
pruivo: got it :) thanks

pruivo:
looks good. waiting for CI.

rvansa:
@pruivo CI results seem fine.

PR #4907, opened 2017-02-27
ISPN-7532 Do not log configuration warnings for templates

tristantarrant:
- Do not log configuration warnings for templates
- Also extract some configuration exceptions to log
- Do not disable expiration reaper in server

https://issues.jboss.org/browse/ISPN-7532

### DIFF
--- a/core/src/main/java/org/infinispan/util/logging/Log.java
+++ b/core/src/main/java/org/infinispan/util/logging/Log.java
@@ -1518,5 +1518,32 @@ public interface Log extends BasicLogger {
    void errorSendingResponse(ReplicableCommand command);
 
    @Message(value = "Unsupported async cache mode '%s' for transactional caches", id = 441)
-   CacheConfigurationException unsupportedAsyncCacheMode(CacheMode cacheModed);
+   CacheConfigurationException unsupportedAsyncCacheMode(CacheMode cacheMode);
+
+   @Message(value = "Invalid cache loader configuration for '%s'.  If a cache loader is configured as a singleton, the cache loader cannot be shared in a cluster!", id = 442)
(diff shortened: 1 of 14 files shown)


core/src/main/java/org/infinispan/util/logging/Log.java
```diff
@@ -1518,5 +1518,32 @@ TimeoutException timeoutWaitingForView(int expectedViewId, int currentViewId,
    void errorSendingResponse(ReplicableCommand command);
 
    @Message(value = "Unsupported async cache mode '%s' for transactional caches", id = 441)
-   CacheConfigurationException unsupportedAsyncCacheMode(CacheMode cacheModed);
+   CacheConfigurationException unsupportedAsyncCacheMode(CacheMode cacheMode);
+
+   @Message("Invalid cache loader configuration for '%s'.  If a cache loader is configured as a singleton, the cache loader cannot be shared in a cluster!", id = 442)
```
danberindei: Missing `value=`.

tristantarrant: It seems I can omit it only if it's the only attribute

tristantarrant:
Fixed it

danberindei:
Seeing test failures now:

```
testNegativeTimeout(org.infinispan.xsite.XSiteStateTransferFileParsingTest)  Time elapsed: 0.022 sec  <<< FAILURE!
org.testng.TestException: 
The exception was thrown with the wrong message: expected "Timeout must be higher or equals than 1 \(one\)." but got "ISPN000449: XSite state transfer timeout must be higher or equals than 1 (one)."
```

tristantarrant:
Fixed the test.

PR #4909, opened 2017-02-27
ISPN-6724 Assign new scripting logging id range

tristantarrant:
https://issues.jboss.org/browse/ISPN-6724

Move the scripting logging to a subrange of tasks to avoid conflicts with infinispan directory provider


danberindei:
Integrated, thanks Tristan!

PR #4910, opened 2017-02-27
ISPN-7488 Unwrap CacheConfigurationException

slaskawi:
https://issues.jboss.org/browse/ISPN-7488

This doesn't have much impact on how Infinispan works but it silences out many weird error messages.

### DIFF
--- a/core/src/main/java/org/infinispan/factories/AbstractComponentRegistry.java
+++ b/core/src/main/java/org/infinispan/factories/AbstractComponentRegistry.java
@@ -148,6 +148,8 @@ public abstract class AbstractComponentRegistry implements Lifecycle, Cloneable 
                invokeInjectionMethod(target, injectMetadata);
             }
          }
+      } catch (IllegalLifecycleStateException e) {
+        throw e;
       } catch (Exception e) {
          throw new CacheConfigurationException("Unable to configure component (type: " + target.getClass() + ", instance " + target + ")", e);
       }


core/src/main/java/org/infinispan/remoting/inboundhandler/GlobalInboundInvocationHandler.java
```diff
@@ -153,7 +154,7 @@ private void runReplicableCommand(ReplicableCommand command, Reply reply, boolea
       try {
          invokeReplicableCommand(command, reply, preserveOrder);
       } catch (Throwable throwable) {
-         if (throwable.getCause() != null && throwable instanceof CompletionException) {
+         if (throwable.getCause() != null && (throwable instanceof CompletionException || throwable instanceof CacheConfigurationException)) {
```
danberindei: I would prefer changing `AbstractComponentRegistry.wireDependencies()` to stop wrapping `IllegalLifecycleStateException`.

wburns: I agree that seems much cleaner to just catch and rethrow IllegalLifecycleStateException.  However this doesn't prevent if we got an InterruptedException wrapped by a CacheConfigurationException.  

So I would say either to just have this change or to have both.

slaskawi: I have checked all places with `wireDependencies` and it turns out only this one catches `IllegalLifecycleStateException` :D So I'm just wondering... maybe we could simplify this and only change the `wireDependencies` method (leaving this place untouched)?

PR #4911, opened 2017-02-27
ISPN-7489 Decreased close connection timeouts

slaskawi:
https://issues.jboss.org/browse/ISPN-7489
https://issues.jboss.org/browse/ISPN-7573

### DIFF
--- a/server/integration/jgroups/src/main/resources/subsystem-templates/cloud-jgroups.xml
+++ b/server/integration/jgroups/src/main/resources/subsystem-templates/cloud-jgroups.xml
@@ -104,7 +104,9 @@
                 <protocol type="FRAG3" />
             </stack>
             <stack name="kubernetes">
-                <?TCP-TRANSPORT?>
+                <transport type="TCP" socket-binding="jgroups-tcp" >
+                    <property name="logical_addr_cache_expiration">360000</property>
+                </transport>
                 <protocol type="kubernetes.KUBE_PING" />
                 <protocol type="MERGE3" />
                 <protocol type="FD_SOCK" socket-binding="jgroups-tcp-fd" />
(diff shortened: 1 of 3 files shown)


slaskawi:
@belaban I lowered those values but I tried to be gentle... 60s should be acceptable for most of the production systems. WDYT?

belaban:
Yes, 60s is good, but in the worst case, you'll get 60s worth of error messages.
Also, `max_retransmit_time` (default: 60s) needs to be added before messages to a left member are stopped.

danberindei:
@slaskawi @belaban Is this kubernetes-related, or should we have the same `conn_close_timeout` in all the default configs?

belaban:
@danberindei @slaskawi My comments were mainly related to reducing the number of ERROR messages shown. Sebastian: can we debug this interactively? I'd like to see how Kubernetes actually terminates the processes...

slaskawi:
Together with @belaban, we found the root issue and fixed it by upgrading JGroups to `4.0.1.Final` and adding `logical_addr_cache_expiration="360000"`.

server/integration/jgroups/src/main/resources/subsystem-templates/cloud-jgroups.xml
```diff
@@ -104,7 +104,9 @@
                 <protocol type="FRAG3" />
             </stack>
             <stack name="kubernetes">
-                <?TCP-TRANSPORT?>
+                <transport type="TCP" socket-binding="jgroups-tcp" >
+                    <property name="logical_addr_cache_expiration">360000</property>
```
slaskawi: I guess @tristantarrant might be worried about this one, but @belaban also changed the timeout in JGroups so it will be removed as soon as we upgrade to `4.0.2`.

gustavonalle:
Integrated, thanks @slaskawi !

PR #4912, opened 2017-02-27
[perf] global command ack collector

pruivo:
includes #4903


pruivo:
run performance tests please
library dist writes

PR #4914, opened 2017-02-27
ISPN-7070

wburns:
CacheNotifierImplInitialTransferDistTest.testIterationBeganAndSegmentNotComplete
random failures

* Make sure reads into listener collection are also synchronized
* Remove listener before replacing notifier that it registered with

https://issues.jboss.org/browse/ISPN-7070

### DIFF
--- a/core/src/test/java/org/infinispan/notifications/cachelistener/CacheNotifierImplInitialTransferDistTest.java
+++ b/core/src/test/java/org/infinispan/notifications/cachelistener/CacheNotifierImplInitialTransferDistTest.java
@@ -13,6 +13,7 @@ import static org.testng.Assert.fail;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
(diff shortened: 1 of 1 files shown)


wburns:
I can't confirm 100% that this fixes it, but it seems the issue could have been caused by an unsynchronized read causing it not to see the updated collection size variable.

pruivo:
I've run it locally for 1h and it didn't fail .I'm going to integrate it and see if it fails again in the future.

danberindei:
Got another failure today: 
https://www.dropbox.com/s/nsmz5k21dnk4h35/CacheNotifierImplInitialTransferDistTest_ISPN-7400_DistributionInfo_caching_20170302.log.gz?dl=0

wburns:
The log seems similar to the previous one.  Everything seems fine with the notifications.  I will add a log message to the listener so we can see exactly which event is not being counted

PR #4908, opened 2017-02-27
ISPN-7528 Store CacheTransaction.keyRead(...) in RepeatableReadEntry

rvansa:
https://issues.jboss.org/browse/ISPN-7528

Depends on #4905

PR #4916, opened 2017-02-28
ISPN-7527 Local write skew check cannot work with persistence

rvansa:
Depends on #4905 and #4908


core/src/main/java/org/infinispan/container/entries/VersionedRepeatableReadEntry.java
```diff
@@ -21,12 +21,12 @@
  * @author Manik Surtani
  * @since 5.1
  */
-public class ClusteredRepeatableReadEntry extends RepeatableReadEntry implements Versioned {
+public class VersionedRepeatableReadEntry extends RepeatableReadEntry implements Versioned {
```
rvansa: I wonder if we need this class at all, since it does not hold any data, or if we should just push the methods to `RepeatableReadEntry`.

pruivo:
@rvansa needs rebase

rvansa:
@pruivo rebased.

core/src/main/java/org/infinispan/container/entries/VersionedRepeatableReadEntry.java
pruivo: :+1:. It think this class was from the time where the version wasn't in the Metadata.
Are you going to remove it in this PR or another one?

core/src/main/java/org/infinispan/transaction/impl/WriteSkewHelper.java
```diff
@@ -46,7 +47,12 @@ public static EntryVersionsMap performWriteSkewCheckAndReturnNewVersions(Version
       for (WriteCommand c : prepareCommand.getModifications()) {
          for (Object k : c.getAffectedKeys()) {
             if (ksl.performCheckOnKey(k)) {
-               ClusteredRepeatableReadEntry entry = (ClusteredRepeatableReadEntry) context.lookupEntry(k);
+               CacheEntry cacheEntry = context.lookupEntry(k);
+               if (!(cacheEntry instanceof VersionedRepeatableReadEntry)) {
+                  // DeltaAware?
```
pruivo: DeltaAware and optimistic+write-skew isn't implemented. you don't have to worry about it

core/src/main/java/org/infinispan/container/entries/VersionedRepeatableReadEntry.java
rvansa: Not in this PR; I could remove it along with deprecating optimistic txs w/o wsc.

rvansa:
Removed those unused imports. I hope I haven't missed any...

pruivo:
looks good. waiting for CI

PR #4920, opened 2017-02-28
[ISPN-7535] Cache creation requires specific permissions when using s…

gaol:
…ecurity manager

Move Util.getClassLoaders() and ReflectionUtil.invokeAccessibly() to doPrivileged() when security manager is enabled

Jira: https://issues.jboss.org/browse/ISPN-7535

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/SecurityActions.java
+++ b/commons/src/main/java/org/infinispan/commons/util/SecurityActions.java
@@ -1,7 +1,12 @@
 package org.infinispan.commons.util;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.util.Arrays;
+
+import org.infinispan.commons.CacheException;
 
 /**
  * Privileged actions for the package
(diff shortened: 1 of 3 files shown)


tristantarrant:
This should be done using the SecurityActions pattern already adopted in other parts of the code, e.g. org.infinispan.jmx.SecurityActions

gaol:
Done, would you please review it again, if it is OK, I will back port it to `8.2.x` and `8.1.x` branches, Thank you! :)

tristantarrant:
I meant it should look like https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/jmx/SecurityActions.java

commons/src/main/java/org/infinispan/commons/util/SecurityActions.java
```diff
@@ -70,4 +75,62 @@ static String getProperty(String name) {
 
       return SysProps.PRIVILEGED.getProperty(name);
    }
+
+   private static <T> T doPrivileged(PrivilegedAction<T> action) {
+      if (System.getSecurityManager() != null) {
+         return AccessController.doPrivileged(action);
+      } else {
+         return action.run();
```
gaol: @tristantarrant  OK, I updated it following the pattern in `org.infinispan.jmx.SecurityActions`. One thing to mention is that   it does not call `Security.doPrivileged(action)` here like what it does in `org.infinispan.jmx.SecurityActions`, because:

- `org.infinispan.security.Security` is in `core` module, this change is done in `commons` module, `core` depends on `commons`.
- Using `Security.doPrivileged(action)` will change the original way when there is no security manager enabled.

slaskawi:
@gaol @tristantarrant Any update on this? Seems like we didn't get any progress since 11 days...

gaol:
@tristantarrant Would you please review this? Thank you very much!

The associated https://issues.jboss.org/browse/JBEAP-8416 is scheduled for EAP 7.0.6, which will component freeze soon.

tristantarrant:
Pushed to master, thanks

PR #4921, opened 2017-02-28
ISPN-7539 Do not configure data container key/value equivalence in tests

anistor:
* this is deprecated and a no-op anyway

https://issues.jboss.org/browse/ISPN-7539


tristantarrant:
Merged, thanks

PR #4922, opened 2017-02-28
ISPN-7533 Enable anonymous logins by default

vjuranek:
https://issues.jboss.org/browse/ISPN-7533


wburns:
I am not sure what overall stance is here, but I personally would say anonymous login should be disabled by default.  But if we do want this enabled by default, we definitely have to document this in the upgrade guide [1].

[1] https://github.com/infinispan/infinispan/blob/master/documentation/src/main/asciidoc/upgrading/upgrading.asciidoc

tristantarrant:
I have recently disabled anonymous logins when authentication is enabled because that is the sane thing to do. Going back to the previous behaviour is a mistake and I reject it. Any objections should be discussed on the mailing list

wburns:
@tristantarrant burning the midnight oil I see

PR #4924, opened 2017-03-01
ISPN-5654 Remove DeadlockDetectingLockManager

danberindei:
https://issues.jboss.org/browse/ISPN-5654


core/src/main/java/org/infinispan/configuration/cache/DeadlockDetectionConfiguration.java
```diff
@@ -30,14 +33,20 @@ static AttributeSet attributeDefinitionSet() {
    /**
     * Time period that determines how often is lock acquisition attempted within maximum time
     * allowed to acquire a particular lock
+    *
+    * @deprecated Since 9.0, always returns {@code -1}.
     */
+   @Deprecated
    public long spinDuration() {
-      return attributes.attribute(SPIN_DURATION).get();
+      return spinDuration.get();
```
rvansa: The field should not be removed, it's public API, but it should be marked `@Deprecated`, too.

danberindei: @rvansa Agree, we should keep it. I'm not sure about adding another @Deprecated annotation, because the class already has it. I added it for the fluent methods, but only because they can be used without referencing the class.

pruivo:
@danberindei looks good. just 2 unused imports in ```DeadlockDetectionConfigurationBuilder```

extended-statistics/src/test/java/org/infinispan/stats/logic/OptimisticLockingTxClusterExtendedStatisticLogicTest.java
```diff
@@ -126,7 +124,6 @@ public void testWriteSkewOnOwner() throws Exception {
       doWriteSkewTest(true);
    }
 
-   @Test (groups = "unstable", description = "https://issues.jboss.org/browse/ISPN-3341")
```
pruivo: the test is still unstable (40% success rate)

pruivo: better, apply this patch to your PR: https://gist.github.com/pruivo/b16f8005135571458b707ad0314be796

danberindei: Sorry, I meant to enable this just so I could get more results from CI, but forgot about it.
Your patch looks good, I just changed the code to use `eventuallyEquals()`. (My new favourite!)

PR #4926, opened 2017-03-02
ISPN-7540 Calling getCache(String, String) - template overrides cache

wburns:
definition

* Throw exception when redefining a cache

https://issues.jboss.org/browse/ISPN-7540


wburns:
Need to verify this doesn't break other modules, core was passing for me at least though.

wburns:
Have a lot of tests in other modules that need to be looked at.

PR #4928, opened 2017-03-02
ISPN-7065 Collect clustered memory statistics.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7065

### DIFF
--- a/core/src/main/java/org/infinispan/stats/impl/AbstractClusterStats.java
+++ b/core/src/main/java/org/infinispan/stats/impl/AbstractClusterStats.java
@@ -1,0 +1,183 @@
+package org.infinispan.stats.impl;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+import org.infinispan.commons.CacheException;
+import org.infinispan.factories.annotations.Inject;
+import org.infinispan.factories.annotations.Start;
+import org.infinispan.jmx.JmxStatisticsExposer;
+import org.infinispan.jmx.annotations.DataType;
+import org.infinispan.jmx.annotations.DisplayType;
+import org.infinispan.jmx.annotations.ManagedAttribute;
+import org.infinispan.jmx.annotations.ManagedOperation;
+import org.infinispan.jmx.annotations.Parameter;
+import org.infinispan.jmx.annotations.Units;
+import org.infinispan.util.TimeService;
+import org.infinispan.util.logging.Log;
+
+/**
+ * @author Ryan Emerson
+ * @since 9.0
+ */
+public abstract class AbstractClusterStats implements JmxStatisticsExposer {
+
+   public static final long DEFAULT_STALE_STATS_THRESHOLD = 3000;
+
+   private TimeService timeService;
+   private volatile long staleStatsThreshold = DEFAULT_STALE_STATS_THRESHOLD;
+   private volatile long statsUpdateTimestamp = 0;
+   volatile boolean statisticsEnabled = false;
+
+   private final Log log;
+   private final AtomicLong resetNanoseconds = new AtomicLong(0);
+   final HashMap<String, Number> statsMap = new HashMap<>();
+
+   AbstractClusterStats(Log log) {
+      this.log = log;
+   }
+
+   abstract void updateStats() throws Exception;
+
+   @Inject
+   void init(TimeService timeService) {
+      this.timeService = timeService;
+   }
+
+   @Start
+   void start() {
+      setStatisticsEnabled(statisticsEnabled);
+   }
+
+   public void reset() {
+      statsMap.clear();
+   }
+
+   @ManagedAttribute(description = "Gets the threshold for cluster wide stats refresh (milliseconds)",
+         displayName = "Stale Stats Threshold",
+         dataType = DataType.TRAIT,
+         writable = true)
+   public long getStaleStatsThreshold() {
+      return staleStatsThreshold;
+   }
+
+   public void setStaleStatsThreshold(long staleStatsThreshold) {
+      this.staleStatsThreshold = staleStatsThreshold;
+   }
+
+   @Override
+   @ManagedOperation(description = "Resets statistics gathered by this component", displayName = "Reset statistics")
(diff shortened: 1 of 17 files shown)


tristantarrant:
These stats seem to be per-cluster and not per-cache. We should probably introduce a ClusterContainerStats MBean to collect global stats like these.

ryanemerson:
@tristantarrant Updated to use a dedicated ClusterContainerStats MBean

core/src/main/java/org/infinispan/stats/impl/AbstractClusterStats.java
```diff
@@ -0,0 +1,162 @@
...
+   public static final long DEFAULT_STALE_STATS_THRESHOLD = 3000;
+
+   protected TimeService timeService;
+   private boolean statisticsEnabled = false;
+   private long staleStatsTreshold = DEFAULT_STALE_STATS_THRESHOLD;
+   private long statsUpdateTimestamp = 0;
+
+   private final Log log;
+   private final AtomicLong resetNanoseconds = new AtomicLong(0);
+   final HashMap<String, Number> statsMap = new HashMap<>();
```
wburns: Looks like it can be private

core/src/main/java/org/infinispan/stats/impl/AbstractClusterStats.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * @author Ryan Emerson
+ * @since 9.0
+ */
+public abstract class AbstractClusterStats implements JmxStatisticsExposer {
+
+   public static final long DEFAULT_STALE_STATS_THRESHOLD = 3000;
+
+   protected TimeService timeService;
```
wburns: I think injecting the `timeService` in this class is a bit nicer.

core/src/main/java/org/infinispan/stats/impl/AbstractClusterStats.java
```diff
@@ -0,0 +1,162 @@
...
+   private boolean statisticsEnabled = false;
```
wburns: I am not sure yet, but maybe we want this volatile? It probably doesn't hurt to misread it though...

core/src/main/java/org/infinispan/stats/impl/AbstractClusterStats.java
```diff
@@ -0,0 +1,162 @@
...
+      this.log = log;
+   }
+
+   abstract void updateStats() throws Exception;
+
+   void reset() {
+      statsMap.clear();
+   }
+
+   public long getStaleStatsTreshold() {
```
wburns: Looks like we need an attribute here. If we can set the value via JMX we should be able to read it at as well.

core/src/main/java/org/infinispan/stats/impl/ClusterContainerStatsFactory.java
```diff
@@ -0,0 +1,20 @@
...
+import org.infinispan.factories.AbstractComponentFactory;
+import org.infinispan.factories.AutoInstantiableFactory;
+import org.infinispan.factories.annotations.DefaultFactoryFor;
+import org.infinispan.stats.ClusterContainerStats;
+
+/**
+ * @author Ryan Emerson
+ * @since 9.0
+ */
+@SuppressWarnings("unused")
```
wburns: Is this needed? I don't have it in compiler, but I am surprised it gives an unused warning for a public class.

core/src/main/java/org/infinispan/stats/impl/AbstractClusterStats.java
ryanemerson: This has to be package private or protected as it is accessed in `ClusterCacheStatsImpl::updateTimeSinceStart` and `ClusterCacheStatsImpl::updateStats`

core/src/main/java/org/infinispan/stats/impl/ClusterContainerStatsFactory.java
ryanemerson: It was to shut Intellij up, I've removed it.

core/src/main/java/org/infinispan/stats/impl/AbstractClusterStats.java
ryanemerson: I attempted this, but I ran into issues with a NullPointerException as @Inject was always being called on the subclass first and the `setStatisticsEnabled(..)` method relies on the TimeService.

ryanemerson:
@wburns Updated

ryanemerson:
Updated again so that `getStaleStatsThreshold() ` is exposed via JMX and DMR, as well as creating a `ClusterStats` interface for common methods.

core/src/main/java/org/infinispan/stats/impl/AbstractClusterStats.java
pruivo: it seems the ```timeService``` is used to reset ```resetNanoseconds```.  I guess that can be done in a start method instead of in the injecting dependencies method.

ryanemerson: Good point, updated.

core/src/main/java/org/infinispan/stats/impl/AbstractClusterStats.java
```diff
@@ -0,0 +1,177 @@
...
+   }
+
+   @ManagedOperation(description = "Gets the threshold for cluster wide stats refresh (milliseconds)",
+         name = "getStaleStatsThreshold",
+         displayName = "Get the cluster wide stats refresh threshold")
+   public long getStaleStatsThreshold() {
+      return staleStatsThreshold;
+   }
+
+   @ManagedOperation(description = "Sets the treshold for cluster wide stats refresh (milliseconds)",
```
wburns: We should be able to use 1 Operation definition just like we have with `isStatisticsEnabled`.

ryanemerson: Done

danberindei:
@wburns @ryanemerson I think `getStatAsLong()`/`getStatAsInt()` would be a bit better.

ryanemerson:
@danberindei updated.

wburns:

PR #4931, opened 2017-03-02
Query Docs

gustavonalle:
Query docs revisited.

* New structure
* Brought back remote queries doc from hiding
* Brain dump
* Marked missing sections as TODO


documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+      .index(Index.PRIMARY_OWNER)
+      .addProperty("default.indexmanager", AffinityIndexManager.class.getName())
+      .addProperty("default.sharding_strategy.nbr_of_shards", "4")
+
+----
+
+The `AffinityIndexManager` by default will have as many shards as Infinispan segments, but this value is configurable as seen in the example above.
+
+The number of shards affects directly the query performance and writing throughput: generally speaking, a high number of shards offers better write throughput but
+has an adverse effect on query latency.
```
Sanne: s/query latency/query performance/ ?

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+
+====== directory-based index manager [[query.directory-based]]
+
+This is the default Index Manager used when no index manager is configured. The `directory-based` index manager is used to manage indexes backed by a local lucene directory. It supports _ram_, _filesystem_ and non-clustered
+_infinispan_ storage.
+
+
+[small]*Filesystem storage*
+
+This is the default storage, and used when index manager configuration is omitted. The index is stored in the filesystem using a link:https://lucene.apache.org/core/6_0_0/core/org/apache/lucene/store/MMapDirectory.html[MMapDirectory].
```
Sanne: the link refers to Lucene 6 .. I doubt it's different but it might confuse someone?
Maybe we could inject the version as a build property? (drawback is that the links might break in future w/o us noticing)

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+It is the recommended storage for local indexes. Although indexes are persistent on disk, they get memory mapped by Lucene and thus offers decent query performance.
```
Sanne: s/offers/offer/

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+      <property name="default.indexBase">${java.io.tmpdir}/baseDir</property>
+   </indexing>
+</replicated-cache>
+----
+
+Infinispan will create a different folder under `default.indexBase` for each entity (index) present in the cache.
+
+[small]*Ram storage*
+
+Index is stored in memory using a link:https://lucene.apache.org/core/6_0_0/core/org/apache/lucene/store/RAMDirectory.html[Lucene RAMDirectory].
```
Sanne: Lucene version (again).

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+    <indexing index="NONE" />
+</local-cache>
+
+----
+
+
+====== near-real-time index manager
+
+Similar to the `directory-based` index manager but takes advantage of the Near-Real-Time features of Lucene. It has better write performance than
+the `directory-based` because it flushes the index to the underlying store less often. The drawback is that it can cause loss of data in case
```
Sanne: s/data/ index updates/ ? Just to clarify that we wouldn't lose *Infinispan data*.

I think the NRT is usually a good option so I'd be careful to not scare people off.

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+This index manager forwards all indexes to an external Elasticsearch server. This is an experimental integration and some features may not be available,
+for example `indexNullAs` is link:https://hibernate.atlassian.net/browse/HSEARCH-2389[not currently supported].
+
+Configuration:
+
+[source,xml]
+----
+<indexing index="LOCAL">
+    <property name="default.indexmanager">elasticsearch</property>
+    <property name="default.elasticsearch.host">http://127.0.0.1:9200</property>
```
Sanne: should we use a made up IP other than localhost?
I would not suggest to run ES nodes locally on the same boxes.

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+</indexing>
+
+----
+
+The index mode should be set to `LOCAL`. More information about Elasticsearch integration can be found at the link:https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#elasticsearch-integration[Hibernate Search manual].
```
Sanne: About the "should be set to `LOCAL`". Should we clarify that it's because the ES cluster can be seen as a "shared index" ?
I'm thinking that we might want to clear this up, as someone might think to run various ES nodes possibly for different Infinispan shards.. yes people are "creative" ;)

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -194,28 +638,92 @@ The auto config properties for distributed caches are:
...
+| default.exclusive_index_use | true | indexing operation in exclusive mode, allowing Hibernate Search to optimize writes
+| default.indexmanager | org.infinispan.query.indexmanager.InfinispanIndexManager | Delegates index writing to a single node in the Infinispan cluster
+| default.reader.strategy | shared | Reuse index reader across several queries, avoiding reopening it
 |==============================================================================
 
+===== Re-indexing [[query.massindexer]]
+
+Occasionally you might need to rebuild the Lucene index by reconstructing it from the data stored in the Cache. You need to rebuild the index if you change the definition of what is indexed on your types, or if you change for example some _Analyzer_ parameter, as Analyzers affect how the index is defined. Also, you might need to rebuild the index if you had it destroyed by some system administration mistake. To rebuild the index just get a reference to the MassIndexer and start it; beware if might take some time as it needs to reprocess all data in the grid!
```
Sanne: s/how the index is defined/how the index is encoded/ ?

Not sure about my proposal, but it's important to stress that it affects the content of the index: beyond the definition.

Sanne: Please add some notice like: "We recommend you test a disaster recovery plan including running the MassIndexer. Understanding its working and tuning options in advance can make it much faster, and nobody wants to find out that their recovery process is going to take a very long time during an emergency: make sure it's tuned correctly so you can perform disaster recovery in short time."

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -194,28 +638,92 @@ The auto config properties for distributed caches are:
...
+===== Hybrid [[query.hybrid]]
 
-==== Lucene Directory
-Infinispan Query isn't aware of where you store the indexes, it just passes the configuration of which _Lucene Directory_ implementation you want to use to the Hibernate Search engine. There are several _Lucene Directory_ implementations bundled, and you can plug your own or add third party implementations: the Directory is the IO API for Lucene to store the indexes.
+TODO
 
-The most common _Lucene Directory_ implementations used with _Infinispan Query_ are:
+===== Mapping Entities
 
+Infinispan relies on Hibernate Search rich API in order to define fine grained configuration for indexing at entity level.
```
Sanne: s/Infinispan relies on Hibernate Search rich API/Infinispan relies on the rich API of Hibernate Search/

And a link to http://hibernate.org/search/ would be appreciated ;-)

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -194,28 +638,92 @@ The auto config properties for distributed caches are:
...
+This configuration includes which fields are annotated, which analyzers should be used, how to map nested objects and so on.
+Detailed documentation is available at link:https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#search-mapping[the Hibernate Search manual].
 
-* Ram - stores the index in a local map to the node. This index can't be shared.
-* Filesystem - stores the index in a locally mounted filesystem. This could be a network shared FS, but sharing this way is generally not recommended.
-* Infinispan - stores the index in a different dedicated Infinispan cache. This cache can be configured as replicated or distributed, to share the index among nodes. See also the dedicated chapter on the Lucene Directory in this guide.
+====== @DocumentId
+Unlike Hibernate Search, using _@DocumentId_ to mark a field as identifier does not apply to Infinispan values; in Infinispan Query the identifier for all _@Indexed_ objects is the key used to store the value. You can still customize how the key is indexed using a combination of _@Transformable_ , custom types and custom _FieldBridge_ implementations.
```
Sanne: "in Infinispan Query" .. 
Do you think "Infinispan Query" was introduced?

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -194,28 +638,92 @@ The auto config properties for distributed caches are:
...
+      CustomKey ck = (CustomKey) customType;
+      return ...
+   }
+}
+
+----
+
+[small]*Registering a Transformer programmatically*
+
+Using this technique, you don't have to annotated your custom key type:
```
Sanne: s/annotated/annotate/

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -232,7 +740,7 @@ properties.put("hibernate.search.[other options]", "[...]");
 Configuration infinispanConfiguration = new ConfigurationBuilder()
       .indexing()
          .enable()
-         .indexLocalOnly(true)
+         .indexLocalOnly(true)             //TODOD FIX CODE BASE
```
Sanne: what's wrong? Make sure to open a JIRA..

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -251,77 +759,191 @@ Assert.assertEquals(cq.getResultSize(), 1);
...
+                         .buildQueryBuilderForClass(Book.class).get()
+                         .range().onField("year").from(2005).to(2010)
+                         .createQuery();
+
+List<Object> results = searchManager.getQuery(luceneQuery).list();
 ----
 
-TIP: This is also available as a `start` JMX operation on the link:{javadocroot}/jmxComponents.html#MassIndexer[MassIndexer MBean]
-registered under the name `org.infinispan:type=Query,manager="{name-of-cache-manager}",cache="{name-of-cache}",component=MassIndexer`.
+For a detailed description of the query capabilities of this DSL, see the relevant section of the link:https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#section-building-lucene-queries[Hibernate Search manual].
```
Sanne: Might prefer to refer to the URLs of specific versions?

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -251,77 +759,191 @@ Assert.assertEquals(cq.getResultSize(), 1);
...
+
+List<Restaurant> nearBy = cacheQuery.list();
+----
+
+More info on http://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#spatial[Hibernate Search manual]
+
+
+====== Obtaining Lucene internal objects
+
+TODO
```
Sanne: Should we omit this?

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -814,9 +1447,302 @@ each query captures the minimal information needed both in terms of number of ma
...
 
-=== More Query DSL samples
 
-Probably the best way to explore using the Query DSL API is to have a look at our tests suite.
-link:https://github.com/infinispan/infinispan/blob/master/query/src/test/java/org/infinispan/query/dsl/embedded/QueryDslConditionsTest.java[QueryDslConditionsTest]
-is a fine example.
+=== Remote Querying [[query.remote]]
+
+While previous Infinispan versions were already providing <<querying_infinispan,indexing and searching>> of Java
+entities to embedded clients, starting with Infinispan 6.0 and the introduction of the new
```
Sanne: you really meant **6.0** ?

Sanne: (I'm not sure as it depends which aspect you mean.. if your really mean the 6.0 "new features" maybe there's no need to explain history anymore?)

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -814,9 +1447,302 @@ each query captures the minimal information needed both in terms of number of ma
...
+[source,java,numbered]
+----
+import org.infinispan.client.hotrod.configuration.ConfigurationBuilder;
+import org.infinispan.client.hotrod.marshall.ProtoStreamMarshaller;
+import org.infinispan.protostream.SerializationContext;
+...
+
+ConfigurationBuilder clientBuilder = new ConfigurationBuilder();
+clientBuilder.addServer()
+    .host("127.0.0.1").port(11234)
```
Sanne: let's make up a different hostname?

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -814,9 +1447,302 @@ each query captures the minimal information needed both in terms of number of ma
...
+----
+
+The key part of creating a query is obtaining the _QueryFactory_ for the remote cache using the
+_org.infinispan.client.hotrod.Search.getQueryFactory()_ method. Once you have this creating the query is similar to
+embedded mode which is covered in link:#infinispan_s_query_dsl[this] section.
+
+
+=== Statistics
+
+Query link:http://docs.jboss.org/hibernate/search/4.4/api/org/hibernate/search/stat/Statistics.html[_Statistics_]
```
Sanne: 4.4 is very old

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -814,9 +1447,302 @@ each query captures the minimal information needed both in terms of number of ma
...
+can be obtained from the _SearchManager_, as demonstrated in the following code snippet.
+
+[source,java]
+----
+SearchManager searchManager = Search.getSearchManager(cache);
+org.hibernate.search.stat.Statistics statistics = searchManager.getStatistics();
+----
+
+TIP: This data is also available via JMX through the link:http://docs.jboss.org/hibernate/search/4.4/reference/en-US/html/search-monitoring.html#d0e7624[Hibernate Search StatisticsInfoMBean]
```
Sanne: change version of Search. (parameters?)

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -814,9 +1447,302 @@ each query captures the minimal information needed both in terms of number of ma
...
+registered under the name `org.infinispan:type=Query,manager="{name-of-cache-manager}",cache="{name-of-cache}",component=Statistics`.
+Please note this MBean is always registered by Infinispan but the statistics are collected only if
+link:#enabling_jmx_statistics[statistics collection is enabled] at cache level.
+
+WARNING: Hibernate Search has its own configuration properties `hibernate.search.jmx_enabled` and `hibernate.search.generate_statistics`
+for JMX statistics as explained link:http://docs.jboss.org/hibernate/search/4.4/reference/en-US/html/search-monitoring.html#d0e7595[here].
```
Sanne: link version again.

gustavonalle:
@Sanne thanks! +1 to push and iterate it afterwards: let me just fix the typos and remove some ancient references

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -1,24 +1,75 @@
...
+are done on top of Java objects.
+
+====  Quick example
+
+We're going to store _Book_ instances in an Infinispan cache called "books". _Book_ instances will be indexed, so we enable indexing for the cache,
+letting Infinispan <<query.autoconfig,configure the indexing automatically>>:
+
+Declarative:
+
+[source,xml]
```
gustavonalle: I don't think so, it's just a snippet of config that is not referenced anywhere else.

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -1,24 +1,75 @@
...
+
+[source,java]
+----
+import org.infinispan.Cache;
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.configuration.cache.Index;
+import org.infinispan.configuration.global.GlobalConfigurationBuilder;
+import org.infinispan.manager.DefaultCacheManager;
+
+GlobalConfigurationBuilder gcb = GlobalConfigurationBuilder.defaultClusteredBuilder();
```
gustavonalle: This programmatic snippet is self contained; it is not related to the configuration above. I'll clarify it.

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
-To enable indexing via XML, you need to add the `<indexing />` element to your cache configuration, and optionally pass additional properties to the embedded Hibernate Search engine:
+====== Index names
+
+Each property inside the `index` element is prefixed with the index name, example:
+
+[source,xml]
+----
+    ...
+    <indexing index="ALL">
+        <property name="org.infinispan.sample.Car.directory_provider">ram</property>
```
gustavonalle: This section is presenting how index naming works; for the `Boat` entity, it is using a shorthand index name. The goal was to show the possible ways of configuring it.

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+queries to all members and aggregate the results. The downside is that Lucene does not allow more than a process writing to the index at the same time, and the coordination of lock acquisitions needs to be done by a proper shared index capable index manager.
+In any case, having a single write lock cluster-wise can lead to some degree of contention under heavy writing.
+
+Infinispan support shared indexes leveraging the <<integrations.adoc#integrations:directory-provider,Infinispan Directory Provider>>, which stores indexes in a separate set of caches. Two index managers are available
+to use shared indexes: InfinispanIndexManager and AffinityIndexManager.
+
+====== Effect of the index mode
+
+Shared indexes should not use the `ALL` index mode since it'd lead to redundant indexing: since there is a single index cluster wide, the entry would get indexed when inserted via Cache API, and another time
+when Infinispan replicates it to another node.
```
gustavonalle: +1 https://issues.jboss.org/browse/ISPN-7560

documentation/src/main/asciidoc/user_guide/query.adoc
gustavonalle: I'd add the notice, but the ISPN MassIndexer has no tuning option  :)

documentation/src/main/asciidoc/user_guide/query.adoc
gustavonalle: not sure I got your question

documentation/src/main/asciidoc/user_guide/query.adoc
Sanne: hummm good point :-/
Refer to index write tuning? Even just setting `default.indexwriter.ram_buffer_size` and the various merge options can get a solid boost.

documentation/src/main/asciidoc/user_guide/query.adoc
Sanne: It doesn't feel like that we explained that "all this stuff" is something we call the "Infinispan Query" component.
There's a "Query" chapter. But I'm not sure it's clear here what we mean with making a comparison between "Hibernate Search" [unlike] "Infinispan Query". So far we just mentioned that Infinispan's _queries_ use Hibernate Search..

gustavonalle: I see, and you're right. I will rephrase that.

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
 
 ----
 
-In this example the index is stored in memory, so when this nodes is shutdown the index is lost: good for a quick demo, but in real world cases you'll want to use the default (store on filesystem) or store the index in Infinispan as well. For the complete reference of properties to define, refer to the link:http://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#search-configuration[Hibernate Search documentation] .
+or programmatically:
+
+[source,java]
+----
+ cacheCfg.indexing()
+      .index(Index.ALL)
```
gustavonalle: I agree with you, in fact this config is called `IndexMode` internally. 

Since we're going final in 2 days, we'll keep as it is for now.  Going forward, there's a plan to deprecated this config and downgrade it to index manager level, since a cache can have more than one entity type, each with its own IndexManager with a different requirement of IndexMode.

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+
+* ALL: all data is indexed, local and remote.
+* LOCAL: only local data is indexed.
+* PRIMARY_OWNER: Only entries containing keys that the node is primary owner will be indexed, regardless of local or remote origin.
+* NONE: no data is indexed. Equivalent to not configure indexing at all.
+
+===== Index Managers [[query.index-manager]]
+
+Index managers are central components in Infinispan Querying responsible for the indexing configuration, distribution and internal lifecycle of several query components such as Lucene's
+_IndexReader_ and _IndexWriter_. Each Index Manager is associated with one of more types of _Directory Provider_, which defines the physical storage of the index.
```
gustavonalle: right, let me rephrase that, the idea is a 1-1 mapping between IM and DP.

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+
+Regarding index distribution, Infinispan can be configured with shared or non-shared indexes.
+
+===== Shared indexes [[query.shared-index]]
+
+A shared index is a single, distributed, cluster-wide index for a certain cache. The main advantage is that the index is visible from every node and can be queried as if the index were local, there is no need to <<query.clustered-query-api,broadcast>>
```
gustavonalle: Not sure, I think `were` sounds better :smile:
Let's ask a native english speaker, wdyt @ryanemerson, "as if the index were local"  or "as if the index was local" ?

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+The _LuceneIndexesLocking_ cache is used to store Lucene locks, and it is a very small cache: it will contains one entry per entity (index).
+
+The _LuceneIndexesMetadata_ cache is used to store info about the logical files that are part of the index, such as names, chunks and sizes and it is also
+small in size.
+
+The _LuceneIndexesData_ cache is the where most of the index is located: it is much bigger then the other two but should be smaller than the data in the cache itself, thanks to Lucene
+efficient storing techniques.
+
+The configuration above is changing the name of those 3 caches and redefining their configuration. Reasons to do so is performance tuning: since two of the 3 caches are small
+and heavily accessed, it makes sense to make them replicated, while the largest cache is configured as distributed. Manually redefining the index caches also
```
gustavonalle: Since you mentioned, I think it became the default very recently, let me double check

Sanne: N.B. this paragraph is talking about optionally "redefining their configuration", so it doesn't really matter what's the default as we're just giving guidance about how to redefine things in a sensible way.

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+[source,xml]
+----
+<replicated-cache name="myCache">
+   <indexing index="ALL">
+      <property name="default.directory_provider">ram</property>
+   </indexing>
+</replicated-cache>
+----
+
+[small]*Infinispan storage*
```
gustavonalle: It can be persistent by configuring those 3 caches with a cache store; even without a cache store, it's less "volatile" than `ram` since the cache will likely be replicated, so there's some redundancy in case of 1 server going down. That's a good point, I will add it to the doc

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -92,73 +144,468 @@ for (Object book : objectList) {
...
+
+===== External indexes
+
+Apart from having shared and non-shared indexes managed by Infinispan itself, it is possible to offload indexing to a third party search engine: currently
+Infinispan supports Elasticsearch as a external index storage.
+
+====== Elasticsearch IndexManager (experimental)
+
+This index manager forwards all indexes to an external Elasticsearch server. This is an experimental integration and some features may not be available,
+for example `indexNullAs` is link:https://hibernate.atlassian.net/browse/HSEARCH-2389[not currently supported].
```
gustavonalle: Right, will clarify it.

documentation/src/main/asciidoc/user_guide/query.adoc
```diff
@@ -251,77 +759,191 @@ Assert.assertEquals(cq.getResultSize(), 1);
...
+--------------------------------------
+
+For more info about Faceted Search, see link:http://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#query-faceting[Hibernate Search Faceting]
+
+====== Spatial Queries
+
+Infinispan also support link:https://en.wikipedia.org/wiki/Spatial_query[Spatial Queries], allowing to combining full-text with restrictions based on distances, geometries or geographic coordinates.
 
-=== Infinispan's Query DSL
+Example, we start by using special annotations in our entity that will be searched, together with latitude and longitude:
```
gustavonalle: Both words fit here, I will move to "spatial"

documentation/src/main/asciidoc/user_guide/query.adoc
ryanemerson: @gustavonalle @yrodiere My vote is for "as if the index was local" as you are referring to a singular index.

Sanne: "were" is the recommended form according to vocabulary as the index is not really local, but both forms are acceptable. I'd pick "were" as it makes the text unequivocal (the index shall not be considered local in the discussed case).

documentation/src/main/asciidoc/user_guide/query.adoc
yrodiere: @Sanne I guess, but when reading this I had the impression that the default was less optimal than this example, and that may lead some users to needlessly redefine the configuration.

Sanne:
That's fair. We provide a safe default which should perform fine in an
"average environment" bit in reality some tuning will always be needed.. As
no one runs on the average ;)

documentation/src/main/asciidoc/user_guide/query.adoc
gustavonalle: Keeping it from now, we can improve it later

documentation/src/main/asciidoc/user_guide/query.adoc
gustavonalle: Yes, lucene IW configs would speed-up the mass indexing. We'll add it later https://issues.jboss.org/browse/ISPN-7560

documentation/src/main/asciidoc/user_guide/query.adoc
gustavonalle: Nothing wrong, just ancient code that needs to get updated :)

documentation/src/main/asciidoc/user_guide/query.adoc
gustavonalle: ok

documentation/src/main/asciidoc/user_guide/query.adoc
gustavonalle: Will sort it out later once variables are available for both lucene and hs https://issues.jboss.org/browse/ISPN-7560

documentation/src/main/asciidoc/user_guide/query.adoc
gustavonalle: This documentation piece was just copied to the new structure, I will remove the old versions for now. There's a lot of docs to be written for remote query still.

documentation/src/main/asciidoc/user_guide/query.adoc
gustavonalle: I saw other grammar theories around that justify the use of `were`, so let's keep it.

documentation/src/main/asciidoc/user_guide/query.adoc
gustavonalle: I will clarify about the defaults

gustavonalle:
Addressed all reviews and added JIRAs for those not addressed, let's integrated this and continue the work on master?

Sanne:
@gustavonalle agreed! Feel free to push..

PR #4941, opened 2017-03-07
ISPN-7500 ClusterListenerDistInitialStateTest emits NPE warning

pruivo:
https://issues.jboss.org/browse/ISPN-7500

### DIFF
--- a/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
+++ b/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
@@ -744,9 +744,16 @@ public class JGroupsTransport extends AbstractTransport implements MembershipLis
          }
          return;
       }
-      dispatcher.sendMessage(
+      CommandAwareRpcDispatcher rpcDispatcher = dispatcher;
+      if (rpcDispatcher == null) {
+         if (trace) {
+            log.trace("Don't send message. Transport is stopped.");
+         }
+         return;
+      }
+      rpcDispatcher.sendMessage(
             toJGroupsAddress(destination),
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -744,9 +744,16 @@ public void sendTo(Address destination, ReplicableCommand rpcCommand, DeliverOrd
          }
          return;
       }
-      dispatcher.sendMessage(
+      CommandAwareRpcDispatcher rpcDispatcher = dispatcher;
+      if (rpcDispatcher == null) {
+         if (trace) {
+            log.trace("Don't send message. Transport is stopped.");
```
slaskawi: Shouldn't this be at least a WARN?

pruivo: why? if the ```dispatcher``` is ```null``` then the ```channel``` is already disconnected and the node left the view. There is no need to spam the log with this message.

wburns: Maybe we should have a private method that does the null check and logs the message?  This way we don't have to duplicate the code check in multiple places.

pruivo: you still need to duplicate the check since the method had to return. I don't think it would save anything...

wburns:
Waiting for https://issues.jboss.org/browse/ISPN-6971 as Dan mentioned he may be modifying this code as well.

wburns:
Closing for now, can be reopened if needed later

PR #4944, opened 2017-03-07
ISPN-7571 Add a declarative JDBC config IT to the wildfly modules

tristantarrant:
https://issues.jboss.org/browse/ISPN-7571

### DIFF
--- a/integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/InfinispanStoreJdbcIT.java
+++ b/integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/InfinispanStoreJdbcIT.java
@@ -2,6 +2,8 @@ package org.infinispan.test.integration.as;
 
 import static org.junit.Assert.assertEquals;
 
+import java.io.IOException;
+
 import org.infinispan.Cache;
 import org.infinispan.Version;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
@@ -32,7 +34,10 @@ public class InfinispanStoreJdbcIT {
 
    @Deployment
    public static Archive<?> deployment() {
-      return ShrinkWrap.create(WebArchive.class, "jdbc.war").addClass(InfinispanStoreJdbcIT.class).add(manifest(), "META-INF/MANIFEST.MF");
+      return ShrinkWrap.create(WebArchive.class, "jdbc.war")
+            .addClass(InfinispanStoreJdbcIT.class)
+            .addAsResource("jdbc-config.xml")
+            .add(manifest(), "META-INF/MANIFEST.MF");
    }
 
    private static Asset manifest() {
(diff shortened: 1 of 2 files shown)


integrationtests/wildfly-modules/src/test/resources/jdbc-config.xml
```diff
@@ -0,0 +1,29 @@
...
+<infinispan
+   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+   xsi:schemaLocation="urn:infinispan:config:9.0 http://www.infinispan.org/schemas/infinispan-config-9.0.xsd
+   urn:infinispan:config:store:jdbc:9.0 http://www.infinispan.org/schemas/infinispan-cachestore-jdbc-config-9.0.xsd"
+   xmlns="urn:infinispan:config:9.0"
+   xmlns:jdbc="urn:infinispan:config:store:jdbc:9.0">
+
+  <cache-container default-cache="default">
+    <local-cache name="default">
+      <locking isolation="REPEATABLE_READ"
```
wburns: Any reason for the locking?

integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/InfinispanStoreJdbcIT.java
```diff
@@ -58,11 +63,31 @@ public void testCacheManager() {
...
+      }
+   }
+
+   @Test
+   public void testXmlConfig() throws IOException {
+      EmbeddedCacheManager cm = null;
+      try {
+         cm = new DefaultCacheManager("jdbc-config.xml");
+         Cache<String, String> cache = cm.getCache("anotherCache");
+         cache.put("a", "a");
```
wburns: Should we verify the cache store is actually present? Using something like

`
PersistenceManager clm = TestingUtil.extractComponent(cache, PersistenceManager.class);
assertEquals(1, clm.getStores(JdbcStringBasedStore.class).size());
`

integrationtests/wildfly-modules/src/test/java/org/infinispan/test/integration/as/InfinispanStoreJdbcIT.java
```diff
@@ -32,7 +34,10 @@
 
    @Deployment
    public static Archive<?> deployment() {
-      return ShrinkWrap.create(WebArchive.class, "jdbc.war").addClass(InfinispanStoreJdbcIT.class).add(manifest(), "META-INF/MANIFEST.MF");
+      return ShrinkWrap.create(WebArchive.class, "jdbc.war")
+            .addClass(InfinispanStoreJdbcIT.class)
+            .addAsResource("jdbc-config.xml")
+            .add(manifest(), "META-INF/MANIFEST.MF");
```
Sanne: This test should pass even w/o explicitly declaring a dependency on the `org.infinispan.persistence.jdbc` module.

One shouldn't repeat himself by having to enable the CacheStore explicitly in two different places:
 - the configuration file
 - the module structure
 - (and a third!) having the depenency

Steps 1 and 3 are widely accepted, so hardly surprising. Step 2 should be handled automatically by Infinispan.

I appreciate you might disagree, but in that case at least throw a better error message pointing out what's missing rather than a generic parser exception.

slaskawi: hmmm do we need an extra JIRA for this?

Sanne: it depends if you all agree .. 

That's my opinion as a "defender of usability" but I might be missing that it's impossible to achieve.

I just hope we can improve.. @tristantarrant ?

tristantarrant: I think the org.infinispan:main module should import all optional store modules.

Sanne: yes that's https://issues.jboss.org/browse/ISPN-7572

In addition, wondering here if for other modules (the ones which we don't know about and hence will not be able to list in org.infinispan:main) should have a better error to point out that this dependency is needed .. ?

Sanne:
I'll take this, to be merged first with @wburns 's suggestions and then I'll use it to follow up with ISPN-7572

Sanne:
merged

PR #4946, opened 2017-03-08
ISPN-5654 Deprecate/remove more deadlock detection-related classes

danberindei:
Follows up on https://github.com/infinispan/infinispan/pull/4924, to remove the interceptor and deprecate classes like `DldGlobalTransaction`.


core/src/main/java/org/infinispan/commands/CommandsFactoryImpl.java
```diff
@@ -226,7 +223,6 @@ public void start() {
       cacheName = ByteString.fromString(cache.getName());
       this.transactional = configuration.transaction().transactionMode().isTransactional();
       this.totalOrderProtocol = configuration.transaction().transactionProtocol().isTotalOrder();
-      this.deadlockDetection = configuration.deadlockDetection().enabled();
       this.distributedOrReplicated = configuration.clustering().cacheMode().isDistributed() || configuration.clustering().cacheMode().isReplicated();
```
pruivo: no longer used as well.

pruivo:
testing locally...

PR #4951, opened 2017-03-08
ISPN-7540 Calling getCache(String, String) - template overrides cache

wburns:
* Added version for deprecation
* Changed internal method name

https://issues.jboss.org/browse/ISPN-7540


danberindei:
Integrated, thanks Will!

PR #4956, opened 2017-03-09
ISPN-7582 Fix and update rolling upgrade tests

vjuranek:
https://issues.jboss.org/browse/ISPN-7582

* Update default HR version
* Improve error message when getting HR codec
* Update README for feature-pack
* Update and fix rolling upgrade tests

### DIFF
--- a/server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
+++ b/server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
@@ -129,7 +129,7 @@
          <cache-container name="local" default-cache="default">
             <global-state />
             <local-cache name="default">
-               <remote-store cache="default" socket-timeout="60000" tcp-no-delay="true" shared="true" hotrod-wrapping="true" purge="false" passivation="false">
+               <remote-store cache="default" socket-timeout="60000" tcp-no-delay="true" shared="true" hotrod-wrapping="true" protocol-version="2.0" purge="false" passivation="false">
                   <remote-server outbound-socket-binding="remote-store-hotrod-server" />
                </remote-store>
             </local-cache>
--- a/server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
+++ b/server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
@@ -129,7 +129,7 @@
          <cache-container name="local" default-cache="default">
             <global-state />
             <local-cache name="default">
-               <remote-store cache="default" socket-timeout="60000" tcp-no-delay="true" shared="true" hotrod-wrapping="true" protocol-version="2.0" purge="false" passivation="false">
+               <remote-store cache="default" socket-timeout="60000" tcp-no-delay="true" shared="true" hotrod-wrapping="true" purge="false" passivation="false">
                   <remote-server outbound-socket-binding="remote-store-hotrod-server" />
                </remote-store>
             </local-cache>
--- a/server/integration/testsuite/build-testsuite.xml
+++ b/server/integration/testsuite/build-testsuite.xml
@@ -168,13 +168,11 @@
                    modifyRemoteDestination="${other.parts}/rcs-remote-destination.xml"
                    modifyInfinispan="${infinispan.parts}/rolling-upgrades-clustered-${new.server.schema.version}.xml"
                    infinispanServerEndpoint="${other.parts}/rolling-upgrades-server-endpoint-${new.server.schema.version}.xml"
-                   addNewHotrodSocketBinding="${other.parts}/socket-binding-hotrod-store.xml"
                    modifyMulticastAddress="${other.parts}/switch-multicast-address.xml"/>
         <transform in="clustered.xml" out="examples/clustered-rest-rolling-upgrade.xml"
                    modifyRemoteDestination="${other.parts}/rest-remote-destination.xml"
                    modifyInfinispan="${infinispan.parts}/rolling-upgrades-clustered-rest-${new.server.schema.version}.xml"
                    infinispanServerEndpoint="${other.parts}/rolling-upgrades-server-endpoint-${new.server.schema.version}.xml"
-                   addNewRestSocketBinding="${other.parts}/socket-binding-rest-store.xml"
                    modifyMulticastAddress="${other.parts}/switch-multicast-address.xml"
                    removeRestSecurity="true"/>
         <transform in="examples/standalone-rest-rolling-upgrade.xml" out="examples/standalone-rest-rolling-upgrade.xml"
--- a/server/integration/testsuite/build-testsuite.xml
+++ b/server/integration/testsuite/build-testsuite.xml
@@ -127,34 +127,34 @@
         <if>
             <equals arg1="${zip.dist.old}" arg2="false" />
             <then>
-                <echo message="Zip for old server distribution was not passed. Skipping." />
+                <echo message="Zip for old server distribution was not passed. Using default old ISPN version ${version.org.infinispan.server-build.old}." />
             </then>
             <else>
                 <echo message="Using file ${zip.dist.old}" />
                 <unzip src="${zip.dist.old}" dest="${server.old.dist}-tmp" />
-                <for param="file">
-                    <path>
-                        <fileset dir="${server.old.dist}-tmp" includes="**/jboss-modules.jar" />
-                    </path>
-                    <sequential>
-                        <propertyregex override="yes" property="dir.in.zip" input="@{file}" regexp=".*[/\\]([^/\\]*)[/\\]jboss-modules\.jar" replace="\1" />
-                    </sequential>
-                </for>
-                <move file="${server.old.dist}-tmp/${dir.in.zip}" tofile="${server.old.dist}" />
-                <delete dir="${server.old.dist}-tmp" />
-                <transform transformDir="${server.old.dist}/standalone/configuration" in="standalone.xml" out="standalone.xml"
-                           removeRestSecurity="true" />
-                <transform transformDir="${server.old.dist}/standalone/configuration" in="clustered.xml" out="clustered.xml"
-                           removeRestSecurity="true"/>
-
-                <copy todir="${server2.old.dist}">
-                    <fileset dir="${server.old.dist}" />
-                </copy>
-                <copy todir="${server3.old.dist}">
-                    <fileset dir="${server.old.dist}" />
-                </copy>
             </else>
         </if>
+        <for param="file">
+            <path>
+                <fileset dir="${server.old.dist}-tmp" includes="**/jboss-modules.jar" />
+            </path>
+            <sequential>
+                <propertyregex override="yes" property="dir.in.zip" input="@{file}" regexp=".*[/\\]([^/\\]*)[/\\]jboss-modules\.jar" replace="\1" />
+            </sequential>
+        </for>
+        <move file="${server.old.dist}-tmp/${dir.in.zip}" tofile="${server.old.dist}" />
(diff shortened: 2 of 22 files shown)


server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
```diff
@@ -141,7 +141,7 @@
          <cache-container name="local" default-cache="default">
             <global-state />
             <local-cache name="default">
-               <remote-store cache="default" socket-timeout="60000" tcp-no-delay="true" shared="true" hotrod-wrapping="true" purge="false" passivation="false">
+               <remote-store cache="default" socket-timeout="60000" tcp-no-delay="true" shared="true" hotrod-wrapping="true" protocol-version="2.0" purge="false" passivation="false">
```
gustavonalle: Better omit the protocol version or put the lastest `2.6`. If using `2.0`, the rolling  ugrade process will not use the remote iterator and will fall back to the (bkoen) legacy behaviour

server/integration/testsuite/src/test/resources/config/infinispan/rolling-upgrades-clustered-7.0.xml
```diff
@@ -1,4 +1,4 @@
-        <subsystem xmlns="urn:infinispan:server:core:8.0" default-cache-container="clustered-new">
+        <subsystem xmlns="urn:infinispan:server:core:7.0" default-cache-container="clustered-new">
```
gustavonalle: IMO we can drop support for 7.0 clusters and test against the latest 8.2.x only

server/integration/testsuite/src/test/resources/config/infinispan/rolling-upgrades-clustered-9.0.xml
```diff
@@ -0,0 +1,21 @@
...
+
+                                     mode="SYNC"
+                                     segments="1"
+                                     owners="2"
+                                     batching="false"
+                                     l1-lifespan="0"
+                                     remote-timeout="30000">
+                    <transaction mode="NONE"/>
+                    <remote-store cache="default" socket-timeout="60000" tcp-no-delay="true" shared="true" hotrod-wrapping="true" protocol-version="2.0" purge="false" passivation="false">
+                        <remote-server outbound-socket-binding="remote-store-hotrod-server"/>
```
gustavonalle: As explained before, better to use a higher `protocol-version`

server/integration/testsuite/src/test/java/org/infinispan/server/test/rollingupgrades/HotRodRollingUpgradesIT.java
```diff
@@ -33,30 +33,7 @@
  */
 @RunWith(Arquillian.class)
 @Category({RollingUpgrades.class})
```
gustavonalle: It seems the `RollingUpgrades` category is not run by default, I think it should be. Maven/Arquillian could be used to provide the old server

mgencur: If they run by default, they will only test a rolling upgrade from "current" to "current" version. You would need to supply a path to an older server version that will be automatically unzipped. But that complicates running the suite in CI.

gustavonalle: @mgencur "current" to "current" is a valid test, we do that on Kubernetes for example to migrate data between same versions with incompatible configurations.
Regarding the `older server`, as I mentioned, we could use maven to grab and unpack the older version, wdyt?

gustavonalle: The servers are all available as artifacts here https://repo1.maven.org/maven2/org/infinispan/server/infinispan-server-build/
I don't expected we to test down to 5.2, but at least the migration between the current stable and the previous one. We could allow testing of older versions behind maven profiles if required.

vjuranek: I'm going to implement testing against `old.server.schema.version` which should be set to previous major ISPN release but can be also easily changes from cmd line/profile to test older releases.

server/integration/testsuite/src/test/resources/config/infinispan/rolling-upgrades-clustered-7.0.xml
vjuranek: I would run tests against last major ISPN release by default (and in CI), but would keep it if someone wants to test older version so that (s)he has everything prepared and can run tests immediatelly without need to change/understand the source code

server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
vjuranek: If we want to use remote iterators, the test would fail for servers which doesn't support HR 2.5 (IIRC implemented in ISPN 8.2, so fails also for 8.0). I'll probably changed it to 2.5 as this is used for generating examples (so that users are adviced to use 2.5 if they can) and generate config file via XSLT for tests, so that HR version can be overwritten by properties if needed

server/integration/testsuite/src/test/java/org/infinispan/server/test/rollingupgrades/HotRodRollingUpgradesIT.java
vjuranek: N.B.: "current" to "current" rolling upgrade is already tested (and runs by default) in `ExampleConfigsIT#testHotRodRollingUpgrades`, resp. `ExampleConfigsIT#testRestRollingUpgrades`

server/integration/testsuite/src/test/resources/config/infinispan/rolling-upgrades-clustered-7.0.xml
gustavonalle: +1

vjuranek:
All comments should be addressed:
* updated to HR version `2.6` which is used by default for rolling upgrades if protocol version is not specified
* removed HR protocol version from example configs (i.e. they are as before) and and new config for HR rolling upgrade is generated, if `hotrod.protocol.version` is specified, this version is used for HR rest store (but `ExampleConfigIT` still uses example config)
* if no `zip.dist.old` is specified, by default is used server version specified by `version.org.infinispan.server-build.old`, currently `8.0.0.Final` and in such case by default HR protocol version `2.0` is used (but still can e overwritten by `hotrod.protocol.version` in case of need, e.g. when specifying old release using `zip.dist.old`)
* all rolling upgrade  now runs by default
* REST rolling upgrade tests are disabled, as REST store seems to be broken (see [ISPN-7617](https://issues.jboss.org/browse/ISPN-7617))

server/integration/testsuite/build-testsuite.xml
```diff
@@ -122,61 +124,57 @@
         <if>
             <equals arg1="${zip.dist.old}" arg2="false" />
             <then>
-                <echo message="Zip for old server distribution was not passed. Skipping." />
+                <echo message="Zip for old server distribution was not passed. Using default old ISPN version." />
```
danberindei: Could we log the default server and protocol versions here?

vjuranek: done

gustavonalle:
Build did not go through, re-triggering...

vjuranek:
HR client testsuite hanged. I took a thread dump and kill the job no to block the machine. Will look on it later today or tomorrow.

vjuranek:
Stuck again. It's caused for some reason by upgrade default HR version to `2.6`. I removed HR protocol upgrade from PR, will investigate and send in separate PR.

vjuranek:
Failed ` HotRodRollingUpgradesDistIT` is very likely caused by [ISPN-7614](https://issues.jboss.org/browse/ISPN-7614) which was fixed yesterday in 4d54c2fd3d5f4aae17912777f114b13d8f95d767. Other failures are unrelated.

wburns:
Seems this is not required for 9.0 as it is just test bugs. Please update if this assumption is incorrect.

gustavonalle:
There's one related test failure:

 HotRodRollingUpgradesDistIT.testHotRodRollingUpgradesDiffVersionsDist  	
junit.framework.AssertionFailedError: Source node entries should NOT be accessible from target node (after RCS disconnection)
    at junit.framework.Assert.fail(Assert.java:57)
    at junit.framework.Assert.assertTrue(Assert.java:22)
    at junit.framework.Assert.assertNull(Assert.java:277)
    at junit.framework.TestCase.assertNull(TestCase.java:447)
    at org.infinispan.server.test.rollingupgrades.HotRodRollingUpgradesDistIT.testHotRodRollingUpgradesDiffVersionsDist(HotRodRollingUpgradesDistIT.java:129)

vjuranek:
@gustavonalle as mentioned in my comment above, this failure is not related to this PR and is caused by [ISPN-7614](https://issues.jboss.org/browse/ISPN-7614), which was already fixed. I did a rebase to cover this fix, it should pass now.

vjuranek:
@gustavonalle after rebase the failure is gone, there's only one failure, which is unrelated to this change

gustavonalle:
@vjuranek needs another rebase

vjuranek:
@gustavonalle rebased. When testing locally after rebase, I hit fail in `HotRodRollingUpgradesDistIT` again, but this is IMHO not related to this change.

```
09:34:26.147 [main] ERROR org.infinispan.commons.test.TestSuiteProgress - Test failed: HotRodRollingUpgradesDistIT.testHotRodRollingUpgradesDiffVersionsDist
junit.framework.AssertionFailedError: Source node entries should NOT be accessible from target node (after RCS disconnection)
```

gustavonalle:
@vjuranek checking

gustavonalle:
@vjuranek  I tried running ```HotRodRollingUpgradesDistIT``` locally, and it fails randomly.

I think the reason why it's because rolling upgrades from 8.0.0.Final are broken, they were on 8.2.x. 

We should support/test only from 8.2.6.Final earliest (and use protocol version "2.5"). Also ```HotRodRollingUpgradesDistIT``` doesn't disconnect the remote store in all nodes and uses ancient Hot Rod versions in the client (PROTOCOL_VERSION_12).

server/integration/versions/pom.xml
```diff
@@ -39,6 +39,7 @@
       <version.org.infinispan>${project.version}</version.org.infinispan>
       <version.org.infinispan.infinispan-management-console>9.0.0.Final</version.org.infinispan.infinispan-management-console>
       <version.org.infinispan.arquillian.container>1.2.0.Beta1</version.org.infinispan.arquillian.container>
+      <version.org.infinispan.server-build.old>8.0.0.Final</version.org.infinispan.server-build.old>
```
gustavonalle: This version is also in server/integration/testsuite/pom.xml, can we keep it in only one place?

vjuranek:
@gustavonalle updated. `HotRodRollingUpgradesDistIT` still fails but it's not related to this change, `disconnectSource` obviously doesn't disconnect the source

gustavonalle:
@vjuranek I believe the fix to the test is just a matter of adding a call to disconnect the remote store from the other target server, which is probably ```hotrod-rolling-upgrade-2-dist```

vjuranek:
rebased. 
Anything else which needs to be fixed @gustavonalle ?

gustavonalle:
Looking

gustavonalle:
@vjuranek Integrated with a fix for the randomly failing test, thanks!

PR #4957, opened 2017-03-09
ISPN-6029 DDL for JDBC store tables should never allow null

pferraro:
https://issues.jboss.org/browse/ISPN-6029

Please cherry-pick into 8.2.x.


ryanemerson:
@pferraro `TableManagerTest.testDrop()` needs updating as it currently only inserts an ID, hence it now fails.

pferraro:
@ryanemerson Updated.

ryanemerson:
@pferraro Integrated, thanks!

PR #4958, opened 2017-03-09
ISPN-7578 EntryWrappingInterceptor should not invoke tx commands synchronously

rvansa:
https://issues.jboss.org/browse/ISPN-7578

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
@@ -64,8 +64,8 @@ import org.infinispan.filter.CompositeKeyFilter;
 import org.infinispan.filter.KeyFilter;
 import org.infinispan.interceptors.DDAsyncInterceptor;
 import org.infinispan.interceptors.InvocationFinallyAction;
-import org.infinispan.interceptors.InvocationStage;
 import org.infinispan.interceptors.InvocationSuccessAction;
+import org.infinispan.interceptors.InvocationSuccessFunction;
 import org.infinispan.interceptors.locking.ClusteringDependentLogic;
 import org.infinispan.metadata.Metadata;
 import org.infinispan.notifications.cachelistener.CacheNotifier;
@@ -138,6 +138,8 @@ public class EntryWrappingInterceptor extends DDAsyncInterceptor {
    private final InvocationFinallyAction
          commitEntriesFinallyHandler = (rCtx, rCommand, rv, t) -> commitContextEntries(rCtx, null, null);
 
+   private final InvocationSuccessFunction prepareHandler = this::prepareHandler;
+
    protected Log getLog() {
       return log;
    }
@@ -181,11 +183,15 @@ public class EntryWrappingInterceptor extends DDAsyncInterceptor {
 
    @Override
    public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand command) throws Throwable {
-      wrapEntriesForPrepare(ctx, command);
-      if (!shouldCommitDuringPrepare(command, ctx)) {
+      return wrapEntriesForPrepareAndApply(ctx, command, prepareHandler);
+   }
+
+   private Object prepareHandler(InvocationContext ctx, VisitableCommand command, Object rv) {
+      if (shouldCommitDuringPrepare((PrepareCommand) command, (TxInvocationContext) ctx)) {
+         return invokeNextThenAccept(ctx, command, commitEntriesSuccessHandler);
+      } else {
          return invokeNext(ctx, command);
       }
-      return invokeNextThenAccept(ctx, command, commitEntriesSuccessHandler);
    }
 
    @Override
@@ -817,19 +823,33 @@ public class EntryWrappingInterceptor extends DDAsyncInterceptor {
              command.isOnePhaseCommit();
    }
 
-   protected final void wrapEntriesForPrepare(TxInvocationContext ctx, PrepareCommand command) throws Throwable {
+   protected final Object wrapEntriesForPrepareAndApply(TxInvocationContext ctx, PrepareCommand command, InvocationSuccessFunction handler) throws Throwable {
       if (!ctx.isOriginLocal() || command.isReplayEntryWrapping()) {
-         for (WriteCommand c : command.getModifications()) {
-            c.setTopologyId(command.getTopologyId());
-            // TODO: we need async invocation since the interceptors may do remote get
-            InvocationStage visitorStage = makeStage(c.acceptVisitor(ctx, entryWrappingVisitor));
-            // Wait for the sub-command to finish. If there was an exception, rethrow it.
-            visitorStage.get();
+         return applyModificationsAndThen(ctx, command, command.getModifications(), 0, handler);
+      }
+      return handler.apply(ctx, command, null);
+   }
 
-            if (c.hasAnyFlag(FlagBitSets.PUT_FOR_X_SITE_STATE_TRANSFER)) {
-               ctx.getCacheTransaction().setStateTransferFlag(Flag.PUT_FOR_X_SITE_STATE_TRANSFER);
+   private Object applyModificationsAndThen(TxInvocationContext ctx, PrepareCommand command, WriteCommand[] modifications, int startIndex, InvocationSuccessFunction handler) throws Throwable {
+      // We need to execute modifications for the same key sequentially. In theory we could optimize
+      // this loop if there are multiple remote invocations but since remote invocations are rare, we omit this.
+      for (int i = startIndex; i < modifications.length; i++) {
+         WriteCommand c = modifications[i];
+         c.setTopologyId(command.getTopologyId());
+         if (c.hasAnyFlag(FlagBitSets.PUT_FOR_X_SITE_STATE_TRANSFER)) {
+            ctx.getCacheTransaction().setStateTransferFlag(Flag.PUT_FOR_X_SITE_STATE_TRANSFER);
+         }
+         Object result = c.acceptVisitor(ctx, entryWrappingVisitor);
+
+         if (!isSuccessfullyDone(result)) {
+            int nextIndex = i + 1;
+            if (nextIndex >= modifications.length) {
+               return makeStage(result).thenApply(ctx, command, handler);
             }
+            return makeStage(result).thenApply(ctx, command,
+                  (rCtx, rCommand, rv) -> applyModificationsAndThen(ctx, command, modifications, nextIndex, handler));
          }
       }
+      return handler.apply(ctx, command, null);
    }
 }
--- a/core/src/main/java/org/infinispan/interceptors/impl/VersionedEntryWrappingInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/VersionedEntryWrappingInterceptor.java
@@ -1,6 +1,7 @@
 package org.infinispan.interceptors.impl;
 
 import org.infinispan.commands.FlagAffectedCommand;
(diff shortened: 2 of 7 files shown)


core/src/test/java/org/infinispan/tx/EntryWrappingInterceptorDoesNotBlockTest.java
```diff
@@ -0,0 +1,310 @@
...
+      cb = new ConfigurationBuilder();
+      cb.clustering().cacheMode(CacheMode.DIST_SYNC).hash().consistentHashFactory(chFactory).numSegments(2);
+      cb.transaction().transactionMode(TransactionMode.TRANSACTIONAL);
+      createCluster(cb, 3);
+      // make sure the caches are started in proper order
+      for (int i = 0; i < 3; ++i) cache(i);
+   }
+
+   public void test1() throws Exception {
+      testMovingStable(this::readWriteKey);
```
danberindei: How about using a DataProvider, something like this:
https://paste.fedoraproject.org/paste/GwE3F5Cv3sQ9X-V1qwwZp15M1UNdIGYhyRLivL9gydE=

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
```diff
@@ -818,19 +828,37 @@ protected boolean shouldCommitDuringPrepare(PrepareCommand command, TxInvocation
...
+
+   private InvocationStage applyModifications(TxInvocationContext ctx, PrepareCommand command, WriteCommand[] modifications, int startIndex) throws Throwable {
+      // We need to execute modifications for the same key sequentially. In theory we could optimize
+      // this loop if there are multiple remote invocations but since remote invocations are rare, we omit this.
+      for (int i = startIndex; i < modifications.length; i++) {
+         WriteCommand c = modifications[i];
+         c.setTopologyId(command.getTopologyId());
+         InvocationStage visitorStage = makeStage(c.acceptVisitor(ctx, entryWrappingVisitor));
+
+         if (visitorStage.isDone()) {
```
rvansa: Okay, I don't have a problem with that. I could also add

```
boolean isResultDone(Object result) {
  return  result instanceof InvocationStage ? ((InvocationStage) result).isDone() : true;
}
```
Anyway, I'll change that to variant without instance creation.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
```diff
@@ -818,19 +828,37 @@ protected boolean shouldCommitDuringPrepare(PrepareCommand command, TxInvocation
...
             if (c.hasAnyFlag(FlagBitSets.PUT_FOR_X_SITE_STATE_TRANSFER)) {
                ctx.getCacheTransaction().setStateTransferFlag(Flag.PUT_FOR_X_SITE_STATE_TRANSFER);
             }
+         } else {
+            int nextIndex = i + 1;
+            if (nextIndex >= modifications.length) {
+               return visitorStage;
+            }
+            return makeStage(visitorStage.thenApply(ctx, command,
+                  (rCtx, rCommand, rv) -> applyModifications(ctx, command, modifications, nextIndex)));
```
rvansa: Uh, not sure if I follow - you want to refactor that  to a more general solution? Let's leave that for a follow up PR then.

core/src/main/java/org/infinispan/interceptors/totalorder/TotalOrderVersionedEntryWrappingInterceptor.java
```diff
@@ -32,6 +34,26 @@
    private static final boolean trace = log.isTraceEnabled();
    private static final EntryVersionsMap EMPTY_VERSION_MAP = new EntryVersionsMap();
 
+   private final InvocationSuccessFunction prepareHandler
+         = (ctx, command, nil) -> invokeNextThenApply(ctx, command, (rCtx, rCommand, rv) -> {
```
rvansa: According to what checkstyle?

rvansa:
@danberindei updated

rvansa:
Rebased & fixed failing tests.

core/src/test/java/org/infinispan/tx/EntryWrappingInterceptorDoesNotBlockTest.java
```diff
@@ -0,0 +1,280 @@
...
+      Traversable<Object> traversable = rwMap.evalMany(map, (BiFunction<Object, EntryView.ReadWriteEntryView<Object, Object>, Object> & Serializable) (value, view) -> {
+         assertFalse(view.find().isPresent());
+         view.set(value);
+         return "r" + index;
+      });
+      return traversable.findAny().orElseThrow(IllegalStateException::new);
+   }
+
+   private Object putDeltaWrite(MagicKey key, int index) {
+      // adding DELTA_WRITE is sufficient to force LoadType.OWNERS, we don't have to do actual delta
```
danberindei: I guess there's no class check because the parameter to a `DELTA_WRITE` put is a `DeltaAware` on the originator and a `Delta` on the remote nodes. Still, it looks weird :)

danberindei:
Got a failure in the new test:

http://ci.infinispan.org/viewLog.html?buildId=55409&buildTypeId=bt9

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
```diff
@@ -817,19 +823,44 @@ protected boolean shouldCommitDuringPrepare(PrepareCommand command, TxInvocation
...
+      // We need to execute modifications for the same key sequentially. In theory we could optimize
+      // this loop if there are multiple remote invocations but since remote invocations are rare, we omit this.
+      for (int i = startIndex; i < modifications.length; i++) {
+         WriteCommand c = modifications[i];
+         c.setTopologyId(command.getTopologyId());
+         Object result = c.acceptVisitor(ctx, entryWrappingVisitor);
+
+         InvocationStage visitorStage = null;
+         boolean done = true;
+         if (result instanceof InvocationStage) {
```
rvansa: Such method does not really fit here; I would have to do the cast in order to throw the exception in the else branch. I could use this && helper method to extract && throw, but I don't find that useful.

rvansa:
@danberindei Rebased, addressed. I've added `asyncInvokeNext(..., Collection<CF<?>> delays)` overload to `BaseAsyncInterceptor` rather than using weird method names.

Wrt the failing test, I was running that locally in a loop and didn't see any failure even without traces; I'll see if CI could reproduce that, and I'll address it with priority if it shows up to be flaky.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
danberindei: Oops, I missed both my own comments and your replies when looking at this again.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
danberindei: Not necessarily a more general solution, I just like making the state explicit so that we know what's being allocated.

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
danberindei: You don't need to extract and throw explicitly: you just need to know that the stage is not completed successfully and you need to call `makeStage(result).thenApply(...)`, which will throw instead.

danberindei:
It's failed every time so far when I ran the core test suite locally, with trace logging enabled.

https://www.dropbox.com/s/1i6k1s3qvjwevne/EntryWrappingInterceptorDoesNotBlockTest_pr_rvansa_ISPN-7578_20170427.log.gz?dl=0

rvansa:
Fixed - thanks a lot for the log, @danberindei

core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java
rvansa: OK, let's give it a go.

galderz:
Needs rebasing

rvansa:
Rebased.

core/src/main/java/org/infinispan/interceptors/impl/VersionedEntryWrappingInterceptor.java
```diff
@@ -73,11 +64,27 @@ public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand comman
          if (onePhaseCommit) {
             commitContextEntries(txInvocationContext, null, null);
          }
-         if (newVersionData != null)
-            return newVersionData;
-
-         return rv;
+         return newVersionData;
       });
+   };
```
danberindei: Extra comma

core/src/main/java/org/infinispan/interceptors/impl/VersionedEntryWrappingInterceptor.java
```diff
@@ -29,33 +32,21 @@
...
+   private final InvocationSuccessFunction prepareHandler = this::prepareHandler;
 
-   @Override
-   public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand command) throws Throwable {
-      VersionedPrepareCommand versionedPrepareCommand = (VersionedPrepareCommand) command;
-      if (ctx.isOriginLocal()) {
-         versionedPrepareCommand.setVersionsSeen(ctx.getCacheTransaction().getVersionsRead());
-      }
-      wrapEntriesForPrepare(ctx, command);
+   private Object prepareHandler(InvocationContext nonTxCtx, VisitableCommand command, Object nil) {
```
danberindei: I guess forgot to mention why I like having a method for every callback: because it allows us to keep related code together, and to minimize the diff when we extract part of a method into a callback :)

core/src/main/java/org/infinispan/interceptors/totalorder/TotalOrderVersionedEntryWrappingInterceptor.java
```diff
@@ -32,6 +35,27 @@
    private static final boolean trace = log.isTraceEnabled();
    private static final EntryVersionsMap EMPTY_VERSION_MAP = new EntryVersionsMap();
 
+   private final InvocationSuccessFunction afterPrepareHandler = (ctx, command, rv) -> {
```
danberindei: Same here, please make the callback a method after `visitPrepareCommand`.

core/src/test/java/org/infinispan/tx/EntryWrappingInterceptorDoesNotBlockTest.java
```diff
@@ -0,0 +1,281 @@
...
+      test(1, operation.f, new MagicKey("moving", cache(0), cache(1)), new MagicKey("stable", cache(0), cache(2)));
+   }
+
+   @Test(dataProvider = "operations")
+   private void testStableMoving(Operation operation) throws Exception {
+      test(1, operation.f, new MagicKey("stable", cache(0), cache(2)), new MagicKey("moving", cache(0), cache(1)));
+   }
+
+   @Test(dataProvider = "operations")
+   private void testMovingMoving(Operation operation) throws Exception {
```
danberindei: Test methods should be public.

rvansa:
@danberindei happy?

Dunno why checkstyle does not catch the extra semicolon...

tristantarrant:
Great stuff @rvansa , merged to master

PR #4959, opened 2017-03-09
ISPN-7579 Replace the non-threadsafe SimpleDateFormatter with DateTimeFormatter

tristantarrant:
https://issues.jboss.org/browse/ISPN-7579

### DIFF
--- a/server/rest/src/main/java/org/infinispan/rest/Server.java
+++ b/server/rest/src/main/java/org/infinispan/rest/Server.java
@@ -7,16 +7,15 @@ import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.io.Serializable;
 import java.io.Writer;
-import java.text.DateFormat;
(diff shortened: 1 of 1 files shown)


server/rest/src/main/java/org/infinispan/rest/Server.java
```diff
@@ -324,7 +318,7 @@ private String formatDate(Date date) {
       if (date == null)
          return null;
       else
-         return Server.DatePatternRfc1123LocaleUS.format(date);
+         return DATE_TIME_FORMATTER.format(Instant.ofEpochMilli(date.getTime()));
```
rvansa: fyi there's `date.toInstant()` method (implemented exactly as you did it).

tristantarrant: Probably a bit more readable...

PR #4960, opened 2017-03-10
ISPN-7591, ISPN-7476 and small ISPN-7517 adjustment

galderz:
All fixes in same PR since they're all related to Hibernate 2LC so I've been testing them all together.

* https://issues.jboss.org/browse/ISPN-7591
* https://issues.jboss.org/browse/ISPN-7476
* Plus a little revert on ISPN-7517 to keep `TestResourceTracker.setThreadTestName` as public.

### DIFF
--- a/core/src/main/java/org/infinispan/context/SingleKeyNonTxInvocationContext.java
+++ b/core/src/main/java/org/infinispan/context/SingleKeyNonTxInvocationContext.java
@@ -81,7 +81,7 @@ public final class SingleKeyNonTxInvocationContext implements InvocationContext 
       if (this.key == null) {
          // Set the key here
          this.key = key;
-      } else if (!this.key.equals(key)) {
+      } else if (!isKeyEquals(key)) {
          throw illegalStateException();
       }
 
@@ -94,10 +94,14 @@ public final class SingleKeyNonTxInvocationContext implements InvocationContext 
 
    @Override
    public CacheEntry lookupEntry(final Object key) {
-      if (this.key != null && this.key.equals(key))
+      if (this.key != null && isKeyEquals(key))
          return cacheEntry;
 
       return null;
+   }
+
+   public boolean isKeyEquals(Object key) {
+      return this.key == key || this.key.equals(key);
    }
 
    @Override
@@ -110,7 +114,7 @@ public final class SingleKeyNonTxInvocationContext implements InvocationContext 
       if (this.key == null) {
          // Set the key here
          this.key = key;
-      } else if (!this.key.equals(key)) {
+      } else if (!isKeyEquals(key)) {
          throw illegalStateException();
       }
 
@@ -119,7 +123,7 @@ public final class SingleKeyNonTxInvocationContext implements InvocationContext 
 
    @Override
    public void removeLookedUpEntry(final Object key) {
-      if (this.key != null && this.key.equals(key)) {
+      if (this.key != null && isKeyEquals(key)) {
          this.cacheEntry = null;
       }
    }
@@ -149,7 +153,7 @@ public final class SingleKeyNonTxInvocationContext implements InvocationContext 
 
    @Override
    public boolean hasLockedKey(final Object key) {
-      return isLocked && this.key.equals(key);
+      return isLocked && isKeyEquals(key);
    }
 
    @Override
--- a/core/src/test/java/org/infinispan/api/APINonTxTest.java
+++ b/core/src/test/java/org/infinispan/api/APINonTxTest.java
@@ -25,7 +25,6 @@ import org.infinispan.lifecycle.ComponentStatus;
 import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.test.SingleCacheManagerTest;
 import org.infinispan.test.fwk.TestCacheManagerFactory;
-import org.testng.AssertJUnit;
 import org.testng.annotations.Test;
 
 /**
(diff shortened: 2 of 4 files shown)


rvansa:
The ugly or-switch in `ExternallyMarshallable` is bloated enough - can't you convert that into static `SortedMap` with custom comparator and register `org.hibernate.*` packages in Hibernate's testsuite?

core/src/main/java/org/infinispan/context/SingleKeyNonTxInvocationContext.java
```diff
@@ -149,7 +153,7 @@ public void setClassLoader(ClassLoader classLoader) {
 
    @Override
    public boolean hasLockedKey(final Object key) {
-      return isLocked && this.key.equals(key);
+      return isLocked && isKeyEquals(key);
```
slaskawi: This should probably be `isKeyEqual` :)

galderz:
> The ugly or-switch in ExternallyMarshallable is bloated enough - can't you convert that into static SortedMap with custom comparator

Nice to have but given that this is just an assert, I don't see an urgent need to fix this right now.

> ... and register org.hibernate.* packages in Hibernate's testsuite?

The filtering capabilities cannot be modified from outside Infinispan source code, and don't intend to add it as indicated in the JIRA.

core/src/main/java/org/infinispan/context/SingleKeyNonTxInvocationContext.java
galderz: Hmmm, why? 

Are you just pointing out the choice of method name? Or something else?

galderz:
@rvansa Added a new commit to switch the white classes to a list and then use `anyMatch` to find if a class is amongst those.

rvansa:
I got that you plan to get rid of that class soon anyway, and that's why you don't like to make that class public and add static method to alter that from Hibernate testsuite?
Btw., why don't you define Externalizer for `StandardCacheEntryImpl` and register it in `LifecycleCallbacks`?

galderz:
> I got that you plan to get rid of that class soon anyway, and that's why you don't like to make that class public and add static method to alter that from Hibernate testsuite?

Hmmm, yeah, that might be a good idea. Simple yet flexible. I'll give that a go.

> Btw., why don't you define Externalizer for StandardCacheEntryImpl and register it in LifecycleCallbacks?

That's not as easy as it sounds. First, from Infinispan I can't get hold of all the instance variables. `disassembledStateText` is not externally available. Secondly, there's no public ctor to be able to take those variables and construct a `StandardCacheEntryImpl` instance. You could maybe hack something with reflection, but not really in the mood.

galderz:
Added one more commit make the `ExternallyMarshallable` public and adding a method to append to the white list.

rvansa:
@galderz That class is not set in stone, I suppose - you can't make it dependent on Infinispan, but you can expose methods.
Anyway, I think that this PR is good as it is and can be integrated.

core/src/main/java/org/infinispan/context/SingleKeyNonTxInvocationContext.java
danberindei: Personally I like `keyEquals()` better :)

core/src/test/java/org/infinispan/api/APINonTxTest.java
```diff
@@ -558,4 +559,33 @@ public void testPutIfAbsentLockCleanup() {
...
+
+   public void testIncorrectKeyGet() {
+      assertNull(cache.get(new IncorrectKey("boo", 1)));
+   }
+
+   public void testIncorrectKeyPut() {
+      cache.put(new IncorrectKey("boo", 1), "blah");
+   }
+
+   static class IncorrectKey {
```
danberindei: How about `BadEqualsKey`?

danberindei:
@galderz IMO `org.hibernate` classes shouldn't be on the blacklist at all... maybe if the blacklist was dynamic too, L2 tests could add them to the blacklist, but I really don't think it should be there by default.

And I'm not convinced about the need to support badly implemented `equals()` either :)

galderz:
> And I'm not convinced about the need to support badly implemented equals() either :)

It's not about supporting it, but about not blowing up with NPE something new with Infinispan 9. We didn't used to blow up like that before.

galderz:
> @galderz IMO org.hibernate classes shouldn't be on the blacklist at all... maybe if the blacklist was dynamic too, L2 tests could add them to the blacklist, but I really don't think it should be there by default.

I think it should since we have different parts within Infinispan itself that externalize Hibernate classes (e.g. query). So, Hibernate is a dependency just like JGroups and Arjuna are. I can't see why Hibernate should be treated differently.

galderz:
> @galderz That class is not set in stone, I suppose - you can't make it dependent on Infinispan, but you can expose methods.

In theory yeah, I could do that, but that would involve convincing Steve and that might not be easy since in his view Infinispan should be transparent in terms of the keys and values that are stored. I can see his point... but it's true that providing externalizers for it would make it more compact for sure and hence would probably give some small performance gain. Until I have some proof of how much gain this could have, I don't think it's worth the time to convince Steve.

danberindei:
> I think it should since we have different parts within Infinispan itself that externalize Hibernate classes (e.g. query). So, Hibernate is a dependency just like JGroups and Arjuna are. I can't see why Hibernate should be treated differently.

The difference is we don't depend on all of Hibernate, only on Hibernate Search, and only in the query module, so the core shouldn't know about it. Besides, we really should be transparent in terms of the keys and values that are stored :)

galderz:
> The difference is we don't depend on all of Hibernate, only on Hibernate Search, and only in the query module, so the core shouldn't know about it.

Good point, I'll remove the hibernate ref.

galderz:
Updated, I can squash some stuff when this is ready :)

galderz:
Rebased and fixed conflict. Happy with this? I can squash commits.

danberindei:
@galderz I'd like to see a more descriptive name for `IncorrectKey` and/or a link to the JIRA in the javadoc, otherwise LGTM.

galderz:
@danberindei I've renamed the key to `FalseEqualsKey` (and test methods) and sorted out the rebase.

core/src/test/java/org/infinispan/api/APINonTxTest.java
```diff
@@ -632,4 +631,33 @@ public void testReplaceAll() {
       assertEquals("hello_es:hola", cache.get("es"));
       assertEquals("hello_cz:ahoj", cache.get("cz"));
    }
+
+   public void testFalseEqualsKeyGet() {
+      assertNull(cache.get(new FalseEqualsKey("boo", 1)));
```
wburns: Hrmm maybe we should have this before and after the put? Since it will never be able to return it.

galderz:
@wburns Addressed your comment

danberindei:
LGTM. Any more comments @wburns @slaskawi ?

danberindei:
Integrated, thanks Galder!

PR #4961, opened 2017-03-10
ISPN-7525 Set topology to null if coordinator has newer topology id

slaskawi:
https://issues.jboss.org/browse/ISPN-7525

### DIFF
--- a/core/src/test/java/org/infinispan/partitionhandling/PreferConsistencyStrategyTest.java
+++ b/core/src/test/java/org/infinispan/partitionhandling/PreferConsistencyStrategyTest.java
@@ -1,17 +1,26 @@
 package org.infinispan.partitionhandling;
 
+
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Optional;
 
+import org.infinispan.commons.hash.MurmurHash3;
+import org.infinispan.distribution.TestAddress;
+import org.infinispan.distribution.ch.impl.DefaultConsistentHashFactory;
 import org.infinispan.partitionhandling.impl.PreferConsistencyStrategy;
+import org.infinispan.remoting.transport.Address;
+import org.infinispan.topology.CacheJoinInfo;
+import org.infinispan.topology.CacheStatusResponse;
+import org.infinispan.topology.CacheTopology;
 import org.infinispan.topology.ClusterCacheStatus;
 import org.infinispan.topology.ClusterTopologyManager;
-import org.infinispan.topology.ClusterTopologyManagerImpl;
+import org.infinispan.topology.PersistentUUID;
 import org.infinispan.topology.PersistentUUIDManager;
 import org.infinispan.topology.PersistentUUIDManagerImpl;
-import org.infinispan.topology.RebalancingStatus;
 import org.infinispan.util.logging.events.EventLogManager;
 import org.infinispan.util.logging.events.impl.EventLogManagerImpl;
+import org.mockito.Mockito;
 import org.testng.Assert;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/infinispan/partitionhandling/PreferConsistencyStrategyTest.java
```diff
@@ -45,5 +77,49 @@ public void testAvoidingNullPointerExceptionWhenUpdatingPartitionWithNullTopolog
...
+      // at this stage current topology ID from CacheStatus is 2
+
+      // The cache response contains non-null stable topology but null current topology
+      CacheStatusResponse responseFromOtherNode = new CacheStatusResponse(null,
+            null,
+            createCacheTopology(1, myAddress, otherAddress),
+            AvailabilityMode.AVAILABLE);
+
+      //when
+      preferConsistencyStrategy.onPartitionMerge(coordinatorStatus, Arrays.asList(responseFromOtherNode));
```
danberindei: I'm not sure what you're trying to test here... can you add some comments?

core/src/test/java/org/infinispan/partitionhandling/PreferConsistencyStrategyTest.java
```diff
@@ -22,19 +30,43 @@
...
+   private EventLogManager eventLogManager;
 
    @BeforeMethod
    public void beforeMethod() {
-      EventLogManager eventLogManager = new EventLogManagerImpl();
-      PersistentUUIDManager persistentUUIDManager = new PersistentUUIDManagerImpl();
-      ClusterTopologyManager topologyManager = new ClusterTopologyManagerImpl();
+      manager = new PersistentUUIDManagerImpl();
+      eventLogManager = new EventLogManagerImpl();
+      topologyManager = new ClusterTopologyManagerImpl() {
```
danberindei: Could use a mock instead.

core/src/test/java/org/infinispan/partitionhandling/PreferConsistencyStrategyTest.java
danberindei: TBH I'm likely the one at fault, because `PreferConsistencyStrategy` and `ClusterCacheStatus` have way too little error checking, and there's no way to know what inputs are acceptable when a cache is in a certain state. E.g. it shouldn't be possible for a cache status response to have a `null` current topology and a non-null stable topology, because a node should only send a `null` current topology if it didn't receive the join response yet.

slaskawi: I was trying to reproduce scenario mentioned in your [comment](https://github.com/infinispan/infinispan/pull/4871#discussion_r102774407): 

> I'm not sure about this change, I'm thinking maybe the coordinator already has a non-null topology in the ClusterCacheStatus and in that case, it should be replaced with a null topology.

And in this part:

> I think the idea was to create an initial topology if the cache has no "full" members yet, but there is one node who sent a join request to the old coordinator and didn't get a response with the initial cache topology before sending its status response. The joiner should be retrying to join with the new coordinator though, so I'm not sure it's worth keeping track of these joiners during the recovery process.
The case I was thinking about before is when there is a merge, but a node also left the cluster in the same view. If that node was the last member of one cache, we should make sure that cache's topology is set to null. Unfortunately you're right, we were trying to create an initial topology here too, and that probably caused a NPE sooner or later...

So in this case I'm setting current and stable topology to `null`. `CacheStatus` has newer topology (id=2) than `CacheStatusResponse` (id=1). So what you think? Is this scenario OK?

core/src/test/java/org/infinispan/partitionhandling/PreferConsistencyStrategyTest.java
slaskawi: +1, will do!

wburns:
Marked as 9.1, or does this need to go into 9.0 @slaskawi ?

core/src/test/java/org/infinispan/partitionhandling/PreferConsistencyStrategyTest.java
slaskawi: Done

slaskawi:
@wburns No, pushing it only to master is fine to me. @danberindei Would you agree with me?

tristantarrant:
Doesn't compile
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.1:testCompile (default-testCompile) on project infinispan-core: Compilation failure
[ERROR] /home/tst/Work/JBoss/infinispan/core/src/test/java/org/infinispan/partitionhandling/PreferConsistencyStrategyTest.java:[88,14] no suitable constructor found for CacheTopology(int,int,<nulltype>,<nulltype>,java.util.List<org.infinispan.remoting.transport.Address>,java.util.List<java.lang.Object>)
[ERROR] constructor org.infinispan.topology.CacheTopology.CacheTopology(int,int,org.infinispan.distribution.ch.ConsistentHash,org.infinispan.distribution.ch.ConsistentHash,org.infinispan.topology.CacheTopology.Phase,java.util.List<org.infinispan.remoting.transport.Address>,java.util.List<org.infinispan.topology.PersistentUUID>) is not applicable
[ERROR] (actual and formal argument lists differ in length)
[ERROR] constructor org.infinispan.topology.CacheTopology.CacheTopology(int,int,org.infinispan.distribution.ch.ConsistentHash,org.infinispan.distribution.ch.ConsistentHash,org.infinispan.distribution.ch.ConsistentHash,org.infinispan.topology.CacheTopology.Phase,java.util.List<org.infinispan.remoting.transport.Address>,java.util.List<org.infinispan.topology.PersistentUUID>) is not applicable
[ERROR] (actual and formal argument lists differ in length)

slaskawi:
I believe this is not valid anymore... closing.

PR #4962, opened 2017-03-10
ISPN-7592 Use correct default cache for REST tests

tristantarrant:
https://issues.jboss.org/browse/ISPN-7592
https://issues.jboss.org/browse/ISPN-7602

### DIFF
--- a/persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
+++ b/persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
@@ -3,10 +3,7 @@ package org.infinispan.persistence.rest;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
 import java.util.HashSet;
-import java.util.Locale;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletionService;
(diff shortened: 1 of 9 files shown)


tristantarrant:
Had to make some further changes. Should be good now.

tristantarrant:
The REST rolling upgrade test will fail, but that is because there is a different bug

tristantarrant:
Added a fix for the REST store issue

persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
```diff
@@ -128,6 +125,7 @@ protected void initChannel(SocketChannel ch) {
       } catch (EncoderException e) {
       }
       this.metadataHelper = Util.getInstance(configuration.metadataHelper(), ctx.getCache().getAdvancedCache().getClassLoader());
+      new HttpResponseHandler(); // Initialize the Netty TypeParameterMatcher
```
wburns: Can you add some more comments around this? This line has a really bad code smell. What happens if this line isn't here for example?

tristantarrant: I have put a log in the description of https://issues.jboss.org/browse/ISPN-7602
I know this looks disgusting, but the REST store really needs an overhaul to use pooling (ISPN-7601) which would trigger proper initialization in the right place.

persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
```diff
@@ -128,6 +125,12 @@ protected void initChannel(SocketChannel ch) {
       } catch (EncoderException e) {
       }
       this.metadataHelper = Util.getInstance(configuration.metadataHelper(), ctx.getCache().getAdvancedCache().getClassLoader());
+      /*
+       * HACK ALERT. Initialize some internal Netty structures early while we are within the scope of the correct classloader to
```
tristantarrant: @wburns better ?

wburns: Seems fine to me

wburns:
Pulling...

wburns:

PR #4966, opened 2017-03-10
ISPN-2435 Creating a ReplaceOperationStressTest

Sanne:
# Preview!

All 6 included tests fail. I'd love your help to figure out if the test is valid... I think so, in case we'll need this fixed critically in both `9` and `8.x`

@pruivo @danberindei @wburns

### DIFF
--- a/core/src/test/java/org/infinispan/stress/ReplaceOperationStressTest.java
+++ b/core/src/test/java/org/infinispan/stress/ReplaceOperationStressTest.java
@@ -1,0 +1,397 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2012 Red Hat Inc. and/or its affiliates and other
+ * contributors as indicated by the @author tags. All rights reserved.
+ * See the copyright.txt in the distribution for a full listing of
+ * individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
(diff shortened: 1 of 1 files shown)


rvansa:
Shouldn't `MOVES % THREADS` be `0`? I got the test stuck at the barrier...

Sanne:
> Shouldn't MOVES % THREADS be 0? I got the test stuck at the barrier...

Yes! thanks. Sorry, that will teach me to toy with parameters at last minute ..

rvansa:
@Sanne It seems that thread with index 0 is doing replace v_0 -> v_0 as its first operation, and it is possible that this succeeds. Other thread wit index X then replaces v_0 -> v_X and succeeds as well - voila, fail.

pruivo:
just to transcript the chat to here:

* use ```PESSIMISTIC``` locking instead of ```OPTIMISTIC``` 
* avoid same ```expected``` and ```targetValue```
* locks are released asynchronously after the operation returns (tx mode). It should do some retries instead of failing immediately.

rvansa:
@Sanne And wrt. transactional caches, you need to enable write skew check (+ RR isolation and simple versioning), otherwise this is expected behaviour. Yes, we know that not having WSC on by default sucks and we keep forgetting to change the default since 7.x or so.

Sanne:
Thanks @rvansa and @pruivo ! Great hints. I somehow forgot about those, yes we discussed it years ago but I guess I thought we were beyond the dark ages :D

Sanne:
I'll close this for now.

PR #4971, opened 2017-03-14
ISPN-7613 Enable write-skew for optimistic + repeatable-read transact…

pruivo:
…ions

* removed repeatable-read + optimistic without write-skew
* default to repeatable-read +  optimistic + write-skew

https://issues.jboss.org/browse/ISPN-7613

note: let CI run to see if I didn't miss a test.


rvansa:
Is default with pessimistic locking also RR or RC?

Btw., it would be nice if transactional putIfAbsents with implicit transactions returned false instead of throwing WSCE, but it's out of scope of this PR.

Sanne:
> Btw., it would be nice if transactional putIfAbsents with implicit transactions returned false instead of throwing WSCE, but it's out of scope of this PR.

:1st_place_medal:

pruivo:
@rvansa pessimistic locking is using RR as a default since we plan to remove RC eventually [1]
@Sanne, @rvansa I'm not sure if it is possible. the ```putIfAbsent``` contract says it must return ```null``` when succeed of the old value. But if the transaction fails to commit, we don't have to old value to return.

pruivo:
updated: fixed ```ConfigurationUnitTest``` and (hopefully) the IT tests.

Sanne:
> I'm not sure if it is possible. the putIfAbsent contract says it must return null when succeed of the old value. But if the transaction fails to commit, we don't have to old value to return.

to be fair I think we can relax that: noone is going to need the "old value" strictly speaking, what matters is that you return the most recent known value so that a subsequent CAS operation has a good chance to succeed, and it be different than null.
So you could just get() the value and return it.. it might be slighly more recent than the one which actually caused the failure, but that's ok, and possibly even a good thing if the user intends to try CAS replacements as a next step.

Sanne:
Let me add that it might be `null` and that complicates things.. but in that case you could try again and actually succeed?

pruivo:
@Sanne I like the idea of retrying the transaction. I've added to the TODO list here: https://issues.jboss.org/browse/ISPN-3927

Sanne:
@pruivo just to be clear, I didn't suggest to retry the whole transaction. Just the putIfAbsent operation could be retried - as the client expectation was "do something if this isn't there" and in practice nothing is there at commit time I think that might be valid.. But for sure needs a bit more thoughts as I didn't think it through..

danberindei:
@Sanne Repeating the `putIfAbsent` inside the transaction won't help, because of repeatable-read. We need to retry the entire transaction so we can see the new value.

Sanne:
@danberindei right but we're talking about implicit transactions. Never mind, that also means repeating "the whole transaction" like @pruivo said is the same..

danberindei:
@Sanne Right, from the user's point of view we're just retrying a `putIfAbsent` operation. I thought you meant somehow retrying only the `PutKeyValueCommand` without doing another prepare+commit round.

core/src/test/java/org/infinispan/api/AbstractRepeatableReadIsolationTest.java
```diff
@@ -240,15 +241,20 @@ private void doIsolationTest(boolean executeOnOwner, boolean initialized, Operat
...
          default:
             fail("Unknown operation " + operation);
             break;
       }
-      tm.commit();
+      try {
+         tm.commit();
+      } catch (RollbackException e) {
+         //write-skew exception
+         log.trace("Expected exception", e);
```
pruivo: optimistic transaction can also succeed.

core/src/test/java/org/infinispan/api/mvcc/repeatable_read/CacheAPIOptimisticTest.java
```diff
@@ -10,4 +10,12 @@
    protected IsolationLevel getIsolationLevel() {
       return IsolationLevel.REPEATABLE_READ;
    }
+
+   @Override
+   public void testRetainAllMethodOfEntryCollection() {
+      //pruivo.note:
+      //write-skew is not stored in ImmortalCacheEntry
+      //should we add equals() to MetadataImmortalCacheEntry and re-implement the test using it?
+      //TBH, it doesn't make much sense to expose our internal cache entries...
```
pruivo: that isn't that simple. it is invoking the ```equals()``` of our internal entries. An ```ImmortalEntry``` is not equals to a ```MetadataImmortalEntry``` or a mortal/transient entry.  I think all of ```entrySet``` code needs to be fixed to check for the ```key``` and ```value``` and not use the ```equals()```.
anyway, it isn't the scope of this PR.

core/src/test/java/org/infinispan/tx/recovery/admin/CommitFailsTest.java
```diff
@@ -89,13 +91,13 @@ public void testForceCommitTxParticipant() throws Exception {
       runTest(where);
    }
 
-   private void assertAllHaveValue(Object key, String newValue) throws Exception {
+   private void assertAllHaveValue(Object key) throws Exception {
```
pruivo: yes it is...

core/src/test/java/org/infinispan/container/versioning/LocalWriteSkewTest.java
```diff
@@ -117,9 +117,14 @@ public void testWriteSkewDisabled() throws Exception {
 
       tm().resume(t);
       cache.put("hello", "world 2");
-      tm().commit();
+      try {
+         tm().commit();
+      } catch (RollbackException e) {
+         //write-skew will be automatically enabled.
+         log.trace("Expected exception", e);
```
pruivo: actually, I'm going to remove the test.

pruivo:
updated

core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
```diff
@@ -327,6 +332,18 @@ public void validate() {
...
          throw log.unsupportedAsyncCacheMode(cacheMode);
       }
+      if (attributes.attribute(TRANSACTION_MODE).get() == TransactionMode.TRANSACTIONAL &&
+            attributes.attribute(LOCKING_MODE).get() == LockingMode.OPTIMISTIC &&
+            !attributes.attribute(USE_EAGER_LOCKING).get() && //this is pessimistic transaction
+            getBuilder().locking().isolationLevel() == IsolationLevel.REPEATABLE_READ &&
+            cacheMode != CacheMode.INVALIDATION_SYNC && //write-skew only for REPL, DIST and LOCAL (async is not supported in tx mode)
+            !getBuilder().invocationBatching().isEnabled()) { //batching can live without WSC
+         getBuilder().locking().writeSkewCheck(true);
+         getBuilder().versioning().enable().scheme(VersioningScheme.SIMPLE);
```
danberindei: I'm not sure about this, I think it will look as if it was modified by the user and it will mess with configuration inheritance. Maybe we can mark the new attribute value as "implicit" so it's not inherited? Otherwise I'd leave them disabled and do this check in a `Configurations.isWriteSkewEnabled()` method.

core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
```diff
@@ -327,6 +332,18 @@ public void validate() {
       if (attributes.attribute(TRANSACTION_MODE).get() == TransactionMode.TRANSACTIONAL && !cacheMode.isSynchronous()) {
          throw log.unsupportedAsyncCacheMode(cacheMode);
       }
+      if (attributes.attribute(TRANSACTION_MODE).get() == TransactionMode.TRANSACTIONAL &&
+            attributes.attribute(LOCKING_MODE).get() == LockingMode.OPTIMISTIC &&
+            !attributes.attribute(USE_EAGER_LOCKING).get() && //this is pessimistic transaction
+            getBuilder().locking().isolationLevel() == IsolationLevel.REPEATABLE_READ &&
+            cacheMode != CacheMode.INVALIDATION_SYNC && //write-skew only for REPL, DIST and LOCAL (async is not supported in tx mode)
```
pruivo: invalidation with tx is a weird thing. it was supported, so I keep it. Since it is invalidation mode, we can support write-skew check.

core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
```diff
@@ -327,6 +332,18 @@ public void validate() {
...
+            !getBuilder().invocationBatching().isEnabled()) { //batching can live without WSC
```
pruivo: I don't understand your question. if batching is enabled, it will use the BatchModeTransactionManager. If not, it would use another TransactionManager implementation.

core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
pruivo: what is the problem here? do you want to inherit an invalid configuration?

core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
```diff
@@ -327,6 +332,18 @@ public void validate() {
       if (attributes.attribute(TRANSACTION_MODE).get() == TransactionMode.TRANSACTIONAL && !cacheMode.isSynchronous()) {
          throw log.unsupportedAsyncCacheMode(cacheMode);
       }
+      if (attributes.attribute(TRANSACTION_MODE).get() == TransactionMode.TRANSACTIONAL &&
+            attributes.attribute(LOCKING_MODE).get() == LockingMode.OPTIMISTIC &&
+            !attributes.attribute(USE_EAGER_LOCKING).get() && //this is pessimistic transaction
```
pruivo: I've already added to the list in https://issues.jboss.org/browse/ISPN-3927

core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
rvansa: I agree with Dan here; `validate()` should be `const` method.

pruivo: @rvansa @danberindei 
I prefer to throw an exception, but that will mess up everything. Like, in 9.0.CR3 all your config is fine and in 9.0.Final, you have exceptions everywhere (cc @tristantarrant). what do you wanna do?

rvansa: As Dan suggested, simply ignoring the configuration with a warning.

pruivo: I'm confused now... if I put a warning, what happens to the part "enable write-skew by default"?

rvansa: Put a warning if the write-skew is disabled. We should set write-skew as `true` by default in `LockingConfiguration`. Then, we should put a warning if the locking mode was set to pessimistic, but write skew was not disabled.
The code should ignore the actual value of write-skew in configuration.

pruivo: I would prefer another approach:
* deprecate write-skew (remove the attributed, but leave the methods)
* set the default versioning to SIMPLE (deprecate as well? I don't see any use case for this)

when we have optimistic + repeatable_read, we know we need versions and write-skew checks.

is everybody ok with it?

rvansa: @pruivo we need to keep the attribute as well, it's public, exported to XSD etc. But I am OK with making changes to that no-op.
The same for `VersioningScheme`, just make sure, that we don't use VersionedRepeatableReadEntry with pessimistic locking.

pruivo:
2 news commits:
* deprecated write-skew and versioning. 
* update the upgrading.doc

I'll squash the commits before it gets integrated!

documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
```diff
@@ -9,6 +9,18 @@ This guide walks you through the process of upgrading Infinispan.
...
+
+The default configuration for transactional caches changed from `READ_COMMITTED` and `OPTIMISTIC` locking to `REPEATABLE_READ` and `OPTIMISTIC` locking with `write-skew` enabled.
+
+Also, using the `REPEATABLE_READ` isolation level and `OPTIMISTIC` locking without `write-skew` enabled is no longer allowed.
+To help with the upgrade, `write-skew` will be automatically enabled in this case.
+
+The following configuration has been deprecated:
+
+* `write-skew`: as said, it is automatically enabled.
+* `<versioning>` and its attributed. It is automatically enabled and configured when needed.
```
rvansa: just a typo: attribute**d**

core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
danberindei: AFAICT `BatchModeTransactionManager` is only used if the user didn't configure another TM with `transactionManagerLookup(GenericTransactionManagerLookup.INSTANCE)`. So in theory someone could use batching with the Arjuna transaction manager.

Off-topic, I see there's a setter in `TransactionConfiguration`, but I have a feeling it doesn't really work.

core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
danberindei: I'm ok with it.

core/src/main/java/org/infinispan/configuration/cache/TransactionConfigurationBuilder.java
pruivo: yep I've noticed and removed this change while ago.
I deprecated the ```TransactionConfiguration.transactionManagerLookup()``` setter.

pruivo:
updated once more :)

core/src/test/java/org/infinispan/distribution/groups/WriteSkewGetGroupKeysTest.java
```diff
@@ -27,12 +25,12 @@
...
-         new WriteSkewGetGroupKeysTest(TestCacheFactory.BACKUP_OWNER).totalOrder(false),
-         new WriteSkewGetGroupKeysTest(TestCacheFactory.BACKUP_OWNER).totalOrder(true),
-         new WriteSkewGetGroupKeysTest(TestCacheFactory.NON_OWNER).totalOrder(false),
-         new WriteSkewGetGroupKeysTest(TestCacheFactory.NON_OWNER).totalOrder(true),
+         new WriteSkewGetGroupKeysTest(TestCacheFactory.PRIMARY_OWNER).totalOrder(false).isolationLevel(IsolationLevel.REPEATABLE_READ),
+         new WriteSkewGetGroupKeysTest(TestCacheFactory.PRIMARY_OWNER).totalOrder(true).isolationLevel(IsolationLevel.REPEATABLE_READ),
+         new WriteSkewGetGroupKeysTest(TestCacheFactory.BACKUP_OWNER).totalOrder(false).isolationLevel(IsolationLevel.REPEATABLE_READ),
+         new WriteSkewGetGroupKeysTest(TestCacheFactory.BACKUP_OWNER).totalOrder(true).isolationLevel(IsolationLevel.REPEATABLE_READ),
+         new WriteSkewGetGroupKeysTest(TestCacheFactory.NON_OWNER).totalOrder(false).isolationLevel(IsolationLevel.REPEATABLE_READ),
+         new WriteSkewGetGroupKeysTest(TestCacheFactory.NON_OWNER).totalOrder(true).isolationLevel(IsolationLevel.REPEATABLE_READ),
```
danberindei: Wasn't this already set in the constructor?

pruivo: I didn't noticed the constructor :)

pruivo:
one more...

pruivo:
rebased!

pruivo:
rebased

core/src/main/java/org/infinispan/configuration/cache/TransactionConfiguration.java
```diff
@@ -186,6 +186,10 @@ public TransactionManagerLookup transactionManagerLookup() {
       return transactionManagerLookup.get();
    }
 
+   /**
+    * @deprecated since 9.0. it doesn't really works since it fetches the transaction manager when starting.
```
danberindei: Rewording: the transaction manager is only looked up once, during cache startup.

core/src/main/java/org/infinispan/configuration/cache/ConfigurationChildBuilder.java
```diff
@@ -44,6 +44,10 @@
 
    TransactionConfigurationBuilder transaction();
 
+   /**
+    * @deprecated since 9.0. Infinispan automatically enable versioning when needed.
```
danberindei: enable*s* here too.

pruivo:
updated! also:
* removed unused imports
* replace expressions by lambdas where possible

danberindei:
Integrated, thanks Pedro!

PR #4975, opened 2017-03-15
ISPN-7619 Use dynamic timeout values in the PersistenceManagerCloseableSupplier

gustavonalle:
https://issues.jboss.org/browse/ISPN-7619

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
@@ -496,6 +496,13 @@ public class CacheLoaderInterceptor<K, V> extends JmxStatsCommandInterceptor {
          this.entrySet = entrySet;
       }
 
+      long calculateTimeoutSeconds() {
+         int minimum = 10;
+         int size = dataContainer.sizeIncludingExpired();
+         if (size < 10_000) return minimum;
+         return Math.round(Math.log1p(size) * 10);
+      }
+
       @Override
       public CloseableIterator<CacheEntry<K, V>> iterator() {
          CloseableIterator<CacheEntry<K, V>> iterator = Closeables.iterator(entrySet.stream());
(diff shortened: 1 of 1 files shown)


danberindei:
@gustavonalle @wburns  I think the timeout is just to retrieve a single entry, I don't think that should depend on the store size.

But anyway, cache streams already have a user-configurable timeout, can't we use that instead of hard-coding a value in `CacheLoaderInterceptor`?

gustavonalle:
@danberindei this timeout is involved in a supplier that obtains data from a cache store, I did not know about any other timeout in the stream, will let @wburns comment on that

Sanne:
+1 to default into some kind of reasonable guess..

Yet in my experience people _really_ prefer to set the desired timeout explicitly on public API of the client. Is that being planned too?

danberindei:
@gustavonalle No idea why 10 seconds is not enough in your case, but `PersistenceManagerCloseableSupplier` definitely calls `closeCondition.await(targetTime - System.nanoTime(), TimeUnit.NANOSECONDS)` for each entry. If the store needs to read all the entries before processing the first one, that looks like a problem in the store...

@Sanne The users can already set a timeout for streams with `cache.entrySet().stream().timeout(time, unit)`, but it's only used for RPCs. We can either use that (at least when the iterator is being created internally by a stream), or we could change the code so that the supplier never times out.

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
```diff
@@ -496,6 +496,13 @@ public WrappedEntrySet(EntrySetCommand command, CacheSet<CacheEntry<K, V>> entry
          this.entrySet = entrySet;
       }
 
+      long calculateTimeoutSeconds() {
+         int minimum = 10;
+         int size = dataContainer.size();
```
wburns: We should call sizeExcludingExpired

wburns:
@danberindei So this is an interesting dilemma. The reason why 10s doesn't work is because of how `DistinctKeyDoubleEntryCloseableIterator` works. It basically exhausts the first iterator completely and for each entry/key it retrieves it keeps a reference to the key. Then it starts the next iterator and only processes an entry if the key doesn't match that entry. So the issue becomes if in memory and the store both contain the same exact data. In that case you will have to iterate over the entire store and find no entries. But depending on the store implementation and number of entries this could take a very long time.

So I suggested @gustavonalle to increase this timeout dynamically to be larger based on the amount of in memory entries to allow the store more time to actually try to find a value.

wburns:
This also brings to light that if passivation is not enabled, we should just iterate over the store and ignore in memory iterator.

gustavonalle:
@wburns fixed

wburns:
Pulling... we will reevaluate usage for 9.1

PR #4983, opened 2017-03-15
ISPN-7586 Rolling Upgrade: use of Remote Store in mode read-only causes data inconsistencies

gustavonalle:
…es data inconsistencies

https://issues.jboss.org/browse/ISPN-7586

### Note for reviewers:

This changes a bit the Hot Rod rolling upgrade procedure. As described in the JIRA, it requires the `RemoteStore` to be in **read-write mode rather than read only**.

To enable that, and to prevent that the Rolling Upgrade itself triggers a write back to the source cluster via the remote store, it will use a new flag `ROLLING_UPGRADE` that will be handled internally in order to avoid such writes.


gustavonalle:
Updated with a bit more testing

gustavonalle:
Closing for now, will need to test another scenario...

gustavonalle:
Updated

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
```diff
@@ -37,46 +51,97 @@
...
    private final int readBatch;
    private final int threads;
-   private volatile Cache<Object, Object> cache;
+   private final ConcurrentHashSet<WrappedByteArray> deletedKeys = new ConcurrentHashSet<>();
+   private byte[] ignoredKey;
+
+   private transient Set<RemoteStore> stores;
+   private transient Cache<Object, Object> cache;
+   private transient ExecutorService executorService;
+   private transient RemoveListener listener = new RemoveListener();
```
wburns: Seems this can be final as well.

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
```diff
@@ -37,46 +51,97 @@
...
 
    @Override
    public void setEnvironment(Cache<Object, Object> cache, Set<Object> inputKeys) {
+      ComponentRegistry cr = cache.getAdvancedCache().getComponentRegistry();
+      PersistenceManager loaderManager = cr.getComponent(PersistenceManager.class);
+      this.stores = loaderManager.getStores(RemoteStore.class);
+      Marshaller marshaller = new MigrationMarshaller();
       this.cache = cache;
+      this.cache.addListener(listener);
+      this.executorService = Executors.newFixedThreadPool(threads);
```
wburns: We should name the threads.

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
```diff
@@ -37,46 +51,97 @@
...
+         if (storeIterator.hasNext()) {
+            RemoteStore store = storeIterator.next();
+            RemoteCache<Object, Object> storeCache = store.getRemoteCache();
+            RemoteStoreConfiguration storeConfig = store.getConfiguration();
+            if (!storeConfig.hotRodWrapping()) {
+               throw log.remoteStoreNoHotRodWrapping(cache.getName());
+            }
+            AtomicInteger counter = new AtomicInteger(0);
+            migrateEntriesWithMetadata(storeCache,counter);
+            gracefulShutdown(executorService);
```
wburns: Put it in finally block?

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
```diff
@@ -37,46 +51,97 @@
...
-      ComponentRegistry cr = cache.getAdvancedCache().getComponentRegistry();
-      PersistenceManager loaderManager = cr.getComponent(PersistenceManager.class);
-      Set<RemoteStore> stores = loaderManager.getStores(RemoteStore.class);
-      Marshaller marshaller = new MigrationMarshaller();
-      byte[] knownKeys;
       try {
-         knownKeys = marshaller.objectToByteBuffer(MIGRATION_MANAGER_HOT_ROD_KNOWN_KEYS);
-      } catch (Exception e) {
-         throw new CacheException(e);
+         Iterator<RemoteStore> storeIterator = stores.iterator();
```
wburns: Any reason this doesn't use forEach loop?

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
```diff
@@ -37,46 +51,97 @@
...
+               long version = metadataValue.getVersion();
+               long created = metadataValue.getCreated();
+               long lastUsed = metadataValue.getLastUsed();
+               Metadata metadata = new EmbeddedMetadata.Builder()
+                     .version(new NumericVersion(version))
+                     .lifespan(lifespan, TimeUnit.SECONDS)
+                     .maxIdle(maxIdle, TimeUnit.SECONDS)
+                     .build();
+               InternalMetadataImpl internalMetadata = new InternalMetadataImpl(metadata, created, lastUsed);
+               executorService.submit(() -> {
```
wburns: Wonder if this is faster by doing batches of inserts instead of 1 callable per.

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
```diff
@@ -37,46 +51,97 @@
...
       this.readBatch = readBatch;
       this.threads = threads;
    }
 
+   @Listener(clustered = true)
+   @SuppressWarnings("unused")
+   private class RemoveListener {
+      @CacheEntryRemoved
+      public void entryRemoved(CacheEntryRemovedEvent event) {
+         deletedKeys.add(new WrappedByteArray((byte[]) event.getKey()));
```
wburns: Is `WrappedByteArray` just used to test equality?

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
```diff
@@ -37,46 +51,97 @@
...
    }
 
    @Override
    public void setEnvironment(Cache<Object, Object> cache, Set<Object> inputKeys) {
+      ComponentRegistry cr = cache.getAdvancedCache().getComponentRegistry();
+      PersistenceManager loaderManager = cr.getComponent(PersistenceManager.class);
+      this.stores = loaderManager.getStores(RemoteStore.class);
+      Marshaller marshaller = new MigrationMarshaller();
       this.cache = cache;
+      this.cache.addListener(listener);
```
wburns: Hrmm wouldn't the listener be null on a remote node?

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
gustavonalle: yes

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
gustavonalle: The migration task runs on each node of the target cluster. Each individual task will read a subset of segments from the source cluster and insert into. So the listener should be on each node.

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
gustavonalle: In theory yes, but I am a bit unsure if having a cache with more than 1 remote store, each one pointing to a different source cluster potentially, being part of a N-1 migration is something useful...

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
gustavonalle: Good question, let me check that.

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
wburns: Ah I see now. In that case you should call `addFilteredListener` instead of normal `addListener`. This way it should reduce a lot of the cluster communications.

gustavonalle: 👍

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
gustavonalle: I need to wait until it is done before returning a value

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
gustavonalle: ok

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
wburns: You need to wait for quiescence, gotcha. In that case I say we should at least put a shutdownNow call as well in the finally block, just in case if we did get an exception.

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/HotRodTargetMigrator.java
```diff
@@ -109,42 +106,34 @@ public long synchronizeData(Cache<Object, Object> cache, int readBatch, int thre
...
-               for (List<Integer> partition : partitions) {
-                  Set<Integer> segmentSet = new HashSet<>();
-                  segmentSet.addAll(partition);
-                  DistributedTask<Integer> task = executor
+            CacheTopologyInfo sourceCacheTopologyInfo = remoteSourceCache.getCacheTopologyInfo();
+            int sourceSegments = sourceCacheTopologyInfo.getNumSegments();
+            List<Address> targetServers = cache.getAdvancedCache().getDistributionManager().getWriteConsistentHash().getMembers();
+
+            List<List<Integer>> partitions = split(range(sourceSegments), targetServers.size());
+            DistributedExecutorService executor = new DefaultExecutorService(cache);
```
wburns: We need to shut this down after we are done with it, else a thread can be leaked.

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
gustavonalle: I couldn't verify any difference when running locally. Granted, CPU quickly saturates by running two clusters in the same JVM. I will do a test in Openstack

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
gustavonalle: I changed it to require at most one remote store in the destination cache.

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/MigrationTask.java
gustavonalle: No difference noted, plus this would add another knob that would interfere with the existing ones at https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/upgrade/TargetMigrator.java#L25

gustavonalle:
@wburns updated.

gustavonalle:
Fixed test failures

tristantarrant:
This looks solid to me. @wburns any more notes ?

gustavonalle:
I've just rebased, moved it to the top of the CI queue

gustavonalle:
The test failure are not related

tristantarrant:
Merged, thanks

PR #4984, opened 2017-03-15
ISPN-7627 Add test to verify off heap and binary storage work with

wburns:
indexing

* Added BINARY and OFF_HEAP indexed tests
* Fixed issue with putAll with BINARY and OFF_HEAP

https://issues.jboss.org/browse/ISPN-7627


gustavonalle:
@wburns There are failing query tests

wburns:
Looking..

wburns:
Yeah from the command line it runs the tests in parallel and causes issues, have to look into how to tweak it.  Looks like the Factory annotation isn't working as I would have hoped :-1:

wburns:
Removed `Factory` annotation as it shouldn't be needed for `ClusteredCacheTest`

wburns:
Tests pass now http://ci.infinispan.org/viewLog.html?buildId=52617&buildTypeId=bt9

wburns:
Let me double check the usage of factory with the subclasses.

wburns:
Okay it is fine since the class doesn't have the factory method, the subclass does :) [1] So this is all good.

[1] https://github.com/infinispan/infinispan/blob/master/core/src/test/java/org/infinispan/test/MultipleCacheManagersTest.java#L437

gustavonalle:
Merged, thanks @wburns

PR #4985, opened 2017-03-15
ISPN-7614 Disabling a store doesn't stop it

danberindei:
https://issues.jboss.org/browse/ISPN-7614

* Check that the store is stopped in ClassLoaderManagerDisablingTest
* Disable reading from/reading to a stopped DummyInMemoryStore
* Make DummyInMemoryStore state private and add accessors


ryanemerson:
@danberindei Integrated, thanks

PR #4989, opened 2017-03-17
ISPN-7636 Update default HR version and fix HR decoder

vjuranek:
https://issues.jboss.org/browse/ISPN-7636

* Use HR 2.6 as default protocol version
* Fix HR decoder to read add listener command correctly

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/ProtocolVersion.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/ProtocolVersion.java
@@ -27,7 +27,7 @@ public enum ProtocolVersion {
    PROTOCOL_VERSION_10(1, 0),
    ;
 
-   public static final ProtocolVersion DEFAULT_PROTOCOL_VERSION = PROTOCOL_VERSION_25;
+   public static final ProtocolVersion DEFAULT_PROTOCOL_VERSION = PROTOCOL_VERSION_26;
 
    private final String version;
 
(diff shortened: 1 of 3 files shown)


gustavonalle:
There some test failures in the Hot Rod client suite:
org.junit.ComparisonFailure: expected:<...protocol version: 2.[5]> but was:<...protocol version: 2.[6]>

wburns:
It sounds like we need a test that verifies the add client listener behavior in the decoder. As long as this test is added to a test that is setup using `HotRodTestingUtils` to configure the server it will find if there is any issue.

vjuranek:
I'm on PTO, will look on dedicated next week (failing test is fixed).

wburns:
Actually looking closer as you mentioned on ISPN-7636 the ClusterInvalidatedNearCacheTest uses the proper test utils. But for some reason the test actually passes fine for me without hanging. I just would love for a test to hang to make sure we plugged all the holes properly when the fix is applied.

wburns:
Okay I see the issue [1] added additional lines to the code but didn't add the appropriate read from the context. It used to not store to the context for the converter information because it was the last one and just returned. With the new protocol version it became the last bytes on the stream for add listener if it is greater than version 2.5.

So I am able to get it to hang with version 2.6 change as well.

[1] https://github.com/vjuranek/infinispan/commit/5af2fd53fae1b3f159b4a310b347ddafc5eca8f7#diff-a4a6292d50595775143111292bcf73afR413

wburns:
Pulling...

wburns:
Integrated into master, thanks @vjuranek !

PR #4992, opened 2017-03-20
ISPN-7632 Remove useEagerLocking configuration

danberindei:
https://issues.jboss.org/browse/ISPN-7632


pruivo:
isn't kind of bad changing public API just before final?

danberindei:
@pruivo I'll let @tristantarrant decide, but I figure this has stayed deprecated long enough that it's not worth maintaining it in all 9.x minor versions.

tristantarrant:
I think we'll do a CR4 at this point. +1 on this removal

pruivo:

PR #4999, opened 2017-03-20
ISPN-7572 Correct the testing configuration for index storage in H2

Sanne:
A follow up on the already-merged ISPN-7572 to fix the regression spotted by the testsuite.

The important bit is setting the caches to `non-transactional` as otherwise the concurrent writes on H2 end up in transaction locks, caused e.g. by the Lucene merger thread and our async deletions of index data.

While at it, I also spotted two configuration inconsistencies.

### DIFF
--- a/integrationtests/as-lucene-directory/src/test/resources/user-provided-infinispan-persistence.xml
+++ b/integrationtests/as-lucene-directory/src/test/resources/user-provided-infinispan-persistence.xml
@@ -10,8 +10,12 @@
         <jmx duplicate-domains="true"/>
 
         <replicated-cache name="LuceneIndexesMetadata" mode="SYNC" remote-timeout="25000">
+            <transaction mode="NONE" />
             <persistence passivation="false">
-                <string-keyed-jdbc-store preload="true" key-to-string-mapper="org.infinispan.lucene.LuceneKey2StringMapper" shared="true">
+                <!-- N.B. "shared" is false which is unusual for a JDBC CacheStore,
+                 but for testing purposes we use an in-memory H2 so each WildFly node
+                 will have an independent CacheStore instance -->
+                <string-keyed-jdbc-store preload="true" key-to-string-mapper="org.infinispan.lucene.LuceneKey2StringMapper" shared="false">
                     <string-keyed-table prefix="IndexMeta">
                         <id-column name="K" type="VARCHAR(255)"/>
                         <data-column name="V" type="BLOB"/>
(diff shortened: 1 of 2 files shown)


integrationtests/as-lucene-directory/src/test/resources/user-provided-infinispan-persistence.xml
```diff
@@ -10,8 +10,12 @@
         <jmx duplicate-domains="true"/>
 
         <replicated-cache name="LuceneIndexesMetadata" mode="SYNC" remote-timeout="25000">
+            <transaction mode="NONE" />
             <persistence passivation="false">
-                <string-keyed-jdbc-store preload="true" key-to-string-mapper="org.infinispan.lucene.LuceneKey2StringMapper" shared="true">
+                <!-- N.B. "shared" is false which is unusual for a JDBC CacheStore,
+                 but for testing purposes we use an in-memory H2 so each WildFly node
+                 will have an independent CacheStore instance -->
+                <string-keyed-jdbc-store preload="true" key-to-string-mapper="org.infinispan.lucene.LuceneKey2StringMapper" shared="false">
```
gustavonalle: Can't the test use a shared H2 database so that it's more aligned with real use cases?

Sanne: It used to be like that, with an H2 configuration to store the database in a file on ~.

But then we had sporadic failures caused by H2 non correctly wiping the database.. 

I think this is also totally valid: we're testing both a non-shared  cacheStore (people might use one?) and we're testing that we're able to configure a JDBC CacheStore, which was the main issue I wanted to resolve as that was broken.

I agree it's unlikely for someone to *combine* the JDBC CacheStore with a non-shared-store, but it's just a test which should pass. The comment addresses that.

gustavonalle: We could've used an in-memory H2 database accessible from each test server via TCP
http://www.h2database.com/html/features.html#in_memory_databases

But anyway, I'll integrate this PR to have the tests fixed.

gustavonalle:
integrated, thanks @Sanne !

integrationtests/as-lucene-directory/src/test/resources/user-provided-infinispan-persistence.xml
Sanne: right, I actually do that in some other module but it requires a bit more coding to start only one of these. I don't remember which other module does it ;)

PR #5000, opened 2017-03-21
ISPN-7002 Early detection of improper query module usage in wildfly

anistor:
Add early detection of the case when the query jar is added just as a pom dependency but not as a proper wildfly module dependency. This allows the application classloading to apparently work but the query module never gets initialised and leads to further failures.

PLEASE do not close the Jira yet.
https://issues.jboss.org/browse/ISPN-7002


danberindei:
Integrated in master and 8.2.x, thanks Adrian!

PR #5002, opened 2017-03-21
ISPN-6976 Log#iterationClosed should not use INFO level logging

Sanne:
- https://issues.jboss.org/browse/ISPN-6976


danberindei:
Integrated, thanks Sanne!

PR #5003, opened 2017-03-21
ISPN-7650 Refactoring two functional test assertions

karesti:
https://issues.jboss.org/browse/ISPN-7650

### DIFF
--- a/core/src/test/java/org/infinispan/api/APINonTxTest.java
+++ b/core/src/test/java/org/infinispan/api/APINonTxTest.java
@@ -1,8 +1,11 @@
 package org.infinispan.api;
 
+import static org.infinispan.test.Exceptions.expectException;
 import static org.infinispan.test.TestingUtil.assertNoLocks;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNull;
+import static org.testng.Assert.assertTrue;
 import static org.testng.AssertJUnit.assertEquals;
-import static org.testng.AssertJUnit.fail;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -25,7 +28,7 @@ import org.testng.annotations.Test;
  * @author Mircea Markus
  * @since 5.1
  */
-@Test (groups = "functional", testName = "api.APINonTxTest")
+@Test(groups = "functional", testName = "api.APINonTxTest")
 public class APINonTxTest extends SingleCacheManagerTest {
 
    @Override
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/infinispan/api/APINonTxTest.java
```diff
@@ -151,19 +143,28 @@ public void testUnsupportedKeyValueCollectionOperations() {
...
-         } catch (ClassCastException e) {
-            // Ignore class cast in expired filtered set because
-            // you cannot really add an Object type instance.
-         }
-         try {
-            col.addAll(newObjCol);
-            assert false : "Should have thrown a UnsupportedOperationException";
-         } catch (UnsupportedOperationException uoe) {
-         }
+         col.add(newObj);
```
rvansa: I think that you want to test adding to both keys and values - this will throw exception with keys and the values won't be tested.
IMO having `Collection[] { keys, values }` is a strange pattern; I would rather use `keys.add(newObj);` and `values.add(newObj)` separately.
Note that instead of try-catch or `@Test(expectedExceptions = ...)` you can also use `Exceptions.expectException(...)`.

core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
```diff
@@ -176,20 +182,21 @@ public void testConditionalReplaceFromNonOwner() {
       initAndTest();
       Cache<Object, String> nonOwner = getFirstNonOwner("k1");
       boolean retval = nonOwner.replace("k1", "valueX", "value2");
-      if (testRetVals) assert !retval : "Should not have replaced";
+      if (testRetVals) assertFalse(retval, "Should not have replaced");
 
       assertOnAllCachesAndOwnership("k1", "value");
 
-      assert !nonOwner.getAdvancedCache().getComponentRegistry().getComponent(DistributionManager.class).getCacheTopology().isWriteOwner("k1");
+      assertFalse(nonOwner.getAdvancedCache().getComponentRegistry().getComponent(DistributionManager.class).getCacheTopology().isWriteOwner("k1"));
```
rvansa: There's a helper method `TestingUtil.extractComponent`

core/src/test/java/org/infinispan/api/APINonTxTest.java
karesti: Ok 👍

karesti:
@rvansa I modified both tests with your comments

ryanemerson:
Integrated, thanks @karesti

rvansa:
Oops, I have noticed only now that it uses TestNG assertions; eventually all codebase should use only the AssertJUnit ones (despite the annotations are TestNG). My fault, I should have pointed it out during the review earlier; it's not an issue, but @karesti in the future please use AssertJUnit.

karesti:
Yes, indeed I realised it just after too. No pb, I will refactor them on my branches to import the good ones if nobody does it before

PR #5004, opened 2017-03-21
Ispn 7651 Add trace message when listener is invoked

wburns:
https://issues.jboss.org/browse/ISPN-7651

### DIFF
--- a/core/src/main/java/org/infinispan/notifications/impl/AbstractListenerImpl.java
+++ b/core/src/main/java/org/infinispan/notifications/impl/AbstractListenerImpl.java
@@ -357,61 +357,54 @@ public abstract class AbstractListenerImpl<T, L extends ListenerInvocation<T>> {
          this.target = target;
          this.method = method;
          this.sync = sync;
-         this.classLoader = new WeakReference<ClassLoader>(classLoader);
+         this.classLoader = new WeakReference<>(classLoader);
          this.subject = subject;
       }
 
       @Override
       public void invoke(final A event) {
-         Runnable r = new Runnable() {
-
-            @Override
-            public void run() {
-               ClassLoader contextClassLoader = null;
-               Transaction transaction = suspendIfNeeded();
-               if (classLoader.get() != null) {
-                  contextClassLoader = SecurityActions.setContextClassLoader(classLoader.get());
-               }
-               try {
-                  if (subject != null) {
-                     try {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/notifications/impl/AbstractListenerImpl.java
```diff
@@ -357,61 +357,61 @@ public ListenerInvocationImpl(Object target, Method method, boolean sync, ClassL
...
+            if (classLoader.get() != null) {
+               contextClassLoader = SecurityActions.setContextClassLoader(classLoader.get());
+            }
+            try {
+               if (subject != null) {
+                  try {
+                     Security.doAs(subject, (PrivilegedExceptionAction<Void>) () -> {
+                        if (trace) {
+                           // Don't want to print out Subject as it could have sensitive information
+                           log.tracef("Invoking listener: %s passing event %s using subject", target, event);
```
ryanemerson: Shouldn't we be using `getLog` here and on L395?

wburns: Good catch. I would say it is better to use getLog and I can remove all of the prior log references an update them as well.

wburns:
Updated all references to use `getLog` so log messages show from the appropriate implementation.

core/src/main/java/org/infinispan/notifications/impl/AbstractListenerImpl.java
danberindei: I find it pretty annoying when I search for a log message in a class and I can't find it because it was actually logged by a base class, so I'm not a big fan of the abstract `getLog()` strategy, but I guess if it's already there it's best to use the same approach everywhere.

core/src/main/java/org/infinispan/notifications/impl/AbstractListenerImpl.java
```diff
@@ -47,9 +47,6 @@
  */
 public abstract class AbstractListenerImpl<T, L extends ListenerInvocation<T>> {
 
-   private static final Log log = LogFactory.getLog(AbstractListenerImpl.class);
```
danberindei: Need to remove the import as well.

wburns: Removed

danberindei:
Integrated, thanks Will!

PR #5006, opened 2017-03-22
ISPN-7653 Add feature to inject input keys

SetoKaiba:
https://issues.jboss.org/browse/ISPN-7653

### DIFF
--- a/core/src/main/java/org/infinispan/distexec/spi/DefaultDistributedTaskLifecycle.java
+++ b/core/src/main/java/org/infinispan/distexec/spi/DefaultDistributedTaskLifecycle.java
@@ -1,5 +1,6 @@
 package org.infinispan.distexec.spi;
 
+import java.util.Collection;
 import java.util.concurrent.Callable;
 
 import org.infinispan.Cache;
@@ -9,7 +10,7 @@ import org.kohsuke.MetaInfServices;
 public class DefaultDistributedTaskLifecycle implements DistributedTaskLifecycle {
 
    @Override
-   public <T, K, V> void onPreExecute(Callable<T> task, Cache<K, V> inputCache) {
+   public <T, K, V> void onPreExecute(Callable<T> task, Cache<K, V> inputCache, Collection<K> inputKeys) {
       // intentionally no-op
    }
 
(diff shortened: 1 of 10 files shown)


gustavonalle:
ok to test

wburns:
Hrmm can you not use [1] instead?

[1] https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/distexec/DistributedCallable.java

SetoKaiba:
@wburns Of course. But it's indeed the missing feature for CDI. You don't need to inject inputCache either. You can use setEnvironment.

SetoKaiba:
The problem CDI solving is to simplify the way to get components.

SetoKaiba:
I aware this missing feature because of implementing PartiallyReindex feature and seeing setEnvironment.

core/src/main/java/org/infinispan/distexec/spi/DefaultDistributedTaskLifecycle.java
```diff
@@ -9,7 +9,7 @@
 public class DefaultDistributedTaskLifecycle implements DistributedTaskLifecycle {
 
    @Override
-   public <T, K, V> void onPreExecute(Callable<T> task, Cache<K, V> inputCache) {
+   public <T, K, V> void onPreExecute(Callable<T> task, Cache<K, V> inputCache, K... inputKeys) {
```
slaskawi: In general the code looks fine to me. 

The only comment that I have is that inputKeys should be a list (instead of array). This would simplify (a little bit but still) the code in `DefaultExecutorService`. Moreover, you should pass `Collections.unmodifiableList()` there.

Good work @SetoKaiba !

SetoKaiba: No. It‘s the same with List. Because the inputKeys in DistributedTaskPart is List<Object> and the keys in DistributedExecuteCommand is Set<Object>. It uses different collection container. No differences from array.
Instead, I used Collections.unmodifiableCollection() now.

SetoKaiba:
It's strange that it uses List in DistributedTaskPart and Set in DistributedExecuteCommand for keys. Anyway, uses collection would be OK for now.

slaskawi:
Looking

slaskawi:
Thanks @SetoKaiba !

PR #5010, opened 2017-03-22
Move multiple top-level classes in a single RemoteCacheManager.java s…

rhusar:
…ource file as static inner classes

Spring cleaning! Looks like an oversight.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
@@ -323,43 +323,42 @@ public class RemoteCacheManager implements RemoteCacheContainer {
       return HotRodConstants.DEFAULT_CACHE_NAME_BYTES;
    }
 
-}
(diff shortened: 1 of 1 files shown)


danberindei:
As long as they don't need to be public, top-level seems fine to me...

rvansa:
@danberindei I think that placing multiple top-level classes into one file is a discouraged practice.

rhusar:
@rvansa is right. This practice is called "mess" officially. :-)

danberindei:
@rhusar That's just Jon Skeet, give me a more credible reference :P

rhusar:
@danberindei  ROFL I see you found my sources :)  Anyway, the way I stumbled upon this is that in IntelliJ IDEA (2017.1) double-clicking on the class file does not open the file itself but rather opens another menu listing out the top level classes defined in that file (what a PITA).

danberindei:
@rhusar That sounds annoying indeed... Ok, just make the classes private, and I'll integrate it.

rhusar:
Oops, fixed.

danberindei:
Integrated, thanks Rado!

PR #5011, opened 2017-03-22
Ispn 7656 DefaultDataContainer.entrySet().stream().toArray(Object[]::new) may fail

wburns:
https://issues.jboss.org/browse/ISPN-7656

### DIFF
--- a/core/src/main/java/org/infinispan/container/DataContainer.java
+++ b/core/src/main/java/org/infinispan/container/DataContainer.java
@@ -99,12 +99,16 @@ public interface DataContainer<K, V> extends Iterable<InternalCacheEntry<K, V>> 
     * Returns a set of keys in the container. When iterating through the container using this method, clients should
     * never call {@link #get(Object)} method but instead {@link #peek(Object)}, in order to avoid changing the order of
     * the underlying collection as a side of effect of iterating through it.
-    *
+    * <p>
+    * This set of keys will include expired entries. If you wish to only retrieve non expired keys please use the
+    * {@link DataContainer#entrySet()} method and retrieve keys from there.
     * @return a set of keys
     */
    Set<K> keySet();
 
    /**
+    * This returns all values in the container including expired entries. If you wish to only receive values that
+    * are not expired it is recommended to use {@link DataContainer#entrySet()} and pull values from there directly.
     * @return a set of values contained in the container
     */
    Collection<V> values();
@@ -116,7 +120,10 @@ public interface DataContainer<K, V> extends Iterable<InternalCacheEntry<K, V>> 
     * <p/>
     * If a client needs to iterate through a mutable set of mutable cache entries, it should iterate the container
     * itself rather than iterating through the return of entrySet().
-    *
+    * <p>
+    * This set is a read only backed view of the entries underneath. This set will only show non expired entries when
+    * invoked. The size method of the set will count expired entries for the purpose of having a O(1) time cost compared
+    * to O(N) if it is to not count expired entries.
     * @return a set of immutable cache entries
     */
    Set<InternalCacheEntry<K, V>> entrySet();
--- a/core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
+++ b/core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
@@ -3,6 +3,9 @@ package org.infinispan.container;
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.mock;
 import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertNotNull;
+import static org.testng.AssertJUnit.assertNull;
+import static org.testng.AssertJUnit.assertTrue;
 
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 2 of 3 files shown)


wburns:
Issue was stream based operation that utilize SIZE, such as toArray caused an issue since our collection is CONCURRENT but it wasn't marked as such so it thought it could determine a stable size. By marking the spliterator as CONCURRENT it doesn't know an explicit size.

wburns:
Also converted test to using junit asserts.

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
```diff
@@ -59,38 +66,39 @@ public void testExpiredData() throws InterruptedException {
...
       dc.put("k", "v", new EmbeddedMetadata.Builder().lifespan(100, TimeUnit.MINUTES).build());
-      Thread.sleep(100);
+      timeService.advance(100);
       assertEquals(1, dc.size());
 
       entry = dc.get("k");
       assertNotNull(entry);
       assertEquals(mortaltype(), entry.getClass());
-      assertTrue(entry.getCreated() <= System.currentTimeMillis());
+      assertTrue(entry.getCreated() <= timeService.wallClockTime());
```
rvansa: `==` with the controlled time service => assertEquals

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
```diff
@@ -283,4 +291,41 @@ public void testGetDuringKeySetLoop() {
...
+
+      results = Arrays.asList(dc.entrySet().stream().toArray(Map.Entry[]::new));
+      assertEquals(2, results.size());
+      expected.remove(k1);
+      assertArrayAndSetContainSame(expected, results);
+
+      timeService.advance(100);
+
+      results = Arrays.asList(dc.entrySet().stream().toArray(Map.Entry[]::new));
+      assertEquals(1, results.size());
```
rvansa: Hmm, shouldn't iterating the container "touch" the entries? At least when running cache/stream ops, non-touching version is necessary for expiration process, too. But it's probably another issue...

core/src/main/java/org/infinispan/container/DataContainer.java
```diff
@@ -116,7 +120,10 @@
     * <p/>
     * If a client needs to iterate through a mutable set of mutable cache entries, it should iterate the container
     * itself rather than iterating through the return of entrySet().
-    *
+    * <p>
+    * This set is a read only backed view of the entries underneath. This set will only show non expired entries when
+    * invoked. The size method of the set will count expired entries for the purpose of having a O(1) time cost compared
+    * to O(N) if it is to not count expired entries.
```
danberindei: Shouldn't we deprecate these methods?

`DataContainer` doesn't extend `Map`, so we don't really need them, and we already have `iterator()` and `iteratorIncludingExpired()` to iterate through the entries.

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
```diff
@@ -283,4 +291,41 @@ public void testGetDuringKeySetLoop() {
 
       assertEquals(10, i);
    }
+
+   public void testEntrySetStreamWithExpiredEntries() {
```
rvansa: Since the set is CONCURRENT, I wonder if it would make sense to create a test that would add/remove entries in the middle of the toArray() call.

core/src/main/java/org/infinispan/container/DataContainer.java
wburns: To be honest I was talking to @tristantarrant mentioning that the API is just all CRAP. I am all for deprecating some of these.

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
wburns: Iterator doesn't touch them. Otherwise it totally throws off the eviction ordering. We can discuss this on dev list if we want to change this maybe?

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
wburns: Sure. I didn't think to go back and change all of these. This test was so outdated :P

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
```diff
@@ -59,38 +66,39 @@ public void testExpiredData() throws InterruptedException {
...
       assertNotNull(entry);
       assertEquals(transienttype(), entry.getClass());
-      assertTrue(entry.getLastUsed() <= System.currentTimeMillis());
+      assertTrue(entry.getLastUsed() <= timeService.wallClockTime());
       long entryLastUsed = entry.getLastUsed();
-      Thread.sleep(100);
+      timeService.advance(100);
       entry = dc.get("k");
       assertTrue(entry.getLastUsed() > entryLastUsed);
       dc.put("k", "v", new EmbeddedMetadata.Builder().maxIdle(0, TimeUnit.MINUTES).build());
```
wburns: Don't know, it used to be there, would have to look closer.

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
```diff
@@ -59,38 +66,39 @@ public void testExpiredData() throws InterruptedException {
       InternalCacheEntry entry = dc.get("k");
```
wburns: Oops thanks.

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
wburns: And actually we could do the Infinispan "touch" without the eviction "touch". So that is possible as well.

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
wburns: I guess to make sure the metadata is overridden properly. I can't say otherwise.  Seems safe to leave.

wburns:
Updated

rvansa:
@wburns What about "a test that would add/remove entries in the middle of the toArray() call"?

wburns:
@rvansa I would say we can add a stress test in a different PR.
And in regards to the other issue, I will make a different PR for that, it has nothing to do with this (I was planning on doing it with another issue that is tightly related).

wburns:
Created #5019 to handle the other issue I mentioned.

core/src/main/java/org/infinispan/container/DataContainer.java
```diff
@@ -99,12 +99,16 @@
     * Returns a set of keys in the container. When iterating through the container using this method, clients should
     * never call {@link #get(Object)} method but instead {@link #peek(Object)}, in order to avoid changing the order of
     * the underlying collection as a side of effect of iterating through it.
-    *
+    * <p>
+    * This set of keys will include expired entries. If you wish to only retrieve non expired keys pleas use the
```
danberindei: pleas*e*

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
danberindei: -1 to touch entries when iterating, both for eviction and for maxIdle expiration. Iterating with a cache store is already slow enough, updating the timestamp of every entry would make it unusable IMO.

wburns: Yeah I forgot about stores.  And iirc we don't even touch store entries on a get currently, have to find JIRA.

wburns: ISPN-3202

wburns:
Fixed typo.

core/src/test/java/org/infinispan/container/SimpleDataContainerTest.java
rvansa: I understand the argument about cache stores (if we ever decide to deal with ISPN-3202, the update should happen asynchronously). However I still think that we should touch the entries in data container - that is cheap (not in this PR, though).

wburns: Sounds good, we can discuss/change this in a different issue.

danberindei:
Integrated, thanks Will!

PR #5012, opened 2017-03-23
Make use of Creaper in server test suite

mgencur:
https://issues.jboss.org/browse/ISPN-7663

This is for preview only.

This is only a part of the migration and I want to get your opinion before I move forward.
Some important notes:
* so far only tests that run with a single node (and not using kill/stop by arquillian) are migrated, this is about one third of all tests (maybe even less)
* the test execution of the single-node tests takes 4 min 12 sec on my machine while on master branch (without Creaper) it is 5 min 33 s
* this would be even faster without these issues that I filed: ISPN-7609, ISPN-7633, ISPN-7662
* I haven't migrated any two-node, three-node tests and any tests that kill the server during the test
* currently there's "SingleNode" JUnit Group and I'd expect creating TwoNode and ThreeNode groups later if we agree that this is the way we want to go, the server(s) are started before the suite and later only modified by Creaper (and sometimes reloaded if required)
* calling Creaper can be made nice by using a *builder* pattern and pushing the changes to Creaper project itself, but that would come later because we'd need to wait for Creaper's release

* the issue I haven't solved is that some classes already had a JUnit group, so when I want to run them only once I had to remove that old JUnit group and add the new SingleNode group (this way they don't run as part of their original group, e.g. Query tests) - not sure it is a big problem, though


tristantarrant:
FYI https://github.com/infinispan/infinispan/pull/5023 fixes ISPN-7662

galderz:
Looks good to me. Anyone else? Needs rebasing when agreed to move this way.

mgencur:
Thanks @galderz , I will continue with changing the server test suite in this way. There are more tests for migration. When I have something ready I will open a new PR. I have to release a new version of infinispan-arquillian-container anyway. For now I'm closing this one.

PR #5013, opened 2017-03-23
Server warning and doc fixes

ryanemerson:
https://issues.jboss.org/browse/ISPN-7658
https://issues.jboss.org/browse/ISPN-7664
https://issues.jboss.org/browse/ISPN-7665
https://issues.jboss.org/browse/ISPN-7666

### DIFF
--- a/documentation/src/main/asciidoc/server_guide/configuration.adoc
+++ b/documentation/src/main/asciidoc/server_guide/configuration.adoc
@@ -322,8 +322,52 @@ The possible attributes for the locking element are:
 While it is possible to configure server caches to be transactional, none of the available protocols offer transaction capabilities.
 
 ==== Loaders and Stores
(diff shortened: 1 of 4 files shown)


server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerConfigurationBuilder.java
```diff
@@ -226,6 +227,9 @@ public GlobalConfiguration getValue() {
                 .allowDuplicateDomains(true);
 
         builder.addModule(PrivateGlobalConfigurationBuilder.class).serverMode(true);
+
+        // ISPN-7644 Set the default cache name so that ISPN000435 warning is not displayed on server startup
+        builder.defaultCacheName(CacheContainer.DEFAULT_CACHE_NAME);
```
tristantarrant: Actually, the server shouldn't even bother with the default cache. If it is starting it, then we need to prevent that.

PR #5015, opened 2017-03-23
ISPN-7461 Cache is not rebalanced on merge

jurakp:
https://issues.jboss.org/browse/ISPN-7461

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1421404

Removed problematic if block.

### DIFF
--- a/core/src/main/java/org/infinispan/topology/DefaultRebalancePolicy.java
+++ b/core/src/main/java/org/infinispan/topology/DefaultRebalancePolicy.java
@@ -20,6 +20,7 @@
 package org.infinispan.topology;
 
 import org.infinispan.distribution.ch.ConsistentHash;
+import org.infinispan.distribution.ch.ConsistentHashFactory;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.jmx.annotations.DataType;
 import org.infinispan.jmx.annotations.ManagedAttribute;
@@ -61,7 +62,7 @@ public class DefaultRebalancePolicy implements RebalancePolicy {
          return;
       }
 
-      if (!cacheStatus.hasJoiners() && isBalanced(cacheStatus.getCacheTopology().getCurrentCH())) {
+      if (!cacheStatus.hasJoiners() && isBalanced(cacheStatus)) {
          log.tracef("Not triggering rebalance for cache %s, no joiners and the current consistent hash is already balanced",
                cacheName);
          return;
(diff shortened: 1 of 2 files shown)


wburns:
@danberindei can you take a look at this? It seems this might be a good one for 9.1, wdyt?

rvansa:
@wburns It's meant for 5.2 :) I don't think we can guarantee much after merge - merge without partition handling is scarcely tested (if at all). A proper solution is related to AP mode - diverged entries merge (I don't know if @ryanemerson is about to work on that given that he has worked on the consistency checker).

ryanemerson:
@rvansa @wburns The plan is for Automatic Merging with various merge strategies to be implemented for 9.1. This is currently a WIP, but you can see the rough design docs [here](https://github.com/infinispan/infinispan/wiki/Conflict-resolution).

wburns:
@rvansa Oh I know that we can't guarantee anything after merge in regards to data. I was more asking because this change only has to do with the topology updates.

danberindei:
Oops, I missed this one... for some reason I'm showing up as a reviewer here but not in the PR list :)

I'm don't think this is the right approach. `DefaultRebalancePolicy.updateCacheStatus()` is called from lots of places, and I'm afraid without this check it could get into an infinite loop. I would rather fix `isBalanced()`, making it more like the check in `startRebalance()`:

         ConsistentHash updatedMembersCH = chFactory.updateMembers(currentCH, newMembers);
         ConsistentHash balancedCH = chFactory.rebalance(updatedMembersCH);
         return balancedCH.equals(currentCH);

jurakp:
I have rewritten it based on @danberindei comment. The main problem was encapsulation of cache status map. I did not want to modify ClusterTopologyManagerImpl just for test, therefore I use refrection to get that map in TestingUtil.

core/src/main/java/org/infinispan/topology/DefaultRebalancePolicy.java
```diff
@@ -84,15 +85,21 @@ public void updateCacheStatus(String cacheName, ClusterCacheStatus cacheStatus)
...
+       if(cacheStatus == null){
+           return true;
+       }
+       synchronized(cacheStatus){
+           CacheTopology cacheTopology = cacheStatus.getCacheTopology();
+           ConsistentHashFactory chFactory = cacheStatus.getJoinInfo().getConsistentHashFactory();
+           ConsistentHash currentCH = cacheTopology.getCurrentCH();
+           ConsistentHash updatedMembersCH = chFactory.updateMembers(currentCH, cacheStatus.getMembers());
+           ConsistentHash balancedCH = chFactory.rebalance(updatedMembersCH);
+           if (currentCH == null || balancedCH.equals(currentCH)) {
```
danberindei: If `currentCH == null`, `updateMembers(currentCH, ...)` would throw a NPE. So we should either check for `null` before, or remove the check here. I don't think it's possible to have a `CacheTopology` with a `currentCH == null`, is it?

jurakp:
@danberindei You were right. I have removed unnecessary null check.

danberindei:
Integrated, thanks Petr!

PR #5019, opened 2017-03-23
ISPN-7669 Evaluate DataContainer size method invocations to make sure

wburns:
they make sense

* Changed size call to use sizeIncludingExpired
* Refactor size method to use another option when possible

https://issues.jboss.org/browse/ISPN-7669

### DIFF
--- a/core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
+++ b/core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
@@ -99,13 +99,14 @@ public class PassivationManagerImpl implements PassivationManager {
       if (enabled && !skipOnStop) {
          long start = timeService.time();
          log.passivatingAllEntries();
+         long count = 0;
          for (InternalCacheEntry e : container) {
+            count++;
             if (trace) log.tracef("Passivating %s", e.getKey());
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
```diff
@@ -99,13 +99,16 @@ public void passivateAll() throws PersistenceException {
       if (enabled && !skipOnStop) {
          long start = timeService.time();
          log.passivatingAllEntries();
+         int count = 0;
          for (InternalCacheEntry e : container) {
+            if (count != Integer.MAX_VALUE) {
+               count++;
```
danberindei: Could use a `long` instead.

wburns: That is what I had originally, but the passivatedEntries log method requires int. Want me to update both?

wburns:
Updated.

wburns:
If we need to @tristantarrant this can be delayed to 9.1, but it is pretty trivial and should help performance a bit for all bulk operations.

tristantarrant:
Merged to master, thanks

PR #5021, opened 2017-03-23
ISPN-7430 Slowdown when using PutAll with transactions

wburns:
* Fix lock availabilty to perform better with multiple keys

https://issues.jboss.org/browse/ISPN-7430


gustavonalle:
Merged, thanks @wburns

PR #5022, opened 2017-03-24
ISPN-7673 object-filter: missing phase from antlr3-maven-plugin

danberindei:
https://issues.jboss.org/browse/ISPN-7673


anistor:

PR #5026, opened 2017-03-24
ISPN-7671 Deprecate DataContainer collection methods

wburns:
* Removed references to collection methods
* Need to update SimpleCacheImpl still

https://issues.jboss.org/browse/ISPN-7671


wburns:
I didn't remove the SimpleCacheImpl references as this should be refactored completely to use some common class between SimpleCacheImpl and EntrySetCommand classes. This should reduce a lot of code, but I didn't want to introduce that so close to Final.

tristantarrant:
Merged, thanks

PR #5028, opened 2017-03-27
ISPN-7300 Operator '-' doesn't work in fulltext querying

anistor:
https://issues.jboss.org/browse/ISPN-7300

### DIFF
--- a/query/src/main/java/org/infinispan/query/dsl/embedded/impl/LuceneQueryMaker.java
+++ b/query/src/main/java/org/infinispan/query/dsl/embedded/impl/LuceneQueryMaker.java
@@ -11,6 +11,7 @@ import org.apache.lucene.index.Term;
 import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.BoostQuery;
+import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.PrefixQuery;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.RegexpQuery;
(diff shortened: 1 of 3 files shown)


query/src/main/java/org/infinispan/query/dsl/embedded/impl/LuceneQueryMaker.java
```diff
@@ -115,6 +116,28 @@ public LuceneQueryMaker(SearchIntegrator searchFactory, FieldBridgeProvider<Type
...
+         BooleanQuery booleanQuery = (BooleanQuery) query;
+         boolean allClausesAreMustNot = true;
+         for (BooleanClause clause : booleanQuery.clauses()) {
+            if (clause.getOccur() != BooleanClause.Occur.MUST_NOT) {
+               allClausesAreMustNot = false;
+               break;
+            }
+         }
+         if (allClausesAreMustNot) {
+            //It is illegal to have only must-not queries, in this case we need to add a positive clause to match everything else.
```
gustavonalle: Why it's illegal?

query/src/main/java/org/infinispan/query/dsl/embedded/impl/LuceneQueryMaker.java
```diff
@@ -115,6 +116,28 @@ public LuceneQueryMaker(SearchIntegrator searchFactory, FieldBridgeProvider<Type
          isAnalyzerRemote = true;
       }
       Query query = makeQuery(parsingResult.getWhereClause());
+
+      // an all negative top level boolean query is not allowed; needs a bit of rewriting
+      if (query instanceof BooleanQuery) {
+         BooleanQuery booleanQuery = (BooleanQuery) query;
+         boolean allClausesAreMustNot = true;
```
gustavonalle: this and next 6 lines can be replaced by 
```booleanQuery.clauses().stream().allMatch(p -> p.getOccur() == BooleanClause.Occur.MUST_NOT);```

File: query/src/main/java/org/infinispan/query/dsl/embedded/impl/LuceneQueryMaker.java [hunk 3]
```diff
@@ -115,6 +116,28 @@ public LuceneQueryMaker(SearchIntegrator searchFactory, FieldBridgeProvider<Type
          isAnalyzerRemote = true;
       }
       Query query = makeQuery(parsingResult.getWhereClause());
+
+      // an all negative top level boolean query is not allowed; needs a bit of rewriting
```
gustavonalle: So this means queries for example like ```-cancelled:true -delayed:true``` for hypothetical flights would not be allowed?

gustavonalle:
LGTM, have a couple of questions though

query/src/main/java/org/infinispan/query/dsl/embedded/impl/LuceneQueryMaker.java
anistor: Nice suggestion. Done.

query/src/main/java/org/infinispan/query/dsl/embedded/impl/LuceneQueryMaker.java
anistor: Because Lucene is weird, and "-text:foo" never matches anything.
But "-text:foo #*:*" does match what we intend.

query/src/main/java/org/infinispan/query/dsl/embedded/impl/LuceneQueryMaker.java
anistor: This comment was referring to the same case I described earlier. So if all clauses are negative I automatically add #*.* to the query or else it will (wrongly) discard everything.

gustavonalle:
@anistor ok, thanks for the clarification

gustavonalle:
Integrated, thanks @anistor !

PR #5029, opened 2017-03-28

alanfx:
JUnit reports

Fix hard coded path separator

### DIFF
--- a/commons-test/src/main/java/org/infinispan/commons/test/PolarionJUnitXMLReporter.java
+++ b/commons-test/src/main/java/org/infinispan/commons/test/PolarionJUnitXMLReporter.java
@@ -189,7 +189,7 @@ public class PolarionJUnitXMLReporter implements IResultListener2, ISuiteListene
       document.pop();
 
       // Reset output directory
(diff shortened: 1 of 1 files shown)


alanfx:
@danberindei here you go! And anyone else can look too. TRIVIAL!!!

commons-test/src/main/java/org/infinispan/commons/test/PolarionJUnitXMLReporter.java
```diff
@@ -189,7 +190,8 @@ protected void generateReport(ISuite suite) {
       document.pop();
 
       // Reset output directory
-      Utils.writeUtf8File(suite.getOutputDirectory().replaceAll("/Surefire suite", ""),
+      Utils.writeUtf8File(
+            suite.getOutputDirectory().replaceAll(String.format("%sSurefire suite", File.pathSeparator), ""),
```
tristantarrant: Actually, since replaceAll takes a regex as the first parameter you could write:
`Utils.writeUtf8File(suite.getOutputDirectory().replaceAll(".Surefire` suite", "")
instead

alanfx:
@tristantarrant and so it is done!

tristantarrant:
Pushed, thanks

PR #5032, opened 2017-03-29
ISPN-7678 Endpoints should not depend on the default cache

tristantarrant:
- Prevents endpoint restart on default cache reconfiguration

ISPN-7683 Unregisters the internal topology cache on server stop

https://issues.jboss.org/browse/ISPN-7678
https://issues.jboss.org/browse/ISPN-7683

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/cache/Configurations.java
+++ b/core/src/main/java/org/infinispan/configuration/cache/Configurations.java
@@ -70,4 +70,8 @@ public class Configurations {
       PrivateGlobalConfiguration config = globalConfiguration.module(PrivateGlobalConfiguration.class);
       return config == null || !config.isServerMode();
    }
+
+   public static boolean isClustered(GlobalConfiguration globalConfiguration) {
+      return globalConfiguration.transport().transport() != null;
+   }
 }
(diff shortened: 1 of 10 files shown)


ryanemerson:
@tristantarrant Tested locally and this fixes the previous problems. Waiting on CI.

tristantarrant:
@ryanemerson I've also added ISPN-7683

tristantarrant:
It seems like the changes are causing a number of regressions in some server tests. I'll come up with a better solution

danberindei:
@tristantarrant couldn't we also restart the cache manager when we restart the server?

tristantarrant:
The issue here is backwards: the server's highly concurrent service design makes the endpoint start as soon as the cache manager service is installed. However this could happen before any actual cache services are started. The Hot Rod endpoint, for legacy reasons, wants to start the default cache (which might not have been configured yet). The fixes we need are:

- make the Hot Rod endpoint start regardless of the presence of any caches
- requesting a cache over Hot Rod should actually wait for the cache (if present obviously) to be started by the server. If the cache is clustered, we should ensure that it is started on all nodes.

tristantarrant:
Closing this for now

PR #5035, opened 2017-03-29
ISPN-6896 Remove scala from server/hotrod module

danberindei:
https://issues.jboss.org/browse/ISPN-6896

The converted Java code isn't very elegant, but it works. We can decide which tests to keep later :)


tristantarrant:
Merged, great stuff @danberindei

PR #5037, opened 2017-03-29
ISPN-7686 Stat getNumberOfEntries should not include expired entries.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7686

@tristantarrant I've only updated the implementation of `CacheMgmtInterceptor` as calling `StatsImpl::getNumberOfEntries` in `DistributedCacheStatsCallable::call` is the equivalent of executing the `SizeCommand` on all nodes. Logic regarding numOwners and distributed mode is already handled in `ClusterCacheStatsImpl`.


core/src/main/java/org/infinispan/interceptors/impl/CacheMgmtInterceptor.java
```diff
@@ -344,12 +359,14 @@ public long getAverageRemoveTime() {
...
-         description = "Number of entries currently in memory including expired entries",
+         description = "Number of entries currently in memory excluding expired entries",
          displayName = "Number of current cache entries",
          displayType = DisplayType.SUMMARY
    )
    public int getNumberOfEntries() {
-      return dataContainer.sizeIncludingExpired();
+      long flags = EnumUtil.bitSetOf(Flag.CACHE_MODE_LOCAL, Flag.SKIP_CACHE_LOAD);
+      SizeCommand command = commandsFactory.buildSizeCommand(flags);
+      return (Integer) invoker.invoke(icf.createInvocationContext(false, UNBOUNDED), command);
```
wburns: By the way if you just want in memory but excluding expired. You could call `dataContainer.size` as this will not count expired entries.

ryanemerson: @wburns Right, that is much simpler!

tristantarrant:
I believe we still don't have a way to show the actual size of the caches / containers in the cluster from the management endpoints, including passivated entries.

wburns:
Hrmm, is this not `Cache.size` ? This should include counts from all stores as well as in memory excluding expired entries.

tristantarrant:
@wburns I am talking about JMX and DMR

wburns:
Oh I misunderstood. You mean they are not exposed through there. 👍

ryanemerson:
@tristantarrant @wburns updated.

tristantarrant:
Pushed to master, thanks

PR #5038, opened 2017-03-29
ISPN-7685 JDBC Sybase upsert fails.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7685


ryanemerson:
It would be nice to get this into final, however a workaround exists (disable upsert) so not urgent!

PR #5041, opened 2017-03-30
Single port design

slaskawi:
Hey,

Please have a look at the Single Endpoint for Infinispan Server design. 

There's also a bonus here... Since our Wiki doesn't allow inline comments, I would like to propose storing designs in a dedicated directory along with the sourcecode. This approach has been successfully used by Kubernetes and I highly encourage us to take this path.

Thanks,
Sebastian


tristantarrant:
Let's use the newly-created repo https://github.com/infinispan/infinispan-designs for this
Using the main infinispan repo creates noise, triggers CI, etc.

PR #5042, opened 2017-03-30
ISPN-6893 Remove scala dependency from all modules

danberindei:
https://issues.jboss.org/browse/ISPN-6893

Remove all the dependencies on scala, so we can close the parent JIRA.


tristantarrant:
testing locally

PR #5045, opened 2017-03-31
Create HotRodMetadata to avoid conflicts

pruivo:
Hot Rod was using Metada.version() and it conflicts with the write-skew versions

under discussion in https://issues.jboss.org/browse/ISPN-7672


rvansa:
I don't see any tests... (but I understand this is a preview).

Why can't you use SimpleClusteredVersion for the purpose HR uses NumericVersion? It seems that if you just move topologyId to first 2 bytes, you get the same, though I don't know if rank is really needed when the versions are generated on primary owner.

pruivo:
I think it is in the other way. You need the rank and not the topology id. The client can contact any node if the primary takes time to reply.

just to recall, there are 2 issues: the first is the use of VersionGenerator assuming a NumericVersionGenerator. It is not true for TX+OL+RR+WS. Second, it uses Metadata.version to store its version. It is bad because we store the versions internally to keep data consistent.

rvansa:
Rank itself has little significance: NVM uses viewId + rank to identify the node, so it's alternative way to encode the address.
So in case of HR, is it the originator who sets the version in metadata? Why can't we just rely on the version generated by WS, and not use any originator-provided version at all?

pruivo:
Rank is the view id + position in the view: https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/container/versioning/NumericVersionGenerator.java#L101

In HR, the originator creates the version: https://github.com/infinispan/infinispan/blob/master/server/hotrod/src/main/java/org/infinispan/server/hotrod/CacheDecodeContext.java#L134

I guess we can use the WS version when it is available. I'm not aware of any issues at the moment. I'll give it a try.

rvansa:
Nope, rank == position in the view, and is 2 bytes long. (view id, rank) is called "version prefix".

Using WS only when it's available does not seem as clean design, either provide it always from the originator (as done before) or always use some version provided by primary (enforce some version generator in server).

pruivo:
@rvansa I think will continue improve this PR. Splitting the version from HR from the entry version. This way it will more future proof. Probably, we need a new version generator for eventual consistent as well.

rvansa:
@pruivo I am working on ISPN-7590 last few weeks and this needs separate mechanism (API in Metadata) - you can't just store one version, but entire invocations history in order to prevent retried application of given command.

PR #5054, opened 2017-04-05
ISPN-7696 RemoveExpired could be called with updated value

wburns:
https://issues.jboss.org/browse/ISPN-7696

### DIFF
--- a/core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
+++ b/core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
@@ -76,9 +76,20 @@ public class ClusterExpirationManager<K, V> extends ExpirationManagerImpl<K, V> 
                  purgeCandidates.hasNext();) {
                InternalCacheEntry<K, V> e = purgeCandidates.next();
                if (e.canExpire()) {
-                  if (ExpiryHelper.isExpiredMortal(e.getLifespan(), e.getCreated(), currentTimeMillis)) {
-                     handleLifespanExpireEntry(e, true);
-                  } else if (ExpiryHelper.isExpiredTransient(e.getMaxIdle(), e.getLastUsed(), currentTimeMillis)) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
```diff
@@ -76,10 +76,14 @@ public void processExpiration() {
...
                if (e.canExpire()) {
-                  if (ExpiryHelper.isExpiredMortal(e.getLifespan(), e.getCreated(), currentTimeMillis)) {
-                     handleLifespanExpireEntry(e, true);
-                  } else if (ExpiryHelper.isExpiredTransient(e.getMaxIdle(), e.getLastUsed(), currentTimeMillis)) {
-                     super.handleInMemoryExpiration(e, currentTimeMillis);
+                  // Have to synchronize on the entry to make sure we see the value and metadata at the same time
+                  synchronized (e) {
+                     long lifespan = e.getLifespan();
+                     if (ExpiryHelper.isExpiredMortal(lifespan, e.getCreated(), currentTimeMillis)) {
+                        handleLifespanExpireEntry(e.getKey(), e.getValue(), lifespan, true);
```
rvansa: You don't need to call this within the synchronized block.

wburns: I am aware. I was being more lazy and to be honest it doesn't save very much as this lock is only on the entry. I can change it up though.

wburns: And actually I also would need synchronization around the `else if` block, which is probably why I didn't separate it. Code will be come quite messy. I will see if I can clean it up more though.

wburns:
Fixed.

rvansa:
Great, let's integrate this.

ryanemerson:
@wburns Integrated, thanks!

PR #5055, opened 2017-04-05
ISPN-7697 DefaultDataContainer.containsKey expires entry without

wburns:
synchronization and notification

https://issues.jboss.org/browse/ISPN-7697


tristantarrant:
Pushed to master, thanks

PR #5060, opened 2017-04-07
ISPN-7704 Support full paths in lookupFileStrict

slaskawi:
https://issues.jboss.org/browse/ISPN-7704

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/AbstractFileLookup.java
+++ b/commons/src/main/java/org/infinispan/commons/util/AbstractFileLookup.java
@@ -10,13 +10,13 @@ import java.net.URI;
 import java.net.URL;
 import java.util.Collection;
 
-import org.infinispan.commons.logging.BasicLogFactory;
+import org.infinispan.commons.logging.Log;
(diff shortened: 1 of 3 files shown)


commons/src/main/java/org/infinispan/commons/util/AbstractFileLookup.java
```diff
@@ -60,11 +60,14 @@ public InputStream lookupFileStrict(String filename, ClassLoader cl) throws File
...
-      String fileName = uri.toString().substring(startIndex + 1);
-
-      return cl.getResourceAsStream(fileName);
+      String scheme = uri.getScheme();
+      switch (scheme) {
+         case "file":
+            String fileName = uri.getRawPath();
+            return new FileInputStream(new File(fileName));
+         default:
+            throw log.schemeNotSupported(scheme);
```
tristantarrant: Don't we want to try classloading as fallback ?

slaskawi: Very good point, let me rewrite this...

PR #5063, opened 2017-04-07
ISPN-7708 Use UnsuccesfulResponse to signalize command failure

rvansa:
https://issues.jboss.org/browse/ISPN-7708


pruivo:
@rvansa I guess the ```WriteResponse``` can be removed and the triangle can use the Successful and Unsuccessful command. wdyt?

commons/src/main/java/org/infinispan/commons/util/Util.java
```diff
@@ -475,6 +476,13 @@ public static String toStr(Object o) {
          return printArray((byte[]) o, false);
       } else if (o == null) {
          return "null";
+      } else if (o.getClass().isArray()) {
```
pruivo: why not invoking ```printArray()```?

rvansa: If I see correctly, `printArray` handles only `byte[]` and this was already handled earlier in this if-else switch. This variant is for generic arrays. I could refactor that to `printArray(Object)`, if you want, but we can't even use `printArray(T[])` since we don't know the type of array statically.

rvansa:
@pruivo OK, it seems we can remove the `WriteResponse` and `TriangleResponseGenerator`. Should I do that in this PR or in a follow-up one?

commons/src/main/java/org/infinispan/commons/util/Util.java
pruivo: I mean, ```Arrays.toString()``` *cof* *cof* 
for some reason, I though we had a generic printArray :(

pruivo:
@rvansa if you don't mind, it can be done in this PR. But if you are busy, I'll do it after it is integrated :)

commons/src/main/java/org/infinispan/commons/util/Util.java
rvansa: There's no generic `Arrays.toString(Object)`. But you're right that I can probably do the `(Object[]) o` cast, since Java arrays suffer the covariance.

rvansa:
@pruivo Rebased & updated.

pruivo:
@rvansa it looks good but I'm not 100% convinced that it works with the triangle because the triangle always need a reply from the primary owner. Let me write some tests to double check if everything is ok.

pruivo:
everything is fine. it is ready to go. just waiting for CI :+1:

rvansa:
@pruivo Will you add those extra tests when integrating this PR, too?

commons/src/main/java/org/infinispan/commons/util/Util.java
danberindei: You're missing some more primitives here, I'm getting a `ClassCastException` in `MemoryBasedEvictionFunctionalTest.testDoubleArray`.

rvansa:
@danberindei That's what you get when running with trace logging by default :) Fixed.

pruivo:
@rvansa I haven't added any test. I guess it is good to go...

pruivo:
integrated! thanks @rvansa !

PR #5064, opened 2017-04-07
ISPN-7590 Proper command retry handling

rvansa:
https://issues.jboss.org/browse/ISPN-7708


rvansa:
run performance tests please
library

rvansa:
Wow, the perf ack shows no difference whatsover - I didn't expect it being *that* good. Now I should test if it doesn't leak memory...

rvansa:
run performance tests please
library

jmarkos:
run performance tests please
library

jmarkos:
run performance tests please
library

wburns:
Looks like a rebase is required now that your other PR is in :)

core/src/main/java/org/infinispan/commands/InvocationRecord.java
```diff
@@ -0,0 +1,271 @@
...
+public final class InvocationRecord {
+   private static final int AUTHORITATIVE = 1;
+   private static final int CREATED = 2;
+   private static final int MODIFIED = 4;
+   private static final int REMOVED = 6;
+
+   public final CommandInvocationId id;
+   // TODO: maybe it would be better to store previous value and metadata and let the command run again
+   public final Object returnValue;
+   // true if this command was written when the node was primary owner of given entry
```
wburns: Seems this is outdated now.

core/src/main/java/org/infinispan/commands/InvocationManager.java
```diff
@@ -0,0 +1,48 @@
...
+public interface InvocationManager {
+   /**
+    * @return Time (in milliseconds) after which an {@link org.infinispan.commands.InvocationRecord} should expire.
+    */
+   long invocationTimeout();
+
+   /**
+    * @return Delegated value from {@link TimeService#wallClockTime()}
+    */
+   long wallClockTime();
```
wburns: I would think we would want to get this directly from `TimeService`

core/src/main/java/org/infinispan/commands/InvocationManager.java
```diff
@@ -0,0 +1,48 @@
...
+    * shouldn't be called - we'll need to wait until the invocations expire. In async caches even the originator stores
+    * the invocation record, because the entry can be stored
+    *
+    * @param commandInvocationId
+    * @param key
+    * @param segment
+    */
+   void notifyCompleted(CommandInvocationId commandInvocationId, Object key, int segment);
+
+   void notifyCompleted(CommandInvocationId commandInvocationId, Collection<?>[] keysBySegment);
```
wburns: Adding a comment as to how this should be used would be great.

core/src/main/java/org/infinispan/commands/impl/InvocationManagerImpl.java
```diff
@@ -0,0 +1,185 @@
...
+            fic.init(dataContainer);
+            fic.setOrigin(rpcManager.getAddress());
+            if (trace) {
+               log.tracef("Invoking %s", fic);
+            }
+            fic.invoke();
+            if (trace) {
+               Set<Object> keys = new HashSet<>(Arrays.asList(fic.keys()));
+               log.tracef("Invalidation completed: " + StreamSupport.stream(dataContainer.spliterator(), false)
+                     .filter(ice -> keys.contains(ice.getKey())).collect(Collectors.toList()));
```
wburns: Minor, but you can use [1] instead to avoid the List

[1] https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#joining--

core/src/main/java/org/infinispan/commands/InvocationManager.java
rvansa: I didn't want to inject both `TimeService` and `InvocationManager` to all commands, as growing the commands increases GC pressure.

core/src/main/java/org/infinispan/commands/impl/InvocationManagerImpl.java
rvansa: Great, thanks for the tip.

rvansa:
@wburns @danberindei As per perfack results, regrettably this PR introduces a regression on writes, which is expected as it simply stores more and sends more messages (I was quite surprised when I first saw the results which showed no difference whatsover). It is possible that we could be resolved, but now we should decide if we value more the performance or correctness in the short-time view: Should we integrate this (assuming that the rest of review comments are handled) and work towards better perf in master (being able to add incremental improvements, by multiple developers), or should we try to keep improving this just as PR?

rvansa:
Regarding correctness, this regrettably does not fix ISPN-3918, and there are more ways how could we address that - it could be discussed in that ticket directly, but I won't try to grow this PR by handling everything.

danberindei:
@rvansa Is there a summary of the changes somewhere?

rvansa:
@danberindei Good point, I should write a summary.

core/src/main/java/org/infinispan/commands/functional/AbstractWriteManyCommand.java
```diff
@@ -1,29 +1,56 @@
...
    Params params;
    // TODO: this is used for the non-modifying read-write commands. Move required flags to Params
    // and make sure that ClusteringDependentLogic checks them.
    long flags;
+   transient InvocationManager invocationManager;
+   transient Set<Object> completedKeys;
+   transient boolean authoritative;
 
-   protected AbstractWriteManyCommand(CommandInvocationId commandInvocationId) {
+   protected AbstractWriteManyCommand(CommandInvocationId commandInvocationId, InvocationManager invocationManager, boolean synchronous) {
```
rvansa: oops, not storing `synchronous`

rvansa:
@danberindei Design info https://github.com/infinispan/infinispan-designs/pull/5

rvansa:
Rebased. I thought the line counts could look more friendly after all gets in; the first three commits could be merged independently, but most of the weight can't be included incrementally.

rvansa:
Closing for now. We have agreed on not passing the return value from primary to backups, but implementing Strict Ordering (retrieving previous value if not present on backup) in all write commands. That way we can guarantee correct previous values as needed for continuous query, regular query and other event consumers.
If the new implementation is ready, it should get in in early 9.2 development.
Questions have risen about the actual need to do this in transactional mode.

PR #5067, opened 2017-04-10
ISPN-7659 Fix build with JDK9 ea-164

danberindei:
https://issues.jboss.org/browse/ISPN-7659

### DIFF
--- a/integrationtests/pom.xml
+++ b/integrationtests/pom.xml
@@ -14,11 +14,70 @@
     <description>Infinispan Integration Tests Parent</description>
     <packaging>pom</packaging>
 
+    <properties>
+        <forkJvmArgs>-Xmx2G</forkJvmArgs>
+        <jigsawForkJvmArgs></jigsawForkJvmArgs>
+    </properties>
+
     <profiles>
         <profile>
             <id>distribution</id>
             <properties>
                 <skipTests>true</skipTests>
+            </properties>
+        </profile>
+        <profile>
+            <id>customForkJvmArgs</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+                <property>
+                    <name>env.MAVEN_FORK_OPTS</name>
+                </property>
+            </activation>
+            <properties>
+                <forkJvmArgs>${env.MAVEN_FORK_OPTS}</forkJvmArgs>
+            </properties>
+        </profile>
+        <profile>
+            <id>jigsaw</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <properties>
+                <!--
+                  java.lang, java.util, java.io, java.lang.invoke, java.lang, reflect, java.util.concurrent, java.time:
+                    core externalizers for JDK types
+                  java.nio, jdk.internal.ref, sun.nio.ch: Netty, server/hotrod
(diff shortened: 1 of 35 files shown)


integrationtests/pom.xml
```diff
@@ -14,13 +14,72 @@
...
+                    --add-opens=java.base/java.security=ALL-UNNAMED
+                    --add-exports=jdk.security.jgss/com.sun.security.sasl.gsskerb=ALL-UNNAMED
+                    --add-exports=java.base/sun.nio.ch=ALL-UNNAMED
+                    --add-opens=java.base/sun.nio.ch=ALL-UNNAMED
+                    --add-exports=java.security.sasl/com.sun.security.sasl.digest=ALL-UNNAMED
+                    --add-exports=java.security.sasl/com.sun.security.sasl.ntlm=ALL-UNNAMED
+                    --add-exports=java.naming/com.sun.jndi.ldap=ALL-UNNAMED
+                    --add-opens=java.management/javax.management=ALL-UNNAMED
+                    --add-exports=java.management/com.sun.jmx.mbeanserver=ALL-UNNAMED
+                    --add-opens=java.base/com.sun.crypto.provider=ALL-UNNAMED
```
tristantarrant: Damn, it seems like we'll also need to whitelist for IBM JDK9 too...

tristantarrant:
Fails with Java 8:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.0.0:shade (default) on project infinispan-embedded-query: Execution default of goal org.apache.maven.plugins:maven-shade-plugin:3.0.0:shade failed. NullPointerException -> [Help 1]

danberindei:
@tristantarrant I forgot to add this in the description: some Maven plugins (e.g. maven-war-plugin) also play tricks with JDK classes, so you need this:

    export MAVEN_OPTS="--add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.text=ALL-UNNAMED --add-opens java.desktop/java.awt.font=ALL-UNNAMED"

integrationtests/pom.xml
danberindei: They had the same package name in IBM JDK8, I'm sure they'll want to preserve backwards compatibility :)

danberindei:
@tristantarrant I had to revert `maven-shade-plugin` back to version 2.4.3, because version 3.0.0 fails with a NPE when one of the dependencies doesn't have a source jar ([MSHADE-247](https://issues.apache.org/jira/browse/MSHADE-247)). There is no `antlr-2.7.7-source.jar` in Maven Central, and it only worked in my local testing because I had an additional repository that did have it. Luckily, 3.0.0 doesn't seem necessary for building on Java 9, but I'm sure we'll need a newer version when we add modules to our code.

danberindei:
@tristantarrant Fixed one more problem: I was adding the `--add-opens` VM options even for JDK8 in the `jcache-tck-runner` module. (Because `wildfly-maven-plugin` needs separate `<javaOpt>` elements, it can't use the `${jigsawForkJvmOpts} property that we use in the other modules.)

tristantarrant:
Looks good !

PR #5071, opened 2017-04-11
ISPN-7631 Cache.entrySet() entries don't implement Map.Entry.equals()…

pruivo:
… correctly

https://issues.jboss.org/browse/ISPN-7631

### DIFF
--- a/core/src/main/java/org/infinispan/container/entries/AbstractInternalCacheEntry.java
+++ b/core/src/main/java/org/infinispan/container/entries/AbstractInternalCacheEntry.java
@@ -1,5 +1,8 @@
 package org.infinispan.container.entries;
 
+import java.util.Map;
+import java.util.Objects;
+
 import org.infinispan.container.DataContainer;
 import org.infinispan.metadata.Metadata;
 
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/infinispan/container/entries/AbstractInternalCacheEntry.java
```diff
@@ -135,4 +138,23 @@ public AbstractInternalCacheEntry clone() {
...
+
+      return Objects.equals(getKey(), that.getKey()) && Objects.equals(getValue(), that.getValue());
+   }
+
+   @Override
+   public final int hashCode() {
+      Object key = getKey();
+      Object value = getValue();
+      int result = key != null ? key.hashCode() : 0;
+      result = 31 * result + (value != null ? value.hashCode() : 0);
```
karesti: I think you can replace the code by 
``` java
31 * Objects.hashCode(key) + Objects.hashCode(value)
```

pruivo: done!

wburns:

PR #5072, opened 2017-04-11
ISPN-7715 Command can reply when the cache has been shut down

rvansa:
https://issues.jboss.org/browse/ISPN-7715


ryanemerson:
@rvansa Integrated, thanks!

PR #5073, opened 2017-04-11
ISPN-7509 TotalOrderStateTransferInterceptor doesn't handle OutdatedT…

pruivo:
…opologyException for read commands

https://issues.jboss.org/browse/ISPN-7509


rvansa:
Nice to see another random failure resolved. The code looks ok (if I see correctly it's just moving methods to parent and removing unnecessary `throws` declarartion). Do you think it would make sense to write a test for this (or adding a variant with TO configuration)?

pruivo:
I don't think so. I just moved the code and it should be properly tested (and DistTotalOrderVersionedStateTransferTest does the trick :))

rvansa:
Then this is good to go in, IMO. If you want to squeeze that before branching, get someone to integrate it.

ryanemerson:
@pruivo Integrated, thanks!

PR #5077, opened 2017-04-12
ISPN-7144 Cross-Site Replication: inconsistent data with multiple sit…

pruivo:
…e masters

* Force single and same route for all requests

https://issues.jboss.org/browse/ISPN-7144

### DIFF
--- a/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
+++ b/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
@@ -201,6 +201,7 @@ public class JGroupsTransport extends AbstractTransport implements MembershipLis
 
       initChannelAndRPCDispatcher();
       addXSiteViewListener();
+      addSiteMasterPicker();
       startJGroupsChannelIfNeeded();
 
       waitForInitialNodes();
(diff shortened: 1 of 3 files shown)


tristantarrant:
Doesn't this change affect both sync and async behaviour ?

pruivo:
yes :(

tristantarrant:
So effectively this is disabling multiple site masters. Wouldn't it be better then to just discourage its use, via a warning, instead of modifying the behaviour ?

pruivo:
a warning doesn't solve the inconsistencies when they happen :)
but if you think it would be better, I'll change but it won't be 100% accurate because you can have a single site-master in site A (Primary Site) e multiple site-masters in site B (Backup Site) .

galderz:
Is multiple site masters enabled by default? I wasn't aware of it actually, what is used for? Failover? Or to have one master connected to one site and another master to a different site? Couldn't find anything in the docu about multiple site masters

pruivo:
@galderz no, the default is a single site master. It is to improve the performance. if you have a single site master, all the requests sent to the remote site need to go through that node. If you have more than one, it picks one site master randomly. Finally, a single site master can be connected to multiple remote sites.

tristantarrant:
Merged to master

tristantarrant:
Merged to 9.0.x too

PR #5089, opened 2017-04-18
added support for named parameter to the like operator of Ickle Query

kazuhira-r:
Using the named parameter for the like operator of the current Ickle Query throws an exception.

For example, if you run the following code
```java
   @Test
   public void testParamLike() throws Exception {
      QueryFactory qf = getQueryFactory();

      Query q = qf.create("from " + getModelFactory().getTransactionTypeName() + " where description like :descriptionParam");

      q.setParameter("descriptionParam", "%january%");

      List<Transaction> list = q.list();

      assertEquals(1, list.size());
   }
```

A ClassCastException is thrown.
```
java.lang.ClassCastException: org.infinispan.objectfilter.impl.syntax.ConstantValueExpr$ParamPlaceholder cannot be cast to java.lang.String
	at org.infinispan.objectfilter.impl.syntax.parser.QueryRendererDelegateImpl.predicateLike(QueryRendererDelegateImpl.java:308)
	at org.infinispan.objectfilter.impl.ql.parse.QueryRenderer.predicate(QueryRenderer.java:3714)
	at org.infinispan.objectfilter.impl.ql.parse.QueryRenderer.searchCondition(QueryRenderer.java:3007)
	at org.infinispan.objectfilter.impl.ql.parse.QueryRenderer.whereClause(QueryRenderer.java:657)
	at org.infinispan.objectfilter.impl.ql.parse.QueryRenderer.querySpec(QueryRenderer.java:510)
	at org.infinispan.objectfilter.impl.ql.parse.QueryRenderer.queryStatement(QueryRenderer.java:379)
	at org.infinispan.objectfilter.impl.ql.parse.QueryRenderer.queryStatementSet(QueryRenderer.java:292)
	at org.infinispan.objectfilter.impl.ql.parse.QueryRenderer.statement(QueryRenderer.java:220)
	at org.infinispan.objectfilter.impl.ql.QueryParser.render(QueryParser.java:88
```

In the QueryRendererDelegateImpl class, the value of the like operator can only handle String.
https://github.com/infinispan/infinispan/blob/9.0.0.Final/object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/parser/QueryRendererDelegateImpl.java#L307-L310

Also, I think that the reason is that the LikeExpr class can not handle the ConstantValueExpr.ParamPlaceholder class.

I think like this, but how is it?

### DIFF
--- a/object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/LikeExpr.java
+++ b/object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/LikeExpr.java
@@ -1,5 +1,7 @@
 package org.infinispan.objectfilter.impl.syntax;
 
+import java.util.Map;
+
 /**
  * @author anistor@redhat.com
  * @since 7.0
@@ -13,10 +15,10 @@ public final class LikeExpr implements PrimaryPredicateExpr {
    public static final char DEFAULT_ESCAPE_CHARACTER = '\\';
 
    private final ValueExpr child;
-   private final String pattern;
+   private final Object pattern;
    private final char escapeChar = DEFAULT_ESCAPE_CHARACTER;
 
-   public LikeExpr(ValueExpr child, String pattern) {
+   public LikeExpr(ValueExpr child, Object pattern) {
       this.child = child;
       this.pattern = pattern;
    }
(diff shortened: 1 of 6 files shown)


object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/LikeExpr.java
```diff
@@ -26,10 +28,27 @@ public ValueExpr getChild() {
       return child;
    }
 
-   public String getPattern() {
+   public Object getPattern() {
```
anistor: This is never used.

object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/LikeExpr.java
```diff
@@ -26,10 +28,27 @@ public ValueExpr getChild() {
...
       return pattern;
    }
 
+   public String getPatternAs(Map<String, Object> namedParameters) {
```
anistor: I'd rename this to ```getPattern```

anistor:
I've added a jira issue to track this : https://issues.jboss.org/browse/ISPN-7757

anistor:
Integrated in master. Thanks @kazuhira-r !

anistor:
I've also backported this to 9.0.x branch: https://github.com/infinispan/infinispan/pull/5101

kazuhira-r:
Thank you for your review and integration!

PR #5092, opened 2017-04-19
ISPN-7479 Add Failover and Execution Policy to ClusterExecutor

wburns:
* Added execution policy to cluster executor
* Added single cluster executor which allows for failover
* Added non cluster executor which also allows failover ?
* Now support timeout for local thread execution

Marking this as preview for now. I still need to add some more sanity checks on how exceptions are handled.  Also some tests are failing due to exception expectations. To be honest CompletableFutures make exception handling without wrapping with CompletionException a nightmare. Currently I am trying to keep the unwrapped exception propagated to the user. The more I deal with this though I am thinking I may want to always return a wrapped CompletionException except for when a TimeoutException occurs. This would make the code a ton easier as well. But I am not sure if this is how a user would want to use it. But it would also allow for the user to distinguish between their callable throwing a TimeoutException and a timeout due to the `timeout` setting.

Also still need to add some more tests around failover most likely.

Also @danberindei if you know of a way of having a `factory` method while not having to rebuild the cluster for each permutation would be awesome :)


core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
```diff
@@ -1019,10 +1023,28 @@ public ClusterExecutor executor() {
...
+            globalComponentRegistry.wireDependencies(hack);
+            timeoutExecutor = hack.timeoutExecutor;
+         }
+         return ClusterExecutors.allSubmissionExcecutor(null, this, null,
+               TransportConfiguration.DISTRIBUTED_SYNC_TIMEOUT.getDefaultValue(), TimeUnit.MILLISECONDS, ForkJoinPool.commonPool(),
+               timeoutExecutor);
+      }
+   }
+
+   static class Hack {
```
wburns: Going to try to add `ClusterExecutor` global component factory instead which can inject it.

wburns:
Updated.

I have fixed the factory issue so now the various policy tests run substantially faster.

Also I have refactored the various Executors so that CompletionExceptions will no longer wrap other Throwables.

core/src/main/java/org/infinispan/manager/ClusterExecutionPolicy.java
```diff
@@ -0,0 +1,90 @@
...
+    * The command can be executed on any node in the cluster
+    */
+   ALL {
+      @Override
+      public boolean include(TopologyAwareAddress thisAddress, TopologyAwareAddress otherAddress) {
+         return true;
+      }
+   },
+   /**
+    * The command can be executed only on the same machine from where the command was executed.  Note this implies
```
tristantarrant: Cacophonic. Better use something like:
The command can be executed only on the same machine from where it is initiated

core/src/main/java/org/infinispan/manager/ClusterExecutor.java
```diff
@@ -35,11 +36,13 @@
     * {@inheritDoc}
     * <p>
     * This command will be ran in the desired nodes, but no result is returned to notify the user of completion or
-    * failure.
+    * failure. This command will return immediately while the callable is processed asynchronously.
```
tristantarrant: s/callable/runnable/

core/src/main/java/org/infinispan/manager/ClusterExecutor.java
```diff
@@ -98,18 +111,52 @@
...
     */
-   <V> CompletableFuture<Void> submitConsumer(SerializableFunction<? super EmbeddedCacheManager, ? extends V> callable,
-                                              TriConsumer<? super Address, ? super V, ? super Throwable> triConsumer);
+   default <V> CompletableFuture<Void> submitConsumer(SerializableFunction<? super EmbeddedCacheManager, ? extends V> callable,
+                                              TriConsumer<? super Address, ? super V, ? super Throwable> triConsumer) {
+      return submitConsumer((Function<? super EmbeddedCacheManager, ? extends V>) callable, triConsumer);
+   }
 
    /**
+    * Sets a duration of which a remote command will timeout. This will cause the command to return a
```
tristantarrant: s/of which/after which/

core/src/main/java/org/infinispan/manager/ClusterExecutor.java
```diff
@@ -98,18 +111,52 @@
...
-    * @return
+    * @param time the duration for the timeout
+    * @param unit what unit the duration is in
+    * @return this executor with a timeout applied for remote commands
     */
    ClusterExecutor timeout(long time, TimeUnit unit);
 
    /**
+    * When a command is submitted it will only be submitted to one node of the available nodes, there is no strict
+    * requirements as to which node is chosen and is implementation specific. Fail over is not used with, if you desire
```
tristantarrant: The phrase about failover seems incomplete

wburns:
Fixed javadoc and thanks for updating title @tristantarrant. I was going to do it this morning and was confused when it was already changed :)

tristantarrant:
Pushed to master, good stuff @wburns

tristantarrant:
Now we need documentation for this together with deprecation of the old distexec stuff

PR #5094, opened 2017-04-19
Rest easy refactoring

slaskawi:
https://issues.jboss.org/browse/ISPN-7734
https://issues.jboss.org/browse/ISPN-7735
https://issues.jboss.org/browse/ISPN-7736
https://issues.jboss.org/browse/ISPN-7737

This is the fundamental part of removing RESTEasy from our Server code base. The plan is to fully implement it and then "switch" the implementation in Server modules part. This way we will be able to benchmark and compare those two implementation before removing RestEasy. 

Highlights:
* [x] REST functionality
* [x] Server is prepared for both HTTP/1.1 and HTTP/2 (however the latter needs some more tricks on the server side, especially [this one](https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/protocols/ssl/ALPNHackSSLEngine.java))
* [x] Authentication is done
* [x] Authorization is done
* [x] Server modules bits are missing https://issues.jboss.org/browse/ISPN-7738
* [x] Check if removing Optional gives us any benefit https://issues.jboss.org/browse/ISPN-7767
* [x] Remove Security domain and use Callback Handlers for authentication https://issues.jboss.org/browse/ISPN-7778


server/rest/src/main/java/org/infinispan/rest/embedded/netty4/SecurityActions.java
```diff
@@ -17,7 +17,7 @@
  * @since 8.2
  * Temporary fork from RestEasy 3.1.0
  */
-final class SecurityActions {
+public final class SecurityActions {
```
tristantarrant: There is a good reason why these are not public. You will need to copy the package to your own.

slaskawi:
@tristantarrant Updated.

server/rest/src/main/java/org/infinispan/rest/server/InfinispanResponse.java
```diff
@@ -0,0 +1,153 @@
...
+import io.netty.handler.codec.http.DefaultFullHttpResponse;
+import io.netty.handler.codec.http.FullHttpResponse;
+import io.netty.handler.codec.http.HttpHeaderNames;
+import io.netty.handler.codec.http.HttpHeaderValues;
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.HttpUtil;
+import io.netty.handler.codec.http.HttpVersion;
+import io.netty.handler.codec.http2.HttpConversionUtil;
+
+public class InfinispanResponse {
```
wburns: Looking at this class it screams that this should be called `HttpResponseBuilder` and doesn't use `Optional` values and just does null checks like normal. There would be a lot of extra Optionals created and all of the Consumers for said Optionals.

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
```diff
@@ -0,0 +1,303 @@
...
+         String key = request.getKey().orElseThrow(NoKeyException::new);
+         String cacheControl = request.getCacheControl().orElse("");
+         boolean returnBody = request.getRawRequest().method() == HttpMethod.GET;
+         CacheEntry<String, byte[]> entry = restCacheManager.getInternalEntry(cacheName, key);
+         MediaType mediaType = request.getMediaType()
+               .map(m -> MediaType.fromMediaTypeAsString(m))
+               .orElse(MediaType.TEXT_PLAIN);
+
+         InfinispanResponse response = InfinispanResponse.inReplyTo(request);
+         response.status(HttpResponseStatus.NOT_FOUND);
```
wburns: Wonder if the Response should just start as `NOT_FOUND` in InfinispanResponse. That or we can just put it in the 2 branches below where it would be used.

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
```diff
@@ -0,0 +1,303 @@
...
+      try {
+         String cacheName = request.getCacheName();
+         AdvancedCache<String, byte[]> cache = restCacheManager.getCache(cacheName);
+         CacheSet<String> keys = cache.keySet();
+         MediaType mediaType = request.getMediaType()
+               .map(m -> MediaType.fromMediaTypeAsString(m))
+               .orElseThrow(NoMediaTypeException::new);
+
+         InfinispanResponse response = InfinispanResponse.inReplyTo(request);
+         response.contentAsBytes(mediaType.getOutputPrinter().print(cacheName, keys));
```
wburns: I wonder if we could send multiple responses for this? There could be a large number of keys and it would be nice to send some to not also consume a large amount of memory on the server at once too.

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
```diff
@@ -0,0 +1,303 @@
...
+   }
+
+   public InfinispanResponse getCacheValues(InfinispanRequest request) throws RestResponseException {
+      try {
+         String cacheName = request.getCacheName();
+         AdvancedCache<String, byte[]> cache = restCacheManager.getCache(cacheName);
+         CacheSet<String> keys = cache.keySet();
+         MediaType mediaType = request.getMediaType()
+               .map(m -> MediaType.fromMediaTypeAsString(m))
+               .orElseThrow(NoMediaTypeException::new);
```
wburns: Is there a reason that we don't use `TEXT_PLAIN` like the single `getCacheValue` method below does?

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
```diff
@@ -0,0 +1,303 @@
...
+         return Integer.MAX_VALUE;
+      } else {
+         return ((int) (expires.getTime() - new Date().getTime()) / 1000);
+      }
+   }
+
+   private OptionalInt minFresh(String cacheControl) {
+      Optional<String> minFreshDirective = Arrays.stream(cacheControl.split(",")).filter(s -> s.contains("min-fresh")).findFirst();
+      return minFreshDirective.map(s -> {
+         String[] equals = s.split("=");
```
wburns: Seems you could do all this map stuff in the above stream. It is done lazily so it would only be performed on the first match anyways.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/Escaper.java
```diff
@@ -0,0 +1,44 @@
...
+
+   private Escaper() {
+
+   }
+
+   static String escapeHtml(String html) {
+      return escapeXml(html);
+   }
+
+   static String escapeXml(String xml) {
```
wburns: I believe there is a standard JRE class that does this already.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/HTMLOutputPrinter.java
```diff
@@ -0,0 +1,32 @@
...
+import org.infinispan.CacheSet;
+import org.infinispan.rest.server.operations.exceptions.ServerInternalException;
+import org.infinispan.rest.server.operations.mediatypes.OutputPrinter;
+
+public class HTMLOutputPrinter implements OutputPrinter {
+
+   @Override
+   public byte[] print(String cacheName, CacheSet<?> keys) {
+      StringBuilder sb = new StringBuilder("<html><body>");
+      Iterator<?> it = keys.iterator();
```
wburns: You could be fancy and do something like:
`keys.stream().map(b -> Escaper.escapeHtml(b)).map(s -> String.format("<a href=\"%s/%s\">%s</a><br/>", cacheName, s, s)).collect(Collectors.joining("<html><body>", "", "</body></html>");`

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/HTMLOutputPrinter.java
```diff
@@ -0,0 +1,32 @@
...
+      while (it.hasNext()) {
+         String key = Escaper.escapeHtml(it.next().toString());
+         sb.append(String.format("<a href=\"%s/%s\">%s</a><br/>", cacheName, key, key));
```
wburns: Concatenation would be a lot faster here and a little easier to read imo.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/HTMLOutputPrinter.java
```diff
@@ -0,0 +1,32 @@
...
+
+      }
+      sb.append("</body></html>");
+      return sb.toString().getBytes(Charset.forName("UTF-8"));
+   }
+
+   @Override
+   public byte[] print(byte[] value) throws ServerInternalException {
+      StringBuilder sb = new StringBuilder("<html><body>");
```
wburns: You can just use concatenation here (compiler will change it to StringBuilder for you) and it will look nicer.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/HTMLOutputPrinter.java
```diff
@@ -0,0 +1,32 @@
...
+      sb.append(new String(value, Charset.forName("UTF-8")));
```
wburns: Also I wonder if we want to not do UTF-8 encoding just to decode it 2 lines below. It would be a lot faster although uglier to instead just encode the surrounding tags and write the byte[] as is.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/JSONOutputPrinter.java
```diff
@@ -0,0 +1,39 @@
...
+   private static class JsonMapperHolder {
+      public static final ObjectMapper jsonMapper = new ObjectMapper();
+   }
+
+   @Override
+   public byte[] print(String cacheName, CacheSet<?> keys) {
+      StringBuilder sb = new StringBuilder("keys=[");
+      Iterator<?> it = keys.iterator();
+      while (it.hasNext()) {
+         sb.append(String.format("\"%s\"", Escaper.escapeJson(it.next().toString())));
```
wburns: I wouldn't format, just append :)

wburns: I would also suggest using a Joiner or streams as the other one.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/XMLOutputPrinter.java
```diff
@@ -0,0 +1,37 @@
...
+   @Override
+   public byte[] print(String cacheName, CacheSet<?> keys) {
+      StringBuilder sb = new StringBuilder("<?xml version=\"1.0\" encoding=\"UTF-8\"?>")
+            .append(System.lineSeparator())
+            .append(System.lineSeparator())
+            .append("<keys>");
+
+      Iterator<?> it = keys.iterator();
+      while (it.hasNext()) {
+         sb.append(String.format("<key>%s</key>", Escaper.escapeXml(it.next().toString())));
```
wburns: Just use append instead of format

server/core/src/main/java/org/infinispan/server/core/transport/NettyChannelInitializer.java
```diff
@@ -53,7 +56,9 @@ public void initializeChannel(Channel ch) throws Exception {
 
          pipeline.addLast("sni", new SniHandler(domainMappingBuilder.build()));
       }
-      pipeline.addLast("decoder", server.getDecoder());
+      if(decoder != null) {
+         pipeline.addLast("decoder", server.getDecoder());
```
slaskawi: Done

server/core/src/main/java/org/infinispan/server/core/AbstractProtocolServer.java
```diff
@@ -155,6 +139,9 @@ public String getHost() {
    }
 
    public int getPort() {
+      if(transport != null) {
```
slaskawi: Yeah I do.... but I've never done spaces between `if` and `brackets` :) My fingers are faster then IDE :)

slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/CacheControl.java
```diff
@@ -0,0 +1,23 @@
...
+public class CacheControl {
+
+   private String cacheControl;
+
+   private CacheControl(String cacheControl) {
+      this.cacheControl = cacheControl;
+   }
+
+   public static CacheControl noCache() {
+      return new CacheControl("no-cache");
```
slaskawi: Currently it's used only here. But OK, I'll extract it to a constant.

server/rest/src/main/java/org/infinispan/rest/server/InfinispanRequest.java
```diff
@@ -0,0 +1,117 @@
...
+         }
+      }
+      return Optional.empty();
+   }
+
+   public Optional<Long> getTimeToLiveSeconds() {
+      try {
+         return Optional.of(Long.parseLong(request.headers().get("timeToLiveSeconds")));
+      } catch (NumberFormatException nfe) {
+         return Optional.empty();
```
slaskawi: Good idea!

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
slaskawi: You're right! Let me fix it.

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
slaskawi: I think it's not the best idea. The same `InfinispanResponse` class is also used in other HTTP methods. I would leave it as is.

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
```diff
@@ -0,0 +1,303 @@
...
+               }
+
+               response.status(HttpResponseStatus.OK);
+               response.lastModified(lastMod);
+               response.cacheControl(calcCacheControl(expires));
+               response.expires(expires);
+               response.timeToLive(meta.lifespan());
+               response.maxIdle(meta.maxIdle());
+
+               if (request.getExtended().isPresent() && supportsExtendedHeaders(request.getExtended().get())) {
```
slaskawi: Pretty much :) https://github.com/infinispan/infinispan/blob/master/server/rest/src/main/java/org/infinispan/rest/Server.java#L359

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/InfinispanResponse.java
slaskawi: Well I wanted the name of this class to stay away from common names used by HTTP frameworks. Currently on REST Server module there are 16 `Http*Builder` classes from Netty and HTTP Components. Let's stay away from that sort of names. But of course I'm open for any other alternatives.

Those `Optional` fields are very convenient when creating `FullHttpResponse` (see `#toNettyHttpResponse`). I prefer to keep it clean rather than doing non-null checks all around.

BTW, I plan to create a JMH test before replacing existing solution. If I spot any performance problem there, I would be happy to replace those Optional fields.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/Escaper.java
slaskawi: Could you please tell me where I can find it? I know there is something like this in Commons but I'm not sure about JRE.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/HTMLOutputPrinter.java
slaskawi: Okey dokey

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/HTMLOutputPrinter.java
slaskawi: Sure

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/HTMLOutputPrinter.java
slaskawi: Sure, let's give it a go.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/HTMLOutputPrinter.java
slaskawi: done

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/JSONOutputPrinter.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/XMLOutputPrinter.java
slaskawi: Done

slaskawi:
@wburns @tristantarrant Comments addressed.

I also added much more tests and implemented handling of `Serializable` values.

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
wburns: Understandable, we should be able to move the line to the other branches then.

server/rest/src/main/java/org/infinispan/rest/server/CacheControl.java
wburns: Well I meant this `noCache` method would be invoked quite often.

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
wburns: No it is fine, just seems odd to use a query param and not actually provide a value for the param. This may be standard usage, just seems odd to me.

server/rest/src/main/java/org/infinispan/rest/server/CacheControl.java
```diff
@@ -0,0 +1,26 @@
...
+   private static final String MAX_AGE = "max-age";
+
+   private String cacheControl;
+
+   private CacheControl(String cacheControl) {
+      this.cacheControl = cacheControl;
+   }
+
+   public static CacheControl noCache() {
+      return new CacheControl(NO_CACHE);
```
wburns: I meant to cache the actual `CacheControl` object so we don't have to create a new one each time.

server/rest/src/main/java/org/infinispan/rest/server/InfinispanRequest.java
wburns: Seems this update didn't come through?

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/ASCIIOutputPrinter.java
```diff
@@ -0,0 +1,19 @@
...
+import java.nio.charset.Charset;
+
+import org.infinispan.CacheSet;
+import org.infinispan.rest.server.operations.mediatypes.OutputPrinter;
+
+public class ASCIIOutputPrinter implements OutputPrinter {
+
+   @Override
+   public byte[] print(String cacheName, CacheSet<?> keys) {
+      return keys.toString().getBytes(Charset.forName("US-ASCII"));
```
wburns: Hrmm I don't know if we can trust the toString to print the contents of a `CacheSet` in all cases. I would say we should do the iterator or stream ourselves to be 100% sure.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/Escaper.java
wburns: Hrmm I may be thinking of something else :( Don't worry about it then.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/HTMLOutputPrinter.java
```diff
@@ -0,0 +1,28 @@
...
+            .map(b -> Escaper.escapeHtml(b.toString()))
+            .map(s -> "<a href=\"" + cacheName + "/" + s + "\">" + s + "</a><br/>")
+            .collect(Collectors.joining("<html><body>", "", "</body></html>"))
+            .getBytes(Charset.forName("UTF-8"));
+   }
+
+   @Override
+   public byte[] print(byte[] value) throws ServerInternalException {
+      String html = "<html><body>";
+      html += new String(value);
```
wburns: Sorry I don't know if I was specific enough for this. What I was saying is that assuming the `byte[]` is already in UTF-8 we could instead statically store the `<html><body>` and `</body></html>` as byte[] encoded as UTF-8. Then when this method is called you just allocate a new byte[] with size equal to all 3 and just use `System.arrayCopy` to copy the contents over.

Something like:
`byte[] result = new byte[head.length + value.length + tail.length];`
`System.arraycopy(head, 0, head.length, result, 0, head.length);`
`System.arraycopy(value, 0, value.length, result, head.length, value.length);`
`System.arraycopy(tail, 0, tail.length, result, head.length + value.length, tail.length);`
`return result`

Where you have

`private static final head byte[] = "<html><body>".getBytes(Charset.forName("UTF-8");`
`private static final tail byte[] = "</body></html>".getBytes(Charset.forName("UTF-8");`

This way you have no encoding required in an invocation since the value byte[] is already in the correct encoding.

Forgive if I messed up my system.array copy arguments :)

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/JSONOutputPrinter.java
```diff
@@ -0,0 +1,50 @@
...
+
+   private static class JsonMapperHolder {
+      public static final ObjectMapper jsonMapper = new ObjectMapper();
+   }
+
+   @Override
+   public byte[] print(String cacheName, CacheSet<?> keys) {
+      return keys.stream()
+            .map(b -> Escaper.escapeJson(b.toString()))
+            .collect(Collectors.joining("keys=", "", "]"))
```
wburns: The only think you have to watch out for is if this is a backed by a distributed cache. In which case the collector here isn't serializable. You can use `CacheCollectors` to be safe unless you know this won't go remote.

wburns: Also did you mean `keys=[` ?

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/UTF8OutputPrinter.java
```diff
@@ -0,0 +1,19 @@
...
+import java.nio.charset.Charset;
+
+import org.infinispan.CacheSet;
+import org.infinispan.rest.server.operations.mediatypes.OutputPrinter;
+
+public class UTF8OutputPrinter implements OutputPrinter {
+
+   @Override
+   public byte[] print(String cacheName, CacheSet<?> keys) {
+      return keys.toString().getBytes(Charset.forName("UTF-8"));
```
wburns: Same thing I don't know if we can rely on the toString implementation of every `CacheSet` impl to be the way we want.

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/XMLOutputPrinter.java
```diff
@@ -0,0 +1,50 @@
...
+   private static class XStreamholder {
+      public static final XStream XStream = new XStream();
+   }
+
+   @Override
+   public byte[] print(String cacheName, CacheSet<?> keys) {
+      return keys.stream()
+            .map(b -> Escaper.escapeXml(b.toString()))
+            .map(s -> "<key>" + s + "</key>")
+            .collect(Collectors.joining("<?xml version=\"1.0\" encoding=\"UTF-8\"?>", "", "</keys>"))
```
wburns: We need a beginning `<keys>` tag it looks like.

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
slaskawi: It's there but in a different place: https://github.com/infinispan/infinispan/pull/5094/files#diff-c1c4ddf87770141381278e928f0bc5b0R57

server/rest/src/main/java/org/infinispan/rest/server/CacheControl.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/JSONOutputPrinter.java
slaskawi: Both done

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/ASCIIOutputPrinter.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/HTMLOutputPrinter.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/UTF8OutputPrinter.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/XMLOutputPrinter.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/InfinispanRequest.java
slaskawi: It's there but in a different place: https://github.com/infinispan/infinispan/pull/5094/files#diff-c1c4ddf87770141381278e928f0bc5b0R57

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
slaskawi: I believe it's OK where it is. Let me explain - we treat `NOT_FOUND` as a fallback here. This status will be replaced as soon as we get lower in this method. 

If you insist I might add it to an `else` statement for this if: `if (entry instanceof InternalCacheEntry)`. But the result will be exactly the same.

slaskawi:
@wburns Comments addressed. 

I also created a JMH performance comparison and it looks like this:
```
Benchmark                                                                         (configurationClassName)  Mode  Cnt     Score     Error  Units
RestPerfComparisonTest.BenchmarkState.perform1KPuts         org.infinispan.rest.profiling.NewServerHandler  avgt   31  1798.611 ± 103.823  ms/op
RestPerfComparisonTest.BenchmarkState.perform1KPuts         org.infinispan.rest.profiling.OldServerHandler  avgt   31  1922.648 ±  86.682  ms/op
RestPerfComparisonTest.BenchmarkState.perform1KPutsAndGets  org.infinispan.rest.profiling.NewServerHandler  avgt   31  3920.441 ± 307.330  ms/op
RestPerfComparisonTest.BenchmarkState.perform1KPutsAndGets  org.infinispan.rest.profiling.OldServerHandler  avgt   31  4313.393 ± 362.056  ms/op
RestPerfComparisonTest.BenchmarkState.perform1KPuts         org.infinispan.rest.profiling.NewServerHandler    ss   31  1878.441 ± 206.807  ms/op
RestPerfComparisonTest.BenchmarkState.perform1KPuts         org.infinispan.rest.profiling.OldServerHandler    ss   31  1998.803 ± 279.613  ms/op
RestPerfComparisonTest.BenchmarkState.perform1KPutsAndGets  org.infinispan.rest.profiling.NewServerHandler    ss   31  3979.594 ± 464.515  ms/op
RestPerfComparisonTest.BenchmarkState.perform1KPutsAndGets  org.infinispan.rest.profiling.OldServerHandler    ss   31  4098.248 ± 182.687  ms/op
```

It seems that the new version is slightly slower than the old one. This means that I have a date with profiler :)

server/rest/src/main/java/org/infinispan/rest/server/InfinispanResponse.java
danberindei: @slaskawi The question is, do we really need a wrapper for `NettyHttpResponse`? :)

In any case, I'd remove the optionals, because escape analysis is not reliable enough. OTOH it requires the lambdas to be inlined (objects passed as arguments to other methods are allocated on heap), but then it's automatically disabled if the method gets too big because of inlining: http://vanillajava.blogspot.ro/2015/01/java-lambdas-and-low-latency.html. So you might get 0 allocations when you test in JMH, but then lots of allocations if `toNettyHttpResponse` gets inlined into a big method.

danberindei:
@slaskawi Seems to hang the hotrod-client test suite, the Jenkins build filled the disk with a 27G infinispan-infinispan-client-hotrod.log.

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
slaskawi: I think it's possible. But lets leave it for now and stick to refactoring.

server/core/src/main/java/org/infinispan/server/core/transport/NettyChannelInitializer.java
slaskawi: Unfortunately this can not be done :( `server.getDecoder()` creates new instance for each call and it seems the decoder can not be shared between piplines. I created (and assigned to you) [ISPN-7765](https://issues.jboss.org/browse/ISPN-7765).

slaskawi:
@danberindei The failures should be fixed now. Thanks for pointing them out.

server/rest/src/main/java/org/infinispan/rest/server/InfinispanResponse.java
slaskawi: > The question is, do we really need a wrapper for NettyHttpResponse? :)

It separates REST-related code from Netty. Yes, we need it.

>In any case, I'd remove the optionals, because escape analysis is not reliable enough. OTOH it requires the lambdas to be inlined (objects passed as arguments to other methods are allocated on heap), but then it's automatically disabled if the method gets too big because of inlining: http://vanillajava.blogspot.ro/2015/01/java-lambdas-and-low-latency.html. So you might get 0 allocations when you test in JMH, but then lots of allocations if toNettyHttpResponse gets inlined into a big method.

I'm glad you brought it up. I'm a big fan of `Optional`s. They simplify the code by accepting lambdas in `ifPresent(lambda_code)` and they allow chaining `optional.map().map().orThrow()`. They are really useful in `InfinispanResponse#toNettyHttpResponse` method. It is also worth to mention that we are getting a hard lesson from the reality resulting in [nearly 80 NullPointerException bugs in last 2 years](https://issues.jboss.org/issues/?jql=project%20%3D%20ISPN%20AND%20issuetype%20%3D%20Bug%20AND%20text%20~%20%22NullPointerException%22%20AND%20created%20%3E%3D%202015-04-27%20AND%20created%20%3C%3D%202017-04-27). @rvansa also asked about possibility to use it in our APIs. Now we are looking at a very concrete code example, which makes it perfect for discussion.

So far my benchmarks ([based on this with emitting JFR files and analysis via JMC](https://github.com/infinispan/infinispan/pull/5094/files#diff-014cb06f820646b99c23c65a47e0c32eR26)) showed no evidence of hot spots in any `Optional` method (nor any other caused by bad inlining). However you are absolutely correct, the JMH test checks REST interface in isolation. In order to test it properly I created [ISPN-7767](https://issues.jboss.org/browse/ISPN-7767). Unfortunately it needs to be evaluated after I update all the wiring code in the server. I plan to use similar testing procedure, start the server with JFR recording, bang it on REST interface and then check the results. Do you have any hints what I should look for?

Please also note that the blog post you shared was created in January 2015 (so probably based on testing done in Q4 2014). It has been quite a long time and I bet OpenJDK engineers implemented a lot of optimizations since then.

server/rest/src/main/java/org/infinispan/rest/server/Http20RequestHandler.java
```diff
@@ -0,0 +1,79 @@
...
+            response = cacheOperations.getCacheValue(infinispanRequest);
+         } else if (request.method() == HttpMethod.DELETE) {
+            if (!infinispanRequest.getKey().isPresent()) {
+               response = cacheOperations.deleteEntireCache(infinispanRequest);
+            } else {
+               response = cacheOperations.deleteCacheValue(infinispanRequest);
+            }
+         }
+      } catch (RestResponseException responseException) {
+         logger.errorWhileReponsing(responseException);
```
danberindei: s/Reponsing/Responding/

server/rest/src/main/java/org/infinispan/rest/server/InfinispanResponse.java
danberindei: > Yes, we need it.

Why? Do we plan to switch away from Netty soon?

> I'm a big fan of Optionals.

I think it's clear that I'm not a big fan of `Optional` :) To me, you're just moving complexity away from where you're using the value to where you're creating the `Optional`, so you're not really gaining anything. E.g. instead of having `Optional` fields and a separate method that transforms those fields into headers, the setters could set add the headers to the response directly.

>  I bet OpenJDK engineers implemented a lot of optimizations since then.

`-XX:MaxBCEAEstimateSize` still defaults to 150 in the latest JDK8 and JDK9 builds.

slaskawi: @danberindei It's only a matter of taste in my opinion. Either using setters or `Optional`s has the same result in Netty response. However if there are solid numbers behind your words (e.g. tough data from JMH or Flight Recorder) I will remove Optionals :)

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/Escaper.java
alanfx: Would this work? https://docs.oracle.com/javase/8/docs/api/index.html?java/net/URLEncoder.html

danberindei: @alanfx Nope, URl-escaping has completely different rules: `<a href` becomes `%3Ca+href` instead of `&lt;a href`.

server/rest/src/main/java/org/infinispan/rest/server/InfinispanResponse.java
danberindei: @slaskawi You know as well as I do that we only notice these performance problems right before the release :)

Speaking of performance, you mentioned in a comment that the new version is slightly slower in JMH, but I didn't see any follow-up on what was causing the slowdown.

slaskawi: Unfortunately it was solely my fault... I was using `URI.parse().getPath()` for obtaining path parameters instead of using Netty's `FullHttpRequest.path()`. I didn't realize parsing URI takes that long. It was a good lesson.

There are still plenty of places to optimize. But I want to get it right first. Then play with performance.

danberindei: Ok, glad to hear it was something easy!

server/rest/src/main/java/org/infinispan/rest/server/operations/mediatypes/printers/Escaper.java
alanfx: @danberindei right, never mind!

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
```diff
@@ -0,0 +1,241 @@
...
+            }
+         }
+         return response;
+      } catch (CacheException cacheException) {
+         throw new NoCacheFoundException(cacheException.getLocalizedMessage());
+      }
+   }
+
+   public InfinispanResponse deleteEntireCache(InfinispanRequest request) throws RestResponseException {
+      try {
```
tristantarrant: I'd rename this to "clearEntireCache" to make semantic sense

server/rest/src/main/java/org/infinispan/rest/server/Http20RequestHandler.java
```diff
@@ -0,0 +1,79 @@
...
+      InfinispanResponse response = InfinispanResponse.asError(NOT_IMPLEMENTED, "");
+
+      try {
+         if (request.method() == HttpMethod.GET) {
+            if (!infinispanRequest.getKey().isPresent()) {
+               response = cacheOperations.getCacheValues(infinispanRequest);
+            } else {
+               response = cacheOperations.getCacheValue(infinispanRequest);
+            }
+         } else if (request.method() == HttpMethod.POST) {
```
tristantarrant: You also need to handle HttpMethod.PUT

slaskawi:
@tristantarrant oooh righ... git push :)

server/rest/src/main/java/org/infinispan/rest/server/Http20RequestHandler.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/Http20RequestHandler.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/InfinispanRequest.java
```diff
@@ -1,27 +1,30 @@
...
 import io.netty.handler.codec.http.FullHttpRequest;
+import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.QueryStringDecoder;
 import io.netty.handler.codec.http2.HttpConversionUtil;
 
 public class InfinispanRequest {
 
    private final FullHttpRequest request;
    private final Optional<String> streamId;
+   private final ChannelHandlerContext nettyCahnnelContext;
```
tristantarrant: spelling Cahnnel

server/rest/src/main/java/org/infinispan/rest/server/authentication/BasicAuthenticator.java
```diff
@@ -0,0 +1,40 @@
...
+import org.infinispan.rest.server.authentication.exceptions.AuthenticationException;
+import org.jboss.resteasy.plugins.server.embedded.SecurityDomain;
+
+public class BasicAuthenticator implements Authenticator {
+
+   private final SecurityDomain domain;
+   private final String authenticateHeader;
+
+   public BasicAuthenticator(SecurityDomain domain, String realm) {
+      this.domain = domain;
```
tristantarrant: Let us get rid of the worthless RestEasy SecurityDomain stuff and use CallbackHandlers instead.

server/rest/src/main/java/org/infinispan/rest/server/InfinispanRequest.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/server/authentication/BasicAuthenticator.java
slaskawi: Let me deal with that when I will be replacing the existing RESTEasy+Netty solution. Then I will be able to move some bits from endpoint to REST. Created [ISPN-7778](https://issues.jboss.org/browse/ISPN-7778) for it.

server/rest/src/main/java/org/infinispan/rest/server/operations/CacheOperations.java
slaskawi: Done

slaskawi:
@gustavonalle mentioned that we could use `-Dio.netty.leakDetectionLevel=paranoid` for testing.

Holmistr:
run performance tests please
cs-rest

galderz:
Re: using Optional

My general feeling is that the most outward facing APIs should expose Optional, e.g. see Functional Map API. However, internally I'm more keen on not using Optional for a couple of reasons:

* Potential performance reasons and reduced memory consumption. Less garbage generated.
* Optional lacks a lot of high order functions to be able to make it really fun to use. E.g. there's no Optional.stream(), so no map, flatmap, fold/collect...etc.

So I'd avoid using Optional internally.

server/rest/src/main/java/org/infinispan/rest/server/InfinispanResponse.java
galderz: I made a general comment about Optionals in the PR. My general feeling is that any new outward facing APIs should expose it but I'd avoid using it internally.

slaskawi:
run performance tests please
cs-rest

slaskawi:
@galderz This actually a 3rd request to remove those `Optional`s, so I tend to agree on removing them (//cc @danberindei @wburns). 

However I would like use this opportunity to identify what's the performance impact of using `Optional`s. We are guessing this for a pretty long time and we can finally compare some numbers. So my plan is the following:

* Leave it as is (for now)
* Change the wiring and use new Netty implementation
* Run perf tests
* Remove `Optional`s
* Re-reun perf tests

This way we will finally learn, whether `Optional`s have performance impact or not. I hope you guys are OK with that.

slaskawi:
run performance tests please
cs-rest

slaskawi:
Tests below will run on Netty (instead of RestEASY + Netty)

slaskawi:
run performance tests please
cs-rest

Holmistr:
run performance tests please
cs-rest

Holmistr:
run performance tests please
cs-rest

slaskawi:
run performance tests please
cs-rest

slaskawi:
run performance tests please
cs-rest

jmarkos:
run performance tests please
cs-rest

slaskawi:
run performance tests please
cs-rest

slaskawi:
run performance tests please
cs-rest

slaskawi:
run performance tests please
cs-rest

slaskawi:
CI Failures are not related.

Regarding to performance [run1](http://perfrepo.mw.lab.eng.bos.redhat.com/reports/tableComparisonReport/5472), [run2](http://perfrepo.mw.lab.eng.bos.redhat.com/reports/tableComparisonReport/5475):
* The new implementation is much faster with 10 readers
* With 10 writers, it's slightly slower (~5% which is in the error margin, so take it with a grain of salt)
* For 100 readers and writers the new implementation is slightly faster.

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/AbstractRESTClientIT.java
```diff
@@ -91,7 +91,7 @@ public void testBasicOperation() throws Exception {
         assertEquals(0, insert.getEntity().getContentLength());
 
         HttpResponse get = rest.get(fullPathKey, initialXML);
-        assertEquals("application/octet-stream", get.getHeaders("Content-Type")[0].getValue());
+        assertEquals("application/octet-stream;charset=UTF-8", get.getHeaders("Content-Type")[0].getValue());
```
tristantarrant: binary payloads should not have a charset. Why is it adding one ?

server/rest/src/main/java/org/infinispan/rest/Http20RequestHandler.java
```diff
@@ -0,0 +1,124 @@
...
+         }
+      } catch (RestResponseException responseException) {
+         logger.errorWhileReponding(responseException);
+         response = responseException.toResponse(infinispanRequest);
+      }
+
+      sendResonse(ctx, request, response.toNettyHttpResponse());
+   }
+
+   protected void sendResonse(ChannelHandlerContext ctx, FullHttpRequest request, FullHttpResponse response) {
```
tristantarrant: sendResponse

server/rest/src/main/java/org/infinispan/rest/InfinispanRequest.java
```diff
@@ -0,0 +1,262 @@
...
+
+   private InfinispanRequest(FullHttpRequest request, ChannelHandlerContext ctx) {
+      this.request = request;
+      queryStringDecoder = new QueryStringDecoder(request.uri());
+      streamId = Optional.ofNullable(request.headers().get(HttpConversionUtil.ExtensionHeaderNames.STREAM_ID.text()));
+      this.nettyChannelContext = ctx;
+
+      //we are parsing /context/cacheName/key
+      //not that the first one is empty!
+      String[] pathParams = queryStringDecoder.path().split("/");
```
tristantarrant: Possible optimization: do not use split, but just iterate through the string and collect parts as you go.

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/AbstractRESTClientIT.java
slaskawi: Fixed. That was a mistake.

server/rest/src/main/java/org/infinispan/rest/Http20RequestHandler.java
slaskawi: Fixed.

slaskawi:
run performance tests please
cs-rest

slaskawi:
Now it's faster in all scenarios :dash:

ryanemerson:
Integrated, awesome work @slaskawi

PR #5095, opened 2017-04-19
ISPN-7743

wburns:
DistributedStreamIteratorTest.testNodeLeavesWhileIteratingOverContainerCausingRehashToLoseValues
fails randomly

* Don't force mainline to wait for rehash - stream does this

Not waiting for rehash allows the remote thread to return earlier which in turn causes less issues with rehash.


wburns:
This is difficult to test to make sure this fixes the random failures, but should help.

danberindei:
@wburns I see, so you don't have a thread dump :) I was wondering if the reason state transfer takes longer is the distributed stream itself using some remote executor threads and blocking it.

wburns:
@danberindei I can't say. But distributed stream only uses 1 remote thread per node at a time. So unless there is a major bug causing it to use more it seems unlikely to be the cause.

danberindei:
Ok, let's see if this works... integrated!

PR #5098, opened 2017-04-20
ISPN-7704 Fixed full path support in AbstractFileLookup

slaskawi:
https://issues.jboss.org/browse/ISPN-7704

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/AbstractFileLookup.java
+++ b/commons/src/main/java/org/infinispan/commons/util/AbstractFileLookup.java
@@ -62,10 +62,11 @@ public abstract class AbstractFileLookup implements FileLookup {
    public InputStream lookupFileStrict(URI uri, ClassLoader cl) throws FileNotFoundException {
       String scheme = uri.getScheme();
       switch (scheme) {
-         case "jar":
          case "file":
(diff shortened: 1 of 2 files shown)


commons/src/main/java/org/infinispan/commons/util/AbstractFileLookup.java
```diff
@@ -65,7 +65,7 @@ public InputStream lookupFileStrict(URI uri, ClassLoader cl) throws FileNotFound
          case "jar":
          case "file":
             String uriAsString = uri.toString();
-            String fileName = uriAsString.substring(uriAsString.lastIndexOf(":"));
+            String fileName = uriAsString.substring(uriAsString.lastIndexOf(":") + 1);
```
tristantarrant: I'd prefer to treat them differently: the 'jar' scheme has its own special syntax, so I'd avoid attempting to collapse parsing logic.
`
case "file":
    return new FileInputStream(new File(uri.getPath()));
case "jar":
   String path = uri.getPath();
   return new FileInputStream(new File(path.substring(path.indexOf(':')+1)));
`

slaskawi: Sure no, problem!

PR #5100, opened 2017-04-24
ISPN-7756 Rename all pseudo-JPA filtering classes to Ickle

anistor:
Don't be scared by the large number of modified files. These are just renamed classes.

https://issues.jboss.org/browse/ISPN-7756


anistor:
The CI failures are "OK" :)

slaskawi:
hmmmm I'm not sure about this PR. It seems to change user-facing APIs. Since we are releasing a minor version I guess we are not allowed to do it (but we can deprecate old APIs and ask users to migrate to the new ones).

anistor:
These are all implementation details of query/continuous query. No API involved.

tristantarrant:
In it goes. Thanks @anistor

PR #5101, opened 2017-04-24
ISPN-7757 Like operator should support named parameters

anistor:
Backporting of https://issues.jboss.org/browse/ISPN-7757

And some ```@SortableField``` minor fixes from upstream.


anistor:
Merged into 9.0.x.

PR #5105, opened 2017-04-26
ISPN-7761 Caffeine

karesti:
Upgrade caffeine version on serve
Exclude caffeine from shaded infinispan-query all

https://issues.jboss.org/browse/ISPN-7761

### DIFF
--- a/all/embedded-query/pom.xml
+++ b/all/embedded-query/pom.xml
@@ -68,6 +68,7 @@
                            <exclude>org.osgi:org.osgi.compendium:jar:</exclude>
                            <exclude>org.hibernate.javax.persistence:hibernate-jpa-2.1-api:jar:</exclude>
                            <exclude>org.infinispan:infinispan-license:jar:</exclude>
+                           <exclude>com.github.ben-manes.caffeine:caffeine:jar:</exclude>
                         </excludes>
                      </artifactSet>
                      <filters>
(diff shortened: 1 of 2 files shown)


Sanne:
thanks @karesti .

This is to address https://issues.jboss.org/browse/ISPN-7761 ?

karesti:
@Sanne well, I don't think this change fixes the 7761, but I would like to help to fix it so I can understand better the packagings. This change could be parti of the solution

karesti:
I will link the PR to the issue, so we can fix it. I don't know if this can be easily fixed, I should speak with @tristantarrant about it. We can unlink the PR later if needed

Sanne:
@karesti I guessed something like that, but I have the advantage of having chatted with you before.. others might feel a bit lost about the missing context.
Especially for @tristantarrant it would be helpful to provide a bit more context about why this change is being proposed.

Maybe open a new JIRA to track that the wrong version was being included in the server?

slaskawi:
@Sanne @karesti Exactly.. I'm a bit lost here... what is the status of this PR? Is it ready to be integrated?

karesti:
@slaskawi yes sorry, not yet ... !

karesti:
@Sanne @slaskawi I think is ok now. Thank you @galderz for the shade explanation ! 💃

ryanemerson:
Integrated, thanks @karesti

ben-manes:
fyi, I released 2.5.0 this week. That adds support for variable expiration (per-entry settings).

@wburns The implementation uses a [hierarchical timer wheel](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/TimerWheel.java) to allow the operations be amortized O(1) and is performed eagerly during normal maintenance work. It is currently optimized for durations less than 7 days, though more wheels and an overflow heap could be added if needed. It could be (or your own variant) a nicer approach than the O(n) scan done by `ExpirationManager`.

wburns:
@ben-manes Thanks for the update!

I will definitely have to take a look. To be honest our expiration code needs some loving :)

PR #5112, opened 2017-04-28
ISPN-7588 Hot Rod Admin ops

tristantarrant:
This PR implements Hot Rod admin operations. It is composed of three separate parts:

https://issues.jboss.org/browse/ISPN-7771
This implements the common bits, including the server modules configuration and a simplistic Embedded Server Admin Operations Handler.

https://issues.jboss.org/browse/ISPN-7588
Is the implementation of the Hot Rod Admin operation in the server module

https://issues.jboss.org/browse/ISPN-7772
Is the implementation of the Hot Rod Admin operation in the client

https://issues.jboss.org/browse/ISPN-7777
Modifies the JCache remote client to use the new admin ops

What is still missing:
- Until Clustered Configuration State is implemented, any additive changes to the server topology will not propagate the new caches. This will be solved by https://issues.jboss.org/browse/ISPN-7776
- Any support for persisting the configuration changes to the server. This will be tackled with specific implementations in https://issues.jboss.org/browse/ISPN-7774 and https://issues.jboss.org/browse/ISPN-7775

The tracking issue for the all of the bits is https://issues.jboss.org/browse/ISPN-7769


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheManagerAdminImpl.java
```diff
@@ -0,0 +1,52 @@
...
+      createCache(name, template, EnumSet.noneOf(AdminFlag.class));
+   }
+
+   @Override
+   public void createCache(String name, String template, EnumSet<AdminFlag> flags) {
+      Map<String, String> params = new HashMap<>(2);
+      params.put(CACHE_NAME, name);
+      if (template != null) params.put(CACHE_TEMPLATE, template);
+      if (flags != null && !flags.isEmpty()) params.put(FLAGS, flagsToString(flags));
+      operationsFactory.newAdminOperation("CACHE_CREATE", params).execute();
```
galderz: Hmmm, wondering... do you need an admin operation like this? Wondering if you could do this with a deployable server task? IOW, we could ship with several admin deployable tasks and each of them does what we need?

galderz: Not only would it simplify the server, but it'd mean any client (Java, Node.js, C/C++/.Net) could just provide a facade around an execute call with certain parameters.

galderz: So far, the only downside I can see here is granularity at the security level. E.g. if cache admin ops are implemented via remote execution ops, you would not be able to apply separate cache ops authorization compared to remote exec.

tristantarrant: Well, the ADMIN operation is just that: an EXEC op with a different name so that it can be handled differently from a security POV

tristantarrant:
@galderz modified to use server tasks. I've also modified the jcache remote client to use the new admin ops instead of DMR ops through wildfly-controller

server/core/src/main/java/org/infinispan/server/core/admin/embeddedserver/CacheCreateTask.java
```diff
@@ -0,0 +1,55 @@
...
+   }
+
+   @Override
+   public Void call() throws Exception {
+      if (isPersistent())
+         throw new UnsupportedOperationException();
+
+      String name = requireParameter("name");
+      String template = getParameter("template");
+      cacheManager.executor().submitConsumer(localManager -> {
```
galderz: Why is this being done async? I don't see much value since there's a call to `join()` and that will block the task.

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/task/ServerTaskEngine.java
```diff
@@ -19,7 +19,7 @@
  * Date: 1/20/16
  * Time: 12:32 PM
  */
-public class ServerTaskEngine implements TaskEngine {
+public class   ServerTaskEngine implements TaskEngine {
```
galderz: Unnecessary? :)

server/core/src/main/java/org/infinispan/server/core/admin/embeddedserver/CacheCreateTask.java
tristantarrant: Ideally the tasks should really return a CompletableFuture so they can be handled completely async. Until then, we sync to ensure that the operation has been applied and we return a result (null) :)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheManagerAdminImpl.java
```diff
@@ -0,0 +1,63 @@
...
+      operationsFactory.newExecuteOperation("@@cache@remove", Collections.singletonMap(CACHE_NAME, string(name))).execute();
+   }
+
+   @Override
+   public void reindexCache(String name) throws HotRodClientException {
+      operationsFactory.newExecuteOperation("@@cache@reindex", Collections.singletonMap(CACHE_NAME, string(name))).execute();
+   }
+
+   private static byte[] flags(EnumSet<AdminFlag> flags) {
+      StringJoiner sj = new StringJoiner(",");
```
wburns: Looks like you can do `flags.stream().collect(Collectors.joining(","));` instead

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/OperationsFactory.java
```diff
@@ -212,7 +216,8 @@ public SizeOperation newSizeOperation() {
    }
 
    public <T> ExecuteOperation<T> newExecuteOperation(String taskName, Map<String, byte[]> marshalledParams) {
-      return new ExecuteOperation<>(codec, transportFactory, cacheNameBytes, topologyId, flags(), clientIntelligence, taskName, marshalledParams);
+      return new ExecuteOperation<>(codec, transportFactory, cacheNameBytes == null ? DEFAULT_CACHE_NAME_BYTES : cacheNameBytes,
+            topologyId, flags(), clientIntelligence, taskName, marshalledParams);
```
wburns: Can we not assign `cacheNameBytes` to `DEFAULT_CACHE_NAME_BYTES` in the constructor if it was null instead and just always have this variable as non null?

jcache/remote/src/main/java/org/infinispan/jcache/remote/RemoteCacheAccess.java
```diff
@@ -1,287 +0,0 @@
...
-
-import org.infinispan.client.hotrod.RemoteCacheContainer;
-import org.infinispan.commons.logging.LogFactory;
-import org.infinispan.commons.util.TypedProperties;
-import org.infinispan.jcache.remote.logging.Log;
-import org.jboss.as.controller.PathAddress;
-import org.jboss.as.controller.client.ModelControllerClient;
-import org.jboss.dmr.ModelNode;
-
-interface RemoteCacheAccess {
```
wburns: Nice!

server/core/src/main/java/org/infinispan/server/core/admin/AdminOperationsHandler.java
```diff
@@ -0,0 +1,60 @@
...
+   public List<Task> getTasks() {
+      return new ArrayList(tasks.values());
+   }
+
+   @Override
+   public <T> CompletableFuture<T> runTask(String taskName, TaskContext context) {
+      Class<? extends AdminServerTask> taskClass = tasks.get(taskName);
+      AdminServerTask<T> task = Util.getInstance(taskClass);
+      task.setTaskContext(context);
+      return CompletableFuture.supplyAsync(() -> {
```
wburns: Should we be using a dedicated thread for this? Using common fork join pool seems a bit odd.

server/core/src/main/java/org/infinispan/server/core/admin/AdminServerTask.java
```diff
@@ -0,0 +1,66 @@
...
+   public final String getName() {
+      return "@@" + getTaskContextName() + "@" + getTaskOperationName();
+   }
+
+   @Override
+   public final void setTaskContext(TaskContext taskContext) {
+      this.cacheManager = taskContext.getCacheManager();
+      Map<String, byte[]> rawParams = (Map<String, byte[]>) taskContext.getParameters().get();
+      parameters = new HashMap<>(rawParams.size());
+      rawParams.entrySet().stream().forEach(entry -> {
```
wburns: You can use collect here as well. http://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-

server/core/src/main/java/org/infinispan/server/core/admin/embeddedserver/CacheRemoveTask.java
```diff
@@ -0,0 +1,46 @@
...
+      return "remove";
+   }
+
+   @Override
+   public Void call() throws Exception {
+      if (isPersistent())
+         throw new UnsupportedOperationException();
+
+      String name = requireParameter("name");
+      cacheManager.executor().submitConsumer(localManager -> {
```
wburns: I have to ask, but since we are in the server is there a reason we don't have the domain controller propagate these types of things?

tristantarrant: this is the first simplistic implementation. Unfortunately there is no built-in way for a server node (which is handling the request) to talk to its domain controller. My plan is to use the wildfly remoting client from the server to the DC

tristantarrant:
@wburns addressed all items

wburns:
I really can't tell if this PR passed the build or not. It looks like the build that contains these commits failed during a build step, but it also pulled in other unrelated changes. @slaskawi can you help me out here :)

slaskawi:
@wburns Right, this can be confusing, so let me explain. 

Jenkins pick up `refs/<pr>/merge` which is generated by Github. It represents this pull request *merged* into branch it was opened against. So whenever you push something into `master` branch, all `refs/<pr>/merge` need to be recalculated and rebuilt. Jenkins represents this as additional *changes*. So it might be a bit misleading.

So in order to check if there are any new failures, you always need to check test failures on the PR and compare them with recent master build. I wanted to simplify this (and failed) many, many times. If we don't want any additional work with checking against master build, we need to have nice `GREEN` build. This way it's obvious whether or not specific PR added introduced new failures.

ryanemerson:
@tristantarrant Getting warmer, but there are still a few related test failures

tristantarrant:
Zero failures (well, one, but that is not my fault and it is fixed in another PR: https://github.com/infinispan/infinispan/pull/5161 )

ryanemerson:

PR #5114, opened 2017-05-01
Fixed bug with JCache loader

rchigvintsev:
Hello everyone! Two days ago I posted a [question](http://stackoverflow.com/questions/43679623/classcastexception-in-case-of-configured-read-through-cache-with-some-cache-load) on StackOverflow related to this issue. Nobody answered so I decided to create a pull request because it seems like a bug to me.

### Short description of issue
When JCache loader is configured loaded values are put in place of keys in cache. It causes ClassCastException after calling of javax.cache.Cache.removeAll(). Iteration on cache entires is also broken (javax.cache.Cache.Entry.getKey() returns value instead of key).

### DIFF
--- a/jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheLoaderAdapter.java
+++ b/jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheLoaderAdapter.java
@@ -1,9 +1,5 @@
 package org.infinispan.jcache.embedded;
 
-import javax.cache.expiry.Duration;
-import javax.cache.expiry.ExpiryPolicy;
-import javax.cache.integration.CacheLoader;
-
 import org.infinispan.commons.logging.LogFactory;
 import org.infinispan.jcache.Exceptions;
 import org.infinispan.jcache.Expiration;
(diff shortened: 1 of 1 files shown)


tristantarrant:
@galderz this is for you. In the meantime @rchigvintsev you should create a Jira on https://issues.jboss.rog please

rchigvintsev:
@tristantarrant Done: https://issues.jboss.org/browse/ISPN-7783

galderz:
@rchigvintsev Looks good! Nice catch. I'm going to integrate it but before doing that I'll just simply edit the title to include the JIRA id.

galderz:
Integrated the fix, it should be available in next 9.1.0 alpha and 9.0.x releases.

galderz:
@rchigvintsev If you have time, it'd be great if you could add the test case to the JCache testsuite. Do you think you can do that? That way we guarantee we won't drop the ball next time :). Thx!

rchigvintsev:
@galderz Sure, I'll try to write some tests within the following few days.

arkaev:
Please, cherry-pick it to 8.x versions... It's huge memory leak

PR #5116, opened 2017-05-03
ISPN-7781 Add java deserial white list for client

galderz:
https://issues.jboss.org/browse/ISPN-7781

* Added new option to Hot Rod client that enables a list of regular
  expressions to be plugged that define classes that can be deserialized
  using standard Java serialization.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
@@ -199,7 +199,7 @@ public class RemoteCacheManager implements RemoteCacheContainer {
 
       createExecutorService();
 
-      listenerNotifier = ClientListenerNotifier.create(codec, marshaller, transportFactory);
+      listenerNotifier = ClientListenerNotifier.create(codec, marshaller, transportFactory, configuration.serialWhitelist());
       transportFactory.start(codec, configuration, defaultCacheTopologyId, listenerNotifier);
 
       synchronized (cacheName2RemoteCache) {
@@ -321,7 +321,7 @@ public class RemoteCacheManager implements RemoteCacheContainer {
       RemoteCacheImpl<?, ?> remoteCache = remoteCacheHolder.remoteCache;
       OperationsFactory operationsFactory = new OperationsFactory(
               transportFactory, remoteCache.getName(), remoteCacheHolder.forceReturnValue, codec, listenerNotifier,
-            asyncExecutorService, configuration.clientIntelligence());
+            asyncExecutorService, configuration);
       remoteCache.init(marshaller, asyncExecutorService, operationsFactory, configuration.keySizeEstimate(), configuration.valueSizeEstimate());
    }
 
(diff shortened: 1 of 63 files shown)


danberindei:
@galderz What's the use case?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
```diff
@@ -192,7 +192,7 @@ public void start() {
          asyncExecutorService = executorFactory.getExecutor(configuration.asyncExecutorFactory().properties());
       }
 
-      listenerNotifier = ClientListenerNotifier.create(codec, marshaller, transportFactory);
+      listenerNotifier = ClientListenerNotifier.create(codec, marshaller, transportFactory, configuration.serialWhitelist());
```
galderz: What do you propose then?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.java
```diff
@@ -282,6 +285,12 @@ public ConfigurationBuilder maxRetries(int maxRetries) {
    }
 
    @Override
+   public ConfigurationBuilder javaSerialWhiteLists(String... regExs) {
```
galderz: Hmmm, why? It's an array of String that's passed in.

rvansa: array ~= list, so the plurality is already in the word "list".

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
danberindei: How about `serializationWhiteList`?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.java
wburns: Also this method should be named something like `add...` since we aren't overriding existing white list with subsequent calls.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.java
```diff
@@ -405,6 +421,9 @@ public ConfigurationBuilder read(Configuration template) {
       this.valueSizeEstimate = template.valueSizeEstimate();
       this.maxRetries = template.maxRetries();
       this.nearCache.read(template.nearCache());
+      for (String className : template.serialWhitelist())
```
wburns: Any reason this doesn't use `addAll` as well?

galderz:
Addressed concerns and rebased

wburns:
Tests look good to me.

wburns:

PR #5124, opened 2017-05-08
ISPN-7784 BytesObjectInput doesn't implement ObjectInput properly

danberindei:
https://issues.jboss.org/browse/ISPN-7784

Return -1 instead of throwing an exception at the end of the buffer.

### DIFF
--- a/core/src/main/java/org/infinispan/marshall/core/BytesObjectInput.java
+++ b/core/src/main/java/org/infinispan/marshall/core/BytesObjectInput.java
@@ -1,10 +1,11 @@
 package org.infinispan.marshall.core;
 
+import java.io.EOFException;
 import java.io.IOException;
 import java.io.ObjectInput;
 
 /**
- * Array backed, expandable {@link ObjectInput} implementation.
+ * Array backed {@link ObjectInput} implementation.
  *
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/marshall/core/BytesObjectInput.java
```diff
@@ -1,5 +1,6 @@
 package org.infinispan.marshall.core;
```
anistor: The javadoc says 'expandable'. Just wondering what that means...

anistor:
Looks good.

core/src/main/java/org/infinispan/marshall/core/BytesObjectInput.java
danberindei: Maybe it's copied from `BytesObjectOutput`, I'll remove it.

galderz: Probably a copy/paste error from `BytesObjectOutput` which is really expandable :)

galderz:
@danberindei Looks good! Thanks :)

anistor:
Integrated in master. Thanks Dan and Galder!

PR #5125, opened 2017-05-08
ISPN-7808 Upgrade to mockito-core 2.7.21

danberindei:
https://issues.jboss.org/browse/ISPN-7808

* Fix JpaStoreIT failures
* Replace usages of deprecated mockito API
* Replace any(Class) with any() or isNull() where necessary
* Add a Mocks utility class to help when delegation is needed
  but Mockito.spy() doesn't work


tristantarrant:
Pushed to master, thanks @danberindei

PR #5130, opened 2017-05-11
ISPN-7827 Additional Documentation for Cluster Executor

wburns:
* Deprecated Distributed Executor

https://issues.jboss.org/browse/ISPN-7827


tristantarrant:
The docs look good, but I think we should really create a dedicated chapter on "Executing code in the grid" where we look at the executor, tasks, scripts and streams

wburns:
Closing, need to redo the documentation to add a new section.

PR #5132, opened 2017-05-12
ISPN-7823 RemoteCacheManager start/stop async

karesti:
https://issues.jboss.org/browse/ISPN-7823

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
@@ -5,6 +5,7 @@ import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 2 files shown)


commons/src/main/java/org/infinispan/commons/api/AsyncLifecycle.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Async Lifecycle interface that defines the lifecycle of components in a non blocking way
+ *
+ * @author Katia Aresti (<a href="mailto:karesti@redhat.com">karesti@redhat.com</a>)
+ * @since 9.1
+ */
+public interface AsyncLifecycle {
```
rvansa: Why do you create the interface when the lifecycle is managed by the user anyway?

karesti: No, I didn't implement async ping because in the PR I understood that, as the start can be done in a call just after passing false, we just need something simple right now concerning the start method that is blocking. 

I created the interface because I wanted to add an async method interface for start, as we do in cache. But the lifecycle interface is used a lot and I was just providing something small to cover the request as I understood it on Galder's issue.

karesti: Maybe what I did here is too small @rvansa and does not worth the integration. I understood it as a very simple request

rvansa: I think it is worth the integration, I just don't think that we need the interface.

karesti: So I remove the interface and I just leave the methods ?

ryanemerson:
Integrated, thanks @karesti

PR #5135, opened 2017-05-15
ISPN-7837 & ISPN-7839 ClusteredQuery fixes

gustavonalle:
https://issues.jboss.org/browse/ISPN-7837
https://issues.jboss.org/browse/ISPN-7839


gustavonalle:
There was only 1 test failure which is not related

query/src/main/java/org/infinispan/query/clustered/DistributedIterator.java
```diff
@@ -81,11 +81,21 @@ public E next() {
       // fetch and return the value
```
karesti: is the variable fetchSize used ?

query/src/main/java/org/infinispan/query/clustered/ClusteredCacheQueryImpl.java
```diff
@@ -123,11 +123,12 @@ private int getNodeMaxResults() {
       List<QueryResponse> responses = invoker.broadcast(command);
 
       for (QueryResponse queryResponse : responses) {
-         ClusteredTopDocs topDocs = new ClusteredTopDocs(queryResponse.getTopDocs(), queryResponse.getNodeUUID());
-
-         resultSize += queryResponse.getResultSize();
-         topDocs.setNodeAddress(queryResponse.getAddress());
-         topDocsResponses.put(queryResponse.getNodeUUID(), topDocs);
+         if (queryResponse.getResultSize() > 0) {
```
karesti: I would add an isEmpty() method here instead of comparing size

query/src/main/java/org/infinispan/query/clustered/ClusteredCacheQueryImpl.java
```diff
@@ -123,11 +123,12 @@ private int getNodeMaxResults() {
...
+            ClusteredTopDocs topDocs = new ClusteredTopDocs(queryResponse.getTopDocs(), queryResponse.getNodeUUID());
```
karesti: I understand this check is done because the queryResponse.getTopDocs() can be null. 
So when creating a ClusteredTopDocs, we can have a NullPointerException if this check is not done in the DistributedIterator.
What do you think about moving creation code into the QueryResponse so we can throw an assertionError on creation ?

In QueryReponse

``` java
public ClusteredTopDocs toClusteredTopDocs() {
      if(nodeTopDocs == null) throw new IllegalStateException("ClusteredTopDocs can't be created from an empty QueryResponse");
      ClusteredTopDocs clusteredTopDocs = new ClusteredTopDocs(this.getTopDocs(), this.getNodeUUID());
      clusteredTopDocs.setNodeAddress(this.getAddress());
      return clusteredTopDocs;
   }
```

query/src/main/java/org/infinispan/query/clustered/DistributedIterator.java
gustavonalle: Doesn't seem so, I think it should be but that's a matter for another JIRA

query/src/main/java/org/infinispan/query/clustered/ClusteredCacheQueryImpl.java
gustavonalle: What about a ```nonEmpty``` method instead?

query/src/main/java/org/infinispan/query/clustered/DistributedIterator.java
gustavonalle: https://issues.jboss.org/browse/ISPN-7843

query/src/main/java/org/infinispan/query/clustered/ClusteredCacheQueryImpl.java
gustavonalle: could be 👍

gustavonalle:
@karesti updated

query/src/main/java/org/infinispan/query/clustered/DistributedIterator.java
karesti: cool :D

gustavonalle:
Tests are ok

anistor:
I'm having a look too.

gustavonalle:
@anistor any review from you?

query/src/main/java/org/infinispan/query/clustered/QueryResponse.java
```diff
@@ -45,6 +45,17 @@ public QueryResponse(NodeTopDocs nodeTopDocs, UUID nodeUUid, int resultSize) {
       this.resultSize = resultSize;
    }
 
+   public boolean nonEmpty() {
```
anistor: Just a matter of style, my feeling is we do not need this method since ```getResultSize``` can be used as well.

gustavonalle: @anistor both ways are fine to me, I was using ```getResultSize``` and @karesti suggested an extra method. Since this class is used internally only, let's keep it

anistor:
LGTM. merging...

anistor:
Thanks @gustavonalle !

PR #5138, opened 2017-05-17
ISPN-6245 Remote query should be able to work with JBoss marshaling, …

anistor:
…compat mode and hibernate-search annotations

https://issues.jboss.org/browse/ISPN-6245


gustavonalle:
@anistor How would this work with a real server? How will the annotated entities be deployed? Probably we should add an IT showing that

anistor:
This has sat for so long at the bottom of my drawer that I forgot about that aspect not being covered here:). Deployment of entities is part of another PR, for https://issues.jboss.org/browse/ISPN-6334, but my solution for that is still very sketchy so I would not include it in alpha1. I'll try an alternative approach for an IT asap: deploy a custom compat mode marshaller which happens to also come with the entities in the same jar. This will prove remote search works with compat mode, JBMAR and no protobuf at all.

gustavonalle:
@anistor This strategy should work, I tried it in an IT  https://github.com/infinispan/infinispan/blob/master/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/hotrod/HotRodCustomMarshallerIteratorIT.java

gustavonalle:
@anistor merged, thanks, looking forward to the IT and the doc on how to use it 👍

anistor:
The remaining bits are now addressed here: https://github.com/infinispan/infinispan/pull/5233

PR #5140, opened 2017-05-18
ISPN-7803 Functional commands-based AtomicMaps

rvansa:
https://issues.jboss.org/browse/ISPN-7803

When this gets in, I'll file PR to deprecate `DeltaAware` et all: https://github.com/rvansa/infinispan/tree/ISPN-7847


documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
```diff
@@ -7,6 +7,16 @@ The Infinispan community
 
 This guide walks you through the process of upgrading Infinispan.
 
+== Upgrading from 9.0 to 9.1
+
+=== (FineGrained)AtomicMap reimplemented
+
+Infinispan now contains a new implementation of both `AtomicMap` and `FineGrainedAtomicMap`, but the semantics has been preserved. The new implementations don't use `DeltaAware` interface but link:../user_guide/user_guide.html#functional_map_api[Functional API] instead.
+
+There are no changes in needed for `AtomicMap`, but it now supports non-transactional use case as well.
```
tristantarrant: remove 'in'

documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
```diff
@@ -7,6 +7,16 @@ The Infinispan community
 
 This guide walks you through the process of upgrading Infinispan.
 
+== Upgrading from 9.0 to 9.1
+
+=== (FineGrained)AtomicMap reimplemented
+
+Infinispan now contains a new implementation of both `AtomicMap` and `FineGrainedAtomicMap`, but the semantics has been preserved. The new implementations don't use `DeltaAware` interface but link:../user_guide/user_guide.html#functional_map_api[Functional API] instead.
```
tristantarrant: s/has/have/
s/don't/does not/
add a 'the' just before the functional api link

documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
```diff
@@ -7,6 +7,16 @@ The Infinispan community
...
+
+There are no changes in needed for `AtomicMap`, but it now supports non-transactional use case as well.
+
+`FineGrainedAtomicMap` now uses link:../user_guide/user_guide.html#grouping[Grouping API] and therefore you need to link:../user_guide/user_guide.html#how_do_i_use_the_grouping_api[enable groups in configuration]. Also it holds entries as regular cache entries, plus one cache entry for cached key set (the map itself). Therefore the cache size or iteration/streaming results may differ. Note that fine grained atomic maps are still supported only on transactional caches.
```
tristantarrant: add 'the' before the grouping api link
'still supported on transactional caches only'

documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
rvansa: According to http://www.wordreference.com/definition/semantics 'semantics' is used with singular verb, despite it's plural noun.

core/src/main/java/org/infinispan/atomic/MergeOnStore.java
```diff
@@ -0,0 +1,25 @@
...
+ * As the atomicity of load & store into persistence layer cannot be guaranteed, it is recommended to use such values
+ * only with {@link org.infinispan.context.Flag#SKIP_CACHE_LOAD} and
+ * {@link org.infinispan.context.Flag#SKIP_CACHE_STORE} or implementing a custom externalizer that will deal with this.
+ */
+public interface MergeOnStore {
+   /**
+    * @param other Actual value stored in the cache, or <code>null</code> if the entry does not exist.
+    * @return Value that will be stored in the cache, or null if it should be removed.
+    */
+   Object merge(Object other);
```
galderz: No generics?

core/src/main/java/org/infinispan/atomic/impl/AtomicKeySetImpl.java
```diff
@@ -0,0 +1,557 @@
...
+         this.key = key;
+      }
+
+      public static <K> Add readFrom(ObjectInput input) throws IOException, ClassNotFoundException {
+         return new Add((K) input.readObject());
+      }
+
+      @Override
+      public Void apply(EntryView.ReadWriteEntryView<Object, Object> view) {
+         if (view.find().isPresent()) {
```
galderz: Optional.map(f).orElseThrow() ? :)

core/src/main/java/org/infinispan/atomic/impl/TransactionHelper.java
```diff
@@ -0,0 +1,84 @@
...
+               transaction = batchContainer.getBatchTransaction();
+            }
+         }
+         return transaction;
+      } catch (SystemException e) {
+         throw new CacheException("Unable to get transaction", e);
+      }
+   }
+
+   <T> T run(Supplier<T> op) {
```
galderz: This is not implementing the correct pattern of dealing with transaction commit/rollback. [This code](https://github.com/infinispan/infinispan/blob/master/core/src/test/java/org/infinispan/test/TestingUtil.java#L1268) shows the correct pattern.

galderz:
Have you benchmarked this and compared with old method?

core/src/main/java/org/infinispan/atomic/MergeOnStore.java
rvansa: No. `ReadCommittedEntry` won't use any generics anyway. The implementation should expect any type of object stored in the cache. I don't see much value in that.

core/src/main/java/org/infinispan/atomic/impl/AtomicKeySetImpl.java
rvansa: Since you call `view.set()`, that would create another object capturing the `view`.

core/src/main/java/org/infinispan/atomic/impl/TransactionHelper.java
rvansa: Then we should fix `CacheImpl.executeCommandAndCommitIfNeeded` and some code in functional maps as well... this is basically inlined ^.

rvansa:
No, regreattably I haven't had a chance to write a benchmark. It will mostly depend on performance of functional commands in general. However I can't argue that it will change dramatically after ISPN-7590 since I don't use `ReadWriteKeyValueCommand` which propagates the previous value to backup.

rvansa:
Rebased.

core/src/main/java/org/infinispan/atomic/impl/TransactionHelper.java
rvansa: Fixed `TransactionHelper`

galderz:
Thx for replies, all looks good. Anyone else wants to look at this?

rvansa:
@galderz Seems there are no further objections.

core/src/main/java/org/infinispan/atomic/AtomicMap.java
```diff
@@ -2,15 +2,14 @@
...
-import org.infinispan.atomic.impl.AtomicHashMap;
 import org.infinispan.util.concurrent.IsolationLevel;
 
 /**
  * This is a special type of Map geared for use in Infinispan.  AtomicMaps have two major characteristics:
  *
  * <ol>
  *    <li>Atomic locking and isolation over the entire collection</li>
- *    <li>Fine-grained serialization of deltas</li>
+ *    <li>Replication of updates through deltas</li>
```
rvansa: I think that 'delta' is a general term, synonym for a 'change'.

core/src/main/java/org/infinispan/atomic/AtomicMapLookup.java
```diff
@@ -142,6 +129,7 @@
     * @param <MK>  key param of the cache
     */
    public static <MK> void removeAtomicMap(Cache<MK, ?> cache, MK key) {
-      cache.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES).remove(key);
+      // This removes any entry from the cache but includes special handling for fine-grained maps
+      FineGrainedAtomicMapProxyImpl.removeMap((Cache<Object, Object>) cache, key);
```
rvansa: Yes, for anything but fine grained it works just like a regular cache.remove(). Non-fine grained maps are just single entries, so no problem with that.

rvansa:
Rebased (this had conflicts with removal of Metadata parameter from CommitManager).

rvansa:
@galderz @wburns anything left?

galderz:
CI running... let's wait for it

rvansa:
@galderz CI finished...

wburns:
I see there are a bunch of failures with DirectoryProviderModuleMemberRegistrationIT. Is this just a random failure that occurs sometimes? Unfortunately I can't tell as easily with Jenkins, wish I could see the history of a given test.

I will poke around, maybe those are using atomic map?

wburns:
It seems they failed here as well http://ci.infinispan.org/job/Infinispan/job/PR-5183/4/#showFailuresLink So it does seem unrelated, guess we have more failures in the test suite now 👎

wburns:
@rvansa is there anything you wanted to squash/rebase before I merged this, there are quite a few commits :)

rvansa:
@wburns No, please keep the commits as-is. Those are logically structured changes.

wburns:

core/src/test/java/org/infinispan/api/mvcc/repeatable_read/WriteSkewTest.java
```diff
@@ -57,6 +57,7 @@ public void setUp() {
       // The default cache is NOT write skew enabled.
       cacheManager = TestCacheManagerFactory.createCacheManager(configurationBuilder);
       configurationBuilder.locking().isolationLevel(IsolationLevel.REPEATABLE_READ);
+      configurationBuilder.clustering().hash().groups().enabled();
```
danberindei: @rvansa This fails with trace logging enabled, because it forces the component registry to create a `DistributionManagerImpl`, and its `start()` methods throws a NPE if the cache manager is not clustered. Do you really need grouping in this class?

rvansa: `testDontFailOnImmediateRemovalOfAtomicMaps` uses fine grained maps, so grouping is necessary. Please use `localAddress` in the trace log instead, and drop the `getAddress()` method.

commons/src/main/java/org/infinispan/commons/api/functional/Param.java
```diff
@@ -86,4 +86,76 @@ public static PersistenceMode valueOf(int ordinal) {
       }
    }
 
+   /**
+    * Normally the cache has to acquire locks during any write operation to guarantee
+    * its correctness. If the application is sure that no concurrent operation occurs,
+    * it is possible to increase performance by setting this param to {@link #SKIP}.
+    * The result of any operation without locking is undefined under presence of concurrent
+    * writes.
```
danberindei: Missing `@since` tag.

commons/src/main/java/org/infinispan/commons/api/functional/Param.java
```diff
@@ -86,4 +86,76 @@ public static PersistenceMode valueOf(int ordinal) {
...
+         return LOCK;
+      }
+
+      public static LockingMode valueOf(int ordinal) {
+         return CACHED_VALUES[ordinal];
+      }
+   }
+
+   /**
+    * Defines where is the command executed.
```
danberindei: Missing `@since` tag.

core/src/main/java/org/infinispan/atomic/MergeOnStore.java
```diff
@@ -0,0 +1,25 @@
...
+ * The intended use is when executing a command with {@link org.infinispan.context.Flag#SKIP_LOCKING}; in that case
+ * we may have two different instances in context at one moment and the writes may be applied in any order, even
+ * in a different order on different owners. Therefore it's strongly recommended that all operations on such entry
+ * are commutative (order-independent).
+ *
+ * As the atomicity of load & store into persistence layer cannot be guaranteed, it is recommended to use such values
+ * only with {@link org.infinispan.context.Flag#SKIP_CACHE_LOAD} and
+ * {@link org.infinispan.context.Flag#SKIP_CACHE_STORE} or implementing a custom externalizer that will deal with this.
+ */
+public interface MergeOnStore {
```
danberindei: This looks a lot like `Delta` to me... do we really want to expose another public API that does the same thing, instead of using a lambda?

I'm not sure what you mean about the atomicity of load and store into persistence layer, don't we allow using atomic maps with persistence?

Also, missing `@since` tag.

rvansa: `Delta` was supposed to work in general, this contract is more narrow - speaking only about atomic application into DC. It is a way to allow the FGAM use case - modifying an entry without having it locked.

FGAMs with persistence would get broken in the old implementation.

PR #5143, opened 2017-05-22
ISPN-6971 Use the JChannel API without MessageDispatcher

danberindei:
https://issues.jboss.org/browse/ISPN-6971

Initially I wanted to expose startRequest/sendCommand all the way up to `RpcManager`, but I became convinced that they're too low-level, so the new methods in `Transport` are now `invokeCommand/invokeCommandOnAll/invokeCommandStaggered/invokeCommands`.

It's not in this PR, but in the future I would like to add `sendToAll`, to replace `sendToMany(null, ...)`.

Also not in this PR, pushing the `ResponseCollector` to `RpcManager` and using using custom collectors to replace `invokeRemotely(...).thenApply(responsesMap -> ...)`.


core/src/main/java/org/infinispan/remoting/transport/jgroups/CommandAwareRpcDispatcher.java
```diff
@@ -236,12 +223,13 @@ private void reply(org.jgroups.blocks.Response response, Object retVal, Replicab
          }
 
          // Always set the NO_FC flag
-         short flags = (short) (req.getFlags() | REPLY_FLAGS_TO_SET & ~REPLY_FLAGS_TO_CLEAR);
-         Message rsp = req.makeReply().setFlag(flags).setBuffer(rsp_buf);
+         short flags = (short) (requestFlags | REPLY_FLAGS_TO_SET & ~REPLY_FLAGS_TO_CLEAR);
```
rvansa: operator priority!

danberindei:
run performance tests please
library dist

danberindei:
run performance tests please
library dist

core/src/main/java/org/infinispan/remoting/transport/MultiTargetRequest.java
```diff
@@ -0,0 +1,164 @@
...
+      synchronized (responseCollector) {
+         if (missingResponses <= 0) {
+            // The request is already completed.
+            return;
+         }
+         // Don't add more responses to the collector after this
+         this.missingResponses = 0;
+      }
+
+      Collection<Address> targetsWithoutResponses = Arrays.stream(targets)
```
pruivo: suggestion:
```
String targetsWithoutResponses = Arrays.stream(targets)
            .filter(Objects::nonNull)
            .map(Object::toString).collect(Collectors.joining(","));
//or Collectors.joining(",", "{", "}")
```

core/src/main/java/org/infinispan/interceptors/impl/QueueAsyncInvocationStage.java
```diff
@@ -78,14 +78,26 @@ private Object invokeDirectly(InvocationContext ctx, VisitableCommand command, I
    @Override
    public void accept(Object rv, Throwable throwable) {
       // We started with a CompletableFuture, which is now complete.
-      invokeQueuedHandlers(rv, throwable);
+      if (trace)
```
pruivo: IMO, the trace message was better in` invokeQueuedHandlers()`

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
```diff
@@ -141,7 +141,7 @@ public void testBothOwnersSuspected(Method m) throws ExecutionException, Interru
       cache(2).getAdvancedCache().getAsyncInterceptorChain().addInterceptor(new DelayingInterceptor(arrival, release), 0);
 
       Future<Object> future = fork(() -> cache(0).get(key));
-      assertTrue(arrival.await(10, TimeUnit.SECONDS));
+      assertTrue(arrival.await(20, TimeUnit.SECONDS));
```
pruivo: are you hiding an issue?

core/src/main/java/org/infinispan/remoting/transport/ResponseCollector.java
```diff
@@ -0,0 +1,45 @@
...
+    * {@link org.infinispan.remoting.responses.CacheNotFoundResponse}.</p>
+    *
+    * <p>Should return a non-{@code null} result if the request should complete, or {@code null}
+    * if it should wait for more responses.
+    * If the method throws an exception, the request will be completed with that exception.
+    *
+    * If the last response is received and {@code onResponse()} still returns {@code null},
+    * {@link #finish()} will also be called to obtain a result.
+    *
+    * <p>Thread safety: {@code onResponse()} will *not* be called concurrently from multiple threads,
```
pruivo: typo: onResponse => addResponse

core/src/main/java/org/infinispan/remoting/transport/ResponseCollector.java
```diff
@@ -0,0 +1,45 @@
...
+    * and the request will not be completed while {@code onResponse()} is running.</p>
+    */
+   T addResponse(Address sender, Response response);
+
+   /**
+    * Called after {@link #addResponse(Address, Response)} returns {@code null} for the last response.
+    *
+    * <p>If {@code finishRequest()} finishes normally, the request will complete with its return value
+    * (even if {@code null}).
+    * If {@code finishRequest()} throws an exception, the request will complete exceptionally with that exception,
```
pruivo: typo: finishRequest => finish

core/src/main/java/org/infinispan/remoting/transport/ResponseCollector.java
```diff
@@ -0,0 +1,45 @@
...
+    *
+    * <p>Thread safety: {@code onResponse()} will *not* be called concurrently from multiple threads,
+    * and the request will not be completed while {@code onResponse()} is running.</p>
+    */
+   T addResponse(Address sender, Response response);
+
+   /**
+    * Called after {@link #addResponse(Address, Response)} returns {@code null} for the last response.
+    *
+    * <p>If {@code finishRequest()} finishes normally, the request will complete with its return value
```
pruivo: typo: finishRequest => finish

core/src/test/java/org/infinispan/manager/SingleClusterExecutorTest.java
```diff
@@ -5,6 +5,7 @@
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
```
pruivo: unused

core/src/test/resources/stacks/tcp.xml
```diff
@@ -2,7 +2,7 @@
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/jgroups-4.0.xsd">
    <!-- Disable the regular thread pool queue for tests so we can set a lower min_threads -->
-   <TCP_NIO2
+   <TCP
```
pruivo: no more NIO2?

core/src/main/java/org/infinispan/remoting/transport/Transport.java
```diff
@@ -178,4 +181,23 @@
     */
    Set<String> getSitesView();
 
+   <T> CompletionStage<T> invokeCommand(Address target, ReplicableCommand command,
```
pruivo: with this commands, I was expecting the remaining `invokeRemotely*` to be deprecated. What is the plan?

core/src/main/java/org/infinispan/remoting/transport/ValidSingleResponseCollector.java
```diff
@@ -0,0 +1,46 @@
...
+import org.infinispan.remoting.responses.Response;
+import org.infinispan.remoting.responses.ValidResponse;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+
+/**
+ * @author Dan Berindei
+ * @since 9.1
+ */
+public abstract class ValidSingleResponseCollector<T> implements ResponseCollector<T> {
```
pruivo: I'm not sure if this class is really useful since all subclasses could use ValidResponseCollector to achieve the same logic.

core/src/main/java/org/infinispan/remoting/transport/AbstractRequest.java
```diff
@@ -0,0 +1,113 @@
...
+ * A remote invocation request.
+ *
+ * <p>Thread-safety: This class and its sub-classes are thread-safe. They use the {@code ResponseCollector}'s monitor
+ * for synchronization, so that collectors usually don't need any explicit synchronization.</p>
+ *
+ * @author Dan Berindei
+ * @since 9.1
+ */
+public abstract class AbstractRequest<T> extends CompletableFuture<T> implements Callable<Void>, Request<T> {
+   protected final ResponseCollector<T> responseCollector;
```
danberindei: I'd keep it here, mostly for symmetry: if I put it in the subclasses it would be private in SingleTargetRequest, and protected in MultiTargetRequest.

core/src/main/java/org/infinispan/remoting/transport/AbstractRequest.java
```diff
@@ -0,0 +1,113 @@
...
+
+   /**
+    * Called when the node received a new cluster view.
+    */
+   public abstract void onNewView(Set<Address> members);
+
+   /**
+    * Called when the timeout task scheduled with {@link #setTimeout(ScheduledExecutorService, long, TimeUnit)} expires.
+    */
+   public abstract void onTimeout();
```
danberindei: Done

core/src/main/java/org/infinispan/remoting/transport/AbstractTransport.java
```diff
@@ -17,6 +17,7 @@
  *
  * @author Manik Surtani
  * @version 4.2
+ * @deprecated Since 9.1, please implement {@link Transport} directly.
  */
 public abstract class AbstractTransport implements Transport {
```
danberindei: Done

core/src/main/java/org/infinispan/remoting/transport/MultiTargetRequest.java
```diff
@@ -0,0 +1,164 @@
...
+public class MultiTargetRequest<T> extends AbstractRequest<T> {
+   @GuardedBy("responseCollector")
+   private final Address[] targets;
+   @GuardedBy("responseCollector")
+   private int missingResponses;
+
+   public MultiTargetRequest(ResponseCollector<T> responseCollector, long requestId, RequestRepository repository,
+                             Collection<Address> targets, Address excluded) {
+      super(requestId, responseCollector, repository);
+      this.targets = new Address[targets.size()];
```
danberindei: For now I'm avoiding streams unless the alternative is particularly painful :)

core/src/main/java/org/infinispan/remoting/transport/MultiTargetRequest.java
```diff
@@ -0,0 +1,164 @@
...
+         if (excluded == null || !excluded.equals(target)) {
+            this.targets[i++] = target;
+         }
+      }
+      this.missingResponses = i;
+      if (missingResponses == 0)
+         complete(responseCollector.finish());
+   }
+
+   public int getMissingResponsesCount() {
```
danberindei: Removed

core/src/main/java/org/infinispan/remoting/transport/MultiTargetRequest.java
danberindei: Done... this is a leftover from the time the `ResponseCollector` was also dealing with timeouts.

core/src/main/java/org/infinispan/remoting/transport/PassthroughMapResponseCollector.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Receive responses from multiple nodes, without checking that the responses are valid.
+ *
+ * @author Dan Berindei
+ * @since 9.1
+ */
+public class PassthroughMapResponseCollector implements ResponseCollector<Map<Address, Response>> {
+   private Map<Address, Response> map;
+
+   public PassthroughMapResponseCollector(List<Address> targets) {
```
danberindei: Done

core/src/test/java/org/infinispan/remoting/AsynchronousInvocationTest.java
```diff
@@ -154,47 +150,21 @@ public void testNonCacheRpcCommands() throws Exception {
...
-      Message oobRequest = serialize(command, true, address);
-      if (oobRequest == null) {
-         log.debugf("Don't test " + command.getClass() + ". it is not Serializable");
-         return;
-      }
       executorService.reset();
       CountDownLatchResponse response = new CountDownLatchResponse();
-      commandAwareRpcDispatcher.handle(oobRequest, response);
+      invocationHandler.handleFromCluster(address, command, response, DeliverOrder.NONE);
+      transport.receiveRequest(address, command, Request.NO_REQUEST_ID, DeliverOrder.NONE);
```
danberindei: Done

core/src/main/java/org/infinispan/interceptors/impl/AsyncInterceptorChainImpl.java
```diff
@@ -230,6 +231,7 @@ public void appendInterceptor(AsyncInterceptor ci, boolean isCustom) {
 
    @Override
    public CompletableFuture<Object> invokeAsync(InvocationContext ctx, VisitableCommand command) {
+      Stages.pushInvocationIdNDC(command, trace);
```
danberindei: Ok, I removed the logging changes :)

I added the NDC because I was trying to figure out a timeout error in a query test, and following the handling of a write command that triggers 10s of writes in the index cache is pretty painful. In the end it turned out to be just trace logging slowing things down, but I still want to add this in another PR.

core/src/main/java/org/infinispan/remoting/transport/jgroups/ClusterView.java
```diff
@@ -0,0 +1,61 @@
...
+import org.infinispan.commons.util.Immutables;
+import org.infinispan.remoting.transport.Address;
+
+public class ClusterView {
+   static final int INITIAL_VIEW_ID = -1;
+   static final int FINAL_VIEW_ID = Integer.MAX_VALUE;
+
+   private final int viewId;
+   private final List<Address> members;
+   private final Set<Address> membersSet;
```
danberindei: We expose the members list via `Transport.getMembers()`, so we need a `List`. I could also make `Request.onNewView()` take a `ClusterView` parameter instead of a `Set<Address>`, but I'm not sure it's worth it.

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
```diff
@@ -255,13 +256,13 @@ public boolean containsKey(Object k) {
 
    @Override
    public InternalCacheEntry<K, V> remove(Object k) {
-      final InternalCacheEntry<K,V>[] reference = new InternalCacheEntry[1];
+      final ByRef<InternalCacheEntry<K,V>> reference = new ByRef<>(null);
```
danberindei: Removed

core/src/main/java/org/infinispan/remoting/transport/jgroups/FilterMapResponseCollector.java
```diff
@@ -0,0 +1,65 @@
...
+import org.infinispan.remoting.transport.Address;
+import org.infinispan.remoting.transport.ValidResponseCollector;
+
+/**
+ * Response collector supporting {@link JGroupsTransport#invokeRemotelyAsync(Collection, ReplicableCommand, ResponseMode, long, ResponseFilter, DeliverOrder, boolean)}.
+ *
+ * <p>This class is not thread-safe by itself. It expects an {@link org.infinispan.remoting.transport.AbstractRequest}
+ * to handle synchronization.</p>
+ */
+public class FilterMapResponseCollector extends ValidResponseCollector<Map<Address, Response>> {
```
danberindei: I do want to remove the `ResponseFilter`, but I'm not sure `ResponseCollector` is stable enough yet. So I'm thinking of marking the `ResponseCollector`-based methods as experimental in 9.1 and only deprecating the `ResponseFilter` ones in 9.2/10.0.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddress.java
```diff
@@ -77,7 +77,7 @@ public void doWriteObject(ObjectOutput output, JGroupsAddress address) throws IO
       public JGroupsAddress doReadObject(ObjectInput unmarshaller) throws IOException, ClassNotFoundException {
          try {
             org.jgroups.Address address = org.jgroups.util.Util.readAddress(unmarshaller);
-            return (JGroupsAddress) JGroupsAddressCache.fromJGroupsAddress(address);
+            return (JGroupsAddress) JGroupsAddressCache.fromJGroupsAddressUnchecked(address);
```
danberindei: It is needed, at least one integration test uses Infinispan in local mode without the JGroups jar and fails to start without this change.

Of course, I could have kept `JGroupsAddressCache.fromJGroupsAddress` with an `Object` parameter, but that caused me lots of trouble when I was moving stuff around, so I'd rather keep it.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -103,251 +114,335 @@
...
-
-   @Override
-   public Log getLog() {
-      return log;
+      probeHandler = new ThreadPoolProbeHandler();
    }
 
    // ------------------------------------------------------------------------------------------------------------------
-   // Lifecycle and setup stuff
+   // querying cluster status
```
danberindei: Removed, I used "rearrange code" and missed some of these section comments.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddress.java
pruivo: are you sure? we are invoking `org.jgroups.util.Util.readAddress(unmarshaller);` before and that would need the `jgroups.jar`

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
    }
 
-   @Override
-   public Address getCoordinator() {
-      return coordinator;
+   private static List<Address> fromJGroupsAddressList(List<org.jgroups.Address> list) {
+      if (list == null || list.isEmpty())
+         return Collections.emptyList();
+
+      List<Address> retval = new ArrayList<>(list.size());
```
danberindei: The listeners are invoked without copying the subgroups lists, so I'd rather keep it immutable.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
-      switch (mode) {
-         case ASYNCHRONOUS:
-            return org.jgroups.blocks.ResponseMode.GET_NONE;
-         case WAIT_FOR_VALID_RESPONSE:
-            return org.jgroups.blocks.ResponseMode.GET_FIRST;
-         case SYNCHRONOUS:
-         case SYNCHRONOUS_IGNORE_LEAVERS:
-            return org.jgroups.blocks.ResponseMode.GET_ALL;
+   private void addRequestHeader(Message message, long requestId) {
+      // TODO Remove the header and store the request id in the buffer
```
danberindei: Mostly to avoid allocating a header instance and reserving a header id.

But you're right, creating our own header is also an option, please comment on https://issues.jboss.org/browse/ISPN-7933 with your arguments for it.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
-      if (rsp.wasReceived()) {
-         if (rsp.hasException()) {
-            log.tracef(rsp.getException(), "Unexpected exception from %s", sender);
-            throw log.remoteException(sender, rsp.getException());
-         } else {
-            response = checkResponse(rsp.getValue(), sender, ignoreLeavers);
-         }
-      } else if (rsp.wasSuspected()) {
-         response = checkResponse(CacheNotFoundResponse.INSTANCE, sender, ignoreLeavers);
+   private boolean shouldBroadcast(Collection<Address> recipients, int membersSize, boolean canSwitchProtocols) {
```
danberindei: I've added some more comments instead :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
-         response = null;
       }
+      return singleTarget;
+   }
 
-      return response;
+   private CompletableFuture<Map<Address, Response>> performAsyncRemoteInvocation(Collection<Address> recipients,
+                                                                                  ReplicableCommand command,
+                                                                                  DeliverOrder deliverOrder,
+                                                                                  boolean totalOrder, boolean rsvp,
```
danberindei: Done

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
    }
 
-   @Override
-   public Set<String> getSitesView() {
-      RELAY2 relay = channel.getProtocolStack().findProtocol(RELAY2.class);
-      RouteStatusListener listener = relay != null ?
-            relay.getRouteStatusListener() : null;
-      return (listener instanceof Supplier) ?
-            ((Supplier<Set<String>>) listener).get() : null;
+   public void sendToAll(ReplicableCommand command, DeliverOrder deliverOrder)
```
danberindei: I'm thinking of adding it to the `Transport` interface, as it's more discoverable than `sendToMany(null)`. WDYT?

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
-            n = new NotifyMerge();
+   public void receiveRequest(Address sender, ReplicableCommand command, long requestId,
+                              DeliverOrder deliverOrder) {
+      try {
+         if (command == null)
+            throw new NullPointerException("Unable to execute a null command!  Message was from " +
+                                                 sender);
+         Reply reply = requestId != Request.NO_REQUEST_ID ? returnValue -> reply(returnValue, requestId, sender,
+                                                                                 command) : Reply.NO_OP;
+         org.jgroups.Address src = ((JGroupsAddress) sender).getJGroupsAddress();
```
danberindei: Done

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
+         reply(new ExceptionResponse(new CacheException("Problems invoking command.", x)), requestId, sender, command);
       }
-
-      JGroupsAddressCache.pruneAddressCache();
    }
 
-   @Override
-   public void suspect(org.jgroups.Address suspected_mbr) {
-      // no-op
+   private void reply(Response response, long requestId, Address originator, ReplicableCommand command) {
```
danberindei: Removed

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
-      // no-op since ISPN-83 has been resolved
+   private void executeCommandFromLocalCluster(final ReplicableCommand cmd, final org.jgroups.Address src,
+                                               final DeliverOrder order, final Reply reply) {
+      invocationHandler.handleFromCluster(fromJGroupsAddress(src), cmd, reply, order);
    }
 
-   // ------------------------------------------------------------------------------------------------------------------
-   // Helpers to convert between Address types
-   // ------------------------------------------------------------------------------------------------------------------
+   private void sendResponse(Address originator, Response response, long requestId, ReplicableCommand command) {
```
danberindei: Done

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
+                                             final DeliverOrder order, final Reply reply) {
+      SiteAddress siteAddress = (SiteAddress) src;
+      ((XSiteReplicateCommand) cmd).setOriginSite(siteAddress.getSite());
+      invocationHandler.handleFromRemoteSite(siteAddress.getSite(), (XSiteReplicateCommand) cmd, reply, order);
    }
 
-   @Override
-   public void unblock() {
-      // no-op since ISPN-83 has been resolved
+   private void executeCommandFromLocalCluster(final ReplicableCommand cmd, final org.jgroups.Address src,
```
danberindei: Removed.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
+         if (trace) {
+            if (requestId != Request.NO_REQUEST_ID) {
+               log.tracef("Received request %d@%s: %s", requestId, sender, command);
+            } else {
+               log.tracef("Received command from %s: %s", sender, command);
+            }
+         }
+         receiveRequest(sender, command, requestId, deliverOrder);
+      } catch (Throwable t) {
+         log.errorf(t, "Error processing request %d@%s", requestId, src);
```
danberindei: Done

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
+            response = (Response) marshaller.objectFromByteBuffer(buffer, offset, length);
+            if (response == null) {
+               response = SuccessfulResponse.SUCCESSFUL_EMPTY_RESPONSE;
+            }
          }
+         if (trace)
+            log.tracef("Received response for request %d@%s from %s: %s", requestId, getAddress(), sender, response);
+         receiveResponse(requestId, sender, response);
+      } catch (Throwable t) {
+         log.errorf(t, "Error processing response for request %d@%s from %s", getAddress(), src);
```
danberindei: Done

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
+         } else {
+            type = SINGLE_MESSAGE;
+            requestId = Request.NO_REQUEST_ID;
+         }
+         Address sender = fromJGroupsAddress(src);
+         if (address == null) {
+            if (trace)
+               log.tracef("Ignoring message received before initial view");
+            if (type == REQUEST) {
+               sendResponse(fromJGroupsAddress(message.src()), CacheNotFoundResponse.INSTANCE, requestId, null);
```
danberindei: Done

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
+            }
+            return;
+         }
+         switch (type) {
+            case SINGLE_MESSAGE:
+               processSingleMessage(src, flags, buffer, offset, length, requestId, sender);
```
danberindei: Done

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +626,845 @@ private void buildChannel() {
...
+               break;
+            case REQUEST:
+               processRequest(src, flags, buffer, offset, length, requestId, sender);
+               break;
+            case RESPONSE:
+               processResponse(src, buffer, offset, length, requestId, sender);
+               break;
+            default:
+               log.error("Invalid message type: " + type + " from " + src);
```
danberindei: Done

core/src/main/java/org/infinispan/executors/LimitedExecutor.java
```diff
@@ -140,12 +140,12 @@ private void runTasks() {
          }
 
          try {
-            NDC.push(name);
+            LogFactory.pushNDC(name, true);
             runnable.run();
          } catch (Throwable t) {
             log.error("Exception in task", t);
```
danberindei: Postponed

core/src/main/java/org/infinispan/interceptors/impl/QueueAsyncInvocationStage.java
danberindei: Postponed

core/src/test/java/org/infinispan/distribution/RemoteGetFailureTest.java
danberindei: Debugging leftovers :)

core/src/main/java/org/infinispan/remoting/transport/ResponseCollector.java
danberindei: Fixed

core/src/main/java/org/infinispan/remoting/transport/ResponseCollector.java
danberindei: Fixed

core/src/main/java/org/infinispan/remoting/transport/ResponseCollector.java
danberindei: Fixed

core/src/test/resources/stacks/tcp.xml
danberindei: Debugging leftovers again :)

core/src/main/java/org/infinispan/remoting/transport/ValidSingleResponseCollector.java
danberindei: The difference is not having to implement `finish()`.

danberindei:
Answered comments, fixed x-site tests.

core/src/main/java/org/infinispan/remoting/transport/MultiTargetRequest.java
pruivo: what is wrong with streams?

core/src/test/java/org/infinispan/remoting/AsynchronousInvocationTest.java
```diff
@@ -59,9 +59,9 @@
 
    private EmbeddedCacheManager cacheManager;
    private DummyTaskCountExecutorService executorService;
-   private CommandAwareRpcDispatcher commandAwareRpcDispatcher;
+   private JGroupsTransport transport;
```
pruivo: the `transport` field can be removed.

core/src/main/java/org/infinispan/remoting/transport/jgroups/ClusterView.java
pruivo: the `List` can be removed in the future and replaced by a `Collection`. Most of our use case is to iterate or find live members. I only find a couple of uses cases where we fetch by index. I still think it is beneficial but I'm ok if you don't wanna do it now.

core/src/main/java/org/infinispan/remoting/transport/jgroups/FilterMapResponseCollector.java
pruivo: :+1: for the filter removal

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddress.java
```diff
@@ -77,7 +77,8 @@ public void doWriteObject(ObjectOutput output, JGroupsAddress address) throws IO
       public JGroupsAddress doReadObject(ObjectInput unmarshaller) throws IOException, ClassNotFoundException {
          try {
             org.jgroups.Address address = org.jgroups.util.Util.readAddress(unmarshaller);
-            return (JGroupsAddress) JGroupsAddressCache.fromJGroupsAddress(address);
+            // Use the unchecked version to avoid loading org.jgroups.Address when the externalizer is instantiated
```
pruivo: I don't know what kind of dark magic is this but you are loading the `Address` when you invoke the `readAddress()`

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
pruivo: :+1:

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +619,784 @@ private void buildChannel() {
...
+   private static List<Address> fromJGroupsAddressList(List<org.jgroups.Address> list) {
+      return Collections.unmodifiableList(list.stream()
+                                              .map(JGroupsAddressCache::fromJGroupsAddress)
+                                              .collect(Collectors.toList()));
    }
 
-   private void waitForInitialNodes() {
-      int initialClusterSize = configuration.transport().initialClusterSize();
-      if (initialClusterSize <= 1)
+   public void receiveClusterView(long viewId, List<Address> members, List<List<Address>> subgroups) {
```
pruivo: can be private IMO

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -466,700 +619,784 @@ private void buildChannel() {
...
 
-      if (remainingNanos <= 0) {
-         throw log.timeoutWaitingForInitialNodes(initialClusterSize, channel.getView().getMembers());
-      }
-
-      log.debugf("Initial cluster size of %d nodes reached", initialClusterSize);
-   }
+         // The first view is installed before returning from JChannel.connect
+         // So we need to set the local address here
+         if (address == null) {
```
pruivo: the other `receiveClusterView()` already sets the address before invoking this method.

core/src/main/java/org/infinispan/remoting/transport/ValidSingleResponseCollector.java
pruivo: and is it worth it?

core/src/main/java/org/infinispan/remoting/transport/Transport.java
```diff
@@ -178,4 +181,23 @@
...
+                                        ResponseCollector<T> collector, DeliverOrder deliverOrder,
+                                        long timeout, TimeUnit unit);
+
+   <T> CompletionStage<T> invokeCommand(Collection<Address> targets, ReplicableCommand command,
```
pruivo: shouldn't be `CompletableFuture`? 
I think the `RpcManager` and the `AsyncInterceptor` use the `CompletableFuture`

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddress.java
danberindei: We do need the jar if the externalizer ever gets executed, but not if it's only instantiated.

core/src/main/java/org/infinispan/remoting/transport/ValidSingleResponseCollector.java
danberindei: I'm not sure yet, but I'm hoping I can implement some of them as lambdas this way :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddress.java
pruivo: exactly. If you only instantiate this class, the old code works fine and this change isn't needed :)

danberindei: In theory, you're right. In practice, if you replace `fromJGroupsAddressUnchecked` with `fromJGroupsAddress`, you'll get an exception like this in the `LdapAuthenticationIT`:

```
org.infinispan.commons.CacheException: Unable to invoke method public void org.infinispan.marshall.core.GlobalMarshaller.start() on object of type GlobalMarshaller
Caused by: org.infinispan.commons.CacheException: Unable to invoke method public void org.infinispan.marshall.core.GlobalMarshaller.start() on object of type GlobalMarshaller
Caused by: java.lang.NoClassDefFoundError: org/jgroups/Address
	at org.infinispan.marshall.core.InternalExternalizers.load(InternalExternalizers.java:180)
	at org.infinispan.marshall.core.GlobalMarshaller.start(GlobalMarshaller.java:125)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.infinispan.commons.util.SecurityActions.lambda$invokeAccessibly$0(SecurityActions.java:91)
	... 89 more
Caused by: java.lang.ClassNotFoundException: org.jgroups.Address from [Module "deployment.a9b4f229-4004-47a6-9113-e651571940af.war:main" from Service Module Loader]
	at org.jboss.modules.ModuleClassLoader.findClass(ModuleClassLoader.java:198)
	at org.jboss.modules.ConcurrentClassLoader.performLoadClassUnchecked(ConcurrentClassLoader.java:363)
	at org.jboss.modules.ConcurrentClassLoader.performLoadClass(ConcurrentClassLoader.java:351)
	at org.jboss.modules.ConcurrentClassLoader.loadClass(ConcurrentClassLoader.java:93)
	... 96 more
```

No idea why, but the method call does trigger the loading of `org.jgroups.Address`.

danberindei: Ok, I think I finally figured out what's happening. 

`JGroupsAddress.Externalizer` call was actually ok, only `JGroupsTopologyAwareAddress.Externalizer` tried to load `org.jgroups.Address` during instantiation. The reason: an implicit cast from `ExtendedUUID` to `org.jgroups.Address`. Once I changed the local variable to be a `org.jgroups.Address` as well, `LdapAuthenticationIT` stopped failing.

I updated the comments and I removed the `JGroupsAddressCache.fromJGroupsAddressUnchecked` method.

core/src/main/java/org/infinispan/remoting/transport/jgroups/ClusterView.java
danberindei: `Transport.getMembers()` is part of the API, and applications may rely in `Transport.getMembers().get(0)` to return the coordinator. We may want to change that in the future, but definitely not in a minor version...

core/src/main/java/org/infinispan/remoting/transport/MultiTargetRequest.java
danberindei: 1. I never know how many objects will be instantiated.
2. Formatting streams-based code on multiple lines is weird, so most of the code I see gives up and puts everything one one line, making it harder to read than the equivalent imperative code.

core/src/main/java/org/infinispan/remoting/transport/Transport.java
danberindei: Actually I'd want the ResponseCollector-based RpcManager API to use a `CompletionStage` as well.

For `BaseAsyncInterceptor.asyncValue()` yes, it was more convenient to assume a `CompletableFuture`, but a) there's always `CompletionStage.toCompletableFuture()` and b) we may want to use our own `CompletionStage` implementation in the future, and `asyncValue()` could be extended to deal with that implementation.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
danberindei: Inlined, and made the other overload protected for `InfinispanNodeFailureTest`.

core/src/main/java/org/infinispan/remoting/transport/MultiTargetRequest.java
pruivo: 1. from `toArray()` javadoc
```
The generator function takes an integer, which is the size of the desired array, and produces an array of the desired size.
```
2. this case is small enough for a single line.

... but ok. I won't insist

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddress.java
pruivo: well, it doesn't make sense. it doesn't explain
1. why isn't failing in master
2. why the Marshaller (and externalizers) is being loaded for a non-clustered global configuration

danberindei:
It's been too long since the last update, so I separated the Optimistic/PessimisticTxPartitionAndMergeTest-related changes into https://github.com/infinispan/infinispan/pull/5236 and the trace logging changes into https://github.com/infinispan/infinispan/pull/5235.

pruivo:
@danberindei I've approved the change (no need to waste more time discussion the issue), but there is one comment that you didn't reply:
```
with this commands, I was expecting the remaining invokeRemotely* to be deprecated. What is the plan?
```

rvansa:
OK, since Pedro approved, I won't try to do another review.

core/src/main/java/org/infinispan/remoting/transport/MultiTargetRequest.java
danberindei: @pruivo When I was talking about how many objects are instantiated, I meant the objects created internally by the stream, which may or may not be elided by the JIT.

danberindei:
@pruivo I didn't answer it specifically because I thought it was too similar to https://github.com/infinispan/infinispan/pull/5143#discussion_r121913911
I thought I'd have some more time to work on the `RpcManager` API changes between this being integrated and 9.1, but that didn't work out as planned, so I'm adding the `@Experimental` tag to the new methods.

danberindei:
Updated to mark all new `Transport` methods as experimental and to move some of them to the `impl` subpackage.

pruivo:
@danberindei you have 8 unused imports :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsAddress.java
danberindei: 1. Because in master the signature of the method is `org.infinispan.remoting.transport.Address fromJGroupsAddress(Object address)`.
2. Because even local caches can have stores.

danberindei:
@pruivo That's late night refactoring for you... fixed!

pruivo:
@danberindei needs rebase

danberindei:
@pruivo Rebased and updated `MultiTargetRequest` to ignore unexpected responses (fixes a random failure in `TotalOrderStateTransferFunctionalTest`).

pruivo:

PR #5146, opened 2017-05-22
ISPN-7838 Fix JBoss Modules NPE in Domain mode

ryanemerson:
https://issues.jboss.org/browse/ISPN-7838

I think we will need this for 9.0.x and probably 8.2.x.

@galderz @tristantarrant For ISPN-7652 it wasn't sufficient to simply ignore the river marshaller because wildfly-core depends on the river marshaller [upon domain startup](https://github.com/wildfly/wildfly-core/blob/master/server/src/main/java/org/jboss/as/server/DomainServerMain.java#L116), hence the NPE we encounter. So instead, we now explicitly depend on the version we need (2.0.0.Beta3), whilst excluding jboss-marshalling from this jar. This resolves the NPE and prevents `rg.infinispan.server.test.client.hotrod.HotRodRemoteCacheIT` from failing (as discussed [here](https://github.com/infinispan/infinispan/pull/4709))

### DIFF
--- a/server/integration/feature-pack/pom.xml
+++ b/server/integration/feature-pack/pom.xml
@@ -566,6 +566,18 @@
         </dependency>
 
         <dependency>
+            <groupId>org.jboss.marshalling</groupId>
+            <artifactId>jboss-marshalling-river</artifactId>
+            <version>${version.org.jboss.marshalling.jboss-marshalling}</version>
+            <exclusions>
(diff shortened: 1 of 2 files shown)


ryanemerson:
@galderz I can confirm that this also affects 8.2.x, so I'll make sure I backport the final fix.

PR #5148, opened 2017-05-22
ISPN-7841 Add stream operations that can operate upon data exclusively

wburns:
* Doesn't work with simple cache or optimistic caches

https://issues.jboss.org/browse/ISPN-7841


wburns:
This should be good, but putting as preview just in case. Trying to think of a nice way of doing the lock stream with a simple cache as well. May have to iterate over the container then inside compute invoke the consumer, but I am not sure if I like this or not.

rvansa:
Test failures don't seem related, resetting labels...

wburns:
I just remembered I missed updating a few places, closing this for now. But at least I know the tests are fine :)

wburns:
@rvansa actually check out #5156 . I couldn't reopen this one since I had rebased the first commit away.

core/src/main/java/org/infinispan/AdvancedCache.java
```diff
@@ -169,6 +169,16 @@
    AuthorizationManager getAuthorizationManager();
 
    /**
+    * Whenever this cache acquires a lock it will do so using the given Object as the owner of said lock.
+    * <p>
+    * This method provides no guarantees as to how this affect locking and should not be used for general usage.
```
wburns: Yeah tbh this is not really designed for users to use. But I can try to be more specific as to how it works or I could be even more forceful in saying this shouldn't be used by an end user (only says not for general usage currently).

rvansa: -100. It's a public API, it's to be used. No hacks. Say what it does, and what it does not (or provide a link to an explanatory document, if you want to keep javadoc concise).

wburns: That is part of the problem is we don't really have a nice way to have this on a non public API. But that is a different issue.

But I wouldn't call it a hack though, just there shouldn't ever be a reason that an end user should ever use this. I can detail what it does in more detail though, which should help curb someone from using it, instead of saying don't use it.

PR #5150, opened 2017-05-23
ISPN-7853 Delay extracting metadata from command (9.0.x)

galderz:
https://issues.jboss.org/browse/ISPN-7853


ryanemerson:
@galderz LGTM, tested locally. Thanks!

PR #5156, opened 2017-05-24
ISPN-7841 Add stream operations that can operate upon data exclusively

wburns:
* Doesn't work with simple cache or optimistic caches

https://issues.jboss.org/browse/ISPN-7841

Reworked this a bit to make sure I cover all of the various methods (includes certain methods not being allowed in the lock section. These are ones that deal with locks and async where the lock transcends the invocation.


wburns:
Hoping CI is still good :)

wburns:
Test failures appear unrelated.

core/src/main/java/org/infinispan/cache/impl/TypeConverterDelegatingAdvancedCache.java
```diff
@@ -744,4 +744,12 @@ public void replaceAll(BiFunction<? super K, ? super V, ? extends V> function) {
       return returned;
    }
 
+   @Override
+   public AdvancedCache<K, V> lockAs(Object lockOwner) {
+      AdvancedCache<K, V> returned = super.lockAs(lockOwner);
+      if (returned != this && returned instanceof TypeConverterDelegatingAdvancedCache) {
+         ((TypeConverterDelegatingAdvancedCache) returned).entryFactory = this.entryFactory;
```
rvansa: why isn't the `entryFactory` copied simply in the wrapper function?

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
```diff
@@ -0,0 +1,231 @@
...
+   }
+
+   @Override
+   public BaseCacheStream distributedBatchSize(int batchSize) {
+      return newOrReuse(realStream.distributedBatchSize(batchSize));
+   }
+
+   @Override
+   public BaseCacheStream segmentCompletionListener(SegmentCompletionListener listener) {
+      throw new UnsupportedOperationException("LockStream doesn't support completion listener");
```
rvansa: Please add comment: is it just not-implemented-yet, or is something principially blocking impl?

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
```diff
@@ -0,0 +1,231 @@
...
+   }
+
+   @Override
+   public BaseCacheStream timeout(long timeout, TimeUnit unit) {
+      return newOrReuse(realStream.timeout(timeout, unit));
+   }
+
+   @Override
+   public Iterator<CacheEntry<K, V>> iterator() {
+      return realStream.iterator();
```
rvansa: Aren't you losing the locked semantics when an user processes the entries using `(spl)iterator()` rather than `forEach()`? Shouldn't you rather let these method throw an exception?

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
```diff
@@ -0,0 +1,231 @@
...
+      private transient LockManager lockManager;
+
+      private CacheEntryConsumer(BiConsumer<Cache<K, V>, ? super CacheEntry<K, V>> realConsumer,
+            Predicate<? super CacheEntry<K, V>> predicate) {
+         this.realConsumer = realConsumer;
+         this.predicate = predicate;
+      }
+
+      @Override
+      public void accept(Cache<K, V> kvCache, CacheEntry<K, V> kvCacheEntry) {
```
rvansa: Have you moved to hungary? :)

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
```diff
@@ -0,0 +1,231 @@
...
+               // Pass the Cache with the owner set to our key so they can write and also it won't unlock that key
+               realConsumer.accept(kvCache.getAdvancedCache().lockAs(key), kvCacheEntry);
+            }
+         } finally {
+            lockManager.unlock(key, key);
+         }
+      }
+
+      private void lock(K key) {
+         KeyAwareLockPromise kalp = lockManager.lock(key, key, 10, TimeUnit.SECONDS);
```
rvansa: Why don't you use locking timeout from cache configuration?

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
```diff
@@ -0,0 +1,231 @@
...
+   public LockedStream<K, V> onClose(Runnable closeHandler) {
+      return newOrReuse(realStream.onClose(closeHandler));
+   }
+
+   @Override
+   public void close() {
+      realStream.close();
+   }
+
+   @SerializeWith(value = CacheEntryConsumer.CacheEntryConsumerExternalizer.class)
```
rvansa: Maybe you could extract this to top-level private class to shorten the marshalled class name.

core/src/main/java/org/infinispan/util/concurrent/locks/impl/DefaultLockManager.java
```diff
@@ -152,7 +152,11 @@ public void unlockAll(Collection<?> keys, Object lockOwner) {
 
    @Override
    public void unlockAll(InvocationContext context) {
-      unlockAll(context.getLockedKeys(), context.getLockOwner());
+      Set<Object> keysToUnlock = context.getLockedKeys();
+      // We remove lock owner from locked keys - we don't unlock this key it is special
+      Object lockOwner = context.getLockOwner();
```
rvansa: Lock owner is a key?

core/src/test/java/org/infinispan/api/APINonTxTest.java
```diff
@@ -661,4 +669,51 @@ public boolean equals(Object obj) {
...
+               throw new RuntimeException(e1);
+            }
+         }
+      }));
+
+      barrier.await(10, TimeUnit.SECONDS);
+
+      Future<Object> putFuture = fork(() -> cache.put(key, "value" + key + "-new"));
+      try {
+         putFuture.get(50, TimeUnit.MILLISECONDS);
```
rvansa: `TestingUtil.shortTimeoutMillis`, `Exceptions.expectException`

core/src/main/java/org/infinispan/LockedStream.java
```diff
@@ -0,0 +1,77 @@
...
+
+import org.infinispan.configuration.cache.TransactionConfiguration;
+import org.infinispan.container.entries.CacheEntry;
+import org.infinispan.util.function.SerializableBiConsumer;
+import org.infinispan.util.function.SerializablePredicate;
+
+/**
+ * Stream that allows for operation upon data solely with side effects by using {@link LockedStream#forEach(BiConsumer)}
+ * where the <b>BiConsumer</b> is invoked while guaranteeing that the entry being passed is properly locked for the
+ * entire duration of the invocation.
```
rvansa: Add an example of usage: I would rather modify the `CacheEntry` than the cache itself.

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
```diff
@@ -0,0 +1,231 @@
...
+/**
+ * Lock Stream implementation that locks a key using the {@link LockManager} before and after executing the various
+ * code.
+ * <p>
+ * This implementation doesn't work properly when using an optimistic transactional cache. Care should be made to prevent
+ * that usage if possible.
+ * @author wburns
+ * @since 9.0
+ */
+public class LockStreamImpl<K, V> implements LockedStream<K, V> {
```
rvansa: Call this Lock**ed**StreamImpl, please

core/src/main/java/org/infinispan/cache/impl/DecoratedCache.java
```diff
@@ -476,22 +577,25 @@ public void removeGroup(String groupName) {
...
    public V putIfAbsent(K key, V value) {
-      return cacheImplementation.putIfAbsent(key, value, cacheImplementation.defaultMetadata, flags);
+      return putIfAbsent(key, value, cacheImplementation.defaultMetadata);
    }
 
    @Override
    public boolean remove(Object key, Object value) {
-      return cacheImplementation.remove(key, value, flags);
+      cacheImplementation.assertKeyValueNotNull(key, value);
+      InvocationContext ctx = cacheImplementation.getInvocationContextWithImplicitTransaction(false, 1);
```
rvansa: Why aren't you setting the lockOwner here? (or in replace...)

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
```diff
@@ -103,12 +103,13 @@ public Object visitGetCacheEntryCommand(InvocationContext ctx, GetCacheEntryComm
 
    // We need this method in here because of putForExternalRead
    protected final Object visitNonTxDataWriteCommand(InvocationContext ctx, DataWriteCommand command) throws Throwable {
-      if (hasSkipLocking(command) || !shouldLockKey(command.getKey())) {
+      Object key;
+      if (hasSkipLocking(command) || ((key = command.getKey()) == ctx.getLockOwner()) || !shouldLockKey(key)) {
```
rvansa: This needs a verbose comment.

rvansa:
One more thing: I'd like to see tests for starting the foreach within transaction, starting a transaction inside the consumer and combination of those.

wburns:
"I think I got how this should work, and while I don't like it, it's acceptable." @rvansa Trust me I really dislike the changes required to get this working. And to be honest it will get even worse when @tristantarrant adds more changes for Subject in the SecureCache being added to the context it will be even worse. If you have a suggestion of how to do this in a different way I am all ears. I myself was going to spend time next week thinking more about it since I have to think about how we can add Subject support.

core/src/main/java/org/infinispan/LockedStream.java
wburns: I personally hate that usage. It is really tailored for single node modifications. I can double check by adding a test if this works properly, but it should though.

core/src/main/java/org/infinispan/LockedStream.java
```diff
@@ -0,0 +1,77 @@
...
+    * The compiler will pick this overload for lambda parameters, making them <code>Serializable</code>
+    * @param predicate the predicate to filter out unwanted entries
+    * @return a LockedStream with the filter applied
+    */
+   default LockedStream<K, V> filter(SerializablePredicate<? super CacheEntry<K ,V>> predicate) {
+      return filter((Predicate<? super CacheEntry<K, V>> ) predicate);
+   }
+
+   /**
+    * Performs an action for each element of this stream.
```
wburns: It is performed only on the primary owner, just like streams and only place that can be done with the lock. And either way should work. I can make the javadoc more clear.

core/src/main/java/org/infinispan/LockedStream.java
```diff
@@ -0,0 +1,77 @@
...
+    * <p>
+    * This method is performed while holding exclusive lock over the given entry and will be released
+    * only after the consumer has completed.
+    * <p>
+    * If using a pessimistic transaction this lock is not held using a transaction and thus the user can start a
+    * transaction in this consumer which also must be completed before returning.
```
wburns: nope the tx will use the same lock owner. It needs to be completed before returning because our code will release the lock. I personally feel that we don't need to say this reason and it is better coding standards since this api is really sync only.

core/src/main/java/org/infinispan/cache/impl/DecoratedCache.java
```diff
@@ -13,19 +14,25 @@
...
 import org.infinispan.metadata.EmbeddedMetadata;
 import org.infinispan.metadata.Metadata;
 import org.infinispan.stream.StreamMarshalling;
 import org.infinispan.stream.impl.local.ValueCacheCollection;
 
-import javax.security.auth.Subject;
-
 /**
- * A decorator to a cache, which can be built with a specific set of {@link Flag}s.  This
+ * * A decorator to a cache, which can be built with a specific set of {@link Flag}s.  This
```
wburns: Thanks conflict resolution at its best :) Missed it.

core/src/main/java/org/infinispan/cache/impl/DecoratedCache.java
```diff
@@ -42,32 +49,39 @@
...
    public DecoratedCache(AdvancedCache<K, V> delegate, Flag... flags) {
-      super(delegate);
+      this(delegate, null, flags);
+   }
 
+   public DecoratedCache(AdvancedCache<K, V> delegate, Object lockOwner, Flag... flags) {
+      super(delegate);
       if (flags == null)
-         throw new IllegalArgumentException("There is no point in using a DecoratedCache without specifying any Flags.");
+         throw new IllegalArgumentException("There is no point in using a DecoratedCache if Flags are set.");
```
wburns: Yeah this was another conflict resolution one that I missed. I fixed the original merge but the comment wasn't updated. I can change the logic though as the previous commit is incorrect as you mentioned.

core/src/main/java/org/infinispan/cache/impl/DecoratedCache.java
```diff
@@ -267,16 +299,22 @@ public boolean replace(K key, V oldValue, V value, long lifespan, TimeUnit lifes
             .maxIdle(maxIdle, maxIdleUnit)
             .build();
 
-      return cacheImplementation.putAsync(key, value, metadata, flags);
+      return putAsync(key, value, metadata);
    }
 
    @Override
    public CompletableFuture<Void> putAllAsync(Map<? extends K, ? extends V> data) {
+      if (lockOwner != null) {
```
wburns: I had thought about that, but couldn't decide. I can make a private method where it passes the method name and do string concat in the exception then.

core/src/main/java/org/infinispan/cache/impl/DecoratedCache.java
```diff
@@ -400,7 +477,7 @@ public boolean isEmpty() {
 
    @Override
    public boolean containsKey(Object key) {
-      return cacheImplementation.containsKey(key, flags);
+      return get(key) != null;
```
wburns: I have to look again but the reason is because contains key required more changes in CacheImpl to get the lock owner to be set properly. I can spend the extra time to fix it though as well.

core/src/main/java/org/infinispan/cache/impl/DecoratedCache.java
wburns: Hrmm weird looking at the diff again part of my changes were reverted... I will have to fix these up.

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
wburns: Sure

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
wburns: I debated about these. I agree throwing an exception would be better now that I think about it more.

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
wburns: The only issue I originally had is because the timeout is only _used_ with transactional caches. So setting this on a non tx cache for this purpose seemed a bit weird. But I can change it to use it either way.

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
wburns: Sorry I am missing the reference. If you mean the variables names, you can blame IntelliJ :D

core/src/test/java/org/infinispan/api/APINonTxTest.java
wburns: This would cause an additional 450 MS test time and the expected exception is already ensured by the try catch. I don't see the need to change these.

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
wburns: Sure

wburns:
"One more thing: I'd like to see tests for starting the foreach within transaction, starting a transaction inside the consumer and combination of those."

Sure

core/src/main/java/org/infinispan/LockedStream.java
rvansa: Not sure if I follow. I've intended to point out that it's a bit ambiguous if you should use `cache.put(...)` or `cacheEntry.setValue(...)` for the modification (or both); in the tests you use the former, so I assume that this is it. This pattern should be mentioned in the javadoc to make it clear for those that don't read testsuite :)
In addition to that, `cacheEntry.setValue(...)` et all should either transparently call `cache.put(...)` (but that might be unsuitable as the user could want to set metadata as well), or throw an exception.

core/src/main/java/org/infinispan/LockedStream.java
rvansa: OK, the reason is not needed, but the fact that it's safe to modify the key in nested tx is.

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
rvansa: https://en.wikipedia.org/wiki/Hungarian_notation

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
rvansa: I think it is used in non-tx caches, too: https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java#L111 which goes to https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java#L215

core/src/main/java/org/infinispan/LockedStream.java
wburns: Either way should work, like I said I can add a test to verify that it works.

core/src/main/java/org/infinispan/LockedStream.java
wburns: Okay.

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
wburns: Yeah I can use https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/configuration/cache/LockingConfiguration.java#L77 For whatever reason I was thinking it was defined on the transaction. Thanks.

core/src/main/java/org/infinispan/cache/impl/TypeConverterDelegatingAdvancedCache.java
wburns: It can't be as the wrapper function is only given the wrapped `AdvancedCache`. It doesn't have access to the old wrapper and thus the factory needed.

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
wburns: The name actually doesn't matter much in this case. This is using the less performant Serializer which actually ships across the entire Serializer class as is. If we wanted this to be as highest performant we would want to extract an actual `AdvancedExternalizer` id. I figured this operation is not about # of invocations as it has to perform the `BiConsumer` for every entry after obtaining the key. This requires a lot less code and doesn't require a reserved ID. Do you think I should also convert this over? It just seemed like overkill.

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
wburns: yeah I blame IntelliJ :) I will change it though.

core/src/main/java/org/infinispan/stream/impl/LockStreamImpl.java
rvansa: I thought that it will just lookup the serializer on local node by name rather than passing the short id but I may be wrong. Anyway, let it be - as you say it's not on the hot path.

core/src/main/java/org/infinispan/cache/impl/TypeConverterDelegatingAdvancedCache.java
rvansa: Check this out (core testsuite seems to pass): https://github.com/rvansa/infinispan/commit/1b71c251ad73870aeed88476dafeb18026fb918a

wburns: Sadly this is just because there isn't a test that has this exact scenario. I will add a 5 line test to `StoreAsBinaryTest` that will reproduce the NPE.

wburns:
Made additional changes and added extra testing. Waiting on CI again though.

core/src/main/java/org/infinispan/cache/impl/DecoratedCache.java
wburns: Unfortunately I can't do this without adding a brand new command.

rvansa: I don't follow...?

wburns: Sorry I wrote this too quickly. I can delegate to `CacheImpl.containsKey`, I would need to add some additional methods for `containsKey` and `get` in `CacheImpl`, just more cod than I would have hoped, but seems fine.

wburns: Actually I find that changing the CacheImpl methods to take the `InvocationContext` makes the `DecoratedCache` a bit nicer. So I will try that out.

wburns:
Updated to clean up `DecoratedCache` to reduce code duplication from `CacheImpl`.

core/src/test/java/org/infinispan/api/APINonTxTest.java
rvansa: OK, failing the timeout wouldn't cause flakiness. But use `expectException` anyway, please.

core/src/test/java/org/infinispan/api/APINonTxTest.java
```diff
@@ -661,4 +669,63 @@ public boolean equals(Object obj) {
...
+
+      forEachFuture.get(10, TimeUnit.SECONDS);
+
+      // The put should replace the value that forEach inserted
+      assertEquals("value" + key + "-other", putFuture.get(10, TimeUnit.SECONDS));
+      // The put should be last since it had to wait until lock was released on forEachWithLock
+      assertEquals("value" + key + "-new", cache.get(key));
+
+      // Make sure the locks were cleaned up properly
+      LockManager lockManager = cache.getAdvancedCache().getComponentRegistry().getComponent(LockManager.class);
```
rvansa: `TestingUtil.extractComponent` is a shorthand for this

core/src/test/java/org/infinispan/api/BaseCacheAPIPessimisticTest.java
```diff
@@ -0,0 +1,156 @@
...
+      }
+
+      CyclicBarrier barrier = new CyclicBarrier(2);
+
+      int key = 4;
+
+      Future<Object> putFuture = fork(() -> {
+         TransactionManager tm = cache.getAdvancedCache().getTransactionManager();
+         try {
+            tm.begin();
```
rvansa: `TestingUtil.withTx`

core/src/test/java/org/infinispan/api/BaseCacheAPIPessimisticTest.java
```diff
@@ -0,0 +1,156 @@
...
+      });
+
+      // Wait until fork thread has alredy locked the key
+      barrier.await(10, TimeUnit.SECONDS);
+
+      LockedStream<Object, Object> stream = cache.getAdvancedCache().lockedStream();
+      Future<?> forEachFuture = fork(() -> stream.filter(e -> e.getKey().equals(key)).forEach((c, e) ->
+            assertEquals("value" + key + "-new", c.put(e.getKey(), String.valueOf(e.getValue() + "-other")))));
+
+      try {
```
rvansa: `expectException`

core/src/test/java/org/infinispan/api/BaseCacheAPIPessimisticTest.java
```diff
@@ -0,0 +1,156 @@
...
+   @Test(dataProvider = "testLockedStreamInTx")
+   public void testLockedStreamInTxCommit(Boolean shouldCommit) throws SystemException, NotSupportedException,
+         HeuristicRollbackException, HeuristicMixedException, RollbackException {
+      for (int i = 0; i < 5; i++) {
+         cache.put(i, "value" + i);
+      }
+
+      TransactionManager tm = cache.getAdvancedCache().getTransactionManager();
+      tm.begin();
+      try {
```
rvansa: `TestingUtil.withTx` and there `if (!shouldCommit) tm.setRollbackOnly()`

core/src/test/java/org/infinispan/api/CacheAPITest.java
```diff
@@ -271,4 +272,113 @@ public void testPutIfAbsentAfterRemoveInTx() throws SystemException, NotSupporte
...
    }
+
+   public void testSizeInExplicitTxWithNonExistent() throws SystemException, NotSupportedException {
+      assertEquals(0, cache.size());
+      cache.put("k", "v");
+
+      TransactionManager tm1 = TestingUtil.getTransactionManager(cache);
+      tm1.begin();
+      try {
+         cache.get("no-exist");
```
rvansa: assertNull? And the withTx... though I know you've just moved this code from another test.

core/src/test/java/org/infinispan/api/BaseCacheAPIPessimisticTest.java
```diff
@@ -0,0 +1,156 @@
...
+            tm.commit();
+         } catch (Exception e1) {
+            throw new RuntimeException(e1);
+         }
+      });
+
+      for (int i = 0; i < 5; i++) {
+         assertEquals("value" + i + "-changed", cache.get(i));
+      }
+   }
```
rvansa: I still miss the case when you have both outer and inner transaction (though from the code I am pretty confident it will work at this moment).

core/src/test/java/org/infinispan/api/mvcc/read_committed/CacheAPIPessimisticTest.java
```diff
@@ -1,17 +1,20 @@
...
  * @since 5.1
  */
 @Test (groups = "functional", testName = "api.mvcc.read_committed.CacheAPIPessimisticTest")
-public class CacheAPIPessimisticTest extends CacheAPIOptimisticTest {
+public class CacheAPIPessimisticTest extends BaseCacheAPIPessimisticTest {
    @Override
-   protected void amend(ConfigurationBuilder cb) {
-      cb.transaction().lockingMode(LockingMode.PESSIMISTIC);
+   protected IsolationLevel getIsolationLevel() {
+      return IsolationLevel.READ_COMMITTED;
```
rvansa: Is there a RR variant for pessimistic caches?

core/src/test/java/org/infinispan/api/mvcc/read_committed/CacheAPIPessimisticTest.java
```diff
@@ -1,17 +1,20 @@
 package org.infinispan.api.mvcc.read_committed;
 
+import org.infinispan.api.BaseCacheAPIPessimisticTest;
+import org.infinispan.api.CacheAPITest;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.transaction.LockingMode;
```
rvansa: I guess this will become an unused import - please check Analyze / Run inspection by name / Unused import in IntellliJ

core/src/test/java/org/infinispan/api/APINonTxTest.java
wburns: I really don't see what `expectException` would buy us here that the try catch doesn't already give us. The usage of `Exceptions` class in my mind is when you have a `Throwable` or something like that and you want to either unwrap it or assert the exact class instance of. The try catch idiom already guarantees that only a `TimeoutException` is caught so if any other is caught it will fail the test and if it isn't thrown the fail will complain as well.

core/src/test/java/org/infinispan/api/APINonTxTest.java
wburns: Sure, I am just used to not using it in non test code :)

core/src/main/java/org/infinispan/AdvancedCache.java
```diff
@@ -559,6 +571,27 @@
...
+    * The stream will not share any ongoing transaction the user may have. Code executed by the stream should be treated
+    * as completely independent. That is any operation performed via the stream will require the user to start their
+    * own transaction or will be done intrinsically on the invocation. Note this means that if you lock a key from
+    * the cache it could cause the stream operation to deadlock.
+    * {@link org.infinispan.configuration.cache.ConfigurationBuilder#simpleCache(boolean)} was set to true. This
+    * restriction may be removed in a future version
+    * @return the locked stream
+    * @since 9.1
+    */
+   LockedStream<K, V> lockedStream() throws UnsupportedOperationException;
```
wburns: I know that I don't, but it makes it more apparent that it can be thrown, which I personally like. Unfortunately in my refactoring of the javadoc I lost the reason as to why this is thrown as you see in the block above re: simple cache. I will fix that.

core/src/main/java/org/infinispan/LockedStream.java
```diff
@@ -0,0 +1,104 @@
...
+    * {@link java.util.Map.Entry#setValue(Object)} method this will be the same as if
+    * {@link java.util.Map#put(Object, Object)} was invoked.
+    * <p>
+    * If using a pessimistic transaction this lock is not held using a transaction and thus the user can start a
+    * transaction in this consumer which also must be completed before returning. A transaction can be started in
+    * the consumer and if done it will share the same lock used to obtain the key.
+    * <p>
+    * Some methods on the provided cache may not work as expected. These include
+    * {@link AdvancedCache#putForExternalRead(Object, Object)}, {@link AdvancedCache#lock(Object[])},
+    * {@link AdvancedCache#lock(Collection)}, {@link AdvancedCache#removeGroup(String)}, all of the async methods on
```
wburns: the problem with async is that it could complete after I have already unlocked the key in the stream. Everything needs to be executed and completed in the Consumer to guarantee that it has the lock during that period.

core/src/main/java/org/infinispan/commands/read/EntrySetCommand.java
```diff
@@ -57,6 +57,10 @@ public LoadType loadType() {
 
    @Override
    public Set<CacheEntry<K, V>> perform(InvocationContext ctx) throws Throwable {
+      Object lockOwner = ctx.getLockOwner();
+      if (ctx.getLockOwner() != null) {
+         return new BackingEntrySet<>(cache.getAdvancedCache().lockAs(lockOwner));
```
wburns: There is nothing stopping a locked stream in another and actually it would work perfectly well, unless someone changed the lock object by invoking `lockAs` themselves, which was why I was trying to tell people they can't guarantee how it would work.

This block though has nothing to do with locked stream, all it is doing is paying attention to the lock owner for the underlying stream. This allows the user to call `setValue` since that needs the appropriate lock owner inside the locked stream.

core/src/test/java/org/infinispan/api/APINonTxTest.java
rvansa: It's only about concise code: 1 line instead of 5, not having to type the `fail()` message (duplicity in catch cause and in message), and expressing that 'now comes something that should throw' ahead.
If you're uncomfortable with that, I won't push you - I just consider it the best practice.

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
```diff
@@ -103,12 +103,16 @@ public Object visitGetCacheEntryCommand(InvocationContext ctx, GetCacheEntryComm
 
    // We need this method in here because of putForExternalRead
    protected final Object visitNonTxDataWriteCommand(InvocationContext ctx, DataWriteCommand command) throws Throwable {
-      if (hasSkipLocking(command) || !shouldLockKey(command.getKey())) {
+      Object key;
+      // We don't lock the key if the lock owner is the same instance as the key. This a special case where the lock
+      // owner was specifically set by locking code to control it. The controlling code will then unlock the key
+      // themselves.
+      if (hasSkipLocking(command) || ((key = command.getKey()) == ctx.getLockOwner()) || !shouldLockKey(key)) {
```
wburns: I had thought about this a while back, forgot who I talked to it about it. The problem as you already mentioned is what do we do in the case of putAll/getAll and others and it became way too hairy. So I went with just a different lock owner as that would work for all.

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
```diff
@@ -103,12 +103,16 @@ public Object visitGetCacheEntryCommand(InvocationContext ctx, GetCacheEntryComm
 
    // We need this method in here because of putForExternalRead
    protected final Object visitNonTxDataWriteCommand(InvocationContext ctx, DataWriteCommand command) throws Throwable {
-      if (hasSkipLocking(command) || !shouldLockKey(command.getKey())) {
+      Object key;
```
wburns: Thanks for pointing it out :)

core/src/main/java/org/infinispan/LockedStream.java
rvansa: Then it's not different from the incomplete user transaction - user has to make sure it waits for all asynchronously invoked operations, and docs should rather state that instead of prohibiting any async calls.

core/src/test/java/org/infinispan/api/APINonTxTest.java
wburns: Sorry there are so many `expectException` methods on the class, I didn't realize there is a `ExceptionRunnable` override one. I was looking at the first ones, which would be considerably more code. I personally feel the try/catch is more readable and direct than the `ExceptionRunnable`, but I can change it to that.

core/src/main/java/org/infinispan/LockedStream.java
wburns: Alright, I normally err on locking things down and adding them back in later, but I agree it is more than able to be used if the user can guarantee they take care :D

core/src/main/java/org/infinispan/LockedStream.java
```diff
@@ -0,0 +1,104 @@
...
+    * <p>
+    * The compiler will pick this overload for lambda parameters, making them <code>Serializable</code>
+    * @param biConsumer the biConsumer to run for each entry under their lock
+    */
+   default void forEach(SerializableBiConsumer<Cache<K, V>, ? super CacheEntry<K, V>> biConsumer) {
+      forEach((BiConsumer<Cache<K, V>, ? super CacheEntry<K, V>>) biConsumer);
+   }
+
+   /**
+    * Sets the timeout for the acquisition of the lock for each entry.
```
wburns: Sure, I am thinking I might want to put it on the class definition though instead of the method itself.

core/src/test/java/org/infinispan/api/BaseCacheAPIPessimisticTest.java
wburns: Sure, thanks for the reminder.

core/src/main/java/org/infinispan/commands/read/EntrySetCommand.java
rvansa: When you start another locked stream holding the lock, the second iteration will go through the same entry. There, it won't be blocked (because the lock owners match), but when the inner iteration will leave the entry, it will unlock it, before the consumer in the outer iteration finished. Or am I mistaken? This sounds like it needs to be confirmed by a test :)

core/src/test/java/org/infinispan/api/mvcc/read_committed/CacheAPIPessimisticTest.java
wburns: https://github.com/wburns/infinispan/blob/85a06a1245df2c655973828aea99aad1b1cadff4/core/src/test/java/org/infinispan/api/mvcc/repeatable_read/CacheAPIPessimisticTest.java

core/src/test/java/org/infinispan/api/APINonTxTest.java
rvansa: All of them use `ExceptionRunnable` - those are just different overrides for nested exceptions. Here I'd use `expectException(TimeoutException.class, () -> putFuture.get(50, TimeUnit.MILLISECONDS));`, but do it your way once you are sure about the options.

core/src/main/java/org/infinispan/commands/read/EntrySetCommand.java
wburns: Yes, iirc the locks don't count invocations, sadly. This was something I had complained about before. To be honest though I don't think it is worth it to change how that behaves here. I can just throw an exception if `lockOwner` was explicitly set before invoking `lockedStream` WDYT?

core/src/test/java/org/infinispan/api/APINonTxTest.java
wburns: Ah yes sorry, I looked at `assertException` and saw the same thing, haha. Gotta love the brain optimizations :)

core/src/main/java/org/infinispan/commands/read/EntrySetCommand.java
rvansa: Sounds like this could work. Just do what you think is the best and provide a test :)

core/src/main/java/org/infinispan/commands/read/EntrySetCommand.java
```diff
@@ -77,7 +81,7 @@ public String toString() {
       @Override
       public CloseableIterator<CacheEntry<K, V>> iterator() {
          Iterator<CacheEntry<K, V>> iterator = new DataContainerRemoveIterator<>(cache);
-         return new EntryWrapperIterator<>(cache, iterator);
+         return Closeables.iterator(iterator);
```
wburns: Sorry there are a bunch of issues with the `EntryWrapper` for `setValue` when using an iterator or forEach on a stream. I will add this back in. I need to log a JIRA to clean this up though as there are spots missing in different areas. I don't think people were expected to call `setValue` as much :) And usages were just bolted on.

core/src/main/java/org/infinispan/AdvancedCache.java
```diff
@@ -559,6 +571,27 @@
...
+    * returned via {@link AdvancedCache#cacheEntrySet()}. The use of this locked stream is that when an entry is
+    * being processed by the user the entry is locked for the invocation preventing a different thread from modifying
+    * it.
+    * <p>
+    * Note that this stream is not supported when using a optimistic transactional or simple cache. Both non transactional and
+    * pessimistc transactions are supported.
+    * <p>
+    * The stream will not share any ongoing transaction the user may have. Code executed by the stream should be treated
+    * as completely independent. That is any operation performed via the stream will require the user to start their
+    * own transaction or will be done intrinsically on the invocation. Note this means that if you lock a key from
```
wburns: Okay

core/src/main/java/org/infinispan/AdvancedCache.java
wburns: I will remove it and just add `@throws` instead.

wburns:
Okay I think I covered all of your comments. Hopefully CI likes this still :)

core/src/main/java/org/infinispan/AdvancedCache.java
```diff
@@ -559,6 +572,36 @@
...
+    * Note that this stream is not supported when using a optimistic transactional or simple cache. Both non transactional and
+    * pessimistc transactions are supported.
+    * <p>
+    * The stream will not share any ongoing transaction the user may have. Code executed by the stream should be treated
+    * as completely independent. That is any operation performed via the stream will require the user to start their
+    * own transaction or will be done intrinsically on the invocation. Note that if there is an ongoing transaction
+    * that has a lock on a key from the cache, that it will cause a deadlock.
+    * <p>
+    * Currently simple cache, {@link org.infinispan.configuration.cache.ConfigurationBuilder#simpleCache(boolean)}
+    * was set to true, functional map {@link org.infinispan.commons.api.functional.FunctionalMap} and optimistic caches,
```
rvansa: `org.infinispan.commons.api.functional.FunctionalMap` is not an AdvancedCache, so it can't support this method. You have mistaken it for `org.infinispan.functional.decorators.FunctionalAdvancedCache` which is a test class?

core/src/main/java/org/infinispan/AdvancedCache.java
```diff
@@ -559,6 +572,36 @@
    CacheSet<CacheEntry<K, V>> cacheEntrySet();
 
    /**
+    * Returns a sequential stream using this Cache as the source. This stream is very similar to using the
+    * {@link CacheStream} returned from the {@link CacheSet#stream()} method of the collection
+    * returned via {@link AdvancedCache#cacheEntrySet()}. The use of this locked stream is that when an entry is
+    * being processed by the user the entry is locked for the invocation preventing a different thread from modifying
+    * it.
+    * <p>
+    * Note that this stream is not supported when using a optimistic transactional or simple cache. Both non transactional and
```
rvansa: both non-transactional **and* pessimistic transactional caches...

core/src/test/java/org/infinispan/api/APINonTxTest.java
```diff
@@ -661,4 +677,99 @@ public boolean equals(Object obj) {
...
+
+      CyclicBarrier barrier = new CyclicBarrier(2);
+
+      int key = 4;
+
+      LockedStream<Object, Object> stream = cache.getAdvancedCache().lockedStream();
+      Future<?> forEachFuture = fork(() -> stream.forEach((c, e) -> {
+         if (e.getKey().equals(key)) {
+            try {
+               barrier.await(10, TimeUnit.SECONDS);
```
rvansa: Maybe I've missed that during my last review but I don't think this test is doing what you want. Here, the consumer and the put that's about to timeout are running concurrently. I suppose that you want actually run two puts that time out, one before consumer invocation and another afterwards, to verify that the command did not unlock the lock:
```java
      Future<?> forEachFuture = fork(() -> stream.forEach((c, e) -> {
         if (e.getKey().equals(key)) {
            try {
               barrier.await(10, TimeUnit.SECONDS);
               barrier.await(10, TimeUnit.SECONDS);
               consumer.accept(c, e);
               barrier.await(10, TimeUnit.SECONDS);
               barrier.await(10, TimeUnit.SECONDS);
            } catch (InterruptedException | BrokenBarrierException | TimeoutException e1) {
               throw new RuntimeException(e1);
            }
         }
      }));
      barrier.await(10, TimeUnit.SECONDS);
      Future<Object> putFutureBefore = fork(() -> cache.put(key, "value" + key + "-new"));
      Exceptions.expectException(TimeoutException.class, () -> putFutureBefore.get(50, TimeUnit.MILLISECONDS));
      barrier.await(10, TimeUnit.SECONDS);
 
      barrier.await(10, TimeUnit.SECONDS);
      barrier.await(10, TimeUnit.SECONDS);
    
      Future<Object> putFutureAfter = fork(() -> cache.put(key, "value" + key + "-new"));
      Exceptions.expectException(TimeoutException.class, () -> putFutureAfter.get(50, TimeUnit.MILLISECONDS));
      barrier.await(10, TimeUnit.SECONDS);

      forEachFuture.get(10, TimeUnit.SECONDS);
```

core/src/main/java/org/infinispan/interceptors/locking/AbstractLockingInterceptor.java
rvansa: Have you handled this already? It's good that you've added the tests, but as I've mentioned below these seem to me are flawed.

core/src/test/java/org/infinispan/api/APINonTxTest.java
rvansa: You could also check directly that the lock is held at the end of the forEach lambda.

wburns: The test is doing exactly what I thought it would, testing that if a consumer is currently running that it would block another write operation.

I however didn't think to verify that after the inner consumer completed that the lock was still held. I can add a check for the lock manager to verify that though as an extra extra precaution.

core/src/main/java/org/infinispan/AdvancedCache.java
wburns: I know that it isn't an `AdvancedCache` :) I was more writing that this couldn't be invoked from an `AdvanceCache` that was backed by a `FunctionalMap`. However as you mentioned that is only in test, so this text is not needed.

core/src/main/java/org/infinispan/AdvancedCache.java
wburns: ok, was confused at first, didn't realize you meant for the next line :)

core/src/test/java/org/infinispan/api/APINonTxTest.java
rvansa: > if a consumer is currently running

You don't block the consumer somewhere in the middle - most likely you'll run the operations in one order or in the other, but non-deterministically, which ain't good for the testsuite.

wburns: No they always run in the same order.

The `forEach` holds the key of the lock the entire time "outer" consumer duration including all calls to barrier.await.

The order goes like (a and b mean they can run at same time)

1. T1 submits stream
2a. T1 waits on barrier
2b. T2 acquires lock and waits on barrier
3a. T2 can run inner consumer and then waits on barrier while still holding the lock
3b. T3 can run put, but will be blocked
4. T1 checks timeout of T3 and finally release T2 and then waits until T2 completes
5. T2 completes consumer and finally releases lock

Maybe the confusion is the introduction of the second "inner" consumer?

wburns: And that is the point of the test, for it to fail if they do indeed run concurrently (aka lock isn't working for some reason) :)

rvansa:
Ok, let's check CI and integrate this when nothing pops up.

tristantarrant:
Merged, thanks !

PR #5157, opened 2017-05-25
[PRVIEW] regular reads

pruivo:
It changes the reads requests and response from OOB messages to Regular messages.
IspnPerfTests shows slightly performance improvement and one order of magnitude lower thread pool usages (concrete number with 8 node, 150 -> 14 threads).

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -52,7 +52,7 @@
    </issueManagement>
 
    <properties>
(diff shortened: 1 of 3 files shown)


pruivo:
run performance tests please
library dist

belaban:
I think this change will only make sense performance-wise (due to lower thread pool usage) once we handle processing of message batches in a non-sequential way.

PR #5158, opened 2017-05-25
Query security actions cleanup

anistor:
Use lambdas for PrivilegedAction.


ryanemerson:
Integrated, thanks @anistor

PR #5162, opened 2017-05-29
ISPN-7866 Remove metadata param from CacheEntry.commit()

rvansa:
https://issues.jboss.org/browse/ISPN-7866

Metadata should be simply updated using `setMetadata()` on the `CacheEntry` in context, instead of passing the instance down.


galderz:
Looks good to me

PR #5163, opened 2017-05-29
ISPN-7867 MetaParams enhancements

rvansa:
https://issues.jboss.org/browse/ISPN-7867

* minor fixes in generics
* add remove and replace methods to MetaParams

### DIFF
--- a/core/src/main/java/org/infinispan/functional/impl/MetaParams.java
+++ b/core/src/main/java/org/infinispan/functional/impl/MetaParams.java
@@ -6,16 +6,17 @@ import java.io.ObjectOutput;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Stream;
 
 import org.infinispan.functional.MetaParam;
-import org.infinispan.commons.marshall.AbstractExternalizer;
 import org.infinispan.commons.util.Experimental;
-import org.infinispan.commons.util.Util;
-import org.infinispan.marshall.core.Ids;
 
 import net.jcip.annotations.NotThreadSafe;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/infinispan/functional/impl/MetaParams.java
```diff
@@ -50,94 +48,186 @@
...
       for (MetaParam<?> meta : metas) {
-         if (meta.getClass().isAssignableFrom(type))
+         if (meta != null && meta.getClass().isAssignableFrom(type))
             return (T) meta;
       }
 
       return null;
    }
 
-   public void add(MetaParam.Writable meta) {
```
galderz: Why change parameter from `MetaParam.Writable` to `MetaParam`?

core/src/main/java/org/infinispan/functional/impl/MetaParamsExternalizer.java
```diff
@@ -0,0 +1,39 @@
...
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.util.Set;
+
+import org.infinispan.commons.api.functional.MetaParam;
+import org.infinispan.commons.marshall.AbstractExternalizer;
+import org.infinispan.commons.util.Util;
+import org.infinispan.marshall.core.Ids;
+
+public final class MetaParamsExternalizer extends AbstractExternalizer<MetaParams> {
```
galderz: Rather than having a specific `MetaParamsExternalizer`, it'd be more useful to have a `MarshallUtil.marshallIterable`...etc that can work for any Iterable, not just MetaParams. The externalizer is not doing anything special you'd not do for any other Iterable.

rvansa: Makes sense. `marshall`is trivial - but should `unmarshall` return an iterable? Or just use `unmarshallArray` and make sure that these two are compatible?

core/src/main/java/org/infinispan/functional/impl/MetaParams.java
rvansa: `MetaParamsInternalMetadata.Builder` implements `Metadata.Builder`, and one method in that (not published in this PR) calls `MetaParams.replace` when it's building the metadata object. The type I use there shouldn't be settable by the user (it's internal info) so it does not implement `MetaParam.Writable`.
I need only replace, but `add`, `addMany` and `replace` should use the same type. And I've changed `MetaParams` to be non-public class, so all modifications should be done only through checked interfaces (using `Writable`).

core/src/main/java/org/infinispan/functional/impl/MetaParamsExternalizer.java
galderz: Just returning Iterable would be be fine. The client might expect a subtype but that should work. Don't think I'd want to force this into any array.

core/src/main/java/org/infinispan/functional/impl/MetaParams.java
galderz: Ok, makes sense.

core/src/main/java/org/infinispan/functional/impl/MetaParamsExternalizer.java
rvansa: The problem with `Iterable` as return type is that the code calling `unmarshall` does not know the number of elements returned (and it's likely to have a use for that), while the `MarshallUtil` knows that. And actually the problem occurs with `marshall` as well, unless you provide number of elements ahead.

Any way, you're right that we should get rid of the externalizer and just expose static methods used by `MetaParamsInternalMetadata.Externalizer`

rvansa:
@galderz updated, I've removed the externalizer completely.

core/src/main/java/org/infinispan/functional/impl/MetaParams.java
```diff
@@ -50,94 +51,186 @@
...
             boolean found = false;
             for (int i = 0; i < metas.length; i++) {
-               if (metas[i].getClass().isAssignableFrom(newMeta.getClass())) {
+               MetaParam<?> m = metas[i];
+               if (m == null) {
+                  hole = i;
+               } else if (m.getClass().isAssignableFrom(newMeta.getClass())) {
                   metas[i] = newMeta;
                   found = true;
+                  break;
```
pruivo: here you can use label (just a suggestion :). `found` can be removed in this case)
```
outer: for (MetaParam newMeta : metaParams) {
...
    metas[i] = newMeta;
    continue outer;
...
}

**Comment by rvansa** [comment]
@pruivo Thanks, updated & rebased.

File: core/src/main/java/org/infinispan/functional/impl/MetaParams.java [hunk 4]
```diff
@@ -50,94 +51,183 @@
...
+      }
+   }
+
+   public <T extends MetaParam> void replace(Class<T> type, Function<T, T> f) {
+      int hole = -1;
+      for (int i = 0; i < metas.length; ++i) {
+         MetaParam<?> m = metas[i];
+         if (m == null) {
+            hole = i;
+         } else if (m != null && m.getClass().isAssignableFrom(type)) {
```
pruivo: minor: no need to check for null again

core/src/main/java/org/infinispan/functional/impl/MetaParams.java
```diff
@@ -50,94 +51,183 @@
...
       }
+      return l == length;
    }
 
-   public void addMany(MetaParam.Writable... metaParams) {
-      if (metas.length == 0) metas = metaParams;
-      else {
+   public void addMany(MetaParam... metaParams) {
+      if (metas.length == 0) {
+         // Arrays in Java are convergent, therefore someone could pass MetaParam.Writable[] and
```
danberindei: covariant?

core/src/main/java/org/infinispan/functional/impl/MetaParams.java
```diff
@@ -50,94 +51,183 @@
...
+            int hole = -1;
             for (int i = 0; i < metas.length; i++) {
-               if (metas[i].getClass().isAssignableFrom(newMeta.getClass())) {
+               MetaParam<?> m = metas[i];
+               if (m == null) {
+                  hole = i;
+               } else if (m.getClass().isAssignableFrom(newMeta.getClass())) {
                   metas[i] = newMeta;
-                  found = true;
+                  continue OUTER;
```
danberindei: I'd rather extract a separate method with the loop body and use `return`, if possible.

core/src/main/java/org/infinispan/functional/impl/MetaParams.java
```diff
@@ -150,62 +240,88 @@ public String toString() {
...
+         skipNulls();
+         return i < metas.length;
       }
 
       @Override
-      public Integer getId() {
-         return Ids.META_PARAMS;
+      public MetaParam<?> next() {
+         try {
+            return metas[i++];
```
danberindei: One can call `it.next().next()` without any `hasNext()` in between, so the `skipNulls()` call should be here.

rvansa:
@danberindei may all your wishes come true

danberindei:
Looks good, waiting for CI. Or is there anything still bothering you @pruivo ?

danberindei:
CI looks pristine, integrating ;)

danberindei:
Integrated, thanks Radim!

PR #5164, opened 2017-05-30
Ispn 7793/map internal external addresses

slaskawi:
https://issues.jboss.org/browse/ISPN-7793

### DIFF
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExternalAddressRoutingTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExternalAddressRoutingTest.java
@@ -1,0 +1,172 @@
+package org.infinispan.client.hotrod;
+
+import static org.infinispan.client.hotrod.test.HotRodClientTestingUtil.killServers;
+import static org.infinispan.server.hotrod.test.HotRodTestingUtil.hotRodCacheConfiguration;
+import static org.infinispan.test.TestingUtil.blockUntilCacheStatusAchieved;
+import static org.infinispan.test.TestingUtil.blockUntilViewReceived;
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.fail;
+
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.infinispan.client.hotrod.impl.AddressMapper;
+import org.infinispan.client.hotrod.test.HotRodClientTestingUtil;
+import org.infinispan.commons.util.Util;
+import org.infinispan.configuration.cache.CacheMode;
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.lifecycle.ComponentStatus;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.server.hotrod.HotRodServer;
+import org.infinispan.util.ControlledConsistentHashFactory;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.Test;
+
+import com.github.terma.javaniotcpproxy.StaticTcpProxyConfig;
+import com.github.terma.javaniotcpproxy.TcpProxy;
+
+@Test(groups = "functional", testName = "client.hotrod.ExternalAddressRoutingTest")
+public class ExternalAddressRoutingTest extends HitsAwareCacheManagersTest {
+
+   private static final int PROXY_TO_SERVER_1_PORT = 9988;
+   private static final int PROXY_TO_SERVER_2_PORT = 9989;
+
+   private static final Map<SocketAddress, SocketAddress> proxyToRealAddresses = new HashMap<>();
+   private static final Map<SocketAddress, SocketAddress> realAddressesToProxy = new HashMap<>();
+
+   private static final String DIST_ONE_CACHE_NAME = "dist-one-cache";
+   private static final String DIST_TWO_CACHE_NAME = "dist-two-cache";
+
+   HotRodServer server1;
+   HotRodServer server2;
+
+   ConfigurationBuilder defaultBuilder;
+   ConfigurationBuilder distOneBuilder;
+   ConfigurationBuilder distTwoBuilder;
+
+   TcpProxy proxyToServer1;
+   TcpProxy proxyToServer2;
+
+   RemoteCacheManager rcm;
+
+   protected ConfigurationBuilder defaultCacheConfigurationBuilder() {
+      return hotRodCacheConfiguration(getDefaultClusteredCacheConfig(CacheMode.DIST_SYNC));
+   }
+
+   @Override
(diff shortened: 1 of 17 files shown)


slaskawi:
More information about this PR: https://github.com/slaskawi/external-ip-proxy

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod.adoc
```diff
@@ -917,6 +917,38 @@ In order to avoid creating a TCP connection on each request (which is a costly o
 
 The number of connections per server, total number of connections, how long should a connection be kept idle in the pool before being closed - all these (and more) can be configured. Please refer to the javadoc of link:{javadocroot}/org/infinispan/client/hotrod/RemoteCacheManager.html[RemoteCacheManager] for a list of all possible configuration elements.
 
+==== Using internal/external IP addresses
+
+In some cases (e.g. OpenStack or Kubernetes with services exposed using load balancers) servers use different IP addresses for clustering and different for communicating with the client.
```
galderz: Will be provide some ready made address mapper instances for the user to use? 

I don't think they should have to implement these interfaces themselves.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExternalAddressRoutingTest.java
```diff
@@ -0,0 +1,194 @@
...
+      RemoteCache<Object, Object> rcOne = rcm.getCache(cacheName);
+      InetSocketAddress serverAddress = new InetSocketAddress(server.getHost(), server.getPort());
+      for (int i = 0; i < 2; i++) {
+         log.infof("Routing put test for key %s", Util.printArray(key, false));
+         rcOne.put(key, "value");
+         assertServerHit(serverAddress, cacheName, i + 1);
+      }
+   }
+
+   byte[] getKeyForServer(HotRodServer primaryOwner, String cacheName) {
```
galderz: We should avoid code duplication, `HotRodClientTestingUtil` contains this method.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExternalAddressRoutingTest.java
```diff
@@ -0,0 +1,194 @@
...
+            .consistentHashFactory(new ControlledConsistentHashFactory(1));
+
+      server1 = addHotRodServer();
+      server2 = addHotRodServer();
+
+      blockUntilViewReceived(manager(0).getCache(), 2);
+      blockUntilCacheStatusAchieved(manager(0).getCache(), ComponentStatus.RUNNING, 10000);
+      blockUntilCacheStatusAchieved(manager(1).getCache(), ComponentStatus.RUNNING, 10000);
+
+      StaticTcpProxyConfig proxyToServer1Configuration = new StaticTcpProxyConfig(PROXY_TO_SERVER_1_PORT, "localhost", server1.getPort());
```
galderz: Wondering if these proxy ports should not be generated randomly making sure the ports are available, otherwise if we run this in a system where ports `9989` or `9988` are taken it would fail.

tristantarrant:
I do not like the concept in here. It should not be the responsibility of the client to perform any kind of mapping/natting. This should either be done by the server itself or by any proxy/NAT sitting in the middle.

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod.adoc
slaskawi: Definitely not! This is something that depends on the environment very heavily. 

For example in Kubernetes and OpenStack there is no way to extract such mapping. This is why I came up with https://github.com/slaskawi/external-ip-proxy which exposes a REST service and allows it.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExternalAddressRoutingTest.java
slaskawi: If we had tests running concurrently than yes :) Currently they are synchronous.

slaskawi:
@galderz This works the other way around than proxy host. With proxy host the server knows the mapping and can "prepare" consistent hash and server list according to that mapping. So it's complementary implementation rather than something that replaces proxy stuff.

slaskawi:
@tristantarrant @rvansa suggested how to create such proxy in the [mailing list](http://lists.jboss.org/pipermail/infinispan-dev/2017-May/017608.html). Please note that with that scenario you *ARE* doing additional network hop (through the empty Infinispan server which acts as a proxy) and get performance penalty.

I dare to say this is the fastest possible solution/architecture that can be achieved with servers sitting in the cloud and client hosted elsewhere. In any other scenario you have additional proxy and get performance penalty.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExternalAddressRoutingTest.java
slaskawi: Thanks! Fixed

tristantarrant:
I still do not see how why the burden should be on the client to know how to map an "internal" address to an "external" address. If you want to provide different "proxy" addresses to different clients depending on where these are, you should expose multiple hotrod endpoints with different mappings.

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod.adoc
galderz: Hmmmm, I have doubts that unless we provide some implementations that can easily be tailored to the environments, people will actually use this.

As a user, I can get around this issue by having some kind of HTTP entry point in the container environment and from that access the different Hot Rod servers without the need of any of these issues.

My point is: if you want people to use this, they should need to do as little as possible to get it working with the most popular container envs. If I were you, at the very least, I'd try to provide something that can easily be used from Kubernetes.

galderz:
> I do not like the concept in here. It should not be the responsibility of the client to perform any kind of mapping/natting. This should either be done by the server itself or by any proxy/NAT sitting in the middle.

^ That's a interesting POV and something I had not noticed. I'd favour this being done in the server too because then it can be applied to any client since we have multiple client implementations.

slaskawi:
@tristantarrant @galderz So the mapping need to happen somewhere, there is no doubts on that. I even tend to agree that it might happen on the server. But how would you supply this custom address mapper? Shall I create a deployable address mapper?

galderz:
> But how would you supply this custom address mapper? Shall I create a deployable address mapper?

Hmmm, there's no address mapper in this PR for a cloud environment, but I guess the idea is that a server should be able to query something in the cloud that given it's private address or some other detail, it can figure out what the public address that the Hot Rod client will need to connect to.

So, not the address mapper should be deployable really. You could create versions that work for major cloud providers and see how it goes. If people ask for it, make it deployable, but should be last resort.

PR #5166, opened 2017-05-30
Create distribution stage to ensure that -Pdistribution builds pass.

ryanemerson:
A test PR to see how much longer a -Pdistribution` build takes. Currently the distribution build is not tested, which resulted in a few issues with the build not being detected until the release process begain for 9.1.0.Alpha1

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -50,6 +50,22 @@ pipeline {
                 }
             }
         }
+
+        stage('distribution') {
+            when {
+                branch 'master'
+            }
+            steps {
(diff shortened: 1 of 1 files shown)


galderz:
@ryanemerson What's missing here? Did CI run as expected? Latest build failed for other reasons.

ryanemerson:
@galderz Hmm I'm not sure tbh. The distribution field appears for the build, but it has an execution time of 19ms which can't be right. @slaskawi any ideas?

danberindei:
@ryanemerson Looks like the distribution step never ran because of `when { master }`. Wouldn't it be simpler to run the build step with `-Pdistribution`?

slaskawi:
Yeah sure... that could work!

PR #5169, opened 2017-05-30
ISPN-7880 The HotRod server build fails if test keys already exist

danberindei:
https://issues.jboss.org/browse/ISPN-7880

Delete keystores before attempting to generate them


ryanemerson:

PR #5170, opened 2017-06-01
Documentation update - JCache Maven dependencies

isavin:
ISPN-6295 made infinispan-core a provided dependency of infinispan-jcache. Update documentation to reflect this change.


wburns:
@galderz guessing we can just close this since we want to rework this right?

galderz:
Yes, I've just submitted a PR to fix, which should be included in 9.0.2

PR #5173, opened 2017-06-01
ISPN-7779 State transfer does not work with protobuf encoded entities

anistor:
* we need to ensure __protobuf_metadata cache is started and ST complete before any cache using protobuf gets started

https://issues.jboss.org/browse/ISPN-7779

### DIFF
--- a/remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/LifecycleManager.java
+++ b/remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/LifecycleManager.java
@@ -117,8 +117,12 @@ public final class LifecycleManager extends AbstractModuleLifecycle {
     */
    @Override
    public void cacheStarting(ComponentRegistry cr, Configuration cfg, String cacheName) {
-      InternalCacheRegistry icr = cr.getGlobalComponentRegistry().getComponent(InternalCacheRegistry.class);
+      GlobalComponentRegistry gcr = cr.getGlobalComponentRegistry();
+      InternalCacheRegistry icr = gcr.getComponent(InternalCacheRegistry.class);
       if (!icr.isInternalCache(cacheName)) {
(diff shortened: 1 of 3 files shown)


tristantarrant:
I think that any Registry cache should behave like this.

gustavonalle:
@tristantarrant +1, we are discussing that the internal caches should always starts before the others

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/LifecycleManager.java
```diff
@@ -117,8 +117,12 @@ private void unregisterProtobufMetadataManagerMBean(GlobalComponentRegistry gcr)
     */
    @Override
    public void cacheStarting(ComponentRegistry cr, Configuration cfg, String cacheName) {
-      InternalCacheRegistry icr = cr.getGlobalComponentRegistry().getComponent(InternalCacheRegistry.class);
+      GlobalComponentRegistry gcr = cr.getGlobalComponentRegistry();
+      InternalCacheRegistry icr = gcr.getComponent(InternalCacheRegistry.class);
       if (!icr.isInternalCache(cacheName)) {
+         ProtobufMetadataManagerImpl protobufMetadataManager = (ProtobufMetadataManagerImpl) gcr.getComponent(ProtobufMetadataManager.class);
+         protobufMetadataManager.addCacheDependency(cacheName);
+
```
gustavonalle: Shouldn't you force the start of the ProtobufMetadata cache somewhere?

anistor: that happens inside ```addCacheDependency```

gustavonalle: right, overlooked that

anistor:
Updated. Made ProtobufMetadataManagerImpl.protobufSchemaCache volatile.

anistor:
Should we add a new Jira to ensure internal caches are started before all other caches?

gustavonalle:
merged, thanks

PR #4873, opened 2017-02-21
ISPN-6645 Scattered cache

rvansa:
Depends on #4862

Currently it does not have a special handling for orderly operations (e.g. node join) and handles state transfer as if node crashed. package-info includes other possible optimizations, but this PR is already big enough.

PR #5175, opened 2017-06-02
ISPN-7884 StackOverflowException caused by GetAllCommand

rvansa:
https://issues.jboss.org/browse/ISPN-7884

I've also removed some parts of PartititionHandlingInterceptor that don't make much sense with the overall picture of lost data handling in mind.
Note that this also addresses problem similar to https://issues.jboss.org/browse/ISPN-7889 for `GetAllCommand`.


rvansa:
When this will get in, I'll send PR for ISPN-7871 https://github.com/rvansa/infinispan/tree/ISPN-7871

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
```diff
@@ -223,7 +225,7 @@ private Object handleReadCommandReturn(InvocationContext rCtx, VisitableCommand
       } else if (ce instanceof AllOwnersLostException) {
          if (trace)
             getLog().tracef("All owners for command %s have been lost.", cmd);
-         return null;
+         return rCommand.acceptVisitor(rCtx, LostDataVisitor.INSTANCE);
```
danberindei: This got me thinking... prior to ISPN-5021, reading a lost key would put an empty entry in the context, but now we just return the `null`, and a future read in the same transaction may return a non-null value.

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
```diff
@@ -160,6 +166,36 @@ public Object visitReadWriteKeyCommand(InvocationContext ctx, ReadWriteKeyComman
...
+   }
+
+   @Override
+   public Object visitReadWriteManyCommand(InvocationContext ctx, ReadWriteManyCommand command) throws Throwable {
+      return handleWriteCommand(ctx, command);
+   }
+
+   @Override
+   public Object visitReadWriteManyEntriesCommand(InvocationContext ctx, ReadWriteManyEntriesCommand command) throws Throwable {
+      return handleWriteCommand(ctx, command);
```
danberindei: Separate commit?

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
```diff
@@ -140,119 +131,90 @@ public final Object visitGetCacheEntryCommand(InvocationContext ctx, GetCacheEnt
...
+      }
 
-         postOperationPartitionCheck(rCtx, dataCommand, dataCommand.getKey());
-      });
+      if (performPartitionCheck(rCtx, dataCommand)) {
+         // We do the availability check after the read, because the cache may have entered degraded mode
+         // while we were reading from a remote node.
+         partitionHandlingManager.checkRead(dataCommand.getKey());
+      }
+      // TODO We can still return a stale value if the other partition stayed active without us and we haven't entered degraded mode yet.
```
danberindei: The only way to fix that would be to always expect a response from a majority of owners, and I don't see us doing that in the foreseeable future, so please remove the `TODO` tag from the comment :)

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
```diff
@@ -140,119 +131,90 @@ public final Object visitGetCacheEntryCommand(InvocationContext ctx, GetCacheEnt
...
+         // We do the availability check after the read, because the cache may have entered degraded mode
+         // while we were reading from a remote node.
+         for (Object key : getAllCommand.getKeys()) {
+            partitionHandlingManager.checkRead(key);
          }
 
-         // TODO We can still return a stale value if the other partition stayed active without us and we haven't entered degraded mode yet.
-      });
+         Map<Object, Object> result = ((Map<Object, Object>) rv);
+         if (result.size() != getAllCommand.getKeys().size()) {
```
danberindei: This check relies on the values not being in the map, which contradicts the comment here: https://github.com/infinispan/infinispan/pull/5175/files#diff-600cdcc892ffcedd5e510f92706ee4a1R278

core/src/test/java/org/infinispan/partitionhandling/NumOwnersNodeCrashInSequenceTest.java
```diff
@@ -188,40 +178,10 @@ public boolean isSatisfied() throws Exception {
...
-            viewMembers.add(((JGroupsAddress) a).getJGroupsAddress());
-      int viewId = where[0].getTransport().getViewId() + 1;
-      View view = View.create(viewMembers.get(0), viewId, viewMembers.toArray(new org.jgroups.Address[viewMembers.size()]));
-
-      log.trace("Before installing new view:" + viewMembers);
-      for (EmbeddedCacheManager ecm : where) {
-         JChannel c = ((JGroupsTransport) ecm.getTransport()).getChannel();
-         ((GMS) c.getProtocolStack().findProtocol(GMS.class)).installView(view);
-      }
+      TestingUtil.installNewView(members.stream().filter(a -> !a.equals(missing)), where);
```
danberindei: No need to keep the test-private method.

core/src/test/java/org/infinispan/partitionhandling/NumOwnersNodeCrashInSequenceTest.java
```diff
@@ -188,40 +178,10 @@ public boolean isSatisfied() throws Exception {
...
-            channel.getProtocolStack().insertProtocol(discard, ProtocolStack.Position.ABOVE, TP.class);
-         } catch (Exception e) {
-            log.warn("Problems inserting discard", e);
-            throw new RuntimeException(e);
-         }
-         View view = View.create(channel.getAddress(), 100, channel.getAddress());
-         ((GMS) channel.getProtocolStack().findProtocol(GMS.class)).installView(view);
-      }
-      TestingUtil.killCacheManagers(cacheManagers);
+      TestingUtil.crashCacheManagers(cacheManagers);
```
danberindei: Same here.

core/src/test/java/org/infinispan/statetransfer/ReadAfterLostDataTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+   public void testRemove() throws Exception {
+      test(ReadAfterLostDataTest::remove, true, false);
+   }
+
+   public void testReplace() throws Exception {
+      test(ReadAfterLostDataTest::replace, true, false);
+   }
+
+   //TODO: We don't test put/remove/replace before topology update as with interceptor these commands
```
danberindei: with triangle?

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
rvansa: That's a valid point. We should wrap the null into context, set skip lookup and add non-existent read version. This is a job for EWI, though, not STI.
And I wonder how to adapt scattered cache for that; In scattered mode the AllOwnersLostException usually leads to retrying with a new topology, because it is thrown any time the primary crashes but we haven't lost the data itself (we just don't know where it is). That's why we cannot wrap & fix null in EWI when we see AOLE.
An alternative would be doing a broadcast from ScatteredDistributionInterceptor, but that would partially duplicate the job of PrefetchInterceptor.
Any way, please log a JIRA and assign it to me.

core/src/test/java/org/infinispan/statetransfer/ReadAfterLostDataTest.java
```diff
@@ -0,0 +1,301 @@
...
+   // invoke RpcManager.sendTo that does not throw an exception when the target is not in view anymore.
+   // These commands rely on next topology (that is blocked) causing the OutdatedTopologyException.
+   // We'd need to execute them in parallel, wait until all of them use RpcManager.sendTo and then unblock
+   // the topology change. That's rather too much tied to the actual implementation.
```
danberindei: I'm not sure what the "after" tests actually check... if the new topology was already installed, then the leavers are no longer in the cluster, and the lost keys now have proper owners.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
danberindei: Done, https://issues.jboss.org/browse/ISPN-7923

Don't know about the scattered cache, but for the regular cache, I think it might be simpler to have the distribution interceptor call into the `PartitionHandlingManager` and let it decide what to do when all the owners were lost (like `TxDistributionInterceptor` does for tx commands).

rvansa: Uh, didn't know about that piece of code. This looks to me as another spaghetti-style breakage of the separation of concerns.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -378,55 +381,157 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
+      public void accept(Map<Address, Response> responseMap, Throwable throwable) {
+         if (throwable != null) {
+            allFuture.completeExceptionally(throwable);
+            return;
+         }
+         Response response = getSingleReadResponseOrFail(responseMap, allFuture);
+         if (response == null) {
+            if (!allFuture.isDone()) {
+               // When we did not get any successful response but the future was not exceptionally completed
+               // that means that the node got suspected and we have to contact other owners
```
rvansa: It's not that easy as using staggered gets - you don't want to send `ClusteredGetAllCommand(keys.filter(k -> isOwner(k) == address))` to each node, because that way you would get the values multiple times. If sending a subset of keys to one node fails, you have to contact multiple other nodes...

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -378,55 +381,157 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
+         } else {
+            allFuture.completeExceptionally(new IllegalStateException("Unexpected response value: " + responseValue));
+         }
+      }
+
+      private void handleMissingResponse() {
+         GlobalTransaction gtx = ctx.isInTxScope() ? ((TxInvocationContext) ctx).getGlobalTransaction() : null;
+
+         LocalizedCacheTopology cacheTopology = checkTopologyId(command);
+         Map<Address, List<Object>> requestedKeys = getKeysByOwner(ctx, keys, cacheTopology, null, contactedNodes);
```
rvansa: You're probably right.. this needs a test, though.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -378,55 +381,157 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
+            clusteredGetAllCommand.setTopologyId(command.getTopologyId());
+            List<Address> contactedNodes = new ArrayList<>(this.contactedNodes.size() + 1);
+            contactedNodes.addAll(this.contactedNodes);
+            contactedNodes.add(pair.getKey());
+            keys.removeAll(pair.getValue());
+            rpcManager.invokeRemotelyAsync(Collections.singleton(pair.getKey()), clusteredGetAllCommand, syncIgnoreLeavers)
+                  .whenComplete(new ClusteredGetAllHandler(allFuture, ctx, command, pair.getValue(), contactedNodes));
+         }
+         if (!keys.isEmpty()) {
+            // current keys don't have any owner left, we have to remove them
```
rvansa: When the response for the clustered get, the entries are wrapped into context. The GetAllCommand.perform then reads those entries and puts the values into the map. If we don't get any response, it's not wrapped, and it's illegal to not find an entry in context when the command is executed.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -493,32 +601,76 @@ private Object handleLocalOnlyReadManyCommand(InvocationContext ctx, VisitableCo
...
+            allFuture.completeExceptionally(throwable);
+            return;
+         }
+         Response response = getSingleReadResponseOrFail(responseMap, allFuture);
+         if (response == null) {
+            if (!allFuture.isDone()) {
+               try {
+                  handleMissingResponse();
+               } catch (Throwable t) {
+                  allFuture.completeExceptionally(t);
```
rvansa: Should I add the comment? It's not duplicate code, this is calling different method... but ok, I could think about de-duplication.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -493,32 +601,76 @@ private Object handleLocalOnlyReadManyCommand(InvocationContext ctx, VisitableCo
...
+            ReplicableCommand remoteCommand = helper.copyForRemote(command, keysForAddress, ctx);
+            Set<Address> target = Collections.singleton(addressKeys.getKey());
+            List<Address> contactedNodes = new ArrayList<>(this.contactedNodes.size() + 1);
+            contactedNodes.addAll(this.contactedNodes);
+            contactedNodes.add(addressKeys.getKey());
+            rpcManager.invokeRemotelyAsync(target, remoteCommand, syncIgnoreLeavers)
+                  .whenComplete(new ReadManyHandler(allFuture, ctx, command, keysForAddress, contactedNodes, pos, helper));
+            pos += keysForAddress.size();
+         }
+         for (int i = destinationIndex + keys.size() - 1; i >= pos; --i) {
```
rvansa: why not? I thing the idea here is not calling keys.size() multiple times... maybe excessive optimisation, but it doesn't hurt.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
danberindei: IDK... the distribution interceptor already stores the value in the context if the remote get was successful, why not give it the means to decide if the remote get was successful properly?

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -493,32 +601,76 @@ private Object handleLocalOnlyReadManyCommand(InvocationContext ctx, VisitableCo
...
+            allFuture.results[i] = LOST_PLACEHOLDER;
```
rvansa: Yes; `null` value is a regular function return value, we can't filter the fixed-size results array for that. And we can't return `null`s for missing entries like nothing happened, since user might expect that the return value from his function is always an instance of `Foo`.

rvansa: I should document that if the entry is known to be lost, the `Traversable` returned by func API might be missing an element.

danberindei: I would treat "known lost" keys the same as non-existent keys.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -833,7 +1005,7 @@ public void applyLocalResult(MergingCompletableFuture allFuture, Object rv) {
 
       @Override
       public Object transformResult(Object[] results) {
-         return Arrays.stream(results);
+         return Arrays.stream(results).filter(o -> o != LOST_PLACEHOLDER);
```
rvansa: The return value inside the interceptor stack is a `Stream`, so no position involved; if the user wants to make the mapping from keys to returned values addressable, he should add the key to the returned value. Maybe this is suboptimal but that's how Galder designed the Traversable.

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
rvansa: Ok, we can move it into https://github.com/infinispan/infinispan-designs/blob/master/Consistency-guarantees-in-Infinispan.asciidoc

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingInterceptor.java
rvansa: That comment addresses the situation when the entry is available but the value is not in the cache; in that case the map contains the key -> null mapping. The mapping is not in the map when some of the keys are actually non-available as all owners have been lost.

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
rvansa: you mean PR? Ok, if you insist...

core/src/test/java/org/infinispan/partitionhandling/NumOwnersNodeCrashInSequenceTest.java
rvansa: this one is overridden in NumOwnersNodeStopInSequenceTest

core/src/test/java/org/infinispan/statetransfer/ReadAfterLostDataTest.java
rvansa: yes

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: You could do that, but not in Distribution interceptor. The PHI has to get some kind of information that the keys were not there.

rvansa: Not that functional commands would be treated properly by PHI, but that's another PR.

rvansa:
Have you removed a comment starting with "IDK..."? Entry wrapped into context with some .isUnavailable()" flag really is an option, though I am not convinced it would simplify things. You can't use that for functional commands which don't wrap anything into context.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: Right, the `handleMissingResponse()` method is different... still, IntelliJ is telling me the code is duplicated :)

It would be nice to have a single place where that comment is required, but I wouldn't mind if you just copy it here.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: This is one thing I would trust HotSpot to optimize away, hoisting the computation before the loop.

I guess it's a matter of style, but the descending loop is surprising to me, especially coming after an ascending loop using the same array. So every time I was reading the code, I would stop for a second and ask myself why it's different :)

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: Thanks for the explanation, I keep thinking in terms of returning a map.

danberindei:
Ok, I think it's pretty clear from my comments that I don't like the retry in `BaseDistributionInterceptor` that you fought so hard to remove for the single-key read commands :)

IMO there are 2 simpler options:

1. Have `BaseDistributionInterceptor` call a `PartitionHandlingManager` method when a key doesn't have any owners in the cluster view, and wrap a `null` in the context if the method doesn't throw any exception. That still requires some logic in BDI to send the request to another owner if the primary is not in the cluster view.
2. Throw an `OutdatedTopologyException` whenever one of the contacted nodes is missing, but only after waiting for the responses from the other nodes and putting their values in the context.

rvansa:
What I was fighting against is checking for a different topology; everything below STI should be done using single topology. The code here (though annoyingly complex) does not retry; it just gets multiple options where to get the value from and tries them in a sequence interrupted on success.

1. This still does the retry; what are the pros? The only thing you've *simplified* is not narrowing the command to be sent to {{CallInteceptor}}. And the con is mixing responsibilities.
2. That does not sound that bad, but you'd be waiting for the new topology. I think you wanted to avoid waiting during reads in the past.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -378,55 +381,157 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
-            if (throwable != null) {
-               allFuture.completeExceptionally(throwable);
-            }
-            if (responseMap.isEmpty()) {
-               allFuture.completeExceptionally(OutdatedTopologyException.INSTANCE);
+         Set<Address> target = Collections.singleton(pair.getKey());
+         rpcManager.invokeRemotelyAsync(target, clusteredGetAllCommand, syncIgnoreLeavers)
+               .whenComplete(new ClusteredGetAllHandler(allFuture, ctx, command, pair.getValue(), target));
+      }
+      return asyncValue(handleRemotelyRetrievedKeys(ctx, allFuture, allFuture)).thenApply(ctx, command, allFuture);
```
rvansa: actually this is wrong; we need to handle remotely retrieved keys only *after* we retrieve them

core/src/test/java/org/infinispan/partitionhandling/NumOwnersNodeCrashInSequenceTest.java
rvansa: `missing` is not final in the places this method is used, I am keeping this

core/src/main/java/org/infinispan/statetransfer/StateTransferInterceptor.java
danberindei: I really meant a separate commit in the same PR, but a separate PR is also fine.

core/src/test/java/org/infinispan/statetransfer/ReadAfterLostDataTest.java
danberindei: I meant should it be "with triangle" instead of "with interceptor"?

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: Good point :)

rvansa:
@danberindei Hi, I've addressed some of the comments and fixed the issue with not asking an already asked node, adding another test to confirm that all works as expected.
I've also refactored some methods used to handle the response and moved them to `ClusteringInterceptor`. Having them down there (as opposed to `BaseDistributionInterceptor`) may seem unnecessary, but I need those methods there for reuse in `ScatteredDistributionInterceptor`, and since I was fixing multiple issues in parallel for scattered cache I've left everything ready for scattered cache.
I've kept the *secondary retrievals* from `handleMissing` methods (because those are not *retries*), though.

danberindei:
@rvansa I don't remember removing any comment, but I may have modified it. Do you still have the notification email?

You're right, I'm not entirely satisfied with my proposals either.

With 1), I just want to make the connection between the distribution interceptor and partition handling clearer, the same it is with entry wrapping. The way I see it, the distribution interceptor already has to know about partition handling, only now needs an OOB signal to the PHI by making the list of results shorter than the list of keys. I'd like us to treat all missing keys the same way, regardless of whether they have owners or not, and use a different communication mechanism between BDI and PHM. If you don't like a direct call and functional commands don't update the context, maybe a `someKeysWereLost` flag on the command itself could work?

Only 2) actually addressed the retry... and indeed, because we're only contacting the first node, the retry can't happen immediately. But don't we already wait for a new topology if the primary owner has a newer topology and returns an `UnsureResponse`? 

You're probably right, it is better to contact the backups before giving up and throwing an `OutdatedTopologyException`, but then I'd argue we need to do the same for `UnsureResponse`. And I'd also like to contact the backups only after we got a reply from all the targets: it slows the retry a bit, but I really don't want to worry about multiple threads retrying in parallel.

rvansa:
@danberindei 
2) We throw OTE if `UnsureResponse` is received and there's no other `SuccessfulResponse`. Then we retry from STI immediately - the 4 phase topology change makes sure that at all times there's at least one node that can respond immediately (though due to timing you may have to do the retry).

1) From design point of view it would be best if we could pass various types of 'Responses' through the whole stack, but it wouldn't work that well on hot path. Adding fields to Get*Commands or to the context would have GC impact, too, but I agree that it would be a bit nicer. I am just afraid that it won't be general enough and we'd soon find a use case when that would be insufficient.

I don't think that 'big' changes are within the scope of this bugfix PR.

danberindei:
2) For GetAll/ReadMany we only contact the primary owner, so there's only going to be one successful/unsure response.
1) I don't want to make big changes either, but considering that on master we don't have any way to signal lost entries to the PHI, we have to make *some* changes, and I'd like the changes to be as clear as possible.

rvansa:
2) You're right, GetAll is retrying the whole operation on the first UnsureResponse. I could attempt fetching the value from other owners before throwing if you want, to make this symmetric with single-entry reads. In either case we wouldn't wait for the next topology and retry immediatelly.

3) OK. I made the change by checking keys vs. map/stream (the part with stream is nasty!). So I assume that you want to go with some flag being raised on `GetKeyValueCommand`, rather than throwing `AllOwnersLostException`, and a collection of missed keys in `GetAllCommand`? Because it can't be 100% symmetric, flag alone is not sufficient for GetAllCommand in current state.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
danberindei: Hey, there was a comment starting with IDK here :)

danberindei:
TBH I was thinking of waiting for a new topology all the time, but your proposal sounds good too :) 

I would like to do the fetching from other owners in `ClusteredGetAllFuture.apply()` and not in `ClusteredGetAllHandler.accept()`, because I'm already seeing `ClusteredGetAllHandler` as a `ResponseCollector`, but it's probably best if I make that change myself once `ResponseCollector` is available.

I was also thinking that `PartitionHandlingInterceptor` is already pretty inconsistent when reporting which keys are unavailable, so reporting all of `GetAllCommand`'s keys all the time wouldn't be a big regression... Still, a collection is the right thing to do.

As for single-key reads, I didn't really want to change their handling in this PR. But I changed my mind when I saw `PessimisticLockingInterceptor.rethrowAndReleaseLocksIfNeeded()` releases all tx locks on an `AllOwnersLostException`.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
rvansa: So do you propose to get rid of PHI completely and solving everything in xDI?

danberindei: Last night I didn't think so, but now I'm leaning towards putting all the checks in DI again:

https://issues.jboss.org/browse/ISPN-7956

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -767,6 +784,9 @@ public MergingCompletableFuture(InvocationContext ctx, int participants, T[] res
 
       @Override
       protected Object result() {
+         if (hasUnsureResponse && lostData) {
```
danberindei: I could use a comment here, I feel like it should be `||` instead of `&&`.

rvansa: No, this is correct; If we've lost data it shouldn't throw but return limited stream. If we've got unsure response - hakuna matata, there has been another response delivering the results. Only if those two combine we'll rather retry.

rvansa:
I am also glad to arrive to the final conclusion that PHI sucks anyway and we should move its responsibilities to xDi :) Now let's merge #4873 :)

gustavonalle:
@danberindei @rvansa There are test failures in the query suite
http://ci.infinispan.org/job/Infinispan/job/PR-5175/42/
http://ci.infinispan.org/job/Infinispan/job/PR-5175/43/

core/src/main/java/org/infinispan/interceptors/impl/ClusteringInterceptor.java
```diff
@@ -114,6 +114,9 @@ public synchronized boolean completeExceptionally(Throwable ex) {
       @Override
       public Object apply(InvocationContext rCtx, VisitableCommand rCommand, Object rv) throws Throwable {
          assert rv == null; // value with which the allFuture has been completed
+         if (hasUnsureResponse) {
```
rvansa: @danberindei I think I've screwed this one as well; Here should be the check for lost data as well.

danberindei: @rvansa Seeing some new test failures?

rvansa: No, I just went through the code and realized that.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: @danberindei this is weird; I can't find this comment on master. And I can't find ed7cf2b879d7b14e9a8f3179386ed50f9e274dcc even in my local history/branches.

danberindei: @rvansa Sorry, I think I forgot to fetch the branch again before integrating.

PR #5179, opened 2017-06-05
ISPN-6676 ISPN-7891 HTTP/2 with TLS/ALPN upgrade

slaskawi:
https://issues.jboss.org/browse/ISPN-6676
https://issues.jboss.org/browse/ISPN-7891

This is the first part of HTTP/2 support. A full to do list looks like the following:
* [x] Implement HTTP/2 support https://issues.jboss.org/browse/ISPN-6676
* [x] Implement upgrade to HTTP/2 using TLS/ALPN negotiation https://issues.jboss.org/browse/ISPN-7891
* [x] Implement upgrade to HTTP/2 using HTTP/1.1 Upgrade header https://issues.jboss.org/browse/ISPN-7892
* [x] Benchmark HTTP/1.1 vs HTTP/2.0 https://issues.jboss.org/browse/ISPN-7893

What's worth to mention is that this implementation doesn't need any statically linked C libraries or Jetty's ALPN Agent!


slaskawi:
The benchmark for HTTP/1.1 vs HTTP/2 is ready. Please note that this benchmark is obviously wrong for many reasons (it has been ran locally, it doesn't involve full server distribution etc). It is also wrong to assume that HTTP/2 is just a new version of HTTP/1.1. It has been rebuilt from the scratch and comparing HTTP/1.1 to HTTP/2 is always like comparing apples to bananas. So please take all the results with a huge grain of salt!

```
Benchmark                                           (httpClientThreads)  (useHttp2)  (useTLS)  Mode  Cnt   Score     Error  Units
Http11VsHttp20Benchmark.BenchmarkState.measure_get                    1        true      true  avgt   10   0.504 ±   0.031  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_get                    1        true     false  avgt   10   0.485 ±   0.020  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_get                    1       false      true  avgt   10   3.391 ±   1.366  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_get                    1       false     false  avgt   10   1.752 ±   0.075  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_get                    2       false      true  avgt   10   1.454 ±   0.069  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_get                    2       false     false  avgt   10   0.836 ±   0.020  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_get                    4       false      true  avgt   10   1.214 ±   0.013  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_get                    4       false     false  avgt   10   0.746 ±   0.020  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_get                    8       false      true  avgt   10   1.711 ±   0.357  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_get                    8       false     false  avgt   10   0.968 ±   0.079  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_put                    1        true      true  avgt   10   0.531 ±   0.016  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_put                    1        true     false  avgt   10   0.511 ±   0.020  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_put                    1       false      true  avgt   10   3.097 ±   0.208  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_put                    1       false     false  avgt   10   1.673 ±   0.119  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_put                    2       false      true  avgt   10   1.508 ±   0.083  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_put                    2       false     false  avgt   10   0.844 ±   0.014  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_put                    4       false      true  avgt   10   1.370 ±   0.221  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_put                    4       false     false  avgt   10   0.818 ±   0.007  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_put                    8       false      true  avgt   10   1.683 ±   0.219  ms/op
Http11VsHttp20Benchmark.BenchmarkState.measure_put                    8       false     false  avgt   10   0.997 ±   0.239  ms/op
```

Configuration details:

- Netty client doesn't like sending requests from multiple threads. It leads to errors in stream ordering. Since HTTP/2 client is async by its nature, I decided to benchamrk HTTP/2 only with 1 thread.
- HTTP/2 client uses OpenSSL for TLS/ALPN whereas HTTP/1.1 uses JSSE.

Findings:

- Without TLS, HTTP/2 produced 50% better results with `GET` opposed to the best HTTP/1.1 results (which was for 4 threads). `PUT`s were 40% faster.
- HTTP/1.1 produced best results with 4 threads in general. My guess is that with 8 the contention was too high. This aligns with an idea that most of default browser configurations use 6 threads for HTTP/1.1.
- HTTP/2 client with OpenSSL produced ~60% better results opposed to the best HTTP/1.1 result (which uses JSSE). It is worth to mention that the results was almost as fast as HTTP/2 client without TLS. This suggests that we might get good benefits by recommending clients to use OpenSSL and we can stick with JSSE for some time.

And finally, some bad news... Most of the HTTP/2 client implementation I tested (OkHttp, Jetty, Netty) have poor to none support for TLS/ALPN. Most of them require Jetty Agent (which needs to be downloaded individually for each JVM version). Only Netty had support for OpenSSL which worked out of the box. But the downside is that Netty requires a tremendous amount of boilerplate code, which I copied from Netty examples (the client supports HTTP/1.1 Upgrade and ALPN). I suppose we can wait till ALPN gets into mainstream and replace it with some other implementation. Otherwise I think it makes sense to start a new Github project with Netty-based HTTP/1.1 and HTTP/2 client.

//cc @tristantarrant @gustavonalle

slaskawi:
@tristantarrant [The boring SSL package has been added in scope test](https://github.com/infinispan/infinispan/pull/5179/files#diff-bbfa0d3456d94ae96cb2f93d77120666R48) and is required for Netty HTTP/2 **client**. The client also needs to be aware of ALPN extension or it won't send a list of application protocols during the handshake.

The server doesn't use OpenSSL at all. It uses `ALPNHackSSLEngine` instead (big bravo to @stuartwdouglas for the implementation), which extends JSSE `SSLEngine` and is capable of negotiating application protocols using ALPN. This way we do not need OpenSSL nor Jetty Agent. It *just* works.

As we have been discussing [on the mailing list](http://lists.jboss.org/pipermail/infinispan-dev/2017-June/017621.html), it might make sense to try migrating into OpenSSL at some point but it's definitely out of the scope for this task.

tristantarrant:
Ok, then amend documentation/src/main/asciidoc/user_guide/server_protocols_rest.adoc to document the supported protocols

slaskawi:
@tristantarrant updated

tristantarrant:
Merged, thanks @slaskawi

PR #5180, opened 2017-06-05
ISPN-7895 ArrayIndexOutOfBoundsException when using off heap with

wburns:
expiration

* Typo with variables fixed
* Offset typo as well
* Added tests for all 3 modes of expiration with and without eviction

https://issues.jboss.org/browse/ISPN-7895

### DIFF
--- a/core/src/main/java/org/infinispan/container/offheap/OffHeapEntryFactoryImpl.java
+++ b/core/src/main/java/org/infinispan/container/offheap/OffHeapEntryFactoryImpl.java
@@ -95,21 +95,22 @@ public class OffHeapEntryFactoryImpl implements OffHeapEntryFactory {
             metadataBytes = new byte[16 + versionBytes.length];
             Bits.putLong(metadataBytes, 0, lifespan);
             Bits.putLong(metadataBytes, 8, timeService.wallClockTime());
-            System.arraycopy(metadataBytes, 0, versionBytes, 16, versionBytes.length);
+            System.arraycopy(versionBytes, 0, metadataBytes, 16, versionBytes.length);
          } else if (lifespan < 0 && maxIdle > -1) {
             type |= TRANSIENT;
             metadataBytes = new byte[16 + versionBytes.length];
             Bits.putLong(metadataBytes, 0, maxIdle);
             Bits.putLong(metadataBytes, 8, timeService.wallClockTime());
-            System.arraycopy(metadataBytes, 0, versionBytes, 16, versionBytes.length);
+            System.arraycopy(versionBytes, 0, metadataBytes, 16, versionBytes.length);
          } else {
             type |= TRANSIENT_MORTAL;
             metadataBytes = new byte[32 + versionBytes.length];
             Bits.putLong(metadataBytes, 0, maxIdle);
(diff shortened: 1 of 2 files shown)


wburns:
This should cherry pick very easily into 9.0.x as well

gustavonalle:
There's a test failure on ```org.infinispan.container.offheap.OffHeapSingleNodeExpirationEvictionTest.testClear```

wburns:
Weird, could have sworn the core test suite all passed for me. I will take
a look.

wburns:
Ah it was because I didn't have the @Test annotation on the test and thus they were sharing a test name between tests and could form a cluster, thus affecting each others tests when ran in parallel.

wburns:
Should be fixed now.

core/src/main/java/org/infinispan/container/offheap/OffHeapEntryFactoryImpl.java
```diff
@@ -95,21 +95,22 @@ public long create(WrappedBytes key, WrappedBytes value, Metadata metadata) {
...
             metadataBytes = new byte[32 + versionBytes.length];
             Bits.putLong(metadataBytes, 0, maxIdle);
             Bits.putLong(metadataBytes, 8, lifespan);
-            Bits.putLong(metadataBytes, 16, timeService.wallClockTime());
-            Bits.putLong(metadataBytes, 24, timeService.wallClockTime());
-            System.arraycopy(metadataBytes, 0, versionBytes, 16, versionBytes.length);
+            long time = timeService.wallClockTime();
+            Bits.putLong(metadataBytes, 16, time);
+            Bits.putLong(metadataBytes, 24, time);
+            System.arraycopy(versionBytes, 0, metadataBytes, 32, versionBytes.length);
```
wburns: Yes it could be. If you want, youcan log a JIRA. I figured we have 8 bits and we aren't even close to using them all so it would be more explicit if they each had their own. (but either way is fine with me).

core/src/test/java/org/infinispan/container/offheap/OffHeapSingleNodeExpirationEvictionTest.java
```diff
@@ -0,0 +1,67 @@
...
+            defaultConfig.expiration().maxIdle(10, TimeUnit.MINUTES);
+            break;
+         case TRANSIENT_MORTAL:
+            defaultConfig.expiration().lifespan(10, TimeUnit.MINUTES).maxIdle(10, TimeUnit.MINUTES);
+            break;
+      }
+      if (eviction) {
+         defaultConfig.memory().evictionType(EvictionType.COUNT).size(1000);
+      }
+      return super.addClusterEnabledCacheManager(defaultConfig);
```
wburns: The problem I had when I made these is that this was only sharing test methods through extension and I didn't want to have to have another line of hiearchy (with MultipleCacheManager and SingleCacheManager test classes in the middle). Do we have a recommended way of doing this now? I haven't looked through the test suite or thought about it enough to know.

tristantarrant:
Pushed to master and 9.0.x

PR #5184, opened 2017-06-07
ISPN-5218 Add batching to the CacheWriter Interface

ryanemerson:
https://issues.jboss.org/browse/ISPN-5218
https://issues.jboss.org/browse/ISPN-7901

I have added the `writeBatch` and `deleteBatch` methods to the `CacheWriter` interface opposed to the `AdvancedCacheWriter` as the `DelegatingCacheWriter` implements the former. Both of these are implemented as default methods in the interface, which fall back to sequentially calling write/delete if a store is unable to perform batch write and/or delete operations.

### DIFF
--- a/core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
+++ b/core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
@@ -8,6 +8,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.infinispan.commons.CacheException;
+import org.infinispan.commons.util.IteratorMapper;
 import org.infinispan.commons.util.Util;
 import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.container.DataContainer;
@@ -44,7 +45,7 @@ public class PassivationManagerImpl implements PassivationManager {
    private DistributionManager distributionManager;
 
    @Inject
-   public void inject(PersistenceManager persistenceManager, CacheNotifier notifier, Configuration cfg, DataContainer container,
+   public void inject(PersistenceManager persistenceManager,CacheNotifier notifier, Configuration cfg, DataContainer container,
                       TimeService timeService, MarshalledEntryFactory marshalledEntryFactory,
                       DistributionManager distributionManager) {
       this.persistenceManager = persistenceManager;
@@ -99,13 +100,11 @@ public class PassivationManagerImpl implements PassivationManager {
       if (enabled && !skipOnStop) {
          long start = timeService.time();
          log.passivatingAllEntries();
-         long count = 0;
-         for (InternalCacheEntry e : container) {
-            count++;
-            if (trace) log.tracef("Passivating %s", e.getKey());
-            persistenceManager.writeToAllNonTxStores(marshalledEntryFactory.newMarshalledEntry(e.getKey(), e.getValue(),
-                                                                                               internalMetadata(e)), BOTH);
-         }
+
+         int count = container.sizeIncludingExpired();
+         Iterable<MarshalledEntry> iterable = () -> new IteratorMapper<>(container.iterator(), e ->
+            marshalledEntryFactory.newMarshalledEntry(e.getKey(), e.getValue(), internalMetadata(e)));
+         persistenceManager.writeBatchToAllNonTxStores(iterable, BOTH, 0);
          log.passivatedEntries(count, Util.prettyPrintTime(timeService.timeDuration(start, TimeUnit.MILLISECONDS)));
       }
    }
--- a/core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
@@ -5,18 +5,16 @@ import static org.infinispan.persistence.manager.PersistenceManager.AccessMode.B
 import static org.infinispan.persistence.manager.PersistenceManager.AccessMode.PRIVATE;
 
 import java.util.List;
-import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Predicate;
 
 import javax.transaction.InvalidTransactionException;
 import javax.transaction.SystemException;
 import javax.transaction.Transaction;
 import javax.transaction.TransactionManager;
 
-import org.infinispan.commands.AbstractVisitor;
 import org.infinispan.commands.FlagAffectedCommand;
 import org.infinispan.commands.VisitableCommand;
-import org.infinispan.commands.functional.AbstractWriteManyCommand;
 import org.infinispan.commands.functional.FunctionalCommand;
 import org.infinispan.commands.functional.ReadWriteKeyCommand;
 import org.infinispan.commands.functional.ReadWriteKeyValueCommand;
@@ -28,7 +26,6 @@ import org.infinispan.commands.functional.WriteOnlyManyCommand;
 import org.infinispan.commands.functional.WriteOnlyManyEntriesCommand;
 import org.infinispan.commands.tx.CommitCommand;
 import org.infinispan.commands.tx.PrepareCommand;
-import org.infinispan.commands.write.ApplyDeltaCommand;
 import org.infinispan.commands.write.ClearCommand;
 import org.infinispan.commands.write.DataWriteCommand;
 import org.infinispan.commands.write.PutKeyValueCommand;
@@ -42,23 +39,21 @@ import org.infinispan.commons.marshall.StreamingMarshaller;
 import org.infinispan.configuration.cache.PersistenceConfiguration;
 import org.infinispan.container.InternalEntryFactory;
 import org.infinispan.container.entries.CacheEntry;
-import org.infinispan.container.entries.InternalCacheEntry;
 import org.infinispan.container.entries.InternalCacheValue;
-import org.infinispan.container.versioning.EntryVersion;
-import org.infinispan.container.versioning.EntryVersionsMap;
 import org.infinispan.context.InvocationContext;
 import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.context.impl.TxInvocationContext;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.factories.annotations.Start;
+import org.infinispan.interceptors.InvocationSuccessAction;
 import org.infinispan.jmx.annotations.MBean;
 import org.infinispan.jmx.annotations.ManagedAttribute;
 import org.infinispan.jmx.annotations.ManagedOperation;
 import org.infinispan.jmx.annotations.MeasurementType;
+import org.infinispan.marshall.core.MarshalledEntry;
 import org.infinispan.marshall.core.MarshalledEntryImpl;
-import org.infinispan.metadata.EmbeddedMetadata;
-import org.infinispan.metadata.Metadata;
 import org.infinispan.persistence.manager.PersistenceManager;
+import org.infinispan.persistence.support.BatchModification;
 import org.infinispan.transaction.xa.GlobalTransaction;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
@@ -83,6 +78,8 @@ public class CacheWriterInterceptor extends JmxStatsCommandInterceptor {
    private InternalEntryFactory entryFactory;
    private TransactionManager transactionManager;
    private StreamingMarshaller marshaller;
+
+   protected InvocationSuccessAction handlePutMapCommandReturn = this::handlePutMapCommandReturn;
 
    private static final Log log = LogFactory.getLog(CacheWriterInterceptor.class);
 
@@ -209,20 +206,27 @@ public class CacheWriterInterceptor extends JmxStatsCommandInterceptor {
 
    @Override
    public Object visitPutMapCommand(InvocationContext ctx, PutMapCommand command) throws Throwable {
-      return invokeNextThenAccept(ctx, command, (rCtx, rCommand, rv) -> {
-         PutMapCommand putMapCommand = (PutMapCommand) rCommand;
-         if (!isStoreEnabled(putMapCommand) || rCtx.isInTxScope())
-            return;
+      return invokeNextThenAccept(ctx, command, handlePutMapCommandReturn);
+   }
 
-         Map<Object, Object> map = putMapCommand.getMap();
-         for (Object key : map.keySet()) {
-            if (isProperWriter(rCtx, putMapCommand, key)) {
-               storeEntry(rCtx, key, putMapCommand);
-            }
-         }
-         if (getStatisticsEnabled())
-            cacheStores.getAndAdd(map.size());
-      });
+   protected void handlePutMapCommandReturn(InvocationContext rCtx, VisitableCommand rCommand, Object rv) {
+      PutMapCommand putMapCommand = (PutMapCommand) rCommand;
+      if (!isStoreEnabled(putMapCommand) || rCtx.isInTxScope())
+         return;
+
+      processIterableBatch(rCtx, putMapCommand, BOTH, key -> !skipSharedStores(rCtx, key, putMapCommand));
+      processIterableBatch(rCtx, putMapCommand, PRIVATE, key -> skipSharedStores(rCtx, key, putMapCommand));
+   }
+
+   protected void processIterableBatch(InvocationContext ctx, PutMapCommand cmd, PersistenceManager.AccessMode mode, Predicate<Object> filter) {
+      if (getStatisticsEnabled())
+         cacheStores.addAndGet(cmd.getMap().size());
+
+      Iterable<MarshalledEntry> iterable = () -> cmd.getMap().keySet().stream()
+            .filter(filter)
+            .map(key -> createMarshalledEntry(ctx, key))
+            .iterator();
+      persistenceManager.writeBatchToAllNonTxStores(iterable, mode, cmd.getFlagsBitSet());
    }
 
    @Override
@@ -350,14 +354,27 @@ public class CacheWriterInterceptor extends JmxStatsCommandInterceptor {
       if (trace) getLog().tracef("Cache loader modification list: %s", modifications);
 
 
-      Updater modsBuilder = new Updater(getStatisticsEnabled());
+      TxBatchUpdater modsBuilder = TxBatchUpdater.createNonTxStoreUpdater(this, persistenceManager, entryFactory, marshaller);
       for (WriteCommand cacheCommand : modifications) {
          if (isStoreEnabled(cacheCommand)) {
             cacheCommand.acceptVisitor(ctx, modsBuilder);
          }
       }
-      if (getStatisticsEnabled() && modsBuilder.putCount > 0) {
-         cacheStores.getAndAdd(modsBuilder.putCount);
+      BatchModification sharedMods = modsBuilder.getModifications();
+      BatchModification nonSharedMods = modsBuilder.getNonSharedModifications();
+
+      persistenceManager.writeBatchToAllNonTxStores(sharedMods.getMarshalledEntries(), BOTH, 0);
+      persistenceManager.writeBatchToAllNonTxStores(nonSharedMods.getMarshalledEntries(), PRIVATE, 0);
+      persistenceManager.deleteBatchFromAllNonTxStores(sharedMods.getKeysToRemove(), BOTH, 0);
+      persistenceManager.deleteBatchFromAllNonTxStores(nonSharedMods.getKeysToRemove(), PRIVATE, 0);
+
+      if (trace) {
+         getLog().tracef("Writing shared batch with #entries=%d and non-shared batch with #entries=%d", sharedMods.getMarshalledEntries().size(), nonSharedMods.getMarshalledEntries().size());
+         getLog().tracef("Deleting shared batch with #entries=%d and non-shared batch with #entries=%d", sharedMods.getKeysToRemove().size(), nonSharedMods.getKeysToRemove().size());
+      }
+
+      if (getStatisticsEnabled() && modsBuilder.getPutCount() > 0) {
+         cacheStores.getAndAdd(modsBuilder.getPutCount());
       }
    }
 
@@ -371,150 +388,6 @@ public class CacheWriterInterceptor extends JmxStatsCommandInterceptor {
 
    protected boolean isProperWriter(InvocationContext ctx, FlagAffectedCommand command, Object key) {
       return true;
-   }
-
-   public class Updater extends AbstractVisitor {
-
-      protected final boolean generateStatistics;
-      int putCount;
-
-      public Updater(boolean generateStatistics) {
-         this.generateStatistics = generateStatistics;
-      }
-
-      @Override
-      public Object visitPutKeyValueCommand(InvocationContext ctx, PutKeyValueCommand command) throws Throwable {
-         return visitSingleStore(ctx, command, command.getKey());
-      }
-
-      @Override
-      public Object visitApplyDeltaCommand(InvocationContext ctx, ApplyDeltaCommand command) throws Throwable {
-         if (isProperWriter(ctx, command, command.getKey())) {
-            if (generateStatistics) putCount++;
-            CacheEntry entry = ctx.lookupEntry(command.getKey());
-            // If the value is null, there is a subsequent remove operation in the transaction and we can ignore
-            // this modification.
-            if (entry.getValue() == null) {
-               return null;
-            }
-            InternalCacheEntry ice;
-            if (entry instanceof InternalCacheEntry) {
-               ice = (InternalCacheEntry) entry;
-            } else {
-               ice = entryFactory.create(entry);
-            }
-            MarshalledEntryImpl marshalledEntry = new MarshalledEntryImpl(ice.getKey(), ice.getValue(), internalMetadata(ice), marshaller);
-            persistenceManager.writeToAllNonTxStores(marshalledEntry, skipSharedStores(ctx, command.getKey(), command) ? PRIVATE : BOTH);
-         }
-         return null;
-      }
-
-      @Override
-      public Object visitReplaceCommand(InvocationContext ctx, ReplaceCommand command) throws Throwable {
-         return visitSingleStore(ctx, command, command.getKey());
-      }
-
-      @Override
-      public Object visitPutMapCommand(InvocationContext ctx, PutMapCommand command) throws Throwable {
-         Map<Object, Object> map = command.getMap();
-         for (Object key : map.keySet())
-            visitSingleStore(ctx, command, key);
-         return null;
-      }
-
-      @Override
-      public Object visitWriteOnlyKeyCommand(InvocationContext ctx, WriteOnlyKeyCommand command) throws Throwable {
-         return visitModify(ctx, command, command.getKey());
-      }
-
-      @Override
-      public Object visitReadWriteKeyValueCommand(InvocationContext ctx, ReadWriteKeyValueCommand command) throws Throwable {
-         return visitModify(ctx, command, command.getKey());
-      }
-
-      @Override
-      public Object visitReadWriteKeyCommand(InvocationContext ctx, ReadWriteKeyCommand command) throws Throwable {
-         return visitModify(ctx, command, command.getKey());
-      }
-
-      @Override
-      public Object visitWriteOnlyManyEntriesCommand(InvocationContext ctx, WriteOnlyManyEntriesCommand command) throws Throwable {
-         return visitManyModify(ctx, command);
-      }
-
-      @Override
-      public Object visitWriteOnlyKeyValueCommand(InvocationContext ctx, WriteOnlyKeyValueCommand command) throws Throwable {
-         return visitModify(ctx, command, command.getKey());
-      }
-
-      @Override
-      public Object visitWriteOnlyManyCommand(InvocationContext ctx, WriteOnlyManyCommand command) throws Throwable {
-         return visitManyModify(ctx, command);
-      }
-
-      @Override
-      public Object visitReadWriteManyCommand(InvocationContext ctx, ReadWriteManyCommand command) throws Throwable {
-         return visitManyModify(ctx, command);
-      }
-
-      @Override
-      public Object visitReadWriteManyEntriesCommand(InvocationContext ctx, ReadWriteManyEntriesCommand command) throws Throwable {
-         return visitManyModify(ctx, command);
-      }
-
-      @Override
-      public Object visitRemoveCommand(InvocationContext ctx, RemoveCommand command) throws Throwable {
-         Object key = command.getKey();
-         if (isProperWriter(ctx, command, key)) {
-            persistenceManager.deleteFromAllStores(key, skipSharedStores(ctx, key, command) ? PRIVATE : BOTH);
-         }
-         return null;
-      }
-
-      @Override
-      public Object visitClearCommand(InvocationContext ctx, ClearCommand command) throws Throwable {
-         persistenceManager.clearAllStores(ctx.isOriginLocal() ? PRIVATE : BOTH);
-         return null;
-      }
-
-      protected Object visitSingleStore(InvocationContext ctx, FlagAffectedCommand command, Object key) throws Throwable {
-         if (isProperWriter(ctx, command, key)) {
-            if (generateStatistics) putCount++;
-            InternalCacheValue sv = entryFactory.getValueFromCtxOrCreateNew(key, ctx);
-            // TODO: we should merge all modifications and write each key only once
-            // If the value in context is null, the transaction contains a subsequent remove,
-            // so we'll ignore this modification
-            if (sv.getValue() != null) {
-               MarshalledEntryImpl me = new MarshalledEntryImpl(key, sv.getValue(), internalMetadata(sv), marshaller);
-               persistenceManager.writeToAllNonTxStores(me, skipSharedStores(ctx, key, command) ? PRIVATE : BOTH);
-            }
-         }
-         return null;
-      }
-
-      protected Object visitModify(InvocationContext ctx, FlagAffectedCommand command, Object key) throws Throwable {
-         if (isProperWriter(ctx, command, key)) {
-            CacheEntry entry = ctx.lookupEntry(key);
-            if (!entry.isChanged()) {
-               return null;
-            } else if (entry.getValue() == null) {
-               persistenceManager.deleteFromAllStores(key, skipSharedStores(ctx, key, command) ? PRIVATE : BOTH);
-            } else {
-               if (generateStatistics) putCount++;
-               InternalCacheValue sv = entryFactory.getValueFromCtxOrCreateNew(key, ctx);
-               MarshalledEntryImpl me = new MarshalledEntryImpl(key, sv.getValue(), internalMetadata(sv), marshaller);
-               persistenceManager.writeToAllNonTxStores(me, skipSharedStores(ctx, key, command) ? PRIVATE : BOTH);
-            }
-         }
-         return null;
-      }
-
-      protected Object visitManyModify(InvocationContext ctx, AbstractWriteManyCommand command) throws Throwable {
-         for (Object key : command.getAffectedKeys()) {
-            visitModify(ctx, command, key);
-         }
-         return null;
-      }
    }
 
    @Override
@@ -536,43 +409,18 @@ public class CacheWriterInterceptor extends JmxStatsCommandInterceptor {
    }
 
    void storeEntry(InvocationContext ctx, Object key, FlagAffectedCommand command) {
-      InternalCacheValue sv = getStoredValue(key, ctx);
-      persistenceManager.writeToAllNonTxStores(new MarshalledEntryImpl(key, sv.getValue(), internalMetadata(sv), marshaller),
-                                               skipSharedStores(ctx, key, command) ? PRIVATE : BOTH, command.getFlagsBitSet());
-      if (trace) getLog().tracef("Stored entry %s under key %s", sv, key);
+      MarshalledEntry entry = createMarshalledEntry(ctx, key);
+      persistenceManager.writeToAllNonTxStores(entry, skipSharedStores(ctx, key, command) ? PRIVATE : BOTH, command.getFlagsBitSet());
+      if (trace) getLog().tracef("Stored entry %s under key %s", entry.getValue(), key);
+   }
+
+   MarshalledEntry createMarshalledEntry(InvocationContext ctx, Object key) {
+      InternalCacheValue sv = entryFactory.getValueFromCtxOrCreateNew(key, ctx);
+      return new MarshalledEntryImpl(key, sv.getValue(), internalMetadata(sv), marshaller);
    }
 
    protected boolean skipSharedStores(InvocationContext ctx, Object key, FlagAffectedCommand command) {
       return !ctx.isOriginLocal() ||
             command.hasAnyFlag(FlagBitSets.SKIP_SHARED_CACHE_STORE);
-   }
-
-   InternalCacheValue getStoredValue(Object key, InvocationContext ctx) {
-      CacheEntry entry = ctx.lookupEntry(key);
-      if (entry instanceof InternalCacheEntry) {
-         return ((InternalCacheEntry) entry).toInternalCacheValue();
-      } else {
-         if (ctx.isInTxScope()) {
-            EntryVersionsMap updatedVersions =
-                  ((TxInvocationContext) ctx).getCacheTransaction().getUpdatedEntryVersions();
-            if (updatedVersions != null) {
-               EntryVersion version = updatedVersions.get(entry.getKey());
-               if (version != null) {
-                  Metadata metadata = entry.getMetadata();
-                  if (metadata == null) {
-                     // If no metadata passed, assumed embedded metadata
-                     metadata = new EmbeddedMetadata.Builder().lifespan(entry.getLifespan()).maxIdle(entry.getMaxIdle())
-                                                              .version(version).build();
-                     return entryFactory.create(entry.getKey(), entry.getValue(), metadata).toInternalCacheValue();
-                  } else {
-                     metadata = metadata.builder().version(version).build();
-                     return entryFactory.create(entry.getKey(), entry.getValue(), metadata).toInternalCacheValue();
-                  }
-               }
-            }
-         }
-
-         return entryFactory.create(entry).toInternalCacheValue();
-      }
    }
 }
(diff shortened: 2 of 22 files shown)


core/src/test/java/org/infinispan/persistence/BaseStoreFunctionalTest.java
```diff
@@ -275,6 +280,25 @@ public void writeToAllNonTxStores(MarshalledEntry marshalledEntry, AccessMode mo
...
+   public void testPutAllBatch() throws Exception {
+      int numberOfEntries = 100;
+      String cacheName = "testPutAllBatch";
+      ConfigurationBuilder cb = new ConfigurationBuilder();
+      cb.read(cacheManager.getDefaultCacheConfiguration());
+      createCacheStoreConfig(cb.persistence(), false);
+      cacheManager.defineConfiguration(cacheName, cb.build());
+
+      Cache<Integer, String> cache = cacheManager.getCache(cacheName);
+      Map<Integer, String> entriesMap = IntStream.range(0, numberOfEntries).boxed()
```
wburns: We should also update `testRemoveCacheWithPassivation` so that it checks the batch method as well.

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
```diff
@@ -35,4 +38,28 @@
...
+   }
+
+   /**
+    * Remove all provided keys from the store in a single batch operation. If this is not supported by the
+    * underlying store, then keys are removed from the store sequentially via {@link #delete(Object)}
+    *
+    * @param keys a List of MarshalledEntry to be removed from the store.
+    * @throws NullPointerException if keys is null.
+    */
+   default void deleteBatch(List<Object> keys) {
```
wburns: I am thinking we want to implement this with `SingleFileStore` as well. In this case it can first order the entries to remove so that the disk updates are done sequentially to more efficiently write to disk. Take a look at the `process` method which does something similar.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stringbased/JdbcStringBasedStore.java
```diff
@@ -257,6 +258,52 @@ private void executeLegacyUpdate(Connection connection, MarshalledEntry entry, S
    }
 
    @Override
+   public void writeBatch(List<MarshalledEntry<? extends K, ? extends V>> marshalledEntries) {
+      // If upsert is not supported, then we must execute the legacy write for each entry; i.e. read then update/insert
+      if (!tableManager.isUpsertSupported()) {
+         marshalledEntries.forEach(this::write);
```
wburns: Up to you but this could also just be `super.writeBatch(marshalledEntries);`

persistence/jdbc/src/test/java/org/infinispan/persistence/jdbc/stringbased/JdbcStringBasedStoreBatchTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/**
+ * @author Ryan Emerson
+ * @since 9.1
+ */
+@Test(groups = "functional", testName = "persistence.jdbc.stringbased.JdbcStringBasedStoreBatchTest")
+public class JdbcStringBasedStoreBatchTest extends AbstractInfinispanTest {
```
wburns: Hrmm wouldn't this now be covered by `JdbcStringBasedStoreTest` since you added a similar method to the BaseStoreTest ?

persistence/jpa/src/main/java/org/infinispan/persistence/jpa/JpaStore.java
```diff
@@ -321,6 +312,56 @@ public boolean delete(Object key) {
...
+               MetadataEntity metadata = getMetadata(key, em);
+               removeEntity(em, entity);
+               if (metadata != null)
+                  removeMetadata(em, metadata);
+            }
+            txn.commit();
+            stats.addBatchRemoveTxCommitted(timeService.time() - txnBegin);
+         } catch (Exception e) {
+            stats.addBatchRemoveTxFailed(timeService.time() - txnBegin);
+            throw new JpaStoreException("Exception caught in deleteBatch()", e);
```
wburns: Maybe only wrap a non `JpaStoreException` ?

persistence/jpa/src/main/java/org/infinispan/persistence/jpa/JpaStore.java
```diff
@@ -321,6 +312,56 @@ public boolean delete(Object key) {
...
 
    @Override
+   public void deleteBatch(List<Object> keys) {
+      EntityManager em = emf.createEntityManager();
+      try {
+         EntityTransaction txn = em.getTransaction();
+         long txnBegin = timeService.time();
+         try {
+            txn.begin();
+            for (Object key : keys) {
```
wburns: I think we should be able to use a `CriteriaDelete` instead since we don't require cascading deletes. This should do the batching for us automatically, instead of having to find and remove each entity.

persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java
```diff
@@ -358,6 +368,34 @@ public MarshalledEntry load(Object key) {
         }
     }
 
+    @Override
+    public void writeBatch(List<MarshalledEntry<? extends K, ? extends V>> marshalledEntries) {
+        try {
+            WriteBatch writeBatch = new WriteBatch();
```
wburns: I know that you didn't write this originally, but it looks like the expiredDb updates in `purge` method are badly in need of this same exact batching.

Also not to mention it scares me that you can only insert 10000 elements into the rocks db store until you will block a thread until the expiration reaper is ran. This needs to be cleaned up @tristantarrant . I would say instead of calling put at https://github.com/infinispan/infinispan/blob/master/persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java#L478 we should invoke offer and if we weren't able to insert to start off a thread to clear this out or do it inline.

core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
```diff
@@ -99,14 +102,14 @@ public void passivateAll() throws PersistenceException {
...
-         long count = 0;
-         for (InternalCacheEntry e : container) {
-            count++;
-            if (trace) log.tracef("Passivating %s", e.getKey());
-            persistenceManager.writeToAllNonTxStores(marshalledEntryFactory.newMarshalledEntry(e.getKey(), e.getValue(),
-                                                                                               internalMetadata(e)), BOTH);
-         }
-         log.passivatedEntries(count, Util.prettyPrintTime(timeService.timeDuration(start, TimeUnit.MILLISECONDS)));
+
+         List<MarshalledEntry> entries = StreamSupport
```
ryanemerson: Good point!  I'll add batching with some default value. It's probably overkill to make this value configurable, WDYT?

core/src/main/java/org/infinispan/persistence/spi/AdvancedCacheWriter.java
```diff
@@ -35,7 +35,7 @@
     * Callback to be notified when an entry is removed by the {@link #purge(java.util.concurrent.Executor,
     * org.infinispan.persistence.spi.AdvancedCacheWriter.PurgeListener)} method.
     */
-   public interface PurgeListener<K> {
+   interface PurgeListener<K> {
```
ryanemerson: I thought that we could remove the public modifier as it is redundant for inner interfaces and therefore doesn't change the accessibility of PurgeListener?

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stringbased/JdbcStringBasedStore.java
ryanemerson: If we call super, we have to call `AdvancedLoadWriteStore.super.writeBatch(marshalledEntries)` as it is a default interface implementation. So calling `marshalledEntries.forEach(this::write)` is more explicit and succinct.

persistence/jdbc/src/test/java/org/infinispan/persistence/jdbc/stringbased/JdbcStringBasedStoreBatchTest.java
ryanemerson: Yep! I wrote this test first and I must have forgot to delete it

core/src/main/java/org/infinispan/interceptors/impl/DistCacheWriterInterceptor.java
```diff
@@ -88,7 +91,9 @@ public Object visitPutMapCommand(InvocationContext ctx, PutMapCommand command) t
          PutMapCommand putMapCommand = (PutMapCommand) rCommand;
          LocalizedCacheTopology cacheTopology = dm.getCacheTopology();
          Map<Object, Object> map = putMapCommand.getMap();
-         int count = 0;
+
```
rvansa: please refactor this whole lambda into `handlePutMapCommandReturn` and store `this::handlePutMapCommandReturn` in the field of the same name...

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -151,18 +151,22 @@ public boolean delete(Object key) {
    }
 
    protected void applyModificationsSync(List<Modification> mods) throws PersistenceException {
+      List<MarshalledEntry> entriesToWrite = new ArrayList<>();
```
rvansa: Would it make sense to coalesce writes to the same key here?

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
rvansa: Why is this a list and not a `Collection<?>`?

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
```diff
@@ -35,4 +38,28 @@
     * @throws PersistenceException in case of an error, e.g. communicating with the external storage
     */
    boolean delete(Object key);
+
+   /**
+    * Persist all provided entries to the store in a single batch update. If this is not supported by the
+    * underlying store, then entries are written to the store sequentially via {@link #write(MarshalledEntry)}
```
rvansa: Please document if the list may contain duplicate entries for the same key (in that case we really need a list, otherwise it could be just a Collection).

wburns:
Yes, sorry about that. I always forget it is dependent on the outer
interface.

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
ryanemerson: I originally went with a List as I thought going forward it would provide us with more flexibility for future use-cases where we may need to write a batch containing duplicates. However, our present use cases (passivation, putAll, async writer) guarantee that there will be no duplicates. So I I'm happy with List or Collection depending on how strong you feel about it

rvansa: Ah, you're right that async write already does the coalescing. In that case I think that Collection would be more appropriate. Documentation still needs these details, though.

However, it seems you've missed one use-case that would benefit from this: `CacheWriterInterceptor$Updater`. If the store does not support `TransactionalCacheWriter`, the update should be still batched.

core/src/main/java/org/infinispan/interceptors/impl/DistCacheWriterInterceptor.java
ryanemerson: What's the benefit of this as the lambda is not stateless? The action taken by the lambda depends on the `InvocationContext` and `PutMapCommand` (flags) that are passed to visitPutMapCommand.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
ryanemerson: I believe this already happens when the modification is created and then stored in the `modifications` map based upon the entries key.

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
ryanemerson: Good spot! I'll update the updater to use batches and perform coalescing. It makes sense for the coalescing to be done in memory opposed to redundant write operations on the store.

core/src/main/java/org/infinispan/interceptors/impl/DistCacheWriterInterceptor.java
rvansa: The lambda implicitly captures `this` (for example to be able to call `isProperWriter`) and that results in an allocation. Regrettably JVM does not optimize this out yet.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
rvansa: True, my bad.

core/src/main/java/org/infinispan/persistence/spi/AdvancedCacheWriter.java
wburns: Yes, sorry about that. I always forget it is dependent on the outer interface.

core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
wburns: We could use the state transfer `chunkSize`. I tend to like using that value for stuff like this :)

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
```diff
@@ -35,4 +38,28 @@
...
+    *
+    * @param entries a List of MarshalledEntry to be written to the store.
+    * @throws NullPointerException if entries is null.
+    */
+   default void writeBatch(List<MarshalledEntry<? extends K, ? extends V>> entries) {
```
wburns: We probably want to do the same thing for writeBatch with SFS as well.

persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java
ryanemerson: I've created [ISPN-7912](https://issues.jboss.org/browse/ISPN-7912) to track this, I'll issue a fix in a separate PR.

ryanemerson:
@wburns @rvansa Updated the PR with the requested changed

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -373,13 +400,35 @@ protected boolean isProperWriter(InvocationContext ctx, FlagAffectedCommand comm
       return true;
    }
 
-   public class Updater extends AbstractVisitor {
-
-      protected final boolean generateStatistics;
+   static class BatchUpdater extends AbstractVisitor {
```
pruivo: what is the benefit of making this class `static`?

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stringbased/JdbcStringBasedStore.java
```diff
@@ -257,6 +258,52 @@ private void executeLegacyUpdate(Connection connection, MarshalledEntry entry, S
...
+         try (PreparedStatement deleteBatch = connection.prepareStatement(tableManager.getDeleteRowSql())) {
+            for (Object key : keys) {
+               String keyStr = key2Str(key);
+               deleteBatch.setString(1, keyStr);
+               deleteBatch.addBatch();
+            }
+            deleteBatch.executeBatch();
+         }
+      } catch (SQLException e) {
+         throw log.sqlFailureWritingBatch(e);
```
pruivo: wrong log message. should be `log.sqlFailureDeletingBatch()`

core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
```diff
@@ -28,10 +30,13 @@
 
 public class PassivationManagerImpl implements PassivationManager {
 
+   private static final int DEFAULT_BATCH_SIZE = 50;
```
pruivo: unused

core/src/main/java/org/infinispan/persistence/manager/PersistenceManager.java
```diff
@@ -1,6 +1,7 @@
 package org.infinispan.persistence.manager;
 
 import java.util.Collection;
+import java.util.List;
```
pruivo: unused import

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -522,6 +522,39 @@ public void writeToAllNonTxStores(MarshalledEntry marshalledEntry, AccessMode ac
...
+         storesMutex.readLock().unlock();
+      }
+   }
+
+   @Override
+   public void deleteBatchFromAllNonTxStores(Collection<Object> keys, AccessMode accessMode, long flags) {
+      if (keys.isEmpty())
+         return;
+
+      log.tracef(String.format("deleteBatchFromAllNonTxStores: keys %s, mode %s, flags %s", keys, accessMode, flags));
```
pruivo: check if trace enabled

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerStub.java
```diff
@@ -2,6 +2,7 @@
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
```
pruivo: unused import

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -505,38 +583,31 @@ public void process(KeyFilter<? super K> filter, final CacheLoaderTask<K, V> tas
...
                keysToLoad.add(new KeyValuePair<>(e.getKey(), e.getValue()));
          }
-         Collections.sort(keysToLoad, new Comparator<KeyValuePair<K, FileEntry>>() {
-            @Override
-            public int compare(KeyValuePair<K, FileEntry> o1, KeyValuePair<K, FileEntry> o2) {
-               long offset1 = o1.getValue().offset;
-               long offset2 = o2.getValue().offset;
-               return offset1 < offset2 ? -1 : offset1 == offset2 ? 0 : 1;
-            }
+         keysToLoad.sort((o1, o2) -> {
```
wburns: I am looking at this again and I can't remember why this has to be in the synchronized block. The comparator doesn't touch any mutable state as offset is final. Do you remember why @danberindei ?

danberindei: @wburns The entries map is not a `ConcurrentHashMap`, it's a `[Linked]HashMap`, so iterating through the entry set (or any read for that matter) needs to be synchronized.

wburns: I know that for entries, but I am talking about the sort here only. The entries map would already have been copied over to the `keysToLoad` collection.

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
```diff
@@ -35,4 +39,30 @@
...
+   /**
+    * Persist all provided entries to the store in a single batch update. If this is not supported by the
+    * underlying store, then entries are written to the store individually via {@link #write(MarshalledEntry)}.
+    * As the execution order of MarshalledEntries is not guaranteed, you should ensure that only a single entry exists
+    * in the Collection for a given key.
+    *
+    * @param entries a Collection of MarshalledEntry to be written to the store.
+    * @throws NullPointerException if entries is null.
+    */
+   default void writeBatch(Collection<MarshalledEntry<? extends K, ? extends V>> entries) {
```
wburns: The more I am thinking about this, this can just be Iterable<MarshalledEntry<...>>. All implementations use it in this way and then PersistenceManagerImpl can pass the Iterable directly and not have to do any batching (as the Iterator does it automatically).

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
danberindei: @wburns Ah ok, once I expanded the diff to see the `synchronized` block my eyes got stuck on the `for` line :) I agree, there's no reason to perform the sort in the `synchronized` block.

wburns: Okay, that is what I thought, but just wanted to be sure.

In this case we could condense the sort with the stream below, but it might not be as readable, up to you @ryanemerson

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -209,20 +212,31 @@ public Object visitReplaceCommand(InvocationContext ctx, ReplaceCommand command)
...
+         for (Object key : putMapCommand.getMap().keySet()) {
+            if (skipSharedStores(rCtx, key, (FlagAffectedCommand) rCommand))
+               nonSharedEntries.add(createMarshalledEntry(rCtx, key));
+            else
+               allWriterEntries.add(createMarshalledEntry(rCtx, key));
          }
+
+         long flags = ((FlagAffectedCommand)rCommand).getFlagsBitSet();
+         persistenceManager.writeBatchToAllNonTxStores(allWriterEntries, BOTH, flags);
+         persistenceManager.writeBatchToAllNonTxStores(nonSharedEntries, PRIVATE, flags);
```
danberindei: `writeBatchToAllNonTxStores` already filters the entries it writes, I think we can add another check in there instead of copying the collection (not necessarily another `filter` call, because I'm not sure the filter allocation is escaped).

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -350,12 +364,25 @@ protected final void store(TxInvocationContext ctx) throws Throwable {
...
+      BatchModification nonSharedMods = modsBuilder.nonSharedModifications;
+
+      persistenceManager.writeBatchToAllNonTxStores(sharedMods.getMarshalledEntries(), BOTH, 0);
+      persistenceManager.writeBatchToAllNonTxStores(nonSharedMods.getMarshalledEntries(), PRIVATE, 0);
+      persistenceManager.deleteBatchFromAllNonTxStores(sharedMods.getKeysToRemove(), BOTH, 0);
+      persistenceManager.deleteBatchFromAllNonTxStores(nonSharedMods.getKeysToRemove(), PRIVATE, 0);
+
+      if (trace) {
+         getLog().tracef("Writing shared batch with #entries=%s and non-shared batch with #entries=%s", sharedMods.getMarshalledEntries().size(), nonSharedMods.getMarshalledEntries().size());
+         getLog().tracef("Deleting shared batch with #entries=%s and non-shared batch with #entries=%s", sharedMods.getKeysToRemove().size(), nonSharedMods.getKeysToRemove().size());
```
danberindei: Why not `%d entries`?

core/src/main/java/org/infinispan/interceptors/impl/DistCacheWriterInterceptor.java
danberindei: @ryanemerson I'm not sure how this looked before @rvansa's comment, but please don't return a lambda from this method: just put the lambda body here, and then `protected final InvocationSuccessAction handlePutMapCommandReturn = this::handlePutMapCommandReturn`.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -368,6 +352,100 @@ public void write(MarshalledEntry<? extends K, ? extends V> marshalledEntry) {
...
+
+      return allocateAndCreateFileEntry(entry, key, data, meta);
+   }
+
+   @Override
+   public void writeBatch(Collection<MarshalledEntry<? extends K, ? extends V>> marshalledEntries) {
+      resizeLock.readLock().lock();
+      try {
+         List<FileEntry> fileEntries = marshalledEntries.stream().map(this::allocateAndCreateFileEntry)
+               .collect(Collectors.toList());
```
danberindei: +1, the JVM may be able to elide the allocation of streams and filters, but it won't elide the allocation of the list's backing array any time soon.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -368,6 +352,100 @@ public void write(MarshalledEntry<? extends K, ? extends V> marshalledEntry) {
...
+      try {
+         synchronized (entries) {
+            for (Object key : keys) {
+               FileEntry entry = entries.get(key);
+               if (entry != null)
+                  keysToLoad.add(new KeyValuePair<>((K) key, entry));
+            }
+         }
+
+         keysToLoad.sort((o1, o2) -> {
```
danberindei: Do we need to delete the entries in a particular order?

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -765,21 +834,24 @@ public SingleFileStoreConfiguration getConfiguration() {
...
 
-      public FileEntry(long offset, int size, int keyLen, int dataLen, int metadataLen, long expiryTime) {
+      public FileEntry(long offset, int size, int keyLen, int dataLen, int metadataLen, long expiryTime, MarshalledEntry marshalledEntry) {
          this.offset = offset;
          this.size = size;
          this.keyLen = keyLen;
          this.dataLen = dataLen;
          this.metadataLen = metadataLen;
          this.expiryTime = expiryTime;
+         this.marshalledEntry = marshalledEntry;
```
danberindei: Isn't this going to keep all serialized values in memory forever?

`FileEntry` should be as small as possible, because we need to keep a `FileEntry` instance in the heap for each entry in the store.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
wburns: The idea being that it would be faster to remove entries from disk in order to prevent from having to skip back and forth (same as process).

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
wburns: The sort is the same reason as process, assuming it is faster to write to disk sequentially to prevent random locations on disk. Although the more I think about it, writes may not have this issue since it should be sequential anyways unless the size is small enough to fill a gap. Wonder if it is worth it here.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
wburns: 👍 good catch, yeah we definitely can't keep that around.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
danberindei: @wburns For `process()` it makes sense because we're reading the entire file, or at least a big part of it. But batches should be much smaller and much more likely to fit in the OS buffers, so I don't think ordering writes would help.

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
danberindei: Not sure what you mean with the automatic batching in `Iterator`, but I hope it's not `IteratorSpliterator`, because those batches are way too big :)

> Implementation Note:
>     The default implementation should usually be overridden. The spliterator returned by the default implementation has poor splitting capabilities, is unsized, and does not report any spliterator characteristics. Implementing classes can nearly always provide a better implementation.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
danberindei: @wburns probably not :)

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
wburns: When you say writes, I can agree regarding the write method since those are already done sequentially (as I mentioned in other comment). However for delete I am not as sure. For a larger batch I would think sorting would matter and in a smaller batch it might not make much difference but the sort also matters less than too.

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
wburns: No, I was referring to the fact that Off Heap should make entries visible per address counter https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java#L590

If these are batched internally by the Stream, I am not sure, but this would be an issue outside the scope of this.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
wburns: I agree :)

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
wburns: And also iirc the IteratorSpliterator "batching" only comes into account when using a parallel stream.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
danberindei: @wburns IMO it doesn't matter either way, because SFS only writes to the OS buffers, and the OS then batches the writes to disk for us.

wburns: Okay I defer to your judgement @danberindei as I am not as familiar with how it would perform in scenarios where the writes are scattered in different positions among the file.

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
danberindei: @wburns Aren't the off heap "batches" much bigger than the store batches should be? I forgot all about what the address counter is, but it looks to me like there's at most 2 of them per processor, and for the store batches even 512 seems like too much to me (e.g. https://stackoverflow.com/a/27855325/55870).

wburns:
Address pointer is essentially a bucket akin to hash map. The default value
is 1 million. You might be looking at the locks which is the number of
processors rounded to nearest power of two multiplied by two. There should
be only 1 entry per address pointer unless you have bad hashes or too low
of address count.

I was going to confirm the stream "batches" are working as I expect but
haven't had a chance to look into it. I will check tonight.

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
wburns: It looks like flatMap totally messes up the stream "batching" and it ends up keeping all entries in memory for a given lock. I created https://issues.jboss.org/browse/ISPN-7922 to look into it. In regards to the database batch, we may have to figure something out regarding that still.

wburns: This is caused due to most likely a bug in flatMap with streams (as another use found it to be even more of a severe issue https://stackoverflow.com/a/29986041/4222962)

danberindei: I'm still not sure how the batching in `OffHeapDataContainer.entrySet()` was supposed to work... the consumer of the stream can't say "give me the elements from the next mapper invocation", it will still read elements from an iterator one by one. At most you could implement a spliterator whose `trySplit()` method uses the lock/address/pointer boundaries, but I doubt the default `flatMap()` will do that even if the [JDK-8075939](https://bugs.openjdk.java.net/browse/JDK-8075939) is fixed.

BTW, even though there isn't a `flatMapToObj` method, you could use `intStream.mapToObj(i -> stream).flatMap(stream -> stream)`.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
ryanemerson: Very good point! I was too busy trying to reduce code repetition that I didn't think about the winder implications :grimacing:

ryanemerson:
@pruivo Regarding `AdvancedCacheWriter` see the PR description

pruivo:
@ryanemerson I was talking about the async cache writer, not the advanced :)

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
wburns: The consumer doesn't say that it says give me the next batch. To be honest I never meant to imply that `OffHeapDataContainer.entrySet` did it's own batching (I thought it would do 1 element at a time as well). Just kinda fell down the rabbit hole :)

But I am glad I looked into it, because `flatMap` is really a collecting intermediate operation, creates "batches", and I was unaware of that behavior. I had assumed it would lazily process each entry of the inner stream, but it first has to collect all entries of that stream. But knowing this I changed the `entrySet` so that it "batches" it per address pointer and if they ever change the behavior, we will be fine as well.

It seems the spec only specifies the source spliterator is the only that has to be lazy (unless a stateful operator is used).

wburns: Also re: `mapToObj.flatMap` I recently saw that usage and I didn't even think to use it for this. Thanks I will give that a go.

ryanemerson:
@pruivo Argh sorry, my eyes failed me! The `AsyncCacheWriter` makes use of the batch methods in `applyModificationsSync`. See [here](https://github.com/ryanemerson/infinispan/blob/850b0b98aabb9b59f587845fdfeed16ac4e08dad/core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java#L168-L169). Thinking about it though, I should create an `AsyncCacheWriter` implementation so that the writeLock is only called once per batch and State makes use of `putAll` on the modifications map.

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
ryanemerson: I'm not sure how it is possible to do this in `PersistenceManagerImpl` taking into account the different logic required for `local | dist` stores which is currently provided by the overrided method `skipSharedStores`.  Also, in `PersistenceManagerImpl` we are only iterating over the configured writers, not the entries themselves.

danberindei: @ryanemerson Ah right, I mistook the loop in `writeBatchToAllNonTxStores` for a loop over the entries. I'd still like to avoid copying the collection when there's a single store, but I guess there's no use worrying about the collection when there are multiple stores.

Or maybe we should accept `writeBatch(Iterable)` after all, and call `writeBatchToAllNonTxStores(() -> Spliterators.iterator(putMapCommand.getMap().entrySet().stream().filter(...).spliterator()))` :)

ryanemerson: @danberindei Sounds good. I have already changed to writeBatch(Iterable) locally, so the change is trivial enough :smile:

wburns: I wonder if we want to add a new Iterator wrapper impl something like IteratorFilter (to be inline with IteratorMapper) or probably better named FilteredIterator (probably would have named MappedIterator if I had thought about it more)

Then we don't have to construct a stream and the new Iterator is pretty light weight.

core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
```diff
@@ -32,6 +34,7 @@
    CacheNotifier notifier;
    Configuration cfg;
    private volatile boolean skipOnStop = false;
+   private int batchSize;
```
pruivo: doesn't seem to be used anywhere

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
```diff
@@ -1,5 +1,9 @@
 package org.infinispan.persistence.spi;
 
+import java.util.Collection;
```
pruivo: unsued imports

pruivo:
@ryanemerson unfortunately yes

wburns:
@ryanemerson @pruivo There is nothing stopping the AsyncCacheWriter from pulling entries from the iterator and then acquiring the correct number of "locks" based on how many you pulled. The implementation can completely decide if it wants to have it's own batch size or if it wants to consume all at once. And as @danberindei mentioned in a previous comment we may want to limit the batch size when doing SQL updates as well as it could cause memory issues.

ryanemerson:
@wburns Ok, I've now overridden the batch methods for `AsyncCacheWriter` so that the write lock is only acquired once, at the cost of an additional iteration of the batch. I have also created [ISPN-7936](https://issues.jboss.org/browse/ISPN-7936) to add batch-size configuration on a per store basis, but I will issue a separate PR for this.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -145,27 +150,47 @@ public void write(MarshalledEntry entry) {
    }
 
    @Override
+   public void writeBatch(Iterable entries) {
```
wburns: Seems this should be `Iterable<MarshalledEntry>`. Then you don't need the cast below.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -145,27 +150,47 @@ public void write(MarshalledEntry entry) {
...
+      putAll(
+            StreamSupport.stream((Spliterator<MarshalledEntry>) entries.spliterator(), false)
+                  .map(me -> new Store(me.getKey(), me))
+                  .collect(Collectors.toList())
+      );
+   }
+
+   @Override
+   public void deleteBatch(Iterable keys) {
```
wburns: Seems this should be `Iterable<Object>`. Then you don't need the cast below.

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -505,38 +503,31 @@ public void process(KeyFilter<? super K> filter, final CacheLoaderTask<K, V> tas
...
                keysToLoad.add(new KeyValuePair<>(e.getKey(), e.getValue()));
          }
-         Collections.sort(keysToLoad, new Comparator<KeyValuePair<K, FileEntry>>() {
-            @Override
-            public int compare(KeyValuePair<K, FileEntry> o1, KeyValuePair<K, FileEntry> o2) {
-               long offset1 = o1.getValue().offset;
-               long offset2 = o2.getValue().offset;
-               return offset1 < offset2 ? -1 : offset1 == offset2 ? 0 : 1;
-            }
+         keysToLoad.sort((o1, o2) -> {
```
wburns: We can move this outside of the synchronized block.

core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
```diff
@@ -99,14 +101,14 @@ public void passivateAll() throws PersistenceException {
...
-         long count = 0;
-         for (InternalCacheEntry e : container) {
-            count++;
-            if (trace) log.tracef("Passivating %s", e.getKey());
-            persistenceManager.writeToAllNonTxStores(marshalledEntryFactory.newMarshalledEntry(e.getKey(), e.getValue(),
-                                                                                               internalMetadata(e)), BOTH);
-         }
-         log.passivatedEntries(count, Util.prettyPrintTime(timeService.timeDuration(start, TimeUnit.MILLISECONDS)));
+
+         final AtomicInteger count = new AtomicInteger();
```
wburns: This unfortunately won't work correctly if we have more than 1 cache writer. I would just store `container.sizeIncludingExpired` and pass that to the log message below. It can be off if we have expired entries, but it should be fine really.

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -209,20 +209,28 @@ public Object visitReplaceCommand(InvocationContext ctx, ReplaceCommand command)
...
+      PutMapCommand putMapCommand = (PutMapCommand) rCommand;
+      if (!isStoreEnabled(putMapCommand) || rCtx.isInTxScope())
+         return;
+
+      processIterableBatch(rCtx, putMapCommand, BOTH, key -> !skipSharedStores(rCtx, key, putMapCommand));
+      processIterableBatch(rCtx, putMapCommand, PRIVATE, key -> skipSharedStores(rCtx, key, putMapCommand));
+   }
+
+   protected void processIterableBatch(InvocationContext ctx, PutMapCommand cmd, PersistenceManager.AccessMode mode, Predicate<Object> filter) {
+      Iterable<MarshalledEntry> iterable = () -> StreamSupport.stream(Spliterators.spliterator(cmd.getMap().keySet(), Spliterator.NONNULL), false)
```
wburns: Why not just `cmd.getMap().keySet().stream()` ?

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -209,20 +209,28 @@ public Object visitReplaceCommand(InvocationContext ctx, ReplaceCommand command)
...
+            .filter(filter)
+            .peek(Void -> {
+               if (getStatisticsEnabled())
+                  cacheStores.incrementAndGet();
```
wburns: Could this just not be incremented by `cmd.getMap().size()` like we did before ?

core/src/main/java/org/infinispan/eviction/impl/PassivationManagerImpl.java
ryanemerson: Argh I didn't think of multiple stores! Good spot.

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
ryanemerson: No idea what I was thinking :confused:

PR #5187, opened 2017-06-08
ISPN-7847 Deprecate DeltaAware

rvansa:
* DeltaAware, CopyableDeltaAware, Delta, DeltaCompositeKey deprecated
* AdvancedCache.applyDelta deprecated, CacheImpl.applyDelta reimplemented using ReadWriteKeyValueCommand, permits to lock only on the main key
* ApplyDeltaCommand and Visitor.visitApplyDeltaCommand deprecated (implementation removed)
* DeltaAwareCacheEntry, DeltaAwareObjectOutput and some internal helpers removed


tristantarrant:
3 related failures

rvansa:
I've removed the three failing tests; these are no longer applicable because `Flag.DELTA_AWARE` is no longer used, and does not make the put requiring previous value on backup owner.
Also adding back the Wait CI Results label - we need to ran whole testsuite (not just core) to check potential problems in tree cache or lucene directory.

galderz:
Last build din't work, but there was a [build](http://ci.infinispan.org/job/Infinispan/job/PR-5187/9/) from 2 days back that run fine. Do you want to wait for another CI run?

rvansa:
@galderz Not really, but I see there is a conflict now :-/

rvansa:
@galderz Rebased.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -1243,7 +1265,14 @@ private final V putIfAbsent(K key, V value, Metadata metadata, long explicitFlag
    @SuppressWarnings("unchecked")
    final V putIfAbsent(K key, V value, Metadata metadata, long explicitFlags, InvocationContext ctx) {
       assertKeyValueNotNull(key, value);
-      PutKeyValueCommand command = createPutIfAbsentCommand(key, value, metadata, explicitFlags, ctx);
+      DataWriteCommand command;
+      if (value instanceof Delta) {
+         command = createApplyDelta(key, (Delta) value, explicitFlags, ctx);
+      } if (value instanceof DeltaAware && (explicitFlags & FlagBitSets.CACHE_MODE_LOCAL) == 0) {
```
pruivo: bug: `else` is missing

core/src/main/java/org/infinispan/atomic/impl/ApplyDelta.java
```diff
@@ -0,0 +1,93 @@
...
+      }
+      Delta delta = (Delta) d;
+      if (view.find().isPresent()) {
+         Object value = view.find().get();
+         DeltaAware deltaAware;
+         if (value instanceof CopyableDeltaAware) {
+            deltaAware = ((CopyableDeltaAware) value).copy();
+         } else if (value instanceof DeltaAware) {
+            try {
+               byte[] bytes = marshaller.objectToByteBuffer(value);
```
pruivo: nice :)

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -272,7 +267,7 @@ private Object handleSecondPhaseCommand(TxInvocationContext ctx, TransactionBoun
       LocalTransaction localTx = (LocalTransaction) ctx.getCacheTransaction();
       LocalizedCacheTopology cacheTopology = dm.getCacheTopology();
       Collection<Address> affectedNodes =
-            isReplicated ? null : cacheTopology.getWriteOwners(getAffectedKeysFromContext(ctx));
+            isReplicated ? null : cacheTopology.getWriteOwners((Collection<Object>) ctx.getAffectedKeys());
```
pruivo: you don't need the cast

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
rvansa: well spotted! :)

rvansa:
@pruivo Thanks, rebased & fixed.

rvansa:
@pruivo Removed imports

pruivo:
@rvansa can I push the button or wait for others reviewers?

rvansa:
@pruivo I don't see why not. I usually pick multiple reviewers, hoping that at least one of them will have the time.

core/src/main/java/org/infinispan/atomic/impl/ApplyDelta.java
```diff
@@ -0,0 +1,93 @@
...
+import org.infinispan.commons.marshall.AdvancedExternalizer;
+import org.infinispan.commons.marshall.Ids;
+import org.infinispan.commons.marshall.Marshaller;
+import org.infinispan.commons.util.Util;
+import org.infinispan.factories.GlobalComponentRegistry;
+
+/**
+ * Replacement for {@link org.infinispan.commands.write.ApplyDeltaCommand} and
+ * {@link org.infinispan.context.Flag#DELTA_WRITE}. Deprecated since {@link Delta},
+ * {@link org.infinispan.atomic.DeltaAware} and {@link org.infinispan.atomic.CopyableDeltaAware} are deprecated, too.
```
danberindei: I don't think you really need the deprecation if it's not visible to applications.

core/src/main/java/org/infinispan/atomic/impl/ApplyDelta.java
danberindei: @rvansa Could it use `MergeOnStore` instead? Not that I think `MergeOnStore` is a good idea, but it seems quite close to the semantics of non-copyable `DeltaAware`.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -875,14 +879,23 @@ boolean lock(Collection<? extends K> keys, long flagsBitSet) {
...
          throw new IllegalArgumentException("Cannot lock empty list of keys");
+      } else if (locksToAcquire.length != 1) {
+         throw new IllegalArgumentException("Only one lock is permitted.");
+      } else if (!Objects.equals(locksToAcquire[0], deltaAwareValueKey)) {
+         throw new IllegalArgumentException("The delta aware key and locked key must match.");
       }
-      InvocationContext ctx = invocationContextFactory.createInvocationContext(true, UNBOUNDED);
-      ApplyDeltaCommand command = commandsFactory.buildApplyDeltaCommand(deltaAwareValueKey, delta,
-            Arrays.asList(locksToAcquire));
+      assertKeyNotNull(deltaAwareValueKey);InvocationContext ctx = invocationContextFactory.createInvocationContext(true, 1);
```
danberindei: split lines

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -875,14 +879,23 @@ boolean lock(Collection<? extends K> keys, long flagsBitSet) {
...
+      ReadWriteKeyValueCommand<K, Object, Object> command = createApplyDelta(deltaAwareValueKey, delta, FlagBitSets.IGNORE_RETURN_VALUES, ctx);
+      invoker.invoke(ctx, command);
+   }
+
+   private ReadWriteKeyValueCommand<K, Object, Object> createApplyDelta(K deltaAwareValueKey, Delta delta, long explicitFlags, InvocationContext ctx) {
+      ReadWriteKeyValueCommand<K, Object, Object> command = commandsFactory.buildReadWriteKeyValueCommand(deltaAwareValueKey, delta,new ApplyDelta<>(marshaller), Params.create());
+                                                                         command.setFlagsBitSet(explicitFlags);
```
danberindei: Way too much indentation.

core/src/main/java/org/infinispan/commands/write/ApplyDeltaCommand.java
```diff
@@ -55,14 +40,7 @@ public Delta getDelta(){
...
    public Object perform(InvocationContext ctx) throws Throwable {
-      CacheEntry contextEntry = ctx.lookupEntry(key);
-      if (contextEntry instanceof DeltaAwareCacheEntry) {
-         DeltaAwareCacheEntry deltaAwareCacheEntry = (DeltaAwareCacheEntry) contextEntry;
-         deltaAwareCacheEntry.appendDelta(delta);
-      } else {
-         throw new IllegalStateException();
-      }
-      return null;
+      throw new UnsupportedOperationException();
```
danberindei: Why keep the command at all?

rvansa: You know that the publicity of interceptors is problematic. I left this as someone might consider it public, and removing the method from `Visitor` could break something.

danberindei: Right, forgot about `visitApplyDeltaCommand`.

galderz:
Rebase needed again :|

rvansa:
@galderz Damned imports.

rvansa:
Fixed conflict in documentation.

pruivo:
@rvansa and @danberindei is there anything missing?

danberindei:
@pruivo Just one more look...

danberindei:
@pruivo I would have liked to have a longer discussion with @rvansa on `MergeOnStore` and why it's useful for the atomic maps but not for `DeltaAware`, but I don't think it would lead to changes to this PR, so I'm fine with integrating.

rvansa:
The purpose of `DeltaAware` was isolating incremental changes from the changing object on origin, and replicate only `Delta`. However, it has some issues (e.g. the one addressed by `CopyableDeltaAware`), sometimes an unclear semantic, and propagation through PKVC required 'extra ifs' as well as custom entry type. All in all it's inferior to use of functional commands, and we don't want to maintain two ways to do the same.

What you wanted to compare is the `MergeOnStore` - definitely a hack to implement FGAMs which should have never been born :) - and `ApplyDeltaCommand` which locks on different keys, and who-knows-on what nodes. And applies deltas, as well.

danberindei:
@rvansa Ok, if you say `MergeOnStore` is a hack that is only ever going to be used by FGAMs I believe you, but you should be aware that it's a public interface in a public package.

rvansa:
@danberindei Once we had to add the check into the code, I made it public since I like extensibility in general. I hope I've defined the behaviour well enough. Maybe it was not a wise choice - if you feel strongly about it, we still have some time to hide the interface. The only thing I'd like to have in mind is that IMO AtomicMap should be moved to its own module in 10.0, rather than staying in core.

wburns:
Sorry this has conflicts now :(

pruivo:
@rvansa needs rebase. @danberindei just press the button when finish the discussion :)

rvansa:
Rebased.

danberindei:
@rvansa Still getting a duplicate externalizer id error on my machine:

    18:51:23,137 ERROR (testng-SingleClusterExecutorTest) [TestSuiteProgress] Test failed: org.infinispan.manager.SingleClusterExecutorTest.testExecutorTriConsumer org.infinispan.manager.EmbeddedCacheManagerStartupException: org.infinispan.commons.CacheConfigurationException: ISPN000423: Duplicate id found! AdvancedExternalizer id=122 is shared by another externalizer (org.infinispan.atomic.impl.ApplyDelta$Externalizer)

danberindei:
@rvansa `MergeOnStore` can't be both a hack and a reasonable public API that should be used by any application at the same time, you have to decide on one :) I vote to make it internal API, because I hope the functional API is going to be good enough to support FGAMs at some point in the future, and I don't want to keep supporting `MergeOnStore` after that.

rvansa:
@danberindei I've corrected the `Ids` and piggy-backed a commit moving `MergeOnStore` to `o.i.container` which is (a bit surprisingly) non-public.

danberindei:
@rvansa That is surprising, considering that `DataContainer` is exposed in the configuration, but I'm fine with it. I'll just wait for another run in CI.

wburns:
Looks like there some related failures http://ci.infinispan.org/job/Infinispan/job/PR-5187/58

wburns:
Actually it looks like this was already failing http://ci.infinispan.org/job/Infinispan/job/master/lastSuccessfulBuild/#showFailuresLink

wburns:

PR #5190, opened 2017-06-09
ISPN-7906 Infinispan Query DSL does not handle inheritance of propert…

anistor:
…ies/fields correctly

https://issues.jboss.org/browse/ISPN-7906


gustavonalle:
Integrated on 9.0.x, thanks @anistor

PR #5191, opened 2017-06-09
Ispn 7827 - Additional documentation for Cluster Executor

wburns:
https://issues.jboss.org/browse/ISPN-7827

Also broke out a new executing code in the grid section


documentation/src/main/asciidoc/user_guide/execute_grid.adoc
```diff
@@ -0,0 +1,163 @@
...
+in the cluster.
+
+This manager was built specifically using Java 8 and such has functional APIs in mind, thus all methods take a functional
+inteface as an argument. Also since these arguments will be sent to other nodes they need to be serializable.  We even
+used a nice trick to ensure our lambdas are immediately Serializable.  That is by having the arguments implement both
+Serializable and the real argument type (ie. Runnable or Function).  The JRE will pick the most specific class when
+determining which method to invoke, so in that case your lambdas will always be serializable.
+
+The manager by default will submit a given command to all nodes in the cluster including the node
+where it was submitted from. It is possible to further reduce which nodes it will execute upon or
```
galderz: Maybe add a little basic example here? The rest of the examples shown are focused on specific options.

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
```diff
@@ -0,0 +1,163 @@
...
+method. The given command will be resubmitted again to a single node until either
+the command completes without exception or the total submission amount is equal to the provided
+failover count. If however a `org.infinispan.util.concurrent.TimeoutException` is encountered
+either from the command or by the command timing out as defined in `timeout` then it will
+terminate the failover process and complete any pending CompletableFutures with the
+exception.
+
+
+==== Example: Dynamically Start and Stop Clustered Cache
+This example shows how you can use the ClusterExecutor to dynamically start and stop a cache.
```
galderz: Will this example be still relevant once we have cluster wide cache starts? I'm wondering if we should have a different example whose lifetime might be longer...

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+----
+
+The first argument is a `Function` that when invoked will pass the `EmbeddedCacheManager` local to each
+node.  Normally this also allows for a return value to be sent back, but unfortunately a `Cache`
+instance is not serializable so we can't send that back to the calling node.  Thus we have to
+return null.
+In this case the second argument `TriConsumer` would be called back for each node and will contain
+who this response is from (address), the return value (if there was one, in our case this is always
+null), and a throwable if a problem occurred.  The value and throwable variables will never both
```
galderz: Hmmm, this is calling for `Either` type, which we define `org.infinispan.commons.util` package. The idea of `Either` is that a type can be either one of two options: `Left` or `Right`. This type comes originally from Haskell, where `Left` is normally used for the error part, and `Right` for the value part.

galderz: So, you'd have a `BiConsumer` of `Adress` and `Either<? super Throwable, ? super V>`.

documentation/src/main/asciidoc/user_guide/server_tasks.adoc
```diff
@@ -0,0 +1,5 @@
+=== Server Tasks
+
+TODO: This section needs to be completed still
```
galderz: I've created [ISPN-7935](https://issues.jboss.org/browse/ISPN-7935) to address that. I'll fix that once its dependencies are fixed.

documentation/src/main/asciidoc/user_guide/streams.adoc
```diff
@@ -534,5 +539,41 @@ public class WordFrequencyExample {
...
+a specific word.
+
+[source,java]
+----
+public class RemoveBadWords {
+   public static void main(String[] args) {
+      // Lines removed
+      String word = ..
+
+      c1.entrySet().parallelStream()
```
galderz: Hmmm, `parallelStream`? From what I've gathered, you need to have a lot of entries for the node-level parallel stream stuff to be worth using it...

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
wburns: Sure I can try to do something else here.

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
wburns: The options below do that, maybe I should say this "... can be done with the various `filterTargets` methods, which are detailed below" ?

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
wburns: Got it, yeah I am fine with this. Maybe I should add this override and deprecate the other (to be removed in 10.0) ?

documentation/src/main/asciidoc/user_guide/server_tasks.adoc
wburns: Thanks 👍

documentation/src/main/asciidoc/user_guide/streams.adoc
wburns: That depends :) In our case we are calling `cache.remove` which can take a while. It all depends on the terminal operator and the how many elements.

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
```diff
@@ -0,0 +1,163 @@
...
+handle cases where an exception occurred during the processing of a given command by retrying
+the command again.
+When this occurs the Cluster Executor will choose a single node again to resubmit the command to
+up to the desired number of failover attempts. Note the chosen node could be any node that passes
+the topology or predicate check. Failover is enabled by invoking the overridden
+link:{javadocroot}/org/infinispan/manager/ClusterExecutor.html#singleNodeSubmission-int-[singleNodeSubmission]
+method. The given command will be resubmitted again to a single node until either
+the command completes without exception or the total submission amount is equal to the provided
+failover count. If however a `org.infinispan.util.concurrent.TimeoutException` is encountered
+either from the command or by the command timing out as defined in `timeout` then it will
```
danberindei: Why treat `TimeoutException` differently? The task could just catch the `TimeoutException` and complete normally, if it doesn't want to be retried.

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
danberindei: +1

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
danberindei: `CompletableFuture` uses `BiConsumer<V, Throwable>`. I wouldn't want to diverge from this, since it's the API that everyone knows.

documentation/src/main/asciidoc/user_guide/streams.adoc
```diff
@@ -534,5 +539,41 @@ public class WordFrequencyExample {
...
+         .filter(e -> e.getValue().contains(word))
+         .forEach((c, e) -> c.remove(e.getKey());
+----
+
+If we carefully note what is serialized and what is not, we notice that only the word along
+with the operations are serialized across to other nods as it is captured by the lambda.
+However the real saving piece is that the cache operation is performed on the primary
+owner thus reducing the amount of network traffic required to remove these values from the
+cache. The cache is not captured by the lambda as we provide a special BiConsumer method
+override that when invoked on each node passes the cache to the BiConsumer
```
danberindei: How about "The cache itself is not serializable, so it cannot be captured by a serializable lambda. Instead, we have to use the `forEach(BiConsumer<Cache<K, V>, ? super V> action)` overload, so that the lambda receives the cache as a parameter when it is executed on each node.`

Also, missing `.` at the end.

documentation/src/main/asciidoc/user_guide/streams.adoc
```diff
@@ -534,5 +539,41 @@ public class WordFrequencyExample {
...
+cache. The cache is not captured by the lambda as we provide a special BiConsumer method
+override that when invoked on each node passes the cache to the BiConsumer 
+
+One thing to keep in mind using the forEach command in this manner is that the underlying
+stream obtains no locks. The cahe remove operation will still obtain locks naturally, but
+the value could have changed from what the stream saw. That means that the entry could
+have been changed after the stream read it but the remove actually removed it.
+
+We have specifically added a new variant which is called LockedStream which will be
+covered in the next section.
```
danberindei: I suggest "a new variant called LockedStream, which does guarantee that the entries seen by the `forEach` action are up to date, and which is covered in the next section."

documentation/src/main/asciidoc/user_guide/streams.adoc
```diff
@@ -534,5 +539,41 @@ public class WordFrequencyExample {
...
+
+If we carefully note what is serialized and what is not, we notice that only the word along
+with the operations are serialized across to other nods as it is captured by the lambda.
+However the real saving piece is that the cache operation is performed on the primary
+owner thus reducing the amount of network traffic required to remove these values from the
+cache. The cache is not captured by the lambda as we provide a special BiConsumer method
+override that when invoked on each node passes the cache to the BiConsumer 
+
+One thing to keep in mind using the forEach command in this manner is that the underlying
+stream obtains no locks. The cahe remove operation will still obtain locks naturally, but
```
danberindei: Typo: cahe

documentation/src/main/asciidoc/user_guide/streams.adoc
danberindei: Shouldn't the guide talk about when parallel streams make sense?

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
```diff
@@ -0,0 +1,163 @@
+==  Executing code in the Grid
+
+The main benefit of a Cache is the ability to very quickly lookup a value
+by its key, even across machines. In fact this use alone is probably the reason
+many users use Infinispan. However Infinispan can provide much more benefit
+that aren't immediately apparent. Since Infinispan is usually used in
```
danberindei: benefit should be benefits

documentation/src/main/asciidoc/user_guide/distributed_executor.adoc
```diff
@@ -1,12 +1,17 @@
 [[DistributedExecutor]]
-== Distributed Execution
+=== Distributed Execution
+
+NOTE: Distributed Executor has been deprecated as of Infinispan 9.1.
+Users should use either a Cluster Executor or Distributed Stream to
+perform the operations they were doing before.
```
danberindei: This is the user guide, so we could say "You should".

core/src/main/java/org/infinispan/distexec/DistributedExecutorService.java
```diff
@@ -45,6 +46,7 @@
  * @author Vladimir Blagojevic
  *
  * @since 5.0
+ * @deprecated since 9.1 Please use {@link EmbeddedCacheManager#executor()} or {@link org.infinispan.CacheStream} instead.
```
danberindei: For symmetry, I'd either replace `{@link EmbeddedCacheManager#executor()}` with `{@link ClusterExecutor}`, or `{@link org.infinispan.CacheStream}` with `{@link org.infinispan.CacheCollection#stream()}`.

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
wburns: I will wait on this. If you want I would say write something to dev list @galderz and we can discuss more there.

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+NOTE: The ClusterExecutor is specifically designed for executing code where the code is not reliant
+upon the data in a cache and is used instead as a way to help users to execute code easily
+in the cluster.
+
+This manager was built specifically using Java 8 and such has functional APIs in mind, thus all methods take a functional
+inteface as an argument. Also since these arguments will be sent to other nodes they need to be serializable.  We even
+used a nice trick to ensure our lambdas are immediately Serializable.  That is by having the arguments implement both
+Serializable and the real argument type (ie. Runnable or Function).  The JRE will pick the most specific class when
+determining which method to invoke, so in that case your lambdas will always be serializable.
```
wburns: Sure :)

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
```diff
@@ -0,0 +1,163 @@
...
+As was mentioned it is possible to limit on which nodes the command will be ran. For example you may
+want to only run a computation on machines in the same rack. Or you may want to perform an operation
+once in the local site and again on a different site. A cluster executor can limit what nodes it sends
+requests to at the scope of same or different machine, rack or site level.
+
+[source,java]
+.SameRack.java
+----
+   EmbeddedCacheManager manager = ...;
+   manager.executor().filterTargets(ClusterExecutionPolicy.SAME_RACK).submit(...)
```
wburns: Okay

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
```diff
@@ -0,0 +1,163 @@
...
+----
+
+To use this you must enable topology aware consistent hashing through
+<<ServerHinting, Server Hinting>>.
+
+==== Timeout
+
+Cluster Executor allows for a timeout to be set per invocation. This defaults to the distributed sync timeout
+as configured on the Transport Configuration. This timeout works in both a clustered and non clustered
+cache manager. There are no guarantees as to if the execution is attempted to be interrupted or not. However
```
wburns: Sounds good

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
wburns: This is how `Distributed Executor` used to work. I agree that having a special case for `TimeoutException` is a bit odd. I created https://issues.jboss.org/browse/ISPN-7978

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
wburns: This sentence was awkward, I cleaned it up a bit.

documentation/src/main/asciidoc/user_guide/streams.adoc
wburns: Changed

documentation/src/main/asciidoc/user_guide/streams.adoc
wburns: Sure

core/src/main/java/org/infinispan/distexec/DistributedExecutorService.java
wburns: Changed to ClusterExecutor

documentation/src/main/asciidoc/user_guide/streams.adoc
wburns: I have added a paragraph to the Parallel Computation section detailing this more.

wburns:
Fixed rework comments.

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
```diff
@@ -0,0 +1,189 @@
...
+a cluster of machines we also have features available that can help utilize
+the entire cluster for performing the user's desired workload.
+
+NOTE: This section covers only executing code in the grid using an embedded cache,
+if you are using a remote cache you should check out
+<<executing_code_in_the_remote_grid, Executing code in the Remote Grid>>.
+
+=== Cluster Executor
+
+Since you have a group of machines, why not off load execution of code to them.
```
tristantarrant: "off-load".
Also I don't like the faux-question style, so I'd turn the sentence to something like "Since you have a group of machines, it makes sense to leverage their combined computing power for executing code on all of them".

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
```diff
@@ -0,0 +1,189 @@
...
+NOTE: The ClusterExecutor is specifically designed for executing code where the code is not reliant
+upon the data in a cache and is used instead as a way to help users to execute code easily
+in the cluster.
+
+This manager was built specifically using Java 8 and such has functional APIs in mind, thus all methods take a functional
+inteface as an argument. Also since these arguments will be sent to other nodes they need to be serializable.  We even
+used a nice trick to ensure our lambdas are immediately Serializable.  That is by having the arguments implement both
+Serializable and the real argument type (ie. Runnable or Function).  The JRE will pick the most specific class when
+determining which method to invoke, so in that case your lambdas will always be serializable.
+It is also possible to use an <<_user_defined_externalizers, Externalizer>> to ensure greatest performance.
```
danberindei: Unless you want to add some benchmark results, I wouldn't boast about "greatest performance" :)
In fact, performance with `Externalizer` should be about the same as with the default serialization, it's only `AdvancedExternalizer` that sends an int id instead of the full class name.

wburns: The link shows both :) I will change it to "to possibly reduce message size further".

documentation/src/main/asciidoc/user_guide/execute_grid.adoc
wburns: Fixed

wburns:
Updated

tristantarrant:
Merged, thanks

PR #5193, opened 2017-06-12
ISPN-7780 multimap embedded

karesti:
https://issues.jboss.org/browse/ISPN-7780

Design doc PR : 

https://github.com/infinispan/infinispan-designs/pull/3


multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
```diff
@@ -0,0 +1,91 @@
...
+      initAndTest();
+      MultimapCache<String, String> nonOwner = multimapCacheCluster.get(getFirstNonOwner("names").getCacheManager().getAddress().toString());
+
+      await(nonOwner.put("names", "julien").thenAccept(v -> TestingUtil.sleepThread(100)));
+      assertOnAllCachesAndOwnership("names", "julien");
+   }
+
+   @Override
+   protected void assertOwnershipAndNonOwnership(Object key, boolean allowL1) {
+      for (MultimapCache<String, String> mc : multimapCacheCluster.values()) {
```
rvansa: you could iterate through `caches()` directly.

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
```diff
@@ -0,0 +1,91 @@
...
+      }
+   }
+
+   @Override
+   protected void assertOnAllCaches(Object key, String value) {
+      for (Map.Entry<String, MultimapCache<String, String>> entry : multimapCacheCluster.entrySet()) {
+         await(entry.getValue().get((String) key).thenAccept(v -> {
+                  assertNotNull(format("values on the key %s must be not null", key), v);
+                  assertTrue(format("values on the key '%s' must contain '%s' on node '%s'", key, value, entry.getKey()), v.contains(value));
+                  // Allow some time for all ClusteredGetCommands to finish executing
```
rvansa: What? You should avoid sleeps wherever possible.

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
```diff
@@ -0,0 +1,91 @@
...
+   protected void initAndTest() {
+      for (MultimapCache<String, String> mc : multimapCacheCluster.values()) {
+         await(mc.size().thenAccept(s -> assertEquals(0, s.intValue())));
+      }
+
+      await(multimapCacheCluster.values().iterator().next().put("names", "oihana"));
+      assertOnAllCachesAndOwnership("names", "oihana");
+   }
+
+   public void testPutFromNonOwner() {
```
rvansa: You probably want to test both owners and non-owners in the same test

multimap/src/test/java/org/infinispan/multimap/impl/EmbeddedMultimapCacheTest.java
```diff
@@ -0,0 +1,163 @@
...
+      return cm;
+   }
+
+   public void testSupportsDuplicates() {
+      assertFalse(multimapCache.supportsDuplicates());
+   }
+
+   public void testPut() {
+      await(
+            multimapCache.put("names", "julien").thenCompose(r1 ->
```
rvansa: You're chaining it wrong, should be:
```java
 multimapCache.put("names", "julien")
   .thenCompose(r -> multimapCache.put("names", "oihana"))
   .thenCompose(r -> multimapCache.put("names", "julien"))
   .thenCompose(r -> multimapCache.get("names"))
```
That applies to the rest of chains as well.

multimap/src/test/java/org/infinispan/multimap/impl/EmbeddedMultimapCacheTest.java
```diff
@@ -0,0 +1,163 @@
...
+      multimapCache = EmbeddedMultimapCacheManagerFactory.get("test", cm, c.build());
+      return cm;
+   }
+
+   public void testSupportsDuplicates() {
+      assertFalse(multimapCache.supportsDuplicates());
+   }
+
+   public void testPut() {
+      await(
```
rvansa: unless you need the timeout, use `CF.join()` instead of `await`

multimap/src/test/java/org/infinispan/multimap/impl/TxEmbeddedMultimapCacheTest.java
```diff
@@ -0,0 +1,181 @@
...
+import org.testng.annotations.Test;
+
+/**
+ * Multimap Cache with transactions in single cache
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.1
+ */
+@Test(groups = "functional", testName = "multimap.TxEmbeddedMultimapCacheTest")
+public class TxEmbeddedMultimapCacheTest extends EmbeddedMultimapCacheTest {
```
rvansa: You should probably test with both pessimistic and optimistic transactions, and pessimistic with both RC and RR isolation levels. Also multi-node tests can reveal more issues than single-node ones, I think (testing both owner and non-owner variant).

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.Collection;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Predicate;
+
+/**
+ * MultimapCache provides the common building block for the two different types of multimap caches that Infinispan
+ * provides: embedded and remote.
+ * <p>
+ * Please see the <a href="http://www.jboss.org/infinispan/docs">Infinispan documentation</a> and/or the <a
```
rvansa: These links are outdated and don't say anything about multimap cache.

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,91 @@
...
+ * href="https://docs.jboss.org/author/display/ISPN/Getting+Started+Guide#GettingStartedGuide-5minutetutorial">5 Minute
+ * Usage Tutorial</a> for more details.
+ * <p/>
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @see <a href="http://www.jboss.org/infinispan/docs">Infinispan documentation</a>
```
rvansa: outdated link

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,91 @@
...
+
+   /**
+    * Returns the number of key-value pairs in this multimap cache.
+    * It doesn't return the distinct number of keys
+    *
+    * For now, this method returns a completed future
+    *
+    * @return long, the size
+    */
+   CompletableFuture<Integer> size();
```
karesti: yes, I was thinking about that earlier too

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,91 @@
...
+    * Returns true when the value has been removed from the key.
+    *
+    * @param key, to be removed
+    * @param value, to be removed
+    * @return boolean, remove result
+    */
+   CompletableFuture<Boolean> remove(K key, V value);
+
+   /**
+    * Removes all the values in every key that matches the given {@link Predicate}.
```
danberindei: @rvansa Actually matches is correct here :)

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
```diff
@@ -0,0 +1,96 @@
...
+   }
+
+   @Override
+   public CompletableFuture<Boolean> remove(K key, V value) {
+      return readWriteMap.eval(key, new RemoveFunction<>(key, value));
+   }
+
+   @Override
+   public CompletableFuture<Void> remove(Predicate<? super V> p) {
+      class Remover {
```
karesti: because at first I was passing this code to the CompletableFuture.runAsync(...)
but it makes no more sense, indeed. But I'm not happy with these implementations anyway, but well, as I said, it's a first step ...

multimap/src/main/java/org/infinispan/multimap/impl/function/PutFunction.java
```diff
@@ -0,0 +1,72 @@
...
+      this.value = value;
+   }
+
+   @Override
+   public Void apply(EntryView.ReadWriteEntryView<K, Collection<V>> entryView) {
+      Collection<V> newValues = new HashSet<>();
+      newValues.add(value);
+
+      try {
+         // Raises NoSuchElementException if the value does not exist
```
karesti: Yes, and tbh I don't know why I did not use it...

multimap/src/main/java/org/infinispan/multimap/impl/function/RemoveFunction.java
```diff
@@ -0,0 +1,93 @@
...
+         Collection<V> values = entryView.get();
+
+         if(value == null) {
+            // Remove the key
+            entryView.remove();
+            removed = true;
+
+         } else if (values.contains(value)){
+            //Remove key/value
+            Collection<V> newValues = new HashSet<>();
```
karesti: when the value is null, I remove the key from the cache. When the value is not null, I remove that particular value in the key, if exists. But the "true" result is not ok if the key does not exist. A test should be added

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
```diff
@@ -0,0 +1,91 @@
...
+public class DistributedMultimapCacheTest extends BaseDistFunctionalTest<String, Collection<String>> {
+
+   private Map<String, MultimapCache<String, String>> multimapCacheCluster = new HashMap<>();
+
+   @Override
+   protected void createCacheManagers() throws Throwable {
+      super.createCacheManagers();
+
+      for (EmbeddedCacheManager cacheManager : cacheManagers) {
+         multimapCacheCluster.put(cacheManager.getAddress().toString(),
```
karesti: because I was testing some things  at first and I forgot to change it back to address ! thanks for pointing it

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
```diff
@@ -0,0 +1,91 @@
...
+               EmbeddedMultimapCacheManagerFactory.get(cacheName, cacheManager, configuration.build()));
+      }
+   }
+
+   @Override
+   protected void initAndTest() {
+      for (MultimapCache<String, String> mc : multimapCacheCluster.values()) {
+         await(mc.size().thenAccept(s -> assertEquals(0, s.intValue())));
```
karesti: I created an assertion for the size that I will share between tests. Distributed test is in very WIP because it fails with eval

multimap/src/test/java/org/infinispan/multimap/impl/EmbeddedMultimapCacheTest.java
danberindei: @karesti please do use a timeout in tests, even if you think you don't need it :)

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
karesti: yes. This test is very wip yet

multimap/src/test/java/org/infinispan/multimap/impl/EmbeddedMultimapCacheTest.java
karesti: ok

multimap/src/test/java/org/infinispan/multimap/impl/TxEmbeddedMultimapCacheTest.java
karesti: good idea

multimap/src/test/java/org/infinispan/multimap/impl/EmbeddedMultimapCacheTest.java
karesti: ok ;)

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
wburns: 👍

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,91 @@
+package org.infinispan.multimap.api;
+
+import java.util.Collection;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Predicate;
+
+/**
+ * MultimapCache provides the common building block for the two different types of multimap caches that Infinispan
+ * provides: embedded and remote.
+ * <p>
```
wburns: We need more general documentation about this class

1. as Radim mentioned whether this supports null values or keys (it shouldn't)
2. talk about supporting duplicates and that this is determined by Object equality
3. talk about it being an async API ?
4. I would talk about all return values being views and thus updates are not reflected
5. Talk about how eviction is only counted per key and doesn't involve values

Also I forgot about this before, but we have no method to iterate over the multi map, guessing users may want this. We can always do that later though.

multimap/src/main/java/org/infinispan/multimap/impl/function/PutFunction.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.HashSet;
+import java.util.NoSuchElementException;
+import java.util.Set;
+
+import org.infinispan.commons.api.functional.EntryView;
+import org.infinispan.commons.marshall.AdvancedExternalizer;
+import org.infinispan.multimap.impl.ExternalizerIds;
+
+/**
+ * The adding function to remove a ket or a key/value on the multimap
```
wburns: This needs to be reworded.

multimap/src/main/java/org/infinispan/multimap/impl/function/RemoveFunction.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.HashSet;
+import java.util.NoSuchElementException;
+import java.util.Set;
+
+import org.infinispan.commons.api.functional.EntryView;
+import org.infinispan.commons.marshall.AdvancedExternalizer;
+import org.infinispan.multimap.impl.ExternalizerIds;
+
+/**
+ * The adding function to remove a ket or a key/value on the multimap
```
wburns: This needs to be reworded.

multimap/src/main/java/org/infinispan/multimap/impl/function/RemoveFunction.java
```diff
@@ -0,0 +1,93 @@
...
+      try {
+         // Raises NoSuchElementException if the key/value does not exist
+         Collection<V> values = entryView.get();
+
+         if(value == null) {
+            // Remove the key
+            entryView.remove();
+            removed = true;
+
+         } else if (values.contains(value)){
```
wburns: Can we not just call `values.remove(value)` ? And then if something was invoked check `isEmpty` is true remove the key?

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
rvansa: @danberindei it's not a the **key** that matches the predicate, since the predicate does not accept the key, therefore *values ... match*

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
wburns: We could just use the ASYNC_OPERATIONS_EXECUTOR to make all the methods "async" that don't yet support async. I personally don't like having the interface Javadoc saying whether a method is ran synchronous or not as it is exposing implementation details.

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
```diff
@@ -0,0 +1,96 @@
...
+
+   @Override
+   public CompletableFuture<Void> put(K key, V value) {
+      return readWriteMap.eval(key, new PutFunction<>(key, value));
+   }
+
+   @Override
+   public CompletableFuture<Collection<V>> get(K key) {
+      SerializableBiFunction<? super Collection<V>, Throwable, Collection<V>> emptyCollectionWhenNull = (vs, throwable) -> vs == null ? new HashSet<>() : vs;
+      return cache.getAsync(key).handle(emptyCollectionWhenNull);
```
wburns: We also need to make sure this is a copy of the collection, just in case. Otherwise if we are in a clustered cache data would get in an inconsistent state if user updates this collection on nodes that own the data.

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
```diff
@@ -0,0 +1,96 @@
...
+         public Remover(Predicate<? super V> p) {
+            this.p = p;
+         }
+
+         public Void removeEveryMatchingValue() {
+            cache.keySet().stream().forEach((c, key) -> c.computeIfPresent(key, (o, o1) -> {
+               Collection<V> values = (Collection<V>) o1;
+               Collection<V> newValues = new HashSet<>();
+
+               for (V v : values) {
```
wburns: Seems we could just use the iterator on the collection and just call remove if it doesn't pass the predicate.

multimap/src/main/java/org/infinispan/multimap/impl/MultimapModuleLifecycle.java
```diff
@@ -0,0 +1,35 @@
...
+import org.infinispan.configuration.global.GlobalConfiguration;
+import org.infinispan.factories.GlobalComponentRegistry;
+import org.infinispan.lifecycle.AbstractModuleLifecycle;
+import org.infinispan.lifecycle.ModuleLifecycle;
+import org.infinispan.multimap.impl.function.PutFunction;
+import org.infinispan.multimap.impl.function.RemoveFunction;
+import org.kohsuke.MetaInfServices;
+
+/**
+ * Used to add exterlizers
```
wburns: Typo

multimap/src/main/java/org/infinispan/multimap/impl/function/BaseFunction.java
```diff
@@ -0,0 +1,14 @@
...
+import org.infinispan.commons.api.functional.EntryView;
+import org.infinispan.util.function.SerializableFunction;
+
+/**
+ * A base function for the multimap updates
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.1
+ */
+abstract class BaseFunction<K, V, R> implements SerializableFunction<EntryView.ReadWriteEntryView<K, Collection<V>>, R> {}
```
wburns: This can just be an interface

multimap/src/main/java/org/infinispan/multimap/impl/function/PutFunction.java
```diff
@@ -0,0 +1,72 @@
...
+   }
+
+   @Override
+   public Void apply(EntryView.ReadWriteEntryView<K, Collection<V>> entryView) {
+      Collection<V> newValues = new HashSet<>();
+      newValues.add(value);
+
+      try {
+         // Raises NoSuchElementException if the value does not exist
+         Collection<V> values = entryView.get();
```
wburns: Hrmm if the entry exists can't we just add the new value to it and set it back in the view? Otherwise the update will get progressively slower.

multimap/src/test/java/org/infinispan/multimap/impl/TxEmbeddedMultimapCacheTest.java
wburns: Also if you can using the factory or @Factory annotation would be preferred over using different classes. Unless there are a lot of differences between how the tests run.

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,91 @@
...
+   CompletableFuture<Boolean> remove(K key, V value);
+
+   /**
+    * Removes all the values in every key that matches the given {@link Predicate}.
+    *
+    * For now, this method returns a completed future
+    *
+    * @param p, the predicate to be tested on every value in the multimap cache
+    */
+   CompletableFuture<Void> remove(Predicate<? super V> p);
```
wburns: Oh also we probably want a `SerializablePredicate` override for this.

karesti: @wburns Yep. I wanted to add it but this class is in commons  and SerializablePredicate is in core ... This joins the pb Radim asked about on a mail last friday.

multimap/src/main/java/org/infinispan/multimap/impl/function/RemoveFunction.java
rvansa: @wburns No, the current value must be treated as immutable (that holds generally for all functional commands).

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
rvansa: nope, as above

multimap/src/main/java/org/infinispan/multimap/impl/function/RemoveFunction.java
wburns: This is just to prevent ConcurrentModificationException ? Cause as soon as you call  `entryView.set` wouldn't that replace the current value even without tx?

wburns: Never mind looking at the code I see it does update the tx context, it is just an issue on the originator.

multimap/src/main/java/org/infinispan/multimap/impl/function/RemoveFunction.java
```diff
@@ -0,0 +1,93 @@
...
+   }
+
+   @Override
+   public Boolean apply(EntryView.ReadWriteEntryView<K, Collection<V>> entryView) {
+      boolean removed = false;
+      try {
+         // Raises NoSuchElementException if the key/value does not exist
+         Collection<V> values = entryView.get();
+
+         if(value == null) {
```
wburns: Spacing

multimap/src/main/java/org/infinispan/multimap/impl/function/RemoveFunction.java
rvansa: @wburns There are situations where the command is retried, even in non-tx case, and that would result in modification being applied twice and/or in a wrong order. Generally the modifications in context should be isolated from the value that is held in DC, as we should be able to roll them back any time before we commit the entry to DC.

wburns: Yes I agree I didn't mean that it only updated the tx context.

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
wburns: Same thing as I posted on there, we should just add an additional interface in commons for this. We can worry about refactoring core in a major version. At least my 2cents.

rvansa: Actually, predicate-based methods won't fly in client-server - Hot Rod is language neutral and therefore can't carry lambdas. That's why the `SerializableFunction` et all are in core and not in commons.
For client-server you'll need to register the filter on server-side, and refer to it by name. Alternatively you could do the filtering using Ickle query.

wburns:
I am more than aware of that, hence my email before. But this doesn't stop
the Java client from performing a lambda on results from the server.

If we want to have separate, but similar, API for embedded and remote I am
fine with that. Then to me this API shouldn't exist in commons. We could
have a super set, such as BaseFeature that you mentioned in the email.

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,91 @@
...
+
+   /***
+    * Returns a view collection of the values associated with key in this multimap cache, if any.
+    * Any changes to the retrieved collection won't change the values in this multimap cache.
+    *
+    * @param key, to be retrieved
+    * @return the collection found in the key when it's present,
+    *         empty Collection when the key is not present
+    */
+   CompletableFuture<Collection<V>> get(K key);
```
tsegismont: Will this be a copy or a view of the values in the cluster?

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,91 @@
...
+    * @param key, to be removed
+    * @param value, to be removed
+    * @return boolean, remove result
+    */
+   CompletableFuture<Boolean> remove(K key, V value);
+
+   /**
+    * Removes all the values in every key that matches the given {@link Predicate}.
+    *
+    * For now, this method returns a completed future
```
tsegismont: Does this behavior apply behavior apply to all methods?

karesti: @tsegismont no, just the methods that are marked with the doc. But this might change soon. I have to speak to the team to see the best way to do it.

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
karesti: @wburns good point. Changing implementation

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
karesti: @wburns I'm +1 on it. I was struggling with this code til I understood why it was failing when I has using the CF.runAsync(...) and to be able to share with you a preview, I followed Radim's suggestion on returning a completable future. I'm going to search how to use the ASYNC_OPERATIONS_EXECUTOR to implement it otherwise. Thanks !

multimap/src/main/java/org/infinispan/multimap/impl/function/PutFunction.java
karesti: haha totally. Sorry guys I'm still so awful in english writing. Promise, I will improve !

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
rvansa: @wburns @karesti I don't agree - the collection stored in cache should be immutable, so we should not create a *copy* but rather wrap it in `Collections.unmodifiableCollection`.

wburns:
Talking to @karesti about the async executor, I guess we can't use this in a tx context as there is no way to get the tx context into the spawned thread. And I don't know if I like having to have a method to inject an InvocationContext or Transaction although that would be possible as part of AdvancedCache API / DecoratedCache. Guess I am fine with it blocking for now :/

rvansa:
@wburns Why can't you suspend and resume the transaction? You can even call
```java
Transaction tx = transactionManager.getTransaction();
executor.execute(() -> {
    transactionManager.resume(tx);
    // do something else
}
```

wburns:
> @wburns  Why can't you suspend and resume the transaction?

@rvansa That was my initial thought as well, but we can't do this as the calling thread has to keep the transaction on it.

> You can even call \<code\>

@rvansa Afaik you can't share the transaction between multiple threads. The old async operations never resumed the transaction in the async thread when an explicit tx was present, they just updated the `InvocationContext`. Are you sure that we can have more than 1 thread associated with the same transaction? You probably know more about the finer details of the xa spec than I do (the javadoc don't seem to detail this case). Would all the various `TransactionManager`s support this?

rvansa:
@wburns As you've said, JTA is not clear in this part so I've asked Narayana guys and they've confirmed that there's no issue with it in JBoss TM. The first thread that calls `tm.commit()`/`tm.rollback()` ends the transaction, and TM prevents multiple invocations. However, if you have an ongoing transaction and other thread commits it while you're doing sth. with that, problems may arise - it's up to the app to not call commit until all participants are ready.
[It seems that](http://fogbugz.atomikos.com/default.asp?community.6.803.3) it was not same with Atomikos but things may have changes since 2009. [Other report](https://stackoverflow.com/questions/6952294/how-to-share-one-transaction-between-multi-threads) that Bitronix does not support that.

wburns:
@rvansa So I am on the fence either way, whichever people prefer. The one thing I do like is that according to that stack overflow post that supposedly JTA spec does allow multiple threads to be associated with the same transaction. I double checked and that line is there at http://download.oracle.com/otndocs/jcp/jta-1.1-spec-oth-JSpec/

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
karesti: ok. I added more doc on the interface

karesti:
@wburns @rvansa concerning the explicit tx, I'm blocking whenever a explicit tx is ongoing for the methods that fail if I don't do so with the actual implementation. I've documented it on the interface.
I'm going to continue with other stuff on this multimap, and I will come to tx later if you don't mind.

multimap/src/main/java/org/infinispan/multimap/impl/ExternalizerIds.java
```diff
@@ -0,0 +1,6 @@
+package org.infinispan.multimap.impl;
+
+public interface ExternalizerIds {
+   Integer PUT_KEY_VALUE_FUNCTION = 3001;
```
karesti: Do you mean update the doc or update the numbers ? 
ClusteredCounters range are 2000-2011. I was assuming I can add another range starting on 3...

multimap/src/main/java/org/infinispan/multimap/impl/MultimapModuleLifecycle.java
karesti: ok

multimap/src/main/java/org/infinispan/multimap/impl/ExternalizerIds.java
rvansa: Just the docs. Please add range for clustered counters if there's not one.

tsegismont:
@karesti maybe a stupid question, but they say it's always worth asking :)

Looking at `EmbeddedMultimapCacheManagerFactory`, I see that a multicache is created from a backing "standard" cache. Does that mean that a multicache can be either distributed or replicated?

And since distribution happens on keys, primary and backups will have the whole collection for a key. Is that correct?

I'm just asking to understand, not saying if it's right or wrong.

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
danberindei: Oops, I should have read the function declaration before replying :)

Then I think s/matches/match/ is not enough, it should be something like "Remove the values that match the given Predicate from all the keys".

tsegismont:
Thanks @rvansa !

commons/src/main/java/org/infinispan/multimap/api/MultimapCache.java
karesti: This will be a copy. Collections.unmodifiableCollection(v)
Changes are not allowed, and if the user needs to modify the collection directly, it should create a new one from the answer.
What do you think ?

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
karesti: ok

multimap/src/test/java/org/infinispan/multimap/impl/EmbeddedMultimapCacheTest.java
karesti: I'm using await

karesti:
Compatibility mode can't work because of https://issues.jboss.org/browse/ISPN-7993

karesti:
The only commit concerned by multimap is the last one, the others have been rebased for testing. ISPN-7993 should be merged before this PR

galderz:
What are we waiting for to integrate this in? It was open early June :\

IMO, this should have included in 9.1.0.Final with follow up JIRAs for missing orthogonal features (transaction support, encoding support...etc) :(

galderz:
To note: commits for ISPN-7993 should be out of this PR.

karesti:
@galderz I removed the rebase from ISPN-7993 and removed the test that verifies this is working

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
```diff
@@ -0,0 +1,382 @@
...
+         MultimapCacheManager multimapCacheManager = EmbeddedMultimapCacheManagerFactory.from(cacheManager);
+         multimapCacheCluster.put(cacheManager.getAddress(), multimapCacheManager.get(cacheName));
+      }
+   }
+
+   @Override
+   protected void initAndTest() {
+      assertMultimapCacheSize(multimapCacheCluster, 0);
+
+      putValuesOnMultimapCache(multimapCacheCluster, NAMES_KEY, OIHANA);
```
rvansa: Why are you inserting the value on all nodes? If this is clustered, you should do it only once.

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
```diff
@@ -0,0 +1,382 @@
...
+import org.infinispan.multimap.api.EmbeddedMultimapCacheManagerFactory;
+import org.infinispan.multimap.api.MultimapCache;
+import org.infinispan.multimap.api.MultimapCacheManager;
+import org.infinispan.remoting.transport.Address;
+import org.testng.annotations.Test;
+
+@Test(groups = "functional", testName = "distribution.DistributedMultimapCacheTest")
+public class DistributedMultimapCacheTest extends BaseDistFunctionalTest<String, Collection<User>> {
+
+   protected Map<Address, MultimapCache<String, User>> multimapCacheCluster = new HashMap<>();
```
rvansa: I would rather store them in a list so that you can call `multimap(0)` in the same way as you do `cache(0)` in core testsuite.

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
```diff
@@ -0,0 +1,382 @@
...
+      return multimapCacheCluster.get(cache.getCacheManager().getAddress());
+   }
+
+   protected MultimapCache getMultimapCacheFirstOwner(String key) {
+      Cache<String, Collection<User>> cache = getFirstOwner(key);
+      return multimapCacheCluster.get(cache.getCacheManager().getAddress());
+   }
+
+   protected MultimapCache getMultimapCacheSecondOwner(String key) {
+      Cache<String, Collection<User>> cache = getSecondNonOwner(key);
```
rvansa: second non owner?

multimap/src/test/java/org/infinispan/multimap/impl/EmbeddedMultimapCacheTest.java
```diff
@@ -0,0 +1,308 @@
...
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.multimap.api.EmbeddedMultimapCacheManagerFactory;
+import org.infinispan.multimap.api.MultimapCache;
+import org.infinispan.multimap.api.MultimapCacheManager;
+import org.infinispan.test.SingleCacheManagerTest;
+import org.infinispan.test.fwk.TestCacheManagerFactory;
+import org.testng.annotations.Test;
+
+/**
+ * Single Multimap Cache Test
```
rvansa: And do we need both single-node and multinode test? Can be the code shared in any way?

multimap/src/test/java/org/infinispan/multimap/impl/StoreTypeMultimapCacheTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.infinispan.configuration.cache.StorageType;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.multimap.api.EmbeddedMultimapCacheManagerFactory;
+import org.infinispan.multimap.api.MultimapCache;
+import org.infinispan.multimap.api.MultimapCacheManager;
+import org.infinispan.remoting.transport.Address;
+import org.testng.annotations.Test;
+import org.junit.Ignore;
+
+//TODO: Integrate https://issues.jboss.org/browse/ISPN-7993 to make it pass
```
rvansa: let's move this test out of the PR if 7993 does not get in before that

persistence/jpa/src/main/java/org/infinispan/persistence/jpa/impl/EntityManagerFactoryRegistry.java
```diff
@@ -10,6 +10,8 @@
 import javax.persistence.spi.PersistenceProvider;
 
 import org.infinispan.commons.util.Util;
+import org.infinispan.factories.scopes.Scope;
```
rvansa: are these imports necessary?

karesti: no, this has been added by error and not removed on amend

multimap/src/test/java/org/infinispan/multimap/impl/EmbeddedMultimapCacheTest.java
karesti: well, configuration is different, I like having the simple test too, I'm going to rework it to avoid unnecessary duplication, I agree on that

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
karesti: not used method anymore, I removed

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
karesti: by error

multimap/src/test/java/org/infinispan/multimap/impl/DistributedMultimapCacheTest.java
```diff
@@ -0,0 +1,382 @@
...
+   }
+
+   @Override
+   protected Object[] parameterValues() {
+      return concat(super.parameterValues(), fromOwner ? Boolean.TRUE : Boolean.FALSE);
+   }
+
+   @Override
+   public Object[] factory() {
+      List testsToRun = new ArrayList<>();
```
karesti: no, this is because I went to tx tests and I was about to do something similar to create tests; I can refactor that, doesn't bother me

multimap/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,226 @@
...
+   /**
+    * Retrieves a CacheEntry corresponding to a specific key in this multimap cache.
+    *
+    * @param key the key whose associated cache entry is to be returned
+    * @return the cache entry to which the specified key is mapped, or
+    *         {@code null} if this multimap contains no mapping for the key
+    *
+    * @since 9.2
+    */
+   CompletableFuture<Optional<CacheEntry<K, Collection<V>>>> getEntry(K key);
```
karesti: I copied the java doc on entry and changed after to optional, didn't realize, thanks for pointing

multimap/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,226 @@
...
+
+   /**
+    * Asynchronous method. Puts a key-value pair in this multimap cache.
+    * <ul>
+    * <li>If this multimap cache supports duplicates, the value will be always added.</li>
+    * <li>If this multimap cache does <i>not support</i> duplicates and the value exists on the key,
+    * nothing will be done.</li>
+    * </ul>
+    *
+    * @param key,   the key to be put
```
karesti: why not, comma are pretty 💃 
TBH, I don't know why or when I took the idea from, and I repeated the , in my javadoc. Removing them

multimap/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,226 @@
...
+ *
+ *    multimapCache.get("k").thenAccept(values -> System.out.println(values.size()));
+ * </pre>
+ * prints the value 2. "k" -> ["v1", "v"]
+ * <p>
+ * Future implementations might evolve to support duplicated values.
+ * <p>
+ * <h2>Transactions</h2> MultimapCache supports implicit transactions without blocking. The following methods block when
+ * they are called in a explicit transaction context. This limitation could be improved in the followin versions if
+ * technically possible. <ul> <li>{@link MultimapCache#size()}</li> <li>{@link MultimapCache#containsEntry(Object,
```
karesti: didn't know for the closing li. Correcting

multimap/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,226 @@
...
+ *    multimapCache.put("k", "v2").join();
+ *
+ *    multimapCache.get("k").thenAccept(values -> System.out.println(values.size()));
+ * </pre>
+ * prints the value 2. "k" -> ["v1", "v"]
+ * <p>
+ * Future implementations might evolve to support duplicated values.
+ * <p>
+ * <h2>Transactions</h2> MultimapCache supports implicit transactions without blocking. The following methods block when
+ * they are called in a explicit transaction context. This limitation could be improved in the followin versions if
```
karesti: yep

multimap/src/main/java/org/infinispan/multimap/api/MultimapCacheManager.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.infinispan.configuration.cache.Configuration;
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+
+public interface MultimapCacheManager<K,V> {
+
+   /**
+    * Defines a named multimap cache's configuration by using the provided configuration
+    * If this cache was already configured either declaritively or programmatically this method will throw a
+    * {@link org.infinispan.commons.CacheConfigurationException}.
```
karesti: I created https://issues.jboss.org/browse/ISPN-8105 to handle that

multimap/src/main/java/org/infinispan/multimap/impl/function/BaseFunction.java
```diff
@@ -0,0 +1,14 @@
...
+import org.infinispan.functional.EntryView;
+import org.infinispan.util.function.SerializableFunction;
+
+/**
+ * A base function for the multimap updates
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+abstract class BaseFunction<K, V, R> implements SerializableFunction<EntryView.ReadWriteEntryView<K, Collection<V>>, R> {}
```
karesti: I created in class at first because I believed code will be shared, but indeed, looks more like that now, so I change it

multimap/src/main/java/org/infinispan/multimap/impl/function/GetFunction.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Serializable function used by {@link org.infinispan.multimap.impl.EmbeddedMultimapCache#get(Object)}
+ * to get a key's value.
+ *
+ * @author Katia Aresti - karesti@redhat.com
+ * @see <a href="http://infinispan.org/documentation/">Marshalling of Functions</a>
+ * @since 9.2
+ */
+public final class GetFunction<K, V> extends BaseFunction<K, V, Collection<V>> {
```
karesti: This is what I had before I realized that encoding was not supported

karesti: I have to restest it properly

multimap/src/main/java/org/infinispan/multimap/impl/function/GetFunction.java
```diff
@@ -0,0 +1,65 @@
...
+
+   public static final AdvancedExternalizer<GetFunction> EXTERNALIZER = new Externalizer();
+   private final K key;
+
+   public GetFunction(K key) {
+      this.key = key;
```
karesti: indeed, everywhere 🐙

multimap/src/main/java/org/infinispan/multimap/impl/function/PutFunction.java
```diff
@@ -0,0 +1,66 @@
...
+ * @since 9.2
+ */
+public final class PutFunction<K, V> extends BaseFunction<K, V, Void> {
+
+   public static final AdvancedExternalizer<PutFunction> EXTERNALIZER = new Externalizer();
+   private final K key;
+   private final V value;
+
+   public PutFunction(K key, V value) {
+      this.key = key;
```
karesti: removing them all

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
```diff
@@ -0,0 +1,176 @@
...
+      }));
+      return null;
+   }
+
+   private Boolean containsEntryInternal(V value) {
+      return cache.entrySet().stream().anyMatch(kCollectionEntry -> kCollectionEntry.getValue().contains(value));
+   }
+
+   private Long sizeInternal() {
+      return cache.values().stream().mapToLong(value -> value.size()).sum();
```
karesti: ok

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
```diff
@@ -0,0 +1,176 @@
...
+            if (!p.test(v))
+               newValues.add(v);
+         }
+         return newValues.isEmpty() ? null : newValues;
+      }));
+      return null;
+   }
+
+   private Boolean containsEntryInternal(V value) {
+      return cache.entrySet().stream().anyMatch(kCollectionEntry -> kCollectionEntry.getValue().contains(value));
```
karesti: ok

multimap/src/main/java/org/infinispan/multimap/api/MultimapCacheManager.java
rvansa: I don't think that such checks are necessary, I just want to document that multimap "x" is just a different view to cache "x".

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCache.java
rvansa: Btw the lambda inside should probably have its externalizer.

tristantarrant:

karesti:
@tristantarrant you scared me there

galderz:
Needs rebasing :|

karesti:
@rvansa @wburns @galderz any objections ?

karesti:
@galderz properly rebased

documentation/src/main/asciidoc/user_guide/multimapcache.adoc
```diff
@@ -0,0 +1,155 @@
+== Multimap Cache
+
+Infinispan Multimap Cache is a new distributed and clustered collection type introduced in Infinispan 9.
```
tristantarrant: 9.2

documentation/src/main/asciidoc/user_guide/multimapcache.adoc
```diff
@@ -0,0 +1,155 @@
...
+Asynchronous that returns the number of key-value pairs in the multimap cache. It doesn't return the distinct number of keys.
+
+==== boolean supportsDuplicates()
+Asynchronous that returns true if the multimap cache supports duplicates. This means that the content of the multimap can be
+'a' -> ['1', '1', '2']. For now this method will always return false, as duplicates are not yet supported.
+The existence of a given value is determined by 'equals' and `hashcode' method's contract.
+
+=== Creating a Multimap Cache
+
+In version 9.2, the MultimapCache is configured as a regular cache. This can be done either by code or XML configuration.
```
tristantarrant: Better not use version numbers, just say "Currently".

documentation/src/main/asciidoc/user_guide/multimapcache.adoc
```diff
@@ -0,0 +1,155 @@
...
+multimapCache = multimapCacheManager.get(multimapCacheName);
+----
+
+==== Server mode
+
+TODO
+
+=== Limitations
+
+In almost every case the Multimap Cache will behave as a regular Cache, but some limitations exist in the first 9.2 version.
```
tristantarrant: again, avoid version numbers.

documentation/src/main/asciidoc/user_guide/user_guide.asciidoc
```diff
@@ -36,3 +36,4 @@ include::rolling_upgrades.adoc[]
 include::extending.adoc[]
 include::architecture.adoc[]
 include::counters.adoc[]
+include::multimapcache.adoc[]
```
tristantarrant: I think we should move both counters and multimapcache docs further up, after query, before CDI.

multimap/src/main/java/org/infinispan/multimap/api/MultimapCache.java
```diff
@@ -0,0 +1,232 @@
...
+ *     multimapCache.put("k", null).join() -> fails
+ *     multimapCache.put("k", "v1").join() -> works and add's v1
+ *     multimapCache.containsKey("k").join() -> true
+ *     multimapCache.remove("k", "v1").join() -> works, removes v1 and as the remaining collection is empty, the key is
+ * removed
+ *     multimapCache.containsKey("k").join() -> false
+ *  </pre>
+ * <p>
+ * <p>
+ * <h2>Duplicates</h2> Today's implementation does not support duplicate values on keys. {@link Object#equals(Object)}
```
tristantarrant: s/Today's/The current/

multimap/src/main/java/org/infinispan/multimap/impl/EmbeddedMultimapCacheManager.java
```diff
@@ -0,0 +1,36 @@
...
+      this.cacheManager = embeddedMultimapCacheManager;
+   }
+
+   @Override
+   public Configuration defineConfiguration(String name, Configuration configuration) {
+      return cacheManager.defineConfiguration(name, configuration);
+   }
+
+   @Override
+   public MultimapCache<K, V> get(String name) {
```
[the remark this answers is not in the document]
karesti: why, I'm using createIfAbsent flag

multimap/src/main/java/org/infinispan/multimap/api/MultimapCacheManager.java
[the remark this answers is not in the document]
karesti: ok

karesti:
@tristantarrant I reported the changes you asked on documentation

tristantarrant:
Merged to master

PR #5201, opened 2017-06-14
Ispn 7924/upgrade kube ping

slaskawi:
https://issues.jboss.org/browse/ISPN-7924
https://issues.jboss.org/browse/ISPN-7928

This PR integrates the latest `KUBE_PING` into Infinispan. You may want to read [Bela's blog](http://belaban.blogspot.ch/2017/05/running-infinispan-cluster-with.html) to get the full spectrum of changes. The important from Infinispan perspective are:

* `KUBE_PING` has only 2 explicit dependencies now - JGroups and OAuth library.
* There is no need to use separate port for discovery (previously `KUBE_PING` used embedded HTTP for this).
* We don't need DMR (!!) dependency
* The new version of `KUBE_PING` uses only `KUBERNETES` prefixes for environmental variables. Previously it was possible to use either `KUBERNETES` or `OPENSHIFT`. This potentially impacts user apps deployed on *OpenShift* (apps deployed on Kubernetes won't get hurt). However it's just a matter of updating configuration.

### DIFF
--- a/all/embedded/pom.xml
+++ b/all/embedded/pom.xml
@@ -140,18 +140,8 @@
       </dependency>
 
       <dependency>
-         <groupId>org.jboss</groupId>
-         <artifactId>jboss-dmr</artifactId>
-         <optional>${uberjar.deps.optional}</optional>
-      </dependency>
-      <dependency>
          <groupId>org.jgroups.kubernetes</groupId>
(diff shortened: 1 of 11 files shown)


tristantarrant:
Merged, thanks

PR #5204, opened 2017-06-15
ISPN-7930 Avoid provided deps unless necessary

galderz:
https://issues.jboss.org/browse/ISPN-7930

* Some of the provided dependencies were added to avoid issues with uber
  jars, but these uber jars are not really designed for Maven-like envs.
* Having these provided dependencies forces users to define the provided
  dependencies on top which is a usability issue.
* So this JIRA reverts these provided dependencies since users should
  not be relying on uber jars as dependencies.


galderz:
Once happy I'll port it to 9.0.x

tristantarrant:
Merged to master and 9.0.x

PR #5205, opened 2017-06-15
ISPN-7710 Remote query server module dependency

galderz:
https://issues.jboss.org/browse/ISPN-7710

* Add remote query server dependency to org.infinispan module in server
  as optional dependency so that CompatibilityProtoStreamMarshaller can
  easily be set in configuration.


tristantarrant:
As a temporary fix I guess this will do. @anistor @gustavonalle will this be obsoleted by the transcoding work or should we consider moving this marshaller to a separate module ?

anistor:
I agree, good for now. Yes, it will be obsoleted by transcoding, and my experiment on deploying lucene analyzers and protobuf enitity marshallers to server, if proven successful. That approach can be applied to encoders too. I commented on the devlist, but you'll have to go back to the future to read my reply.

anistor:
Integrated in master and 9.0.x. thanks @galderz !

PR #5209, opened 2017-06-16
ISPN-7114: Datahash event no longer ignored.

ryanemerson:
A mistake when refactoring.


rvansa:
OK, looks really trivial.

PR #5211, opened 2017-06-16
Ispn 7900 Provide entrySet, values, keySet implementation for RemoteCache

wburns:
https://issues.jboss.org/browse/ISPN-7900


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
```diff
@@ -522,11 +527,156 @@ protected void set(K key, V value) {
...
+   public CloseableIteratorCollection<V> values() {
+      return new ValuesCollection();
+   }
+
+   private class KeySet extends AbstractCollection<K> implements CloseableIteratorSet<K> {
+
+      @Override
+      public CloseableIterator<K> iterator() {
+         return new RemovableCloseableIterator<>(
+               new CloseableIteratorMapper<>(retrieveEntries(null, batchSize), e -> (K) e.getKey()),
```
rvansa: Seems like you should use some filter-converter to drop the values before going over the wire.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
```diff
@@ -522,11 +527,156 @@ protected void set(K key, V value) {
...
+
+      @Override
+      public boolean remove(Object o) {
+         return RemoteCacheImpl.this.remove(o) != null;
+      }
+
+      @Override
+      public boolean removeAll(Collection<?> c) {
+         boolean removedSomething = false;
+         for (Object removeMe : c) {
```
rvansa: s/`removeMe`/`key`/ ?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
```diff
@@ -522,11 +527,156 @@ protected void set(K key, V value) {
...
+      }
+   }
+
+   private CloseableIterator<Entry<K, V>> castIterator(CloseableIterator iterator) {
+      return iterator;
+   }
+
+   private boolean removeEntry(Map.Entry<K, V> entry) {
+      K key = entry.getKey();
+      VersionedValue<V> versionedValue = RemoteCacheImpl.this.getVersioned(key);
```
rvansa: Would it be possible to retrieve the entries including the version, so you don't have to retrieve it second time?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
```diff
@@ -522,11 +527,156 @@ protected void set(K key, V value) {
...
+
+      @Override
+      public boolean contains(Object o) {
+         return RemoteCacheImpl.this.containsValue(o);
+      }
+
+      @Override
+      public boolean remove(Object o) {
+         Objects.requireNonNull(o);
+         try (CloseableIterator<V> iter = iterator()) {
```
rvansa: Do you fancy O(n<sup>2</sup>) traversal with removal? :) I was about to suggest filter-converter dropping the keys before I've realized that you need the keys for effective removals.

documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
```diff
@@ -26,6 +26,22 @@ There are no changes needed for `AtomicMap`, but it now supports non-transaction
 
 `FineGrainedAtomicMap` now uses the link:../user_guide/user_guide.html#grouping[Grouping API] and therefore you need to link:../user_guide/user_guide.html#how_do_i_use_the_grouping_api[enable groups in configuration]. Also it holds entries as regular cache entries, plus one cache entry for cached key set (the map itself). Therefore the cache size or iteration/streaming results may differ. Note that fine grained atomic maps are still supported on transactional caches only.
 
+=== RemoteCache keySet/entrySet/values
+
+RemoteCache now implements all of the collection backed methods from `Map` interface.
+Previously `keySet` was implemented, however it was a deep copy. This has now changed
+and it is a backing sett. That is that the set retrieves the updated values on each
```
rvansa: typo

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod.adoc
```diff
@@ -104,19 +105,67 @@ assert !cache.containsKey("car") : "Value must have been removed!";
...
+requiring them all be stored in the client memory at once if the user does not want.
+These collections adhere to the `Map` specification being that `add` and `addAll`
+are not supported but all other methods are supported.
+
+One thing to note is the `Iterator.remove` and `Set.remove` or `Collection.remove`
+methods require more than 1 round trip to the server to operate. You can check
+out the link:{javadocroot}/org/infinispan/client/hotrod/RemoteCache.html[RemoteCache]
+Javadoc to see more details about these and the other methods.
+
+*Iterator Uage*
```
rvansa: typo

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod.adoc
```diff
@@ -104,19 +105,67 @@ assert !cache.containsKey("car") : "Value must have been removed!";
...
+
+The iterator method of these collections uses `retrieveEntries` internally, which
+is described below. If you nottice `retrieveEntries` takes an argument for the batch
```
rvansa: typo

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod.adoc
```diff
@@ -104,19 +105,67 @@ assert !cache.containsKey("car") : "Value must have been removed!";
...
+size. There is no way to provide this to the iterator. As such the batch size can be
+configured via system property `infinispan.client.hotrod.batch_size` or through
+the link:{javadocroot}/org/infinispan/client/hotrod/configuraion/ConfigurationBuilder.html#batchSize-int-[ConfigurationBuilder]
```
rvansa: imo configuration should be the preferred place; why do we need the system property at all? JCache? Is there a way to add proprietary configuration when using JCache?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
wburns: `Remove` is fine, I am guessing you mean for `removeAll`? That was just an oversight, I meant to implement that.

wburns: And actually looking closer I have to implement all of the `iterator` methods to make sure the iterator is closed. Upon closer inspection it looks like draining the entire iterator doesn't close it for remote. @gustavonalle is that correct?

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod.adoc
wburns: I am not sure why as well? @galderz might know? I just added both as we have for other things.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
wburns: Let me take another look, I just assumed from the API (contains Map.Entry) it wasn't, but I can debug and see the underlying impl. And if not we could include it as a converter.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
gustavonalle: yes

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
wburns: Shouldn't really matter, but sure.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
wburns: Yes I thought about that, however I wasn't sure if we wanted to register a filter/converter automatically. I figured at worse case it could always be added back in later if needed. I was hoping to get a little discussion here about that anyways.

gustavonalle: @wburns the filter and converter could be available on the server on start, and the ```retrieveEntries``` could just reference either one (keys only or values only) when needed.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
rvansa: Uuh, my bad. I've mistaken that for the iterator.remove() operation. Sorry about that.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
rvansa: If the entries retrieved through `retrieveEntries` don't contain that info (that would be no surprise), the `Converter` on server side has access to the `Metadata`, so it could add those to the returned entries.

wburns: Yep my thought exactly!

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
wburns: Yeah I get confused looking back and forth with these all the time too, haha.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
wburns: I will look into adding this then, guessing it shouldn't be too bad.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
gustavonalle: ```retrieveEntriesWithMetadata``` gives you the versions

wburns: Ah yes overlooked that, thanks!

wburns:
Updated per comments, still looking into filter/converter though as this was only thing not done.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
wburns: Sadly you can only supply a `KeyValueFilterConverter` which doesn't allow for null values. So I guess I will have it just return an empty byte[]. I will add a converter called `to_empty_bytes`. Seem alright to you guys?

wburns:
Added in key value filter converter as well for keySet.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
rvansa: I would register it under exactly the fully qualified class name, e.g. `org.infinispan.filter.ToEmptyBytes`

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
```diff
@@ -138,6 +146,34 @@ public HotRodDecoder getDecoder() {
       return new HotRodDecoder(cacheManager, transport, this, this::isCacheIgnored);
    }
 
+   /**
+    * Class used to allow for remote clients to essentially ignore the value by returning an empty byte[].
+    */
+   @SerializeWith(value=ToBytesKeyValueFilterConverter.ToBytesKeyValueFilterConverterExternalizer.class)
+   static class ToBytesKeyValueFilterConverter extends AbstractKeyValueFilterConverter {
```
rvansa: I thought you wanted to use To**Empty**Bytes...

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
```diff
@@ -348,6 +349,13 @@ public boolean containsKey(Object key) {
    }
 
    @Override
+   public boolean containsValue(Object value) {
+      Objects.requireNonNull(value);
+      // TODO: This would be more efficient if done on the server itself and requested via protocol
```
galderz: Yeah, could be done with a pre-registered remote server task.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
```diff
@@ -522,11 +530,212 @@ protected void set(K key, V value) {
...
+      }
+
+      @Override
+      public void clear() {
+         RemoteCacheImpl.this.clear();
+      }
+
+      @Override
+      public boolean contains(Object o) {
+         // Have to close the stream, just in case stream terminates early
```
galderz: This is another one that could be more efficient if implemented as a remote server task?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
wburns: Sure.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
wburns: Yeah I kept changing my mind and didn't make them the same :)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
wburns: Yes, actually this and below are essentially the same thing. I thought I had refactored this one to call the other, let me do that and I will just put the TODO on the bottom one.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
wburns: Yes it is the same thing :)

wburns:
Fixing some test failures as well, will have to wait for run to verify they pass after I update PR.

wburns:
Okay fixed test failures and recent comments. This has to wait for a CI run to verify no other failures though.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
```diff
@@ -522,11 +529,216 @@ protected void set(K key, V value) {
...
+            V value = RemoteCacheImpl.this.get(entry.getKey());
+            return value != null && value.equals(entry.getValue());
+         }
+         return false;
+      }
+
+      @Override
+      public boolean remove(Object o) {
+         Map.Entry<K, V> entry = toEntry(o);
+         return entry != null && RemoteCacheImpl.this.removeEntry(entry);
```
rvansa: `RemoteCacheSupport` contains `remove(K key, V value)` which throws `UnsupportedOperationException`. I think either this method should throw that, too (and then `removeEntry` could be deleted) or plain `RemoteCache.remove(K, V)` should be supported, too.

wburns: I had originally implemented `remove(K key, V value)` to do what my added `removeEntry` does. The problem is if I added that method I feel like I also would have to add all the various `replace` methods which I wasn't looking forward to.

WDYT?

rvansa: In that case just don't support removal on the collection. Though, showing the 'correct' `replace` wouldn't hurt too much, especially if you include a javadoc comment that users should prefer getWithMetadata + replaceWithVersion.

wburns: I will add replace, it shouldn't be too bad.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
```diff
@@ -522,11 +529,216 @@ protected void set(K key, V value) {
...
+
+   private class KeySet extends AbstractCollection<K> implements CloseableIteratorSet<K> {
+
+      @Override
+      public CloseableIterator<K> iterator() {
+         return new RemovableCloseableIterator<>(
+               // Use the to_empty_bytes filter converter to reduce value payload
+               new CloseableIteratorMapper<>(retrieveEntries(
+                     "org.infinispan.server.hotrod.HotRodServer$ToEmptyBytesKeyValueFilterConverter",
+//                     null,
```
wburns: Fixed

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
```diff
@@ -522,11 +529,216 @@ protected void set(K key, V value) {
...
+   public CloseableIteratorCollection<V> values() {
+      return new ValuesCollection();
+   }
+
+   private class KeySet extends AbstractCollection<K> implements CloseableIteratorSet<K> {
+
+      @Override
+      public CloseableIterator<K> iterator() {
+         return new RemovableCloseableIterator<>(
+               // Use the to_empty_bytes filter converter to reduce value payload
```
wburns: Fixed

wburns:
Addressed comments, also added in support for `replace` and `remove` overridden methods on RemoteCache.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCache.java
```diff
@@ -80,6 +83,55 @@
...
+    */
+   boolean remove(Object key, Object value);
+
+   /**
+    * {@inheritDoc}
+    * <p>
+    * This method requires 2 round trips to the server. The first to retrieve the value and version and a second to
+    * replace the key with the version if the value matches. If possible user should use
+    * {@link RemoteCache#getWithMetadata(Object)} and
+    * {@link RemoteCache#replaceWithVersion(Object, Object, long)} if possible.
```
tristantarrant: cacophony alert: if possilbe .... if possible

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCache.java
```diff
@@ -80,6 +83,55 @@
...
+   @Override
+   boolean replace(K key, V oldValue, V newValue);
+
+   /**
+    * {@inheritDoc}
+    * <p>
+    * This method requires 2 round trips to the server. The first to retrieve the value and version and a second to
+    * replace the key with the version if the value matches. If possible user should use
+    * {@link RemoteCache#getWithMetadata(Object)} and
+    * {@link RemoteCache#replaceWithVersion(Object, Object, long, long, TimeUnit, long, TimeUnit)} if possible.
```
tristantarrant: cacophony alert: if possilbe .... if possible

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCache.java
```diff
@@ -80,6 +83,55 @@
...
+   @Override
+   boolean replace(K key, V oldValue, V value, long lifespan, TimeUnit unit);
+
+   /**
+    * {@inheritDoc}
+    * <p>
+    * This method requires 2 round trips to the server. The first to retrieve the value and version and a second to
+    * replace the key with the version if the value matches. If possible user should use
+    * {@link RemoteCache#getWithMetadata(Object)} and
+    * {@link RemoteCache#replaceWithVersion(Object, Object, long, long, TimeUnit, long, TimeUnit)} if possible.
```
tristantarrant: cacophony alert: if possilbe .... if possible

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCache.java
wburns: Fixing

tristantarrant:
This needs documentation bits:
* update the "basic api" paragraph related to bulk ops in the Java Hot Rod client chapter
* update the upgrading guide to indicate that these bulk ops no longer fetch everything

wburns:
@tristantarrant these should be covered at
https://github.com/infinispan/infinispan/pull/5211/files#diff-dd93d1e8f5a73963f966b6e41981bbb8R29
https://github.com/infinispan/infinispan/pull/5211/files#diff-5d6e8f3e7a67d3b6c70c9ebc0dc8912eR108

Did you want more detail on anything there?

tristantarrant:
Age is a terrible thing, I completely missed that. LGTM

alanfx:
Age beats the alternative!

gustavonalle:
@wburns WDYT of an overload keySet/EntrySet/values that takes a KVfilterConverter string?

wburns:
@gustavonalle I think it could be interesting, however it might cause more confusion. This would only affect `iterator`, `spliterator` and `stream`, other methods would be unaffected. Also we have to return a `Collection<Object>` etc. since we can't rely on the generics any longer.

If we provided this, we should also provide a way to override the `batchSize` as well I would think.

But either way I am not so sure about that addition in this PR. Maybe we can think about this further on dev list and create another issue?

gustavonalle:
@wburns yes, I wasn't thinking about adding it to this PR

tristantarrant:
Merged to master. Thanks @wburns

PR #5220, opened 2017-06-20
ISPN-7952 OffHeapSingleNodeExpirationEvictionTest failures

wburns:
* Fixed offset issue

https://issues.jboss.org/browse/ISPN-7952

### DIFF
--- a/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
@@ -347,7 +347,7 @@ public class OffHeapDataContainer implements DataContainer<WrappedBytes, Wrapped
 
    @Override
    public int size() {
-      long time = timeService.time();
+      long time = timeService.wallClockTime();
       long count = entryStream().filter(e -> !e.isExpired(time)).count();
       if (count > Integer.MAX_VALUE) {
          return Integer.MAX_VALUE;
@@ -587,7 +587,7 @@ public class OffHeapDataContainer implements DataContainer<WrappedBytes, Wrapped
 
    @Override
    public Iterator<InternalCacheEntry<WrappedBytes, WrappedBytes>> iterator() {
-      long time = timeService.time();
+      long time = timeService.wallClockTime();
       return entryStream().filter(e -> !e.isExpired(time)).iterator();
    }
 
--- a/core/src/test/java/org/infinispan/container/offheap/OffHeapSingleNodeExpirationEvictionTest.java
+++ b/core/src/test/java/org/infinispan/container/offheap/OffHeapSingleNodeExpirationEvictionTest.java
@@ -1,8 +1,16 @@
 package org.infinispan.container.offheap;
 
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertNotNull;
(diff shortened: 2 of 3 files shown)


wburns:
This can also be cherry picked easily in 9.0.x

gustavonalle:
@wburns test still failing

core/src/test/java/org/infinispan/container/offheap/OffHeapSingleNodeExpirationEvictionTest.java
```diff
@@ -64,4 +69,41 @@ protected EmbeddedCacheManager addClusterEnabledCacheManager(ConfigurationBuilde
...
+            assertEquals(storedTime, entry.getLifespan());
+            assertEquals(storedTime, entry.getMaxIdle());
+//            assertBetweenTimes(beforeInsert, entry.getCreated(), afterInsert);
+//            assertBetweenTimes(beforeInsert, entry.getLastUsed(), afterInsert);
+            break;
+      }
+   }
+
+   // TODO: this is commented out because currently you can't retrieve the created or last used information from the cache
+   // See https://issues.jboss.org/browse/ISPN-7954 if this has been fixed.
```
danberindei: Couldn't you use `DataContainer.get()` instead of `Cache.get()`?

wburns: I thought of doing this instead, but I had I guessed incorrectly I had fixed the issue (I will change it over to verify for sure these are written properly). I really wish I could reproduce this issue on my box :(

wburns:
Okay my changes were correct, unfortunately it was the wrong bug!

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -587,7 +587,7 @@ public void executeTask(KeyValueFilter<? super WrappedBytes, ? super WrappedByte
 
    @Override
    public Iterator<InternalCacheEntry<WrappedBytes, WrappedBytes>> iterator() {
-      long time = timeService.time();
+      long time = timeService.wallClockTime();
```
wburns: This was the actual bug btw. So depending on the current system time it would show itself or not. Hence why I was never able to see the issue.

core/src/test/java/org/infinispan/container/offheap/OffHeapSingleNodeExpirationEvictionTest.java
wburns: Updated to make sure metadata for created/lastUsed is updated properly.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
danberindei: @wburns Have you ever tried running the entire core test suite with trace logging enabled? That does it for me almost every time.

wburns: Not this test :) This test is all because of the system clock rolling over or not. I am able to reproduce it now, because my system clock has rolled over.

danberindei:
Integrated, thanks Will!

PR #5224, opened 2017-06-21
ISPN-7884 Test fix & rename

rvansa:


danberindei:
Integrated, thanks Radim!

PR #5226, opened 2017-06-22
BZ-1373412 - [GSS](6.4.z) TimeoutException when request ends up on …

JiriOndrusek:
…non-coordinator node (invalidation cache, web session passivation and shared cache store)

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1373412
No upstream required

### DIFF
--- a/core/src/test/java/org/infinispan/interceptors/InvalidationInterceptorMockTest.java
+++ b/core/src/test/java/org/infinispan/interceptors/InvalidationInterceptorMockTest.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2011 Red Hat, Inc. and/or its affiliates.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301 USA
+ */
+
+package org.infinispan.interceptors;
+
+import org.infinispan.commands.CommandsFactory;
+import org.infinispan.commands.tx.PrepareCommand;
+import org.infinispan.commands.write.InvalidateCommand;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -85,21 +85,25 @@ public void init(CommandsFactory factory) {
...
             lockKeyAndCheckOwnership(ctx, command.getKey(), lockTimeout, skipLocking);
          }
          return invokeNextInterceptor(ctx, command);
       } catch (Throwable t) {
+         releaseRemote = false;
          releaseLocksOnFailureBeforePrepare(ctx);
          throw t;
       } finally {
          if (!ctx.isInTxScope()) lockManager.unlockAll(ctx);
+         if(releaseRemote) releaseRemoteIfNeeded(ctx, command, cdl.localNodeIsPrimaryOwner(command.getKey()));
```
danberindei: This seems wrong, the locks should only be released on commit/rollback. If we release the lock immediately after reading it, then we might as well not lock it at all, like we do when the `FORCE_WRITE_LOCK` flag isn't used.

tristantarrant: @JiriOndrusek can you address @danberindei 's comment ?

JiriOndrusek: @danberindei I see my mistake, I'll look for the correct solution.

JiriOndrusek:
New fix is prepared.
@danberindei Could you look into it?

core/src/test/java/org/infinispan/interceptors/InvalidationInterceptorMockTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+        RpcManager rpcManager = mock(RpcManager.class);
+        when(rpcManager.getTransport()).thenReturn(mock(Transport.class));
+
+        invalidationInterceptor.inject(rpcManager);
+        invalidationInterceptor.injectDependencies(commandsFactory);
+
+        invalidationInterceptor.invalidateAcrossCluster(true, null, ctx);
+
+        if(inTransaction) {
```
danberindei: Please add a space after `if`.

danberindei:
@JiriOndrusek You're missing the `EntryWrappingVisitor.visitInvalidateCommand()` addition from https://github.com/infinispan/infinispan/pull/1726/files. This causes `SyncInvalidationTest` to fail, because the `InvalidationCommand` wrapped in a `PrepareCommand` is not processed and the value is not removed.

Also, please remove the `[]` around `BZ-1373412` from the commit message.

JiriOndrusek:
@danberindei all changes you suggested are commited into PR

danberindei:
Integrated, thanks Jiri!

PR #5213, opened 2017-06-19
ISPN-7910 Add serializable overloads to the functional API

rvansa:
https://issues.jboss.org/browse/ISPN-7910

PR #5228, opened 2017-06-22
ISPN-8072 Convenience methods in ReadWriteEntryView

rvansa:
* Add some convenience default methods to ReadWriteEntryView

### DIFF
--- a/core/src/main/java/org/infinispan/functional/EntryView.java
+++ b/core/src/main/java/org/infinispan/functional/EntryView.java
@@ -4,6 +4,7 @@ import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 import org.infinispan.commons.util.Experimental;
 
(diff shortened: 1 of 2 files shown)


rvansa:
@danberindei I haven't noticed that #5213 was already merged and adapted the docs as you've suggested (you seem to come commenting to PRs when these are already merged). Trying to use `ReadWriteMap` in the example I've noticed that current API does not allow updating the value through a one-liner, so I've added couple of convenience methods.
Could move that under another JIRA, though.

core/src/main/java/org/infinispan/functional/EntryView.java
```diff
@@ -107,6 +108,88 @@ private EntryView() {
...
+       * Sets this value to the result of <code>updateFunction</code> applied
+       * to the current value, or <code>null</code> if absent.
+       * If <code>updateFunction</code> returns <code>null</code>, the mapping
+       * is removed.
+       *
+       * @param updateFunction
+       * @return <code>null</code>
+       * @since 9.1
+       */
+      default Void update(Function<? super V, ? extends V> updateFunction) {
```
tristantarrant: I don't understand how $SUBJECT applies to the changes. Should you be using SerializableFunction here ?

danberindei:
@rvansa Before a crash wiped out my Firefox session yesterday, I still had a tab open with the scattered cache PR open ;)

I don't see the need for the new helper methods, either, as it should be quite easy to achieve the same result by using the `Optional` methods on the result of `find()`.

rvansa:
Something went wrong with this PR, as the serializable overloads went already in, and the documentation seems already malformed... My motivation was that I could not write a simple `value -> value + 1` update inliner in the docs, so I thought that it could come handy to users as well. And I wanted to piggyback it to the serialization changes/docs.

@danberindei I find `view -> view.updateIfPresent(value -> value + 1)` more readable than `view -> view.find().ifPresent(value -> view.set(value + 1))`
@tristantarrant The function applied to the view is not serialized, so we don't need the overloads.

In any way I should reword the PR & file a new JIRA for that.

rvansa:
Created another JIRA for this.

danberindei:
IMO there's a pretty good reason you can't do a simple `value -> value + 1`, and that's because the value can be missing, and you'll want to initialize with `1` instead of ignoring that case like `updateIfPresent` does.

Another reason not to add these methods is that they use `null` for missing/removed values, but `find()` uses `Optional.empty()`, and I don't like mixing the two approaches in the same interface.

core/src/main/java/org/infinispan/functional/EntryView.java
```diff
@@ -118,6 +119,88 @@ private EntryView() {
...
+
+      /**
+       * Set this value to value provided by the <code>supplier</code> if there
+       * is no current mapping.
+       *
+       * @param supplier
+       * @return <code>null</code>
+       * @since 9.1
+       */
+      default Void setIfAbsent(Supplier<? extends V> supplier) {
```
danberindei: `set` takes an immediate value, so `setIfAbsent` should do the same (not that I think a method like that would be very useful).

rvansa:
OK, let's close this, then.

PR #5230, opened 2017-06-23
ISPN-7979 Add licenses to distributions

tristantarrant:
This is a blocker for release (required by legal).
The format of the generated license descriptor and table has also been set by legal.

This is what it looks like: http://www.dataforte.net/listing/infinispan/licenses/licenses.html

### DIFF
--- a/distribution/src/main/resources/licenses/antlr-bsd.txt
+++ b/distribution/src/main/resources/licenses/antlr-bsd.txt
@@ -1,0 +1,11 @@
+Copyright (c) 2012 Terence Parr and Sam Harwell
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
+    Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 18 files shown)


distribution/src/main/resources/licenses/antlr-bsd.txt
```diff
@@ -0,0 +1,11 @@
+Copyright (c) 2012 Terence Parr and Sam Harwell
```
slaskawi: Just a thought - perhaps we could download this file during the build? Otherwise we'd need to maintain all header information (copyright year etc).

tristantarrant:
Unfortunately we cannot guarantee that licenses will stay in the same
location...

wburns:
@tristantarrant this is all I need to do re: licenses for community? Does @vblagoje have to add something to the console then?

tristantarrant:
For infinispan we need to add the licenses to server as well as the console
licenses. Vittorio and Galder need to add all licenses to the native
clients.

wburns:
Integrating, I will take care of server in another PR.

PR #5231, opened 2017-06-26
ISPN-7975 NullPointerException in PutMapHelper.transformResult

rvansa:
* add AdvancedCache.getAndPutAll method
* includes fix in MultiSubCommandInvoker (affects L1)
* includes fix of UnsureResponse handling for GetAllCommand

https://issues.jboss.org/browse/ISPN-7975

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/impl/MultiSubCommandInvoker.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/MultiSubCommandInvoker.java
@@ -46,7 +46,7 @@ public class MultiSubCommandInvoker implements InvocationSuccessFunction {
    public Object apply(InvocationContext rCtx, VisitableCommand rCommand, Object rv) throws Throwable {
       if (subCommands.hasNext()) {
          VisitableCommand newCommand = subCommands.next();
-         return interceptor.invokeNext(rCtx, newCommand);
+         return interceptor.invokeNextThenApply(rCtx, newCommand, this);
       } else {
          return finalStage;
       }
--- a/core/src/test/java/org/infinispan/statetransfer/ReadAfterLostDataTest.java
+++ b/core/src/test/java/org/infinispan/statetransfer/ReadAfterLostDataTest.java
@@ -15,8 +15,6 @@ import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
 import org.infinispan.Cache;
-import org.infinispan.commands.CommandsFactory;
-import org.infinispan.commands.write.PutMapCommand;
 import org.infinispan.functional.EntryView.ReadEntryView;
 import org.infinispan.functional.EntryView.ReadWriteEntryView;
 import org.infinispan.functional.FunctionalMap.ReadOnlyMap;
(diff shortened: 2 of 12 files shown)


core/src/main/java/org/infinispan/interceptors/impl/MultiSubCommandInvoker.java
```diff
@@ -46,7 +46,7 @@ public static Object invokeEach(InvocationContext ctx, Iterator<VisitableCommand
    public Object apply(InvocationContext rCtx, VisitableCommand rCommand, Object rv) throws Throwable {
       if (subCommands.hasNext()) {
          VisitableCommand newCommand = subCommands.next();
-         return interceptor.invokeNext(rCtx, newCommand);
+         return interceptor.invokeNextThenApply(rCtx, newCommand, this);
```
rvansa: Np, that's just an oversight :)

core/src/test/java/org/infinispan/statetransfer/ReadAfterLostDataTest.java
```diff
@@ -217,11 +213,7 @@ private void invokeOperation(Cache<Object, Object> cache, BiFunction<Cache<Objec
       for (Object key : keys) {
          writeMap.put(key, "other" + (i++));
       }
-      ComponentRegistry cr = cache.getAdvancedCache().getComponentRegistry();
-      InvocationContext ctx = cr.getComponent(InvocationContextFactory.class).createInvocationContext(true, keys.size());
-      PutMapCommand putMapCommand = cr.getComponent(CommandsFactory.class).buildPutMapCommand(writeMap, null, 0);
-      ctx.setLockOwner(putMapCommand.getKeyLockOwner());
-      return (Map<?, ?>) cache.getAdvancedCache().getAsyncInterceptorChain().invoke(ctx, putMapCommand);
+      return cache.getAdvancedCache().getAndPutAll(writeMap);
```
danberindei: So you already had a test that used a `PutMapCommand` without `IGNORE_RETURN_VALUES`... do you know why it didn't fail with the NPE?

rvansa:
The failure in `ClusteredDistCacheTest#testPutForExternalRead` should not be related, seems like related to https://issues.jboss.org/browse/ISPN-3709

core/src/main/java/org/infinispan/interceptors/impl/MultiSubCommandInvoker.java
danberindei: True, it didn't break any tests :D

core/src/test/java/org/infinispan/statetransfer/ReadAfterLostDataTest.java
rvansa: Because all the keys were set to `"valueX"` - the NPE is thrown only when the keys are not found in the cache.

rvansa: That's why the added test uses two local and two remote keys, one of which is always without mapping.

gustavonalle:
LGTM

danberindei:
Ok, integrating...

danberindei:
Done, thanks Radim!

PR #5232, opened 2017-06-26
ISPN-7961 Cross-site replication of functional commands

rvansa:
* Fix handling of ComputeCommand and ComputeIfAbsentCommand
* In transaction replicate values from transactional context (replaying
  modifications in remote site may have different results)

https://issues.jboss.org/browse/ISPN-7961


pruivo:
@rvansa needs rebase

core/src/main/java/org/infinispan/commands/functional/WriteOnlyKeyCommand.java
```diff
@@ -8,6 +8,7 @@
 import org.infinispan.commands.CommandInvocationId;
 import org.infinispan.commands.Visitor;
 import org.infinispan.commands.write.ValueMatcher;
+import org.infinispan.functional.EntryView;
```
pruivo: unused import

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -358,44 +360,55 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
          }
          return invokeNext(ctx, command);
       }
+      GetAllSuccessHandler getAllSuccessHandler = new GetAllSuccessHandler(command);
```
rvansa: Yes, I need a way to fetch multiple entries (`remoteGetAll`) from `TxDistributionInterceptor`, because when xsite is on, we will need the final value of the entry to replicate xsite.

pruivo: I'm not convinced. the read commands aren't sent to the backup site.

rvansa: You got me wrong; When you execute functional modification in a transaction, by default you don't fetch the whole entry into originator's context, just the version for WSC. However, currently the xsite is implemented in such a way that the whole transaction is replicated xsite by the originator. And since the xsite may not have the previous value, we can't just replay the func modification in there - we need to send the whole updated value. Therefore, if xsite is on, we need to fetch the modified entry in order to have the updated value on the originator.
Since some commands work on multiple entries, I need some kind of `remoteGetAll` to get hold of those values. Note that at some places the commands use just multiple `remoteGet`s but that's a bug, in fact https://issues.jboss.org/browse/ISPN-7889

core/src/main/java/org/infinispan/xsite/BackupSenderImpl.java
```diff
@@ -286,54 +297,53 @@ private boolean isOffline(String site) {
...
          return Collections.emptyList();
       }
       List<WriteCommand> filtered = new ArrayList<>(modifications.size());
+      // Note: the result of replication of transaction with flagged operations may be actually different.
+      // We use last-flag-bit-set wins strategy.
+      // All we can do is to assume that if the user plays with unsafe flags he won't modify the entry once
+      // in a replicable and another time in a non-replicable way
+      //TODO: what if after functional command the modified entry is not in the context?
+      Map<Object, WriteCommand> lastModifyingCommand = new HashMap<>();
       for (WriteCommand writeCommand : modifications) {
```
rvansa: Reverse iteration is much better, done.

core/src/main/java/org/infinispan/xsite/BaseBackupReceiver.java
```diff
@@ -83,79 +83,59 @@ protected final void assertAllowInvocation() {
...
       }
 
       @Override
-      public Object visitPutMapCommand(InvocationContext ctx, PutMapCommand command) throws Throwable {
-         Metadata metadata = command.getMetadata();
-         backupCache.putAll(command.getMap(),
-                            metadata.lifespan(), TimeUnit.MILLISECONDS,
-                            metadata.maxIdle(), TimeUnit.MILLISECONDS);
+      public Object visitClearCommand(InvocationContext ctx, ClearCommand command) throws Throwable {
+         backupCache.clear();
```
danberindei: Missed another TODO here ;)

core/src/test/java/org/infinispan/xsite/AbstractTwoSitesTest.java
```diff
@@ -107,6 +111,11 @@ protected void adaptLONConfiguration(BackupConfigurationBuilder builder) {
       //no-op
    }
 
+   public AbstractTwoSitesTest use2Pc(boolean use2Pc) {
+      this.use2Pc = use2Pc;
+      return this;
+   }
```
danberindei: This should be grouped with the other parameters (and with `parameters()` as well).

core/src/test/java/org/infinispan/xsite/backupfailure/NonTxBackupFailureTest.java
```diff
@@ -136,9 +137,17 @@ public void testPutMapFailure() {
...
 
+      assertTrue(failureInterceptor.writeOnlyManyEntriesFailed);
       for (int i = 0; i < 100; i++) {
          final int keyIndex = i;
-         eventuallyEquals("v" + keyIndex, () -> cache("LON", keyIndex % 2).get("k" + keyIndex));
+         // When the policy is set to fail, the failure may fail local cluster operation and the value
+         // is not written. This happens when the failure is thrown on remote primary owner - we don't
+         // commit local entries until distribution interceptor returns and this now throws an exception.
+         // This used to work when we were replicating cross-site from origin only after everything was
+         // committed - the replication failure then did not affect local cluster state.
```
danberindei: Isn't that the case now? `BaseBackupInterceptor` is still before `EntryWrappingInterceptor`...

core/src/test/java/org/infinispan/xsite/statetransfer/failures/SiteConsumerTopologyChangeTest.java
```diff
@@ -55,6 +55,7 @@ public void testXSiteDuringJoin() throws InterruptedException, ExecutionExceptio
       doXSiteStateTransferDuringTopologyChange(TopologyEvent.JOIN);
    }
 
+   // TODO: this test is flaky - ISPN-6872
```
danberindei: Perhaps you should move it to the "unstable_xsite" group.

core/src/test/java/org/infinispan/xsite/CacheOperationsTest.java
```diff
@@ -0,0 +1,237 @@
...
+      for (Object key : keys) {
+         assertEquals(null, backup.get(key));
+      }
+   }
+
+   private static Map<Object, Object> map(Object[] keys, String value) {
+      return Stream.of(keys).collect(Collectors.toMap(Function.identity(), ignored -> value));
+   }
+}
+
```
danberindei: checkstyle error

core/src/main/java/org/infinispan/functional/impl/MetaParamsInternalMetadata.java
```diff
@@ -6,6 +6,7 @@
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.StreamSupport;
```
pruivo: unused import

core/src/main/java/org/infinispan/functional/Param.java
```diff
@@ -126,14 +126,20 @@ public static LockingMode valueOf(int ordinal) {
...
        * is not an owner of given entry, the entry is not stored in the cache. If the command reads a value and
        * the entry is not available locally, null entry is provided instead.
        */
-      LOCAL;
+      LOCAL,
+      /**
+       * Command is executed only in the current site (same as {@link #CLUSTER}, but it is not sent for backup
+       * to other sites)
+       */
+      SITE_ONLY;
```
rvansa: I'll rename `CLUSTER` to `ALL`, we can mess with this a bit due to the `@Experimental` flag. It's not used anywhere, as it's the default anyway. And `LOCAL_SITE` sounds good.

I'll add some backdoor to the documentation of `CLUSTER` :)

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -358,44 +360,55 @@ public Object visitGetAllCommand(InvocationContext ctx, GetAllCommand command) t
...
       }
-      return asyncValue(allFuture).thenApply(ctx, command, allFuture);
+      return allFuture;
    }
 
    protected void handleRemotelyRetrievedKeys(InvocationContext ctx, List<?> remoteKeys) {
    }
 
-   private class ClusteredGetAllHandler implements BiConsumer<Map<Address, Response>, Throwable> {
+   private class ClusteredGetAllHandler<C extends FlagAffectedCommand & TopologyAffectedCommand> implements BiConsumer<Map<Address, Response>, Throwable> {
```
rvansa: Maybe, but I wouldn't add changes in command hierarchy into this PR.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -456,17 +469,15 @@ private void handleMissingResponse(Response response) {
...
             synchronized (allFuture) {
-               Set<?> strippedKeys = new HashSet<>(allFuture.localCommand.getKeys());
-               strippedKeys.removeAll(keys);
-               // We can't just call command.setKeys() - interceptors might compare keys and actual result set
-               allFuture.localCommand = cf.buildGetAllCommand(strippedKeys, command.getFlagsBitSet(), command.isReturnEntries());
-               allFuture.lostData = true;
+               try {
+                  remoteGetAllHandler.onKeysLost(keys);
+               } catch (Throwable t) {
+                  allFuture.completeExceptionally(t);
```
rvansa: `completeExceptionally` must be called within a synchronized block, see `ClusteredGetAllFuture` javadoc. I could add another `allFuture.isDone()` check before the sync block on line 430, but we need to sync the responses.

core/src/test/java/org/infinispan/xsite/backupfailure/NonTxBackupFailureTest.java
rvansa: In the past:
1. origin -> primary
2. primary commit
3. origin commit
4. origin backup -> exception, but data are committed locally

now:
1. origin -> primary
2. primary commit
3. primary backups -> exception
4. origin gets exception, does not commit

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: Calling `visitGetAllCommand` would also mean that it would call the interceptors below, executing the command and then checking if it loaded all entries or throwing OTE (as write commands should). I think that the error handling was the main reason why I went for the refactoring.

rvansa:
Fixed or commented.

core/src/main/java/org/infinispan/functional/impl/MetaParams.java
```diff
@@ -288,6 +290,25 @@ void merge(MetaParams other) {
       return new It();
    }
 
+   @Override
+   public Spliterator<MetaParam<?>> spliterator() {
+      return Spliterators.spliterator(iterator(), length, Spliterator.DISTINCT | Spliterator.NONNULL);
+   }
+
+   public MetaParam.Writable[] toWritableMetas() {
+      int writable = 0;
```
rvansa: That probably creates another intermediate array, and this is on the hot path...

rvansa: Btw., I think that @karesti will introduce Metadata setter to the func API, and this will be another use case for that since the double conversion is unfortunate.

karesti: merge PR should be merged for this to happen, I guess

rvansa: Yes, but none should block each other. The current solution is fine, if we decide to not include the setter.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: I'm not sure I understand the javadoc of `ClusteredGetAllFuture` either... 

Completing `allFuture` will run some interceptor callbacks synchronously, but not all of them. E.g. the state transfer interceptor callback would suspend the invocation to wait for a new topology, and that would release the `allFuture` monitor, but the new topology could arrive while `remoteGetAllHandler.onKeysLost(keys)` is running, and you'd still have 2 threads accessing the same invocation context.

So I don't think synchronizing `completeExceptionally()` helps, we need to cancel all the `BaseDistributionInterceptor`-related callbacks before completing `allFuture` instead.

rvansa: When `completeExceptionally` executes before the other responses processing, the future will be marked as done and as soon as the other responses get into the synchronized block, these will check `isDone` and return.
If the response is being processed in sync block, running `completeExceptionally` and the related callbacks will be blocked until we finish the processing.

danberindei: Sure, having the handlers check `isDone()` works. But the `ClusteredGetAllFuture` javadoc only mentions synchronizing around `completeExceptionally()`, and that's not enough.

And yes, please add an `isDone()` check before entering the `synchronized` block.

rvansa:
Accidentally merged myself :-/

PR #5233, opened 2017-06-26
ISPN-6245 Integration test and fix for deploying custom compat marshaller

anistor:
Besides using a FQN to reference classes for compat marshaller and indexed entities (in server) we also accept an extended class name, composed of moduleName:slotName:fullyQualifiedClassName. See the updated docs for more info.

https://issues.jboss.org/browse/ISPN-6245
https://issues.jboss.org/browse/ISPN-6334

### DIFF
--- a/documentation/src/main/asciidoc/server_guide/configuration.adoc
+++ b/documentation/src/main/asciidoc/server_guide/configuration.adoc
@@ -499,48 +499,44 @@ By default each protocol stores data in the cache in the most efficient format f
 
 ----
 
-To specify a custom server-side compatibility marshaller use the "marshaller" attribute:
+To specify a custom server-side compatibility marshaller use the "marshaller" attribute which can hold either an
+ordinary fully qualified class name or an 'extended' class name which is composed of a JBoss Modules module identifier,
+a slot name, and a fully qualified class name separated by the ':' character, (eg. "com.acme.my-module-name:my-slot:com.acme.CustomMarshaller").
+
+By using a plain fully qualified class name you will be able to reference only 'stock' marshallers provided by Infinispan,
+like "org.infinispan.query.remote.CompatibilityProtoStreamMarshaller". This option is provided for conciseness and backwards
+compatibility. The second option of using 'extended' class names is more generic and is preferred as it allows you to
+really add new marshallers from an external module of your choice.
+
+The module identifier and slot follow the rules imposed by JBoss Modules, and the designated module must be available to
+the server either in the 'modules' folder or be deployed through the deployer service in the 'deployments' folder. In
+the second case its module identifier will be of the form 'deployment.my-jar-name.jar' and the slot will be 'main'.
+
+NOTE: The referenced module will become an automatic dependency for the Cache referencing it, so in case the jar
+defining the module is redeployed, the Cache will be restarted.
 
 [source,xml]
 ----
+<!-- using an alternative, 'stock', marshaller -->
+<compatibility marshaller="org.infinispan.query.remote.CompatibilityProtoStreamMarshaller"/>
 
-<compatibility marshaller="com.acme.CustomMarshaller"/>
+or
 
-----
+<!-- using a marshaller deployed as a module -->
+<compatibility marshaller="com.acme.my-gadgets-module:main:com.acme.CustomCompatMarshaller"/>
 
-Your custom marshaller needs to be on the classpath of the Infinispan module. You can add it by either:
+or
 
-- copying your jar to
-+
-    modules/system/layers/base/org/infinispan/main
-+
-and editing the module definition to include the jar as resource-root:
-+
-[source,xml]
-.modules/system/layers/base/org/infinispan/main/modules.xml
-----
-
-<resources>
-    ...
-    <resource-root path="acme-custom-marshallers.jar"/>
-    ...
-</resources>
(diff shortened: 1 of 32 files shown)


anistor:
Just the commit titled 'ISPN-6245 wip' is relevant. The others are just cosmetic issues. The last commit is not a complete fix.

gustavonalle:
@anistor Nice to finally see "deployment of entity classes on server" working 👍

anistor:
This is a variant that does not use the per cache manager or per cache (to be discussed) explicit module dependency configuration yet. Instead it uses just the 'extended class name' approach.

gustavonalle:
@anistor A variant which is more sound than piggybacking the entity classes in another deployments :)

gustavonalle:
There seems to be a related failure
org.infinispan.server.test.client.hotrod.HotRodCustomMarshallerIteratorIT

gustavonalle:
@anistor the commits are still with WIP

anistor:
I may have forgot to push.

gustavonalle:
@anistor We need to update docs on this. The info in the server guide recommends hacking the internal modules to deploy the compat marshaller

http://infinispan.org/docs/stable/server_guide/server_guide.html#protocol_interoperability

anistor:
yikes! I'm on it. The config of indexable classes in server needs a mention in docs too.

anistor:
Test and docs fixed!

documentation/src/main/asciidoc/server_guide/configuration.adoc
```diff
@@ -499,49 +499,45 @@ By default each protocol stores data in the cache in the most efficient format f
...
+ordinary fully qualified class name or an 'extended' class name which is composed of a JBoss Modules module identifier,
+a slot name, and a fully qualified class name separated by the ':' character, (eg. "com.acme.my-module-name:my-slot:com.acme.CustomMarshaller").
 
-[source,xml]
-----
-
-<compatibility marshaller="com.acme.CustomMarshaller"/>
+By using a plain fully qualified class name you will be able to reference only 'stock' marshallers provided by Infinispan,
+like "org.infinispan.query.remote.CompatibilityProtoStreamMarshaller". This option is provided for conciseness and backwards
+compatibility. The second option of using 'extended' class names is more generic and is preferrd as it allows you to
```
gustavonalle: typo: preferrd

documentation/src/main/asciidoc/server_guide/configuration.adoc
```diff
@@ -499,49 +499,45 @@ By default each protocol stores data in the cache in the most efficient format f
...
+really add new marshallers from an external module of your choice.
 
-----
+The module identifier and slot follow the rules imposed by JBoss Modules, and the designated module must be available to
+the server either in the 'modules' folder or be deployed through the deployer service in the 'deployments' folder. In
+the second case its module identifier will be of the form 'deployment.my-jar-name.jar' and the slot will be 'main'.
 
-Your custom marshaller needs to be on the classpath of the Infinispan module. You can add it by either:
+NOTE: The referenced module will become an automatic dependency for the Cache referencing it, so in case the jar
+defining the module is redeployed, the Cache will be restarted.
```
gustavonalle: Is it possible to not restart the cache automatically (potentially causing data loss)? Is this restart something we control or is part of Wildfly internals? 
cc @tristantarrant

anistor:
The restart is part of how the managed service container works and wildfly, consequently.

So why would you update a class if you do not want it reloaded?

documentation/src/main/asciidoc/server_guide/configuration.adoc
anistor: fixed.

gustavonalle:
> So why would you update a class if you do not want it reloaded?

I woudn't want to figure out suprised that my data is gone after I dropped a jar. If I get a WARN saying "restart required", I would do it at my earliest convenience, maybe one server at a time, in a Sunday morning :)

anistor:
There are no surprises here. It's documented :)

gustavonalle:
The issue is not everyone is well intentioned and also errors can happen. Fortunately there's a way to disable this auto-deployment  https://docs.jboss.org/author/display/WFLY10/Deployment+Scanner+configuration

gustavonalle:
LGTM, no more comments, nicely done @anistor !  👍

anistor:
I've sent an email to devlist inviting everybody to comment on the approach, as I'm, planing to reuse it, but my email was held hostage by the server. 

So until it is received by everybody, maybe @galderz @tristantarrant @Sanne would like to have a look too?

wburns:

documentation/src/main/asciidoc/server_guide/configuration.adoc
tristantarrant: In theory everything can be controlled, as long as the handlers can apply the change at runtime.
For dependencies however this is tricky, since I imagine that a deployment is actually a removal and addition and the in-between situation would be undefined.

galderz: Hmmmm, this is a bit unclear. If the marshaller is an actual module in `modules` directory, I don't remember any capability to redeploy of those jars. The only redeployment available is for jars that are in `deployments` folder, which do have their own module id... This needs further clarification.

In fact, it's unclear in the documentation if jars that are put in `deployments` jar will be picked. This used to be the case for filter/event marshallers...

I lean more towards only supporting jars in `modules` folder.

PR #5235, opened 2017-06-27
ISPN-7985 Don't create the DistributionManager in local caches

danberindei:
https://issues.jboss.org/browse/ISPN-7985

Fixes NullPointerException when the cache is local and
trace logging is enabled.


wburns:

PR #5236, opened 2017-06-27
ISPN-6997 PessimisticTxPartitionAndMergeDuringRuntimeTest.testOriginatorIsolatedPartition random failures

danberindei:
https://issues.jboss.org/browse/ISPN-6997
https://issues.jboss.org/browse/ISPN-7960

* Handle lock failures with a regular rollback command.
* Don't send TxCompletionNotificationCommands from
  PessimisticLockingInterceptor.
* Add lock recipients to the affected nodes even in case of failure.
* Send commit/rollback to all affected nodes, even if the cache
  topology did not change.
* Do not change protocols when the cluster changes size 
* Do not create a remote transaction when the originator is not a
  cluster member.
* Remove the originator check in TxInterceptor.

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
@@ -105,21 +105,17 @@ public class TxDistributionInterceptor extends BaseDistributionInterceptor {
 
    @Override
    public Object visitComputeCommand(InvocationContext ctx, ComputeCommand command) throws Throwable {
-      /**
-       * Contrary to functional commands, compute() needs to return the new value returned by the remapping function.
-       * Since we can assume that old and new values are comparable in size, fetching old value into context is acceptable
-       * and it is more efficient in case that we execute more subsequent modifications than sending the return value.
-       */
+      // Contrary to functional commands, compute() needs to return the new value returned by the remapping function.
+      // Since we can assume that old and new values are comparable in size, fetching old value into context is acceptable
+      // and it is more efficient in case that we execute more subsequent modifications than sending the return value.
       return handleTxWriteCommand(ctx, command, command.getKey());
    }
 
    @Override
    public Object visitComputeIfAbsentCommand(InvocationContext ctx, ComputeIfAbsentCommand command) throws Throwable {
-      /**
-       * Contrary to functional commands, compute() needs to return the new value returned by the remapping function.
-       * Since we can assume that old and new values are comparable in size, fetching old value into context is acceptable
-       * and it is more efficient in case that we execute more subsequent modifications than sending the return value.
-       */
+      // Contrary to functional commands, compute() needs to return the new value returned by the remapping function.
+      // Since we can assume that old and new values are comparable in size, fetching old value into context is acceptable
+      // and it is more efficient in case that we execute more subsequent modifications than sending the return value.
       return handleTxWriteCommand(ctx, command, command.getKey());
    }
 
@@ -239,15 +235,15 @@ public class TxDistributionInterceptor extends BaseDistributionInterceptor {
          }
 
          TxInvocationContext<LocalTransaction> localTxCtx = (TxInvocationContext<LocalTransaction>) rCtx;
-         Collection<Address> affectedNodes =
-               dm.getCacheTopology().getWriteOwners(getAffectedKeysFromContext(localTxCtx));
-         Collection<Address> recipients = isReplicated ? null : affectedNodes;
+         LocalTransaction localTx = localTxCtx.getCacheTransaction();
+         LocalizedCacheTopology cacheTopology = dm.getCacheTopology();
+         Collection<Address> writeOwners = cacheTopology.getWriteOwners(getAffectedKeysFromContext(localTxCtx));
+         localTx.locksAcquired(writeOwners);
+         Collection<Address> recipients =
+               isReplicated ? null : localTx.getCommitNodes(writeOwners, cacheTopology);
          CompletableFuture<Object> remotePrepare =
                prepareOnAffectedNodes(localTxCtx, (PrepareCommand) rCommand, recipients);
-         return asyncValue(remotePrepare.thenApply(o -> {
-            localTxCtx.getCacheTransaction().locksAcquired(affectedNodes);
-            return o;
-         }));
+         return asyncValue(remotePrepare);
       });
    }
 
@@ -300,14 +296,14 @@ public class TxDistributionInterceptor extends BaseDistributionInterceptor {
    protected void checkTxCommandResponses(Map<Address, Response> responseMap,
          TransactionBoundaryCommand command, TxInvocationContext<LocalTransaction> context,
          Collection<Address> recipients) {
-      OutdatedTopologyException outdatedTopologyException = null;
+      LocalizedCacheTopology cacheTopology = checkTopologyId(command);
       for (Map.Entry<Address, Response> e : responseMap.entrySet()) {
          Address recipient = e.getKey();
          Response response = e.getValue();
          if (response == CacheNotFoundResponse.INSTANCE) {
-            // No need to retry if the missing node wasn't a member when the command started.
-            if (command.getTopologyId() == stateTransferManager.getCacheTopology().getTopologyId()
-                  && !rpcManager.getMembers().contains(recipient)) {
+            // Prepare/Commit commands are sent to all affected nodes, including the ones that left the cluster.
+            // We must not register a partial commit when receiving a CacheNotFoundResponse from one of those.
+            if (!cacheTopology.getMembers().contains(recipient)) {
                if (trace) log.tracef("Ignoring response from node not targeted %s", recipient);
             } else {
                if (checkCacheNotFoundResponseInPartitionHandling(command, context, recipients)) {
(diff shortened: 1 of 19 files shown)


core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -301,13 +301,18 @@ protected void checkTxCommandResponses(Map<Address, Response> responseMap,
          TransactionBoundaryCommand command, TxInvocationContext<LocalTransaction> context,
          Collection<Address> recipients) {
       OutdatedTopologyException outdatedTopologyException = null;
+      if (command.getTopologyId() != stateTransferManager.getCacheTopology().getTopologyId()) {
```
rvansa: Why are you checking STM and not DM? Just use `checkTopologyId` in here

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -130,11 +128,9 @@ protected Object handleReadManyCommand(InvocationContext ctx, FlagAffectedComman
             }
          }
          return makeStage(stage).andExceptionally(ctx, command, (rCtx, rCommand, t) -> {
```
rvansa: This would deserve some code cleanup, you don't need to register the handler.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -130,11 +128,9 @@ protected Object handleReadManyCommand(InvocationContext ctx, FlagAffectedComman
...
-            releaseLocksOnFailureBeforePrepare(rCtx);
             throw t;
          });
       } catch (Throwable t) {
```
rvansa: Same as above, remove the try/catch clause completely.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -587,25 +587,12 @@ public Address getAddress() {
          log.tracef("dests=%s, command=%s, mode=%s, timeout=%s", recipients, rpcCommand, mode, timeout);
       Address self = getAddress();
       boolean ignoreLeavers = mode == ResponseMode.SYNCHRONOUS_IGNORE_LEAVERS || mode == ResponseMode.WAIT_FOR_VALID_RESPONSE;
-      List<Address> members = getMembers();
-      if (mode.isSynchronous() && recipients != null && !members.containsAll(recipients)) {
-         if (!ignoreLeavers) { // SYNCHRONOUS
-            Address suspect = recipients.stream().filter(a -> !members.contains(a)).findFirst().orElse(null);
-            CompletableFuture<Map<Address, Response>> future = new CompletableFuture<>();
-            future.completeExceptionally(new SuspectException(
```
rvansa: where is the exception thrown now instead?

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -684,9 +684,14 @@ private void applyTransactions(Address sender, Collection<TransactionInfo> trans
...
                tx = transactionTable.getRemoteTransaction(gtx);
                if (tx == null) {
-                  tx = transactionTable.getOrCreateRemoteTransaction(gtx, transactionInfo.getModifications());
-                  // Force this node to replay the given transaction data by making it think it is 1 behind
-                  ((RemoteTransaction) tx).setLookedUpEntriesTopology(topologyId - 1);
+                  try {
+                     tx = transactionTable.getOrCreateRemoteTransaction(gtx, transactionInfo.getModifications());
+                     // Force this node to replay the given transaction data by making it think it is 1 behind
+                     ((RemoteTransaction) tx).setLookedUpEntriesTopology(topologyId - 1);
+                  } catch (IllegalStateException e) {
```
rvansa: If this is supposed to be catch-all-and-ignore clause, make it so.

core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
```diff
@@ -369,20 +369,31 @@ private RemoteTransaction getOrCreateRemoteTransaction(GlobalTransaction globalT
...
+      return remoteTransactions.compute(globalTx, (gtx, existing) -> {
+         if (existing != null) {
+            if (trace)
+               log.tracef("Remote transaction already registered: %s", existing);
+            return existing;
+         } else {
+            if (isTransactionCompleted(globalTx)) {
+               throw new IllegalStateException("Cannot create remote transaction " + globalTx + ", already completed");
+            }
+            if (!rpcManager.getTransport().getMembers().contains(gtx.getAddress())) {
```
rvansa: Why are you checking the members inside the compute block? I smell race conditions since the membership change does not have to lock this map.

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
danberindei: Because completion didn't find `dm` :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
danberindei: JGroups marks as suspected any recipient that's not a member of the cluster view at the time the request is sent, and that gets handled in `Responses`.

core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
danberindei: Good catch, I was incorrectly assuming that the iteration in `cleanupLeaverTransactions()` would block if an entry was locked by `compute()`. I'm moving the originator check after `compute()`.

danberindei:
@rvansa I tried to explain what's happening in the JIRA comments. Basically UNICAST3 retransmits all messages sent to nodes from a different partition after the merge, if the cluster didn't stay split for long (https://issues.jboss.org/browse/JGRP-2194), and the ISPN-6997 commit is to make sure that if a `LockControlCommand` is delivered after the merge, the follow-up `PrepareCommand` or `RollbackCommand` is delivered as well.

Unfortunately, that fix breaks something else: if the split happens during commit, the originator's partition is still available when the target is suspected, so the commit is retried, and somehow it arrives on the other partition (not sure why, but it doesn't matter, it could just as well be retransmitted during the merge, and delivered before our transport is notified of the merged view). The remote node sees a new topology id, replays the modifications, and then rolls back the transaction because the originator isn't in its cluster view. The problem is that the transaction is not supposed to be rolled back in either partition, the originator is supposed to commit it after the merge instead, so the ISPN-7960 commit avoids the rollback by moving that originator check from `TxInterceptor.verifyRemoteTransaction()` to `TransactionTable.getOrCreateTransaction()`.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
rvansa: `Responses` don't throw `SuspectException`, does it mean that now you will get empty response map when sending with `SYNCHRONOUS`?

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
```diff
@@ -604,6 +604,8 @@ private Object verifyRemoteTransaction(RemoteTxInvocationContext ctx, AbstractTr
             log.tracef("Rolling back remote transaction %s because either already completed (%s) or originator no longer in the cluster (%s).",
                        globalTransaction, alreadyCompleted, originatorMissing);
          }
+         // The rollback command only marks the transaction as completed in invokeAsync()
+         txTable.markTransactionCompleted(globalTransaction, false);
```
rvansa: Is this the only place that marks transaction for rollback when the lock command fails? Or is it up to the user to handle exception being thrown by the operation and mark the tx?

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
danberindei: No, `JGroupsTransport.checkRsp()` still throws a `SuspectException` -- I had forgotten that `Responses` returns the entire `Rsp` and the `Response` is extracted later.

wburns:
Integrated into master, thanks @danberindei

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
rvansa: This comment got hidden, could you respond @danberindei ?

danberindei: Missed it, sorry. It would have been nice if I didn't, because I would have realized I can remove the `markTransactionCompleted()` call (at this point we know the transaction was already marked as completed) and some more dead code.

Now to the answer... `verifyRemoteTransaction` doesn't care if the lock command failed or not, only if the originator left the cluster (removed now) or if another thread already marked the transaction as completed (either because the originator left the cluster, or because the lock/prepare RPC timed out and the originator already sent out a rollback command).

`TxInterceptor` never marked the transaction for rollback anyway, only released the current command's locks, because of the `FAIL_SILENTLY` flag. IMO it would have been much better if the flag only applied to non-optimistic lock timeouts, but it applies to all exceptions, so it's `InvocationContextInterceptor` that marks the transaction for rollback (and only the TM transaction on the originator, not the cache transaction).

PR #5240, opened 2017-06-27
ISPN-7986 Hot RodTransaction (Server only)

pruivo:
https://issues.jboss.org/browse/ISPN-7986

Hot Rod server is able to handle transaction from a client.
The Hot Rod client will be implemented later.
Optimistic and Total Order transactions aren't supported yet due to a bug between hot rod operations and versioning.

question: should the protocol be increase to 2.7? or only when the client is implemented?


wburns:
I would personally say we may want to bump the protocol to 3.0. Adding transactions is a pretty big change to me. Anyone else have an opinion?

This seems like it would be 9.2.0 as well right?

pruivo:
@wburns I would like to have this PR in 9.1. Although the server can handle TX, the client can't so it is harmless. The version could be incremented when the client is implemented as well.

gustavonalle:
IMO the version should be increased now to 2.7, even without the clients. Great opportunity to test older clients (2.6) with newer servers (2.7)  :)

wburns:
I am guessing we will need a CR2 build then to contain this. This seems like an awful lot of changes for a Final build.

wburns:
Although @tristantarrant would control that.

tristantarrant:
Too late for 9.1. this will land early on in 9.2

rvansa:
Wrt versions - I second 2.7 (server side). With the advent of builiding blocks (counters, multimaps...) I would like to see 3.0 being modular to accommodate these.

tristantarrant:
I've never liked the single global protocol version approach. It would make
more sense to leave the global version just as a way to indicate the
general operation framing and format, version the individual ops (á la
JGroups) and return the supported (versioned) ops as part of PING.
e.g. PUT PUT2 GET ...

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TxState.java
```diff
@@ -0,0 +1,183 @@
...
+   private static List<WriteCommand> copyModifications(Collection<WriteCommand> modifications) {
+      return modifications == null || modifications.isEmpty() ?
+            null :
+            Collections.unmodifiableList(new ArrayList<>(modifications));
+   }
+
+   private static List<WriteCommand> unmarshallModifications(ObjectInput input)
+         throws IOException, ClassNotFoundException {
+      List<WriteCommand> modifications = MarshallUtil.unmarshallCollection(input, ArrayList::new);
+      return modifications == null ? null : Collections.unmodifiableList(modifications);
```
rvansa: do we need to be *that* defensive when the `getModifications` is package-private?

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TxState.java
```diff
@@ -0,0 +1,183 @@
...
+            null;
+   }
+
+   TxState rollback() {
+      return status.canAdvance(TxStatus.ROLLED_BACK) ?
+            new TxState(globalTransaction, TxStatus.ROLLED_BACK, null) :
+            null;
+   }
+
+   public enum TxStatus {
```
rvansa: I don't fancy another enum describing tx state... can't you use `javax.transaction.Status` codes and a little switch in static `canAdvance`?

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_wire.adoc
```diff
@@ -2013,6 +2014,151 @@ Here are the current values for each flag: +
...
+
+|OnePhaseCommit
+|byte
+|When it is set to `1`, the server will use one-phase-commit if available (XA only)
+
+|Number of keys
+|vInt
+|The number of keys
+
+3+^|For each key
```
rvansa: You should mention that the server requires all keys to be distinct here.

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TransactionDecodeContext.java
```diff
@@ -0,0 +1,105 @@
...
+   TransactionDecodeContext(AdvancedCache<byte[], byte[]> cache, XidImpl xid) {
+      this.xid = xid;
+      this.rpcManager = cache.getRpcManager();
+      this.localAddress = rpcManager != null ? rpcManager.getAddress() : null;
+      ComponentRegistry registry = cache.getComponentRegistry();
+      this.commandsFactory = registry.getCommandsFactory();
+      this.transactionTable = registry.getComponent(TransactionTable.class);
+      this.clusteringDependentLogic = registry.getComponent(ClusteringDependentLogic.class);
+      this.serverTransactionTable = registry.getComponent(ServerTransactionTable.class);
+      this.tm = (EmbeddedTransactionManager) cache.getTransactionManager();
```
rvansa: I don't really understand *why* we need a JTA transaction manager at all, when we force it to be the `EmbeddedTransactionManager` anyway. I think the transaction manager should be only on the user-code side, not in server. Please elaborate.

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_wire.adoc
```diff
@@ -129,15 +129,16 @@ When topology-aware or hash-distribution-aware clients will send 0 until they
...
-| Transaction Type    | 1 byte     | This is a 1 byte field, containing one
-of the following well-known supported transaction types (For this version of
-the protocol, the only supported transaction type is 0): +
-+0+ = Non-transactional call, or client does not support transactions.
-The subsequent TX_ID field will be omitted. +
-+1+ = X/Open XA transaction ID (XID). This is a well-known, fixed-size format.
-| Transaction Id      | byte array |  The byte array uniquely identifying the
-transaction associated to this call. Its length is determined by the
-transaction type. If transaction type is 0, no transaction id will be present.
+//TODO this isn't implemented. Implement it will break compatbility with older clients
```
pruivo: there isn't nothing to remove. Decoder10 reads the transactionType but in Decoder2x it don't read it. 
I checked the client code as well and only the Codec13 sends the transactionType. Starting from Codec20, it isn't sent.

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_wire.adoc
```diff
@@ -2013,6 +2014,151 @@ Here are the current values for each flag: +
...
+| Length of key.
+Note that the size of a vInt can be up to 5 bytes which in theory can produce bigger numbers than `Integer.MAX_VALUE`.
+However, Java cannot create a single array that’s bigger than `Integer.MAX_VALUE`, hence the protocol is limiting vInt array lengths to `Integer.MAX_VALUE`.
+
+|Key
+|byte array
+|Byte array containing the key
+
+|Version Read
+|long
```
pruivo: what? why complicated? The `RemoteCache` interface uses long (e.g. `removeWithVersion()`), and I don't think we are going to change it. The only issue we have is that the hot rod version is overwriting the entry version (optimistic transactions) and it shouldn't.

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_wire.adoc
```diff
@@ -2013,6 +2014,151 @@ Here are the current values for each flag: +
...
+|Byte
+|The first 2 bits are the read status: +
++0x01+ = `NOT_READ` +
++0x02+ = `NON_EXISTING` +
+And the 3rd bit is set when it is a remove operation:
++0x04+ = `REMOVE_OPERATION` +
+
+|Lifespan
+|vInt
+|Number of seconds that a entry during which the entry is allowed to life.
```
pruivo: ok

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_wire.adoc
```diff
@@ -2013,6 +2014,151 @@ Here are the current values for each flag: +
...
+
+|Xid
+|XID
+|The transaction ID (XID)
+
+|===
+
+.Response from prepare, commit and rollback request.
+
+Response 0x3C, 0x3E and 0x40
```
pruivo: I think so. In the docs there are a bunch of operations with the same response format and different response code (e.g. putIfAbsent, replace)

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_wire.adoc
```diff
@@ -2013,6 +2014,151 @@ Here are the current values for each flag: +
...
+|The XA code representing the prepare response. +
+Can be `XA_OK(0)`, `XA_RDONLY(3)` or any of the error codes (see `XaException`). +
+This field isn't present if the response state is different from `Successful`.
+
+|===
+
+
+.XID Formtat
+
+The XID in the requests has the following format:
```
pruivo: why? the transaction doesn't live in the server. only in the client. The script wouldn't have access to the transaction context (keys read/written by the tx) in the server.

server/hotrod/src/main/java/org/infinispan/server/hotrod/CacheDecodeContext.java
```diff
@@ -58,6 +75,168 @@ public RequestParameters getParams() {
...
+    */
+   TransactionResponse rollbackTransaction() {
+      checkConfigurationAndTransactionManager();
+      return finishTransaction(new RollbackTransactionDecodeContext(cache, (XidImpl) operationDecodeContext));
+   }
+
+   /**
+    * Handles a prepare request from a client
+    */
+   Response prepareTransaction() throws Exception {
```
pruivo: ok

server/hotrod/src/main/java/org/infinispan/server/hotrod/CacheDecodeContext.java
```diff
@@ -58,6 +75,168 @@ public RequestParameters getParams() {
...
+      checkConfigurationAndTransactionManager();
+
+      PrepareTransactionContext context = (PrepareTransactionContext) operationDecodeContext;
+      if (context.isEmpty()) {
+         //the client can optimize and avoid contacting the server when no data is written.
```
pruivo: Our implementation would sent it but I'm not preventing a user implementation to send an empty modification set.

server/hotrod/src/main/java/org/infinispan/server/hotrod/CacheDecodeContext.java
```diff
@@ -58,6 +75,168 @@ public RequestParameters getParams() {
...
+   }
+
+   /**
+    * Commits or Rollbacks the transaction (second phase of two-phase-commit)
+    */
+   private TransactionResponse finishTransaction(SecondPhaseTransactionDecodeContext txContext) {
+      try {
+         txContext.perform();
+      } catch (HeuristicMixedException e) {
+         return createTransactionResponse(header, XAException.XA_HEURMIX);
```
pruivo: I'll double check but I'm pretty sure that the core logs the exception.

server/hotrod/src/main/java/org/infinispan/server/hotrod/CacheDecodeContext.java
```diff
@@ -58,6 +75,168 @@ public RequestParameters getParams() {
...
+         if (isTrace) {
+            log.tracef("Unable to start transaction %s", context.getXid());
+         }
+         return decoder.createNotExecutedResponse(header, null);
+      }
+
+      //forces the write-lock. used by pessimistic transaction. it ensures the key is not written after is it read and validated
+      //optimistic transaction will use the write-skew check.
+      AdvancedCache<byte[], byte[]> readCache = cache.withFlags(Flag.FORCE_WRITE_LOCK);
+      boolean failed = false;
```
pruivo: ok

server/hotrod/src/main/java/org/infinispan/server/hotrod/CacheDecodeContext.java
```diff
@@ -301,6 +484,36 @@ ComponentRegistry getCacheRegistry(String cacheName) {
...
+            log.tracef("Operation %s wasn't read.", write);
+         }
+         return true;
+      }
+      CacheEntry<byte[], byte[]> entry = readCache.getCacheEntry(write.key);
+      if (write.wasNonExisting()) {
+         if (isTrace) {
+            log.tracef("Key didn't exist for operation %s. Entry is %s", write, entry);
+         }
+         return entry == null;
```
pruivo: ok

server/hotrod/src/main/java/org/infinispan/server/hotrod/CacheDecodeContext.java
```diff
@@ -301,6 +484,36 @@ ComponentRegistry getCacheRegistry(String cacheName) {
...
+      }
+      if (isTrace) {
+         log.tracef("Checking version for operation %s. Entry is %s", write, entry);
+      }
+      return entry != null && write.versionRead == extractVersion(entry.getMetadata().version());
```
pruivo: I'll use the extractVersion until the issue is fixed. Otherwise, it won't be compliant since the getWithMetadata uses this method to send the version back to the client (and it is this version that is sent to validate)

server/hotrod/src/main/java/org/infinispan/server/hotrod/Decoder2x.java
```diff
@@ -332,32 +345,120 @@ public void customReadHeader(HotRodHeader header, ByteBuf buffer, CacheDecodeCon
...
+   private Optional<PrepareTransactionContext> readPrepareTxContext(ByteBuf buffer) {
+      return readXid(buffer)
+            .flatMap(xid -> readMaybeByte(buffer)
+                  .flatMap(onePhaseCommit -> readMaybeVInt(buffer)
+                        .map(numberOfWrites ->
+                              new PrepareTransactionContext(xid, onePhaseCommit == 1, numberOfWrites))));
+   }
+
+   private Optional<TransactionWrite> readTransactionWrite(HotRodHeader header, ByteBuf byteBuf) {
+      return readMaybeRangedBytes(byteBuf).flatMap(key ->
```
pruivo: I don't like it either but it seems the way to avoid blocking the reader thread for data.

server/hotrod/src/main/java/org/infinispan/server/hotrod/Decoder2x.java
```diff
@@ -332,32 +345,120 @@ public void customReadHeader(HotRodHeader header, ByteBuf buffer, CacheDecodeCon
...
+            readMaybeLong(byteBuf).flatMap(versionRead ->
+                  readMaybeByte(byteBuf).flatMap(control -> {
+                     if (ControlByte.REMOVE_OP.hasFlag(control)) {
+                        return Optional.of(new TransactionWrite(key, versionRead, control, null, null, null));
+                     }
+                     return readLifespanOrMaxIdle(byteBuf, hasFlag(header, ProtocolFlag.DefaultLifespan))
```
pruivo: yes, I'm an idiot :)

server/hotrod/src/main/java/org/infinispan/server/hotrod/Decoder2x.java
```diff
@@ -332,32 +345,120 @@ public void customReadHeader(HotRodHeader header, ByteBuf buffer, CacheDecodeCon
...
+
+   private Optional<XidImpl> readXid(ByteBuf byteBuf) {
+      return readMaybeSignedInt(byteBuf)
+            .flatMap(formatId -> readMaybeRangedBytes(byteBuf)
+                  .flatMap(gtxId -> readMaybeRangedBytes(byteBuf)
+                        .map(branchId -> XidImpl.create(formatId, gtxId, branchId))));
+   }
+
+   private Optional<Integer> readLifespanOrMaxIdle(ByteBuf buffer, boolean useDefault) {
+      return readMaybeVInt(buffer).map(stream -> {
```
pruivo: copy-pasted... I'll change it.

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/ServerTransactionOriginatorChecker.java
```diff
@@ -0,0 +1,50 @@
...
+      return isOriginatorMissing(gtx, rpcManager.getMembers());
+   }
+
+   @Override
+   public boolean isOriginatorMissing(GlobalTransaction gtx, Collection<Address> liveMembers) {
+      return !liveMembers.contains(gtx.getAddress()) && isNonClientTransaction(gtx);
+   }
+
+   private boolean isNonClientTransaction(GlobalTransaction gtx) {
+      return globalTxTable.values().stream()
```
pruivo: a client address where?

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TxFunctions.java
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Functions and Predicated and their externalizer.
+ * <p>
+ * To be used on cache streams.
+ *
+ * @author Pedro Ruivo
+ * @since 9.1
+ */
+public class TxFunctions {
```
pruivo: I can try but I found out it sends the function when topology is changing...

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TxState.java
pruivo: ok, no more defensive code :)

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TxState.java
pruivo: ok

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TransactionDecodeContext.java
pruivo: All the core code interacts with TransactionManager and Transaction interface. I don't see any other way around exception re-write the core... What idea do you have in mind?

pruivo:
Updated:
* advanced version to 2.7
* refactor the expiration decoding  (it was bugged)

rvansa:
Re [entry version as long]: In general, Infinispan supports and `EntryVersion` implementations; you're right that with current `RemoteCache` the full version info is omitted. It's not 'only overwriting the entry version', it's breaking any use of versions right now.
As I've said, we need a plan for `EntryVersion` in general. Either we enforce `NumericVersion` everywhere (which would be unfortunate IMO, as we might find good use cases for that - e.g. SC uses SimpleClusteredVersion, too - I know we currently don't support that together with Hot Rod, but...) or we accept other version types at least in the protocol, eventually adjusting RemoteCache API as well (temporarily scrambling other versions to long, but not ignoring parts). Do you see any other options?

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_wire.adoc
rvansa: And that's part of the problem. But running a script within the scope of a transaction and not participating in that transaction is not correct. If you don't want to deal with that, clearly describe the limitations in docs.

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/ServerTransactionOriginatorChecker.java
rvansa: `GlobalTransaction.getAddress()` would return an instance of `Address` that does not match to any node address.

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TxFunctions.java
rvansa: Ah, when the cache has joined the replicated cache but did not receive all data yet, it probably has to execute a stream operation remotely. In any way, doing the transaction originator check using a RPC is a no-go IMO.

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TransactionDecodeContext.java
rvansa: What core code? Interceptors should interact only with `TxInvocationContext`, only CacheWriterInterceptor and CacheNotifierImpl suspend the running tx. I've found that applying state in `StateConsumerImpl` creates it's own transaction using the TM, but I am not sure how much that'd be necessary there.

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/ServerTransactionOriginatorChecker.java
pruivo: that will break the transactions. GlobalTransaction needs to be unique and your suggestion would make all nodes to create the GlobalTransaction with the same Address. 
I guess I can create a client address per server/cache manager if that is ok for you. I'm open to other suggestions as well :)

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TransactionDecodeContext.java
pruivo: but the CacheImpl interacts with the TransactionManager to map the Transaction > LocalTransaction.

pruivo:
the plan is to remove it from the public interface. We support EntryVersion but we don't. If you try to configure an optimistic tx cache with a custom VersionGenerator, it will give you an error. If you configure a custom VersionGenerator with any other type of cache, it will be simply ignored. It shouldn't be public anyway. 
And the eventual consistency stuff will probably needs another implementation....

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/ServerTransactionOriginatorChecker.java
rvansa: Yep, that sounds good.

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/TransactionDecodeContext.java
rvansa: You don't have to use CacheImpl directly in the server; I suggest overriding `DecoratedCache.readContext`/`writeContext` and pass the transaction identifier to the created context directly.

pruivo: it isn't possible. I can't decorate the cache and I really don't want to re-implement the cache.
```
java.lang.ClassCastException: org.infinispan.cache.impl.EncoderCache cannot be cast to org.infinispan.cache.impl.CacheImpl
        at org.infinispan.cache.impl.DecoratedCache.<init>(DecoratedCache.java:70) ~[infinispan-core-9.1.1-SNAPSHOT.jar:9.1.1-SNAPSHOT]
        at org.infinispan.cache.impl.DecoratedCache.<init>(DecoratedCache.java:56) ~[infinispan-core-9.1.1-SNAPSHOT.jar:9.1.1-SNAPSHOT]
```

pruivo:
Updated:
* new commit with GlobalTransaction created with a ClientAddress. It avoid iterate over the values to find if it is a client transaction or not.
* new commit with removal of TransactionManager... it is a bit hacky.

pruivo:
updated:
* new commit: added factory method to tests (test with different transaction modes)

ps. commits will be squashed when review is finished.

tristantarrant:

pruivo:
reopened PR for 9.2 alpha

pruivo:
rebased

galderz:
New failure?

```
org.infinispan.server.test.jmx.management.JmxManagementIT.testCacheManagerAttributes
```

pruivo:
should be fixed now :+1:

galderz:
Integrating, thanks @pruivo! :)

PR #5244, opened 2017-06-27
ISPN-7979 Add licenses to distributions

wburns:
* Added server distribution

https://issues.jboss.org/browse/ISPN-7979


wburns:
Have to update console packages to show in a better way for html.

wburns:
Also looking closer, the wildfly-server-provisioning-maven-plugin doesn't generate the versions for the licenses file... sigh

PR #5246, opened 2017-06-28
ISPN-7988 ScatteredSyncFuncTest#testCompute* failures

rvansa:
https://issues.jboss.org/browse/ISPN-7988

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
@@ -1,5 +1,6 @@
 package org.infinispan.interceptors.distribution;
 
+import org.infinispan.commands.AbstractVisitor;
 import org.infinispan.commands.FlagAffectedCommand;
 import org.infinispan.commands.MetadataAwareCommand;
 import org.infinispan.commands.TopologyAffectedCommand;
@@ -53,6 +54,7 @@ import org.infinispan.distribution.group.impl.GroupManager;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.functional.impl.FunctionalNotifier;
 import org.infinispan.interceptors.InvocationSuccessAction;
+import org.infinispan.interceptors.InvocationSuccessFunction;
 import org.infinispan.interceptors.impl.ClusteringInterceptor;
 import org.infinispan.metadata.EmbeddedMetadata;
 import org.infinispan.metadata.Metadata;
@@ -132,6 +134,7 @@ public class ScatteredDistributionInterceptor extends ClusteringInterceptor {
          cacheNotifier.notifyCacheEntryRemoved(entry.getKey(), entry.getValue(), entry.getMetadata(), false, rCtx, (ClearCommand) rCommand);
       }
    };
+   private InvocationSuccessFunction handleWritePrimaryResponse = this::handleWritePrimaryResponse;
 
    @Inject
    public void injectDependencies(GroupManager groupManager, ScatteredVersionManager svm, TimeService timeService,
@@ -186,8 +189,7 @@ public class ScatteredDistributionInterceptor extends ClusteringInterceptor {
          } else { // not primary owner
             CompletableFuture<Map<Address, Response>> rpcFuture =
                rpcManager.invokeRemotelyAsync(info.writeOwners(), command, defaultSyncOptions);
-            return asyncValue(rpcFuture.thenApply(responseMap ->
-                  handleWritePrimaryResponse(ctx, command, responseMap)));
+            return asyncValue(rpcFuture).thenApply(ctx, command, handleWritePrimaryResponse);
          }
       } else { // remote origin
          if (info.isPrimary()) {
@@ -283,11 +285,10 @@ public class ScatteredDistributionInterceptor extends ClusteringInterceptor {
       }
    }
 
-   private <T extends DataWriteCommand & MetadataAwareCommand> Object handleWritePrimaryResponse(
-         InvocationContext ctx, T command, Map<Address, Response> responseMap) {
-      Response response = getSingleResponse(responseMap);
+   private Object handleWritePrimaryResponse(InvocationContext ctx, VisitableCommand command, Object rv) throws Throwable {
+      Response response = getSingleResponse((Map<Address, Response>) rv);
       if (!response.isSuccessful()) {
-         command.fail();
+         ((DataWriteCommand) command).fail();
          if (response instanceof UnsuccessfulResponse) {
             return ((UnsuccessfulResponse) response).getResponseValue();
          } else {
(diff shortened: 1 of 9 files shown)


danberindei:
@karesti rvansa I trust your judgement, I have no idea what the fix is doing :) 

@rvansa Lots of extended statistics test failures in CI for this PR. And there's another test failure that might be related: https://issues.jboss.org/browse/ISPN-7995

rvansa:
@karesti Please review again, changing the `ValueMatcher` was a step into wrong direction.

karesti:
@rvansa at the time I think I used the valuematcher, and it failed. but looking at you code seemed good to me, so I thought something was missing from my side at the time.

rvansa:
@karesti Weill it is an option to do things; however it would be unrelated. And with version history the value matchers will be gone anyway, so I've left it there.

danberindei:
@rvansa I know you said the fix is simple in the JIRA comment, but I still don't know what went wrong in the first place :)

rvansa:
@danberindei The thing that went wrong was that the compute command was successful even if it did not match the criteria (absent/present). Also, compute command has to send the whole new value from primary to originator and store it (if successful) because originator does not know the previous value - it cannot generate it in order to store that for backup.

core/src/test/java/org/infinispan/scattered/ScatteredSyncFuncTest.java
```diff
@@ -19,4 +24,37 @@ public ScatteredSyncFuncTest() {
    protected void assertOwnershipAndNonOwnership(Object key, boolean allowL1) {
       Utils.assertOwnershipAndNonOwnership(caches, key);
    }
+
+   public void testCompute() {
```
karesti: In general, all the test I made before concerned more scenarios.

1) compute/computeIfPresent/computeIfAbsent with existing value in the cache
2) compute/computeIfPresent for an existing value that should be removed if the lambda computes null. computeIfAbsent should do nothing
3) compute/computeIfPresent/computeIfAbsent when a value is not present (computeIfPresent) should do nothing

There are 2 commands Compute for compute/computeIfPresent and ComputeIfAbsent command. You may wanna cover all the scenarios

danberindei:
@rvansa Why does the originator need to send the previous value to the backup? Doesn't the primary send the updates to the backup?

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
```diff
@@ -186,8 +189,13 @@ public void injectDependencies(GroupManager groupManager, ScatteredVersionManage
          } else { // not primary owner
             CompletableFuture<Map<Address, Response>> rpcFuture =
                rpcManager.invokeRemotelyAsync(info.writeOwners(), command, defaultSyncOptions);
-            return asyncValue(rpcFuture.thenApply(responseMap ->
-                  handleWritePrimaryResponse(ctx, command, responseMap)));
+            return asyncValue(rpcFuture.thenApply(responseMap -> {
+               try {
+                  return handleWritePrimaryResponse(ctx, command, responseMap);
+               } catch (Throwable t) {
+                  throw CompletableFutures.asCompletionException(t);
```
danberindei: With `asyncValue(rpcFuture).thenApply((rCtx, rCommand, responseMap) -> ...)` you wouldn't need the `try..catch` (but you would need a cast for the `responseMap`).

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
```diff
@@ -1059,4 +1035,117 @@ public final Object visitGetKeysInGroupCommand(InvocationContext ctx,
...
+         if (command.isReturnValueExpected()) {
+            if (!(responseValue instanceof MetadataImmortalCacheValue)) {
+               throw new CacheException("Expected MetadataImmortalCacheValue as response but it is " + responseValue);
+            }
+            MetadataImmortalCacheValue micv = (MetadataImmortalCacheValue) responseValue;
+            version = micv.getMetadata().version();
+            returnValue = micv.getValue();
+         } else {
+            if (!(responseValue instanceof EntryVersion)) {
+               throw new CacheException("Expected NumericVersion as response but it is " + responseValue);
```
danberindei: Should it be "Expected EntryVersion"?

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
```diff
@@ -1059,4 +1035,117 @@ public final Object visitGetKeysInGroupCommand(InvocationContext ctx,
...
+      }
+
+      @Override
+      public Object visitReadWriteManyCommand(InvocationContext ctx, ReadWriteManyCommand command) throws Throwable {
+         throw new UnsupportedOperationException();
+      }
+
+      @Override
+      public Object visitReadWriteManyEntriesCommand(InvocationContext ctx, ReadWriteManyEntriesCommand command) throws Throwable {
+         throw new UnsupportedOperationException();
```
danberindei: You could implement `AbstractVisitor` and only implement `handleDefault` to throw `UnsupportedOperationException`.

core/src/test/java/org/infinispan/scattered/ScatteredSyncFuncTest.java
rvansa: I'll add those tests...

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
rvansa: I could, but that way I'd get checkstyle warnings for not handling these other commands. I haven't figured out how to disable the checkstyle for given inner class in IntelliJ :-/

rvansa:
I've increased coverage in `ScatteredSyncFuncTest` and applied Dan's suggestions.

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
rvansa: What separate classes? I use visitor as switch on command type.

danberindei: My mistake, I saw the method names `handleDataWriteCommand` and `handleComputeCommand` and I assumed that `ComputeCommand` is not a `DataWriteCommand`. Since `handleWritePrimaryResponse` casts the command to a `DataWriteCommand`, I further assumed that there must be another method using a `PrimaryResponseHandler` with a `ComputeCommand`, and that you have a neat separation between the two use cases :)

danberindei:
Looks good to me, @karesti feel free to integrate it if you don't have any more comments and tests pass.

PR #5251, opened 2017-06-30
ISPN-7998 Multi-key write commands don't work with union CHs

danberindei:
The distribution interceptors assume all segments have
CH.getNumOwners() owners, but in a union CH that's not true.

https://issues.jboss.org/browse/ISPN-7998

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/distribution/NonTxDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/NonTxDistributionInterceptor.java
@@ -116,7 +116,7 @@ public class NonTxDistributionInterceptor extends BaseDistributionInterceptor {
             map.put(member, segments);
          }
          map.remove(rpcManager.getAddress());
-      } else if (ch.getNumOwners() > 1) {
+      } else {
          for (Integer segment : segments) {
             List<Address> owners = ch.locateOwnersForSegment(segment);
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/infinispan/interceptors/distribution/NonTxDistributionInterceptor.java
```diff
@@ -700,7 +700,7 @@ public int containerSize(Collection<Object> list) {
 
       @Override
       public boolean shouldRegisterRemoteCallback(ReadWriteManyCommand cmd, ConsistentHash ch) {
```
pruivo: maybe remove the `ConsistentHash` from the method signature?

core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
```diff
@@ -150,7 +150,7 @@ private Object handleRemotePutMapCommand(InvocationContext ctx, PutMapCommand co
       final ConsistentHash ch = cacheTopology.getWriteConsistentHash();
```
pruivo: can be removed now :)

danberindei:
@pruivo Updated to remove the unnecessary parameter/local.

pruivo:

PR #5253, opened 2017-06-30
ISPN-7997 DistributedStreamIteratorTest.testLocallyForcedStream random failures

danberindei:
https://issues.jboss.org/browse/ISPN-7997
https://issues.jboss.org/browse/ISPN-8092


rvansa:
Why have you deleted the byte[] test?

danberindei:
@rvansa Because it didn't test anything since the `Equivalence` classes were removed. If you think it's still useful, please rename it to something else.

danberindei:
No longer causes failures in the scattered stream tests now that ISPN-8002 is in. There are some `DirectoryProviderModuleMemberRegistrationIT` failures in the latest run, but I believe they are unrelated.

core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
```diff
@@ -90,4 +100,52 @@ public String convert(Object key, String value, Metadata metadata) {
...
+
+   private static class TestDefaultConsistentHashFactory
+         extends BaseControlledConsistentHashFactory<DefaultConsistentHash> {
+      TestDefaultConsistentHashFactory() {
+         super(new DefaultTrait(), 3);
+      }
+
+      @Override
+      protected List<Address> createOwnersCollection(List<Address> members, int numberOfOwners, int segmentIndex) {
+         if (members.size() < 3)
```
pruivo: you probably want to remove this condition otherwise members[0] will be always the primary owner of all the segments.

core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
```diff
@@ -90,4 +100,52 @@ public String convert(Object key, String value, Metadata metadata) {
...
+            return members;
+
+         switch (segmentIndex) {
+            case 0:
+               return Arrays.asList(members.get(0), members.get(1)).subList(0, numberOfOwners);
+            case 1:
+               return Arrays.asList(members.get(1), members.get(2)).subList(0, numberOfOwners);
+            case 2:
+               return Arrays.asList(members.get(2), members.get(1)).subList(0, numberOfOwners);
```
pruivo: should members[0] be the backup owner?

pruivo:
@danberindei I think the PREVIEW can be removed now?
are you planning to fix ScatteredStreamIteratorTest in this PR?

core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
danberindei: That shouldn't make any difference. I'll add a comment.

core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
danberindei: Why?

danberindei:
@pruivo I guess there was a reason to keep the preview label after all :)

I'll try to debug `ScatteredStreamiteratorTest` again, I thought it would be fixed by https://github.com/infinispan/infinispan/pull/5258

danberindei:
@pruivo @rvansa I think the `ScatteredStreamIteratorTest` are caused by https://issues.jboss.org/browse/ISPN-8092

core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
pruivo: because it only has one segment assign to it :)
```
segment_0 => 0 & 1
segment_1=> 1 & 2
segment_2 => 2 & 0?
```

pruivo:
@danberindei 
* why change the cache names in the stress tests?
* may mark ScatteredStreamIteratorTest as unstable? I don't feel comfortable to integrate a PR that fixes 1 test and fails 3 more :)

core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
danberindei: The tests need keys with owners 2&1 and with owners 1&2. I could add another segment, but none of the `MagicKeys` generated by the tests would use it.

pruivo: it looks like you have 3 segments already :)

danberindei:
@pruivo 1) Because having the full class name as the cache name is really annoying, and if someone sees the pattern in the stress test they might think it's a good idea.
2) I don't intend to integrate this before the ISPN-8092 fix.

core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
danberindei: @pruivo Yes, because there are other test methods that need primary owner 0.

danberindei:
Updated to include a fix for ISPN-8092 and to use `int[][]` even more in the test consistent hash factories.

pruivo:
@danberindei looks good to me. I'm going to wait until CI finishes.

pruivo:

PR #5254, opened 2017-06-30
ISPN-7963 Relocate Infinispan Hibernate cache provider

galderz:
Relocate Infinispan Hibernate cache provider, along with all its related history to Infinispan repository.

Please bear in mind the following:

1. The method I used to bring the cache provider along with all its history is [this one](https://stackoverflow.com/a/24693985/186429).

2. The new Maven location of the cache provider would be: `org.infinispan:infinispan-hibernate-cache`, located under `hibernate-cache` folder. Thoughts/suggestions welcome :)

3. I have not changed the code from tabs to spaces because it'd make a mess in terms of annotated commits or blames when trying to figure out the cause of each line change. Radim and I have been working on this code base since 2009 and I think it'd be detrimental to the quality of the code to do such tabs to spaces change. We are not working on this code on a daily basis, so it's important to be able to easily figure out why/when each line of code was added. Blanket style related changes are a PITA from this perspective. Hence, I created a hibernate-cache specific checkstyle file to avoid build failing. There might be easier ways to do this.

Some TODOs still left:

1. Remove licensing comments from files. This is easy to do and won't affect tracing back code changes.

2. Make the hibernate-cache pom.xml produce OSGI bundles.

3. When happy, port over this code to 9.0.x branch. If master ends up moving to 9.2.x before this is in, we'd also port this to 9.1.x branch.

4. Make the corresponding changes in the Hibernate git repository to remove the cache provider and add artifact relocation details.

5. Change the package from `org.hibernate.cache.infinispan` to `org.infinispan.hibernate.cache`

Remember: this is a preview PR, so focus on things like location, Maven coordinates, overall style...etc. Once those are agreed, I can finish the remaining TODOs that affect this repository and address any little details.


galderz:
One last thing: testsuite should be passing without any errors.

Sanne:
@galderz great progress!

What about the licensing terms? I see you maintained the LGPL header - which is the right thing to do - but I wonder if there are some other areas in the Infinispan project in which we need to clarify that some modules are not ASL2.

galderz:
> What about the licensing terms? I see you maintained the LGPL header - which is the right thing to do - but I wonder if there are some other areas in the Infinispan project in which we need to clarify that some modules are not ASL2.

Hmmm, I maintained LGPL but not by choice but because I've addressed this question yet. TBH, I'd like to move it to ASL2, same as the rest of Infinispan code, but I've not looked at the steps to do that. If that's not feasible, I'd leave it as LGPL.

Sanne:
> TBH, I'd like to move it to ASL2, same as the rest of Infinispan code

We could explore that but remember you can't "just do it". Requires several levels of approval and to get permissions in writing from any person who ever contributed code to it.

I don't think that's worth the effort, as there's no strategical benefit in changing it.

galderz:
> We could explore that but remember you can't "just do it". Requires several levels of approval and to get permissions in writing from any person who ever contributed code to it.

Yeah, I'm aware. In this particular case, I think it mostly bogs down to Steve, Radim and myself, but I'd need to look at all the commits.

Let's see what others think about this topic.

galderz:
> Yeah, I'm aware. In this particular case, I think it mostly bogs down to Steve, Radim and myself, but I'd need to look at all the commits.

Actually, there's a few more... So yeah, maybe just leave it as is and mention that this module has LGPL license. @tristantarrant @rvansa thoughts?

Sanne:
Some more thoughts:
- since we're not deleting the original yet, and the package root name is `org.hibernate` changing license might be confusing
- I see no value in changing this, as it's only useful when combined with Hibernate ORM: people having a problem with LGPL are not your target audience.

danberindei:
@galderz You can change tabs to spaces in the whole history with [git-filter-branch](https://stackoverflow.com/a/4130589). IntelliJ seems much better lately about allowing different code styles without excessive configuration, so I'm not sure it's worth it, but this is our only chance to rewrite all the commits.

Definitely -1 for changing tabs to spaces in a separate commit: even if `git blame -w` is supposed to ignore whitespace changes, I'm not sure how well IDEs support it (I couldn't find the option in IntelliJ).

As to the license, I would favor ASL2 so that users feel free-er to go in and modifying the code, and they can think about contributing their changes back later. But I haven't written any line of code here, so it's not like I have a vote :)

Sanne:
@danberindei good idea. While at it could you also include the fixes while filtering the branch so the code looks better?

More seriously: I'd refrain from changing formatting. For a while still you'll be maintaining both repositories, if you then have to backport any patch to the Hibernate repository, it's going to be complex enough.

hibernate-cache/src/test/java/org/hibernate/test/cache/infinispan/functional/TombstoneTest.java
```diff
@@ -321,7 +321,7 @@ public void testEvictPutFromLoadDuringUpdate() throws Exception {
 
    private Future<?> blockedPutFromLoad(CyclicBarrier putFromLoadBarrier) throws InterruptedException, BrokenBarrierException, TimeoutException {
       BlockingInterceptor blockingInterceptor = new BlockingInterceptor(putFromLoadBarrier, PutKeyValueCommand.class, false, true);
```
rvansa: the blocking interceptor should be async-ified, too

galderz:
> since we're not deleting the original yet, and the package root name is org.hibernate, changing license might be confusing

^ True. I forgot to mention that my plan is to change the package from `org.hibernate.cache.infinispan` to `org.infinispan.hibernate.cache`. I'll add it to the TODOs above.

> I see no value in changing this, as this cache integration is only useful when combined with Hibernate ORM: people having a problem with LGPL are not your target audience.

^ V true. Since this is used with ORM and ORM is LGPL, no point in changing to ASL2.

hibernate-cache/src/test/java/org/hibernate/test/cache/infinispan/util/ExpectingInterceptor.java
```diff
@@ -57,39 +58,36 @@ private static String source() {
...
-               log.tracef("Testing condition %s", condition);
-               if (condition.success != null && condition.success != succeeded) {
-                  log.trace("Condition test failed, succeeded: " + succeeded);
-               } else if (condition.predicate.test(ctx, command)) {
-                  assert condition.action != null;
-                  log.trace("Condition succeeded");
-                  toExecute.add(condition.action);
-                  if (condition.removeCheck == null || condition.removeCheck.getAsBoolean()) {
-                     iterator.remove();
+      return invokeNextAndFinally( ctx, command, new InvocationFinallyAction() {
```
rvansa: move handler to method please

hibernate-cache/src/test/java/org/hibernate/test/cache/infinispan/stress/entities/Address.java
```diff
@@ -1,6 +1,7 @@
 <<<<<<< HEAD
 <<<<<<< HEAD
 <<<<<<< HEAD
+<<<<<<< HEAD
```
rvansa: this looks like a failed conflict resolution

galderz:
> More seriously: I'd refrain from changing formatting. For a while still you'll be maintaining both repositories, if you then have to backport any patch to the Hibernate repository, it's going to be complex enough.

^ Another good point. I'm with Sanne on avoiding formatting changes. Eventually code changes and we could move it slowly, as needed to spaces and other preferred formats.

rvansa:
@galderz For some reason GH is showing my comments as outdated; please check them anyway.

rvansa:
@galderz @Sanne Any backport *will* be manual, since the interceptors have changed, maybe there will be even difference in Hibernate SPI. Not sure about vanilla git, but I bet IntelliJ can resolve whitespace-only conflict automatically. Even github allows you to add `?w=0` to URL to avoid whitespace conflicts.

I think we could live with `}`s on separate line, but each file has to be formatted consistently. Otherwise it's just mess, and mess in code indicates bugs.

tristantarrant:
My POV:
- no to formatting changes
- no to license changes
- this could land in 9.1 since it isn't really new code

galderz:
> I don't agree with @Sanne, you should change tabs to spaces and generally fix formatting. But that should not be done as a separate commit, but on each commit in the history. And there won't be a better time to do that than now, because any later you'd have to use the separate commit.

Interesting idea, I had not thought of that. If done that way, the formatting would be one with each commit and that'd be fine from a history perspective. I will give it a go and see what I can achieve.

galderz:
One more thing: before trying to fix this formatting stuff, I'd like to make sure the rest of the code looks fine, so I'll first try to address @rvansa's individual comments. As a final step, I'd try to address each commit's formatting.

galderz:
If anyone has any tips/tricks of how to fix the formatting at each commit, I'm all ears :) @rvansa

galderz:
> I don't agree with @Sanne, you should change tabs to spaces and generally fix formatting. But that should not be done as a separate commit, but on each commit in the history. And there won't be a better time to do that than now, because any later you'd have to use the separate commit.

^ @rvansa I've started looking into this and this seems like a huge job. There are 246 commits, so if you got back to the first one and fix formatting there, then commit, you find yourself that every single commit afterwards it's going to have conflicts due to style differences. So, let's say you fix those and reformat, then comes the next commit, which again has conflicts due to style...etc. From my POV, trying to do this just spirals out of proportion for the gains to be had...

One more thing:

> While the IDEs are getting better, if people don't use the same settings, any updates will result in unneeded changes. See for how long have we had trouble just with the order of imports.

We've had problems with imports because people would let their IDEs reorder/optimise the imports. If people would have left them as they are it'd be fine :). This is something configurable. Just don't let the IDE mess up the existing imports. 

Personally, I've had more issues backporting patches to other versions as a result of that blanket change to reorder all imports.

galderz:
Btw, I've noticed that when trying to rebase with master, I didn't fix all the conflicts in the branch, so the code won't compile. I'm fixing it, but I think i'm going to start with a brand new branch. Once it's working I'll post it here and we can continue the discussion.

As it stands, my position on the topics here are:

* No style changes other than code that had to be changed to get compilation and testsuite working.
* License stays as is.

The remaining TODOs are still valid.

@rvansa I will address your individual comments in the new branch.

hibernate-cache/src/main/java/org/hibernate/cache/infinispan/access/LockingInterceptor.java
```diff
@@ -32,43 +34,41 @@
...
-		try {
-			// Clear any metadata; we'll set them as appropriate in TombstoneCallInterceptor
-			command.setMetadata(null);
+   protected final InvocationFinallyAction unlockAllReturnCheckCompletableFutureHandler = new InvocationFinallyAction() {
+      @Override
+      public void accept(InvocationContext rCtx, VisitableCommand rCommand, Object rv, Throwable throwable) throws Throwable {
+         lockManager.unlockAll(rCtx);
+         if (rv instanceof CompletableFuture) {
+            try {
+               ((CompletableFuture) rv).join();
```
galderz: Hmmm, not sure I understand what the code should look like. Could you be more precise?

hibernate-cache/src/main/java/org/hibernate/cache/infinispan/access/TombstoneCallInterceptor.java
```diff
@@ -169,22 +168,23 @@ private Object setValue(MVCCEntry e, Object value) {
...
-	private void setFailed(PutKeyValueCommand command) {
+	private Object setFailed(InvocationContext ctx, PutKeyValueCommand command) {
 		// This sets command to be unsuccessful, since we don't want to replicate it to backup owner
 		command.setValueMatcher(ValueMatcher.MATCH_NEVER);
-		try {
-			command.perform(null);
-		}
-		catch (Throwable ignored) {
-		}
+      return invokeNextAndExceptionally( ctx, command, (rCtx, rCommand, throwable) -> {
```
galderz: Hmmm, so?

hibernate-cache/src/main/java/org/hibernate/cache/infinispan/access/TxInvalidationInterceptor.java
```diff
@@ -119,15 +127,17 @@ public Object visitLockControlCommand(TxInvocationContext ctx, LockControlComman
...
-		Object retval = invokeNextInterceptor( ctx, command );
-		if ( command.isSuccessful() && !ctx.isInTxScope() ) {
-			if ( keys != null && keys.length != 0 ) {
-				if ( !isLocalModeForced( command ) ) {
-					invalidateAcrossCluster( isSynchronous( command ), keys, ctx );
-				}
-			}
-		}
-		return retval;
+      return invokeNextAndHandle( ctx, command, (rCtx, rCommand, rv, throwable) -> {
```
galderz: I'm not sure you can do this one, since the handler depends on the `keys` parameter. If it's capturing an external variable, a new lambda needs to be created each time.

hibernate-cache/src/main/java/org/hibernate/cache/infinispan/util/LockOwner.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.ObjectOutput;
+
+import org.infinispan.commands.CommandInvocationId;
+
+final class LockOwner {
+
+	private LockOwner() {
+	}
+
+	static void writeTo(ObjectOutput out, Object lockOwner) throws IOException {
```
galderz: That comes from the command's key lock owner, which normally is the command invocation id.

hibernate-cache/src/main/resources/org/hibernate/cache/infinispan/builder/infinispan-configs-local.xml
```diff
@@ -17,20 +17,22 @@
 
       <!-- Default configuration is appropriate for entity/collection caching. -->
       <local-cache-configuration name="entity" simple-cache="true" statistics="false" statistics-available="false">
+         <locking isolation="READ_COMMITTED"/>
```
galderz: True.

However, there's a problem with `QueryRegionImplTest`. That test runs with transactional caches and the default isolation level is the RR. When running with RR, the test breaks. That's why I forced RC in the configurations.

What would be the correct way to solve this? I went for the option of setting isolation to RC everywhere, but really it was only needed for `replicated-query` cache.

Maybe the test should run in non-transactional mode?

hibernate-cache/src/test/java/org/hibernate/test/cache/infinispan/functional/TombstoneTest.java
galderz: It's already been async-ified. Can you be more specific?

hibernate-cache/src/test/java/org/hibernate/test/cache/infinispan/stress/entities/Address.java
galderz: Yeah, I realised I had not fix this stuff. I've created a new branch where I've done this properly and hence this is fixed there.

galderz:
@rvansa All comments marked as 👍 have been fixed in `t_2lc_v4` branch that I've kept separately. Let's try to clarify the remaining comments and I can open a new PR.

galderz:
Closing this. See #5268

hibernate-cache/src/main/java/org/hibernate/cache/infinispan/access/LockingInterceptor.java
rvansa:
```
if (rv instanceof CompletableFuture) {
   return asyncValue((CompletableFuture) rv));
} else {
   return rv;
}
```

hibernate-cache/src/main/java/org/hibernate/cache/infinispan/access/TombstoneCallInterceptor.java
rvansa: Therefore you should remove method `setFailed` completely and just call `command.fail()` on the places where you call it.

hibernate-cache/src/main/resources/org/hibernate/cache/infinispan/builder/infinispan-configs-local.xml
rvansa: I see that the query cache is also non-transactional, is the test overriding that? (probably it is running in both variants). In any case, it should not pollute the default configuration.

And breaking a test with RR sounds fishy, have you investigated?

PR #5255, opened 2017-06-30
ISPN-7996 ClusterTopologyManagerTest.testAbruptLeaveAfterGetStatus2 r…

rvansa:
…andom failures

### DIFF
--- a/core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
+++ b/core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
@@ -105,6 +105,10 @@ public class ScatteredStateConsumerImpl extends StateConsumerImpl {
       // and ScatteringInterceptor would check according to the new CH.
       for (int segment = 0; segment < newWriteCh.getNumSegments(); ++segment) {
          if (!newWriteCh.isSegmentLocalToNode(rpcManager.getAddress(), segment)) {
+            // Failed key transfer could move segment to OWNED state concurrently with installing a new topology.
+            // Therefore we cancel the transfers before installing the topology, preventing any unexpected segment
+            // state updates. Cancelling moves the segments to NOT_OWNED state.
+            cancelTransfers(Collections.singleton(segment));
             svm.unregisterSegment(segment);
          }
       }
@@ -161,7 +165,7 @@ public class ScatteredStateConsumerImpl extends StateConsumerImpl {
                log.debug("Failed confirming revoked segments", throwable);
             }
             for (int segment : addedSegments) {
-               svm.notifyKeyTransferFinished(segment, false);
+               svm.notifyKeyTransferFinished(segment, false, false);
             }
             notifyEndOfStateTransferIfNeeded(cacheTopology.getTopologyId(), cacheTopology.getRebalanceId());
          }
@@ -193,7 +197,7 @@ public class ScatteredStateConsumerImpl extends StateConsumerImpl {
       if (!isTransferringKeys) {
          log.trace("No keys in transfer, finishing segments " + segments);
          for (int segment : segments) {
-            svm.notifyKeyTransferFinished(segment, false);
+            svm.notifyKeyTransferFinished(segment, false, false);
          }
          notifyEndOfStateTransferIfNeeded(cacheTopology.getTopologyId(), cacheTopology.getRebalanceId());
       }
@@ -221,8 +225,10 @@ public class ScatteredStateConsumerImpl extends StateConsumerImpl {
                transfers.remove(inboundTransfer);
                if (transfers.isEmpty()) {
                   transfersBySegment.remove(segment);
-                  log.trace("All transfer tasks have completed.");
-                  svm.notifyKeyTransferFinished(segment, inboundTransfer.isCompletedSuccessfully());
+                  if (trace) {
+                     log.tracef("All transfer tasks for segment %d have completed.", segment);
+                  }
+                  svm.notifyKeyTransferFinished(segment, inboundTransfer.isCompletedSuccessfully(), inboundTransfer.isCancelled());
                   switch (completedSegments.size()) {
                      case 0:
                         completedSegments = Collections.singleton(segment);
--- a/core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
+++ b/core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
@@ -253,7 +253,10 @@ public class ScatteredVersionManagerImpl<K> implements ScatteredVersionManager<K
 
    @Override
    public synchronized void unregisterSegment(int segment) {
-      segmentStates.set(segment, SegmentState.NOT_OWNED);
+      SegmentState previous = segmentStates.getAndSet(segment, SegmentState.NOT_OWNED);
+      if (trace) {
+         log.tracef("Unregistered segment %d (previous=%s)", segment, previous);
+      }
       CompletableFuture<Void> blockedFuture = blockedFutures.get(segment);
       if (blockedFuture != null) {
          blockedFuture.completeExceptionally(new CacheException("The segment is no longer owned."));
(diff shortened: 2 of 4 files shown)


rvansa:
@danberindei The test passes, and it seems I haven't broken anything...

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
```diff
@@ -105,6 +105,10 @@ protected void beforeTopologyInstalled(int topologyId, boolean startRebalance, C
       // and ScatteringInterceptor would check according to the new CH.
       for (int segment = 0; segment < newWriteCh.getNumSegments(); ++segment) {
          if (!newWriteCh.isSegmentLocalToNode(rpcManager.getAddress(), segment)) {
+            // We cancel any existing transfers as failed key transfer could move segment
+            // to OWNED state if it fails, concurrently with installing the topology.
```
danberindei: Can a failed key transfer not fail? And is the key transfer the one installing the topology, or is it someone else?

core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
```diff
@@ -267,15 +270,21 @@ public boolean isVersionActual(int segment, EntryVersion version) {
...
+         update = SegmentState.VALUE_TRANSFER;
+      } else if (cancelled) {
+         // The transfer is cancelled when a newer topology is being installed.
+         update = SegmentState.NOT_OWNED;
+      } else {
+         update = SegmentState.OWNED;
+      }
+      // It is possible that the segment is not in KEY_TRANSFER state, but can be in BLOCKED states as well
+      // when the CONFIRM_REVOKED_SEGMENTS failed.
+      // Another possibility is when there is already another rebalance starting, then it might be NOT_OWNED
```
danberindei: Is this still possible? If yes, can we make it impossible, so that a key transfer for rebalance X always finishes before we start rebalance `X+1`?

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
rvansa: If the key transfer fails, we assume that the node has crashed. This node then has to take the ownership, there's not much else to do when we've lost data.
If the key transfer is cancelled, it is because there is a newer topology installed and we don't seize the ownership, it's up to the new topology to select who will be the next owner.

Key transfer is not installing a topology. When a rebalance starts, all new segments are marked as `BLOCKED` and we wait until the previous owner confirms that it has revoked them. Then the key transfer starts (followed by value transfer) and when everything is done, we confirm the topology calling `StateTransferManager.notifyEndOfStateTransfer`.

core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
rvansa: When two nodes crash in sequence, the other rebalance can be sent out before all nodes confirm the previous rebalance. If both crashed were coords (at their time), you can imagine that keeping track of who finished would be problematic - that's why we support canceling the state transfer even in dist/repl caches.

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
danberindei: I meant if it's already a "failed key transfer", there's no question of "if it fails". Also, the phrasing makes it sound as if the key transfer installs a topology concurrently with changing the segment's state... it would be better to explain in the comment what could cause a concurrent topology update and why we need to prevent concurrent updates of the segment state.

core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
danberindei: If I understand the problem correctly, our solution for dist/repl caches is to use one `LimitedExecutor` per cache so that the new rebalance cannot start until the old topology update was processed and the old inbound transfers were all cancelled. The new coord always sends a topology update before sending the rebalance, so it doesn't matter if it was 2 crashes in a row, crash + join, or join + crash. The only complication is that the topology update and the rebalance start commands are not ordered, so we need to create the fake topology update in `LTMI.resetLocalTopologyBeforeRebalance`.

Now `ScatteredStateConsumerImpl.beforeTopologyInstalled` also cancels old inbound transfers, so I thought you won't have unfinished transfers from a previous rebalance any more.

rvansa: I've already forgot the details, but it seems that `notifyKeyTransferFinished` can be called from `SSCI.onTaskCompletion` which is being executed when the keys transfer was just cancelled, because of the other rebalance starting. This other rebalance has reset the status of the segment. However, the transfer should be cancelled *before* the segment is unregistered... So I don't know if that can still happen in practice.

Anyway, it doesn't hurt because we're just allowing to do a no-op NOT_OWNED -> NOT_OWNED.

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
rvansa: I'll remove the duplicate 'if fails'... KT does not install a topology but it can change segment state. Nothing prevents a new topology to come during KT, and topology changes segment state, too. Therefore in order to synchronize those two events we kill the former one.

core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
rvansa: Can't reproduce the situation...

core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
```diff
@@ -267,15 +270,21 @@ public boolean isVersionActual(int segment, EntryVersion version) {
...
-         previous = segmentStates.get(segment);
-         // It is possible that the segment is not in KEY_TRANSFER state, but can be in BLOCKED states as well
-         // when the CONFIRM_REVOKED_SEGMENTS failed.
-         log.tracef("Finishing transfer for segment %d = %s -> %s", segment, previous, update);
-      } while (!segmentStates.compareAndSet(segment, previous, update));
+   public void notifyKeyTransferFinished(int segment, boolean expectValues, boolean cancelled) {
+      SegmentState update;
+      if (expectValues) {
+         update = SegmentState.VALUE_TRANSFER;
+      } else if (cancelled) {
```
pruivo: shouldn't check `cancelled` before `expectValues`?

rvansa: `expectValues` is `true` only if the transfer was not cancelled, see `ScatteredStateConsumerImpl:231`. Though, I admit that this is a risk for the future.

rvansa:
Rebased, reordered the check (@pruivo) & removed the comment that probably does not apply anymore (@danberindei)

core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
```diff
@@ -267,15 +270,21 @@ public boolean isVersionActual(int segment, EntryVersion version) {
...
+         assert !expectValues;
+      } else if (expectValues) {
+         update = SegmentState.VALUE_TRANSFER;
+      } else {
+         update = SegmentState.OWNED;
+      }
+      // It is possible that the segment is not in KEY_TRANSFER state, but can be in BLOCKED states as well
+      // when the CONFIRM_REVOKED_SEGMENTS failed.
+      SegmentState previous = segmentStates.getAndSet(segment, update);
+      log.tracef("Finished transfer for segment %d = %s -> %s", segment, previous, update);
```
pruivo: `if (trace)` is missing.

danberindei:
Integrated, thanks Radim!

PR #5259, opened 2017-07-03
ISPN-2082 JdbcStringBasedCacheStore: ORA-24816 when storing BLOB valu…

ryanemerson:
…es > 4000 bytes

-- PreparedStatements are now prepared in the TableManager
implementation to allow for greater flexibility for each database
vendor.

https://issues.jboss.org/browse/ISPN-2082


ryanemerson:
This should probably be backported to 9.0.x as well

danberindei:
Integrated in master and 9.0.x, thanks Ryan!

PR #5261, opened 2017-07-04
ISPN-8006 Key/value bodies should use printArray

galderz:
https://issues.jboss.org/browse/ISPN-8006

* To avoid logging too long key/values


galderz:
Please apply it to 9.0.x too

tristantarrant:
Merged to both master and 9.0.x

PR #5262, opened 2017-07-04
ISPN-7224 Synchronous get

slaskawi:
https://issues.jboss.org/browse/ISPN-7224

Here's the missing part of [ISPN-7224](https://issues.jboss.org/browse/ISPN-7224) for synchronous GET methods. In the future we may reimplement `#get(Object, Callable<T>)` method to support different locking strategy but for now it would be an overkill. `StampedLock` seems cheap enough to do the trick.

### DIFF
--- a/spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
+++ b/spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
@@ -1,8 +1,12 @@
 package org.infinispan.spring.provider;
 
+import java.util.Map;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 
+import org.infinispan.commons.api.BasicCache;
 import org.springframework.cache.Cache;
 import org.springframework.cache.support.SimpleValueWrapper;
 import org.springframework.util.Assert;
@@ -15,6 +19,8 @@ import org.springframework.util.Assert;
  * @author Sebastian Laskawiec
  */
 class CacheDelegate implements Cache {
+
+   private final Map<Object, ReentrantLock> synchronousGetLocks = new ConcurrentHashMap<>();
 
    //Implemented as a static holder class for backwards compatibility.
    //Imagine a situation where a client has new integration module and old Spring version. In that case
@@ -31,7 +37,7 @@ class CacheDelegate implements Cache {
    /**
     * @param nativeCache underlying cache
     */
-   public CacheDelegate(final org.infinispan.commons.api.BasicCache<Object, Object> nativeCache) {
+   public CacheDelegate(final BasicCache<Object, Object> nativeCache) {
       Assert.notNull(nativeCache, "A non-null Infinispan cache implementation is required");
       this.nativeCache = nativeCache;
    }
@@ -71,13 +77,31 @@ class CacheDelegate implements Cache {
 
    @Override
    public <T> T get(Object key, Callable<T> valueLoader) {
-      return (T) nativeCache.computeIfAbsent(key, keyToBeInserted -> {
+      ReentrantLock lock = null;
+      T value = (T) nativeCache.get(key);
+      if (value == null) {
+         lock = synchronousGetLocks.computeIfAbsent(key, k -> new ReentrantLock());
+         lock.lock();
          try {
-            return valueLoader.call();
-         } catch (Exception e) {
-            throw ValueRetrievalExceptionResolver.throwValueRetrievalException(key, valueLoader, e);
+            if ((value = (T) nativeCache.get(key)) == null) {
+               try {
+                  T newValue = valueLoader.call();
+                  // we can't use computeIfAbsent here since in distributed embedded scenario we would
+                  // send a lambda to other nodes. This is the behavior we want to avoid.
+                  value = (T) nativeCache.putIfAbsent(key, newValue);
+                  if (value == null) {
+                     value = newValue;
+                  }
+               } catch (Exception e) {
+                  throw ValueRetrievalExceptionResolver.throwValueRetrievalException(key, valueLoader, e);
+               }
+            }
+         } finally {
+            lock.unlock();
+            synchronousGetLocks.remove(key);
          }
-      });
+      }
+      return value;
    }
 
    /**
--- a/spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheTest.java
+++ b/spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheTest.java
@@ -1,9 +1,14 @@
 package org.infinispan.spring.provider;
 
 import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertNotNull;
 import static org.testng.AssertJUnit.assertNull;
 import static org.testng.AssertJUnit.assertSame;
 import static org.testng.AssertJUnit.assertTrue;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.configuration.global.GlobalConfigurationBuilder;
@@ -12,7 +17,6 @@ import org.infinispan.spring.support.embedded.InfinispanNamedEmbeddedCacheFactor
 import org.infinispan.test.SingleCacheManagerTest;
 import org.infinispan.test.fwk.TestCacheManagerFactory;
 import org.springframework.cache.Cache;
-import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
@@ -30,8 +34,8 @@ import org.testng.annotations.Test;
  * @author Marius Bogoevici
  *
  */
-@Test(testName = "spring.provider.SpringCacheCacheTest", groups = {"unit", "smoke"})
-public class SpringCacheCacheTest extends SingleCacheManagerTest {
+@Test(testName = "spring.provider.SpringCacheTest", groups = {"unit", "smoke"})
+public class SpringCacheTest extends SingleCacheManagerTest {
 
    protected final static String CACHE_NAME = "testCache";
 
@@ -53,12 +57,6 @@ public class SpringCacheCacheTest extends SingleCacheManagerTest {
    public void setUp() throws Exception {
       this.nativeCache = createNativeCache();
       this.cache = createCache(this.nativeCache);
-   }
-
-   @AfterMethod
-   public void tearDown() throws Exception {
-      this.nativeCache = null;
-      this.cache = null;
    }
 
    @Test
@@ -210,6 +208,43 @@ public class SpringCacheCacheTest extends SingleCacheManagerTest {
 
       //then
       assertEquals("test", valueFromCache.get());
+   }
+
+   /*
+    * In this test Thread 1 should exclusively block Cache#get method so that Thread 2 won't be able to
+    * insert "thread2" string into the cache.
+    *
+    * The test check this part of the Spring spec:
+    * Return the value to which this cache maps the specified key, obtaining that value from valueLoader if necessary.
+    * This method provides a simple substitute for the conventional "if cached, return; otherwise create, cache and return" pattern.
+    * @see http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/Cache.html#get-java.lang.Object-java.util.concurrent.Callable-
+    */
+   @Test
+   public void testValueLoaderWithLocking() throws Exception {
+      //given
+      CountDownLatch waitUntilThread1LocksValueGetter = new CountDownLatch(1);
+
+      //when
+      Future<String> thread1 = fork(() -> cache.get("test", () -> {
+         waitUntilThread1LocksValueGetter.countDown();
+         return "thread1";
+      }));
+
+      Future<String> thread2 = fork(() -> {
+         waitUntilThread1LocksValueGetter.await(30, TimeUnit.SECONDS);
+         return cache.get("test", () -> "thread2");
+      });
+
+      String valueObtainedByThread1 = thread1.get();
+      String valueObtainedByThread2 = thread2.get();
+
+      Cache.ValueWrapper valueAfterGetterIsDone = cache.get("test");
+
+      //then
+      assertNotNull(valueAfterGetterIsDone);
+      assertEquals("thread1", valueAfterGetterIsDone.get());
+      assertEquals("thread1", valueObtainedByThread1);
+      assertEquals("thread1", valueObtainedByThread2);
    }
 
    @Test
--- a/spring/spring4/spring4-remote/src/test/java/org/infinispan/spring/provider/SpringRemoteCacheTest.java
+++ b/spring/spring4/spring4-remote/src/test/java/org/infinispan/spring/provider/SpringRemoteCacheTest.java
@@ -1,0 +1,93 @@
+package org.infinispan.spring.provider;
+
+import static org.infinispan.server.hotrod.test.HotRodTestingUtil.hotRodCacheConfiguration;
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertNotNull;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Future;
+
+import org.infinispan.client.hotrod.RemoteCacheManager;
+import org.infinispan.client.hotrod.configuration.ConfigurationBuilder;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.server.hotrod.HotRodServer;
+import org.infinispan.server.hotrod.test.HotRodTestingUtil;
+import org.infinispan.test.SingleCacheManagerTest;
+import org.infinispan.test.fwk.TestCacheManagerFactory;
+import org.springframework.cache.Cache;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+@Test(testName = "spring.provider.SpringRemoteCacheTest", groups = "functional")
+public class SpringRemoteCacheTest extends SingleCacheManagerTest {
+
+   private static final String TEST_CACHE_NAME = "spring.remote.cache.Test";
+
+   private RemoteCacheManager remoteCacheManager;
+   private HotRodServer hotrodServer;
+
+   @Override
+   protected EmbeddedCacheManager createCacheManager() throws Exception {
+      cacheManager = TestCacheManagerFactory.createCacheManager(hotRodCacheConfiguration());
+      cacheManager.defineConfiguration(TEST_CACHE_NAME, cacheManager.getDefaultCacheConfiguration());
+      cache = cacheManager.getCache(TEST_CACHE_NAME);
+
+      return cacheManager;
+   }
+
+   @BeforeClass
+   public void setupRemoteCacheFactory() {
+      hotrodServer = HotRodTestingUtil.startHotRodServer(cacheManager, 0);
+      ConfigurationBuilder builder = new ConfigurationBuilder();
+      builder.addServer().host("localhost").port(hotrodServer.getPort());
+      remoteCacheManager = new RemoteCacheManager(builder.build());
+   }
+
+   @AfterClass
+   public void destroyRemoteCacheFactory() {
+      remoteCacheManager.stop();
+      hotrodServer.stop();
+   }
+
+   /*
+    * In this test Thread 1 should exclusively block Cache#get method so that Thread 2 won't be able to
+    * insert "thread2" string into the cache.
+    *
+    * The test check this part of the Spring spec:
+    * Return the value to which this cache maps the specified key, obtaining that value from valueLoader if necessary.
+    * This method provides a simple substitute for the conventional "if cached, return; otherwise create, cache and return" pattern.
+    * @see http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/Cache.html#get-java.lang.Object-java.util.concurrent.Callable-
+    */
+   @Test(timeOut = 30_000)
+   public void testValueLoaderWithLocking() throws Exception {
+      //given
+      final SpringRemoteCacheManager springRemoteCacheManager = new SpringRemoteCacheManager(remoteCacheManager);
+      final SpringCache cache = springRemoteCacheManager.getCache(TEST_CACHE_NAME);
+
+      CountDownLatch waitUntilThread1LocksValueGetter = new CountDownLatch(1);
+
+      //when
+      Future<String> thread1 = fork(() -> cache.get("test", () -> {
+         waitUntilThread1LocksValueGetter.countDown();
+//         /TimeUnit.MILLISECONDS.sleep(10);
+         return "thread1";
+      }));
+
+      Future<String> thread2 = fork(() -> {
+         waitUntilThread1LocksValueGetter.await();
+         return cache.get("test", () -> "thread2");
+      });
+
+      String valueObtainedByThread1 = thread1.get();
+      String valueObtainedByThread2 = thread2.get();
+
+      Cache.ValueWrapper valueAfterGetterIsDone = cache.get("test");
+
+      //then
+      assertNotNull(valueAfterGetterIsDone);
+      assertEquals("thread1", valueAfterGetterIsDone.get());
+      assertEquals("thread1", valueObtainedByThread1);
+      assertEquals("thread1", valueObtainedByThread2);
+   }
+}


spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
```diff
@@ -71,13 +76,33 @@ public ValueWrapper get(final Object key) {
...
-      return (T) nativeCache.computeIfAbsent(key, keyToBeInserted -> {
-         try {
-            return valueLoader.call();
-         } catch (Exception e) {
-            throw ValueRetrievalExceptionResolver.throwValueRetrievalException(key, valueLoader, e);
+      long stamp = synchronousGetLock.readLock();
+      try {
+         T value = (T) nativeCache.get(key);
+         if (value == null) {
+            AtomicBoolean valueObtained = new AtomicBoolean();
```
rvansa: `ByRef.Boolean`

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
```diff
@@ -212,6 +209,51 @@ public void testValueLoaderWithNoPreviousValue() {
...
+         try {
+            waitUntilThread1LocksValueGetter.await();
+         } catch (InterruptedException e) {
+            throw new AssertionError(e);
+         }
+         cache.get("test", () -> "thread2");
+         waitUntilBothThreadsFinish.countDown();
+      });
+
+      waitUntilBothThreadsFinish.await();
```
rvansa: This does not check correctly against exceptions, just call `.get()` on the future that `fork()` gives you

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
```diff
@@ -212,6 +209,51 @@ public void testValueLoaderWithNoPreviousValue() {
...
+      });
+
+      fork(() -> {
+         Thread.currentThread().setName("Thread2");
+         try {
+            waitUntilThread1LocksValueGetter.await();
+         } catch (InterruptedException e) {
+            throw new AssertionError(e);
+         }
+         cache.get("test", () -> "thread2");
```
rvansa: `assertEquals("thread1", ...)`

slaskawi:
Yes and no. As for ISPN-7224 [we agreed to implement as a client side locking](https://issues.jboss.org/browse/ISPN-7224?focusedCommentId=13393019&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13393019). For embedded we can use transactions but we will probably need to wait a little bit until Hot Rod gets them (later on we can always replace existing implementation with some locking strategy).

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
slaskawi: The reason I took `AtomicBoolean` is that I wanted to be 100% sure that thread which is spinning inside the `while` loop will observe its change. Otherwise we can not guarantee that the supplier method will be called exactly once (and this is the contract). Can we guarantee this with `ByRef.Boolean`?

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
slaskawi: No, we are not interested here of the output. The supplier matters here.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
slaskawi: hmmm interesting... I'll give it a go. BTW, exceptions are tested in another test - `testValueLoaderWithExceptionWhenLoading`.

rvansa:
@slaskawi When you execute `computeIfAbsent` on a dist cache, there are couple of possible outcomes:
1) valueLoader is non-serializable, at some point it fails attempting to serialize the lambda and fail
2) valueLoader is serializable and it will be actually executed on multiple remote nodes
3) valueLoader is non-serializable but you test on single-node cache, and it will work

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
rvansa: Actually you shouldn't use `ByRef.Boolean`, just a `break`. If I am not mistaken the `valueObtained` does not escape the scope of the current looping thread.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
rvansa: And how do you know that the supplier was not invoked without storing the value, then?

rvansa:
If you want to run something with a clustered cache, please test it on a clustered cache.

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
slaskawi: Agree. I had to done some refresher of "Java concurrently in practice" :)

slaskawi:
@rvansa RT `computeIfAbsent` - I think I just need to avoid sending lambdas.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
slaskawi: The problem is that this supplier blocks until the other one finishes. Let me go through the test in steps (perhaps I did some mistake and I just don't understand what you're trying to point me out):

1. Thread 1 and 2 start is undetermined order. They are both supplied to the common executor via `fork` method.
2. Let's assume Thread 2 goes first (just for the sake of explanation).
3. Thread 2 hits `waitUntilThread1LocksValueGetter.await();` and waits (since the latch count is 1).
4. Thread 1 can go now.
5. In supplier method of a Thread 1 unblocks Thread 2 giving it chance to go (`waitUntilThread1LocksValueGetter.countDown();` and `TimeUnit.MILLISECONDS.sleep(10);`).
6. If there was no proper synchronization, the value set in the cache would be "thread2" at this point (set by Thread 2). 
7. But if everything is fine, the implementation will block Thread 2 from progressing because of exclusive lock (at this point Thread 1 already owns write lock and sleeping (sleeping does not release locks)).
8. Thread 1 wakes up and writes the value in cache.
9. Thread 1 releases the lock
10. Thread 2 progresses (since the test is waiting at `thread1.get()` or `thread2.get`).
11. Thread 2 notices that the value is there so it doesn't write anything to the cache.
12. `Cache.ValueWrapper valueAfterGetterIsDone = cache.get("test");` obtains the value which should be `thread1` (since Thread 1 have done the write).

Note that I tried to write this test in pretty generic way. So that I could refactor it in the future.

Does this make sense? Or I made it even worse? :D

danberindei: @slaskawi The question is what do you expect `cache.get("test", () -> "thread2")` to return. Should it return `thread1`, `thread2`, or it could be any of them?

If it could be any of them, then the test is fine, but if it should be `thread1`, then you need to assert that. Personally, I'd return the result of `cache.get(...)` from both `fork` lambdas, and then `assertEquals("thread1", thread1.get()); assertEquals("thread1", thread2.get());`

And if the implementation guarantees is that the lambda in thread 2 is never executed, then you should throw a `TestException` instead of returning `thread2`.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
```diff
@@ -212,6 +210,50 @@ public void testValueLoaderWithNoPreviousValue() {
...
+            return "thread1";
+         });
+      });
+
+      Future<?> thread2 = fork(() -> {
+         Thread.currentThread().setName("Thread2");
+         try {
+            waitUntilThread1LocksValueGetter.await();
+         } catch (InterruptedException e) {
+            throw new AssertionError(e);
```
danberindei: It's usually simpler to use the `Callable<Void>` overload, then you don't need to catch exceptions in the lambda.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
```diff
@@ -212,6 +210,50 @@ public void testValueLoaderWithNoPreviousValue() {
...
+    * @see http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/Cache.html#get-java.lang.Object-java.util.concurrent.Callable-
+    */
+   @Test(timeOut = 30_000)
+   public void testValueLoaderWithLocking() throws Exception {
+      //given
+      CountDownLatch waitUntilThread1LocksValueGetter = new CountDownLatch(1);
+
+      //when
+      Future<?> thread1 = fork(() -> {
+         Thread.currentThread().setName("Thread1");
```
danberindei: The existing thread name is `ForkThread-1,SpringCacheCacheTest`, no need to set it again.

spring/spring4/spring4-remote/src/test/java/org/infinispan/spring/provider/SpringRemoteCacheManagerTest.java
```diff
@@ -159,4 +161,51 @@ public final void getNativeCacheShouldReturnTheRemoteCacheManagerSuppliedAtConst
...
+    * In this test Thread 1 should exclusively block Cache#get method so that Thread 2 won't be able to
+    * insert "thread2" string into the cache.
+    *
+    * The test check this part of the Spring spec:
+    * Return the value to which this cache maps the specified key, obtaining that value from valueLoader if necessary.
+    * This method provides a simple substitute for the conventional "if cached, return; otherwise create, cache and return" pattern.
+    * @see http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/Cache.html#get-java.lang.Object-java.util.concurrent.Callable-
+    */
+   @Test(timeOut = 30_000)
+   public void testValueLoaderWithLocking() throws Exception {
```
danberindei: Should a `CacheManagerTest` test cache operations?

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
rvansa: @slaskawi The synchronization might be correct, but after Thread 2 being unblocked it might not see the value `thread1` and run the supplier. The supplier value could not be written for some reason, but it would be returned.

Or, the synchronization could not work and Thread 1 wouldn't see the updated value either, ending up with value `thread 1`.

slaskawi: @rvansa Thread 2 is guaranteed to see the value inserted by Thread 1. Thread 1 acquires read lock and then upgrades it into write lock (by `long upgradedLockStamp = synchronousGetLock.tryConvertToWriteLock(stamp)`). Once the lock upgrade is successful then Thread 1 checks the value in cache again.

The same logic pattern is applied to Thread 2. It needs to wait until Thread 1 release the write lock. And operations under the same write lock are guaranteed to be seen. Unless I missed something.

@danberindei Ok, let me add an assert for the second get.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
slaskawi: Done

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
slaskawi: Done

spring/spring4/spring4-remote/src/test/java/org/infinispan/spring/provider/SpringRemoteCacheManagerTest.java
slaskawi: Moved to `SpringRemoteCacheTest`. Thanks for pointing this out.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
rvansa: @slaskawi No, you did not missed anything, that's how it *should* work. But the purpose of any test is to verify that the implementation works how you think it does.

slaskawi: @rvansa I'm a bit confused right now. Could you please tell me what is your request here? Shall I add/remove anything? Maybe you could leave me a code snippet? Or maybe everything is fine and I just didn't get it?

rvansa: The whole discussion started with my request to assert anything that can be asserted, not taking things for granted.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheTest.java
```diff
@@ -212,6 +210,44 @@ public void testValueLoaderWithNoPreviousValue() {
...
+
+      //when
+      Future<String> thread1 = fork(() -> cache.get("test", () -> {
+         waitUntilThread1LocksValueGetter.countDown();
+         TimeUnit.MILLISECONDS.sleep(10);
+         return "thread1";
+      }));
+
+      Future<String> thread2 = fork(() -> {
+         waitUntilThread1LocksValueGetter.await();
```
danberindei: One more thing @slaskawi, please use a timeout here (e.g. 10 seconds).

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheCacheTest.java
danberindei: @rvansa @slaskawi I think this is properly covered in the latest version.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheTest.java
slaskawi: I used the timeout on `@Test` annotation on this method. Do we also need additional timeout here?

rvansa: @slaskawi that timeout applies only to the main thread (it gets interrupted after the timeout), but here you're forking off another thread.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheTest.java
```diff
@@ -212,6 +210,44 @@ public void testValueLoaderWithNoPreviousValue() {
...
+   /*
+    * In this test Thread 1 should exclusively block Cache#get method so that Thread 2 won't be able to
+    * insert "thread2" string into the cache.
+    *
+    * The test check this part of the Spring spec:
+    * Return the value to which this cache maps the specified key, obtaining that value from valueLoader if necessary.
+    * This method provides a simple substitute for the conventional "if cached, return; otherwise create, cache and return" pattern.
+    * @see http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/Cache.html#get-java.lang.Object-java.util.concurrent.Callable-
+    */
+   @Test(timeOut = 30_000, invocationCount = 100)
```
danberindei: @slaskawi shouldn't `invocationCount = 100` be removed?

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheTest.java
danberindei: @rvansa indeed. `AbstractInfinispanTest.killSpawnedThreads()` will eventually kill the thread, but it's better to make tests as explicit as possible.

In fact I avoid `@Test(timeout = ...)` as much as I can, just like `@Test(expectedExceptions = ...)`, because they're too coarse and it's hard for the reader to figure out which of the statements you are expecting to block/throw an exception. The fact that TestNG hasn't yet figured out how to run such test methods in the same thread as their configuration methods (requiring `TestResourceTracker.testThreadStarted()` calls before starting a cache manager) makes me want to use them even less.

tristantarrant:

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
```diff
@@ -71,13 +75,38 @@ public ValueWrapper get(final Object key) {
...
-         try {
-            return valueLoader.call();
-         } catch (Exception e) {
-            throw ValueRetrievalExceptionResolver.throwValueRetrievalException(key, valueLoader, e);
+      long stamp = synchronousGetLock.readLock();
+      try {
+         T value = (T) nativeCache.get(key);
+         if (value == null) {
+            while (true) {
+               long upgradedLockStamp = synchronousGetLock.tryConvertToWriteLock(stamp);
```
danberindei: Is the loop really necessary? The way I see it, either the upgrade succeeds the first time and you get a write lock, or it didn't succeed and `writeLock()` blocks until you get a write lock.

spring/spring4/spring4-embedded/src/test/java/org/infinispan/spring/provider/SpringCacheTest.java
```diff
@@ -212,6 +210,44 @@ public void testValueLoaderWithNoPreviousValue() {
...
+    */
+   @Test
+   public void testValueLoaderWithLocking() throws Exception {
+      //given
+      CountDownLatch waitUntilThread1LocksValueGetter = new CountDownLatch(1);
+
+      //when
+      Future<String> thread1 = fork(() -> cache.get("test", () -> {
+         waitUntilThread1LocksValueGetter.countDown();
+         TimeUnit.MILLISECONDS.sleep(10);
```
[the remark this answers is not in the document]
slaskawi: This wait is unnecessary. I just wanted to give other thread a chance to run but after thinking about this for a while - it doesn't make sense.

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
slaskawi: This works slightly different than that. Here are relevant parts from [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/StampedLock.html):

> ...The state of a StampedLock consists of a version and mode. Lock acquisition methods return a stamp that represents and controls access with respect to a lock state; "try" versions of these methods may instead return the special value zero to represent failure to acquire access...

> ...This class also supports methods that conditionally provide conversions across the three modes. For example, method tryConvertToWriteLock(long) attempts to "upgrade" a mode, returning a valid write stamp if (1) already in writing mode (2) in reading mode and there are no other readers or (3) in optimistic mode and the lock is available...

So effectively if you want to achieve upgrade behavior, you need to use `tryConvertToWriteLock` and pass read lock stamp into that method. With `writeLock` you just get another lock for writing.

BTW, if you look into `else` block, which happens when I can't upgrade the lock, I just obtain a write lock and do stuff in the next iteration.

spring/spring4/spring4-remote/src/test/java/org/infinispan/spring/provider/SpringRemoteCacheTest.java
```diff
@@ -0,0 +1,94 @@
...
+      //given
+      final SpringRemoteCacheManager springRemoteCacheManager = new SpringRemoteCacheManager(remoteCacheManager);
+      final SpringCache cache = springRemoteCacheManager.getCache(TEST_CACHE_NAME);
+
+      CountDownLatch waitUntilThread1LocksValueGetter = new CountDownLatch(1);
+
+      //when
+      Future<String> thread1 = fork(() -> cache.get("test", () -> {
+         waitUntilThread1LocksValueGetter.countDown();
+         TimeUnit.MILLISECONDS.sleep(10);
```
rvansa: The other test does not have this delay; why is it needed here?

rvansa:
@slaskawi Looks almost OK, though you use single stamped lock for all keys; computing value for key `test1` then blocks computing that for `test2`. Do you want to keep it that way, or create a map of stamped locks?

spring/spring4/spring4-remote/src/test/java/org/infinispan/spring/provider/SpringRemoteCacheTest.java
slaskawi: Removed, thanks for noticing!

slaskawi:
@rvansa That's a very good point. I believe we should have a map of stamped locks. Update is coming right up.

galderz:
@slaskawi Did you add any updates since?

slaskawi:
Yes, @rvansa's comments have been addressed. It's ready to be reviewed.

galderz:
@rvansa Can you give this one last look?

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
```diff
@@ -16,6 +20,8 @@
  */
 class CacheDelegate implements Cache {
 
+   private final Map<Object, StampedLock> synchronousGetLocks = new HashMap<>();
```
rvansa: I might be missing something obvious, but this should be a *concurrent* map, shouldn't it?

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
```diff
@@ -71,13 +77,54 @@ public ValueWrapper get(final Object key) {
...
+                     stamp = lock.writeLock();
+                  }
+               }
+            }
+         }
+         return value;
+      } finally {
+         if (lock != null) {
+            lock.unlock(stamp);
+            synchronousGetLocks.remove(key);
```
rvansa: This does not sound safe. Suppose that:
T1 creates and acquires lock
T2 retrieves the lock and blocks
T1 unlocks the lock and removes it
T2 can proceed, enters the critical section
T3 comes and creates another lock, does not get blocked by T2...

slaskawi: > T1 creates and acquires lock

Assuming T1 acquires the Read Lock at line 93.

> T2 retrieves the lock and blocks

T2 successfully upgraded its lock at line 97 and now blocks at line 102.

> T1 unlocks the lock and removes it

Impossible. T1 tries to upgrade its Read Lock to Write Lock (just as a reminder, Write Locks are exclusive) by executing `lock.tryConvertToWriteLock(stamp);` (line 97). This returns `0` since T2 is holding the write lock and sitting at line 102. T1 enters the `else` block and unlocks the Read Lock. Now T1 also blocks but on line 116, it's trying to acquire the Write Lock exclusively but it's waiting till T2 releases it (`writeLock()` is a blocking method).

Unless I misunderstood this, the situation you showed seems to be impossible in reality.

rvansa: No, I meant a situation where:

> T1 acquires read lock, upgrades to write lock on 97
> T2 tries to acquire a read lock and blocks on 93
> T1 unlocks and removes the lock
> T2 unblocks and has read-lock acquired
> T3 comes and does not see the same instance of lock as T2

The T3 can now get the computed value (if it did not expire/be evicted in the meantime) or whatever; the thing is that you have a concept of critical section here and might use different instances of locks in concurrent threads visiting this section.

galderz:
@slaskawi @rvansa 9.2.0.CR1 is being released later today, is this still going in?

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
slaskawi: hmmm good point. I believe we need to hire some visiting thread counting mechanism here. The idea is to remove the lock only if the current thread is the last one visiting this section.

slaskawi:
@galderz Nope, not a chance a get it till CR1

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
danberindei: @slaskawi You're only using the read lock after `cache.get(key)` already returned `null` once, so I think you can skip it and use a `DefaultLockManager` to manage the locking.

tristantarrant:
7 months for a 3-file PR is probably a record...

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
slaskawi: Ok, so how about simplifying this code and synchronizing on the key? This way it will be super easy to read and will probably do what we want. The only downside is that the key we got in an argument might be already locked by the client code. However this is the thing we could just document. 

Looking for up/down vote from @rvansa and @danberindei.

slaskawi: @danberindei This is the implementation of a shared Cache Delegate. It can be used in both remote and embedded scenario. Therefore I can not use `DefaultLockManager`.

slaskawi: @rvansa @danberindei I drastically simplified locking here. Please take another look at it.

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
```diff
@@ -16,6 +20,8 @@
  */
 class CacheDelegate implements Cache {
 
+   private final Map<Object, ReentrantLock> synchronousGetLocks = new HashMap<>();
```
rvansa: not a concurrent/synchronized one

slaskawi: Fixed

tristantarrant:
At last, it's in.

spring/spring4/spring4-common/src/main/java/org/infinispan/spring/provider/CacheDelegate.java
```diff
@@ -71,13 +77,31 @@ public ValueWrapper get(final Object key) {
...
+                  if (value == null) {
+                     value = newValue;
+                  }
+               } catch (Exception e) {
+                  throw ValueRetrievalExceptionResolver.throwValueRetrievalException(key, valueLoader, e);
+               }
+            }
+         } finally {
+            lock.unlock();
+            synchronousGetLocks.remove(key);
```
danberindei: I feel we should have a comment here to explain that removing the lock from `synchronousGetLocks` allows 2 threads to invoke `valueLoader` in parallel, but that's fine because the `get(key, valueLoader)` javadoc allows it.

PR #5266, opened 2017-07-05
ISPN-7936 Add configurable batch-size to AbstractStoreConfiguration.

ryanemerson:
https://issues.jboss.org/browse/ISPN-7936


persistence/jpa/src/main/java/org/infinispan/persistence/jpa/configuration/JpaStoreConfigurationBuilder.java
```diff
@@ -29,8 +30,9 @@ public JpaStoreConfigurationBuilder entityClass(Class<?> entityClass) {
       return self();
    }
 
+   @Deprecated
    public JpaStoreConfigurationBuilder batchSize(long batchSize) {
-      attributes.attribute(BATCH_SIZE).set(batchSize);
+      attributes.attribute(MAX_BATCH_SIZE).set((int) batchSize);
```
ryanemerson: The casting isn't ideal, but at the same time will anybody need a batch > Integer.MAX_VALUE?

wburns: Not normally, but some people may pass in `Long.MAX_VALUE`, which would convert to -1 :(

To be honest a lot of our config code is missing validations. For example we wouldn't want a number 0 or less passed in as well.

However in regards to `> Integer.MAX_VALUE` I would just use `Integer.MAX_VALUE` as the batch size.

ryanemerson: I've updated the code so it handles inputs less/greater than int boundaries. It's worth noting that it's valid for the user to enter a negative max-batch-size as this indicates no batching should occur.

persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java
```diff
@@ -371,16 +371,31 @@ public MarshalledEntry load(Object key) {
     @Override
     public void writeBatch(Iterable<MarshalledEntry<? extends K, ? extends V>> marshalledEntries) {
         try {
+            // Store a list of WriteBatch objects so that we can perform all marshalling before acquiring the semaphore
+            List<WriteBatch> writeBatches = new ArrayList<>();
+            int batchSize = 0;
             WriteBatch writeBatch = new WriteBatch();
-            for (MarshalledEntry entry : marshalledEntries)
+            for (MarshalledEntry entry : marshalledEntries) {
```
wburns: Hrmm I don't really like the way batching is done here. We end up putting _all_ entries into batches, but retain all batches. We should really write each batch as it is filled as we do with the other stores and then clear it.

documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
```diff
@@ -48,6 +48,9 @@ Interfaces `DeltaAware`, `Delta` and `CopyableDeltaAware` have been deprecated.
 
 Any partial updates to an entry should be replaced using the link:../user_guide/user_guide.html#functional_map_api[Functional API].
 
+=== Store Changes
```
wburns: Might as well add a line or 2 to the User Guide as well. Also maybe we should call this `Store Batch Size Changes` or something.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/configuration/TableManipulationConfigurationBuilder.java
```diff
@@ -37,9 +37,9 @@
    }
 
    /**
-    * Repetitive DB operations this are batched according to this parameter. This is an optional parameter, and if it
-    * is not specified it will be defaulted to {@link TableManager#DEFAULT_BATCH_SIZE}.
+    * Deprecated. Please use {@link org.infinispan.configuration.cache.AbstractStoreConfigurationBuilder#maxBatchSize(int)} instead.
     */
+   @Deprecated
    public S batchSize(int batchSize) {
       attributes.attribute(BATCH_SIZE).set(batchSize);
```
ryanemerson: It is exposed via TableManager::getBatchSize() unfortunately

wburns: I meant that the config would directly update `maxBatchSize` and use that when it creates the `TableManager`. There is no reason to store the same config option in 2 places.

ryanemerson:
@wburns Updated

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/configuration/TableManipulationConfigurationBuilder.java
```diff
@@ -37,11 +36,11 @@
...
 
    /**
-    * Repetitive DB operations this are batched according to this parameter. This is an optional parameter, and if it
-    * is not specified it will be defaulted to {@link TableManager#DEFAULT_BATCH_SIZE}.
+    * Deprecated. Please use {@link org.infinispan.configuration.cache.AbstractStoreConfigurationBuilder#maxBatchSize(int)} instead.
     */
+   @Deprecated
    public S batchSize(int batchSize) {
-      attributes.attribute(BATCH_SIZE).set(batchSize);
+      attributes.attribute(AbstractJdbcStoreConfiguration.MAX_BATCH_SIZE).set(batchSize);
```
wburns: I was actually thinking more along something like `builder.batchSize(batchSize)` by invoking the batchSize method on the `StoreConfiguration` parent. This then requires changing the create method though as well. I guess it isn't really worth changing then maybe.

ryanemerson: I don't think this is as easy it seems, as `TableManipulationConfigurationBuilder` utilises it's own AttributeSet, not the one inherited from `AbstractStoreConfigurationBuilder`, so we can't just call `maxBatchSize` on the parent. Furthermore, `TableManipulation` does not extend `AbstractStoreConfiguration`.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/configuration/TableManipulationConfiguration.java
```diff
@@ -101,6 +103,7 @@ public int fetchSize() {
    /**
     * @return the size of batches to process.  Guaranteed to be a power of two.
     */
+   @Deprecated
```
wburns: Please put this in the javadoc as well stating they should be using the one on the StoreConfiguration.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/configuration/TableManipulationConfigurationBuilder.java
wburns: Well you can call it on the parent one :) That is why I was saying we would have to change create method, being that it would take another argument.

The thing I am worrying about is that with this change a user having a custom cache store may not be using the batch size properly as it should now be configured on the store level, but their store is most likely using the TableManipulation one. I added another comment that should help alleviate more of my concern.

persistence/jpa/src/main/java/org/infinispan/persistence/jpa/JpaStore.java
```diff
@@ -62,7 +62,7 @@
 @ConfiguredBy(JpaStoreConfiguration.class)
 public class JpaStore<K, V> implements AdvancedLoadWriteStore<K, V> {
    private static final Log log = LogFactory.getLog(JpaStore.class);
-   private static boolean trace = log.isTraceEnabled();
+   private static final boolean trace = log.isTraceEnabled();
```
wburns: :+1:

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/configuration/TableManipulationConfigurationBuilder.java
```diff
@@ -30,18 +29,20 @@
...
 
    TableManipulationConfigurationBuilder(AbstractJdbcStoreConfigurationBuilder<?, B> builder) {
       super(builder);
       attributes = TableManipulationConfiguration.attributeSet();
    }
 
    /**
-    * Repetitive DB operations this are batched according to this parameter. This is an optional parameter, and if it
-    * is not specified it will be defaulted to {@link TableManager#DEFAULT_BATCH_SIZE}.
+    * Deprecated. Please use {@link org.infinispan.configuration.cache.AbstractStoreConfigurationBuilder#maxBatchSize(int)} instead.
```
wburns: Need to put `@deprecated` at the beginning so it shows up properly in Javadoc.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/configuration/TableManipulationConfigurationBuilder.java
```diff
@@ -155,7 +156,7 @@ AttributeSet attributes() {
 
    @Override
    public TableManipulationConfiguration create() {
-      return new TableManipulationConfiguration(attributes.protect());
+      return new TableManipulationConfiguration(attributes.protect(), batchSize);
```
wburns: Sadly this didn't come out as well I would have hoped, but it seems a little better. Unfortunately due to the disconnectedness of the builders, there seems to be no way to pass the value back. This is fine though, we just need to make sure to remove this in 10.0 :)

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/configuration/TableManipulationConfigurationBuilder.java
wburns: Also I was more meaning this should be on the `TableManager` since that is part of the SPI. Sorry if I added some confusion there.

wburns:
I would like to verify that a run passed on CI recently before I integrate but it seems to be down :-1:

wburns:

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/configuration/TableManipulationConfigurationBuilder.java
```diff
@@ -30,18 +29,20 @@
...
 
    TableManipulationConfigurationBuilder(AbstractJdbcStoreConfigurationBuilder<?, B> builder) {
       super(builder);
       attributes = TableManipulationConfiguration.attributeSet();
    }
 
    /**
-    * Repetitive DB operations this are batched according to this parameter. This is an optional parameter, and if it
-    * is not specified it will be defaulted to {@link TableManager#DEFAULT_BATCH_SIZE}.
+    * @deprecated Please use {@link org.infinispan.configuration.cache.AbstractStoreConfigurationBuilder#maxBatchSize(int)} instead.
```
danberindei: It would have been nice to say `@deprecated Since 9.1, ...`

PR #5267, opened 2017-07-06
ISPN-7599 Jolokia support

slaskawi:
https://issues.jboss.org/browse/ISPN-7599

This PR is a proposed approach for enabling Jolokia in the server. It uses Agent Bond, which provides a single Java agent for Jolokia (OpenShift) and Prometheus (Kubernetes). 

Highlights:

* Agent Bond was put into the modules (it would look weird it had been placed elsewhere).
* Jolokia bootstrap script is placed in a separate file. It may be also reused in domain mode.
* The agent is turned off by default and in order to enable it, one needs to supply `--jmx` parameter.

Questions and open issues:

* [x] Shall we use it in domain mode? Or are we good with DMR?
* [x] Do we want to bootstrap it with additional parameter? Or should we leave it enabled by default (allocates yet another port in the host).
* [x] Windows script needs to be ironed out (but probably *after* we agree whether this approach is something that we want or not).
* [x] In Docker bootstrap script there is a clash with `--jmx` parameter. It just needs to be debugged and fixed.
* [ ] `agent.conf` needs to be altered for Kubernetes/OpenShift. It needs to have the following content: `jolokia=port=8778,protocol=https,caCert=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt,clientPrincipal=cn=system:master-proxy,useSslClientAuthentication=true,extraClientCheck=true,host=0.0.0.0,discoveryEnabled=false`. The question where do we want to do the substitution? In a separate conf file? In Docker image? Or maybe put this on users (they can use ConfigMap to override this file on their own).
* [ ] With the above point in mind, the Management Console should operate also on https.

//cc @vblagoje @ryanemerson @tristantarrant @gustavonalle @galderz

### DIFF
--- a/documentation/src/main/asciidoc/server_guide/management.adoc
+++ b/documentation/src/main/asciidoc/server_guide/management.adoc
@@ -65,3 +65,10 @@ In domain mode the port will be the port of the Remoting connector for the Serve
 </subsystem>
 
 ----
+
+=== Jolokia
+
+You can also expose JMX beans using link:https://jolokia.org/[Jolokia]. In order to do this, just run the server with
+additional parameter `--jmx`, for example: `./standalone.xml -c cloud.xml --jmx`. Jolokia configuration is stored in
+`agent.conf` file. It is possible to override this file by specifying it after `--jmx` parameter. For example:
+`./standalone.sh -c cloud.xml --jmx my-config.properties`.
--- a/server/integration/feature-pack/pom.xml
+++ b/server/integration/feature-pack/pom.xml
@@ -629,6 +629,17 @@
             <artifactId>oauth</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>agent-bond-agent</artifactId>
+            <version>${version.agent-bond}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
     </dependencies>
 
     <build>
--- a/server/integration/feature-pack/src/main/resources/content/bin/agent.conf
+++ b/server/integration/feature-pack/src/main/resources/content/bin/agent.conf
@@ -1,0 +1,1 @@
+jolokia=port=8778
--- a/server/integration/feature-pack/src/main/resources/content/bin/jolokia.bat
+++ b/server/integration/feature-pack/src/main/resources/content/bin/jolokia.bat
@@ -1,0 +1,5 @@
+set AGENT_BOND_JAR=
+for /r %JBOSS_HOME%\modules\system\layers\base\io\fabric8\agent-bond\main %%X in (*.jar) do (
+  set AGENT_BOND_JAR=%%X
+)
+set JAVA_OPTS=%JAVA_OPTS% -javaagent:%AGENT_BOND_JAR%=%1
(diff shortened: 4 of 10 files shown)


vblagoje:
Very cool @slaskawi Looking into details!

ryanemerson:
@slaskawi LGTM. A few points/questions:

- What's the plan for Jolokia and embedded? As we're modifying the console to utilise Jolokia, enabling it in embedded will enable the console to work in this context with no additional effort (AFAIK). 
- Regarding agent.conf, should we just allow an optional file parameter after `--jmx`? If no file is specified then just utilise the default agent.conf.
- If possible we should output something to the server log detailing stating what port Jolokia is operating on. Currently we have "Http management interface listening on ...", it would be nice to have the same for Jolokia (especially when the user hasn't specified their own agent.conf file).

slaskawi:
@ryanemerson Thank Ryan. Answering questions:

> What's the plan for Jolokia and embedded? As we're modifying the console to utilise Jolokia, enabling it in embedded will enable the console to work in this context with no additional effort (AFAIK).

This one is very tricky since we don't control the deployment. Users can use Infinispan embedded in Vert.x, Spring Boot, WF or any other container. Each of them might require different types of actions. Just to give you some examples - in Spring Boot one might use Jolokia Java agent (or Agent Bond) whereas on WF one might use Jolokia deployed as WAR.

I honestly don't have any good idea how to solve this apart from documenting it. Maybe @tristantarrant or @galderz will chime in with some good idea.

> Regarding agent.conf, should we just allow an optional file parameter after --jmx? If no file is specified then just utilise the default agent.conf.

I didn't mention it but it's already done this way. Just type './standalone.sh --jmx my-config.properties' and Agent Bond will pick it up.

> If possible we should output something to the server log detailing stating what port Jolokia is operating on. Currently we have "Http management interface listening on ...", it would be nice to have the same for Jolokia (especially when the user hasn't specified their own agent.conf file).

This one is also tricky. Jolokia prints a small information by itself:

```
13:15:59,935 INFO  [org.jboss.modules] (main) JBoss Modules version 1.5.2.Final
I> No access restrictor found, access to any MBean is allowed
Jolokia: Agent started with URL http://127.0.0.1:8778/jolokia/
13:16:00,066 INFO  [org.jboss.msc] (main) JBoss MSC version 1.2.6.Final
...
```

But I'm not sure how to access Java agent from server modules. I will need to ask some WF folks whether it's possible.

slaskawi:
Answering my own questions:

* We should bootstrap it separately since any Java Agent might affect performance in some way
* I didn't enable Jolokia for Domain Mode. I think we should leave it in DMR mode. Of course we might change our mind in any period of time.

vblagoje:
I looked through the PR and I agree with your proposal @slaskawi . I also agree that we should not have Jolokia enabled for domain mode because functionality provided by domain controller can not be easily replicated in some JMX Bean or wrapper and we'll only confuse everyone who is using Infinispan in a particular deployment setup. Let's do one thing at a time and leave domain mode to DMR for now.

vblagoje:
Integrated with 9e65d83 and ac25f43 Thanks @slaskawi

tristantarrant:
I did not want this merged in 9.1. I'll revert it.

vblagoje:
Ok @tristantarrant cc @slaskawi

slaskawi:
@tristantarrant Why not? I would really appreciate it from OpenShift side of things. With this I could enable Jolokia stuff directly from OSE console. It's pretty nice, have a look: https://developers.redhat.com/blog/2016/03/30/jolokia-jvm-monitoring-in-openshift/

tristantarrant:
Let us not introduce new dependencies/features in the last few days before the release. Additionally this needs amending of the license.xml file.

slaskawi:
@tristantarrant Ok, thanks.

tristantarrant:

PR #5269, opened 2017-07-06
ISPN-1149 Allow the protocol servers to choose their own port

tristantarrant:
https://issues.jboss.org/browse/ISPN-1149

### DIFF
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/test/HotRodClientTestingUtil.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/test/HotRodClientTestingUtil.java
@@ -48,7 +48,7 @@ public class HotRodClientTestingUtil {
    private static final int DEFAULT_PORT = 15232;
 
    public static HotRodServer startHotRodServer(EmbeddedCacheManager cacheManager, HotRodServerConfigurationBuilder builder) {
-      return startHotRodServer(cacheManager, findFreePort(), builder);
+      return startHotRodServer(cacheManager, 0, builder);
    }
 
    private static boolean isBindException(Throwable e) {
--- a/server/core/src/main/java/org/infinispan/server/core/AbstractProtocolServer.java
+++ b/server/core/src/main/java/org/infinispan/server/core/AbstractProtocolServer.java
@@ -1,6 +1,8 @@
 package org.infinispan.server.core;
 
+import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.net.ServerSocket;
 
 import javax.management.DynamicMBean;
 import javax.management.MBeanServer;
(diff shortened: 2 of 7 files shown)


client/hotrod-client/src/test/java/org/infinispan/client/hotrod/test/HotRodClientTestingUtil.java
```diff
@@ -48,7 +48,7 @@
    private static final int DEFAULT_PORT = 15232;
 
    public static HotRodServer startHotRodServer(EmbeddedCacheManager cacheManager, HotRodServerConfigurationBuilder builder) {
-      return startHotRodServer(cacheManager, findFreePort(), builder);
```
slaskawi: Perhaps we can remove `#findFreePort` method completely?

gustavonalle: I'd rather keep it; it is often needed for tests

server/core/src/main/java/org/infinispan/server/core/AbstractProtocolServer.java
```diff
@@ -83,7 +85,15 @@ public final void start(A configuration, EmbeddedCacheManager cacheManager) {
    }
 
    protected void startTransport() {
-      InetSocketAddress address = new InetSocketAddress(configuration.host(), configuration.port());
+      int port = configuration.port();
```
pruivo: I don't think that this code is needed. From the `InetSocketAddress` javadoc:
```
 A port number of zero will let the system pick up an ephemeral port in a bind operation
```
https://docs.oracle.com/javase/8/docs/api/java/net/InetSocketAddress.html#InetSocketAddress-java.net.InetAddress-int-

server/hotrod/src/main/java/org/infinispan/server/hotrod/configuration/HotRodServerConfigurationBuilder.java
```diff
@@ -125,6 +125,9 @@ public void validate() {
       if (proxyHost == null && host == null) {
          throw log.missingHostAddress();
       }
+      if (proxyPort < 0 && port == 0 && !startTransport) {
```
pruivo: why are you checking the `proxyPort`? I don't see why you can't have a `proxyPort` with a "ephemeral port".

tristantarrant: Right, I need to think about this a bit more

danberindei: @pruivo The (reverse) proxy has to know about the server's port in order to forward requests, I'm not sure how that would work if the server port changes on each restart.

tristantarrant:

PR #5271, opened 2017-07-07
ISPN-7993 Encoding support on functional maps

karesti:
https://issues.jboss.org/browse/ISPN-7993

Can be reviewed, but :
* @gustavonalle is working on a refactoring that highly impacts this PR (where I grouped the EncodingClasses in a single object, and he is doing a similar refactoring)


karesti:
failing tests seem unrelated

commons/src/main/java/org/infinispan/commons/dataconversion/EncodingUtils.java
```diff
@@ -42,4 +42,10 @@ public static Object toStorage(Object toStore, Encoder encoder, Wrapper wrapper)
       return wrapper.wrap(encoder.toStorage(toStore));
    }
 
+   public static boolean isWrapped(Object obj, Wrapper wrapper) {
+      if (wrapper == null) {
```
rvansa: `Objects.requireNonNull`?

core/src/main/java/org/infinispan/cache/impl/CacheEncoders.java
```diff
@@ -0,0 +1,77 @@
...
+import org.infinispan.commons.dataconversion.EncodingUtils;
+import org.infinispan.commons.dataconversion.Wrapper;
+import org.infinispan.marshall.core.EncoderRegistry;
+
+/**
+ * @since 9.2
+ */
+public class CacheEncoders {
+
+   private Encoder keyEncoder;
```
rvansa: These fields should be final, and then you could make them public, instead of using getters.

core/src/main/java/org/infinispan/functional/impl/MapFactory.java
```diff
@@ -0,0 +1,33 @@
+package org.infinispan.functional.impl;
+
+import org.infinispan.functional.FunctionalMap;
+
+/**
+ * Created by katiaaresti on 29/06/17.
+ */
+public class MapFactory {
```
rvansa: Hmm, this means that encoding won't work with existing code... I would rather see ReadOnlyMapImpl et all aware of Encoded* classes, maybe even get the encoder functionality directly in the ReadOnlyMapImpl et all. There's huge code duplication anyway.

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -182,7 +183,7 @@ public Object visitReadWriteKeyCommand(InvocationContext ctx, ReadWriteKeyComman
 
    @Override
    public Object visitWriteOnlyManyEntriesCommand(InvocationContext ctx, WriteOnlyManyEntriesCommand command) throws Throwable {
-      return handleTxWriteManyEntriesCommand(ctx, command, command.getEntries(), (c, entries) -> new WriteOnlyManyEntriesCommand(c).withEntries(entries));
+      return handleTxWriteManyEntriesCommand(ctx, command, command.getEntries(), (c, entries) -> new WriteOnlyManyEntriesCommand(c, componentRegistry).withEntries(entries));
```
rvansa: You've made the lambda capture `this` reference; please store the lambda in a field. Ofc if you don't remove the `componentRegistry` param completely, see my comment on ReadOnlyManyCommand

core/src/main/java/org/infinispan/commands/functional/ReadOnlyManyCommand.java
```diff
@@ -9,38 +9,67 @@
...
+      this.encodingClasses = encodingClasses;
       this.setFlagsBitSet(params.toFlagsBitSet());
+      init(componentRegistry);
    }
 
    public ReadOnlyManyCommand() {
    }
 
-   public ReadOnlyManyCommand(ReadOnlyManyCommand c) {
+   public ReadOnlyManyCommand(ReadOnlyManyCommand c, ComponentRegistry componentRegistry) {
```
rvansa: This is a copy-constructor, so it shouldn't need to re-wire anything. Is there a problem with `this.cacheEncoders = c.cacheEncoders`?

core/src/main/java/org/infinispan/commands/functional/ReadOnlyManyCommand.java
```diff
@@ -9,38 +9,67 @@
...
       this.setFlagsBitSet(c.getFlagsBitSet());
+      this.encodingClasses = c.encodingClasses;
+      init(componentRegistry);
+   }
+
+   public void init(ComponentRegistry componentRegistry) {
+      if (encodingClasses != null) {
+         componentRegistry.wireDependencies(this);
+      }
+      if (encodingClasses != null) {
```
rvansa: wiring twice?

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
```diff
@@ -28,7 +28,7 @@
  * Lifecycle - <tt>EmbeddedCacheManager</tt>s have a lifecycle (it implements {@link Lifecycle}) and
  * the default constructors also call {@link #start()}.  Overloaded versions of the constructors are available, that do
  * not start the <tt>CacheManager</tt>, although it must be kept in mind that <tt>CacheManager</tt>s need to be started
- * before they can be used to create <tt>Cache</tt> instances.
+ * before they can be used to readWriteMap <tt>Cache</tt> instances.
```
rvansa: auto refactoring?

core/src/test/java/org/infinispan/functional/FunctionalEncoderTest.java
```diff
@@ -0,0 +1,29 @@
+package org.infinispan.functional;
+
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.configuration.cache.StorageType;
+import org.testng.annotations.Test;
+
+@Test(groups = "functional", testName = "functional.FunctionalEncoderTest")
+public class FunctionalEncoderTest extends FunctionalMapTest {
```
rvansa: I think that we need a test using the `AdvancedCache.withEncoders`, too. I am not convinced that all the encodings/decodings happen at the right places. You should also test transactional case (as you've missed encoding on Mutations), so basing this just on FunctionalMapTest is not sufficient.
E.g. I wonder why the testsuite did not catch the `Ids.BI_FUNCTION_MAPPER`, could you check?

core/src/main/java/org/infinispan/commands/functional/ReadOnlyKeyCommand.java
```diff
@@ -42,6 +66,7 @@ public void writeTo(ObjectOutput output) throws IOException {
       output.writeObject(key);
       output.writeObject(f);
       Params.writeObject(output, params);
+      output.writeObject(encodingClasses);
```
gustavonalle: @rvansa This will be addressed in https://issues.jboss.org/browse/ISPN-8004.

gustavonalle: Both wrappers serialization could be encoded in a single byte; each encoder would be another byte, so a total overhead of 3 bytes would be added per command

core/src/main/java/org/infinispan/commands/functional/AbstractWriteKeyCommand.java
```diff
@@ -2,26 +2,41 @@
...
+import org.infinispan.factories.ComponentRegistry;
 import org.infinispan.functional.impl.Params;
 
 public abstract class AbstractWriteKeyCommand<K, V> extends AbstractDataWriteCommand implements FunctionalCommand<K, V> {
 
    Params params;
    ValueMatcher valueMatcher;
    boolean successful = true;
+   EncodingClasses encodingClasses;
+   CacheEncoders cacheEncoders = new CacheEncoders();
```
karesti: empty encoders handle the code where no encoding exist and so Entry code is the same with and without encoding. This can be discussed from a performance/memory point of view and try to make it better !

core/src/main/java/org/infinispan/commands/functional/ReadOnlyKeyCommand.java
karesti: @rvansa I said that in the description, @gustavonalle is working on a PR whose refactoring will be the one. My encoding classes are more something I did here because passing 4 classes every time was too heavy ; specially knowing that I will have to change all that after Gustavo

core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyValueCommand.java
```diff
@@ -144,20 +151,30 @@ public LoadType loadType() {
...
+            .append(", valueMatcher=").append(valueMatcher)
+            .append(", successful=").append(successful)
+            .append(", encodingClasses=").append(encodingClasses)
+            .append("}")
+            .toString();
    }
 
    @Override
    public Mutation toMutation(K key) {
       return new Mutations.ReadWriteWithValue<>(value, f);
```
karesti: I totally left the tx mode off yet, I admit

core/src/main/java/org/infinispan/commands/functional/functions/InjectableWrappper.java
```diff
@@ -0,0 +1,8 @@
+package org.infinispan.commands.functional.functions;
+
+import org.infinispan.factories.ComponentRegistry;
+
+public interface InjectableWrappper {
```
karesti: yes, I would like to merge the "merge" PR so I can pick up this ! 💃

core/src/main/java/org/infinispan/commands/functional/TxReadOnlyKeyCommand.java
```diff
@@ -65,7 +67,7 @@ public Object perform(InvocationContext ctx) throws Throwable {
       if (f != null) {
          ret = f.apply(rw);
       }
-      return snapshot(ret);
+      return snapshot(ret, new CacheEncoders());
```
karesti: hm, yes and no. I didn't work on the tx part. So encoding + tx are not supported in this PR (no test done on that). Something to be done

core/src/main/java/org/infinispan/compat/BiConsumerMapper.java
```diff
@@ -0,0 +1,75 @@
...
+   public static class Externalizer implements AdvancedExternalizer<BiConsumerMapper> {
+
+      @Override
+      public Set<Class<? extends BiConsumerMapper>> getTypeClasses() {
+         return Collections.singleton(BiConsumerMapper.class);
+      }
+
+      @Override
+      public Integer getId() {
+         return Ids.BI_FUNCTION_MAPPER;
```
karesti: yes, I added this class but I wanted to delete it because I was not using it anymore ! and I forgot

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
karesti: intellij did a big refactoring i did not want, and I didn't see this class !

core/src/main/java/org/infinispan/cache/impl/EncodingClasses.java
```diff
@@ -0,0 +1,80 @@
...
+
+   public Class<? extends Wrapper> getKeyWrapperClass() {
+      return keyWrapperClass;
+   }
+
+   public Class<? extends Wrapper> getValueWrapperClass() {
+      return valueWrapperClass;
+   }
+
+   public static class Externalizer implements AdvancedExternalizer<EncodingClasses> {
```
karesti: I don't think so, for this I'm just gonna wait for @gustavonalle code so I can do it like him

core/src/main/java/org/infinispan/cache/impl/EncoderValueMapper.java
```diff
@@ -62,14 +58,13 @@ public Integer getId() {
 
       @Override
       public void writeObject(ObjectOutput output, EncoderValueMapper object) throws IOException {
-         output.writeObject(object.valueEncoderClass);
-         output.writeObject(object.valueWrapperClass);
+         output.writeObject(object.encodingClasses);
```
karesti: yep, again, gustavo

core/src/main/java/org/infinispan/commands/functional/AbstractWriteKeyCommand.java
karesti: I'm not convinced 100% about what I did here, but I didn't want to pass every time the same code all over again !

core/src/main/java/org/infinispan/commands/functional/ReadOnlyManyCommand.java
karesti: I think I did this because at some points of the code when this constructor is called, dependencies - cache encoders - are null.

core/src/main/java/org/infinispan/commands/functional/ReadOnlyManyCommand.java
karesti: deleted

core/src/main/java/org/infinispan/commands/functional/AbstractWriteKeyCommand.java
rvansa: My point was that I'd like to make this class immutable, and the default would be set during injection.

core/src/main/java/org/infinispan/functional/FunctionalMap.java
```diff
@@ -86,6 +86,14 @@
    ComponentStatus getStatus();
 
    /**
+    * is encoded
```
karesti: ok

core/src/main/java/org/infinispan/cache/impl/EncodingClasses.java
rvansa: Maybe you got me wrong: Encoders could be user-supplied, so there's more magic up to gustavo, but `EncodingClasses` won't be marshalled somewhere arbitrarily in the object graph. You know when you are marshalling/unmarshalling that, so you don't have to send the externalizer id.

core/src/main/java/org/infinispan/commands/functional/ReadOnlyManyCommand.java
rvansa: Then fix it in the original (copied) object, please.

core/src/test/java/org/infinispan/functional/FunctionalEncoderTest.java
karesti: for the Bifunction mapper, simply because it is not used haha. I wanted to delete but I  didn't see I left it.
Ok to add tx tests and so the support. And ok to test with Advanced Caches and not just the store type.

core/src/main/java/org/infinispan/cache/impl/EncodingClasses.java
karesti: ok

core/src/main/java/org/infinispan/cache/impl/CacheEncoders.java
```diff
@@ -0,0 +1,77 @@
...
+   }
+
+   public CacheEncoders(Encoder keyEncoder, Wrapper keyWrapper, Encoder valueEncoder, Wrapper valueWrapper) {
+      this.keyEncoder = keyEncoder;
+      this.keyWrapper = keyWrapper;
+      this.valueEncoder = valueEncoder;
+      this.valueWrapper = valueWrapper;
+   }
+
+   public void grabEncodersFromRegistry(EncoderRegistry encoderRegistry, EncodingClasses encodingClasses) {
```
karesti: ok

core/src/main/java/org/infinispan/commands/functional/AbstractWriteKeyCommand.java
karesti: ok

karesti:
@rvansa @gustavonalle I added this issue to treat Tx + Encoding in functional maps in another issue
https://issues.jboss.org/browse/ISPN-8090

core/src/main/java/org/infinispan/functional/impl/EntryViews.java
```diff
@@ -32,81 +35,106 @@ private EntryViews() {
...
          EntryBackedReadOnlyView view = (EntryBackedReadOnlyView) ret;
-         return (R) new ReadOnlySnapshotView(view.key(), view.entry.getValue(), view.entry.getMetadata());
+         return (R) new ReadOnlySnapshotView(view.key(), view.find().orElse(null), view.entry.getMetadata());
+      } else if (ret instanceof NoValueReadOnlyView) {
+         NoValueReadOnlyView view = (NoValueReadOnlyView) ret;
+         return (R) new ReadOnlySnapshotView(view.key(), null, null);
       }
 
-      return ret;
+      return (R) cacheEncoders.valueFromStorage(ret);
```
karesti: I remembered I did the conversion there too for some reason, some tests not passing, event if at first I didn't put anything there. But this is probably a wrong solution I did because the code was not working properly somewhere else. I'm going to remove it and see what it happens

commons/src/main/java/org/infinispan/commons/dataconversion/EncodingUtils.java
karesti: I aligned to the rest of the class where IllegalArgumentException is thrown instead of NullPointerException. But this refactoring can indeed be done.

core/src/main/java/org/infinispan/functional/impl/EntryViews.java
```diff
@@ -280,24 +314,33 @@ public String toString() {
...
       }
 
       @Override
       public Optional<V> find() {
-         return Optional.ofNullable(prevValue);
+         return Optional.ofNullable((V) cacheEncoders.valueFromStorage(prevValue));
+      }
+
+      public V getCurrentValue() {
+         return (V) cacheEncoders.valueFromStorage(entry.getValue());
```
karesti: This is used for snapshot. the getValue gives PrevValue there, and I need the newValue encoded.

core/src/main/java/org/infinispan/commands/functional/ReadOnlyManyCommand.java
karesti: probably another thing was wrong before because it seems to work properly now

core/src/main/java/org/infinispan/functional/impl/EntryViews.java
rvansa: sure, I just said that the single usage of the method could be inlined

core/src/main/java/org/infinispan/functional/impl/EntryViews.java
```diff
@@ -32,81 +35,106 @@ private EntryViews() {
...
       @Override
       public K key() {
-         return entry.getKey();
+         return (K) cacheEncoders.keyFromStorage(entry.getKey());
       }
 
       @Override
       public Optional<V> find() {
-         return entry == null ? Optional.empty() : Optional.ofNullable(entry.getValue());
+         return entry == null ? Optional.empty() : Optional.ofNullable((V) cacheEncoders.valueFromStorage(entry.getValue()));
```
karesti: @rvansa you mean that if find() is called several times, to have the decoded value already decoded and available ?

karesti: @rvansa I did something in the last commit to follow your comment, but I'm not really 100% I understand. In any case, you can review this PR on the commits I did and tell me if you see anything else that should be fixed/refactored beyond the transactions which will be supported in another ticket

gustavonalle:
FYI, I am working atm on https://issues.jboss.org/browse/ISPN-8004 which will avoid the cost of shipping those classes around.

tristantarrant:

PR #5258, opened 2017-07-03
ISPN-8002 ScatteredStreamIteratorTest.waitUntilProcessingResults random failures

rvansa:
https://issues.jboss.org/browse/ISPN-8002

Piggybacking little fix in JGroupsTransport.

PR #5272, opened 2017-07-07
Ispn 8033 stream rehash is broken with a map operation

wburns:
https://issues.jboss.org/browse/ISPN-8033

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/distribution/DistributionBulkInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/DistributionBulkInterceptor.java
@@ -77,12 +77,13 @@ public class DistributionBulkInterceptor<K, V> extends DDAsyncInterceptor {
       });
    }
 
-   protected static <K, V> Supplier<CacheStream<CacheEntry<K, V>>> supplier(Cache<K, V> cache, CacheStream<CacheEntry<K, V>> stream) {
+   protected static <K, V> Supplier<CacheStream<CacheEntry<K, V>>> supplier(Cache<K, V> cache,
+         Supplier<CacheStream<CacheEntry<K, V>>> streamSupplier) {
       if (cache.getCacheConfiguration().clustering().cacheMode().isScattered()) {
          // ignore tombstones
-         return () -> stream.filter(entry -> entry.getValue() != null);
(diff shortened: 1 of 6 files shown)


pruivo:
looks good to me, except this suspicious test:
```
org.infinispan.scattered.stream.ScatteredStreamIteratorTest.waitUntilProcessingResults[SCATTERED_SYNC, tx=false]
```

wburns:
Hrmm this change shouldn't cause that, my guess is that it is a random failure. I will check into it though.

danberindei:
@pruivo @wburns should have been fixed by https://github.com/infinispan/infinispan/pull/5258

wburns:
Looking at [1] it seems that scattered cache still skipped a segment (as it is null), which is what #5258 fixed. My guess is that there is another reason outside of this that can cause a segment to be skipped.

This PR is just to make sure that the stream is generated fresh each time a rehash occurs.

[1] http://ci.infinispan.org/job/Infinispan/job/PR-5272/7/testReport/junit/org.infinispan.scattered.stream/ScatteredStreamIteratorTest/waitUntilProcessingResults_SCATTERED_SYNC__tx_false_/

wburns:
The only thing I can think of if this PR is causing this failure is that Scattered cache iteration requires the same stream instance to be returned every time from the supplier. But that defeats the purpose of the supplier and actually ends up applying intermediate operations (map/flatMap/filter) multiple times, which is why map/flatMap don't work then.

pruivo:
ok. going to push the button :)

wburns:
Yeah sadly I think @rvansa needs to look into it again possibly. But I know that this fixes a real issue, that you can reproduce every time with the test change [1] I have.

[1] https://github.com/infinispan/infinispan/pull/5272/files#diff-d7679d36fd7db7278a4661ebd6adbd0cR91

pruivo:
Integrated! thanks @wburns

danberindei:
@wburns Maybe you should have moved the test to to unstable group? Let's try to avoid at least the test failures we know about in master :)

wburns:
@danberindei sounds fine to me. We should probably update the unstable ones every couple days until we finally have green (making sure to create JIRAs to fix them obviously)

PR #5273, opened 2017-07-09
Support and document the new `local_heap` configuration property for Query configuration

Sanne:
https://issues.jboss.org/browse/ISPN-8034

The "ram" short name to use Lucene's RAMDirectory implementation was renamed to "local_heap" in Hibernate Search to make its function clearer in the context of Infinispan as there are many kinds of ram.

The "ram" keyword was hardcoded into the Infinispan project though, so this needs to be updated and documented.


tristantarrant:
Looks good. I'd like a note in the upgrading.adoc file too please

Sanne:
> Looks good. I'd like a note in the upgrading.adoc file too please

Right! thanks, I didn't know about that one. Will do asap.

Sanne:
Updated the migration guide.

tristantarrant:
Pushed to master, thanks !

PR #5277, opened 2017-07-10
[9.0.x] ISPN-7991 Improve Compatibility between Rest and HotRod for text content

gustavonalle:
Created no-arg ctor to be used as compat mode marshaller

https://issues.jboss.org/browse/ISPN-7991

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/marshall/StringMarshaller.java
+++ b/commons/src/main/java/org/infinispan/commons/marshall/StringMarshaller.java
@@ -6,7 +6,7 @@ import java.nio.charset.Charset;
 import org.infinispan.commons.io.ByteBuffer;
 import org.infinispan.commons.io.ByteBufferImpl;
 
-public final class StringMarshaller extends AbstractMarshaller {
+public class StringMarshaller extends AbstractMarshaller {
 
    final Charset charset;
 
--- a/commons/src/main/java/org/infinispan/commons/marshall/UTF8StringMarshaller.java
+++ b/commons/src/main/java/org/infinispan/commons/marshall/UTF8StringMarshaller.java
@@ -1,0 +1,11 @@
+package org.infinispan.commons.marshall;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public final class UTF8StringMarshaller extends StringMarshaller {
+
+   public UTF8StringMarshaller() {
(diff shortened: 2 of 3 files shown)


commons/src/main/java/org/infinispan/commons/marshall/UTF8Marshaller.java
```diff
@@ -0,0 +1,29 @@
+package org.infinispan.commons.marshall;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+
+import org.infinispan.commons.io.ByteBuffer;
+import org.infinispan.commons.io.ByteBufferImpl;
+
+public final class UTF8Marshaller extends AbstractMarshaller {
```
tristantarrant: 9.0.x and master have https://github.com/infinispan/infinispan/blob/master/commons/src/main/java/org/infinispan/commons/marshall/StringMarshaller.java
So we should probably have a default constructor for that which uses UTF-8 (as well as backporting the class to 8.2.x)

gustavonalle: I prefer to have multiple instances of marshaller, for each charset, since it needs to used in the server xml, and for clarity, having a ```marshaller="UTF8Marshaller"``` is more descriptive than ```marshaller="StringMarshaller"```. What if someone needs a different charset?

gustavonalle: I could make UTF8Marshaller (and the others) extend StringMarshaller, how does that sound?

tristantarrant: Yes, that is also acceptable

gustavonalle: Too bad StringMarshaller is ```final```  😄

tristantarrant: Since it is final now I doubt removing the qualifier is going to cause incompatibilities...

anistor:
Some comments from https://github.com/infinispan/infinispan/pull/5278 should be applied here too.

anistor:
Integrated in 9.0.x. Thanks @gustavonalle !

PR #5279, opened 2017-07-10
ISPN-7990 Provide a way to remove an entry from all indexes

gustavonalle:
https://issues.jboss.org/browse/ISPN-7990

### DIFF
--- a/query/src/main/java/org/infinispan/query/backend/QueryInterceptor.java
+++ b/query/src/main/java/org/infinispan/query/backend/QueryInterceptor.java
@@ -9,6 +9,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.transaction.TransactionManager;
 import javax.transaction.TransactionSynchronizationRegistry;
@@ -188,7 +190,7 @@ public final class QueryInterceptor extends DDAsyncInterceptor {
    public Object visitComputeCommand(InvocationContext ctx, ComputeCommand command) throws Throwable {
       InternalCacheEntry internalCacheEntry = dataContainer.get(command.getKey());
       Object stateBeforeCompute = internalCacheEntry != null ? internalCacheEntry.getValue() : null;
-      return invokeNextThenAccept(ctx, command, (rCtx, rCommand, rv) -> processComputeCommand(((ComputeCommand) rCommand), rCtx, stateBeforeCompute, rv,null));
+      return invokeNextThenAccept(ctx, command, (rCtx, rCommand, rv) -> processComputeCommand(((ComputeCommand) rCommand), rCtx, stateBeforeCompute, rv, null));
    }
 
    @Override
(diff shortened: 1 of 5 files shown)


gustavonalle:
@karesti Once you have the test to reproduce the issue of lack of values in the context during topology change, you can test against this branch.

rvansa:
@gustavonalle Needs rebase.

rvansa:
LGTM, @anistor could you bless this?

anistor:
Sure, I'll have a look after the meeting.

gustavonalle:
Rebased

query/src/main/java/org/infinispan/query/backend/QueryInterceptor.java
```diff
@@ -232,6 +234,18 @@ public void purgeIndex(Class<?> entityType) {
       purgeIndex(null, entityType);
    }
 
+   /**
+    * Remove entries from all indexes by key
+    */
+   void removeFromIndexes(TransactionContext transactionContext, Object key) {
+      transactionContext = transactionContext == null ? makeTransactionalEventContext() : transactionContext;
```
anistor: Assigning same value to itself is a bit strange. Why not do it like ```if (transactionContext == null) { transactionContext = makeTransactionalEventContext();
}```   ?

query/src/main/java/org/infinispan/query/backend/SearchWorkCreator.java
```diff
@@ -31,4 +31,7 @@
     * @return collection of Work instances
     */
    Collection<Work> createPerEntityWorks(Object entity, Serializable id, WorkType workType);
+
+   Collection<Work> createPerEntityIdWorks(Serializable id, IndexedTypeIdentifier entityType, WorkType workType);
```
anistor: A bit of javadoc would be nice.

query/src/main/java/org/infinispan/query/backend/QueryInterceptor.java
gustavonalle: This used everywhere in the QI

query/src/main/java/org/infinispan/query/backend/SearchWorkCreator.java
gustavonalle: ok. Will change a bit this method, it does not need to return a collection

query/src/main/java/org/infinispan/query/backend/QueryInterceptor.java
```diff
@@ -9,6 +9,8 @@
 import java.util.Set;
```
gustavonalle: This came from the "official" formatter unfortunately. Should we avoid in general to reformat comments?

gustavonalle:
Updated.

query/src/main/java/org/infinispan/query/backend/QueryInterceptor.java
anistor: I we don't change any content, then better not reformat them.

query/src/main/java/org/infinispan/query/backend/QueryInterceptor.java
anistor: Indeed, and it looks a bit silly. Let's leave like that in the name of tradition :)

anistor:

PR #5280, opened 2017-07-11
ISPN-7754 forEach method

karesti:
https://issues.jboss.org/browse/ISPN-7754

### DIFF
--- a/core/src/main/java/org/infinispan/Cache.java
+++ b/core/src/main/java/org/infinispan/Cache.java
@@ -5,6 +5,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
@@ -443,4 +444,18 @@ public interface Cache<K, V> extends BasicCache<K, V>, BatchingCache, FilteringL
                      SerializableBiFunction<? super K, ? super V, ? extends V> remappingFunction) {
       return compute(key, (BiFunction<? super K, ? super V, ? extends V>) remappingFunction);
    }
+
+   /**
+    * {@inheritDoc}
+    * <p>
+    * When this method is used on a clustered cache, either replicated or distributed, the biconsumer will be serialized
+    * to owning nodes to perform the operation in the most performant way. However this means the biconsumer must
+    * have an appropriate {@link org.infinispan.commons.marshall.Externalizer} or be {@link java.io.Serializable} itself.
+    * <p>
+    * For transactional caches, whenever the values of the caches are collections, and the mapping function modifies the collection, the collection
+    * must be copied and not directly modified, otherwise whenever rollback is called it won't work.
+    * This limitation could disappear in following releases if technically possible.
+    */
+   @Override
+   void forEach(BiConsumer<? super K, ? super V> action);
 }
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/infinispan/Cache.java
```diff
@@ -443,4 +444,18 @@ default V compute(K key,
...
    }
+
+   /**
+    * {@inheritDoc}
+    * <p>
+    * When this method is used on a clustered cache, either replicated or distributed, the biconsumer will be serialized
+    * to owning nodes to perform the operation in the most performant way. However this means the biconsumer must
+    * have an appropriate {@link org.infinispan.commons.marshall.Externalizer} or be {@link java.io.Serializable} itself.
+    * <p>
+    * For transactional caches, whenever the values of the caches are collections, and the mapping function modifies the collection, the collection
```
rvansa: It's not specific to collections: the values in cache must not be mutated, period, and we'll rather enforce that than remove the limitation.

core/src/main/java/org/infinispan/Cache.java
```diff
@@ -443,4 +444,18 @@ default V compute(K key,
...
+    * must be copied and not directly modified, otherwise whenever rollback is called it won't work.
+    * This limitation could disappear in following releases if technically possible.
+    */
+   @Override
+   void forEach(BiConsumer<? super K, ? super V> action);
```
rvansa: Where is the `SerializableBiConsumer` overload?

core/src/main/java/org/infinispan/security/impl/SecureCacheImpl.java
```diff
@@ -357,6 +358,12 @@ public V compute(K key, BiFunction<? super K, ? super V, ? extends V> remappingF
    }
 
    @Override
+   public void forEach(BiConsumer<? super K, ? super V> action) {
+      authzManager.checkPermission(subject, AuthorizationPermission.WRITE);
```
rvansa: Shouldn't this be READ?

core/src/main/java/org/infinispan/Cache.java
rvansa: Err, actually this is a consumer, and you should not modify the collection - what would you do with the copy?

core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
```diff
@@ -396,4 +400,38 @@ public void testComputeIfAbsentFromNonOwner() throws InterruptedException {
...
+   static final Map<Object, String> forEachStructure = new ConcurrentHashMap<>();
+   static final AtomicInteger forEachOffset = new AtomicInteger();
+
+   static int populateNextForEachStructure(String obj) {
+      int offset = forEachOffset.getAndIncrement();
+      forEachStructure.put(offset, obj);
+      return offset;
+   }
+
+   public void testForEach() throws Exception {
```
rvansa: This test is silly. Accessing a static variable on (potentially) remote node...

core/src/main/java/org/infinispan/security/impl/SecureCacheImpl.java
karesti: yes, it should be

core/src/main/java/org/infinispan/Cache.java
karesti: this is me that does copy/paste without thinking twice sometimes. So well, the comment is not correct, indeed !!

core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
karesti: 😢

tristantarrant:

PR #5284, opened 2017-07-11
Ispn 4099 no rehash event

wburns:
https://issues.jboss.org/browse/ISPN-4099

### DIFF
--- a/core/src/main/java/org/infinispan/notifications/cachelistener/NotifyHelper.java
+++ b/core/src/main/java/org/infinispan/notifications/cachelistener/NotifyHelper.java
@@ -6,6 +6,7 @@ import org.infinispan.commands.write.RemoveCommand;
 import org.infinispan.commands.write.WriteCommand;
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.context.InvocationContext;
+import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.functional.impl.EntryViews;
 import org.infinispan.functional.impl.FunctionalNotifier;
 import org.infinispan.metadata.Metadata;
@@ -14,6 +15,10 @@ public class NotifyHelper {
    public static void entryCommitted(CacheNotifier notifier, FunctionalNotifier functionalNotifier,
                                      boolean created, boolean removed, boolean expired, CacheEntry entry,
                                      InvocationContext ctx, FlagAffectedCommand command, Object previousValue, Metadata previousMetadata) {
+      // We only notify if there is no state transfer flag
+      if (FlagBitSets.extractStateTransferFlag(ctx, command) != null) {
+         return;
+      }
       boolean isWriteOnly = (command instanceof WriteCommand) && ((WriteCommand) command).isWriteOnly();
       if (removed) {
          if (command instanceof RemoveCommand) {
@@ -57,8 +62,8 @@ public class NotifyHelper {
             // command is read-write.
             if (!isWriteOnly)
                functionalNotifier.notifyOnModify(
-                  EntryViews.readOnly(entry.getKey(), previousValue, previousMetadata),
-                  EntryViews.readOnly(entry));
+                     EntryViews.readOnly(entry.getKey(), previousValue, previousMetadata),
+                     EntryViews.readOnly(entry));
 
             functionalNotifier.notifyOnWrite(() -> EntryViews.readOnly(entry));
          }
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/infinispan/notifications/cachelistener/NotifyHelper.java
```diff
@@ -14,54 +17,73 @@
...
-      if (removed) {
-         if (command instanceof RemoveCommand) {
-            ((RemoveCommand) command).notify(ctx, previousValue, previousMetadata, false);
-         } else if (command instanceof InvalidateCommand) {
-            notifier.notifyCacheEntryInvalidated(entry.getKey(), entry.getValue(), entry.getMetadata(), false, ctx, command);
-         } else {
-            if (expired) {
-               notifier.notifyCacheEntryExpired(entry.getKey(), previousValue, previousMetadata, ctx);
+      // We only notify if there is no state transfer flag
+      if (extractStateTransferFlag(ctx, command) == null) {
```
karesti: There are a lot of if/else, could it be possible to do something like

```java

if (containsStateTransferFlag(ctx, command)) 
 // We only notify if there is no state transfer flag
   return ;

// the content of the if
```

core/src/main/java/org/infinispan/notifications/cachelistener/NotifyHelper.java
```diff
@@ -14,54 +17,73 @@
...
+                        EntryViews.readOnly(entry));
 
-            functionalNotifier.notifyOnWrite(() -> EntryViews.readOnly(entry));
+               functionalNotifier.notifyOnWrite(() -> EntryViews.readOnly(entry));
+            }
+         }
+      }
+   }
+
+   private static Flag extractStateTransferFlag(InvocationContext ctx, FlagAffectedCommand command) {
```
karesti: Could this method return a `boolean` instead of the flag ?

wburns: This was actually copied from [1]. I want to have it one place, but I couldn't decide where yesterday. The more I think about this morning I am thinking that NotifyHelper should be fine until we can find a better place, if any. My other thought was maybe on [2], but it seems oddly specific for it. WDYT?

[1] https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/interceptors/impl/EntryWrappingInterceptor.java#L550
[2] https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/context/impl/FlagBitSets.java

core/src/main/java/org/infinispan/notifications/cachelistener/NotifyHelper.java
wburns: I had thought about this, but I tend to not like multiple locations in a method that can return. Thinking on it more it seems harmless enough, so it should be fine.

core/src/main/java/org/infinispan/notifications/cachelistener/NotifyHelper.java
karesti: FlagBitSets.java seems a better place to me than NotifyHelper. It's related to the flags and the utility method makes sense to me there. It might be in a specific utility class `FlagExtractor ?` if any other code related to flags extracting or check is duplicated in core.

wburns:
Updated

core/src/main/java/org/infinispan/notifications/cachelistener/NotifyHelper.java
wburns: Put it in FlagBitSets for now.

pruivo:
looks good! waiting for CI results...

pruivo:
@wburns, there is a suspicious failure: `org.infinispan.client.hotrod.event.ClientClusterExpirationEventsTest.testSimpleExpired`

wburns:
Checking.

wburns:
Initial look seems unrelated as there is no rehash occurring in the test, but I will look a little closer.

wburns:
The failure for some reason saw a different key expiration, which makes no sense for the test as it only uses a single key. My guess is that somehow 2 parallel tests interacted

pruivo:
so, it isn't related to the changes in the PR... can I push it? or you wanna try to fix it?

pruivo:
integrated! thanks @wburns !

wburns:
Created https://issues.jboss.org/browse/ISPN-8067 to handle the test failure.

PR #5293, opened 2017-07-14
ISPN-7114: Fix MergePolicy test failures.

ryanemerson:


ryanemerson:
I seem to have fixed the issues with the MergePolicy tests, but there is now a failure on the ConflictManagerTest. I'll fix all in this PR.

core/src/test/java/org/infinispan/conflict/impl/ConflictManagerTest.java
```diff
@@ -87,7 +87,7 @@ public void testGetAllVersionsDuringStateTransfer() throws Exception {
       });
       fork(() -> partition(0).merge(partition(1)));
 
-      Map<Address, InternalCacheValue<Object>> versionMap = versionFuture.get(STATE_TRANSFER_DELAY * 8, TimeUnit.MILLISECONDS);
+      Map<Address, InternalCacheValue<Object>> versionMap = versionFuture.get(STATE_TRANSFER_DELAY * 16, TimeUnit.MILLISECONDS);
```
rvansa: How are those magic multipliers chosen?

ryanemerson: Good point. Originally there was some logic to the multiplier, however this is no longer  (maybe never was) applicable. I've  updated the value to be explicitly 60 seconds, which is the same timeout used when waiting for a rebalance.

danberindei:
@ryanemerson please remove the `:` from the commit message.

danberindei:
The latest failure I've seen locally is from another test method:

    16:33:08,072 ERROR (testng-Test:[]) [TestSuiteProgress] Test failed: org.infinispan.conflict.impl.ConflictManagerTest.testGetConflictsDuringStateTransfer[DIST_SYNC]
    org.testng.TestException: 
    Method org.infinispan.test.fwk.NamedTestMethod@8bf055db should have thrown an exception of class java.lang.IllegalStateException

My guess is that `delayStateTransferCompletion()` is called too late, after the relevant `StateResponseCommand` was already received (and the same with the rehash listener). To make it more reliable, the test should use a `CountDownLatch` or maybe an `o.i.t.f.CheckPoint` to make sure that the `StateResponseCommand` is blocked and that it's not unblocked before `getConflicts(0)` gets called. That way the test also doesn't have to wait for 4 seconds when `getConflicts(0)` finishes immediately.

Optional request: I suggest using `Exceptions.expectException(..., () -> getConflicts(0))` or `Exceptions.expectExecutionException(..., conflictsFuture) instead of the `@Test(expectedException = ...)` annotation, to make it clearer where the exception is thrown.

ryanemerson:
@danberindei Good points. Everything seems fine on my machine, now lets see what CI says.

core/src/test/java/org/infinispan/conflict/impl/ConflictManagerTest.java
```diff
@@ -203,9 +198,8 @@ public void testCacheOperationOnConflictStream() {
       assertTrue(cache.isEmpty());
    }
 
-   @Test(expectedExceptions = CacheException.class)
    private void testNoEntryMergePolicyConfigured() {
```
pruivo: I'm not sure if it runs the private methods as test.

core/src/test/java/org/infinispan/conflict/impl/ConflictManagerTest.java
```diff
@@ -74,61 +77,53 @@ public void testGetAllVersionsDuringStateTransfer() throws Exception {
...
 
       Future<Map<Address, InternalCacheValue<Object>>> versionFuture = fork(() -> {
-         while (!listener.isRehashInProgress.get())
-            TestingUtil.sleepThread(100);
-
+         listener.latch.await();
          return getAllVersions(0, key);
       });
       fork(() -> partition(0).merge(partition(1)));
+      new Timer().schedule(new TimerTask() {
```
pruivo: I really don't like tests based on timers... isn't possible to do the following:
```
listener.latch.await()
check not versionFuture.isDone()
latch.counterDown()
versionMap = versionFuture.get(...)
```

pruivo: edit: or invoke the `fork()` after `listener.latch.await()`

core/src/test/java/org/infinispan/conflict/impl/ConflictManagerTest.java
```diff
@@ -74,61 +75,48 @@ public void testGetAllVersionsDuringStateTransfer() throws Exception {
...
+      getCache(0).addListener(listener);
+      CountDownLatch latch = new CountDownLatch(1);
+      delayStateTransferCompletion(latch);
 
       Future<Map<Address, InternalCacheValue<Object>>> versionFuture = fork(() -> {
-         while (!listener.isRehashInProgress.get())
-            TestingUtil.sleepThread(100);
-
+         listener.latch.await();
+         fork(latch::countDown);
```
danberindei: I don't think you need a fork here.

danberindei: Actually I think the count-down doesn't belong here. I'd change the test to something like this:

``` 
      // Trigger the merge and wait for state transfer to begin
      Future<?> mergeFuture = fork(() -> partition(0).merge(partition(1)));
      listener.latch.await();

      Future<Map<Address, InternalCacheValue<Object>>> versionFuture = fork(() -> getAllVersions(0, key));

      // Check that getAllVersions doesn't return while state transfer is in progress
      Exceptions.expectException(TimeoutException.class, () -> versionFuture.get(500, TimeUnit.MILLISECONDS));

      // Allow and wait for state transfer to finish
      latch.countDown();
      mergeFuture.get(30, TimeUnit.SECONDS);

      // Check the results
      Map<Address, InternalCacheValue<Object>> versionMap = versionFuture.get(10, TimeUnit.SECONDS);
      assertTrue(versionMap != null);
      assertTrue(!versionMap.isEmpty());
      assertEquals(String.format("Returned versionMap %s", versionMap),2, versionMap.size());
      versionMap.values().forEach(icv -> assertEquals(value, icv.getValue()));
```

core/src/test/java/org/infinispan/conflict/impl/ConflictManagerTest.java
```diff
@@ -277,21 +266,30 @@ private void dropClusteredGetCommands() {
...
 
       @Override
       public void handle(CacheRpcCommand command, Reply reply, DeliverOrder order) {
          if (command instanceof StateResponseCommand) {
-            TestingUtil.sleepThread(STATE_TRANSFER_DELAY);
+            StateResponseCommand stc = (StateResponseCommand) command;
+            boolean isLastChunk = stc.getStateChunks().stream().anyMatch(StateChunk::isLastChunk);
+            if (isLastChunk) {
+               try {
+                  latch.await();
```
danberindei: I fear this will hang during teardown if the test fails and doesn't call `latch.countDown()`. I have some thoughts about making it easier to close latches in tests, but in the meantime I suggest adding a timeout here.

danberindei:
@ryanemerson Sorry for the delayed review. Latches come with their own set of problems, as you can see, but I'm sure the end result will be a reliable test.

ryanemerson:
@danberindei No worries, thanks for the insights. Updated and pushed

danberindei:
Integrated, thanks Ryan!

PR #5297, opened 2017-07-18
Random failing test

karesti:
Try fix on ...

* computeIfPresent
* computeIfAbsent
* compute

... related tests in distributed mode

### DIFF
--- a/core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
+++ b/core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
@@ -63,9 +63,9 @@ public class DistSyncFuncTest extends BaseDistFunctionalTest<Object, String> {
       List l3 = getCacheTopology(c3).getDistribution(key).writeOwners();
       List l4 = getCacheTopology(c4).getDistribution(key).writeOwners();
 
-      assertEquals("L1 "+l1+" and L2 "+l2+" don't agree.", l1, l2);
-      assertEquals("L2 "+l2+" and L3 "+l3+" don't agree.", l2, l3);
-      assertEquals("L3 "+l3+" and L4 "+l4+" don't agree.", l3, l4);
+      assertEquals("L1 " + l1 + " and L2 " + l2 + " don't agree.", l1, l2);
+      assertEquals("L2 " + l2 + " and L3 " + l3 + " don't agree.", l2, l3);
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
```diff
@@ -306,27 +306,19 @@ public void testComputeFromNonOwner() throws InterruptedException {
       // compute function applied
       initAndTest();
       Object retval = getFirstNonOwner("k1").compute("k1", (k, v) -> "computed_" + k + "_" + v);
-      if (testRetVals) assertEquals("computed_k1_value", retval);
       asyncWait("k1", ComputeCommand.class);
-      assertOnAllCachesAndOwnership("k1", "computed_k1_value");
+      if (testRetVals) assertEquals("computed_k1_value", retval);
+      eventuallyAssertOnAllCachesAndOwneship("k1", "computed_k1_value");
```
rvansa: Why do you need the `eventually`? `asyncWait` should do the trick, shouldn't it?

karesti: well, it does not, apparently. I thought this "eventually" way makes thing more stable

rvansa: The flakiness might demonstrate a problem in the code that's not handled. Tests should not be fixed by loosening the requirements unless we know what's going on.

karesti: When I see a test that passes all the time for a while but in jenkins I get suddenly a => Expecting [k1] to equal [null] on cache [DistSyncFuncTest[DIST_SYNC, tx=false]-NodeB-50117] but was [computed_k1_value]. Owners are [Cache &apos;dist&apos;@DistSyncFuncTest[DIST_SYNC, tx=false]-NodeC-19559, Cache &apos;dist&apos;@DistSyncFuncTest[DIST_SYNC, tx=false]-NodeD-7512] 

telling me that a non owner has not deleted a value that was supposed to be deleted when looking directly in every cache, my guess is that the Command has not been executed yet as fast as it should be in every cache and the check is happening before that happens. The asyncWait of the command does nothing in the case of DistSyncFuncTest.

But you are right, the thing is that I'm not able to reproduce this problem. I'm going to open a ticket for this and check closer after I finish one thing I'm working on.

PR #5299, opened 2017-07-19
ISPN-7086 Fix memached server not reading all of a value

tristantarrant:
- Do not be cautious about reading past the buffer since ReplayingDecoder will handle this
- Replace recursive calls with iterative

https://issues.jboss.org/browse/ISPN-7086


tristantarrant:
Fixed the checkstyle warning

galderz:
Thx @tristantarrant, integrated

PR #5300, opened 2017-07-19
[9.0.x] ISPN-7086 Fix memached server not reading all of a value

tristantarrant:
- Do not be cautious about reading past the buffer since ReplayingDecoder will handle this
- Replace recursive calls with iterative

https://issues.jboss.org/browse/ISPN-7086


server/memcached/src/main/java/org/infinispan/server/memcached/MemcachedDecoder.java
```diff
@@ -1049,6 +1049,7 @@ private MemcachedOperation toRequest(String commandName, Boolean endOfOp, ByteBu
                String line = readDiscardedLine(buffer); // Read rest of line to clear the operation
                log.debugf("Unexpected operation '%s', rest of line contains: %s", commandName, line);
             }
+            System.out.printf("Readable bytes %d\n", buffer.readableBytes());
```
slaskawi: `System.out`?

tristantarrant:
Yeah, silly temporary debug that got left in unintentionally. I've removed and repushed both branches

galderz:
This is integrated too.

PR #5308, opened 2017-07-20
ISPN-5476 Cross-site tests should run in parallel

danberindei:
https://issues.jboss.org/browse/ISPN-5476

May also fix https://issues.jboss.org/browse/ISPN-7024

Create a TEST_RELAY2 protocol that allows setting a different bridge cluster name for each test.

### DIFF
--- a/core/src/main/java/org/infinispan/xsite/XSiteAdminOperations.java
+++ b/core/src/main/java/org/infinispan/xsite/XSiteAdminOperations.java
@@ -273,14 +273,14 @@ public class XSiteAdminOperations {
                            "The remote site will be bring back online",
                      name = "pushState")
    public final String pushState(@Parameter(description = "The destination site name", name = "SiteName") String siteName) {
-      String status = bringSiteOnline(siteName);
-      if (!SUCCESS.equals(status)) {
-         return String.format("Unable to pushState to '%s'. %s", siteName, status);
-      }
       try {
+         String status = bringSiteOnline(siteName);
+         if (!SUCCESS.equals(status)) {
+            return String.format("Unable to pushState to '%s'. %s", siteName, status);
+         }
          stateTransferManager.startPushState(siteName);
       } catch (Throwable throwable) {
-         log.debugf(throwable, "Unable to pushState to '%s'.", siteName);
+         log.xsiteAdminOperationError("pushState", siteName, throwable);
          return String.format("Unable to pushState to '%s'. %s", siteName, throwable.getLocalizedMessage());
       }
       return SUCCESS;
@@ -313,7 +313,7 @@ public class XSiteAdminOperations {
                      description = "Clears the state transfer status.",
                      name = "ClearPushStateStatus")
    public final String clearPushStateStatus() {
-      return performOperation(() -> stateTransferManager.clearClusterStatus());
+      return performOperation("clearPushStateStatus", "(local)", () -> stateTransferManager.clearClusterStatus());
    }
 
    @ManagedOperation(displayName = "Cancel Push Status",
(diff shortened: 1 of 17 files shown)


tristantarrant:
This doesn't fix ISPN-7024 since SiteDownFailoverTest inherits from AbstractXSiteTest and you've only patched AbstractTwoSitesTest.

tristantarrant:
Ignore my completely useless comment above, as the changes to AbstractXSiteTest have nothing to do with the overall fix. However, I have seen  SiteDownFailoverTest fail locally with your PR

tristantarrant:
Apologies, I had not installed infinispan-core. I should stop commenting out of my ...

tristantarrant:
Actually, this seems to fix the issue.

pruivo:
should the test-xsite profile being removed? I guess it was there only to make sure the test runs sequentially...

core/src/test/java/org/infinispan/test/fwk/TEST_RELAY2.java
```diff
@@ -0,0 +1,44 @@
...
+      super.parseSiteConfiguration(map);
+
+      String testName = TestResourceTracker.getCurrentTestName();
+      map.forEach((s, siteConfig) -> {
+         List<RelayConfig.BridgeConfig> bridges = siteConfig.getBridges();
+         for (int i = 0; i < bridges.size(); i++) {
+            RelayConfig.BridgeConfig bridgeConfig = bridges.get(i);
+            String config =
+                  (String) TestingUtil.extractField(RelayConfig.PropertiesBridgeConfig.class, bridgeConfig, "config");
+            if (testName != null) {
```
pruivo: I guess you can check the `testName` before the `forEach()`

pruivo:
`NonTxOfflineTest.testPutWithFailures` is failing when I use more than 1 thread

danberindei:
@pruivo I haven't seen `NonTxOfflineTest` fail locally, but I'm making the keys an instance field just in case.
I've was hesitating to remove the `test-xsite` profile, but I guess there's no point in delaying it.

core/src/test/java/org/infinispan/test/fwk/TEST_RELAY2.java
danberindei: I updated the code to change the cluster name in both cases.

pruivo:
`NonTxOfflineTest` seems fine now...
You have a new failure in cli-interpreter `SiteStatementTest`. I guess it was never run before... 
* The test `testContainerOperations()` needs to define `"another-cache"` in NYC site, otherwise it will ignore the xsite command (it returns a `CacheNotFoundResponse`)
* I'm unable to reproduce `testSiteStateTransfer()` locally... so no idea what is going on...

danberindei:
Updated to fix `SiteStatementTest`. I haven't reproduced `testSiteStateTransfer()` locally either, let's wait for a couple more runs in CI.

danberindei:
Updated again to add a method `AbstractXSiteTest.waitForSites()` and improve logging for failed xsite state push.

core/src/main/java/org/infinispan/xsite/XSiteAdminOperations.java
```diff
@@ -280,7 +280,7 @@ public final String pushState(@Parameter(description = "The destination site nam
       try {
          stateTransferManager.startPushState(siteName);
       } catch (Throwable throwable) {
-         log.debugf(throwable, "Unable to pushState to '%s'.", siteName);
+         log.errorf(throwable, "Unable to pushState to '%s'.", siteName);
```
pruivo: it should be in `Log` interface.

danberindei: @pruivo I'd rather keep it the same as the error string that's returned to the user.

pruivo: don't we have a policy that all error messages should be i8n?

danberindei: True... ok, I've i18n-ed the error message and added it to the other operations as well.

pruivo:
:+1:

tristantarrant:
Merged, thanks

PR #5311, opened 2017-07-20
Bump schemas to 9.2

gustavonalle:


gustavonalle:
What is the policy of schemas updates in the server? This PR bumps them to 9.2, but wasn't really sure if they should stay on 9.0

cdi/embedded/src/test/resources/infinispan.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <infinispan
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-      xsi:schemaLocation="urn:infinispan:config:9.1 http://www.infinispan.org/schemas/infinispan-config-9.1.xsd"
```
tristantarrant: I think that we should probably use resource filtering for many of these configuration files and use the infinispan.core.schema.version property which lives in the bom

gustavonalle: It crossed my mind, the downside is that IDE wwill not be able to validate the XML properly and potentially the XML will be invalid, WDYT?

danberindei: Couldn't we remove the `xsi:schemaLocation` and configure the IDE to use the local XSD for "urn:infinispan:config:9.1"? Before Alpha1 there isn't going to be a 9.2 schema on `www.infinispan.org` anyway.

gustavonalle: We could, but have you seen the local xsd? It is not valid because it contains placeholders to be processed by maven.

gustavonalle: Actually all the XSDs in the src tree are invalid; I'd prefer to have them valid, and a XSL to inject the default values extracted from the src

tristantarrant:

gustavonalle:
Rebasing and re-testing...

gustavonalle:
Rebased

cdi/embedded/src/test/resources/infinispan.xml
tristantarrant: this is a task for @ryanemerson

query/src/test/resources/configuration-parsing-test.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <infinispan
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-      xsi:schemaLocation="urn:infinispan:config:9.1 http://www.infinispan.org/schemas/infinispan-config-9.1.xsd"
-      xmlns="urn:infinispan:config:9.1">
+      xsi:schemaLocation="urn:infinispan:config:9.2 http://www.infinispan.org/schemas/infinispan-config-9.1.xsd"
```
tristantarrant: You forgot to bump the actual version in the URL (for this and many other files)

cdi/embedded/src/test/resources/infinispan.xml
ryanemerson: Isn't it possible just to point the IDE to the schema in `target/` opposed to `src/`, that way it will be a valid `.xsd` file. Granted you will have to run `mvn install` first.

query/src/test/resources/configuration-parsing-test.xml
gustavonalle: fixing

gustavonalle: probably I did not look at files changed after rebasing...

gustavonalle: Nope, my search and replace was not ideal :)

gustavonalle:
Updated

tristantarrant:
Merged, thanks !

PR #5315, opened 2017-07-24
ISPN-7919 Expose ResponseCollector in the RpcManager interface

danberindei:
https://issues.jboss.org/browse/ISPN-7919
https://issues.jboss.org/browse/ISPN-8120

I've only changed the `invokeRemotelyAsync()` calls to `invokeCommand*` in this PR, there will be at least one separate PR to address all the `invokeRemotely` (sync) calls.


extended-statistics/src/main/java/org/infinispan/stats/wrappers/ExtendedStatisticRpcManager.java
```diff
@@ -78,13 +81,79 @@ public ExtendedStatisticRpcManager(RpcManager actual, CacheStatisticManager cach
...
+      return null;
+   }
+
+   @Override
    public CompletableFuture<Map<Address, Response>> invokeRemotelyAsync(Collection<Address> recipients,
                                                                         ReplicableCommand rpc,
                                                                         RpcOptions options) {
       long start = timeService.time();
       CompletableFuture<Map<Address, Response>> future = actual.invokeRemotelyAsync(recipients, rpc, options);
-      updateStats(rpc, options.responseMode().isSynchronous(), timeService.timeDuration(start, NANOSECONDS), recipients);
```
rvansa: Uh, so extended statistics were wrong for a long while...

hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/UnorderedDistributionInterceptor.java
```diff
@@ -64,43 +61,52 @@ public Object visitPutKeyValueCommand(InvocationContext ctx, PutKeyValueCommand
...
 		}
 		else {
-			owners = writeCH.locateOwners(command.getKey());
+			List<Address> owners = cacheTopology.getDistribution(command.getKey()).writeOwners();
 			if (owners.contains(rpcManager.getAddress())) {
-            List<Address> finalOwners = owners;
-            return invokeNextAndHandle( ctx, command, (rCtx, rCommand, rv, throwable) ->
-                  invokeRemotelyAsync(finalOwners, rCtx, (WriteCommand) rCommand));
+				return invokeNextAndHandle( ctx, command, (rCtx, rCommand, rv, throwable) ->
+                  invokeRemotelyAsync(owners, rCtx, (WriteCommand) rCommand));
```
rvansa: please fix the 
```java
}
else {
```
formatting when you're modifying that part anyway, and convert that file to spaces instead of tabs

query/src/test/java/org/infinispan/query/affinity/TrackingRpcManager.java
```diff
@@ -28,6 +31,49 @@
...
+                                                Function<Address, ReplicableCommand> commandGenerator,
+                                                ResponseCollector<T> collector, RpcOptions rpcOptions) {
+      targets.forEach(target -> rpcCollector.addRPC(
+            new RpcDetail(getAddress(), commandGenerator.apply(target), cacheName, Collections.singleton(target))));
+      return delegate.invokeCommands(targets, commandGenerator, collector, rpcOptions);
+   }
+
+   @Override
+   public <T> T blocking(CompletionStage<T> request) {
+      return null;
```
rvansa: delegate?

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerInterceptor.java
```diff
@@ -315,8 +319,8 @@ public Object visitRemoveCommand(InvocationContext ctx, RemoveCommand command) t
             StringBuilder sb = new StringBuilder();
             for (FileDescriptor fd : serializationContext.getFileDescriptors().values()) {
                if (fd.isResolved()) {
-                  cmd = commandsFactory.buildRemoveCommand(fd.getName() + ERRORS_KEY_SUFFIX, null, flagsBitSet);
-                  invoker.invoke(ctx, cmd);
+                  writeCommand = commandsFactory.buildRemoveCommand(fd.getName() + ERRORS_KEY_SUFFIX, null, flagsBitSet);
```
rvansa: set topology

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerInterceptor.java
```diff
@@ -301,11 +303,13 @@ public Object visitRemoveCommand(InvocationContext ctx, RemoveCommand command) t
          if (shouldIntercept(key)) {
             // lock .errors key
             long flagsBitSet = copyFlags(command);
-            VisitableCommand cmd = commandsFactory.buildLockControlCommand(ERRORS_KEY_SUFFIX, flagsBitSet, null);
-            invoker.invoke(ctx, cmd);
+            LockControlCommand lockCommand = commandsFactory.buildLockControlCommand(ERRORS_KEY_SUFFIX, flagsBitSet, null);
+            lockCommand.setTopologyId(command.getTopologyId());
+            invoker.invoke(ctx, lockCommand);
```
rvansa: could you file a JIRA to make this async?

core/src/main/java/org/infinispan/configuration/cache/ClusteringConfiguration.java
```diff
@@ -36,11 +38,19 @@ static AttributeSet attributeDefinitionSet() {
...
+      this.invalidationBatchSize = attributes.attribute(INVALIDATION_BATCH_SIZE);
       this.hashConfiguration = hashConfiguration;
       this.l1Configuration = l1Configuration;
       this.stateTransferConfiguration = stateTransferConfiguration;
       this.syncConfiguration = syncConfiguration;
       this.partitionHandlingConfiguration  = partitionHandlingStrategy;
+
+      // Expose the true value for users of attributes().attribute(REMOTE_TIMEOUT).get()
+      // and attributes().attribute(REMOTE_TIMEOUT).addListener()
+      syncConfiguration.attributes().attribute(SyncConfiguration.REPL_TIMEOUT)
```
danberindei: I'll create a separate JIRA, but the changes have to be in this PR because `RpcManager.getSyncRpcOptions()` really needs the `remoteTimeout` listener to work -- no test actually checked the listeners before.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -119,8 +121,8 @@ public final Object visitGetKeysInGroupCommand(InvocationContext ctx, GetKeysInG
          return invokeNext(ctx, command);
       }
       Address primaryOwner = dm.getCacheTopology().getDistribution(command.getGroupName()).primary();
-      CompletableFuture<Map<Address, Response>> future = rpcManager.invokeRemotelyAsync(
-            Collections.singleton(primaryOwner), command, defaultSyncOptions);
+      CompletionStage<Map<Address, Response>> future = rpcManager
+            .invokeCommand(primaryOwner, command, new MapResponseCollector(false, 1), rpcManager.getSyncRpcOptions());
```
danberindei: Actually I wanted to use `SingleResponseCollector` for all the RPCs with a single target, but I only changed a few so far. Of course, I don't want to use that for long either: ideally I'd like to use custom `ResponseCollector` implementations everywhere.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -171,33 +178,33 @@ public final Object visitClearCommand(InvocationContext ctx, ClearCommand comman
...
+                                return;
+                             }
+                             InternalCacheEntry ice = ((InternalCacheValue) responseValue).toInternalCacheEntry(key);
+                             if (rvrl != null) {
+                                rvrl.remoteValueFound(ice);
+                             }
+                             wrapRemoteEntry(ctx, key, ice, isWrite);
+                             return;
+                          }
+                          throw handleMissingSuccessfulResponse(r);
```
danberindei: No, `UnsureResponse`s are ignored by `RemoteGetResponseCollector`. This handler only gets an `UnsureResponse` if there were no successful/exceptional responses.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -848,15 +866,14 @@ public Object visitReadOnlyKeyCommand(InvocationContext ctx, ReadOnlyKeyCommand
...
-               }
-               throw handleMissingSuccessfulResponse(responses);
-            }));
+         CompletionStage<Response> rpc = rpcManager
+               .invokeCommand(owners, remoteCommand, new RemoteGetResponseCollector(), rpcManager.getSyncRpcOptions());
+         return asyncValue(rpc.thenApply(rsp -> {
+            if (rsp.isSuccessful()) {
+               return unwrapFunctionalResultOnOrigin(ctx, key, ((SuccessfulResponse) rsp).getResponseValue());
+            }
+            throw handleMissingSuccessfulResponse(rsp);
```
danberindei: That's exactly what happens.

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
```diff
@@ -271,8 +274,9 @@ public void injectDependencies(GroupManager groupManager, ScatteredVersionManage
       Address backup = getNextMember(cacheTopology);
       if (backup != null) {
          // error responses throw exceptions from JGroupsTransport
-         CompletableFuture<Map<Address, Response>> rpcFuture =
-            rpcManager.invokeRemotelyAsync(Collections.singletonList(backup), command, defaultSyncOptions);
+         MapResponseCollector collector = new MapResponseCollector(false, 1);
+         CompletionStage<Map<Address, Response>> rpcFuture =
+               rpcManager.invokeCommand(backup, command, collector, rpcManager.getSyncRpcOptions());
```
danberindei: I usually let IntelliJ decide for me, but I agree this looks better.

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -597,7 +610,7 @@ protected RpcOptions createRpcOptions() {
 
    private class TxReadOnlyManyHelper extends ReadOnlyManyHelper {
       @Override
-      public ReplicableCommand copyForRemote(ReadOnlyManyCommand command, List<Object> keys, InvocationContext ctx) {
+      public ReadOnlyManyCommand<?, ?, ?> copyForRemote(ReadOnlyManyCommand command, List<Object> keys, InvocationContext ctx) {
```
danberindei: Hey, complain to the guy who made `ReadOnlyManyCommand` generic :)

core/src/main/java/org/infinispan/interceptors/impl/InvalidationInterceptor.java
```diff
@@ -209,8 +214,9 @@ public Object visitLockControlCommand(TxInvocationContext ctx, LockControlComman
          LockControlCommand lockControlCommand = (LockControlCommand) rCommand;
          boolean sync = !lockControlCommand.isUnlock();
          ((LocalTxInvocationContext) rCtx).remoteLocksAcquired(rpcManager.getTransport().getMembers());
-         CompletableFuture<Map<Address, Response>> remoteInvocation =
-               rpcManager.invokeRemotelyAsync(null, lockControlCommand, getBroadcastRpcOptions(sync));
+         CompletionStage<Map<Address, Response>> remoteInvocation =
+               rpcManager.invokeCommandOnAll(lockControlCommand, new MapResponseCollector(true),
```
danberindei: Indeed, I'm writing a `VoidResponseCollector` for this.

core/src/main/java/org/infinispan/interceptors/impl/InvalidationInterceptor.java
```diff
@@ -308,13 +315,14 @@ public Object visitPutMapCommand(InvocationContext ctx, PutMapCommand command) t
          // If the cache uses 2PC it's possible that the remotes will commit the invalidation and the originator rolls back,
          // but this does not impact consistency and the speed benefit is worth it.
          command = commandsFactory.buildPrepareCommand(txCtx.getGlobalTransaction(), Collections.singletonList(invalidateCommand), true);
+         command.setTopologyId(invalidateCommand.getTopologyId());
+      }
+      if (synchronous) {
```
danberindei: Yeah, it is annoying to have these extra checks in the interceptors. I still like the idea of keeping the sync and async cases separate in `RpcManager`, but maybe I should add something to `BaseRpcInterceptor` instead.

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
```diff
@@ -20,6 +24,78 @@
...
+   <T> CompletionStage<T> invokeCommands(Collection<Address> targets,
+                                         Function<Address, ReplicableCommand> commandGenerator,
+                                         ResponseCollector<T> collector, RpcOptions rpcOptions);
+
+   /**
+    * Block on a request and return its result.
+    *
+    * @since 9.2
+    */
+   <T> T blocking(CompletionStage<T> request);
```
danberindei: How would using a `CompletableFuture` argument improve this method?

core/src/main/java/org/infinispan/remoting/rpc/RpcManagerImpl.java
```diff
@@ -85,8 +98,26 @@ private void start() {
...
+   @Stop
+   private void stop() {
+      configuration.clustering()
+                   .attributes().attribute(ClusteringConfiguration.REMOTE_TIMEOUT)
+                   .removeListener(updateRpcOptions);
+   }
+
+   private void updateRpcOptions(Attribute<Long> attribute, Long oldValue) {
+      syncRpcOptions = new RpcOptions(DeliverOrder.NONE, attribute.get(), TimeUnit.MILLISECONDS);
+      totalSyncRpcOptions = new RpcOptions(DeliverOrder.TOTAL, attribute.get(), TimeUnit.MILLISECONDS);
```
danberindei: I think not, but I wanted the name to say that it's sync so I didn't have to check :)

core/src/main/java/org/infinispan/remoting/rpc/RpcManagerImpl.java
```diff
@@ -107,23 +138,164 @@ public String getPendingViewAsString() {
...
+
+      if (!statisticsEnabled) {
+         return t.invokeCommand(target, cacheRpc, collector, rpcOptions.deliverOrder(),
+                                rpcOptions.timeout(), rpcOptions.timeUnit());
+      }
+
+      long startTimeNanos = timeService.time();
+      CompletionStage<T> invocation;
+      try {
+         invocation = t.invokeCommand(target, cacheRpc, collector, rpcOptions.deliverOrder(),
```
danberindei: I have already decided to do that, but there are some leftovers from `invokeRemotelyAsync` lying around.

danberindei: Actually that's not true, I only did it for target-side exceptions. If `JChannel.send` throws a `RuntimeException` it'll reach the caller directly, but I think that's fine.

core/src/main/java/org/infinispan/remoting/rpc/RpcManagerImpl.java
```diff
@@ -107,23 +138,164 @@ public String getPendingViewAsString() {
...
+                                      rpcOptions.timeout(), rpcOptions.timeUnit());
+      } catch (Exception e) {
+         return errorReplicating(e);
+      }
+      return invocation.handle((response, throwable) -> updateStatistics(startTimeNanos, response, throwable));
+   }
+
+   private void checkTopologyId(ReplicableCommand command) {
+      if (command instanceof TopologyAffectedCommand && ((TopologyAffectedCommand) command).getTopologyId() < 0) {
+         throw new IllegalArgumentException("Command does not have a topology id");
```
danberindei: It would work, it's just that I don't like `assert` and the fact that it's not active all the time. Too many bad memories running tests without `-ea` from long time ago :)

core/src/main/java/org/infinispan/remoting/rpc/RpcManagerImpl.java
```diff
@@ -165,72 +328,34 @@ public String getPendingViewAsString() {
...
-         if (statisticsEnabled) replicationCount.incrementAndGet();
-         if (trace) log.tracef("Response(s) to %s is %s", rpcs, result);
-         return result;
-      } catch (InterruptedException e) {
-         Thread.currentThread().interrupt();
-         throw new CacheException("Thread interrupted while invoking RPC", e);
+         Function<Address, ReplicableCommand> commandGenerator = address -> {
+            ReplicableCommand rpc = rpcs.get(address);
+            // Set the topology id of the command, in case we don't have it yet
+            setTopologyId(rpc);
```
danberindei: I'm not going to change the behaviour of existing methods in this PR.

core/src/main/java/org/infinispan/remoting/rpc/RpcManagerImpl.java
```diff
@@ -259,25 +380,34 @@ public void sendTo(Address destination, ReplicableCommand command, DeliverOrder
 
    @Override
    public void sendToMany(Collection<Address> destinations, ReplicableCommand command, DeliverOrder deliverOrder) {
-      if (trace) {
-         log.tracef("%s invoking %s to list %s ordered by %s", t.getAddress(), command, destinations, deliverOrder);
+      // Set the topology id of the command, in case we don't have it yet
+      setTopologyId(command);
```
danberindei: I'm not going to change the behaviour of existing methods in this PR.

core/src/main/java/org/infinispan/remoting/rpc/RpcOptions.java
```diff
@@ -19,6 +21,14 @@
    private final DeliverOrder deliverOrder;
 
    /**
+    * @since 9.2
+    */
+   @SuppressWarnings("deprecation")
```
danberindei: Because it's delegating to a deprecated constructor.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
rvansa: I see, `r` is then generated by the `finish` method. That's a bit unintuitive, and requires an extra check in here: what about passing successful and unsure response handlers directly to the collector? Unsure would be called only during finish, ofc.

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
rvansa: Java gives us the liberty to use or not use generics when appropriate. In the stack we have no clues about key/value types, so it's better to omit them here.

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
rvansa: Using CF.join(). But I've noticed that that the impl uses non-spinning variant, so we probably need the utility method. I would still rather put the helper into `CompletableFutures` than to `RpcManager`.

core/src/main/java/org/infinispan/remoting/rpc/RpcOptions.java
rvansa: yes, but why do you suppress the warning at all?

core/src/test/java/org/infinispan/tx/dld/ControlledRpcManager.java
```diff
@@ -125,11 +122,11 @@ protected Object beforeInvokeRemotely(ReplicableCommand command) {
    }
 
    @Override
-   protected Map<Address, Response> afterInvokeRemotely(ReplicableCommand command, Map<Address, Response> responseMap, Object argument) {
+   protected Object afterInvokeRemotely(ReplicableCommand command, Object responseObject, Object argument) {
```
danberindei: I thought it would make it harder for implementers, but it looks about the same so I'm changing it.

extended-statistics/src/main/java/org/infinispan/stats/wrappers/ExtendedStatisticRpcManager.java
danberindei: Unfortunately yes.

hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/UnorderedDistributionInterceptor.java
danberindei: I would prefer keeping these files with the original formatting. If we do decide to change the formatting, I'd prefer a separate commit changing it everywhere.

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerInterceptor.java
danberindei: Why not, https://issues.jboss.org/browse/ISPN-8136

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerInterceptor.java
danberindei: Actually it shouldn't be necessary to set the topology here. The command will go through the `StateTransferInterceptor`, which will both set the topology and retry if necessary, so I'll remove all the `setTopologyId()` calls from this interceptor.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
danberindei: I don't see a need for an extra check, can you elaborate?

As the callback is added to the request I think it should be pretty clear that it can't see responses before the request is completed. In principle I'm open to the idea of using more customized collectors, but I don't see the case for it here.

rvansa: There already is a check in this handler, on line 187: `if (r instanceof SuccessfulResponse) {`

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
danberindei: @rvansa IMO it's better to say explicitly that we don't care what the key/value types are with wildcards, instead of pretending that the command is not generic.

It's true that in this case the command is basically only used in the interceptor stack, where we don't know what the key/value types are, so it's probably better to pretend it was never generic, but doing that means more compilation warnings.

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
danberindei: The reason I put it in `RpcManager` is that the exception re-wrapping seemed RPC-specific. I'll experiment with moving it to `CompletableFutures`, but in the meantime I've marked it as `@Experimental` and I hope to remove all its uses before 9.2.

core/src/main/java/org/infinispan/remoting/rpc/RpcOptions.java
danberindei: Because it means one less warning, and warnings are annoying?

core/src/main/java/org/infinispan/remoting/rpc/RpcManagerImpl.java
```diff
@@ -259,25 +380,34 @@ public void sendTo(Address destination, ReplicableCommand command, DeliverOrder
...
       }
    }
 
-   private void errorReplicating(Exception e) {
-      log.unexpectedErrorReplicating(e);
+   private <T> T errorReplicating(Throwable t) {
+      log.unexpectedErrorReplicating(t);
       if (statisticsEnabled) replicationFailures.incrementAndGet();
-      rethrowAsCacheException(e);
+      return rethrowAsCacheException(t);
```
danberindei: It looks the same to me, so I'd rather not change it here.

danberindei:
Closing for now to free some CI resources.

PR #5317, opened 2017-07-25
ISPN-8097 ScatteredCrashInSequenceTest.testSplit failures

rvansa:
https://issues.jboss.org/browse/ISPN-8097

### DIFF
--- a/core/src/main/java/org/infinispan/commands/write/InvalidateVersionsCommand.java
+++ b/core/src/main/java/org/infinispan/commands/write/InvalidateVersionsCommand.java
@@ -6,7 +6,12 @@ import org.infinispan.container.DataContainer;
 import org.infinispan.container.versioning.InequalVersionComparisonResult;
 import org.infinispan.container.versioning.SimpleClusteredVersion;
 import org.infinispan.persistence.manager.OrderedUpdatesManager;
+import org.infinispan.statetransfer.StateTransferLock;
+import org.infinispan.statetransfer.StateTransferManager;
 import org.infinispan.util.ByteString;
+import org.infinispan.util.concurrent.CompletableFutures;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
 
 import java.io.IOException;
 import java.io.ObjectInput;
@@ -19,8 +24,13 @@ import java.util.concurrent.CompletableFuture;
  * @author Radim Vansa &lt;rvansa@redhat.com&gt;
  */
 public class InvalidateVersionsCommand extends BaseRpcCommand {
+   private static final Log log = LogFactory.getLog(InvalidateVersionsCommand.class);
+   private static final boolean trace = log.isTraceEnabled();
+
    public static final int COMMAND_ID = 67;
 
+   // This is the topologyId in which this command is valid (in case that it comes from the state transfer)
+   private int topologyId;
    private Object[] keys;
    private int[] topologyIds;
    private long[] versions;
@@ -29,6 +39,8 @@ public class InvalidateVersionsCommand extends BaseRpcCommand {
 
    protected DataContainer dataContainer;
    protected OrderedUpdatesManager orderedUpdatesManager;
+   protected StateTransferLock stateTransferLock;
+   protected StateTransferManager stateTransferManager;
 
    public InvalidateVersionsCommand() {
       this(null);
@@ -38,37 +50,65 @@ public class InvalidateVersionsCommand extends BaseRpcCommand {
       super(cacheName);
    }
 
-   public InvalidateVersionsCommand(ByteString cacheName, Object[] keys, int[] topologyIds, long[] versions, boolean removed) {
+   public InvalidateVersionsCommand(ByteString cacheName, int topologyId, Object[] keys, int[] topologyIds, long[] versions, boolean removed) {
       super(cacheName);
+      this.topologyId = topologyId;
       this.keys = keys;
       this.topologyIds = topologyIds;
       this.versions = versions;
       this.removed = removed;
    }
 
-   public void init(DataContainer dataContainer, OrderedUpdatesManager orderedUpdatesManager) {
+   public void init(DataContainer dataContainer, OrderedUpdatesManager orderedUpdatesManager,
+                    StateTransferLock stateTransferLock, StateTransferManager stateTransferManager) {
       this.dataContainer = dataContainer;
       this.orderedUpdatesManager = orderedUpdatesManager;
+      this.stateTransferLock = stateTransferLock;
+      this.stateTransferManager = stateTransferManager;
    }
 
    @Override
    public CompletableFuture<Object> invokeAsync() {
-      for (int i = 0; i < keys.length; ++i) {
-         Object key = keys[i];
-         if (key == null) break;
-         SimpleClusteredVersion version = new SimpleClusteredVersion(topologyIds[i], versions[i]);
-         dataContainer.compute(key, (k, oldEntry, factory) -> {
-            if (oldEntry == null) {
-               return null;
+      if (stateTransferLock != null) {
+         stateTransferLock.acquireSharedTopologyLock();
+      }
+      try {
+         if (topologyId >= 0 && stateTransferManager != null) {
+            int currentTopologyId = stateTransferManager.getCacheTopology().getTopologyId();
+            if (topologyId > currentTopologyId) {
+               if (trace) {
+                  log.tracef("Delaying command %s, current topology id %d", currentTopologyId);
+               }
+               return stateTransferLock.topologyFuture(topologyId).thenCompose(nil -> invokeAsync());
+            } else if (topologyId < currentTopologyId) {
+               log.ignoringInvalidateVersionsFromOldTopology(this.topologyId, currentTopologyId);
+               if (trace) {
+                  log.tracef("Ignored command is %s", this);
+               }
+               return CompletableFutures.completedNull();
             }
-            SimpleClusteredVersion localVersion = (SimpleClusteredVersion) oldEntry.getMetadata().version();
-            InequalVersionComparisonResult result = localVersion.compareTo(version);
-            if (result == InequalVersionComparisonResult.BEFORE || (removed && result == InequalVersionComparisonResult.EQUAL)) {
-               return null;
-            } else {
-               return oldEntry;
-            }
-         });
+         }
+         for (int i = 0; i < keys.length; ++i) {
+            Object key = keys[i];
+            if (key == null) break;
+            SimpleClusteredVersion version = new SimpleClusteredVersion(topologyIds[i], versions[i]);
+            dataContainer.compute(key, (k, oldEntry, factory) -> {
+               if (oldEntry == null) {
+                  return null;
+               }
+               SimpleClusteredVersion localVersion = (SimpleClusteredVersion) oldEntry.getMetadata().version();
+               InequalVersionComparisonResult result = localVersion.compareTo(version);
+               if (result == InequalVersionComparisonResult.BEFORE || (removed && result == InequalVersionComparisonResult.EQUAL)) {
+                  return null;
+               } else {
+                  return oldEntry;
+               }
+            });
+         }
+      } finally {
+         if (stateTransferLock != null) {
+            stateTransferLock.releaseSharedTopologyLock();
+         }
       }
       return orderedUpdatesManager.invalidate(keys).thenApply(nil -> null);
    }
@@ -90,6 +130,7 @@ public class InvalidateVersionsCommand extends BaseRpcCommand {
 
    @Override
    public void writeTo(ObjectOutput output) throws IOException {
+      output.writeInt(topologyId);
       // TODO: topology ids are mostly the same - sort the arrays according to topologyIds and use compaction encoding
       output.writeInt(keys.length);
       for (int i = 0; i < keys.length; ++i) {
@@ -107,6 +148,7 @@ public class InvalidateVersionsCommand extends BaseRpcCommand {
 
    @Override
    public void readFrom(ObjectInput input) throws IOException, ClassNotFoundException {
+      topologyId = input.readInt();
       keys = new Object[input.readInt()];
       topologyIds = new int[keys.length];
       versions = new long[keys.length];
@@ -124,7 +166,8 @@ public class InvalidateVersionsCommand extends BaseRpcCommand {
 
    @Override
    public String toString() {
-      StringBuilder sb = new StringBuilder("InvalidateVersionsCommand{removed=").append(removed).append(": ");
+      StringBuilder sb = new StringBuilder("InvalidateVersionsCommand{topologyId=").append(topologyId)
+            .append(", removed=").append(removed).append(": ");
       if (keys.length > 0 && keys[0] != null) {
          sb.append(keys[0]).append(" -> ").append(versions[0]);
       } else {
--- a/core/src/main/java/org/infinispan/container/entries/RemoteMetadata.java
+++ b/core/src/main/java/org/infinispan/container/entries/RemoteMetadata.java
@@ -14,6 +14,11 @@ import java.io.ObjectOutput;
 import java.util.Set;
 
 /**
+ * This is a metadata type used by scattered cache during state transfer. The address points to node which has last
+ * known version of given entry: During key transfer RemoteMetadata is created and overwritten if another response
+ * with higher version comes. During value transfer the address is already final and we request the value + metadata
+ * only from this node.
+ *
  * @author Radim Vansa &lt;rvansa@redhat.com&gt;
  */
 public class RemoteMetadata implements InternalMetadata {
--- a/core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
@@ -156,7 +156,7 @@ public class ScatteredDistributionInterceptor extends ClusteringInterceptor {
 
       DistributionInfo info = cacheTopology.getDistribution(command.getKey());
       if (info.primary() == null) {
-         throw AllOwnersLostException.INSTANCE;
+         throw OutdatedTopologyException.INSTANCE;
       }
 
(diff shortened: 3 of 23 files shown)


rvansa:
(not sure if the test fix label applies, since it's not a test problem)

pruivo:
basically, the only change it setting the topology id in the get. It looks good to me :) 
IMO, the test-fix label applies if you are fixing a test that is randomly (or not) failing :)

pruivo:
there are still failures in CI `ScatteredCrashInSequenceTest`

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
```diff
@@ -207,68 +214,97 @@ private Object prefetchKeysIfNeededAndInvokeNext(InvocationContext ctx, Visitabl
...
-         } else if ((metadata instanceof RemoteMetadata)) {
-            Address backup = ((RemoteMetadata) metadata).getAddress();
-            future = retrieveRemoteValue(Collections.singleton(backup), key);
-         } else {
-            future = retrieveRemoteValue(null, key);
+      // We want to throw OTE with the proper requested topology early if the command's topology does not match.
+      // Moreover, OTE thrown this way has requested topology id set; all queries about owner etc. must match
+      // command's topology id. If it did not the cache topology could change between SDI and STI
+      // (when the exception pops up) and we would set next topology id too high, while the OTE/AOLE was thrown
+      // due to older topology being present in SDI.
```
danberindei: I'd rather point the reader to the `ScatteredDistributionInterceptor` (?) method that checks the topology id and throws the OTE instead. This looks like too much detail and too far from the implementation to stay true for long.

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
```diff
@@ -207,68 +214,97 @@ private Object prefetchKeysIfNeededAndInvokeNext(InvocationContext ctx, Visitabl
...
       if (transferedKeys != null) {
-         return asyncInvokeNext(ctx, command, retrieveRemoteValues(ctx, transferedKeys).toCompletableFuture());
+         return asyncInvokeNext(ctx, command, retrieveRemoteValues(ctx, transferedKeys, command.getTopologyId()).toCompletableFuture());
       } else {
          return invokeNext(ctx, command);
       }
    }
 
-   private InvocationStage retrieveSingleRemoteValue(InvocationContext ctx, Object key) {
+   private InvocationStage retrieveSingleRemoteValue(InvocationContext ctx, Object key, int topologyId) {
```
danberindei: Is the method name right? At the very least, it's too similar to `retrieveRemoteValue`, which is quite different.

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
```diff
@@ -207,68 +214,97 @@ private Object prefetchKeysIfNeededAndInvokeNext(InvocationContext ctx, Visitabl
...
    }
 
-   private InvocationStage retrieveRemoteValues(InvocationContext ctx, List<?> keys) {
+   // TODO: this is not completely aligned with single-entry prefetch
+   private InvocationStage retrieveRemoteValues(InvocationContext ctx, List<?> keys, int topologyId) {
+      if (trace) {
+         log.tracef("Prefetching entries for keys %s using broadcast", keys);
+      }
       ClusteredGetAllCommand command = commandsFactory.buildClusteredGetAllCommand(keys, FlagBitSets.SKIP_OWNERSHIP_CHECK, null);
+      command.setTopologyId(topologyId);
```
danberindei: No explanation needed here?

rvansa: TBH I haven't seen any problems with that but I thought that adding the id won't harm. Should i add `// not really tested`? :)

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
rvansa: Right, it sounds a bit confusing. `retrieveSingleRemoteValue` runs first the local lookup (for the case that we already have the key) and then invokes `retrieveRemoteValue`. How would you like to call this method which does first local and then remote lookup? Just `retrieveSingleValue`?

Writing this comment I've realized that it may not work properly, though; Using the RemoteMetadata is correct only during the value transfer, but during key transfer the metadata could point to a node with outdated copy. I'll change that.

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
danberindei: It's just odd to have such a long explanation in one method and no explanation in the another one when they do the same thing :)

TBH it's probably better to not have a comment in `PrefetchInterceptor`, and have `ScatteredDistributionInterceptor` throw an exception if the topology id is not set. In https://github.com/infinispan/infinispan/pull/5315 I also made `RpcManager.invokeCommand*()` throw an exception if the topology id is not set instead of using the current topology id, like `RpcManager.invokeRemotely()` does.

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
danberindei: I'd probably call it `invokeGetCacheEntryCommand`.

rvansa: That (naming) wouldn't be right: it invokes the command *and* then does the remote lookup. The returned InvocationStage is completed when both those complete.

rvansa:
I've been running the tests 100 times w/ trace logs and 12 times w/o logs without failure :-/

rvansa:
Updated, setting the topology id all the time, including for local commands (those could do the retry now, too - I think that it's little price for reduced complexity).

danberindei:
@rvansa I don't think it's conceptually ok to throw an OTE for `CACHE_MODE_LOCAL` commands, because you're not doing an RPC, but you're still waiting for another RPC to happen so that the new topology is installed.

There's also a problem with `throw new OutdatedTopologyException(command.getTopologyId())` in `ScatteredDistributionInterceptor.checkTopology()`. There's an undocumented assumption in `(Base)StateTransferInterceptor` that the `requestedTopologyId` of any caught OTE is always greater than the command's initial topology id, and some of the `handleXReturn` methods ignore it, using `command.getTopologyId() + 1` all the time.

OTOH the `(Base)StateTransferInterceptor.handleXReturn` methods that do not ignore the `requestedTopologyId` won't block at all, because the command's initial topology id is guaranteed to be less than the current topology id.

And finally, what does it mean for an exception to "pop up"?

rvansa:
I agree that you must not wait for RPC with CACHE_MODE_LOCAL command, but I thought that the OTE is thrown only if the topology is changed between STI and xDI. Then we should retry immediatelly, because the next topology should be already installed. You can't have OTEs originating remotely.

Though, there really seems to be something fishy about BaseSTI.nextTopology and getNextTopologyId, I'll check it out.

Exception popping up means being propagated up the stack. I'll change the wording.

danberindei:
@rvansa I was thinking about when the primary owner is `null`, but I guess that would throw an AOLE instead of an OTE. Either way, a `CACHE_MODE_LOCAL` command should not wait for a new topology with a non-null primary owner.

rvansa:
@danberindei OK, I'll just remove the topologyId check for local commands in distribution interceptors, does that sound good to you?

tristantarrant:
Looks like this is still failing http://ci.infinispan.org/job/Infinispan/job/PR-5317/7/testReport/junit/org.infinispan.partitionhandling/ScatteredCrashInSequenceTest/testSplit5_SCATTERED_SYNC_/

rvansa:
@tristantarrant Yes, I've noticed that CI is still failing. So this will be a fix for the problems that I can reproduce locally.

danberindei:
@rvansa Skipping the topology check for local commands sounds good. I'll keep the JIRA open when I integrate this :)

pruivo:
@rvansa I have the `testSplit2` failing when I run the test class with single thread.

pruivo:
trace logs of `testSplit4`
https://drive.google.com/file/d/0ByatUPYii_o5UVpEWHFfZnFJekk/view?usp=sharing

I've noticed that NodeC receives an InvalidateVersionCommand right after applied the state transfer... I'm not sure if it was supposed to do that or not...

rvansa:
@pruivo Looks like you've hit even different issue than the one in CI. I'll fix that as well.

rvansa:
Okay, I've fixed both the InvalidateVersionsCommand coming from the past topology, and the other issue that had troubled CI: the test was using `StateSequencer` to block rebalance confirmations, but if none of them came it got stuck as well. Replaced with one-way synchronization.

pruivo:
@rvansa checkstyle error in CI

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
```diff
@@ -714,7 +712,7 @@ private Object handlePutMapOnOrigin(InvocationContext ctx, PutMapCommand command
             localEntries.put(key, new MetadataImmortalCacheValue(entry.getValue(), metadata));
             commitSingleEntryIfNewer(ctxEntry, ctx, command);
          } else if (info.primary() == null) {
-            throw AllOwnersLostException.INSTANCE;
+            throw new AllOwnersLostException(cacheTopology.getTopologyId());
```
rvansa: There's no owner in CH - the key is lost.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
```diff
@@ -237,13 +232,24 @@ private Object handleReadCommandReturn(InvocationContext rCtx, VisitableCommand
...
          // we can return null only after the next topology is installed. If partition handling is enabled we decide
          // only based on the availability status.
          // In other cache modes, during partition the exception is already handled in PartitionHandlingInterceptor,
          // and if the handling is not enabled, we can't but return null.
          if (cacheConfiguration.clustering().cacheMode().isScattered()) {
-            requestedTopologyId = nextTopology(cmd, currentTopologyId);
+            AllOwnersLostException aole = (AllOwnersLostException) ce;
+            // It is possible that a command arrives with higher topology id than the one of current node; such
+            // command can cause execution of another command with CACHE_MODE_LOCAL flag (not checking topology id
+            // vs. command id) and this could find no owner in the current cache topology.
```
rvansa: It was your requirement that a read command never blocks, waiting for another topology.

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
```diff
@@ -207,68 +214,96 @@ private Object prefetchKeysIfNeededAndInvokeNext(InvocationContext ctx, Visitabl
...
+      }
+      CacheEntry entry = (CacheEntry) rv;
+      Metadata metadata = entry != null ? entry.getMetadata() : null;
+      CompletableFuture<InternalCacheValue> future;
+      int segment = keyPartitioner.getSegment(cmd.getKey());
+      if (metadata != null && metadata.version() != null && svm.isVersionActual(segment, metadata.version())){
+         return null;
+      } else if ((metadata instanceof RemoteMetadata) && svm.getSegmentState(segment) == ScatteredVersionManager.SegmentState.VALUE_TRANSFER) {
+         // The RemoteMetadata is valid only during value transfer - in blocked state there shouldn't be any such
+         // entry and during key transfer we could see metadata pointing to a node with outdated information.
```
rvansa: It's not outdated RemoteMetadata, but RM pointing to an outdated copy. During key transfer all nodes send their versions and RM is added/overwritten if another node has higher version. Until we get all copies (and we're not sure about it until value transfer) there might be a RM that comes from a node with outdated information.

core/src/main/java/org/infinispan/commands/write/InvalidateVersionsCommand.java
```diff
@@ -38,39 +45,63 @@ public InvalidateVersionsCommand(ByteString cacheName) {
...
+               SimpleClusteredVersion localVersion = (SimpleClusteredVersion) oldEntry.getMetadata().version();
+               InequalVersionComparisonResult result = localVersion.compareTo(version);
+               if (result == InequalVersionComparisonResult.BEFORE || (removed && result == InequalVersionComparisonResult.EQUAL)) {
+                  return null;
+               } else {
+                  return oldEntry;
+               }
+            });
+         }
+         return orderedUpdatesManager.invalidate(keys).handle((nil, t) -> {
```
rvansa: what's the difference between DC and storing in persistence?

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
danberindei: The way I see it, a key is lost iif `cache.get(key)` returns `null` or throws an `AvailabilityException`. Scattered cache may have a single owner in the CH, but it behaves as if it had 1 primary owner and 1 backup owner.

core/src/main/java/org/infinispan/commands/write/InvalidateVersionsCommand.java
danberindei: I wasn't thinking about persistence, but about not blocking while holding the shared topology lock in general.

`StateTransferLockImpl.ownershipLock` is non-fair, so keeping the read lock for a long time may delay write lock acquisition (IOW topology updates) for a long time. Of course, it may be simpler to make `ownershipLock` fair, but then we'd have to profile it and make sure it's not slower than then non-fair version when no locks are acquired.

core/src/main/java/org/infinispan/interceptors/impl/BaseStateTransferInterceptor.java
danberindei: That was for distributed mode. I thought scattered mode always had to block until the primary (as determined by the originator) has the value, and the primary can't apply the value if it doesn't have the right topology.

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
danberindei: Maybe you could add some javadoc to `RemoteMetadata` to explain what `getAddress()` means during each phase?

rvansa: added

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
rvansa: Still, by definition of "all owners" this is what happens. I prefer to throw OTE on events where we detect that current topology id does not match to command topology id, rather than any time we don't know what else to do and hope we could know it later.

rvansa:
@pruivo @danberindei Forgot to comment, the checkstyle problems have been fixed and now there's Success in CI (First time I see that!)

pruivo:
@rvansa success? I see a new failure:
http://ci.infinispan.org/job/Infinispan/job/PR-5317/24/testReport/org.infinispan.scattered.statetransfer/PushTransferTest/testNodeJoin/

rvansa:
@pruivo @danberindei Ok, this time it asked for a bigger overhaul. I've found out that rebalance confirmations can be sent before the push-transfer (incl. invalidation) finishes, so I've adjusted `StateTransferProvider/Consumer#onTopology` to return CF and confirm the rebalance only after both complete the update, instead of calling `StateTransferManager.notifyRebalanceEnd` which I've removed.

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
danberindei: I guess I'm fixating too much on the "lost" part :)

But IMO we don't need a separate exception every time we need to tell `StateTransferInterceptor` to wait for a new topology, because that just adds duplicate code in `StateTransferInterceptor`. Except here the topology id in `AllOwnersLostException` is the "known bad" topology id and in `OutdatedTopologyException` it's the "expected" topology id, so it's going to be harder to figure out if `STI` is trying to do the same thing for both in the future.

core/src/main/java/org/infinispan/container/entries/RemoteMetadata.java
```diff
@@ -14,6 +14,11 @@
 import java.util.Set;
 
 /**
+ * This is a metadata type used by scattered cache during state transfer. The address points to node which has last
+ * known version of given entry: During key transfer RemoteMetadata is created and overwritten if another response
+ * with higher version comes. During value transfer the address is already final and we request the value + metadata
+ * only from this node.
```
danberindei: Nice, this is exactly what I was looking for!

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
```diff
@@ -207,68 +214,96 @@ private Object prefetchKeysIfNeededAndInvokeNext(InvocationContext ctx, Visitabl
...
+         }
+         SuccessfulResponse successfulResponse = (SuccessfulResponse) response;
+         InternalCacheValue icv = (InternalCacheValue) successfulResponse.getResponseValue();
+         if (icv == null) {
+            continue;
+         }
+         Metadata metadata = icv.getMetadata();
+         if (metadata instanceof RemoteMetadata) {
+            // not sure if this can happen, but let's be on the safe side
+            throw OutdatedTopologyException.INSTANCE;
```
danberindei: I think it can happen if the remote get is delayed a lot and the key ownership changes from A to B and then to A again.

rvansa: Right, topology check should make this impossible.

danberindei:
Integrated, thanks Radim!

PR #5328, opened 2017-07-31
ISPN-8137 computeIfPresent Ignoring radom failing test

karesti:
https://issues.jboss.org/browse/ISPN-8137

### DIFF
--- a/core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
+++ b/core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
@@ -352,10 +352,11 @@ public class DistSyncFuncTest extends BaseDistFunctionalTest<Object, String> {
       expectException(RemoteException.class, () -> getFirstNonOwner("k1").computeIfPresent("k1", mappingToException));
 
       // remove if after compute value is null
-      retval = getFirstNonOwner("k1").computeIfPresent("k1", (v1, v2) -> null);
-      asyncWait("k1", ComputeCommand.class);
-      if (testRetVals) assertNull(retval);
-      assertRemovedOnAllCaches("k1");
+      // TODO: Fix https://issues.jboss.org/browse/ISPN-8074
+//      retval = getFirstNonOwner("k1").computeIfPresent("k1", (v1, v2) -> null);
+//      asyncWait("k1", ComputeCommand.class);
+//      if (testRetVals) assertNull(retval);
+//      assertRemovedOnAllCaches("k1");
 
       // do nothing if absent
       retval = getFirstNonOwner("notThere").computeIfPresent("notThere", (k, v) -> "add_" + k);


core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
```diff
@@ -352,10 +352,11 @@ public void testComputeIfPresentFromNonOwner() throws InterruptedException {
...
       // remove if after compute value is null
-      retval = getFirstNonOwner("k1").computeIfPresent("k1", (v1, v2) -> null);
-      asyncWait("k1", ComputeCommand.class);
-      if (testRetVals) assertNull(retval);
-      assertRemovedOnAllCaches("k1");
+      // TODO: Fix https://issues.jboss.org/browse/ISPN-8074
+//      retval = getFirstNonOwner("k1").computeIfPresent("k1", (v1, v2) -> null);
+//      asyncWait("k1", ComputeCommand.class);
+//      if (testRetVals) assertNull(retval);
+//      assertRemovedOnAllCaches("k1");
```
danberindei: @karesti we prefer not to comment out code, because it will never run and that doesn't help figure out what's wrong. The "proper" way to disable a test is to put a `@Test(groups = "unstable")` annotation on the test method. Does this fail just in the base test, or does it fail in some of the tests that extend `DistSyncFuncTest` as well?

We also don't create a new JIRA just to disable a test, instead we use the JIRA that's already open for fixing the test.

karesti: @danberindei This test fails randomly in this test but in any extending it, that's why I did it that way.
Concerning the JIRA ignoring the test and linking to the other JIRA that corrects the possible bug behind, @tristantarrant told me to do it like that (at least, this is what I understood)

karesti: @danberindei if I extract the test to another test and I add it to `@Test(groups = "unstable")`, will be added to the unstable group on the extending classes ?

tristantarrant:
My words were "yes, issue a PR to ignore it and mention the Jira which tracks its fix"

danberindei:
@karesti I think we can skip disabling the test, I've been able to reproduce it pretty reliably on my machine after I added a random delay in `InvalidateL1Command.perform()`:

https://github.com/infinispan/infinispan/compare/master...danberindei:ISPN-8074_DistSyncFuncTest.testComputeIfPresentFromNonOwner_reproducer

The problem seems to be that the L1 invalidation isn't synchronous, so the compute operation can finish before the L1 entry is removed from all the nodes.

core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
danberindei: @karesti I'm not sure. I never really grokked TestNG's behaviour when it comes to inheritance, so I have to test every time to make sure I get the behaviour I want.

PR #5329, opened 2017-07-31
ISPN-8004 Optimize serialization of encoders

gustavonalle:
https://issues.jboss.org/browse/ISPN-8004

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/dataconversion/EncodingUtils.java
+++ b/commons/src/main/java/org/infinispan/commons/dataconversion/EncodingUtils.java
@@ -4,6 +4,7 @@ package org.infinispan.commons.dataconversion;
  * Utilities to encode/decode keys and values from caches.
  *
  * @since 9.1
+ * @deprecated Use the org.infinispan.encoding.DataConversion obtained from the AdvancedCache.
  */
 public final class EncodingUtils {
 
--- a/core/src/main/java/org/infinispan/encoding/DataConversion.java
+++ b/core/src/main/java/org/infinispan/encoding/DataConversion.java
@@ -6,12 +6,15 @@ import java.io.ObjectOutput;
 import java.util.Set;
 
 import org.infinispan.commons.dataconversion.ByteArrayWrapper;
+import org.infinispan.commons.dataconversion.CompatModeEncoder;
 import org.infinispan.commons.dataconversion.Encoder;
 import org.infinispan.commons.dataconversion.IdentityEncoder;
 import org.infinispan.commons.dataconversion.Wrapper;
 import org.infinispan.commons.marshall.AbstractExternalizer;
 import org.infinispan.commons.marshall.Ids;
+import org.infinispan.commons.marshall.Marshaller;
 import org.infinispan.commons.util.Util;
+import org.infinispan.configuration.cache.CompatibilityModeConfiguration;
 import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.marshall.core.EncoderRegistry;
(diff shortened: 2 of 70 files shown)


server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
```diff
@@ -379,11 +372,11 @@ public void onCacheEvent(CacheEntryEvent<byte[], byte[]> event) {
             }
             Object k = event.getKey();
             Object v = event.getValue();
-            if (keyEncoder.isStorageFormatFilterable() && k != null) {
-               k = fromStorage(k, keyEncoder, keyWrapper);
+            if (keyDataConversion.isStorageFormatFilterable() && k != null) {
```
karesti: is it necessary to test that k is not null ? the fromStorage already checks that it seems to me

core/src/main/java/org/infinispan/encoding/DataConversion.java
```diff
@@ -0,0 +1,162 @@
...
+      } else {
+         CompatibilityModeConfiguration compatibility = configuration.compatibility();
+         Marshaller compatMarshaller = compatibility.marshaller();
+         this.encoder = new CompatModeEncoder(compatMarshaller);
+      }
+   }
+
+   public Object fromStorage(Object stored) {
+      checkConverters();
+      if (stored == null) return null;
```
karesti: I would check first if the value is null

karesti:
@gustavonalle I'm going to advance other stuff, so I guess I will rebase your work from master 👍 
Most of the time we use key/value together. I would prefer to have to pass 1 single class as CacheEncoders as I did in my branch, but I'm ok on this too

commons/src/main/java/org/infinispan/commons/dataconversion/EncodingUtils.java
```diff
@@ -4,6 +4,7 @@
  * Utilities to encode/decode keys and values from caches.
  *
  * @since 9.1
+ * @deprecated Use the org.infinispan.encoding.DataConversion obtained from the AdvancedCache.
  */
 public final class EncodingUtils {
```
gustavonalle: I deprecated because this class is public, it's not used because I already removed the (internal) usages

core/src/main/java/org/infinispan/encoding/DataConversion.java
gustavonalle: +1

gustavonalle:
> I would prefer to have to pass 1 single class as CacheEncoders

I think it'd be fine too, but as you said streams/filter sometimes require only the K or V to be converted and we can avoid shipping everything around, specially because the converters will include more (media type) soon.

server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
gustavonalle: this seems like a leftover from previous iterations, let me clean it up

tristantarrant:

gustavonalle:
Rebased and dressed reviews

tristantarrant:
I think we should add a section to the "contributing" docs listing known encoder ids and reference it from the encoder interface.

gustavonalle:
@tristantarrant Alternatively, I can enhance the javadoc of the encoder interface about those Ids, so documentations stays together with the code, wdyt?

gustavonalle:
Needs rebase, will come back to this shortly

gustavonalle:
rebased

galderz:
@anistor @karesti Anything to add here?

karesti:
@gustavonalle seems good to me too 💃

wburns:
Integrated into master, thanks @gustavonalle !

PR #5331, opened 2017-08-01
ISPN-6952 Fixed naming and added more debug information

slaskawi:
https://issues.jboss.org/browse/ISPN-6952

The test itself seems pretty stable on my environment. I'm afraid there's some interference with some other test and that causes failures. Added some more debugging info and will be retriggering it over and over again...

### DIFF
--- a/jcache/commons/src/test/java/org/infinispan/jcache/AbstractTwoCachesBasicOpsTest.java
+++ b/jcache/commons/src/test/java/org/infinispan/jcache/AbstractTwoCachesBasicOpsTest.java
@@ -9,9 +9,11 @@ import static org.testng.Assert.assertTrue;
 
 import java.io.Serializable;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


danberindei:
@slaskawi Have you considered that the additional listener invocations could be leftovers from previous test methods in the same class?

slaskawi:
@danberindei I actually considered this hypothesis and rejected it. All tests de-register their listeners so this might happen only if one of those test failed. Unfortunately all tests apart from this one are green. Having said that I hope to hunt down the root cause by printing out the listener content.

danberindei:
@slaskawi I think `testListenerDiscardingFilter` doesn't remove the 2nd listener. And none of the listeners when the test fails before the `deregisterCacheEntryListener(...)` line.

I also think the listener content won't help, because all the test methods use the same `keyX/valueX` pattern. Maybe you should change the tests to include the method name in the key/value, and I definitely suggest logging the invocation when it happens instead of waiting for the test to fail.

danberindei:
@slaskawi One more thing, I see `JCacheTwoCachesExpirationTest` also has the same name+package in embedded and remote. The other 2 remote tests are in a separate package, `org.infinispan.jcache.remote`, we could put all 4 of them there and skip putting `Remote` in the class name. WDYT?

danberindei:
@slaskawi got a failure in the embedded test as well:
http://ci.infinispan.org/job/Infinispan/job/PR-5353/2/testReport/junit/org.infinispan.jcache/JCacheTwoCachesAnnotationsTest/testResult_embedded_/

tristantarrant:
I'll look into this

PR #5332, opened 2017-08-01
ISPN-8142 Serialize global security configuration

tristantarrant:
https://issues.jboss.org/browse/ISPN-8142

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/serializing/XMLExtendedStreamWriter.java
+++ b/core/src/main/java/org/infinispan/configuration/serializing/XMLExtendedStreamWriter.java
@@ -4,19 +4,19 @@ import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 
 public interface XMLExtendedStreamWriter extends XMLStreamWriter {
-    void writeAttribute(String localName, String[] values) throws XMLStreamException;
+   void writeAttribute(String localName, String[] values) throws XMLStreamException;
 
-    void writeAttribute(String prefix, String namespaceURI, String localName, String[] values) throws XMLStreamException;
+   void writeAttribute(String prefix, String namespaceURI, String localName, String[] values) throws XMLStreamException;
 
-    void writeAttribute(String namespaceURI, String localName, String[] values) throws XMLStreamException;
+   void writeAttribute(String namespaceURI, String localName, String[] values) throws XMLStreamException;
 
-    void writeAttribute(String localName, Iterable<String> value) throws XMLStreamException;
+   void writeAttribute(String localName, Iterable<String> value) throws XMLStreamException;
 
-    void writeAttribute(String prefix, String namespaceURI, String localName, Iterable<String> value) throws XMLStreamException;
+   void writeAttribute(String prefix, String namespaceURI, String localName, Iterable<String> value) throws XMLStreamException;
 
-    void writeAttribute(String namespaceURI, String localName, Iterable<String> value) throws XMLStreamException;
+   void writeAttribute(String namespaceURI, String localName, Iterable<String> value) throws XMLStreamException;
 
-    void setUnspecifiedElementNamespace(String namespace);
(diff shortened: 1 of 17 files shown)


tristantarrant:

tristantarrant:
Brought back to life

core/src/main/java/org/infinispan/configuration/serializing/XMLExtendedStreamWriter.java
```diff
@@ -4,19 +4,19 @@
 import javax.xml.stream.XMLStreamWriter;
 
 public interface XMLExtendedStreamWriter extends XMLStreamWriter {
-    void writeAttribute(String localName, String[] values) throws XMLStreamException;
+   void writeAttribute(String localName, String[] values) throws XMLStreamException;
```
ryanemerson: Not sure we should include these changes if they are just cosmetic.

ryanemerson:
@tristantarrant I assume this will also want backporting to 9.1.x?

tristantarrant:
Possibly

core/src/main/java/org/infinispan/security/impl/CacheRoleImpl.java
```diff
@@ -54,4 +54,23 @@ public String toString() {
...
+   @Override
+   public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+
+      CacheRoleImpl cacheRole = (CacheRoleImpl) o;
+
+      if (mask != cacheRole.mask) return false;
+      if (name != null ? !name.equals(cacheRole.name) : cacheRole.name != null) return false;
+      return permissions != null ? permissions.equals(cacheRole.permissions) : cacheRole.permissions == null;
```
pruivo: `permissions` can't be `null`, otherwise the constructor would throw a NPE.

core/src/main/java/org/infinispan/security/impl/CacheRoleImpl.java
```diff
@@ -54,4 +54,23 @@ public String toString() {
...
+   }
+
+   @Override
+   public int hashCode() {
+      int result = name != null ? name.hashCode() : 0;
+      result = 31 * result + (permissions != null ? permissions.hashCode() : 0);
```
pruivo: `permissions` can't be `null`, otherwise the constructor would throw a NPE.

core/src/main/java/org/infinispan/security/impl/CacheRoleImpl.java
```diff
@@ -54,4 +54,23 @@ public String toString() {
...
 
+   @Override
+   public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+
+      CacheRoleImpl cacheRole = (CacheRoleImpl) o;
+
+      if (mask != cacheRole.mask) return false;
+      if (name != null ? !name.equals(cacheRole.name) : cacheRole.name != null) return false;
```
pruivo: `name` can be `null`?

wburns:
👍 about singleton or enum. I was actually just going to mention that but you beat me to the punch @pruivo

tristantarrant:
PrincipalRoleMappers cannot be immutable as they need to have the PrincipalRoleMapperContext injected in them

tristantarrant:
Addressed @pruivo 's minor suggestions to CacheRoleImpl

pruivo:
@tristantarrant yes they have the context injected, but they are ignoring it
```
@Override
   public void setContext(PrincipalRoleMapperContext context) {
      // Do nothing
   }
```

tristantarrant:
Unfortunately the server uses a different way to instantiate the mappers (using the class name), so turning them into singletons is not that simple.

tristantarrant:
@pruivo I pushed.

pruivo:
@tristantarrant  it looks good to me. 
@ryanemerson the cosmetic changes are OK to me. it can be integrated if CI is fine!

PR #5334, opened 2017-08-01
ISPN-8147 Add multi-threaded test for compareAndSet operation

mgencur:
https://issues.jboss.org/browse/ISPN-8147

### DIFF
--- a/counter/src/test/java/org/infinispan/counter/StrongCounterTest.java
+++ b/counter/src/test/java/org/infinispan/counter/StrongCounterTest.java
@@ -16,6 +16,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicIntegerArray;
 
 import org.infinispan.counter.api.CounterConfiguration;
 import org.infinispan.counter.api.CounterManager;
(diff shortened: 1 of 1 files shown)


counter/src/test/java/org/infinispan/counter/StrongCounterTest.java
```diff
@@ -96,6 +96,58 @@ public void testCompareAndSet(Method method) {
       }
    }
 
+   public void testCompareAndSetConcurrent(Method method) throws ExecutionException, InterruptedException, TimeoutException {
+      final int numThreadsPerNode = 2;
```
mgencur: Increasing this to 10 currently leads to https://issues.jboss.org/browse/ISPN-8134

pruivo: I though only the number of nodes in the cluster would originate ISPN-8134

mgencur: Oh. Sorry. It is a little different timeout. Not ISPN-8134

mgencur:
It was an issue with modulo. I've pushed the corrected version. Now it runs even with 8 threads per node.

counter/src/test/java/org/infinispan/counter/StrongCounterTest.java
```diff
@@ -96,6 +96,58 @@ public void testCompareAndSet(Method method) {
...
+               long update = previousValue + 1;
+               barrier.await();
+               //all threads calling compareAndSet at the same time, only one should succeed
+               boolean ret = counter.compareAndSet(previousValue, update);
+               if (ret) {
+                  previousValue = update;
+               } else {
+                  previousValue = counter.getValue();
+               }
+               retValues.add(ret);
```
danberindei: You could assert here that no other operation succeeded, then the main thread would only have to check that all workers finished without an exception.

danberindei:
Doesn't it make sense to run `testUniqueReturnValues` with more than one thread per node as well?

OTOH considering that there could be 14 other tests running in parallel I suggest only using 2 (or maybe 4) threads per node, and maybe adding a separate `testCompareAndSetConcurrentStress` method to run with 8 threads per node.

counter/src/test/java/org/infinispan/counter/StrongCounterTest.java
mgencur: How? By storing return values in a synchronized block to e.g. ArrayList and then checking that just one succeeded?

mgencur:
Pushed a new version. Thoughts?

counter/src/test/java/org/infinispan/counter/StrongCounterTest.java
danberindei: I was thinking that you don't need to store the result of every operation, something like this:

      final AtomicBoolean retValue = new AtomicBoolean(false);
      ...
               if (ret) {
                  assertTrue(retValue.compareAndSet(false, true));
               }
               barrier.await();
               retValue.set(false);

But this would only check that there's at most one successful operation, you still need something on the outside to check that there's at least one. So I was wrong, please go back to your original code.

danberindei:
Integrated, thanks Martin!

PR #5336, opened 2017-08-01
ISPN-8148 Use X-Forwarded-For header as source IP if available

tristantarrant:
https://issues.jboss.org/browse/ISPN-8148

### DIFF
--- a/server/rest/src/main/java/org/infinispan/rest/logging/RestAccessLoggingHandler.java
+++ b/server/rest/src/main/java/org/infinispan/rest/logging/RestAccessLoggingHandler.java
@@ -19,6 +19,7 @@ public class RestAccessLoggingHandler {
    private final static Log log = LogFactory.getLog(RestAccessLoggingHandler.class, Log.class);
 
    private final static String NANO_TIME = "NanoTime";
+   private final static String X_FORWARDED_FOR = "X-Forwarded-For";
 
    private boolean isEnabled() {
       return log.isTraceEnabled();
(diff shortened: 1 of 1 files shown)


tristantarrant:

tristantarrant:
Thanks @johnou, I'll see what I can do.

johnou:
@tristantarrant something like this..

```java
String forwardedFor = request.headers().getAsString(X_FORWARDED_FOR);
String remoteAddr = ctx.channel().remoteAddress().toString();
if (trustedProxyIpMatcher.matches(remoteAddr) && forwardedFor != null) {
    // if there are multiple addresses in the header, return the last in the list which is not trusted address
    if (forwardedFor.contains(",")) {
        String[] forwardedIPs = StringUtils.tokenizeToStringArray(forwardedFor, ",", /* trimTokens */ true, /* ignoreEmptyTokens */ true)
        for (int index = forwardedIPs.length - 1; index >= 0; index--) {
            forwardedFor = forwardedIPs[index];
            if (!trustedProxyIpMatcher.matches(forwardedFor)) {
                break;
            }
        }
    }
    return forwardedFor;
}
```

slaskawi:
I'm having doubts if `X-Forwarded-For` functionality works in current implementation. We don't interpret this header at all, [neither do Netty](https://github.com/netty/netty/search?utf8=%E2%9C%93&q=X-Forwarded-For&type=). So this leads to a question - does it magically work out-of-the-box and if not (that's my assumption), is there any reason for fixing audit logs?

tristantarrant:
The fact that netty doesn't implicitly include the header doesn't matter: it's just a named header.

johnou:
@slaskawi unfortunately that's not the case, it doesn't work out of the box, essentially the application needs to know if the remote peer is trusted eg. a load balancer.

tristantarrant:
If we want to prevent header spoofing, it is not something that should be done within a logger that can be disabled. We should instead add something like ip access control to the HTTP endpoint (like https://access.redhat.com/solutions/18412 )

tristantarrant:
The source IP whitelisting issue https://issues.jboss.org/browse/ISPN-8327

tristantarrant:
I'll reopen this when I have an implementation of ISPN-8327 to go with it.

PR #5342, opened 2017-08-02
[DO NOT MERGE] Remove the unstable test group

danberindei:
Remove the unstable group from all the unstable tests so they run in CI.

Do not integrate, this is just to see how many of the unstable tests are still failing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,10 +162,10 @@
       <bnd.consumer.policy>$&lt;range;[==,+)&gt;</bnd.consumer.policy>
       <bnd.provider.policy>$&lt;range;[==,=+)&gt;</bnd.provider.policy>
       <buildDirectory>target</buildDirectory>
-      <defaultExcludedJUnitGroups>org.infinispan.commons.test.categories.Unstable,org.infinispan.commons.test.categories.Stress,org.infinispan.commons.test.categories.Profiling</defaultExcludedJUnitGroups>
+      <defaultExcludedJUnitGroups>org.infinispan.commons.test.categories.Stress,org.infinispan.commons.test.categories.Profiling</defaultExcludedJUnitGroups>
       <!-- unstable and stress because we can have them inside a functional class, for example -->
-      <defaultExcludedTestNGGroups>unstable,stress</defaultExcludedTestNGGroups>
-      <defaultTestNGGroups>functional,unit,xsite,arquillian</defaultTestNGGroups>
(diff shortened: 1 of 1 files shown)


slaskawi:
I love this idea!!! Let's see what CI says about it... we will know what point are we really at?

tristantarrant:
I'd prefer if we had a dedicated CI job for unstable and we moved all currently disabled or ignored tests to the unstable group. We would get daily reports and we would have more incentive to promote tests out of unstable (well, "incentive" is a loose word :)

danberindei:
I would prefer a separate CI job too, I only opened this PR because I think creating a separate job would take much longer.

galderz:
Needs rebase

slaskawi:
@tristantarrant @galderz @danberindei I think there's no need to create a custom job for ustable tests. The biggest question is what shall we do if a test is green in unstable group but fails when running with the rest of tests. IMO such a job does not provide any information to us and increases complexity (different profiles) and add maintenance burden. 

What I would prefer to do instead, is to simply Ignore this test (in other words put it into quarantine) and assign someone to look into it. What we could also do is to set a threshold for ignored tests. Once we exceed the threshold, we will fail the main build.

rvansa:
@danberindei Should this PR still hang around?

danberindei:
@slaskawi if there's no CI build running unstable tests, how do you know that a test is "green in unstable group"? And if you think that asking someone to look into an unstable test that hasn't ran in CI for more than a month, please pick one test from the spreadsheet and let me know how it goes :)

danberindei:
@rvansa Tristan has configured Jenkins to run the build on the original branch, not on the merge, so it's not that useful any more, but I think the history might still help the brave souls who try to work on tests that have been marked unstable for a long time.

tristantarrant:
Not only that, but new tests have been moved to unstable as well as harmonizing testng/junit unstable categories. We should probably have a nightly job which enables the unstable profile.

ryanemerson:
@danberindei Is this PR going to be resurrected any time soon? Or can we close it for now?

danberindei:
@ryanemerson it was never intended to be merged, just as a way to run the unstable tests in CI from time to time. I haven't made any progress on creating an unstable tests job in Jenkins.

slaskawi:
@danberindei So if the intention of this PR is to run unstable tests, I guess we need a separate Jenkins job. I would prefer to close this PR (please reopen if you don't agree with me).

PR #5343, opened 2017-08-02
ISPN-8074 testComputeIfPresentFromNonOwner correction

karesti:
https://issues.jboss.org/browse/ISPN-8074

### DIFF
--- a/core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
+++ b/core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
@@ -9,6 +9,7 @@ import static org.testng.AssertJUnit.assertTrue;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


karesti:
@danberindei just a preview of the direction of the correction I've taken.

core/src/test/java/org/infinispan/distribution/DistSyncFuncTest.java
```diff
@@ -39,6 +42,32 @@ public DistSyncFuncTest() {
       testRetVals = true;
    }
 
+   protected Map<Cache<?, ?>, ReplListener> listenerLookup;
```
pruivo: -1
I would prefer to have an empty `waitForInvalidateL1InNonOwners()` method here and override it in the async test instead of moving "unused" listeners to the sync test.

danberindei:
@karesti I think we can close this PR, since the problem is with the production code and not with the test.

karesti:
@danberindei agreed

PR #5344, opened 2017-08-02
ISPN-4038 CustomFailurePolicyTxTest.testPrepareFailure random failures

pruivo:
https://issues.jboss.org/browse/ISPN-4038

### DIFF
--- a/core/src/test/java/org/infinispan/xsite/backupfailure/tx/BaseBackupTxFailureTest.java
+++ b/core/src/test/java/org/infinispan/xsite/backupfailure/tx/BaseBackupTxFailureTest.java
@@ -3,8 +3,9 @@ package org.infinispan.xsite.backupfailure.tx;
 import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertNull;
 
-import javax.transaction.xa.XAException;
+import javax.transaction.RollbackException;
 
+import org.infinispan.commons.CacheException;
 import org.infinispan.configuration.cache.BackupFailurePolicy;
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/infinispan/xsite/backupfailure/tx/BaseBackupTxFailureTest.java
```diff
@@ -33,11 +32,10 @@ protected void createSites() {
       backup("LON").getAdvancedCache().getAsyncInterceptorChain().addInterceptor(failureInterceptor, 1);
    }
 
-   @Test(groups = {"xsite", "unstable"})
    public void testPrepareFailure() {
       failureInterceptor.enable();
       try {
-         Exceptions.expectException(XAException.class, () -> cache("LON", 0).put("k", "v"));
+         Exceptions.expectException(CacheException.class, () -> cache("LON", 0).put("k", "v"));
```
tristantarrant: You could use Exceptions.expectException(CacheException.class, RollbackException.class, () -> cache("LON", 0).put("k", "v"));
I noticed the XAException is now a suppressed exception inside the RollbackException, but exploring the whole exception hierarchy might be overkill

pruivo: The `XaException` are never seen by the user. they are only there for interaction between the `TransactionManager` and the `XaResource`/`Synchronization`. 
I'll change to check for `RollbackException`

tristantarrant:
Good stuff. merged.

tristantarrant:
Merged

PR #5345, opened 2017-08-02
ISPN-8167 Archive debug logs in CI

danberindei:
https://issues.jboss.org/browse/ISPN-8167

Personally I'm hoping to get more info about the `SiteStatementTest` random failure...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -47,6 +47,9 @@ pipeline {
                 junit testResults: '**/target/*-reports*/*.xml',
                         testDataPublishers: [[$class: 'ClaimTestDataPublisher']],
                         healthScaleFactor: 100
+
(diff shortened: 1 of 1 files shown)


pruivo:
@danberindei it looks good to me... but aren't we having problems with disk space? when are they deleted?

danberindei:
@pruivo The zipped logs are ~ 40MB per build, which means < 10GB for 200 builds, so we should be fine. We ran out of disk space on the Jenkins server mostly because of a few 8.2.x builds that had > 2GB of logs (ISPN-8170).

pruivo:
ok. going to merge

pruivo:

tristantarrant:
Tried using tar cJf. The savings are great:

41852 logs-PR-5365-2.zip  17152 logs.tar.xz

PR #5346, opened 2017-08-02
ISPN-8169 DirectoryImplementerTests.testConfigureAsyncDeletes random …

gustavonalle:
…failures

https://issues.jboss.org/browse/ISPN-8169

### DIFF
--- a/lucene/lucene-directory/src/test/java/org/infinispan/lucene/impl/DirectoryImplementerTests.java
+++ b/lucene/lucene-directory/src/test/java/org/infinispan/lucene/impl/DirectoryImplementerTests.java
@@ -1,33 +1,32 @@
 package org.infinispan.lucene.impl;
 
-import java.io.IOException;
-import java.util.Set;
-import java.util.concurrent.SynchronousQueue;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.testng.Assert.assertEquals;
 
-import org.apache.lucene.analysis.standard.StandardAnalyzer;
-import org.apache.lucene.document.Document;
-import org.apache.lucene.document.Field;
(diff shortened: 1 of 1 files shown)


gustavonalle:
Looks like CI is not picking this one

anistor:
I noticed that. I'm running this locally in a loop to see how airtight it is.

gustavonalle:
Even in master I can't replicated the failure in my local environment, even constraining severely the CPU to make it as slow as possible. There must be some serious hiccups happening on CI every now and then causing wild thread scheduling discrepancies

anistor:
I think the initial version of the test may be wrong because it relied on the invocation of the ```afterExecute``` method of the threadpool, but the threadpool is shut down right after the second commit to the indexwriter. ```shutdown``` method guarantees that the previously submitted tasks are still executed but it does not say if the other 'contracts' of the threadpool will be honoured, like the execution of the ```afterExecute``` method, so we may have been relying on the wrong hook and the ```deletedSegments``` set may have not been updated correctly after the execution of the delete task. 

Anyway, I like the new version of this test. It's so much clearer than the previous.

I'm still running it trying to break it, but it does not seem to fail, so I'll merge it soon.

lucene/lucene-directory/src/test/java/org/infinispan/lucene/impl/DirectoryImplementerTests.java
```diff
@@ -102,67 +101,43 @@ public void testGetIndexNameAndToString() throws IOException {
...
-      }
-
-      private String extractSegmentName(String fileName) {
-         if (!fileName.startsWith("_")) {
-            return null;
-         } else {
-            return fileName.substring(1, fileName.indexOf('.'));
-         }
-      }
+         verify(executor).execute(any(Runnable.class));
```
anistor: Here I'd be very picky, and say ```verify(executor).execute(any(DirectoryLucene.DeleteTask.class))```

gustavonalle:
Updated @anistor

anistor:

anistor:
Backported also to 8.2.x and 9.0.x.

PR #5350, opened 2017-08-03
ISPN-8114 Random failures in loading from Hibernate Cache

galderz:
Fixed NaturalIdInvalidationTest so that it waits for end invalidation messages.

### DIFF
--- a/hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/PutFromLoadValidator.java
+++ b/hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/PutFromLoadValidator.java
@@ -586,6 +586,9 @@ public class PutFromLoadValidator {
 						continue;
 					}
 					long now = regionFactory.nextTimestamp();
+               if (trace) {
+                  log.tracef("endInvalidatingKey(%s#%s, %s) remove invalidator from %s", cache.getName(), key, lockOwnerToString(lockOwner), pending);
+               }
 					pending.invalidate(now);
 					pending.addInvalidator(lockOwner, valueForPFER, now);
 				}
--- a/hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/AbstractFunctionalTest.java
+++ b/hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/AbstractFunctionalTest.java
@@ -6,6 +6,8 @@
  */
 package org.infinispan.test.hibernate.cache.functional;
 
+import static org.infinispan.test.TestingUtil.wrapInboundInvocationHandler;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -15,7 +17,10 @@ import java.util.function.Predicate;
 import org.hibernate.Session;
 import org.hibernate.boot.Metadata;
 import org.hibernate.boot.spi.MetadataImplementor;
+import org.infinispan.commands.remote.CacheRpcCommand;
+import org.infinispan.hibernate.cache.util.EndInvalidationCommand;
 import org.infinispan.hibernate.cache.util.FutureUpdate;
+import org.infinispan.hibernate.cache.util.InfinispanMessageLogger;
 import org.infinispan.hibernate.cache.util.TombstoneUpdate;
 import org.hibernate.cache.internal.SimpleCacheKeysFactory;
 import org.hibernate.cache.spi.RegionFactory;
@@ -33,6 +38,9 @@ import org.hibernate.resource.transaction.backend.jdbc.internal.JdbcResourceLoca
 import org.hibernate.resource.transaction.backend.jta.internal.JtaTransactionCoordinatorBuilderImpl;
 import org.hibernate.resource.transaction.spi.TransactionCoordinatorBuilder;
 
+import org.infinispan.remoting.inboundhandler.DeliverOrder;
+import org.infinispan.remoting.inboundhandler.PerCacheInboundInvocationHandler;
+import org.infinispan.remoting.inboundhandler.Reply;
 import org.infinispan.test.hibernate.cache.util.ExpectingInterceptor;
 import org.hibernate.testing.BeforeClassOnce;
 import org.hibernate.testing.junit4.BaseNonConfigCoreFunctionalTestCase;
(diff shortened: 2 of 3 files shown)


galderz:
By the way, the reason I end up wrapping the command and adding the latches there is because I needed for the latch count down to happen within the `remote-thread-` thread invocation. The invocation handler gets executed in the `jgroups-` thread.

galderz:
CI run a couple of times the testsuite of Hibernate Cache ([build 1](http://ci.infinispan.org/job/Infinispan/job/PR-5350/1/) and [build 3](http://ci.infinispan.org/job/Infinispan/job/PR-5350/3/)) and the testsuite run fine.

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/AbstractFunctionalTest.java
```diff
@@ -239,4 +253,114 @@ protected CountDownLatch expectPutWithValue(AdvancedCache cache, Predicate<Objec
...
+      private final CountDownLatch latch;
+
+      public ExpectingInboundInvocationHandler(PerCacheInboundInvocationHandler delegate, CountDownLatch latch) {
+         this.delegate = delegate;
+         this.latch = latch;
+      }
+
+      @Override
+      public void handle(CacheRpcCommand command, Reply reply, DeliverOrder order) {
+         if (command instanceof EndInvalidationCommand) {
```
rvansa: You don't need to create a new command at all. Just call
```java
delegate.handle(command, response -> {
    latch.countDown();
    reply.reply(response);
}, order);
```

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/AbstractFunctionalTest.java
```diff
@@ -239,4 +253,114 @@ protected CountDownLatch expectPutWithValue(AdvancedCache cache, Predicate<Objec
 			return new CountDownLatch(0);
 		}
 	}
+
+   protected CountDownLatch expectAfterEndInvalidation(AdvancedCache cache, int numInvalidates) {
+      CountDownLatch latch = new CountDownLatch(numInvalidates);
+      wrapInboundInvocationHandler(cache, handler -> new ExpectingInboundInvocationHandler(handler, latch));
```
rvansa: You don't ever remove that handler. I suggest adding a `List<Runnable> cleanup` collection as a field and executing tasks from that from `@AfterMethod`. `expectAfterEndInvalidation` should add removal of the wrapper to that collection.

hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/PutFromLoadValidator.java
```diff
@@ -586,6 +586,9 @@ public boolean beginInvalidatingWithPFER(Object lockOwner, Object key, Object va
 						continue;
 					}
 					long now = regionFactory.nextTimestamp();
+               if (trace) {
```
galderz: Hmm, why? I've not changed anything in the rest of the method.

If you look at the files changed adding `?ts=3` at the end (e.g. [this URL](https://github.com/infinispan/infinispan/pull/5350/files?ts=3)), you won't have issues seeing where the changes happen.

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/cluster/NaturalIdInvalidationTest.java
```diff
@@ -41,6 +42,9 @@
 
 	private static final InfinispanMessageLogger log = InfinispanMessageLogger.Provider.getLog(NaturalIdInvalidationTest.class);
 
+	@Rule
```
galderz: Indentation is fine. 3 spaces.

The rest of indentation is tabs, so you as in the previous comment, adjust the tab to be 3 spaces passing in `?ts=3`

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/AbstractFunctionalTest.java
galderz: I'll look into that.

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/AbstractFunctionalTest.java
galderz: I'll double check that.

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/AbstractFunctionalTest.java
galderz: > I suggest adding a List<Runnable> cleanup collection as a field and executing tasks from that from @AfterMethod.

I disagree with that because it's not symmetric.

I do agree with the need to remove the handler, but it should be symmetric. If the handler is added in the test method, it should be removed in the test method itself.

galderz: Pushed a commit to fix this.

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/AbstractFunctionalTest.java
galderz: Pushed another commit to fix this.

hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/PutFromLoadValidator.java
rvansa: Even with `ts=3` the `if` is off one char to the right...

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/AbstractFunctionalTest.java
rvansa: OK, I personally prefer the test to focus on test logic, instead of having a big `try { ... } finally { /* cleanup */ }` switch (which we have there anyway, though). Cleanup is what `@AfterMethod`s are for. But be it your way...

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/cluster/NaturalIdInvalidationTest.java
```diff
@@ -55,9 +59,8 @@
 	}
 
 	@Test
-   @Ignore("Randomly failing on CI - ISPN-8114")
 	public void testAll() throws Exception {
-		log.info( "*** testAll()" );
+      log.infof("*** %s", name.getMethodName());
```
rvansa: Are these logs needed? TestNG should print out `Test starting: ...`

rvansa:
@galderz Do you intend any further cosmetic changes or should I merge?

hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/PutFromLoadValidator.java
galderz: It's not. The extra space is due to the diff `+` character.

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/AbstractFunctionalTest.java
galderz: From a reader perspective, creating something within a test method and then cleaning it up in an `@AfterMethod` is asymmetric, so the user has to look somewhere else to find out where the cleanup is done... This might be necessary if there's any bugs in the clearing itself. Also, related to this, I don't know which kind of guarantees you get in terms of which thread will execute `@AfterMethod`. In theory, it should be the same thread as test method one but I've never trusted these testing frameworks. If the thread is for whatever is different, then it can get messy to figure out issues.

I agree with you that the try/finally clause is a bit clunky, but I feel it makes it clear that whatever I create in this section will be cleared there, so helps the reader IMO, and you have full guarantees of who executes the finally block.

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/functional/cluster/NaturalIdInvalidationTest.java
galderz: TestNG only prints `Test starting: testAll()`, leaving out all the test parameters, so trying to figure out which testAll is being executed becomes quite a mission. I've personally had to go through this pain analysing TRACE logs.

In other tests, the thread name has been adjusted to have the parameters embedded within the test name. However, in this particular test this does not seem to be easy to do (all threads are named `Thread-xxx`... something to do with how the test is designed... Changing this seemed messy, so instead went for the simpler fix, which is to print out not only the test method name but also the test parameters. This is enough to figure which `Thread-` belongs to what.

If you know how to fix it, be my guest, creating a separate JIRA for it :)

galderz:
> @galderz Do you intend any further cosmetic changes or should I merge?

Let me squash the commits first and then it should be ready.

galderz:
@rvansa I've pushed a squashed commit.

PR #5351, opened 2017-08-03
ISPN-4042 Fix test and re-enable it

tristantarrant:
https://issues.jboss.org/browse/ISPN-4042


tristantarrant:
@danberindei I decided to re-enable this even though the description mentioned that it might not make sense. I couldn't find another eviction + storeasbinary test, so this one covers it.

danberindei:
@tristantarrant Is there anything different about eviction when store as binary is enabled? Back when I added the comment, the test had a `MockMarshalledValueInterceptor` to test that eviction never called `MarshalledValueInterceptor.createMarshalledValue()`, but Will removed that with ISPN-7177.

BTW, `MarshalledValuesManualEvictionTest` is in the same situation.

tristantarrant:
Closing this

PR #5354, opened 2017-08-04
ISPN-6456 OptimisticTxPartitionAndMergeDuringPrepareTest.testOriginat…

ryanemerson:
…orIsolatedPartition fails randomly

https://issues.jboss.org/browse/ISPN-6456

I believe this should also fix the `PessimisticTxPartiionAndMergeDuringPrepareTest# testPrimaryOwnerIsolatedPartition` failures.

From the logs, I can see that in the failure cases node C (the pirmary owner of key 2) never updates its TransactionTable minTopology to the topology established after a merge, therefore it never removes the old transaction from its records.

### DIFF
--- a/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
+++ b/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
@@ -390,7 +390,12 @@ public class TransactionTable implements org.infinispan.transaction.TransactionT
             if (topologyId < minTxTopologyId) {
                if (trace)
                   log.tracef("Changing minimum topology ID from %d to %d", minTxTopologyId, topologyId);
-               minTxTopologyId = topologyId;
+               minTopologyRecalculationLock.lock();
+               try {
+                  minTxTopologyId = topologyId;
+               } finally {
+                  minTopologyRecalculationLock.unlock();
+               }
             }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
```diff
@@ -390,7 +390,12 @@ private RemoteTransaction getOrCreateRemoteTransaction(GlobalTransaction globalT
             if (topologyId < minTxTopologyId) {
                if (trace)
                   log.tracef("Changing minimum topology ID from %d to %d", minTxTopologyId, topologyId);
-               minTxTopologyId = topologyId;
+               minTopologyRecalculationLock.lock();
```
pruivo: I'm not sure if it makes much sense to lock here. at lest, include the `if (topologyId < minTxTopologyId)` in the locked section.

ryanemerson: Yes you're correct. I encountered a test failure shortly after issuing the PR anyway :confounded:

danberindei:
@ryanemerson it's the other way around: `TransactionTable.calculateMinTopologyId()` can only update the minimum topology id after all the transactions with the old id have been removed.

PR #5355, opened 2017-08-04
Fix OptimisticTxPartitionAndMergeDuringCommitTest

pruivo:
wrong assumption during when writing the test make it fail randomly.
the transaction is committed in the majority partition and checking and the locks are released. checking for locks during the split can fail depending if the locks are still acquired or not (locks are released asynchronously)


tristantarrant:
Merged to master

PR #5356, opened 2017-08-04
Jenkinsfile: Remove the deploy stage and run git clean

danberindei:


pruivo:
so no more snapshots are deployed? should we add a daily snapshot deployment?

tristantarrant:
Once I have the "fatci" EC2 machine properly configured to only run when necessary, I want to deploy snapshots nightly from that.

danberindei:
@pruivo If nobody comes up and says they need the CI snapshots, I'd rather not publish them.

@tristantarrant If it's just a nightly snapshot, we can do it from the existing `ci-slave-fast-1`. Unless you mean it should happen once per day *and* after a successful build...

tristantarrant:
"should happen once per day and after a successful build." < that was my intention

pruivo:
ack.  
@danberindei, I guess it can be integrated or do you want to test/change something else?

tristantarrant:
Merged to master

PR #5357, opened 2017-08-04
ISPN-8168 Index corruption with topology changes

gustavonalle:
https://issues.jboss.org/browse/ISPN-8168

Removed ```DeltaAware``` from the lucene directory. Did not measure any performance difference when running clusters locally. 

Once the functional commands issues related to previous values and topology change are resolved (likely on ```9.2```), we can re-introduce the replication of the ```FileListCacheValue``` using func. commands


tristantarrant:
Merged to master, thanks !

PR #5358, opened 2017-08-07
ISPN-8027 Wait for topologies to be set

galderz:
https://issues.jboss.org/browse/ISPN-8027

* Region tests relying on asynchronous commands can't be retried when
  outdated topologies are encountered. This is fine from a functional
  perspective, but it can cause impredictable test results.

### DIFF
--- a/hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/AbstractGeneralDataRegionTest.java
+++ b/hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/AbstractGeneralDataRegionTest.java
@@ -13,6 +13,7 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import org.hibernate.SessionFactory;
 import org.hibernate.Transaction;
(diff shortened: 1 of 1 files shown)


rvansa:
Do I understand correctly that with async caches, the OTE it thrown but the command won't be retried? As this relates to timestamps & query caches, that means that the command won't be executed and queries will be stale (not fixing themselves eventually)?

If this is true, we need to deal with it in core, not just avoid that situation in the tests. At least logging a proper JIRA.

galderz:
> Do I understand correctly that with async caches, the OTE it thrown but the command won't be retried?

Yes, that's correct.

> As this relates to timestamps & query caches, that means that the command won't be executed and queries will be stale (not fixing themselves eventually)? If this is true, we need to deal with it in core, not just avoid that situation in the tests. At least logging a proper JIRA.

The risk of stale queries has always been there since we choose asynchronous communications for timestamps cache (I assume you remember the default cluster configuration docu [here](https://docs.jboss.org/hibernate/orm/5.2/userguide/html_single/Hibernate_User_Guide.html#caching-provider-infinispan). It's true that in this case, the issue does not get repaired, but all sorts of errors/exceptions could lead to such situations... Not sure we can handle/predict all of them TBH.

However, I do agree that core could do a better job of handling this kind of issues by retrying asynchronous commands if an outdated topology is found, even if they're asynchronous. I'll create a JIRA for this.

If we want to support such scenarios, we need specific tests that reliable/predictable test such scenarios. We shouldn't have tests that might randomly encounter such situations, hence I think the specific fix in this PR is still valid.

galderz:
@rvansa I've created [https://issues.jboss.org/browse/ISPN-8182](ISPN-8282) to track this requirement.

tristantarrant:
Merged to master.

PR #5363, opened 2017-08-08
ISPN-8175 LocalCacheStateTransferTest random failures

pruivo:
https://issues.jboss.org/browse/ISPN-8175

I was unable to reproduce it locally but I *think* that it fails because of `finish_receive` is sent asynchronously in `NYC` site. 
Change the `assert` to `eventually`


gustavonalle:
This is pretty tough to reproduce (if not impossible) when running the test in isolation. I was able to reproduce it when running the full ```core``` suite, and not very often.

gustavonalle:
Integrated, thanks!

PR #5365, opened 2017-08-08
ISPN-8183 ClassFinder logs lots of warnings in CI

danberindei:
https://issues.jboss.org/browse/ISPN-8183

Change the rules for whether a classpath element belongs to Infinispan:
for files, the file name should contain "infinispan",
and for directories, they should have a subdirectory "org/infinispan".


tristantarrant:
Only 112 warnings now. Looks good.

PR #5368, opened 2017-08-08
ISPN-8150 SingleFileStoreStressTest intermittent failure

ryanemerson:
https://issues.jboss.org/browse/ISPN-8150

### DIFF
--- a/core/src/test/java/org/infinispan/persistence/file/SingleFileStoreStressTest.java
+++ b/core/src/test/java/org/infinispan/persistence/file/SingleFileStoreStressTest.java
@@ -19,13 +19,14 @@ import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.infinispan.Cache;
+import org.infinispan.commons.io.ByteBuffer;
+import org.infinispan.commons.io.ByteBufferImpl;
 import org.infinispan.commons.marshall.StreamingMarshaller;
 import org.infinispan.commons.util.Util;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.marshall.core.MarshalledEntry;
 import org.infinispan.marshall.core.MarshalledEntryImpl;
-import org.infinispan.persistence.manager.PersistenceManager;
 import org.infinispan.test.SingleCacheManagerTest;
 import org.infinispan.test.TestingUtil;
 import org.infinispan.test.fwk.TestCacheManagerFactory;
@@ -40,7 +41,8 @@ import org.testng.annotations.Test;
  */
 @Test(groups = "unit", testName = "persistence.file.SingleFileStoreStressTest")
 public class SingleFileStoreStressTest extends SingleCacheManagerTest {
-   public static final String CACHE_NAME = "testCache";
+   private static final String CACHE_NAME = "testCache";
+   private static final String TIMES_STRING = "123456789_";
    private String location;
 
    @AfterClass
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/infinispan/persistence/file/SingleFileStoreStressTest.java
```diff
@@ -191,65 +188,53 @@ public void testFileTruncation() throws ExecutionException, InterruptedException
...
-      MILLISECONDS.sleep(200);
+      store.purge(null, null);
       long length3 = file.length();
 
-      executor.shutdown();
-
       // Verify that file size decreases and the file shrinks
       assertTrue(length2 <= length1);
-      assertTrue(length3 < length2);
+      assertTrue(String.format("Length2=%d, Length3=%d", length2, length3), length3 <= length2);
```
rvansa: So the store may not shrink?

core/src/test/java/org/infinispan/persistence/file/SingleFileStoreStressTest.java
```diff
@@ -323,7 +306,7 @@ private void populateStoreRandomValues(int NUM_KEYS, int MAX_VALUE_SIZE, SingleF
                                          StreamingMarshaller marshaller, List<String> keys) {
       for (int j = 0; j < NUM_KEYS; j++) {
          String key = "key" + j;
-         String value = key + "_value_" + j + times("123456789_", new Random().nextInt(MAX_VALUE_SIZE / 10));
+         String value = key + "_value_" + j + times(new Random().nextInt(MAX_VALUE_SIZE / 10));
```
rvansa: Please reformat the parameters to camelCase.

core/src/test/java/org/infinispan/persistence/file/SingleFileStoreStressTest.java
danberindei: Yeah, I think the store is definitely not shrinking all the time. This has been failing (randomly) for a long time, see https://issues.jboss.org/browse/ISPN-5730.

danberindei: Looking at the test some more, it's sleeping for 200ms, but the `SingleFileStore.purge()` actually does everything on the caller thread. So the test could pass a `WithinThreadExecutor` and skip the wait.

I also suggest testing `store.getFileSize()` before `file.size()`, just in case there's some caching going on in `File.size()`. I'd go even further and use `TestingUtil.extractField` to print the used+free entries if the `store.getFileSize()` check fails.

rvansa: I thought this test was about making sure that it does shrink, changing the condition in here voids the test.

ryanemerson: The reason it is not shrinking every time because the merging of entries in the first call to purge results in large FileEntries being created before the end of file. So when we attempt to write more entries, the existing FileEntries before the end of file are sufficient to accommodate the new entries and so when we call purge again, there are no files at the end of the file to be truncated. I agree with @rvansa that my original changes void the test. I am going to push some changes that will ensure that there is always at least one free FileEntry at the end of the file and hence a truncation will occur.

core/src/test/java/org/infinispan/persistence/file/SingleFileStoreStressTest.java
```diff
@@ -202,54 +200,47 @@ public void testFileTruncation() throws ExecutionException, InterruptedException
       // Do some reading/writing entries with random size
       final CountDownLatch stopLatch = new CountDownLatch(1);
       Future[] writeFutures = new Future[NUM_WRITER_THREADS];
-      for (int i = 0; i < NUM_WRITER_THREADS; i++) {
+      for (int i = 0; i < NUM_WRITER_THREADS; i++)
```
danberindei: -1

danberindei:
Integrated, thanks Ryan!

PR #5371, opened 2017-08-09
ISPN-5170 Re-enable MemcachedClusteredStatsTest

tristantarrant:
https://issues.jboss.org/browse/ISPN-5170

I've fixed the mbean name and re-enabled the test. I've run it in a loop locally for a couple of hourse and I haven't had it fail once...


ryanemerson:
@tristantarrant Merged, thanks!

PR #5377, opened 2017-08-11
Enable log appending for the server testsuite

tristantarrant:


tristantarrant:
Changed to filter out some excessive chattiness (e.g. from ApacheDS)

PR #5384, opened 2017-08-21
ISPN-8206 Update smoke test suite

mgencur:
https://issues.jboss.org/browse/ISPN-8206

Smoke tests are run for majority of Infinispan modules. When there are no smoke tests in the module it means that the functionality is covered by other modules (e.g. query-dsl is covered by query tests) or they are not part of the product. The smoke test suite will be run for the product on some less frequently used platforms. I'm going to create another PR for the product when this PR is integrated.

### DIFF
--- a/cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/cache/DefaultCacheTest.java
+++ b/cdi/embedded/src/test/java/org/infinispan/cdi/embedded/test/cache/DefaultCacheTest.java
@@ -20,7 +20,7 @@ import org.testng.annotations.Test;
  * @author Pete Muir
  * @author Kevin Pollet <kevin.pollet@serli.com> (C) 2011 SERLI
  */
-@Test(groups = "functional", testName = "cdi.test.cache.embedded.DefaultCacheTest")
+@Test(groups = {"functional", "smoke"}, testName = "cdi.test.cache.embedded.DefaultCacheTest")
 public class DefaultCacheTest extends Arquillian {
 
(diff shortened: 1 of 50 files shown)


mgencur:
I also sent an email to infinispan-dev list about smoke test suite for hibernate-cache module. There are no smoke tests for the module in this PR. I may add them later if needed.

slaskawi:
It's probably very late in the cycle, but scratching my head about a
fundamental question - Why do we need smoke tests? When are we going to use
them? What shall we do if smoke testsuite is green but running all tests
indicate some failures?

> --

SEBASTIAN ŁASKAWIEC

INFINISPAN DEVELOPER

Red Hat EMEA <https://www.redhat.com/>
<https://red.ht/sig>

mgencur:
As I said, the smoke test suite will be run for the product on some less frequently used platforms. So on some platforms we'll be running all tests and on some platforms just smoke tests. If the smoke test suite is green on one platform but the full test suite has failures on other (more important and frequently used platforms) then these failures would probably be on the less frequent platforms as well. 

The goal is simply to reduce the time for running and investigating test suites on obscure platforms when we have so many random failures. I know there's an effort to reduce the number of random failures to zero but that is only for a single platform, AFAIK.

mgencur:
I've added a clean up of TMP directory to BeforeClass phase of SingleFileStoreFunctionalTest to prevent reading data stored in previous releases.

tristantarrant:
Merged, thanks

PR #5386, opened 2017-08-21
ISPN-8114 Ignore failing test again

galderz:
Test is still failing in master, it needs further looking into.


gustavonalle:
Integrated, thanks @galderz

PR #5387, opened 2017-08-22
Fix broken Links in performance guide

kevcodez:


rvansa:
Actually I think that the problem is using `9.1` in https://github.com/infinispan/infinispan/blob/master/documentation/pom.xml#L19 while the docs are on URL using `9.1.x`; @tristantarrant ?

Anyway, thanks for pointing out that those links are broken.

kevcodez:
@rvansa It is the reason why. Other links are also done using simple back navigation, thus my proposed fix. I do not know if the difference in version is on purpose. This way, the version does not matter.

Sure, no problem :)

tristantarrant:
Merged, thanks

PR #5388, opened 2017-08-23
ISPN-8207 Update eviction tests and add off-heap eviction listener

vjuranek:
https://issues.jboss.org/browse/ISPN-8207
https://issues.jboss.org/browse/ISPN-8216


core/src/test/java/org/infinispan/eviction/impl/OffHeapEvictionFunctionalTest.java
```diff
@@ -0,0 +1,20 @@
+package org.infinispan.eviction.impl;
+
+import org.infinispan.configuration.cache.StorageType;
+import org.testng.annotations.Test;
+
+import jdk.nashorn.internal.ir.annotations.Ignore;
```
tristantarrant: Checkstyle is failing because of unused imports

vjuranek:
ops, sorry, fixed

core/src/test/java/org/infinispan/eviction/impl/ObjectEvictionFunctionalTest.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @author vjuranek
+ * @since 9.2
+ */
+@Test(groups = {"functional", "smoke"}, testName = "eviction.ObjectEvictionFunctionalTest")
+public class ObjectEvictionFunctionalTest extends BaseEvictionFunctionalTest {
+
+   @Override
+   protected StorageType getStorageType() {
+      return StorageType.OBJECT;
```
wburns: We should be able to make this into a `@Factory` method similar to https://github.com/infinispan/infinispan/blob/master/core/src/test/java/org/infinispan/container/offheap/OffHeapSingleNodeExpirationEvictionTest.java#L45.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -223,7 +224,9 @@ private void ensureSize() {
                      addressToRemove, currentSize, maxSize);
             }
             try {
+               InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(addressToRemove);
```
wburns: I wonder if we could change this so we lazily ressurect this object. This way if no listener is installed we wouldn't have to read this object. But that would probably require changing `evictionManager` API. So maybe for ISPN 10? @tristantarrant WDYT?

vjuranek:
@wburns thanks, switched to factory method.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
tristantarrant: Sure.

tristantarrant:
Merged to master, thanks

PR #5389, opened 2017-08-23
Fix random failures in DistAsyncFuncTest

rvansa:
The L1 registration can survive from one test to another, therefore
expecting InvalidateL1Commands is rather unreliable. Instead, we'll
monitor to which nodes the invalidation was actually sent and block
the test until we see those updates applied after each write command.


core/src/test/java/org/infinispan/distribution/DisabledL1Test.java
```diff
@@ -22,11 +22,11 @@ public void testRemoveFromNonOwner() {
       for (Cache<Object, String> c : caches) assert c.isEmpty();
```
karesti: I know this is not the purpose of your PR, but this test seems to me unnecessary as it is, unless it depends on DistSyncFuncTest instead of BaseDistFunctionalTest

core/src/test/java/org/infinispan/distribution/DistAsyncFuncTest.java
```diff
@@ -39,12 +56,44 @@ protected void createCacheManagers() throws Throwable {
...
+            @Override
+            public CompletableFuture<Map<Address, Response>> invokeRemotelyAsync(Collection<Address> recipients, ReplicableCommand rpc, RpcOptions options) {
+               ReplicableCommand command = rpc;
+               if (rpc instanceof SingleRpcCommand) {
+                  command = ((SingleRpcCommand) rpc).getCommand();
+               }
+               if (command instanceof InvalidateL1Command) {
+                  InvalidateL1Command invalidateL1Command = (InvalidateL1Command) command;
+                  if (recipients != null) {
+                     recipients.stream().map(address -> caches.stream()
```
karesti: for lisibility, could we pass just cache address already filtered here ?

karesti: forget about it, I just got too long to understand

core/src/test/java/org/infinispan/distribution/DisabledL1Test.java
rvansa: Removed...

rvansa:
@karesti There was one (non-random) test failure as this PR is a cherry-pick from other branch, I've fixed that one.

tristantarrant:
Merged, thanks

danberindei:
@rvansa I think the idea was for the test to also check that the cache properly keeps track of L1 requestors, so the L1 invalidations are sent to the nodes that the test expects. Or is there a separate test for that?

rvansa:
@danberindei There should be a separate test, synchronous variant was not checking the invalidations. The current test messed together the invalidations from previous puts (e.g. from the intialization). We could add an assertion which invalidations should appear in `expectedL1Invalidations` as well.

PR #5391, opened 2017-08-24
ISPN-8222 Jenkins should archive all logs

slaskawi:
https://issues.jboss.org/browse/ISPN-8222


slaskawi:
Just some numbers for comparison:

* PR-5390 archive takes 18 MB
* this PR takes 19 MB

So we should be good in terms of disk space. @danberindei Good job with `xz` compression!

danberindei:
Integrated, thanks Sebastian!

PR #5392, opened 2017-08-24
ISPN-8186 Avoid client socket connection to itself

galderz:
https://issues.jboss.org/browse/ISPN-8186

* When a server stops, the client can still try to establish a
connection with a closing server until the new view has been
received.
* When opening the connection, there's a chance that
the client port and the local port are the same (since the
server por might have been vacated while stopping.
* The fix proposed verifies that if a connection is established,
the port and local port are not the same, in which case the socket
is closed and retried.


galderz:
Tests failing, looking into it. Closing for now.

galderz:
Hmmm, can't seem to get CI to find the commit. I'll open a new PR.

PR #5399, opened 2017-08-30
ISPN-6952 JCacheTwoCachesAnnotationsTest no need to clear caches befo…

tristantarrant:
…re killing CMs

https://issues.jboss.org/browse/ISPN-6952


tristantarrant:
I have a new strategy. Will push it soon.

tristantarrant:
@danberindei here is my latest proposal:
- create/destroy explicit cachemanagers only before/after test (and not per-method)
- clear caches after method

The implicit cachemanager is created on the fist invocation to JCacheAnnotatedClass's methods and destroyed on undeployment.

danberindei:
Integrated, thanks Tristan!

PR #5400, opened 2017-08-31
ISPN-8248 Missing timeout exceptions in distributed executor tests

danberindei:
https://issues.jboss.org/browse/ISPN-8248

Use a latch instead of Thread.sleep().


tristantarrant:
Merged to master, thanks

PR #5401, opened 2017-08-31
Jenkinsfile: debug build failures caused by git clean

danberindei:


danberindei:
Updated to use strace (which I installed manually on the agents).

danberindei:
`strace` shows that the Karaf JVM is still running and is creating new files while `git clean` is recursively removing the directory:

    rmdir("integrationtests/osgi/target/pax/2aebbeb2-13f0-4315-90e1-c8702da04aee") = 0
    rmdir("integrationtests/osgi/target/pax") = -1 ENOTEMPTY (Directory not empty)

I've updated the PR to ignore the exit code of `git clean`, and separately I'm working on upgrading Karaf to fix the shutdown.

tristantarrant:
Mered

PR #5403, opened 2017-09-01
ISPN-8248 Missing timeout exceptions in distributed executor tests

danberindei:
https://issues.jboss.org/browse/ISPN-8248

Make SleepingSimpleCallable close the latch automatically

The previous PR didn't close/open the latch for some test methods, so they could pass or fail depending on the order they ran in.


tristantarrant:
Merged, thanks

PR #5404, opened 2017-09-04
ISPN-8255 Smoke test suite for hibernate-cache module

mgencur:
https://issues.jboss.org/browse/ISPN-8255


commons-test/src/main/java/org/infinispan/commons/test/categories/Smoke.java
```diff
@@ -0,0 +1,7 @@
+package org.infinispan.commons.test.categories;
+
+/**
+ * {@link org.junit.experimental.categories.Category} tag for smoke tests.
+ */
+public class Smoke {
```
rvansa: Is there any advantage to using class/final class/interface as the category class? JUnit wiki uses interface in the example code.

mgencur: Both a class and interface work. This is just a marker that the test class annotated with this type should be picked by the test suite runner. I don't see any difference.

mgencur: Thinking about it a little more. We can build a hierarchy of categories with both interfaces and non-final classes.

galderz:
Let's wait for CI, the rest looks good to me.

rvansa:
CI looks good, integrating.

PR #5405, opened 2017-09-04
ISPN-8026 Size related fixes in Hibernate Cache tests

galderz:
A bunch of fixes to avoid the following problems:
* Async puts from tests that linger affecting other tests. This applies to both remove put from load calls, and remove end invalidation calls.
* Wait for end invalidation and putFromLoad calls before asserting cache size. Although cache gets might work fine, getting an accurate size count requires all remote ops to end well, so that streams can be correctly processed, filtered and counted.

### DIFF
--- a/core/src/main/java/org/infinispan/container/DefaultDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/DefaultDataContainer.java
@@ -413,6 +413,7 @@ public class DefaultDataContainer<K, V> implements DataContainer<K, V> {
          while (it.hasNext()) {
             InternalCacheEntry<K, V> entry = it.next();
             if (includeExpired || !entry.canExpire()) {
+               log.tracef("Return next entry %s", entry);
                return entry;
             } else {
                if (!initializedTime) {
@@ -420,10 +421,14 @@ public class DefaultDataContainer<K, V> implements DataContainer<K, V> {
                   initializedTime = true;
                }
                if (!entry.isExpired(now)) {
+                  log.tracef("Return next entry %s", entry);
                   return entry;
+               } else {
+                  log.tracef("%s is expired", entry);
                }
             }
          }
+         log.tracef("Return next null");
          return null;
       }
 
--- a/core/src/main/java/org/infinispan/stream/impl/TerminalFunctions.java
+++ b/core/src/main/java/org/infinispan/stream/impl/TerminalFunctions.java
@@ -43,11 +43,15 @@ import org.infinispan.IntCacheStream;
 import org.infinispan.LongCacheStream;
 import org.infinispan.commons.marshall.Externalizer;
 import org.infinispan.commons.marshall.SerializeWith;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
 
 /**
  * Static factory class used to provide marshallable terminal operations
  */
 final class TerminalFunctions {
+   private static final Log log = LogFactory.getLog(TerminalFunctions.class);
+
    private TerminalFunctions() { }
 
    public static <T> Function<Stream<T>, Boolean> allMatchFunction(Predicate<? super T> predicate) {
@@ -615,7 +619,9 @@ final class TerminalFunctions {
 
       @Override
       public Long apply(Stream<T> stream) {
-         return stream.count();
+         long count = stream.count();
+         log.tracef("Count value is %d", count);
+         return count;
       }
 
       public static final class CountFunctionExternalizer implements Externalizer<CountFunction> {
--- a/hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/PutFromLoadValidator.java
+++ b/hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/PutFromLoadValidator.java
@@ -224,7 +224,7 @@ public class PutFromLoadValidator {
          chain.addInterceptor(nonTxPutFromLoadInterceptor, entryWrappingPosition);
 			validator.nonTxPutFromLoadInterceptor = nonTxPutFromLoadInterceptor;
 		}
-		log.debug("New interceptor chain is: " + cache.getInterceptorChain());
+		log.debug("New interceptor chain is: " + cache.getAsyncInterceptorChain());
 
 		CacheCommandInitializer cacheCommandInitializer = cache.getComponentRegistry().getComponent(CacheCommandInitializer.class);
 		cacheCommandInitializer.addPutFromLoadValidator(cache.getName(), validator);
@@ -587,7 +587,7 @@ public class PutFromLoadValidator {
 					}
 					long now = regionFactory.nextTimestamp();
                if (trace) {
-                  log.tracef("endInvalidatingKey(%s#%s, %s) remove invalidator from %s", cache.getName(), key, lockOwnerToString(lockOwner), pending);
+                  log.tracef("beginInvalidatingKey(%s#%s, %s) remove invalidator from %s", cache.getName(), key, lockOwnerToString(lockOwner), pending);
                }
 					pending.invalidate(now);
 					pending.addInvalidator(lockOwner, valueForPFER, now);
--- a/hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/PutFromLoadValidator.java
+++ b/hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/PutFromLoadValidator.java
@@ -185,7 +185,7 @@ public class PutFromLoadValidator {
 	public static void addToCache(AdvancedCache cache, PutFromLoadValidator validator) {
       AsyncInterceptorChain chain = cache.getAsyncInterceptorChain();
       List<AsyncInterceptor> interceptors = chain.getInterceptors();
-      log.debug("Interceptor chain was: " + interceptors);
+      log.debugf("Interceptor chain was: ", interceptors);
       int position = 0;
 		// add interceptor before uses exact match, not instanceof match
 		int invalidationPosition = 0;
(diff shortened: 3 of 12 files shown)


galderz:
One more thing, I've noticed this failure in CI:

    org.infinispan.test.hibernate.cache.functional.cluster.NaturalIdInvalidationTest.testAll[nonstrict, REPL_SYNC]

It's likely that failure is due to leftover putFromLoad calls that affect other tests. This is precisely the scenario this PR is trying to fix.

So, let's wait for CI to run for this PR and see CI after this PR is integrated before checking if any more of this failure exists.

galderz:
I've fixed the checkstyle error

galderz:
Last two runs failed in a weird way, but before that [this run](http://ci.infinispan.org/job/Infinispan/job/PR-5405/2/) went fine. No Hibernate Cache errors there.

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
```diff
@@ -413,6 +413,7 @@ public void onEntryRemoved(Entry<K, InternalCacheEntry<K, V>> entry) {
          while (it.hasNext()) {
             InternalCacheEntry<K, V> entry = it.next();
             if (includeExpired || !entry.canExpire()) {
+               log.tracef("Return next entry %s", entry);
```
rvansa: If you committed these on purpose, it should be guarded by the `if (trace) { ... }`

core/src/main/java/org/infinispan/stream/impl/TerminalFunctions.java
```diff
@@ -615,7 +619,9 @@ private CountFunction() { }
 
       @Override
       public Long apply(Stream<T> stream) {
-         return stream.count();
+         long count = stream.count();
+         log.tracef("Count value is %d", count);
```
rvansa: even though it's trace level, I am not sure if we should log it...

hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/PutFromLoadValidator.java
```diff
@@ -224,7 +224,7 @@ public static void addToCache(AdvancedCache cache, PutFromLoadValidator validato
          chain.addInterceptor(nonTxPutFromLoadInterceptor, entryWrappingPosition);
 			validator.nonTxPutFromLoadInterceptor = nonTxPutFromLoadInterceptor;
 		}
-		log.debug("New interceptor chain is: " + cache.getInterceptorChain());
+		log.debug("New interceptor chain is: " + cache.getAsyncInterceptorChain());
```
rvansa: `debugf` please

hibernate-cache/src/main/java/org/infinispan/hibernate/cache/access/TombstoneCallInterceptor.java
```diff
@@ -194,19 +194,8 @@ public Object visitSizeCommand(InvocationContext ctx, SizeCommand command) throw
...
-				}
-			}
-		}
-		finally {
-         it.close();
-		}
-		return size;
+      return (int) decoratedCache.cacheEntrySet().stream()
+         .filter(CacheFilters.predicate(Tombstone.EXCLUDE_TOMBSTONES))
+         .count();
```
rvansa: the `(int)` cast can cause overflow, use `Math.min(Integer.MAX_VALUE, (...).count())` or `Math.toIntExact(...)`.

hibernate-cache/src/main/java/org/infinispan/hibernate/cache/util/Tombstone.java
```diff
@@ -191,7 +195,9 @@ private ExcludeTombstonesFilter() {}
 
 		@Override
 		public boolean accept(Object key, Object value, Metadata metadata) {
-			return !(value instanceof Tombstone);
+         boolean b = !(value instanceof Tombstone);
+         log.tracef("Is value %s for key %s is tombstone? %b", value, key, b);
```
rvansa: I wouldn't log here...

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/AbstractRegionAccessStrategyTest.java
```diff
@@ -269,6 +278,31 @@ protected void putFromLoadTest(final boolean useMinimalAPI, boolean isRemoval) t
 		}
 	}
 
+   protected void assertPutFromLoadLatches(CountDownLatch[] latches) {
+      if (!await(latches[0]) && !await(latches[1]))
```
rvansa: `assertTrue("One of...", await(latches[0]) || await(latches[1]));` ?

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/AbstractRegionAccessStrategyTest.java
```diff
@@ -184,39 +186,42 @@ protected void putFromLoadTest(final boolean useMinimalAPI, boolean isRemoval) t
 		final CountDownLatch writeLatch2 = new CountDownLatch(1);
 		final CountDownLatch completionLatch = new CountDownLatch(2);
 
+      CountDownLatch[] putFromLoadLatches = new CountDownLatch[2];
```
rvansa: it seems that you could have a separate variable for each latch rather than an array

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/AbstractRegionAccessStrategyTest.java
```diff
@@ -592,4 +652,6 @@ private CountDownLatch expectRemotePutFromLoad(AdvancedCache localCache, Advance
 		}
 		return putFromLoadLatch;
 	}
+
```
rvansa: extra newlines?

core/src/main/java/org/infinispan/stream/impl/TerminalFunctions.java
galderz: Definitely needs logging. The count is computed out of values in local and remotes, and there's no way to see where each part of the count comes from without this.

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
galderz: Does not need `if (trace)` because there's nothing precomputed before calling `tracef`. Adding `if (trace)` in this case is repeating the check the `log.tracef` does inside without no extra optimisation to be gained here.

hibernate-cache/src/main/java/org/infinispan/hibernate/cache/util/Tombstone.java
galderz: Why not? I added this because I had no way of knowing which of the two entries in the cache that was being returned (instead of expected one entry) was a tombstone...

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/AbstractRegionAccessStrategyTest.java
galderz: That would require quite a bit of rearchitecture of the code, which I wanted to avoid just to fix it.

If you separate into two variables, these variables need to be modified inside the lambda which is not allowed, so then you need to start creating separate Runnable classes that take all the state variables touched as parameters... etc.

TBH, the code is quite messy since it's calling for Callables rather this manual way of tracking exceptions...

galderz:
Added a commit fixing those small issues agreed by both.

@rvansa There might be some log messages that should be removed, but the ones pointed out are mostly fine. The ones that I'm maybe considering removing are the ones inside the data container, particularly in the get() call. Once I squash I'll review them and decide which ones to remove.

hibernate-cache/src/test/java/org/infinispan/test/hibernate/cache/AbstractRegionAccessStrategyTest.java
rvansa: I see, sorry - the broken indentation got me confused.

rvansa:
@galderz 170 related failures in CI :-/

galderz:
> @galderz 170 related failures in CI :-/

LOL... I'll fix that

galderz:
I've fixed the issue, I'll squash and resend shortly

galderz:
@rvansa I've squashed commits.

galderz:
I've just checked CI and still shows NaturalIdInvalidationTest failures. These are not specifically fixed by the commits in this PR, though I was hoping they might. Also, these failures were already present independent on this PR. So let's integrate this PR and I'll look into these failures later.

    org.infinispan.test.hibernate.cache.functional.cluster.NaturalIdInvalidationTest.testAll[nonstrict, REPL_SYNC]
    org.infinispan.test.hibernate.cache.functional.cluster.NaturalIdInvalidationTest.testAll[nonstrict, DIST_SYNC]

galderz:
I've created [ISPN-8263](https://issues.jboss.org/browse/ISPN-8263) to track that.

rvansa:
@galderz OK, there are still 8 commits - will you squash them?

galderz:
@rvansa Yeah, I wanted to keep the remaining commits separate since each fixes a particular situation, and the remaining two add logging and fix the counter overflow issue. I'll rebase ASAP...

galderz:
@rvansa Rebased

rvansa:
Integrated, thanks!

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
danberindei: @galderz It's not exactly the same: because `trace` is a static final, when it's disabled the JIT doesn't have to generate any code for the `log.tracef` call. While the log level check is much simpler in log4j2 than in log4j1, it's still more than 0 instructions.

Also, "all `tracef` calls must be guarded" is easier to follow than "all `tracef` calls that have more than 3 parameters or a method call must be guarded" ;)

PR #5408, opened 2017-09-05
ISPN-8227 OptimisticTxPartitionAndMergeDuringPrepareTest.testPrimaryOwnerIsolatedPartition fails randomly

rvansa:
https://issues.jboss.org/browse/ISPN-8227

### DIFF
--- a/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
+++ b/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
@@ -33,6 +33,7 @@ import org.infinispan.commands.remote.recovery.TxCompletionNotificationCommand;
 import org.infinispan.commands.tx.RollbackCommand;
 import org.infinispan.commands.write.WriteCommand;
 import org.infinispan.commons.CacheException;
+import org.infinispan.commons.util.ByRef;
 import org.infinispan.commons.util.CollectionFactory;
 import org.infinispan.commons.util.Util;
 import org.infinispan.configuration.cache.Configuration;
@@ -354,6 +355,7 @@ public class TransactionTable implements org.infinispan.transaction.TransactionT
     * Atomicity: this method supports concurrent invocations, guaranteeing that all threads will see the same
     * transaction object.
     */
+   // TODO: consider returning null instead of throwing exception when the transaction is already completed
    public RemoteTransaction getOrCreateRemoteTransaction(GlobalTransaction globalTx, WriteCommand[] modifications) {
       return getOrCreateRemoteTransaction(globalTx, modifications, currentTopologyId);
    }
@@ -370,8 +372,7 @@ public class TransactionTable implements org.infinispan.transaction.TransactionT
 
       int viewId = rpcManager.getTransport().getViewId();
       if (!rpcManager.getTransport().getMembers().contains(globalTx.getAddress())) {
-         throw new IllegalStateException(
-               "Cannot create remote transaction " + globalTx + ", the originator is not in the cluster view");
+         throw log.remoteTransactionOriginatorNotInView(globalTx);
       }
 
       RemoteTransaction newTransaction = modifications == null ? txFactory.newRemoteTransaction(globalTx, topologyId)
@@ -383,7 +384,7 @@ public class TransactionTable implements org.infinispan.transaction.TransactionT
             return existing;
          } else {
             if (isTransactionCompleted(gtx)) {
-               throw new IllegalStateException("Cannot create remote transaction " + gtx + ", already completed");
+               throw log.remoteTransactionAlreadyCompleted(gtx);
             }
             if (trace)
                log.tracef("Created and registered remote transaction %s", newTransaction);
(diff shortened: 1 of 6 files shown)


tristantarrant:
Its counterpart (duringCommit) failed at least once in CI: http://ci.infinispan.org/job/Infinispan/job/PR-5408/10/testReport/junit/org.infinispan.partitionhandling/OptimisticTxPartitionAndMergeDuringCommitTest/testOriginatorIsolatedPartitionWithDiscard_DIST_SYNC_/

Can the fix be applied to both ?

rvansa:
@tristantarrant Should be, IMO... I'll try to reproduce it locally.

rvansa:
Hmm, the original test failed every now and then on my machine, but I can't get the failure from CI.

pruivo:
aren't you hitting https://issues.jboss.org/browse/ISPN-8232? it is going to require some work...

core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
```diff
@@ -471,10 +473,17 @@ public void remoteTransactionCommitted(GlobalTransaction gtx, boolean onePc) {
    }
 
    public final RemoteTransaction removeRemoteTransaction(GlobalTransaction txId) {
-      RemoteTransaction removed = remoteTransactions.remove(txId);
-      if (trace) log.tracef("Removed remote transaction %s ? %s", txId, removed);
-      releaseResources(removed);
-      return removed;
+      ByRef<RemoteTransaction> removed = new ByRef<>(null);
+      // we need to mark the transaction inside compute() even if it does not exist
```
danberindei: Why not mark the tx as completed before removing it from the map?

rvansa:
@pruivo That might be the cause for the other failure; here I've identified a race in data structures access and I am fixing that one.

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
```diff
@@ -167,7 +168,15 @@ public void run() {
             ArrayDeque<BlockingRunnable> readyList = new ArrayDeque<>(size);
             for (Iterator<BlockingRunnable> iterator = blockedTasks.iterator(); iterator.hasNext(); ) {
                BlockingRunnable runnable = iterator.next();
-               if (runnable.isReady()) {
+               boolean ready;
+               try {
+                  ready = runnable.isReady();
+               } catch (Exception e) {
+                  log.debugf(e, "Failed to check ready state of %s, dropping.", runnable);
```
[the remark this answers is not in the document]
rvansa: Would you like me to add a `getCommand()` method to `BlockingRunnable` iface? I've rather added `toString()` impl to `DefaultTopologyRunnable`; I could add it to other implementing classes as well.

core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
rvansa: It seems I could; I thought I have an counterexample but probably I was mistaken.

rvansa: Actually I need to see if the tx was `successful`; trying to mark before remove/compute means that I'd need to do another lookup into `remoteTransaction` anyway.

core/src/main/java/org/infinispan/util/concurrent/BlockingTaskAwareExecutorServiceImpl.java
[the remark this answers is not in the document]
danberindei: Adding `toString()` to all the `BlockingRunnable` impls is fine.

core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
danberindei: I missed that, let's keep it in `compute()` then.

rvansa:
@danberindei Added those `toString()`s.

core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
```diff
@@ -370,7 +372,7 @@ private RemoteTransaction getOrCreateRemoteTransaction(GlobalTransaction globalT
 
       int viewId = rpcManager.getTransport().getViewId();
       if (!rpcManager.getTransport().getMembers().contains(globalTx.getAddress())) {
-         throw new IllegalStateException(
+         throw new CacheException(
```
tristantarrant: Possibly this deserves a localized message

core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
```diff
@@ -383,7 +385,7 @@ private RemoteTransaction getOrCreateRemoteTransaction(GlobalTransaction globalT
             return existing;
          } else {
             if (isTransactionCompleted(gtx)) {
-               throw new IllegalStateException("Cannot create remote transaction " + gtx + ", already completed");
+               throw new CacheException("Cannot create remote transaction " + gtx + ", already completed");
```
tristantarrant: Localized message ?

rvansa:
Localized the exceptions.

tristantarrant:
Merged, thanks

PR #5410, opened 2017-09-06
Remove DecoratedCache from docs

rvansa:
* DecoratedCache was moved to private package in ISPN-4074


tristantarrant:
Merged, thanks

PR #5413, opened 2017-09-07
ISPN-8249 EvictionFunctionalTest.testSimpleExpirationMaxIdle random failure

rvansa:
* Fix OffHeapDataContainer.compute() and optimize some other methods
* Trigger expired entries processing manually
* Drop EvictionFunctionalTest.testMultiThreaded which does not test anything

https://issues.jboss.org/browse/ISPN-8249

### DIFF
--- a/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
@@ -182,6 +182,10 @@ public class OffHeapDataContainer implements DataContainer<WrappedBytes, Wrapped
     */
    protected void performPut(long newAddress, WrappedBytes key) {
       long address = memoryLookup.getMemoryAddress(key);
+      performPut(address, newAddress, key);
+   }
+
+   protected void performPut(long address, long newAddress, WrappedBytes key) {
       boolean shouldCreate = false;
       // Have to start new linked node list
       if (address == 0) {
@@ -352,19 +356,12 @@ public class OffHeapDataContainer implements DataContainer<WrappedBytes, Wrapped
    public int size() {
       long time = timeService.wallClockTime();
       long count = entryStream().filter(e -> !e.isExpired(time)).count();
-      if (count > Integer.MAX_VALUE) {
-         return Integer.MAX_VALUE;
-      }
-      return (int) count;
+      return (int) Math.min(count, Integer.MAX_VALUE);
    }
 
    @Override
    public int sizeIncludingExpired() {
-      long currentSize = size.get();
-      if (currentSize > Integer.MAX_VALUE) {
-         return Integer.MAX_VALUE;
-      }
-      return (int) currentSize;
+      return (int) Math.min(size.get(), Integer.MAX_VALUE);
    }
 
    @Override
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -494,9 +491,13 @@ public void evict(WrappedBytes key) {
...
       try {
          checkDeallocation();
-         // TODO: this could be more efficient
-         passivator.passivate(get(key));
-         remove(key);
+         long address = memoryLookup.getMemoryAddress(key);
+         if (address != 0) {
+            // TODO: this could be more efficient
+            InternalCacheEntry<WrappedBytes, WrappedBytes> ice = performGet(address, key);
+            passivator.passivate(ice);
```
wburns: We should do a null check for the ice and only perform these 2 lines if it exists. This way we prevent a possible loop around the memory address linked list if it is present.

wburns: We should add an if check here for ice and only do this line and the next if it is not null. Prevents another possible loop through the linked list if it exists.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -509,13 +510,16 @@ public void evict(WrappedBytes key) {
       lock.lock();
       try {
          checkDeallocation();
-         InternalCacheEntry<WrappedBytes, WrappedBytes> prev = get(key);
+         long address = memoryLookup.getMemoryAddress(key);
+         InternalCacheEntry<WrappedBytes, WrappedBytes> prev = address == 0 ? null : performGet(address, key);
```
wburns: Are you just trying to avoid the read lock acquisition here?

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -509,13 +510,16 @@ public void evict(WrappedBytes key) {
...
          InternalCacheEntry<WrappedBytes, WrappedBytes> result = action.compute(key, prev, internalEntryFactory);
-         if (result != null) {
+         if (prev == result) {
```
wburns: Good catch.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
rvansa: I am trying to avoid any redundancy and organize publicly accessible vs. internal. Both the lock acquisition and deallocation check should be in the publicly accessible methods, and public methods should call only internal...

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
rvansa: OK, I was only inlining the calls, but I can add it.

rvansa:
@wburns added the null check in `evict`

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
wburns: No worries, I just wanted to make sure I wasn't missing something that was "fixed" and didn't quite understand.

wburns:

PR #5414, opened 2017-09-08
Ispn karaf upgrade

danberindei:
ISPN-8261 OSGi integration tests can leak the Karaf process
https://issues.jboss.org/browse/ISPN-8261

* Kill any Karaf containers that are still running at the end of the tests
* Go back to using only categories PerSuite, PerClass, and PerMethod
  Partially reverting https://github.com/infinispan/infinispan/pull/5378
* Avoid NPE in JUnitTestListener when the container fails to start
* Add dependency on core to the RocksDB feature
* Remove explicit dependency to org.osgi.enterprise
 The extra dependency caused a reload, which reloaded the
 org.ops4j.pax.logging.pax-logging-api bundle, which in
 turn reloaded all the infinispan bundles while the tests
 were running.
* Improve logging when the pax logging service is being restarted
* Update ServiceFinder to use FrameworkUtil
* Add example dependencies for debugging through surefire and the Karaf test container

ISPN-8262 Upgrade to Karaf 4.1.2
https://issues.jboss.org/browse/ISPN-8262

* Add explicit dependencies for the wrap feature (wrap: protocol)


tristantarrant:
Merged, good work !

PR #5416, opened 2017-09-08
ISPN-8239 Update marshalling tests and documentation

vjuranek:
https://issues.jboss.org/browse/ISPN-8239

### DIFF
--- a/core/src/test/java/org/infinispan/marshall/DefensiveCopyTest.java
+++ b/core/src/test/java/org/infinispan/marshall/DefensiveCopyTest.java
@@ -7,6 +7,7 @@ import java.net.URL;
 import java.net.URLClassLoader;
 
 import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.configuration.cache.StorageType;
 import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.test.SingleCacheManagerTest;
 import org.infinispan.test.data.Key;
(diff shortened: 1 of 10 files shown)


core/src/test/java/org/infinispan/marshall/DefensiveCopyTest.java
```diff
@@ -50,7 +51,7 @@ public void testSafetyAfterRetrieving() {
       Person cachedPerson = this.<Integer, Person>cache().get(k);
       assertEquals(person, cachedPerson);
       cachedPerson.setName("Mr Digweed");
-      assertEquals(new Person("Mr Coe"), cache.get(k));
+      assertEquals(person, cache.get(k));
```
rvansa: This is not right; if `person == cachedPerson` and Infinispan did not make a defensive copy, this would succceed.

documentation/src/main/asciidoc/user_guide/marshalling.adoc
```diff
@@ -436,6 +439,7 @@ This is the list of Externalizer identifiers that are used by Infinispan based m
 |Infinispan Scripting Module:|1800 - 1849
 |Infinispan Server Event Logger Module:|1850 - 1899
 |Infinispan Remote Store:|1900 - 1999
+|Infinispan Counters:|2000 - 2100
```
rvansa: Counters will have only 50 reserved ids, see https://github.com/infinispan/infinispan/pull/5193/files#diff-09c3fc56ad86ab15b4e36314131a7d59

vjuranek: fixed

core/src/test/java/org/infinispan/marshall/DefensiveCopyTest.java
vjuranek: If I understand tests correctly, this is tested in previous `testOriginalReferenceSafety`, this one should test that modification to instance obtained from cache doesn't affect instance in the cache and thus creating new object in assert is redundant.

rvansa:
Integrated, thanks.

PR #5117, opened 2017-05-04
ISPN-7680 - enabled karaf-maven-plugin to validate Karaf features

galderz:
Fixed conflicts in @paoloantinori's fix. Thx @paoloantinori for contributing!

PR #5418, opened 2017-09-08
ISPN-7680 Enable karaf-maven-plugin to validate Karaf features

danberindei:
https://issues.jboss.org/browse/ISPN-7680

Replacement for https://github.com/infinispan/infinispan/pull/5117. I fixed some indentation issues in-place, but I changed the plugin versions in a separate commit.

If would have been nice if I'd found this before finishing https://github.com/infinispan/infinispan/pull/5414, as having the proper dependencies would have avoided a lot of stepping through Karaf code :)

### DIFF
--- a/all/embedded/pom.xml
+++ b/all/embedded/pom.xml
@@ -134,6 +134,13 @@
       </dependency>
 
       <dependency>
+        <groupId>org.apache.karaf.features</groupId>
+        <artifactId>framework</artifactId>
+        <type>kar</type>
+        <scope>provided</scope>
+      </dependency>
+
+      <dependency>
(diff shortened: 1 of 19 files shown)


danberindei:
This is breaking `SampleConfigFilesCorrectnessTest` in `core`, I'll push an update soon.

danberindei:
The new Karaf framework was adding the `paxlogging-api` dependency to the test classpath, and it was replacing log4j2's `LoggerContext` with a custom implementation. I couldn't remove the dependency completely, but I excluded all its dependencies.

tristantarrant:
Merged, thanks

PR #5419, opened 2017-09-08
ISPN-8174 LocalStreamIteratorWithPassivationTest random failures

wburns:
* Disabled test as we don't support concurrent activation with iteration

https://issues.jboss.org/browse/ISPN-8174

### DIFF
--- a/core/src/test/java/org/infinispan/stream/LocalStreamIteratorWithPassivationTest.java
+++ b/core/src/test/java/org/infinispan/stream/LocalStreamIteratorWithPassivationTest.java
@@ -75,7 +75,7 @@ public class LocalStreamIteratorWithPassivationTest extends DistributedStreamIte
       TestObjectStreamMarshaller sm = new TestObjectStreamMarshaller();
       PersistenceManager pm = null;
       try {
-         store.write(new MarshalledEntryImpl(loaderKey, loaderValue, null, sm));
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/infinispan/stream/LocalStreamIteratorWithPassivationTest.java
```diff
@@ -113,7 +113,7 @@ public void testConcurrentActivation() throws InterruptedException, ExecutionExc
       }
    }
 
-   @Test
+   @Test(enabled = false, description = "This requires supporting concurrent activation in cache loader interceptor")
```
tristantarrant: You are disabling a test as well as changing it ?

wburns: This test didn't really test anything, hence why it was working most of the time :D

This test is supposed to test the filtering of the store to work while a concurrent activation occurs. Which we don't support concurrent activation with stores when doing iteration or state transfer. In those cases we will actually miss entries. This test was updated to more properly reflect that. Hence why it must now be disabled, since it doesn't work until we can finally fix the concurrent activation bug.

tristantarrant: Understood

tristantarrant:
Merged.

PR #5420, opened 2017-09-08
ISPN-8110 OffHeapBoundedDataContainer.ensureSize() busy loop takes too

wburns:
long

* Replaced yield with lock acquisition and retry

https://issues.jboss.org/browse/ISPN-8110

This should help resolve timeout issues in test when TRACE is enabled.

### DIFF
--- a/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
@@ -14,8 +14,10 @@ import org.infinispan.metadata.Metadata;
  * Data Container implementation that stores entries in native memory (off-heap) that is also bounded.  This
  * implementation uses a simple LRU doubly linked list off-heap guarded by a single lock.
  * <p>
- * The link list consists of 28 bytes (3 longs and 1 int).  The first long is the actual entry address, the second is the
- * previous pointer, the third is the next pointer and lastly the int is the hashCode of the key to retrieve the lock.
+ * The link list is represented by firstAddress as the had of the list and lastAddress as the tail of the list. Each
+ * entry in the list consists of 28 bytes (3 longs and 1 int), the first long is the actual entry address, the second is
+ * a pointer to the previous element in the list, the third is the next pointer and lastly the int is the hashCode of
+ * the key to retrieve the lock. The hashCode is required to know which lock to use when trying to read the entry.
  * @author wburns
  * @since 9.0
  */
@@ -190,34 +192,66 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
       super.performClear();
    }
 
+   /**
+    * This method repeatedly removes the head of the LRU list until there the current size is less than or equal to `maxSize`.
+    *
+    * We need to hold the LRU lock in order to check the current size and to read the head entry,
+    * and then we need to hold the head entry's write lock in order to remove it.
+    * The problem is that the correct acquisition order is entry write lock first, LRU lock second,
+    * and we need to hold the LRU lock so that we know which entry write lock to acquire.
+    *
+    * To work around it, we first try to acquire the entry write lock without blocking.
+    * If that fails, we release the LRU lock and we acquire the locks in the correct order, hoping that
+    * the LRU head doesn't change while we wait. Because the entry write locks are striped, we actually
+    * tolerate a LRU head change as long as the new head entry is in the same lock stripe.
+    * If the LRU list head changes, we release both locks and try again.
+    */
    private void ensureSize() {
+
       while (true) {
          long addressToRemove;
          Lock entryWriteLock;
          lruLock.lock();
          try {
-            if (currentSize > maxSize) {
-               // Retrieve the hashCode so we can lock it to verify the address is still present
-               int hashCode = UNSAFE.getInt(firstAddress + 24);
-               entryWriteLock = locks.getLockFromHashCode(hashCode).writeLock();
-               if (!entryWriteLock.tryLock()) {
-                  // Attempts to release the lru lock and reacquire the write lock are problematic as underlying allocator
-                  // may return same memory address between points, so there is no way to verify we have same object
-                  // in an efficient way - just force loop back around to allow other threads to get lruLock
-                  addressToRemove = 0;
-               } else {
-                  addressToRemove = UNSAFE.getLong(firstAddress);
-               }
-            } else {
-               // This is the only way to break out of loop
+            if (currentSize <= maxSize) {
                break;
+            }
+            int hashCode = UNSAFE.getInt(firstAddress + 24);
+            entryWriteLock = locks.getLockFromHashCode(hashCode).writeLock();
+            if (!entryWriteLock.tryLock()) {
+               addressToRemove = 0;
+            } else {
+               addressToRemove = UNSAFE.getLong(firstAddress);
             }
          } finally {
             lruLock.unlock();
          }
 
-         // We can proceed with removal if it is non zero.  It would be 0 if the firstAddress changed
-         // It is assumed we own the entryWriteLock lock as well so we can read the keys.
+         if (addressToRemove == 0) {
+            entryWriteLock.lock();
+            try {
+               lruLock.lock();
+               try {
+                  if (currentSize <= maxSize) {
+                     break;
+                  }
+                  int hashCode = UNSAFE.getInt(firstAddress + 24);
+                  Lock innerLock = locks.getLockFromHashCode(hashCode).writeLock();
+                  if (innerLock == entryWriteLock) {
+                     addressToRemove = UNSAFE.getLong(firstAddress);
+                  } else {
+                     addressToRemove = 0;
+                  }
+               } finally {
+                  lruLock.unlock();
+               }
+            } finally {
+               if (addressToRemove == 0) {
+                  entryWriteLock.unlock();
+               }
+            }
+         }
+
          if (addressToRemove != 0) {
             if (trace) {
                log.tracef("Removing entry: %d due to eviction due to size %d being larger than maximum of %d",
@@ -230,9 +264,6 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
             } finally {
                entryWriteLock.unlock();
             }
-         } else {
-            // Just to let another thread possibly continue and release its lock
-            Thread.yield();
          }
       }
    }
--- a/core/src/test/java/org/infinispan/container/offheap/OffHeapBoundedSingleNodeTest.java
+++ b/core/src/test/java/org/infinispan/container/offheap/OffHeapBoundedSingleNodeTest.java
@@ -26,7 +26,7 @@ import org.testng.annotations.Test;
 @Test(groups = "functional", testName = "container.offheap.OffHeapBoundedSingleNodeTest")
 public class OffHeapBoundedSingleNodeTest extends OffHeapSingleNodeTest {
 
-   private static final int COUNT = 100;
+   private static final int COUNT = 51;
 
    @Override
    protected void createCacheManagers() throws Throwable {


wburns:
@danberindei Can you confirm if this fixes ISPN-8111 for you?

danberindei:
@wburns `OffHeapBoundedSingleNodeTest` passes for me now, but it takes 9.1s, so it would very likely fail on a CI agent. Is there any particular reason for the off-heap tests to use transactions? And in the following log fragment only the first and last message seem to say anything useful :)

```
10:53:30,936 TRACE (testng-Test:[]) [BoundedOffHeapDataContainer] Removing entry: 140173309347200 due to eviction due to size 101 being larger than maximum of 100
10:53:30,937 TRACE (testng-Test:[]) [UnsafeWrapper] Copying memory of object null offset by 140173309347216 to [B@6df3e41f offset by 16 with a total of 17 bytes
10:53:30,937 TRACE (testng-Test:[]) [UnsafeWrapper] Copying memory of object null offset by 140173309347233 to [B@1248ff4a offset by 16 with a total of 11 bytes
10:53:30,938 TRACE (testng-Test:[]) [UnsafeWrapper] Copying memory of object null offset by 140173309347244 to [B@3423369b offset by 16 with a total of 0 bytes
10:53:30,938 TRACE (testng-Test:[]) [UnsafeWrapper] Copying memory of object null offset by 140173309347244 to [B@4c847eef offset by 16 with a total of 13 bytes
10:53:30,939 TRACE (testng-Test:[]) [UnsafeWrapper] Copying memory of object null offset by 140173309347216 to [B@3e25b87c offset by 16 with a total of 17 bytes
10:53:30,939 TRACE (testng-Test:[]) [UnsafeWrapper] Copying memory of object null offset by 140173309347233 to [B@6b42a25d offset by 16 with a total of 11 bytes
10:53:30,940 TRACE (testng-Test:[]) [UnsafeWrapper] Retrieved long value 0 from address 140173309347208
10:53:30,940 TRACE (testng-Test:[]) [UnsafeWrapper] Copying memory of object null offset by 140173309347216 to [B@4678c640 offset by 16 with a total of 17 bytes
10:53:30,941 TRACE (testng-Test:[]) [UnsafeWrapper] Copying memory of object null offset by 140173309347233 to [B@2964bb66 offset by 16 with a total of 11 bytes
10:53:30,941 TRACE (testng-Test:[]) [UnsafeWrapper] Copying memory of object null offset by 140173309347244 to [B@3a91006e offset by 16 with a total of 0 bytes
10:53:30,942 TRACE (testng-Test:[]) [UnsafeWrapper] Copying memory of object null offset by 140173309347244 to [B@443f1d2b offset by 16 with a total of 13 bytes
10:53:30,942 TRACE (testng-Test:[]) [UnsafeWrapper] Retrieved long value 140173306430272 from address 140173309347200
10:53:30,946 TRACE (testng-Test:[]) [BoundedOffHeapDataContainer] Removing first LRU node at 140173306430272
```

danberindei:
There are some HotRod server failures, but I'm going to assume those are unrelated.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -209,13 +209,47 @@ private void ensureSize() {
...
+         // If this was 0 that means we couldn't acquire the write lock with a try so we have to do a regular lock
+         if (addressToRemove == 0) {
+            // Acquire the write lock first and then double check the size before removing
+            entryWriteLock.lock();
+            try {
+               // Now we have to verify the size was greater than max again
+               lruLock.lock();
+               try {
+                  if (currentSize > maxSize) {
+                     // Retrieve the hashCode and make sure it has the same write lock owner
```
danberindei: Which hash code? And does the code below really check the owner?

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -209,13 +209,47 @@ private void ensureSize() {
...
+                     if (innerLock == entryWriteLock) {
+                        // If they are the same we can remove the entry
+                        addressToRemove = UNSAFE.getLong(firstAddress);
+                     } else {
+                        // If they weren't the same it means we have a different head now and we have to loop back around
+                        addressToRemove = 0;
+                     }
+                  } else {
+                     // This and above are the only ways to break out of entire loop
+                     break;
```
danberindei: Again, `if (currentSize <= maxSize) break;` looks better to me.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
wburns: It is the hash code of the head element in the lru stack. We use that to find which lock to acquire.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
wburns: Sure

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -209,13 +209,47 @@ private void ensureSize() {
                   addressToRemove = UNSAFE.getLong(firstAddress);
                }
             } else {
-               // This is the only way to break out of loop
+               // This and below are the only ways to break out of entire loop
```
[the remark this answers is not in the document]
wburns: Sure

I can write how this works in a big paragraph at the beginning.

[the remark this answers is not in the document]
wburns: Sure

I can write how this works in a big paragraph at the beginning.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
wburns: It is the hash code of the head element in the lru stack. We use that to find which lock to acquire.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
wburns: Sure

wburns:
Updated

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -190,34 +191,75 @@ protected void performClear() {
...
+    * <ul>
+    *    <li>First we unlock the lru node lock</li>
+    *    <li>Then we wait on the key write lock until it can be acquired</li>
+    *    <li>Once the write lock is acquired we reacquire the lru node lock (this order is required to prevent deadlock)</li>
+    *    <li>We check the size again just to make sure we should attempt to remove an entry</li>
+    *    <li>Lastly we check the hashCode of the head pointer again make sure it shares the same lock
+    *        (we don't care if the entry changed as long as it shares the same lock)</li>
+    *    <li>If lock is the same we unlock lru node and then finally remove the entry and release write lock</li>
+    * </ul>
+    * This entire process is repeated again until the current size is smaller or equal to the maximum size.
```
danberindei: I was thinking of something more high-level, like this:

    * This method repeatedly removes the head of the LRU list until there the current size is less than or equal to `maxSize`.
    *
    * We need to hold the LRU lock in order to check the current size and to read the head entry,
    * and then we need to hold the head entry's write lock in order to remove it.
    * The problem is that the correct acquisition order is entry write lock first, LRU lock second,
    * and we need to hold the LRU lock so that we know which entry write lock to acquire.
    *
    * To work around it, we first try to acquire the entry write lock without blocking.
    * If that fails, we release the LRU lock and we acquire the locks in the correct order, hoping that
    * the LRU head doesn't change while we wait. Because the entry write locks are striped, we actually
    * tolerate a LRU head change as long as the new head entry is in the same lock stripe.
    * If the LRU list head changes, we release both locks and try again.

Writing this got me thinking that we could improve the algorithm a little: if the current head entry has a different lock stripe at the end, we can skip the release and re-acquisition of the LRU lock for the next iteration.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -16,6 +16,7 @@
  * <p>
  * The link list consists of 28 bytes (3 longs and 1 int).  The first long is the actual entry address, the second is the
  * previous pointer, the third is the next pointer and lastly the int is the hashCode of the key to retrieve the lock.
+ * The hashCode is required    to know which lock to use when trying to read the entry.
```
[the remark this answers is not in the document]
wburns: Updating docs.

Questions answered:
1. When an entry is removed or accessed and it is not the head or tail we have to update the before and after pointers. Without the previous node address, we would have to traverse the entire linked list to find the before.
2. This should be using something like malloc behind the scenes, but I don't know what various systems would allocate it as. At worse case I would hope the 28 bytes were allocated in a 32 byte spacing.
3. Do you mean instead of allocating memory for the entry and lru independently that we allocate 1 section of memory that holds both? I had originally done this. Unfortunately this cannot be done. The reason is that I can't read an entry unless I have at least the read lock stripe for that entry (or modify unless I have write lock). There is no reasonable way to store the hashCode for every single element in the lru list, since I would need them all to write lock before and after pointers to update all references properly. I hope this is enough to try to understand what I mean, but we can talk more if needed.
4. The size and `OffHeapDataContainer` is how many elements there are. The one in `BoundedOffHeapDataContainer` is either how many elements there are or how many bytes it is taking up. The latter is done in a way where it doesn't care which as it is handled by the `sizeCalculator` to determine which is used.

wburns:
Updated docs. I am still looking at the last option of not releasing the lruLock but I am not too optimistic at this point.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
danberindei: Yes, that's what I meant. And you're right, tracking whether `lruLock` is acquired adds quite a bit of complexity, so let's leave it out.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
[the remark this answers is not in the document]
danberindei: 1. Ok
2. I'd say 32 bytes is the minimum, malloc needs to store the allocation length inside the block and maybe some extra pointers (traditionally it uses linked lists to manage the free blocks, and the list pointer(s) should be in the blocks themselves).
3. Yeah, I'd like to talk more about this. I was thinking that the hash code should only change when the entry is released and replaced with something else, so it should be possible to use double-checked locking for the hash code. Access to the LRU pointers would still be guarded by `lruLock` instead of the entry locks.
4. I was thinking you could avoid the size update in OffHeapDataContainer for bounded containers, but doing it only for one `sizeCalculator` is probably too much.

danberindei:
Integrated, thanks Will!

A final question: do we have any tests for off-heap with a store enabled?

vjuranek:
> do we have any tests for off-heap with a store enabled?

not right now, but I'm working on extending off-heap test suite, tests with off heap and single file store are already included ([here](https://github.com/vjuranek/infinispan/commit/d44e61d22a67e0fa5dbdbb4072b95affef6a9f16)) and passing

danberindei:
@vjuranek that's good news, but I forgot to mention I'm mostly interested in scenarios with passivation enabled :)

I was asking because I don't see a call to `PassivationManager.passivate()` in/around `ensureSize()`, only in evict(). And if we do call `PassivationManager.passivate()` while holding the entry write lock, we probably need more than 2*cores stripes.

wburns:
@danberindei Good point, we should have some, I will create a new JIRA for that.

PR #5421, opened 2017-09-11
Overloaded collect() to take SerializableSupplier

galderz:
This is just a quick preview to see if @wburns likes this idea :)

Whenever a collect() is called in a distributed environment, you have to wrap up in a serializable supplier method call to make it serializable, e.g.

    Map<Integer, Long> totalPerHour = cache.values().stream()
       .collect(
           CacheCollectors.serializableCollector(() -> Collectors.groupingBy(
               e -> getHourOfDay(e.departureTs),
               Collectors.counting()
           )));

This is a bit clunky and on Thursday I was wondering if we could overload `collect` to take a `SerializableSupplier` in `CacheStream`.

At first glance, it seems to work. I had to use a slightly different CollectionSupplier class to the one used by CacheCollectors.serializableCollector to get the types to work.

One slightly odd thing about this is IntelliJ, which seems to indicate there are some compilation errors in BaseStreamTest, but when I run it, it's just runs fine.

@wburns WDYT? Have you considered doing this before?

### DIFF
--- a/core/src/main/java/org/infinispan/CacheStream.java
+++ b/core/src/main/java/org/infinispan/CacheStream.java
@@ -318,6 +318,8 @@ public interface CacheStream<R> extends Stream<R>, BaseCacheStream<R, Stream<R>>
    @Override
    <R1, A> R1 collect(Collector<? super R, A, R1> collector);
 
+   <R1, A> R1 collect(SerializableSupplier<Collector<? super R, A, R1>> supplier);
+
    /**
     * Same as {@link CacheStream#collect(Supplier, BiConsumer, BiConsumer)} except that the various arguments must
     * also implement <code>Serializable</code>
--- a/core/src/main/java/org/infinispan/util/AbstractDelegatingCacheStream.java
+++ b/core/src/main/java/org/infinispan/util/AbstractDelegatingCacheStream.java
@@ -1,5 +1,8 @@
 package org.infinispan.util;
 
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.Optional;
@@ -28,6 +31,9 @@ import org.infinispan.CacheStream;
 import org.infinispan.DoubleCacheStream;
 import org.infinispan.IntCacheStream;
 import org.infinispan.LongCacheStream;
+import org.infinispan.commons.marshall.Externalizer;
+import org.infinispan.commons.marshall.SerializeWith;
+import org.infinispan.stream.CacheCollectors;
 import org.infinispan.util.function.SerializableBiConsumer;
 import org.infinispan.util.function.SerializableBiFunction;
 import org.infinispan.util.function.SerializableBinaryOperator;
@@ -318,6 +324,11 @@ public class AbstractDelegatingCacheStream<R> implements CacheStream<R> {
    }
 
    @Override
+   public <R1, A> R1 collect(SerializableSupplier<Collector<? super R, A, R1>> supplier) {
+      return collect(new CollectorSupplier2<>(supplier));
+   }
+
+   @Override
    public Optional<R> min(Comparator<? super R> comparator) {
       return castStream(underlyingStream).min(comparator);
    }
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/infinispan/CacheStream.java
```diff
@@ -318,6 +318,8 @@
    @Override
    <R1, A> R1 collect(Collector<? super R, A, R1> collector);
 
+   <R1, A> R1 collect(SerializableSupplier<Collector<? super R, A, R1>> supplier);
```
wburns: Some Javadoc here would be great.

core/src/main/java/org/infinispan/util/AbstractDelegatingCacheStream.java
```diff
@@ -318,6 +324,11 @@ public void close() {
    }
 
    @Override
+   public <R1, A> R1 collect(SerializableSupplier<Collector<? super R, A, R1>> supplier) {
+      return collect(new CollectorSupplier2<>(supplier));
```
wburns: Why not just call `return collect(CacheCollectors.serializableCollector(supplier));` ? Then we won't need the additional inner class.

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
```diff
@@ -219,7 +218,7 @@ public void testObjSortedCollector() {
 
       List<Map.Entry<Integer, String>> list = createStream(entrySet).sorted(
             (e1, e2) -> Integer.compare(e1.getKey(), e2.getKey())).collect(
-            CacheCollectors.serializableCollector(Collectors::toList));
+            () -> Collectors.toList());
```
wburns: `Collectors::toList` should still work right?

core/src/main/java/org/infinispan/CacheStream.java
wburns: Also do we want the non Serializable version override? In case if user wants to provide different serialization? This means just making this method default and calling the other.

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
galderz: It doesn't seem to:

    Error:(220, 75) java: incompatible types: cannot infer type-variable(s) R1,A,capture#1 of ?,T
    (argument mismatch; bad return type in method reference
      java.util.stream.Collector<T,capture#1 of ?,java.util.List<T>> cannot be converted to 
      java.util.stream.Collector<? super java.util.Map.Entry<java.lang.Integer,java.lang.String>,A,R1>)

It does work with the fix above... WDYT?

core/src/main/java/org/infinispan/util/AbstractDelegatingCacheStream.java
galderz: Doesn't work because the types don't match up:

    Error:(328, 37) java: method serializableCollector in class org.infinispan.stream.CacheCollectors cannot be applied to given types;
    required: org.infinispan.util.function.SerializableSupplier<java.util.stream.Collector<T,?,R>>
    found: org.infinispan.util.function.SerializableSupplier<java.util.stream.Collector<? super R,A,R1>>
    reason: cannot infer type-variable(s) T,R
    (argument mismatch;
    org.infinispan.util.function.SerializableSupplier<java.util.stream.Collector<? super R,A,R1>> 
    cannot be converted to
    org.infinispan.util.function.SerializableSupplier<java.util.stream.Collector<T,?,R>>)

To avoid causing much havoc to start with, I created `CollectorSupplier2` that's fully typed. CollectorSupplier is not fully typed on typed returned by the Supplier. I do agree both should use same collector supplier eventually... I'll play with it.

core/src/main/java/org/infinispan/util/AbstractDelegatingCacheStream.java
```diff
@@ -441,4 +452,66 @@ public DoubleCacheStream flatMapToDouble(Function<? super R, ? extends DoubleStr
    public DoubleCacheStream flatMapToDouble(SerializableFunction<? super R, ? extends DoubleStream> mapper) {
       return flatMapToDouble((Function<? super R, ? extends DoubleStream>) mapper);
    }
+
+//   public static <T, R> Collector<T, ?, R> serializableCollector(SerializableSupplier<Collector<T, ?, R>> supplier) {
```
rvansa: needs cleanup

core/src/main/java/org/infinispan/util/AbstractDelegatingCacheStream.java
```diff
@@ -441,4 +452,66 @@ public DoubleCacheStream flatMapToDouble(Function<? super R, ? extends DoubleStr
...
+      public Function<A, R> finisher() {
+         return getCollector().finisher();
+      }
+
+      @Override
+      public Set<Characteristics> characteristics() {
+         return getCollector().characteristics();
+      }
+
+      public static final class CollectorSupplierExternalizer
```
rvansa: I'd call this just `Externalizer`, the full name is determined by outer class.

core/src/main/java/org/infinispan/CacheStream.java
rvansa: Why do you need to specify the accumulator type `A`? If you replace this with `?` the `Collectors::toList` in Will's comment below works.

galderz:
> Why do you need to specify the accumulator type A? If you replace this with ? the Collectors::toList in Will's comment below works.

Why would you not? The standard method is defined as:

    <R1, A> R1 collect(Collector<? super R, A, R1> collector);

I don't see why the one that takes the Supplier should be any different.

core/src/main/java/org/infinispan/CacheStream.java
galderz: Replied below, for some reason the UI was not letting me reply here directly.

wburns:
To be honest the `A` type is really only there for the `Collector` implementation from what I have seen. The `A` type is not actually used by the `Stream` as it doesn't really care about this. It only cares about the input and output type. If you notice every single `Collectors` method returns a `Collector` with `?` as the middle type [1]. I would just make it `?`.

Also Java sucks when it deals with wildcards in nested generics, so I would suggest removing the `? super R` and replace it with just `R`. If we do this and make the definition `<R1> R1 collect(SerializableSupplier<Collector<R, ?, R1>> supplier);` then all of the above comments will just work as is.

[1] https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html

rvansa:
@galderz Because you should accept the most generic variant you can use, that's a common rule when designing API. And you don't care what type of accumulator the collector uses.
Even the `Collector` javadocs state that the acumulator type is _often hidden as an implementation detail_. And `Collectors.toList` does not expose that detail.

Regrettably the inner generic cannot bound wildcard `?` to the generic type (I think that it's related to [this problem](https://stackoverflow.com/questions/6293871/java-generics-nested-collection-of-wildcard), but if you allow it here, the `Collectors::toList` can happily work. The direct `collect` method does not have this limitation, and that's why it is free to capture the accumulator type.

```java
class A<T> {}

void test() {
   Supplier<A<?>> bar = () -> new A<Object>();
   foo(bar); // <- does not compile :-(
}

<X> void foo(Supplier<A<X>> a) {}
```

rvansa:
@wburns What's the problem with `super R`? I wouldn't limit it unless necessary.

wburns:
I agree @rvansa. The problem is just that Java doesn't really like wildcards in nested types when you are applying that as a different generic. We can workaround it with an ugly cast in the impl, which I am fine with if you wanted to do that way.

galderz:
Thx @wburns and @rvansa for the details. I'll look into them :)

galderz:
Closing PR to work on the improvements and missing bits. I'll re-open when ready.

galderz:
Hmm, can't reopen cos it thinks it was force-pushed or recreated :\

galderz:
Followed up on #5440

rvansa:
@galderz FYI there's a cure for that: log your updated branch commit id, checkout commit id of the closed PR (618b75fce57f454a2828a41869c931d9a88cd7c1 here), force push that into your branch, reopen the PR, and force push the updated commit id into the branch.

PR #5427, opened 2017-09-12
ISPN-8280 RemoteStoreRawValuesTest.testStopStartDoesNotNukeValues fails

wburns:
randomly

* Added null check in case if expired entry is removed at same time

https://issues.jboss.org/browse/ISPN-8280


danberindei:
@wburns Who could be removing the entry in this test?

wburns:
I can't confirm, but my guess is maybe the expiration reaper? Either way it is entirely possible for any user to remove an entry between when the get sees it is expired and it tries to invoked compute (lock the container for the given key).

tristantarrant:
Merged, thanks

PR #5432, opened 2017-09-13
Remove internal use of deprecated configurations

ryanemerson:
https://issues.jboss.org/browse/ISPN-8300
https://issues.jboss.org/browse/ISPN-8294
https://issues.jboss.org/browse/ISPN-8308
https://issues.jboss.org/browse/ISPN-8313

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/parsing/Parser.java
+++ b/core/src/main/java/org/infinispan/configuration/parsing/Parser.java
@@ -61,8 +61,6 @@ import org.infinispan.configuration.global.ThreadPoolConfigurationBuilder;
 import org.infinispan.configuration.global.TransportConfigurationBuilder;
 import org.infinispan.conflict.EntryMergePolicy;
 import org.infinispan.conflict.MergePolicies;
-import org.infinispan.eviction.EvictionStrategy;
-import org.infinispan.eviction.EvictionThreadPolicy;
 import org.infinispan.eviction.EvictionType;
 import org.infinispan.factories.threads.DefaultThreadFactory;
 import org.infinispan.partitionhandling.PartitionHandling;
@@ -1668,10 +1666,10 @@ public class Parser implements ConfigurationParser {
    }
 
    private void parseStoreAsBinary(final XMLExtendedStreamReader reader, final ConfigurationBuilderHolder holder) throws XMLStreamException {
+      log.elementDeprecatedUseOther(Element.STORE_AS_BINARY, Element.MEMORY);
       ConfigurationBuilder builder = holder.getCurrentConfigurationBuilder();
       Boolean binaryKeys = null;
       Boolean binaryValues = null;
-      builder.storeAsBinary().enable();
       builder.memory().storageType(StorageType.BINARY);
       for (int i = 0; i < reader.getAttributeCount(); i++) {
          ParseUtils.requireNoNamespaceAttribute(reader, i);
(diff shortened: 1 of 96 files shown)


core/src/test/java/org/infinispan/configuration/parsing/UnifiedXmlFileParsingTest.java
```diff
@@ -523,11 +523,6 @@ private static void configurationCheck70(EmbeddedCacheManager cm) {
       assertTrue(c.dataContainer().dataContainer() instanceof QueryableDataContainer);
       assertTrue(c.dataContainer().<byte[]>keyEquivalence() instanceof AnyEquivalence);
       assertTrue(c.dataContainer().<byte[]>valueEquivalence() instanceof AnyEquivalence);
-
-      c = cm.getCacheConfiguration("store-as-binary");
```
wburns: We should verify that BINARY storage type has been enabled instead of just removing.

core/src/test/resources/configs/unified/9.0.xml
```diff
@@ -198,14 +198,16 @@
             <property name="foo">bar</property>
          </data-container>
       </local-cache>
-      <local-cache name="store-as-binary">
```
wburns: This would should be converted to use `<memory> <binary/> <memory>` so it can pass the 70 test I mention above.

core/src/test/resources/configs/unified/9.1.xml
```diff
@@ -198,14 +198,16 @@
             <property name="foo">bar</property>
          </data-container>
       </local-cache>
-      <local-cache name="store-as-binary">
```
wburns: Same as 9.0.xml above.

danberindei:
@wburns I'd also like to keep the deprecated attributes/elements in the input XMLs for now. 

@ryanemerson I guess we need a separate cache for each `memory` sub-element.

ryanemerson:
@wburns I've updated the Serializer to no longer serialize deprecated configurations where possible (data-container is still present) and updated the parser etc. Let me know what you think. I'm going to add an additional commit(s) which removes all erroneous uses of the EvictionConfiguration and StoreAsBinaryConfiguration

core/src/main/java/org/infinispan/configuration/parsing/Parser.java
```diff
@@ -1870,33 +1865,22 @@ protected void parseTransaction(XMLExtendedStreamReader reader, ConfigurationBui
...
-               builder.eviction().strategy(EvictionStrategy.valueOf(value));
+            case STRATEGY:
+            case THREAD_POLICY:
+            case TYPE:
+               log.ignoreXmlElement(attribute);
                break;
-            }
-            case MAX_ENTRIES: {
+            case MAX_ENTRIES:
                log.evictionMaxEntriesDeprecated();
```
wburns: We can remove this now too since all of eviction is deprecated.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -263,7 +261,7 @@ public void preload() {
       long start = timeService.time();
 
 
-      final long maxEntries = getMaxEntries();
+      final long maxEntries = configuration.memory().size();
```
wburns: Hrmm this defaults to -1, should we not set this to `Long.MAX_VALUE` or ignore if it is -1?

hibernate-cache/src/main/java/org/infinispan/hibernate/cache/TypeOverrides.java
```diff
@@ -60,6 +60,7 @@ public EvictionStrategy getEvictionStrategy() {
     *
     * @param evictionStrategy String defining eviction strategy allowed.
     *                         Possible values are defined in {@link EvictionStrategy}
+	 * @deprecated
```
wburns: Nitpick: whitespace

core/src/main/java/org/infinispan/configuration/parsing/Parser.java
ryanemerson: @wburns Hmm, this has got me thinking. Can we make these changes to the parser for 9.1.x, or do we need to hold off this for 9.2.x?

tristantarrant:
Merged, thanks

PR #5434, opened 2017-09-13
Hibernate Search 5.8.0.Final

Sanne:
https://issues.jboss.org/browse/ISPN-8302

N.B. The modules released by Hibernate Search are now aiming for WildFly 11 as primary targed. I see no reason for them not to work on WildFly 10 though.
Hibernate ORM also released modules compatible with WildFly 11; in that case those modules are not going to work on WildFly 10, so I'm intentionally not following up with an update of the Hibernate ORM component until Infinispan updates its target WildFly version as well.


tristantarrant:
Merged, thanks.
We are not  updating to WF 11 until it goes final.

PR #5435, opened 2017-09-14
ISPN-8297 Create tests for partition handling strategies in server mode

jmarkos:
http://issues.jboss.org/browse/ISPN-8297


server/integration/testsuite/src/test/java/org/infinispan/server/test/partitionhandling/PartitionHandlingIT.java
```diff
@@ -0,0 +1,252 @@
...
+      int segmentOwnedByNode = -1;
+      for (Map.Entry<SocketAddress, Set<Integer>> socketAddressToSegments : cache.getCacheTopologyInfo().getSegmentsPerServer().entrySet()) {
+         if (socketAddressToSegments.getKey().toString().equals(serverAddress)) {
+            segmentOwnedByNode = socketAddressToSegments.getValue().stream().findFirst().get();
+         }
+      }
+
+      int i = 0;
+      while (hash.getSegment(marshall(rcm, "key" + i)) != segmentOwnedByNode) {
+         i++;
```
jmarkos: This can actually enter an infinite loop, trying to figure out why.

server/integration/testsuite/src/test/java/org/infinispan/server/test/partitionhandling/PartitionHandlingIT.java
```diff
@@ -0,0 +1,252 @@
...
+      denyReadWritesCache3.put(server3OwnedKey, "value");
+      assertEquals("value", denyReadWritesCache3.get(server3OwnedKey));
+      assertThrowsHotRodClientException(() -> denyReadWritesCache3.put(server1OwnedKey, "value"));
+      assertThrowsHotRodClientException(() -> denyReadWritesCache3.get(server1OwnedKey));
+
+      healCluster();
+   }
+
+   @Test
+   public void testAllowReads() throws InterruptedException {
```
ryanemerson: Other than the cache name, the content of `testAllowReads` and `testDenyReadWrites` is the same. We should just create one method with this logic that takes the cache name as a param and simply call this in the test methods with the appropriate name.

server/integration/testsuite/src/test/java/org/infinispan/server/test/partitionhandling/PartitionHandlingIT.java
```diff
@@ -0,0 +1,252 @@
...
+      } catch (HotRodClientException hce){
+         // OK
+      }
+   }
+
+   private String getKeyOwnedByNode(String serverAddress, RemoteCacheManager rcm, RemoteCache cache, int owners) {
+      SegmentConsistentHash hash = createHash(cache, NUM_SEGMENTS, owners);
+
+      int segmentOwnedByNode = -1;
+      for (Map.Entry<SocketAddress, Set<Integer>> socketAddressToSegments : cache.getCacheTopologyInfo().getSegmentsPerServer().entrySet()) {
```
ryanemerson: Can we not call `hash.getSegmentsByServer()` here?

server/integration/testsuite/src/test/java/org/infinispan/server/test/partitionhandling/PartitionHandlingIT.java
```diff
@@ -0,0 +1,252 @@
...
+   @InfinispanResource(CONTAINER1)
+   RemoteInfinispanServer server1;
+   String server1Address;
+
+   @InfinispanResource(CONTAINER2)
+   RemoteInfinispanServer server2;
+
+   @InfinispanResource(CONTAINER3)
+   RemoteInfinispanServer server3;
+   String server3Address;
```
ryanemerson: Is there any reason for these variables not to be private?

server/integration/testsuite/src/test/java/org/jgroups/protocols/PARTITION_HANDLING.java
```diff
@@ -0,0 +1,155 @@
...
+ * @author Radim Vansa (rvansa@redhat.com)
+ */
+@MBean(description = "Enables custom partition handling.")
+public class PARTITION_HANDLING extends Protocol {
+
+   protected static final short PROTOCOL_ID = (short) 0x51A7;
+   protected static Log log = LogFactory.getLog(PARTITION_HANDLING.class);
+
+   protected int nodeIndex = -1;
+   protected Set<Integer> allowedNodes;
```
ryanemerson: Vars can be private

jmarkos:
@ryanemerson addressed all comments and fixed the loop (ran the tests 40 times locally without any issues). Not 100% happy with the solution, but I cannot access the rebalance status via JMX during an arquillian test for whatever reason (works fine outside arquillian so cannot create an issue for it). Might look into it more later...

server/integration/testsuite/src/test/java/org/infinispan/server/test/partitionhandling/PartitionHandlingIT.java
danberindei: I think you should check that `segmentOwnedByNode != -1` before entering the loop. Optionally you could allow the key to map to any segment owned by `serverAddress`, not just the first one.

server/integration/testsuite/src/test/java/org/infinispan/server/test/partitionhandling/PartitionHandlingIT.java
```diff
@@ -0,0 +1,257 @@
...
+            int emptyIndex = 0;
+            while (mySegmentOwners[integer][emptyIndex] != null) {
+               emptyIndex++;
+            }
+            mySegmentOwners[integer][emptyIndex] = server;
+         }
+      }
+      SegmentConsistentHash result = new SegmentConsistentHash();
+      result.init(mySegmentOwners, segments);
+      return result;
```
danberindei: Kind of annoying that you need all of this just to call `getSegment(key)`... wouldn't something like this be better? 

`((TcpTransportFactory) ((InternalRemoteCacheManager) remoteCacheManager).getTransportFactory()).getConsistentHash(RemoteCacheManager.cacheNameBytes(cache.getName()))` instead

server/integration/testsuite/src/test/java/org/jgroups/protocols/PARTITION_HANDLING.java
```diff
@@ -0,0 +1,155 @@
...
+ * This is custom protocol used for partition handling. What it basically does is, that it discards specific
+ * JGroups messages from some allowed nodes. Do not use this class directly (via JGroups Probe), use {@link PartitionHandlingController} instead.
+ * <p>
+ * The class file is inserted as a JAR into the test servers (see pom.xml).
+ *
+ * @author Jiri Holusa (jholusa@redhat.com)
+ * @author Radim Vansa (rvansa@redhat.com)
+ */
+@MBean(description = "Enables custom partition handling.")
+public class PARTITION_HANDLING extends Protocol {
```
danberindei: I wonder if you could do the same thing using `DISCARD.addIgnoredMember()`/`resetIgnoredMembers()`. I have no idea if they work over JMX/probe, but I'd like to avoid the extra jar if possible.

rhusar: I guess I have the same question: could this be achieved by DISCARD and if not, why not? If this is generally useful it should ideally reside in JGroups repo which solves the "extra jar" problem, but mainly this way it can be reused in other projects like WildFly.

jmarkos: Looking at DISCARD it seems it could be used, yes. I'll try to use it and see if there are any issues.

jmarkos: So I found the problem - with probe, you can only call methods that have primitive types as arguments, and the DISCARD.addIgnoreMember method expects an argument of type Address. So DISCARD could be used, but I'd need to make changes to it and we'd need to wait for a JGroups release with them.

jmarkos: Or we could use DISCARD only on the node that is alone in a partition, but that would limit the possible splits on clusters of size 4 and more (you wouldn't be able to do 2 partitions of 2 nodes). Since we only test with max 3 servers this would work.

rhusar: > with probe, you can only call methods that have primitive types as arguments

Nice find; JGroups Jira + PR would be nice.

>  Since we only test with max 3 servers this would work.

Looks like this could work to. Would you mind giving it a try? I would be curious to see how fast you can get the partitions to merge then.

jmarkos:
Updated. Using DISCARD now with the limitation to 3 servers, which is ok for our testsuite.

tristantarrant:
CI reports related failures http://ci.infinispan.org/job/Infinispan/job/PR-5435/3/

jmarkos:
Added a small fix (so that a failure in 1 test doesn't cause failures in other tests), but that still doesn't fix the main problem which is that the cluster didn't split properly. I've tried reproducing this on a CI machine but didn't get a failure in over 80 runs, so I'm a bit stuck.

danberindei:
@jmarkos @tristantarrant @mgencur we discussed reducing the number of clusters we start in the integration tests, but this test adds another minute to the test suite (not just starting the servers, but also waiting for the partition/merge), so I can't help wonder if we really need it.

rhusar:
@danberindei Maybe this helps, only regarding speeding up split/merge: I have explored this in WF testsuite since I am having the same problem with high execution time, by installing views manually (similar to what merge tests in jgroups testsuite do). When this works, splits and merge happen instantly: https://github.com/rhusar/wildfly/commit/735664a2be9cce732a1e236b33713f62496cbd34#diff-25841019034dfe0d9f3a4f3fc5a3a6b9R87 ...unfortunately, for some reason jgroups sometimes drops the message and the test needs to wait for merge logic to kick in.

ryanemerson:
@danberindei Good point. I think it's worth us adding additional split-brain scenarios to the core tests instead though, as we currently only test a 4 node cluster {A,B,C,D} that splits {A,B}, {C,D}

danberindei:
@ryanemerson I agree we could have tests with 3 nodes in core, and in core it makes sense to install views manually instead of waiting for `FD` and `MERGE3`. But I should mention we already have different split scenarios in `ThreeWaySplitAndMergeTest` and `XPartitionAndMergeDuringYTest`.

ryanemerson:
@danberindei Sorry I should have been more specific, I meant that currently we only have tests for conflict resolution in the aforementioned split brain scenario.

mgencur:
Right, it is bad that these new tests add another minute to the test suite. We need to change the strategy about which tests will go to the server test suite. It is a bit unfortunate to find this out only when this PR has been finalized (after several rounds) but it probably makes more sense to add more tests to the core test suite and prevent growing the server test suite execution time.

jmarkos:
I could imagine leaving at least the test for the custom entry merge policy (from https://github.com/infinispan/infinispan/pull/5545), as that's something that's most probable to break in the future. Would be useful even without the cluster split, just starting the server with the configured custom policy.

mgencur:
You're right @jmarkos , the custom entry merge policy requires a specific deployment to the server. This needs to be in the server test suite, IMO.

ryanemerson:
Closing this PR. I will create a test for the server to ensure that custom policies are deployed as expected as part of a future PR that will contain other enhancements for partition handling

PR #5437, opened 2017-09-15
ISPN-8195 Transaction fails to commit when a node crashes

danberindei:
https://issues.jboss.org/browse/ISPN-8195
https://issues.jboss.org/browse/ISPN-8237

ISPN-8195 Transaction fails to commit when a node crashes
ISPN-8237 State transfer doesn't replace L1 entry with regular entry

Should fix (rare) random failures in ConcurrentNonOverlappingLeaveTest.


danberindei:
Updated.

pruivo:

PR #5438, opened 2017-09-15
ISPN-8111 OffHeapBoundedSingleNodeTest.testMultiThreaded fails with

wburns:
trace logging enabled

* Disabled transactions which should let it run faster

https://issues.jboss.org/browse/ISPN-8111


danberindei:
Please change `OffHeapSingleNodeTest` and `OffHeapMultiNodeTest` as well.

Speaking of which, does the off-heap implementation work differently in single-node vs multi-node?

wburns:
No, should be the same. Just wanted to test different scenarios is all. Obviously missed stores :(

wburns:
Updated to disable transactions in other tests as well.

danberindei:
Integrated, thanks Will!

PR #5440, opened 2017-09-15
ISPN-8281 Overloaded collect() to take SerializableSupplier

galderz:
https://issues.jboss.org/browse/ISPN-8281

Follows up on #5421 to fix some of the issues found.

One thing that I still can't to work is the `Collectors::toList` thing point out by @wburns [here](https://github.com/infinispan/infinispan/pull/5421#discussion_r138057754). I get error:

```
Error:(220, 75) java: incompatible types: cannot infer type-variable(s) R1,capture#1 of ?,T
    (argument mismatch; bad return type in method reference
      java.util.stream.Collector<T,capture#1 of ?,java.util.List<T>> cannot be converted to java.util.stream.Collector<? super java.util.Map.Entry<java.lang.Integer,java.lang.String>,?,R1>)
```

Any ideas?

### DIFF
--- a/core/src/main/java/org/infinispan/CacheStream.java
+++ b/core/src/main/java/org/infinispan/CacheStream.java
@@ -1,5 +1,6 @@
 package org.infinispan;
 
+import java.io.Serializable;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.Optional;
@@ -308,7 +309,8 @@ public interface CacheStream<R> extends Stream<R>, BaseCacheStream<R, Stream<R>>
     * prevents the usage of {@link java.util.stream.Collectors} class.  However you can use the
     * {@link org.infinispan.stream.CacheCollectors} static factory methods to create a serializable wrapper, which then
     * creates the actual collector lazily after being deserialized.  This is useful to use any method from the
-    * {@link java.util.stream.Collectors} class as you would normally.</p>
+    * {@link java.util.stream.Collectors} class as you would normally.
+    * Alternatively, you can call {@link #collect(SerializableSupplier)} too.</p>
     * @param collector
     * @param <R1> collected type
     * @param <A> intermediate collected type if applicable
@@ -318,7 +320,58 @@ public interface CacheStream<R> extends Stream<R>, BaseCacheStream<R, Stream<R>>
    @Override
    <R1, A> R1 collect(Collector<? super R, A, R1> collector);
 
-   <R1, A> R1 collect(SerializableSupplier<Collector<? super R, A, R1>> supplier);
+   /**
+    * Performs a <a href="package-summary.html#MutableReduction">mutable
+    * reduction</a> operation on the elements of this stream using a
+    * {@code Collector} that is lazily created from the {@code SerializableSupplier}
+    * provided.
+    *
+    * This method behaves exactly the same as {@link #collect(Collector)} with
+    * the enhanced capability of working even when the mutable reduction
+    * operation has to run in a remote node and the operation is not
+    * {@link Serializable} or otherwise marshallable.
+    *
+    * So, this method is specially designed for situations when the user
+    * wants to use a {@link Collector} instance that has been created by
+    * {@link java.util.stream.Collectors} static factory methods.
+    *
+    * In this particular case, the function that instantiates the
+    * {@link Collector} will be marshalled according to the
+    * {@link Serializable} rules.
+    *
+    * @param supplier The supplier to create the collector that is specifically serializable
+    * @param <R1> The resulting type of the collector
+    * @return the collected value
+    * @since 9.2
+    */
+   <R1> R1 collect(SerializableSupplier<Collector<? super R, ?, R1>> supplier);
+
+   /**
+    * Performs a <a href="package-summary.html#MutableReduction">mutable
+    * reduction</a> operation on the elements of this stream using a
+    * {@code Collector} that is lazily created from the {@code Supplier}
+    * provided.
+    *
+    * This method behaves exactly the same as {@link #collect(Collector)} with
+    * the enhanced capability of working even when the mutable reduction
+    * operation has to run in a remote node and the operation is not
+    * {@link Serializable} or otherwise marshallable.
+    *
+    * So, this method is specially designed for situations when the user
+    * wants to use a {@link Collector} instance that has been created by
+    * {@link java.util.stream.Collectors} static factory methods.
+    *
+    * In this particular case, the function that instantiates the
+    * {@link Collector} will be marshalled using Infinispan
+    * {@link org.infinispan.commons.marshall.Externalizer} class or one of its
+    * subtypes.
+    *
+    * @param supplier The supplier to create the collector
+    * @param <R1> The resulting type of the collector
+    * @return the collected value
+    * @since 9.2
+    */
+   <R1> R1 collect(Supplier<Collector<? super R, ?, R1>> supplier);
 
    /**
     * Same as {@link CacheStream#collect(Supplier, BiConsumer, BiConsumer)} except that the various arguments must
--- a/core/src/main/java/org/infinispan/CacheStream.java
+++ b/core/src/main/java/org/infinispan/CacheStream.java
@@ -318,6 +318,8 @@ public interface CacheStream<R> extends Stream<R>, BaseCacheStream<R, Stream<R>>
    @Override
    <R1, A> R1 collect(Collector<? super R, A, R1> collector);
 
+   <R1, A> R1 collect(SerializableSupplier<Collector<? super R, A, R1>> supplier);
+
    /**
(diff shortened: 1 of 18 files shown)


galderz:
Oh, one final thing: I'm happy to hold this off till 9.2.x has been branched off since it's new functionality. @tristantarrant thoughts?

core/src/test/java/org/infinispan/stream/BaseStreamIteratorEvictionTest.java
```diff
@@ -43,7 +44,7 @@ public void testExpiredEntryNotReturned() throws InterruptedException {
       Thread.sleep(TimeUnit.SECONDS.toMillis(expectedTime) + 50);
 
       Map<Object, String> results;
-      try (Stream<CacheEntry<Object, String>> stream = cache.getAdvancedCache().cacheEntrySet().stream()) {
+      try (CacheStream<CacheEntry<Object, String>> stream = cache.getAdvancedCache().cacheEntrySet().stream()) {
```
wburns: Why do these need to be changed to `CacheStream` ?

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
```diff
@@ -219,7 +218,7 @@ public void testObjSortedCollector() {
 
       List<Map.Entry<Integer, String>> list = createStream(entrySet).sorted(
             (e1, e2) -> Integer.compare(e1.getKey(), e2.getKey())).collect(
-            CacheCollectors.serializableCollector(Collectors::toList));
+            () -> Collectors.toList());
```
wburns: Man I wish that nested generics work a bit better.
Sad that the `serializableCollector` works but not the new way :(

wburns:
Actually we should update https://github.com/infinispan/infinispan/blob/0221635f0340dbe793ddf7e46a627793f17e01ad/documentation/src/main/asciidoc/user_guide/streams.adoc#serialization to mention this new usage.

core/src/test/java/org/infinispan/stream/BaseStreamIteratorEvictionTest.java
galderz: It's due to the implementation of `mapFromStream`. In there, I use now the overloaded `collect` method but that's only available in `CacheStream` API, hence the change. Anything you'd change?

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
galderz: Yeah :(. I'm going to try to build a small test case that's easier to share with StackOverflow or submit as bug to Oracle.

galderz:
@wburns Good catch, I'll update that.

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
rvansa: @galderz So you have decided not to use wildcard for accumulator type?

galderz: I've been able to replicate [this standalone](https://gist.github.com/galderz/a44468018e5ecd270fd7e0e28db5b499) and I've submitted a [question](https://stackoverflow.com/questions/46274846/java-compilation-fail-with-method-reference-but-works-with-lambda) to StackOverflow. Let's see if we can get an answer.

tristantarrant:
[ERROR] /home/infinispan/workspace/Infinispan_PR-5440-KKOWBIXVBM6GU3BQP5XYF2MAFZ7TZPROS4BIWL7RTKAYLWKFNRNA/core/src/test/java/org/infinispan/stream/BaseStreamIteratorEvictionTest.java:8:8: Unused import - java.util.stream.Stream. [UnusedImports]
[ERROR] /home/infinispan/workspace/Infinispan_PR-5440-KKOWBIXVBM6GU3BQP5XYF2MAFZ7TZPROS4BIWL7RTKAYLWKFNRNA/core/src/test/java/org/infinispan/stream/BaseStreamIteratorTest.java:14:8: Unused import - java.util.stream.Stream. [UnusedImports]
[ERROR] /home/infinispan/workspace/Infinispan_PR-5440-KKOWBIXVBM6GU3BQP5XYF2MAFZ7TZPROS4BIWL7RTKAYLWKFNRNA/core/src/test/java/org/infinispan/stream/DistributedStreamIteratorTxTest.java:10:8: Unused import - java.util.stream.Stream. [UnusedImports]
[ERROR] /home/infinispan/workspace/Infinispan_PR-5440-KKOWBIXVBM6GU3BQP5XYF2MAFZ7TZPROS4BIWL7RTKAYLWKFNRNA/core/src/test/java/org/infinispan/stream/stress/DistributedStreamRehashStressTest.java:30:8: Unused import - org.infinispan.stream.CacheCollectors. [UnusedImports]

galderz:
Thx @tristantarrant, I'll check that

core/src/test/java/org/infinispan/stream/BaseStreamTest.java
galderz: Fix for this line of code: https://stackoverflow.com/a/46276537/186429

galderz: I didn't know you could force types with method references. FYI of others:

    Collectors::<Map.Entry<Integer, String>>toList

core/src/main/java/org/infinispan/CacheStream.java
```diff
@@ -319,6 +321,59 @@
...
+    * In this particular case, the function that instantiates the
+    * {@link Collector} will be marshalled according to the
+    * {@link Serializable} rules.
+    *
+    * @param supplier The supplier to create the collector that is specifically serializable
+    * @param <R1> The resulting type of the collector
+    * @return the collected value
+    * @since 9.2
+    */
+   <R1> R1 collect(SerializableSupplier<Collector<? super R, ?, R1>> supplier);
```
wburns: I find it a lot easier to have this method be a default method that just passes the `supplier` to the non serializable method variant. This way all implementations only have to implement the method once. And we don't really care how it is serialized in the impl, just that it is.

core/src/test/java/org/infinispan/stream/BaseStreamIteratorEvictionTest.java
wburns: I see now, thanks.

documentation/src/main/asciidoc/user_guide/streams.adoc
```diff
@@ -118,18 +118,16 @@ This relies on the spec to pick the most specific method as defined link:https:/
...
-class doesn't produce Serializable instances.  Thus if you need to use these, you can use the newly provided
-link:{javadocroot}/org/infinispan/stream/CacheCollectors.html[CacheCollectors]
-class which allows for a `Supplier<Collector>` to be provided.  This instance could then use the
-link:{jdkdocroot}/java/util/stream/Collectors.html[Collectors]
-to supply a `Collector` which is not serialized. You can read more details about how the
-collector peforms in a distributed fashion at link:user_guide.html#distributed_stream_execution[distributed execution].
+class doesn't produce Serializable instances.
+Thus if you need to use these, you should call the overloaded `collect`methods that take `Supplier<Collector>`.
+This can be serialized enabling the collector to be lazily instantiated in remote nodes.
+You can read more details about how the collector peforms in a distributed fashion at link:user_guide.html#distributed_stream_execution[distributed execution].
```
wburns: We should probably leave in a blurb about the `CacheCollectors` still. This is useful if your is defined as a `Stream` interface which doesn't have the new collect method taking a Supplier.

galderz: Not sure what you exactly want... mind sending a PR to my branch?

documentation/src/main/asciidoc/user_guide/streams.adoc
```diff
@@ -118,18 +118,16 @@ This relies on the spec to pick the most specific method as defined link:https:/
 
 In our previous example we used a `Collector` to collect all the results into a `Map`.
 Unfortunately the link:{jdkdocroot}/java/util/stream/Collectors.html[Collectors]
-class doesn't produce Serializable instances.  Thus if you need to use these, you can use the newly provided
-link:{javadocroot}/org/infinispan/stream/CacheCollectors.html[CacheCollectors]
```
wburns: I meant more that we should keep this paragraph in here in some way. We can reduce it, but just so people know about the `CacheCollectors` in case if they don't use the `CacheStream` interface.

galderz: Hmmm... I've rebased and fixed other bits. Don't have time today to have an in depth look into this. Back on Friday, feel free to send me a PR if you have a clear idea and have time.

PR #5442, opened 2017-09-18
ISPN-7140 Concurrent modifications succeed in pessimistic cache

pruivo:
Change pessimistic locking. See the javadoc in PessimistLockingInterceptor.

It should fix random failures of InfinispanNodeFailureTest

https://issues.jboss.org/browse/ISPN-7140
https://issues.jboss.org/browse/ISPN-5076

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
@@ -1,6 +1,7 @@
 package org.infinispan.interceptors.locking;
 
 import java.util.Collection;
+import java.util.Collections;
 
 import org.infinispan.commands.CommandsFactory;
 import org.infinispan.commands.DataCommand;
@@ -8,40 +9,62 @@ import org.infinispan.commands.FlagAffectedCommand;
 import org.infinispan.commands.control.LockControlCommand;
 import org.infinispan.commands.tx.PrepareCommand;
 import org.infinispan.commands.write.DataWriteCommand;
+import org.infinispan.context.Flag;
 import org.infinispan.context.InvocationContext;
 import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.context.impl.TxInvocationContext;
 import org.infinispan.factories.annotations.Inject;
-import org.infinispan.interceptors.InvocationSuccessFunction;
-import org.infinispan.statetransfer.StateTransferManager;
-import org.infinispan.transaction.impl.LocalTransaction;
+import org.infinispan.interceptors.distribution.TxDistributionInterceptor;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
 /**
- * Locking interceptor to be used by pessimistic caches.
- * Design note: when a lock "k" needs to be acquired (e.g. cache.put("k", "v")), if the lock owner is the local node,
- * no remote call is performed to migrate locking logic to the other (numOwners - 1) lock owners. This is a good
- * optimisation for  in-vm transactions: if the local node crashes before prepare then the replicated lock information
- * would be useless as the tx is rolled back. OTOH for remote hotrod/transactions this additional RPC makes sense because
- * there's no such thing as transaction originator node, so this might become a configuration option when HotRod tx are
- * in place.
- *
- * Implementation note: current implementation acquires locks remotely first and then locally. This is required
- * by the deadlock detection logic, but might not be optimal: acquiring locks locally first might help to fail fast the
- * in the case of keys being locked.
+ * Locking interceptor to be used by pessimistic transactions.
+ * <p>
+ * Design (from 9.1):
+ * <p>
+ * This interceptor is skipped when {@link Flag#SKIP_LOCKING} is used or when all affected keys are already locked.
+ * Read-only commands can acquire the locks when {@link Flag#FORCE_WRITE_LOCK}.
+ * <p>
+ * The new design has 2 concepts. The primary lock and the backup lock. The primary lock is acquired in the primary
+ * owner and it is exclusive; only one transaction can acquire it. The backup lock is acquired by the backup owner and
+ * it is shared; multiple transactions can acquire simultaneously. The backup locks are associated with the topology id
+ * and they are only used when the primary owner leaves or crashes.
+ * <p>
+ * This algorithm contacts all owners and acquires the primary lock in the primary owner and the backup lock in the
+ * backup owners. So, lock acquisition (single or multiple keys) happens in a single RPC to all the owners of the
+ * key(s). This is triggered in {@link TxDistributionInterceptor#visitLockControlCommand(TxInvocationContext,
+ * LockControlCommand)}.
+ * <p>
+ * The {@link Flag#CACHE_MODE_LOCAL} changes this behaviour; it makes the lock acquisition only local. So, only local
+ * the primary (if primary owner) or backup lock (if backup owner) is acquired. No RPC is create and if the node isn't
+ * an owner, no locks are acquired.
+ * <p>
+ * <p>
+ * In case of primary owner changes, there are 2 cases.
+ * <p>
+ * 1) a new node joins and it is the new primary owner. In this case, the old primary owner transfers the lock
+ * information to the new primary owner.
+ * <p>
+ * 2) the primary owner crashes or leaves. In this case, a backup owner will be promoted to the primary owner. All
+ * primary lock requests are hold until its previous backup locks are released. Note that a transaction only acquires
+ * the lock once. If a transaction requests a primary lock but has the backup lock acquired, it doesn't grant the
+ * primary lock if there are other backup locks from other transactions.
+ * <p>
+ * Know Issue: There is small deadlock probably is 2 or more transaction try to acquire the lock of a key and the
+ * primary owner dies before any of them receives a reply. In that case, none of the transactions (and future
+ * transactions) are able to acquire the primary lock until all backup lock are cleared. Eventually, the transactions
+ * will rollback and the backup lock will be released. One way to solve it is to acquire the backup locks after the
+ * primary lock. However, this will increase the performance penalty for all locks by generating 2 RPCs.
  *
  * @author Mircea Markus
+ * @author Pedro Ruivo
  */
 public class PessimisticLockingInterceptor extends AbstractTxLockingInterceptor {
    private static final Log log = LogFactory.getLog(PessimisticLockingInterceptor.class);
-   public static final boolean trace = log.isTraceEnabled();
-
-   private final InvocationSuccessFunction localLockCommandWork =
-         (rCtx, rCommand, rv) -> localLockCommandWork(rCtx, (LockControlCommand) rCommand);
+   private static final boolean trace = log.isTraceEnabled();
 
    private CommandsFactory cf;
-   private StateTransferManager stateTransferManager;
 
    @Override
    protected Log getLog() {
@@ -49,78 +72,23 @@ public class PessimisticLockingInterceptor extends AbstractTxLockingInterceptor 
    }
 
    @Inject
-   public void init(CommandsFactory factory, StateTransferManager stateTransferManager) {
+   public void init(CommandsFactory factory) {
       this.cf = factory;
-      this.stateTransferManager = stateTransferManager;
    }
 
    @Override
-   protected final Object visitDataReadCommand(InvocationContext ctx, DataCommand command)
-         throws Throwable {
-      if (!readNeedsLock(ctx, command)) {
-         return invokeNext(ctx, command);
-      }
-
-      if (!readNeedsLock(ctx, command)) {
-         return invokeNext(ctx, command);
-      }
-
-      Object key = command.getKey();
-      if (!needRemoteLocks(ctx, key, command)) {
-         acquireLocalLock(ctx, command);
-         return invokeNext(ctx, command);
-      }
-
-      TxInvocationContext txContext = (TxInvocationContext) ctx;
-      LockControlCommand lcc = cf.buildLockControlCommand(key, command.getFlagsBitSet(),
-            txContext.getGlobalTransaction());
-      return invokeNextThenApply(ctx, lcc, (rCtx, rCommand, rv) -> {
-         acquireLocalLock(rCtx, command);
-         return invokeNext(rCtx, command);
-      });
-
-   }
-
-   private boolean readNeedsLock(InvocationContext ctx, FlagAffectedCommand command) {
-      return ctx.isInTxScope() && command.hasAnyFlag(FlagBitSets.FORCE_WRITE_LOCK) && !hasSkipLocking(command);
-   }
-
-   private void acquireLocalLock(InvocationContext ctx, DataCommand command) throws InterruptedException {
-      if (trace)
-         log.tracef("acquireLocalLock");
-      final TxInvocationContext txContext = (TxInvocationContext) ctx;
-      Object key = command.getKey();
-      lockOrRegisterBackupLock(txContext, key, getLockTimeoutMillis(command));
-      txContext.addAffectedKey(key);
+   protected final Object visitDataReadCommand(InvocationContext ctx, DataCommand command) throws Throwable {
+      return readNeedsLock(ctx, command) ?
+             invokeLockCommandAndContinue(Collections.singleton(command.getKey()), command, (TxInvocationContext<?>) ctx, command.getTopologyId()) :
+             invokeNext(ctx, command);
    }
 
    @Override
-   protected Object handleReadManyCommand(InvocationContext ctx, FlagAffectedCommand command, Collection<?> keys) throws Throwable {
-      Object maybeStage;
-      if (!readNeedsLock(ctx, command)) {
-         maybeStage = invokeNext(ctx, command);
-      } else {
-         if (!needRemoteLocks(ctx, keys, command)) {
-            acquireLocalLocks(ctx, command, keys);
-            maybeStage = invokeNext(ctx, command);
-         } else {
-            // Acquire the remote locks first, then the local locks
-            final TxInvocationContext txContext = (TxInvocationContext) ctx;
-            LockControlCommand lcc = cf.buildLockControlCommand(keys, command.getFlagsBitSet(),
-                  txContext.getGlobalTransaction());
-            maybeStage = invokeNextThenApply(ctx, lcc, (rCtx, rLockCommand, rv) -> {
-               acquireLocalLocks(rCtx, command, keys);
-               return invokeNext(rCtx, command);
-            });
-         }
-      }
-      return maybeStage;
-   }
-
-   private void acquireLocalLocks(InvocationContext ctx, FlagAffectedCommand command, Collection<?> keys)
-         throws InterruptedException {
-      lockAllOrRegisterBackupLock((TxInvocationContext<?>) ctx, keys, getLockTimeoutMillis(command));
-      ((TxInvocationContext<?>) ctx).addAllAffectedKeys(keys);
+   protected Object handleReadManyCommand(InvocationContext ctx, FlagAffectedCommand command, Collection<?> keys,
+         int topologyId) throws Throwable {
+      return readNeedsLock(ctx, command) ?
+             invokeLockCommandAndContinue(keys, command, (TxInvocationContext) ctx, topologyId) :
+             invokeNext(ctx, command);
    }
 
    @Override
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -1,126 +1,84 @@
...
+ * <p>
+ * The new design has 2 concepts. The primary lock and the backup lock. The primary lock is acquired in the primary
+ * owner and only on transaction can acquire at each give time. The backup lock exists in the backup owner and multiple
+ * transactions can acquire simultaneously. The backup lock is associated to the current topology id and the primary
+ * lock can only be acquired if doesn't exists any backup lock with older topology id.
+ * <p>
+ * Since the primary owner doesn't acquire backup locks, they are only used when the primary ownership of a key is
+ * changed.
+ * <p>
+ * The new algorithm will always acquire both. So, lock acquisition (single or multiple keys) will generate a RPC to all
```
rvansa: This is unclear: acquire both locks on each node, or just primary on primaries and backup on the others? Or does it just mean that a LockCommand will contact all owners? Don't try to describe "new" algorithm assuming that the "old" i known, just describe how it is.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -1,126 +1,84 @@
...
+ * the owners of the key(s). This is triggered in {@link TxDistributionInterceptor#visitLockControlCommand(TxInvocationContext,
+ * LockControlCommand)}.
+ * <p>
+ * The {@link Flag#CACHE_MODE_LOCAL} changes this behaviour. It makes the lock acquisition only local. So, only the
+ * primary or backup lock is acquired. No RPC is create.
+ * <p>
+ * Know Issue: There is small deadlock probably is 2 or more transaction try to acquire the same lock and the primary
```
rvansa: You haven't explain how the backup lock is used (just that it's acquired non-exclusively), so this note is missing a "big picture". Does the to-be-primary node automatically convert its backup lock into primary lock? What if it has the lock locked multiple times?

pruivo:
@rvansa updated. let me know if some part isn't clear or if you have more questions :)

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -190,95 +145,89 @@ public Object visitLockControlCommand(TxInvocationContext ctx, LockControlComman
       if (!ctx.isInTxScope())
          throw new IllegalStateException("Locks should only be acquired within the scope of a transaction!");
 
-      boolean skipLocking = hasSkipLocking(command);
-      if (skipLocking) {
+      if (command.isUnlock()) { //TODO double check unlock. it isn't used anymore!
+         if (ctx.isOriginLocal()) {
+            throw new AssertionError(
```
rvansa: Either use `assert ctx.isOriginLocal() : "There's ..." or `throw new IllegalStateException(...)`

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -190,95 +145,89 @@ public Object visitLockControlCommand(TxInvocationContext ctx, LockControlComman
...
-               if (trace)
-                  log.tracef("Already own locks on keys: %s, skipping remote call", command.getKeys());
-               return true;
-            }
-         } else {
-            if (trace)
-               log.tracef("Single key %s and local, skipping remote call", command.getSingleKey());
-            return localLockCommandWork(ctx, command);
+      if (hasSkipLocking(command) || areAllKeysAlreadyLocked(ctx, command.getKeys())) {
+         //lock control with skip locking?? that is weird...
```
rvansa: I would throw exception for LockCommand with skip locking

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -190,95 +145,89 @@ public Object visitLockControlCommand(TxInvocationContext ctx, LockControlComman
...
+                handleLocalLockCommand(ctx, command) :
+                handleRemoteLockCommand(ctx, command);
       }
+   }
 
-      return invokeNextThenApply(ctx, command, localLockCommandWork);
+   private boolean readNeedsLock(InvocationContext ctx, FlagAffectedCommand command) {
+      return ctx.isInTxScope() &&
+             ctx.isOriginLocal() &&
+             command.hasAnyFlag(FlagBitSets.FORCE_WRITE_LOCK) &&
```
rvansa: `command.getFlagBitSet() & (FlagBitSets.FORCE_WRITE_LOCK | FlagBitSets.SKIP_LOCKING) == FlagBitSets.FORCE_WRITE_LOCK`
Maybe we should throw exception when both flags are used together.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -190,95 +145,89 @@ public Object visitLockControlCommand(TxInvocationContext ctx, LockControlComman
...
    }
 
-   private boolean localLockCommandWork(InvocationContext ctx, LockControlCommand command)
-         throws InterruptedException {
-      TxInvocationContext<?> txInvocationContext = (TxInvocationContext<?>) ctx;
-      if (ctx.isOriginLocal()) {
-         txInvocationContext.addAllAffectedKeys(command.getKeys());
-      }
+   private boolean areAllKeysAlreadyLocked(TxInvocationContext<?> context, Collection<?> keys) {
+      return context.getAffectedKeys().containsAll(keys);
```
rvansa: If you execute put with skip locking, and then putMap with the same key, the key is already in the context but it's not locked... amiright?

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -134,54 +102,41 @@ public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand comman
...
    @Override
-   protected <K> Object handleWriteManyCommand(InvocationContext ctx, FlagAffectedCommand command, Collection<K> keys, boolean forwarded) throws Throwable {
-      Object maybeStage;
-      if (hasSkipLocking(command)) {
-         maybeStage = invokeNext(ctx, command);
+   protected <K> Object handleWriteManyCommand(InvocationContext ctx, FlagAffectedCommand command, Collection<K> keys,
+         boolean forwarded, int topologyId) throws Throwable {
+      final TxInvocationContext<?> txCtx = (TxInvocationContext<?>) ctx;
+      if (skipLockingForKeys(command, txCtx, keys)) {
+         txCtx.addAllAffectedKeys(keys);
```
rvansa: Related to https://github.com/infinispan/infinispan/pull/5442/files#r139728211 why do you add the keys here when these are not being locked?

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -134,54 +102,41 @@ public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand comman
...
    }
 
    @Override
    protected Object visitDataWriteCommand(InvocationContext ctx, DataWriteCommand command)
          throws Throwable {
-      Object maybeStage;
-      Object key = command.getKey();
-      if (hasSkipLocking(command)) {
+      if (!ctx.isInTxScope()) {
+         //we don't know how to lock keys in non-tx contexts
```
rvansa: PFER needs to be locked the same way as in non-tx mode - lock it here and unlock after `invokeNext`.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -134,54 +102,41 @@ public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand comman
...
-            // Mark the key as affected even with SKIP_LOCKING
-            ((TxInvocationContext<?>) ctx).addAffectedKey(key);
-         }
-         maybeStage = invokeNext(ctx, command);
+         return invokeNext(ctx, command);
+      }
+      final Object key = command.getKey();
+      final TxInvocationContext<?> txCtx = (TxInvocationContext<?>) ctx;
+      if (skipLockingForKey(command, txCtx, key)) {
+         txCtx.addAffectedKey(key);
```
rvansa: same as https://github.com/infinispan/infinispan/pull/5442/files#r139892231

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
rvansa: I've noticed that there is a `InvocationContext.hasLockedKey` method.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -190,95 +145,89 @@ public Object visitLockControlCommand(TxInvocationContext ctx, LockControlComman
...
+      //we need to acquire locks
+      LockControlCommand lcc = cf.buildLockControlCommand(keys, command.getFlagsBitSet(), ctx.getGlobalTransaction());
+      lcc.setTopologyId(topologyId);
+      return makeStage(handleLocalLockCommand(ctx, lcc))
+            .thenApply(ctx, command, (rCtx, rCommand, rv) -> invokeNext(rCtx, rCommand));
    }
+
+   private Object handleLocalLockCommand(TxInvocationContext context, LockControlCommand command)
+         throws InterruptedException {
+      //TODO blocking method. change it to non-blocking? [minor, only blocks local commands]
```
rvansa: Yes, this is unacceptable because you would be blocking `xxxAsync` calls!

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
pruivo: `hasLockedKey` only check if this node has the primary lock acquired. this is false when the transaction is running in the backup owner or a non-owner. On other hand, the affected keys contains all the touched keys.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
pruivo: because the key is touched by the transaction.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
pruivo: PFER is invoking `visitNonTxDataWriteCommand` and not this method.

pruivo:
updated!

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
rvansa: OK, is there any other way to get non-tx write command here? (If in doubt I usually throw exception and if testsuite doesn't fail, it's good to rule out one weird special case with error).

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
rvansa: Yes, but previously the key was added to the affected set only if it was locked. Also, you test the locking (both primary and backup) through `context.getAffectedKeys().contains(key)` - so you need to decide; should this set contain only locked or all touched keys?

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -134,151 +102,137 @@ public Object visitPrepareCommand(TxInvocationContext ctx, PrepareCommand comman
...
+      //we need to acquire locks
+      LockControlCommand lcc = cf.buildLockControlCommand(keys, command.getFlagsBitSet(), ctx.getGlobalTransaction());
+      lcc.setTopologyId(topologyId);
+      return makeStage(handleLocalLockCommand(ctx, lcc))
+            .thenApply(ctx, command, (rCtx, rCommand, rv) -> invokeNext(rCtx, rCommand));
+   }
+
+   private Object handleLocalLockCommand(TxInvocationContext context, LockControlCommand command)
+         throws InterruptedException {
+      //TODO blocking method. change it to non-blocking? [minor, only blocks local commands]
```
rvansa: Maybe you've missed my previous comment here: I believe that this part must not be blocking as it would block `xxxAsync` methods.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
pruivo: touched keys. I didn't change the semantic of the set. See: https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java#L167

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
pruivo: you are the one who added the if-condition (https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java#L164) IMO, it doesn't make sense to have non-tx commands here...

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
pruivo: that is another issue. I won't fix it in this PR

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
rvansa: OK, my bad that `handleWriteManyCommand` does not add them to the set correctly.
In that case, though, you need to change https://github.com/infinispan/infinispan/pull/5442/files#diff-922309dc5606a439d0cda2f8ab6ba470R186 and https://github.com/infinispan/infinispan/pull/5442/files#diff-922309dc5606a439d0cda2f8ab6ba470R190 because touched does not imply locked.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
rvansa: I don't recall why I did that; could you replace `invokeNext` with `throw new IllegalStateException()`?

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
rvansa: ☹️ Please create a JIRA and mention it in the comment then.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
pruivo: changing to `assert` if you don't mind

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
pruivo: touched keys implied they are copied to the transaction context. the key may change after being copied to the tx context. locking doesn't ensure nothing that's why we didn't do it before. 
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java#L206

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
pruivo: created a jira: https://issues.jboss.org/browse/ISPN-8336

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
rvansa: What do you mean by "the key may change after being copied to the tx context"?
Locking does ensure that unless there's another operation with SKIP_LOCKING, this operation does atomically updates DC & persistence. I don't claim that the behaviour in master is correct.
Suppose that a transaction has been executed with one write using SL flag, and another write (on the same key) without the SL flag. The second operation must acquire the lock despite the key was touched before.

pruivo: using your example, the first write copies the key to the transaction context without locking the key. between the first and the second write, any transaction may commit. Acquire the locking in the second write doesn't giving you anything since you no longer have access to the most updated value. 

we said it in the javadoc:
```
So this flag is useful only as an optimization when the same key is written once and never again, or as an unsafe optimisation if the period between writes on the same key is large enough to make a race condition never happen in practice. If this is unclear, avoid it.
```

IMO, the behavior in master is correct

rvansa: @pruivo It gives you the guarantee that the cache won't end up with V1 in DC and V2 in cachestore, if there's a concurrent write writing V2. The javadoc also says "Bypasses lock acquisition for **this invocation**". You're right that you lose the correctness of the return value in the second write, but it's the same as if you execute a `get()` without forcing a write lock.
The period between the write w/SL and the second+concurrent write can be very long, so it matches even the suggestion later on.

pruivo: you're telling me what is written in the javadoc. the user is warned about the inconsistency between DC and CS. If he is using the flag, he knows what to expect. 
However, I still don't see why acquiring the locking in the second command would fix. Except if you wanna fix all the inconsistencies created by the SL flag. In that case, create a new JIRA :)

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
danberindei: @rvansa @pruivo in an optimistic cache, the backup lock would be converted to a primary lock by replaying the prepare command during commit. Here I assume there is no change, just that an existing backup lock for key `k` prevents any transaction from acquiring a primary lock for `k`. So other transactions cannot lock `k`, but the transaction that acquired the primary lock on the old primary owner can still commit.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
pruivo: BTW, forgot to say, if you want to remove the SKIP_LOCKING support from transactional caches, I'm 100% in favor :) SKIP_LOCKING and CACHE_MODE_LOCAL are the 2 flags that doesn't fit in the transactional  semantics

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -190,95 +135,89 @@ public Object visitLockControlCommand(TxInvocationContext ctx, LockControlComman
...
+   private boolean areAllKeysAlreadyLocked(TxInvocationContext<?> context, Collection<?> keys) {
+      return context.getAffectedKeys().containsAll(keys);
+   }
 
-      if (command.isUnlock()) {
-         if (ctx.isOriginLocal()) throw new AssertionError(
-               "There's no advancedCache.unlock so this must have originated remotely.");
-         return false;
-      }
+   private boolean isKeysAlreadyLocked(TxInvocationContext<?> context, Object key) {
```
danberindei: s/Keys/Key/

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
danberindei: This new algorithm sounds very much like the old algorithm, the only difference seems to be the removal of `isSingleKeyAndLocal` optimization. (Which I believe was supposed to mean a single key in the transaction, so it was broken all along.)

@pruivo are you sure we can't fix that optimization, maybe at the expense of some more work during state transfer? Requiring an extra RPC when the local node is the primary owner would definitely impact the latency of HotRod writes to pessimistic caches.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -1,126 +1,94 @@
...
+ * owner and it is exclusive; only one transaction can acquire it. The backup lock is acquired by the backup owner and
+ * it is shared; multiple transactions can acquire simultaneously. The backup locks are associated with the topology id
+ * and they are only used when the primary owner leaves or crashes.
+ * <p>
+ * This algorithm contacts all owners and acquires the primary lock in the primary owner and the backup lock in the
+ * backup owners. So, lock acquisition (single or multiple keys) happens in a single RPC to all the owners of the
+ * key(s). This is triggered in {@link TxDistributionInterceptor#visitLockControlCommand(TxInvocationContext,
+ * LockControlCommand)}.
+ * <p>
+ * The {@link Flag#CACHE_MODE_LOCAL} changes this behaviour; it makes the lock acquisition only local. So, only local
```
danberindei: s/local//

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -1,126 +1,94 @@
...
+ * <p>
+ * In case of primary owner changes, there are 2 cases.
+ * <p>
+ * 1) a new node joins and it is the new primary owner. In this case, the old primary owner transfers the lock
+ * information to the new primary owner.
+ * <p>
+ * 2) the primary owner crashes or leaves. In this case, a backup owner will be promoted to the primary owner. All
+ * primary lock requests are hold until its previous backup locks are released. Note that a transaction only acquires
+ * the lock once. If a transaction requests a primary lock but has the backup lock acquired, it doesn't grant the
+ * primary lock if there are other backup locks from other transactions.
```
danberindei: Shouldn't this be "If a transaction has acquired the backup lock, it will not try to acquire the primary lock" ?

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
```diff
@@ -1,126 +1,94 @@
...
+ * <p>
+ * Know Issue: There is small deadlock probably is 2 or more transaction try to acquire the lock of a key and the
+ * primary owner dies before any of them receives a reply. In that case, none of the transactions (and future
+ * transactions) are able to acquire the primary lock until all backup lock are cleared. Eventually, the transactions
+ * will rollback and the backup lock will be released. One way to solve it is to acquire the backup locks after the
+ * primary lock. However, this will increase the performance penalty for all locks by generating 2 RPCs.
```
danberindei: IMO the originator should release the backup lock (only for this key, not for all the keys in the transaction) and retry if any it gets a `CacheNotFoundResponse` or `OutdatedTopologyException` from any of the targets.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
rvansa: @pruivo I don't care about inconsistencies due to SL in the first command - that's really expected. We don't try to fix those inconsistencies by design.
Acquiring the lock properly by the second command would guarantee that such transaction ends up with the cache in consistent state.
While ignoring SL in tx caches would be tempting, I think it would break the current implementation of atomic maps the least.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
rvansa: @danberindei You can't just release a lock and reacquire it - you'd be leaving a window to put in some value not observed by the modifying transaction. Unless we go with WSC in pessimistic caches, which would be my preference.

danberindei: @rvansa The lock release and re-acquisition would both happen before the `lock()` operation finished on the originator. If a transaction tries to acquire the lock after reading it, there's already a window for other transactions to modify the value.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
danberindei: @rvansa I agree that we use `SKIP_LOCKING` (and `CACHE_MODE_LOCAL`) with transactions too much internally in order to eliminate it now. Maybe if we had a way to do at least local writes without transactions in tx cache (like PFER, but more generic).

OTOH the `SKIP_LOCKING` javadoc does say it is for keys that do not change often, so a key being written to twice in the same transaction is clearly out of scope, and I wouldn't spend too much time trying to fix consistency in that case.

rvansa: @danberindei @pruivo As you wish, guys. We can fix it when we need it for internal works as we want to hide the flag anyway.
@pruivo May I ask you to document `getAffectedKeys()` that this is used for both *touched* and *locked* keys despite the key may not be actually locked?

rvansa: ... and maybe a TODO in the interceptor. We've made a decision, let's document it so that we won't question it again after couple of months.

core/src/main/java/org/infinispan/interceptors/locking/PessimisticLockingInterceptor.java
rvansa: @danberindei OK, I thought you suggest to re-acquire the lock during commit.

danberindei: @rvansa No, that would really require WSC :)

pruivo:
@danberindei yes it can be fixed with a more expensive state transfer. I don't know which idea do you have in mind, but it would require the state transfer to contact all nodes to collect the locked keys for the segments the "new primary owner" has received. Then, it acquires the lock. The backup lock can be removed since it isn't required anymore.

I don't mind to drop this changes and fix it later. @rvansa wdyt?

rvansa:
@pruivo If this would mean a big regression to Hot Rod (though only with tx caches), it could be better to rethink it. If you're unsure how to implement it, we could have a breakout session in Barcelona. Or you could write a design doc first.

pruivo:
@danberindei @rvansa closing this PR for now. I've added a breakout to the agenda. We can also discuss the affected key and the other sets used by transactions (remove, merge or leave it as well)

danberindei:
@pruivo Did we end up discussing this during the F2F? I don't remember reaching any conclusion...

pruivo:
@danberindei no. we didn't have time :(

PR #5447, opened 2017-09-19
Jenkins logs/artifacts improvements

danberindei:
Archive the dump/dumpstream files generated by Surefire
Log the correct thread name after we call Thread.setName()


tristantarrant:
Merged

PR #5448, opened 2017-09-19
Ispn 7865 Reactive stream iterator implementation

wburns:
https://issues.jboss.org/browse/ISPN-7865
https://issues.jboss.org/browse/ISPN-8338

The main benefits of this PR are that we use far less dedicated threads when doing an iterator process. In the old model we would use:

- 2 threads (originating node user and async for local iteration process)
- N-1 threads (1 for every remote node)

In the new model we only dedicate 
- 1 thread (user)

All other threads are only used briefly to retrieve a batch size of entries and immediately return and become idle, allowing for more concurrent requests.

Also the entire iteration process has been overhauled to provide:

1. Lazily marshalled iteration process (this process reduces iteration cost to only 1 iteration in most cases) - old method required at least 3 iterations
2. Prioritized entry usage with decreased locking contention. Remote entries are prioritized allowing for local entries to be returned when there is a gap in remote entries being available, allowing for less time that the iterator is blocking.
3. Less memory usage as currently we cap the amount of entries to a fixed number instead of a size that scaled with the number of nodes.

These were achieved with some of the things below:

- Adding [IntSet](https://github.com/infinispan/infinispan/pull/5448/files#diff-a34e2e625f8c44b3658fa436af27ac42R1) interface and usage across streams (prevents unnecessary boxing and unboxing) - This will need to be taken further in another PR
- [IteratorHandler](https://github.com/infinispan/infinispan/pull/5448/files#diff-ada1ae1b816c6f6a6f0f1af88e987735R1) now keeps track of open iteration requests (This way it is only used during a request and doesn't keep a thread used always on each node for iteration).
- This also has a special [iterator response](https://github.com/infinispan/infinispan/pull/5448/files#diff-127bd12873f1097653e69ef8602c60e8R25) that does not require the size of the response as well as optionally supplying the fact that segments were lost until marshalling.
- Changed [ClusterStreamManager](https://github.com/infinispan/infinispan/pull/5448/files#diff-83112cef2238f7c65212a112e94aeb8aR188) to return a new `RemoteIteratorPublisher` that allows for asynchronous retrieval of remote entries from remote nodes that also has separate callbacks for segment completion and lost.
- Created composite [PriorityMergingProcessor](https://github.com/infinispan/infinispan/pull/5448/files#diff-c48245ea653b177d754a04d4cf001c2cR39) which is a reactive streams processor that combines multiple `Publisher`s so that entries from each publisher are enqueued up to their configured batch size. Then the iterator will pull from the publishers in order until it finds a value. This allows for the remotely retrieved entries to be prioritized over locally retrieved entries which should provide iterator less time blocking waiting for an entry.
- Priority processor also [asynchronously requests](https://github.com/infinispan/infinispan/pull/5448/files#diff-c48245ea653b177d754a04d4cf001c2cR237) new entries from existing remote nodes after a certain amount of entries have been consumed, further trying to ensure that entries are always available.
- Currently this only submits request to [4](https://github.com/infinispan/infinispan/pull/5448/files#diff-be6f435a8edfd997f62d9a90ad59582bR942) remote nodes concurrently. This could be tweaked, however it was found that this amount kept memory usage lower as we only used up `batchSize * 4` for remote nodes at max but still gave good iterator request throughput. Maybe at some point this should be dynamic in the future or should we send to all nodes concurrently?


wburns:
Forgot to add test for intStream for both RangeSet and SmallIntSet, have to add that first as well.

wburns:
Test failure http://ci.infinispan.org/job/Infinispan/job/PR-5448/3/consoleFull#-12420200015fcafecb-7e79-4bda-8584-4be4ccf4b80e

Cause: https://gist.github.com/wburns/ea01383958d2644aca32fa0b2d32f67a

wburns:
CI ran earlier and there were 2 failures, which are unrelated.

Also @tristantarrant can you make sure to double check the security part.

tristantarrant:
Needs rebase, @wburns

wburns:
Rebased

wburns:
Sigh, CI seems to be stuck again. However the thread dump shows just CI related stuff in logs. I don't believe this is caused by this PR. Restarting it again though, just in case.

wburns:
Can confirm that other jobs are stuck as well. So just CI hiccup again :(

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -72,11 +78,12 @@
 import org.infinispan.stream.impl.intops.object.PeekOperation;
 import org.infinispan.stream.impl.termop.object.ForEachBiOperation;
 import org.infinispan.stream.impl.termop.object.ForEachOperation;
-import org.infinispan.stream.impl.termop.object.NoMapIteratorOperation;
-import org.infinispan.util.CloseableSuppliedIterator;
 import org.infinispan.util.RangeSet;
```
ryanemerson: should use `org.infinispan.commons.util.RangeSet`

core/src/main/java/org/infinispan/stream/impl/ClusterStreamManagerImpl.java
```diff
@@ -2,46 +2,71 @@
...
+import org.infinispan.remoting.responses.ExceptionResponse;
 import org.infinispan.remoting.responses.Response;
+import org.infinispan.remoting.responses.SuccessfulResponse;
+import org.infinispan.remoting.rpc.ResponseMode;
 import org.infinispan.remoting.rpc.RpcManager;
+import org.infinispan.remoting.rpc.RpcOptions;
 import org.infinispan.remoting.transport.Address;
 import org.infinispan.remoting.transport.jgroups.SuspectException;
+import org.infinispan.stream.impl.intops.IntermediateOperation;
 import org.infinispan.util.RangeSet;
```
ryanemerson: `org.infinispan.commons.util.RangeSet`

core/src/main/java/org/infinispan/stream/impl/IteratorHandler.java
```diff
@@ -0,0 +1,244 @@
...
+import org.infinispan.notifications.cachemanagerlistener.event.ViewChangedEvent;
+import org.infinispan.remoting.transport.Address;
+import org.infinispan.stream.impl.intops.IntermediateOperation;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+
+/**
+ * Handler for storing iterators between invocations so that remote node can request data in chunks.
+ * @author wburns
+ * @since 9.1
```
ryanemerson: 9.2

core/src/main/java/org/infinispan/stream/impl/IteratorHandler.java
```diff
@@ -0,0 +1,244 @@
...
+         return this;
+      }
+   }
+
+   /**
+    * Given two Runnables, return a Runnable that executes both in sequence,
+    * even if the first throws an exception, and if both throw exceptions, add
+    * any exceptions thrown by the second as suppressed exceptions of the first.
+    */
+   private static Runnable composeWithExceptions(Runnable a, Runnable b) {
```
ryanemerson: Is it worth creating a StreamUtil class for code like this that is repeated?

core/src/main/java/org/infinispan/stream/impl/PriorityMergingProcessor.java
```diff
@@ -0,0 +1,298 @@
...
+         try {
+            signalled.set(true);
+            condition.signalAll();
+         } finally {
+            lock.unlock();
+         }
+      }
+   }
+
+   static final class QueueSubscriber<T> extends AtomicReference<Subscription> implements Subscriber<T>, AutoCloseable {
```
ryanemerson: Can be private

core/src/main/java/org/infinispan/stream/impl/IteratorHandler.java
wburns: It isn't really specific to streams per se. I can add it to `commons.util.Util` as this is really just a Java 8 thing that allows composition of exceptions when combining runnables.

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
wburns: Sure I can update all of these.

core/src/main/java/org/infinispan/stream/impl/tx/TxClusterStreamManager.java
```diff
@@ -90,19 +98,60 @@ public void forgetOperation(Object id) {
...
+
    private static class TxExcludedKeys<K> extends AbstractDelegatingMap<Integer, Set<K>> {
       private final Map<Integer, Set<K>> map;
       private final Map<Integer, Set<K>> ctxMap;
 
-      private TxExcludedKeys(Map<Integer, Set<K>> map, LocalTxInvocationContext ctx, ConsistentHash hash) {
+      private TxExcludedKeys(Map<Integer, Set<K>> map, LocalTxInvocationContext ctx, ToIntFunction<Object> intFunction) {
          this.map = map;
-         this.ctxMap = contextToMap(ctx, hash);
+         this.ctxMap = contextToMap(ctx, intFunction);
```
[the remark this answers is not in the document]
wburns: Not sure I understand the question. The `convertToMap` eagerly converts all the entries into a map keyed by the segment, so that we can very easily find which entries map to which. This is important as the context could be checked multiple times and we don't want to compute it every time.

core/src/main/java/org/infinispan/stream/impl/IteratorHandler.java
ryanemerson: Sounds good

core/src/main/java/org/infinispan/stream/impl/tx/TxClusterStreamManager.java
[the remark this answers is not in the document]
ryanemerson: The `ctxMap` is always generated in the constructor of `TxExcludedKeys`, so why can't we just have the `contextToMap` function in the scope of `TxClusterStreamManager` and pass the calculated `ctxMap` to the constructor of `TxExcludedKeys`. That way we don't have to pass both the `ctx` and `intFunction` instances to the constructor of `TxExcludedKeys` (or to `contextToMap` as both are class variables).

Furthermore, and I'm not sure if this is the case, but if `ctx.getLookedUpEntries()` will always return the same entries for a given ctx, then we can just compute `ctxMap` once.

wburns:
Updated

wburns:
Tests all passed, rebasing though.

wburns:
Rebased the pom.xml changes. Nothing else has changed since last test run.

core/src/main/java/org/infinispan/stream/impl/tx/TxClusterStreamManager.java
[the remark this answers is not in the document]
wburns: In this case we have to calculate this map per invocation. The ctx can and will change between calls. It can even change between the constructor being called and invoking the remote stream operation.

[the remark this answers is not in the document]
wburns: TxClusterStreamManager constructor is called when user invokes `stream` or `parallelStream` in a transactional context. There is nothing preventing them from then changing the context before invoking a terminal operation on that stream (which finally calls `remoteStream*` methods).

core/src/main/java/org/infinispan/stream/impl/intops/FlatMappingOperation.java
```diff
@@ -0,0 +1,22 @@
+package org.infinispan.stream.impl.intops;
+
+import java.util.stream.BaseStream;
+import java.util.stream.Stream;
+
+/**
+ * Interface to signify that an {@link IntermediateOperation} is a flat map operation. This also provides proper
+ * generics for converting a flat map as a map operation resulting in a Stream containing the proper stream
+ * @author wburns
+ * @since 9.0
```
danberindei: Shouldn't these be 9.2 instead?

PR #5452, opened 2017-09-21
ISPN-8326 LevelDB test methods executed when SkipOnOS us used

mgencur:
https://issues.jboss.org/browse/ISPN-8326

### DIFF
--- a/commons-test/src/main/java/org/infinispan/commons/test/skip/SkipOnOsListener.java
+++ b/commons-test/src/main/java/org/infinispan/commons/test/skip/SkipOnOsListener.java
@@ -4,12 +4,33 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import org.testng.IAnnotationTransformer;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.AfterGroups;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.AfterSuite;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeGroups;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.BeforeSuite;
 import org.testng.annotations.ITestAnnotation;
+import org.testng.annotations.Test;
 
 public class SkipOnOsListener implements IAnnotationTransformer {
+
+   private static final Class[] TESTNG_ANNOTATIONS = new Class[] {
+           BeforeSuite.class,
+           AfterSuite.class,
+           BeforeClass.class,
+           AfterClass.class,
+           BeforeMethod.class,
+           AfterMethod.class,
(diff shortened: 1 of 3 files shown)


danberindei:
@mgencur probably unrelated, but there are some failures in `DirectoryProviderModuleMemberRegistrationIT`:

Caused by: org.h2.jdbc.JdbcSQLException: Unique index or primary key violation: "UK9QV6YHJQM8IAFTO8QK452GX8H_INDEX_8 ON PUBLIC.MEMBER(EMAIL) VALUES ( /* key:1 */ 2, 'davide@mailinator.com', 37.780392, -122.513898, 'Davide D''Altoe', '2125551235')"; SQL statement:
insert into Member (email, latitude, longitude, name, phone_number, id) values (?, ?, ?, ?, ?, ?) [23505-173]

mgencur:
@danberindei , am I supposed to do something about that? My changes are related to TestNG listener and this test uses JUnit so I don't see how they can be related.

danberindei:
@mgencur I just needed a confirmation that it's unrelated. Of course, after reading your comment I also found https://issues.jboss.org/browse/ISPN-7983 :)

tristantarrant:
Wouldn't it be easier to just use maven properties to include/exclude os-specific tests ?

mgencur:
That would be a completely different approach and probably easier from the beginning:)

danberindei:
@mgencur @tristantarrant I wouldn't go so far as to involve Maven in this... I think the problem is just that `SkipOnOSListener` is actually an `IAnnotationTransformer`, so it doesn't see the inherited methods. It also means that methods annotated with `@SkipOnOS` but not with with `@Test` always run.

Personally I'd prefer an explicit `assumeOS(OS.UNIX)` in the test method, with `assumeOS` throwing a `SkipException`, like in this article:

http://blog.xebia.com/conditionally-running-tests-in-testng/

I'd also nitpick about Solaris actually being in the UNIX family, but I have a feeling Solaris is an option because more because of random test failures then because of fundamental differences between Solaris and Linux :)

mgencur:
@danberindei , I think it would not solve the problem with inherited methods. You would have to override each method from the superclass anyway and place assumeOS in it. You can't do that in the superclass' methods because, like in this case, one sub-class is supposed to run (the Java implementation of LevelDB) and the other one should be skipped (the JNI impl.).
You can place the SkipOnOs on each overridden method in the sub-class too but the advantage is that it can be placed on the whole class as well.

danberindei:
@mgencur this seems to skip all the methods, with only a small glitch in IntelliJ (`assumeOS` still has the in-progress animation after the test finished):


    @BeforeClass
    public void assumeOS() {
       throw new SkipException("Skipped on Linux");
    }

mgencur:
@danberindei , so you're requesting the removal of all SkipOn* classes and annotations and implementing  this new solution?

danberindei:
@mgencur yes, I actually opened a new PR: https://github.com/infinispan/infinispan/pull/5559

PR #5454, opened 2017-09-22
ISPN-8334 ISPN-8307 Passivate enties evicted from off-heap containter

vjuranek:
https://issues.jboss.org/browse/ISPN-8334
https://issues.jboss.org/browse/ISPN-8307

### DIFF
--- a/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
@@ -257,13 +257,16 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
                log.tracef("Removing entry: %d due to eviction due to size %d being larger than maximum of %d",
                      addressToRemove, currentSize, maxSize);
             }
+            InternalCacheEntry<WrappedBytes, WrappedBytes> ice = null;
             try {
-               InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(addressToRemove);
-               passivator.passivate(ice);
+               ice = offHeapEntryFactory.fromMemory(addressToRemove);
                performRemove(addressToRemove, offHeapEntryFactory.getKey(addressToRemove));
                evictionManager.onEntryEviction(Collections.singletonMap(ice.getKey(), ice));
             } finally {
                entryWriteLock.unlock();
+               if (ice != null) {
+                  passivator.passivate(ice);
+               }
             }
          }
       }
--- a/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
@@ -259,6 +259,7 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
             }
             try {
                InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(addressToRemove);
+               passivator.passivate(ice);
                performRemove(addressToRemove, offHeapEntryFactory.getKey(addressToRemove));
                evictionManager.onEntryEviction(Collections.singletonMap(ice.getKey(), ice));
             } finally {
--- a/core/src/test/java/org/infinispan/eviction/impl/EvictionWithPassivationTest.java
+++ b/core/src/test/java/org/infinispan/eviction/impl/EvictionWithPassivationTest.java
@@ -9,16 +9,20 @@ import java.util.concurrent.TimeUnit;
 
 import org.infinispan.Cache;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.configuration.cache.StorageType;
 import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.persistence.dummy.DummyInMemoryStoreConfigurationBuilder;
 import org.infinispan.test.SingleCacheManagerTest;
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -259,6 +259,7 @@ private void ensureSize() {
             }
             try {
                InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(addressToRemove);
+               passivator.passivate(ice);
```
wburns: I think if we can move the `passivate` call to after we unlock the `entryWriteLock` would be better. This way we won't block other off heap reads and writes for the same lock while performing the passivation, which can take a while. We will still hold onto the ISPN lock for the given key so it should be fine.

wburns:
Also just a reminder to whomever integrates, this will want to be backported to 9.1.x as well.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
wburns: We should probably also update https://github.com/vjuranek/infinispan/blob/1398f14e570068daa48c00ef2d4c4b0bcd1c0391/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java#L499 as well to make that after we unlock the `lock`.

vjuranek:
PR updated (with minor changes to reduce allocations when reading entry and and to avoid updating LRU node when evicting an entry)

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
danberindei: @wburns Which ISPN lock? Passivation should not acquire the `LockManager` lock, because regular writes ignore it on the backup owners.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -257,12 +257,16 @@ private void ensureSize() {
...
+            InternalCacheEntry<WrappedBytes, WrappedBytes> ice = null;
             try {
-               InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(addressToRemove);
+               ice = offHeapEntryFactory.fromMemory(addressToRemove);
                performRemove(addressToRemove, offHeapEntryFactory.getKey(addressToRemove));
                evictionManager.onEntryEviction(Collections.singletonMap(ice.getKey(), ice));
             } finally {
                entryWriteLock.unlock();
+               if (ice != null) {
+                  passivator.passivate(ice);
```
danberindei: Passivation potentially invokes listeners and throws an exception, so I would put it after the finally block.

core/src/test/java/org/infinispan/eviction/impl/EvictionWithPassivationTest.java
```diff
@@ -9,34 +9,53 @@
...
       // size being equal to full container size
       cleanup = CleanupPhase.AFTER_METHOD;
    }
 
-   private final int EVICTION_MAX_ENTRIES = 2;
+   @Factory
+   public Object[] factory() {
+      return new Object[] {
+            new EvictionWithPassivationTest().withStorage(StorageType.BINARY),
+            new EvictionWithPassivationTest().withStorage(StorageType.OBJECT),
```
danberindei: Is `BINARY` any different from `OBJECT` when it comes to eviction/passivation?

danberindei:
@vjuranek I'd also like to see more tests extended for off-heap passivation, e.g. `EvictionWithPassivationAndConcurrentOperationsTest` or `OffHeapBoundedSingleNodeTest`.

core/src/test/java/org/infinispan/eviction/impl/EvictionWithPassivationTest.java
wburns: It shouldn't be, but no reason not to just run all 3 :)

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
wburns: Sorry yes there is no lock, in which case we have to use the striped lock from the data container to guarantee consistency.

wburns:
Sorry @vjuranek I didn't think it totally through, we can't passivate the entry after unlocking the lock. This would allow a read lock to be acquired and could possibly see a time frame where the entry is not present in the off heap container and not in the store. We have to passivate in the lock critical section, however the ordering in that section doesn't matter since it is a read write lock and we have the write lock (although it would be a bit cleaner to ensure it is passivated before we removed).

wburns:
And actually that is a good point @danberindei implementing `EvictionWithPassivationAndConcurrentOperationsTest` would provide guarantees that my last comment doesn't occur.

wburns:
Integrated the first commit into master and 9.1, thanks @vjuranek !

vjuranek:
First, sorry for the delay.
I looked on the test today again and it seems to me, that the tests (especially `EvictionWithPassivationAndConcurrentOperationsTest`) cannot be done until [ISPN-8385](https://issues.jboss.org/browse/ISPN-8385) is fixed - or at least I failed to find a way how to determine if the cache entry is in memory when underlying `datacontainer.containsKey()` cannot be used.

tristantarrant:
Closing this.

PR #5455, opened 2017-09-22
ISPN-8333 HotRod client test suite fails with java.net.BindException …

mgencur:
…when run through IKVM

https://issues.jboss.org/browse/ISPN-8333

This is rather a workaround for IKVM. The constructor of ServerSocket(0) in findFreePort() method throws the "socket closed" exception. I've checked my ephemeral port range on Windows and there are plenty of ports. It looks like this is a problem directly in IKVM because running the HotRod client tests separately on Windows works fine and the ports are reserved correctly.


galderz:
Looks good, integrating

PR #5457, opened 2017-09-22
ISPN-8323 Cache.evict(...) throws NPE if entry was already evicted and

wburns:
an eviction listener is present

* Use regular for loop instead of Collectors

https://issues.jboss.org/browse/ISPN-8323

### DIFF
--- a/core/src/test/java/org/infinispan/eviction/impl/EvictionFunctionalTest.java
+++ b/core/src/test/java/org/infinispan/eviction/impl/EvictionFunctionalTest.java
@@ -91,6 +91,22 @@ public class EvictionFunctionalTest extends SingleCacheManagerTest {
       assertTrue(evictionListener.evictedEntries.size() > CACHE_SIZE);
    }
 
+   public void testEvictNonExistantEntry() {
+      String key = "key";
+      String value = "some-value";
+      cache.put(key, value);
+
+      cache.evict(key);
+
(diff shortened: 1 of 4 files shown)


wburns:
This needs to be backported to 9.1.x as well after this is integrated.  Should cherrypick as is.

core/src/test/java/org/infinispan/eviction/impl/EvictionFunctionalTest.java
```diff
@@ -91,6 +91,22 @@ public void testSimpleEvictionMaxEntries() throws Exception {
...
+      cache.put(key, value);
+
+      cache.evict(key);
+
+      assertEquals(1, evictionListener.evictedEntries.size());
+
+      // Make sure if we evict again that it doesn't increase count
+      cache.evict(key);
+
+      // TODO: this seems like a bug, but many tests rely on this - maybe change later
```
danberindei: I'm sure if we had many tests that evicted an entry twice we'd have noticed the NPE earlier :)

wburns: Yeah but this comment is the fact that an evict causes an event even when there as nothing to evict. :(

tristantarrant:
Merged to master and 9.1.x

PR #5458, opened 2017-09-25
ISPN-5783 Cache.put(null, null) leaks an implicit transaction

danberindei:
https://issues.jboss.org/browse/ISPN-5783

### DIFF
--- a/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
+++ b/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
@@ -24,6 +24,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.function.IntFunction;
 
 import javax.security.auth.Subject;
 import javax.transaction.SystemException;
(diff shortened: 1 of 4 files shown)


rvansa:
While your code is more concise, I think that the try-catch block should be in the method that creates the context (e.g. calls `DecoratedCache.writeContext`).

danberindei:
@rvansa IMO `tryRollbackInjected()` and `executeCommandAndCommitIfNeeded()` should be called at the same level.

It would have been even better if we created the JTA transaction only in `executeCommandAndCommitIfNeeded`, but preserving the `DecoratedCache`-level locking was getting a bit ugly.

rvansa:
@danberindei Yes, that would be even better, but if we can't solve the difficulties with that, I wouldn't scatter the release of resources to even another layer of calls.

pruivo:
@danberindei well, you could change the internal methods to receive a `Function<Integer, InvocationContext>` instead of `InvocationContext` and invoke the function after the `assert()`. You can create the context in the `executeCommandAndCommitIfNeeded()` and it fits the `DecoratedCache.writeContext()` method signature. 
But I'm ok with all the try-catch stuff...

wburns:
@pruivo I like that :+1: Also could be `IntFunction<InvocationContext>`

danberindei:
run performance tests please

danberindei:
@pruivo @wburns I wanted the most common case to use a constant lambda, so I made it a `Function<CacheImpl, InvocationContext>`. Maybe a field would work just as well, so I'll also get some results with `IntFunction<InvocationContext>`.

pruivo:
Well, I don't think the function/lambda would have much of a impact in the code. But if you want to micro-optimize this is my suggestion:
* `Supplier<InvocationContext>` for single key context (and PFER)
* `IntFunction<InvocationContext>` for multi-key context
* remove the PFER code from internal `putIfAbsent` and create a method for it. 
:)

danberindei:
@pruivo Looks like it does have a bit of an impact, -3.57 % for repl reads on 100 threads.

danberindei:
run performance tests please

jmarkos:
@danberindei Sorry I had to stop the perfack, I forgot to update it when infinispan version was bumped to 9.2.0-SNAPSHOT. I'll run it tomorrow morning, to also have correct upstream numbers to compare with. I'm afraid the results from 11 days ago were also not correct, sorry :(

rvansa:
@jmarkos Could you automatize the script to take a proper version? I know we don't change versions *that* often, but I think this is getting annoying.

jmarkos:
@rvansa It's not that much about taking a proper version, the problem is that we do radargun plugins for each minor version (infinispan90, infinispan91, etc). Would you be okay with a plugin named infinispan9x? (I don't like a plugin named infinispan91 with 9.2.0-SNAPSHOT inside.)

jmarkos:
run performance tests please

alanfx:
@jmarkos @rvansa what about using a plugin named `infinispan-latest-master` for PerfAck and logging the version?

rvansa:
@jmarkos Not sure what you're asking for... in the RG codebase, we should have distinct plugins for minors. When you're duplicating the plugin using specific snapshot, call it whatever you want. I don't see a problem with plugin `xxx` that is a copy of `infinispan91` plugin with 9.2.0-SNAPSHOT in, as long as it compiles and runs. Of course we'll have to add another plugin if you want to adapt new features/some backwards compatibility is broken (since RG may use unsupported API).

PR #5461, opened 2017-09-25
ISPN-8339 Cache Stream response should send asynchronously

wburns:
https://issues.jboss.org/browse/ISPN-8339

### DIFF
--- a/core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
+++ b/core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
@@ -6,6 +6,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ForkJoinPool;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


wburns:
All tests passed http://ci.infinispan.org/job/Infinispan/job/PR-5461/

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -247,8 +267,9 @@ public void dataRehashed(DataRehashedEvent<K, V> event) {
       operation.setSupplier(() -> getStream(cacheEntrySet, parallelStream, segments, keysToInclude, keysToExclude));
       operation.handleInjection(registry);
       R value = operation.performOperation();
-      rpc.invokeRemotely(Collections.singleton(origin), factory.buildStreamResponseCommand(requestId, true,
-              Collections.emptySet(), value), rpc.getDefaultRpcOptions(true));
+      CompletableFuture<Map<Address, Response>> completableFuture = rpc.invokeRemotelyAsync(Collections.singleton(origin),
```
rvansa: If `trace` is not on, you could even use `RpcManager.sendTo`. Not sure about the elegance of non-trivial differences in `if (trace) { ... }`

wburns: Hrmm I never realized that method was there, thanks for the pointer. I should be able to refactor it to a common method that does all the hard work.

wburns: Ah actually that makes the test behave differently as well as I would have to optionally replace the correct method invocation. I would say for now it can stay as it is to reduce complexity.

tristantarrant:
Merged to master, thanks

PR #5463, opened 2017-09-25
ISPN-8340 Cache.size in a REPL cache should be performed locally only

wburns:
https://issues.jboss.org/browse/ISPN-8340

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
@@ -26,6 +26,7 @@ import org.infinispan.commands.read.AbstractDataCommand;
 import org.infinispan.commands.read.GetAllCommand;
 import org.infinispan.commands.read.GetCacheEntryCommand;
 import org.infinispan.commands.read.GetKeyValueCommand;
+import org.infinispan.commands.read.SizeCommand;
 import org.infinispan.commands.remote.ClusteredGetAllCommand;
 import org.infinispan.commands.remote.ClusteredGetCommand;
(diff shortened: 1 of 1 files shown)


danberindei:
@wburns off-topic, but this has been nagging me for a while... do we really need `SizeCommand`? I mean calling `keySet().stream().count()` directly from `CacheImpl.size()` sounds simpler to me than creating a command and running it through all the interceptors only to do the same thing.

wburns:
@danberindei I agree and tbh I wanted to do this when I changed this a long time ago. But I believe some people were worried about custom interceptors and such. I am all game for removing the command and such. I guess I would only be able to deprecate the Visitor method though. Although it wouldn't actually do anything anymore. @tristantarrant What kind of rules do we have on that?

tristantarrant:
I don't think we have any special rules. I'm all for dropping the command (and we should probably have "laxer" rules for maintaining backwards compatibility within the interceptor chain).

wburns:
So looking closer, unfortunately there are 2 interceptors that actually mess all of this up: `VersionedCallInterceptor` and `TombstoneCallInterceptor`. These do the same thing as the command, in which case I would have to also move this logic into `CacheImpl`. Doing so would make this pretty ugly not existing in the interceptors.

So I am thinking maybe we should keep the interceptors and I can just change this to check for REPL in the `BaseDistributionInterceptor` and update the command flags there. Otherwise I could add a `Predicate` argument to the SizeCommand so that this is passed down to the actual `SizeCommand` invocation.

@danberindei @rvansa WDYT?

wburns:
Updated to use `BaseDistributionInterceptor` to update the `CACHE_MODE_LOCAL` flag to the `SizeCommand` when using a REPL cache.

tristantarrant:
Merged, thanks

PR #5465, opened 2017-09-26
ISPN-5475 Narayana should use a volatile store for all tests

danberindei:
https://issues.jboss.org/browse/ISPN-5475


tristantarrant:
Didn't we already do this a while back. It seems to creep back in every now and then...

tristantarrant:
@danberindei , no ObjectStore files archived, so I guess this is looking good

danberindei:
@tristantarrant we did do something, but AFAICT it only worked for core.

danberindei:
Ready for review, I removed the archive step.

tristantarrant:
Merged thanks

PR #5469, opened 2017-09-27
ISPN-8348 Some methods from cache API do not work properly with off-heap

wburns:
* Fixed more bugs with encoder cache wrapper
* Document equality issue with off heap and binary

https://issues.jboss.org/browse/ISPN-8348
https://issues.jboss.org/browse/ISPN-8359
https://issues.jboss.org/browse/ISPN-8355


wburns:
This should be backported to 9.1.x as well.

gustavonalle:
LGTM, but build failed due to checkstyle errors

wburns:
Oops fixed.

wburns:
I have to look at some stuff still.

core/src/main/java/org/infinispan/commands/CommandsFactoryImpl.java
```diff
@@ -200,7 +201,8 @@ public void setupDependencies(DataContainer container, CacheNotifier<Object, Obj
                                  OrderedUpdatesManager orderedUpdatesManager, StateTransferLock stateTransferLock) {
       this.dataContainer = container;
       this.notifier = notifier;
-      this.cache = cache;
+      // We shouldn't use encoded entries or anything else here
+      this.cache = AbstractDelegatingCache.unwrapCache(cache);
```
gustavonalle: FYI, ```advancedCache.withEncoding(IdentityEncoder.class)``` will avoid doing any conversion with the cache. 
Not sure if the test failures are caused by this, but the encoding stuff does not support "naked" caches atm.

wburns:
I also found https://issues.jboss.org/browse/ISPN-8355 while fixing others. Let me update the PR and see how many failures there are now. In core there was just some around `keyFiliter` but there could be more.

wburns:
Added in delegates for cache stream to properly encode keys provided from `CacheStream` interface. This way all encoding should be done in the `EncoderCache` for Streams and the underlying Stream just starts the stream in the uncoded state.

wburns:
Hoping that CI passes now :)

wburns:
My only concern is that now a stream that uses an `EncodedCache` which is all of them has 1 additional wrapper around it. If we find this is too costly we could leak the encoding code into the various `CacheStream` implementations do the unwrapping of the filtered keys as well.

gustavonalle:
Is CI dead?

gustavonalle:
retriggered manually

gustavonalle:
@wburns LGTM.

WRT additional wrapping, it seems to me that an extra wrapper compared to the amount of work typically performed by a stream operation is not significant, do you have a test to run some benchmarks?

wburns:
@gustavonalle There are no benchmark tests. But thinking more I agree, this would make inconsequential different for all DIST/REPL caches. Local/Inval cache would see the biggest difference but it should still be very negligible compared to the iteration of all entries.

gustavonalle:
Merged in master, thanks @wburns

PR #5472, opened 2017-09-28
 Fix the OSGi integration tests module so that it runs correctly

tristantarrant:
https://issues.jboss.org/browse/ISPN-8365


pruivo:
no JIRA?

tristantarrant:
Added

pruivo:
integrated! thanks @tristantarrant

PR #5473, opened 2017-09-28
ISPN-8346 Reduce the number of clustered listener in clustered counters

pruivo:
https://issues.jboss.org/browse/ISPN-8346

Review notes:
All the notifications triggered by the cache are handled by CounterManagerNotificationManager. 
Each counter registers on it together with the user listeners.


counter/src/main/java/org/infinispan/counter/impl/listener/CounterManagerNotificationManager.java
```diff
@@ -0,0 +1,229 @@
...
+    *
+    * @param counterName            The counter's name.
+    * @param generator              The counter's {@link CounterEvent} generator.
+    * @param topologyChangeListener The counter's listener to topology change. It can be {@code null}.
+    * @throws IllegalStateException If the counter with that name is already registered.
+    */
+   public void registerCounter(ByteString counterName, CounterEventGenerator generator,
+         TopologyChangeListener topologyChangeListener) {
+      if (counters.putIfAbsent(counterName, new Holder(generator, topologyChangeListener)) != null) {
+         throw new IllegalStateException();
```
tristantarrant: If this is likely to be triggered by wrong user code, maybe we should use a proper exception message. Otherwise, leave it as it is.

pruivo: If it is throw, it is our (read my) fault. When a counter instance is created, in the CounterManager, it registers in the notification manager. Since I have only one instance per counter, this register must happen only once (we don't support counters removal).
Maybe I should change to assert?

tristantarrant: No need.

tristantarrant: No need.

tristantarrant:
Merged, thanks

PR #5480, opened 2017-09-29
Revert "ISPN-7986 Hot RodTransaction (Server only)"

pruivo:
Reverts infinispan/infinispan#5240


pruivo:
I guess the revert button doesn't work very well :)

PR #5484, opened 2017-10-02
ISPN-7919 Expose ResponseCollector in the RpcManager interface 

danberindei:
https://issues.jboss.org/browse/ISPN-7919
https://issues.jboss.org/browse/ISPN-8120

Opening a new PR because I force-pushed on the https://github.com/infinispan/infinispan/pull/5315 branch...

Note that I don't intend the `invokeCommand` calls to stay like this forever. My goal is to use `ResponseCollector` implementations almost everywhere we now use `invokeRemotelyAsync().thenApply()`.


wburns:
Seems to be quite a bit of failures, unlikely to not be unrelated.

http://ci.infinispan.org/job/Infinispan/job/PR-5484/4/#showFailuresLink

galderz:
One failure:

    org.infinispan.distribution.DistAsyncFuncTest.testComputeIfPresentFromNonOwner[DIST_ASYNC, tx=false, groupers=true]

core/src/main/java/org/infinispan/remoting/transport/impl/MapResponseCollector.java
```diff
@@ -23,36 +24,43 @@
  * @author Dan Berindei
  * @since 9.1
  */
-public class SyncMapResponseCollector extends ValidResponseCollector<Map<Address, Response>> {
+public class MapResponseCollector extends ValidResponseCollector<Map<Address, Response>> {
    private final HashMap<Address, Response> map;
    private final boolean ignoreLeavers;
```
pruivo: I think the boolean can be removed and have 2 implementation. this is my suggestion (ps. the same idea for the remaining response collectors!)

```
public abstract class MapResponseCollector extends ValidResponseCollector<Map<Address, Response>> {
   public static MapResponseCollector defaultCollector() {
      return new DefaultCollector(4);
   }

   public static MapResponseCollector ignoreLeaversCollector() {
      return new IgnoreLeaversCollector(4);
   }

   public static MapResponseCollector defaultCollector(int size) {
      return new DefaultCollector(size);
   }

   public static MapResponseCollector ignoreLeaversCollector(int size) {
      return new IgnoreLeaversCollector(size);
   }

...
   private static class DefaultCollector extends MapResponseCollector {

      private DefaultCollector(int expectedSize) {
         super(expectedSize);
      }

      @Override
      protected Map<Address, Response> addTargetNotFound(Address sender) {
         recordException(ResponseCollectors.remoteNodeSuspected(sender));
         return null;
      }
   }

   private static class IgnoreLeaversCollector extends MapResponseCollector {

      private IgnoreLeaversCollector(int expectedSize) {
         super(expectedSize);
      }

      @Override
      protected Map<Address, Response> addTargetNotFound(Address sender) {
         map.put(sender, CacheNotFoundResponse.INSTANCE);
         return null;
      }
   }
```

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
```diff
@@ -114,7 +216,9 @@
     * @param deliverOrder  the {@link org.infinispan.remoting.inboundhandler.DeliverOrder}.
     * @return a new {@link RpcOptionsBuilder} with the default options and the response mode and deliver mode set by the
     * parameters.
+    * @deprecated Since 9.2, please use {@link #getSyncRpcOptions()} instead.
     */
+   @Deprecated
    RpcOptionsBuilder getRpcOptionsBuilder(ResponseMode responseMode, DeliverOrder deliverOrder);
```
pruivo: should the `RpcOptionsBuilder` be deprecated also? Doesn't seems to be used anywhere since you are using the `RpcOption` directly.

core/src/main/java/org/infinispan/remoting/rpc/RpcManagerImpl.java
```diff
@@ -106,24 +137,165 @@ public String getPendingViewAsString() {
       return pendingCH != null ? pendingCH.getMembers().toString() : "null";
    }
 
+   @Override
+   public <T> CompletionStage<T> invokeCommand(Address target, ReplicableCommand command,
+                                               ResponseCollector<T> collector, RpcOptions rpcOptions) {
+      checkTopologyId(command);
```
pruivo: Most of the method looks the same. can't we avoid that code duplication? 
```
invokeCommand(Address...)
invokeCommand(Collection<Address> targets...)
invokeCommandOnAll(...)
invokeCommandStaggered(...)
```

core/src/test/java/org/infinispan/remoting/rpc/RpcManagerTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+      CompletionStage<Response> stage2 =
+            rpcManager.invokeCommand(address(1), clusteredGetCommand, SingleResponseCollector.INSTANCE,
+                                     rpcManager.getSyncRpcOptions());
+      assertEquals(SuccessfulResponse.SUCCESSFUL_EMPTY_RESPONSE, stage2.toCompletableFuture().join());
+   }
+
+   @Test
+   public void testInvokeCommandCollection() throws Exception {
+   }
```
pruivo: where is the test?

core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
```diff
@@ -401,6 +405,7 @@ private void sendToBackups(DistributionInfo distributionInfo, DataWriteCommand c
       if (trace) {
          log.tracef("Command %s got sequence %s for segment %s", id, sequenceNumber, segmentId);
       }
+      // TODO Should we use sendToAll in replicated mode?
```
pruivo: there is no triangle in replicated mode :)

core/src/main/java/org/infinispan/configuration/cache/ClusteringConfiguration.java
```diff
@@ -25,6 +25,8 @@ static AttributeSet attributeDefinitionSet() {
    }
 
    private final Attribute<CacheMode> cacheMode;
+   private final Attribute<Long> remoteTimeout;
+   private final Attribute<Integer> invalidationBatchSize;
```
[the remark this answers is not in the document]
rvansa: `invalidationBatchSize` is used by scattered cache.

core/src/main/java/org/infinispan/interceptors/distribution/TriangleDistributionInterceptor.java
danberindei: Yes, but there could be :)

core/src/main/java/org/infinispan/remoting/rpc/RpcManagerImpl.java
danberindei: I wanted to avoid creating another lambda per invocation on top of the updateStatistics one. I'll have to check if escape analysis can remove the allocation, otherwise I'd rather keep the duplication.

core/src/main/java/org/infinispan/remoting/transport/impl/MapResponseCollector.java
danberindei: Yeah, your proposal is probably better. 

I didn't want to make these collectors too fancy because they were supposed to be only a crutch until we start using custom collectors on each call site (and avoid adding a `CompletableFuture` callback to transform the responses into something else). But considering how long this PR has taken, worked on them a little more.

core/src/main/java/org/infinispan/configuration/cache/ClusteringConfiguration.java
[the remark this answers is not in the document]
danberindei: I forgot to modify the read method to use the direct field instead of the attribute collection.

core/src/main/java/org/infinispan/remoting/transport/impl/MapResponseCollector.java
pruivo: can you elaborate your idea? I'm curious :)

core/src/test/java/org/infinispan/remoting/rpc/RpcManagerTest.java
```diff
@@ -0,0 +1,180 @@
...
+import org.infinispan.test.Exceptions;
+import org.infinispan.test.MultipleCacheManagersTest;
+import org.infinispan.test.TestingUtil;
+import org.testng.annotations.Test;
+
+/**
+ * @author Dan Berindei
+ * @since 9.2
+ */
+public class RpcManagerTest extends MultipleCacheManagersTest {
```
pruivo: we no longer annotate the class with the groups and test name?

core/src/main/java/org/infinispan/interceptors/distribution/ScatteredDistributionInterceptor.java
```diff
@@ -732,8 +744,14 @@ public Object visitClearCommand(InvocationContext ctx, ClearCommand command) thr
       // local mode clear will have unpredictable results
       svm.clearInvalidations();
       if (ctx.isOriginLocal() && !isLocalModeForced(command)) {
-         RpcOptions rpcOptions = isSynchronous(command) ? syncIgnoreLeavers : defaultAsyncOptions;
-         return makeStage(asyncInvokeNext(ctx, command, rpcManager.invokeRemotelyAsync(null, command, rpcOptions)))
+         if (!isSynchronous(command)) {
```
pruivo: I would invert the `if` here too

core/src/main/java/org/infinispan/remoting/transport/impl/SingletonMapResponseCollector.java
```diff
@@ -13,18 +13,33 @@
...
+import org.infinispan.remoting.transport.jgroups.JGroupsTransport;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
 
 /**
  * Response collector supporting {@link JGroupsTransport#invokeRemotelyAsync(Collection, ReplicableCommand, ResponseMode, long, ResponseFilter, DeliverOrder, boolean)}.
  *
  * @author Dan Berindei
- * @since 9.1
+ * @since 9.2
```
pruivo: why did the version change?

core/src/main/java/org/infinispan/remoting/transport/impl/SingletonMapResponseCollector.java
```diff
@@ -13,18 +13,33 @@
...
  */
-class SingletonMapResponseCollector
+public class SingletonMapResponseCollector
       extends ValidSingleResponseCollector<Map<Address, Response>> {
+   private static final Log log = LogFactory.getLog(SingletonMapResponseCollector.class);
+   private static final SingletonMapResponseCollector VALID_ONLY = new SingletonMapResponseCollector(false);
+   private static final SingletonMapResponseCollector IGNORE_LEAVERS = new SingletonMapResponseCollector(true);
+
    private final boolean ignoreLeavers;
```
pruivo: I would have done the same thing as in `MapResponseCollector`

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -543,18 +541,23 @@ private void awaitTotalOrderTransactions(CacheTopology cacheTopology, boolean is
       }
    }
 
-   protected void notifyEndOfStateTransferIfNeeded() {
+   protected boolean notifyEndOfStateTransferIfNeeded() {
       if (waitingForState.get()) {
          if (hasActiveTransfers()) {
-            return;
+            log.tracef("notifyEndOfStateTransferIfNeeded: no active transfers");
```
pruivo: prepend `if (trace)`?

core/src/main/java/org/infinispan/configuration/cache/SyncConfiguration.java
```diff
@@ -24,19 +24,20 @@
...
 
    SyncConfiguration(AttributeSet attributes) {
       this.attributes = attributes.checkProtection();
-      remoteTimeout = attributes.attribute(ClusteringConfiguration.REMOTE_TIMEOUT);
+      remoteTimeout = attributes.attribute(REPL_TIMEOUT);
    }
 
    /**
-    * This is the timeout used to wait for an acknowledgment when making a remote call, after which
+    * This is the timeout used to wait for an acknowledgment when ma      remoteTimeout.set()
```
pruivo: typo?

core/src/main/java/org/infinispan/remoting/transport/Transport.java
```diff
@@ -248,6 +263,6 @@
    @Experimental
    <T> CompletionStage<T> invokeCommands(Collection<Address> targets,
                                          Function<Address, ReplicableCommand> commandGenerator,
-                                         ResponseCollector<T> responseCollector, long timeout,
-                                         DeliverOrder deliverOrder);
+                                         ResponseCollector<T> collector, DeliverOrder deliverOrder,
```
pruivo: hmm isn't the `Transport` kind of public API? you can set it via `TransportConfigurationBuilder.transport(Transport);`

core/src/main/java/org/infinispan/remoting/transport/impl/VoidResponseCollector.java
```diff
@@ -0,0 +1,71 @@
...
+ * a node is suspected and {@code ignoreLeavers == true}.
+ *
+ * @author Dan Berindei
+ * @since 9.2
+ */
+public class VoidResponseCollector extends ValidResponseCollector<Void> {
+   private static VoidResponseCollector VALID_ONLY = new VoidResponseCollector(false);
+   private static VoidResponseCollector IGNORE_LEAVERS = new VoidResponseCollector(true);
+
+   private final boolean ignoreLeavers;
```
pruivo: same as `MapResponseCollector`?

core/src/main/java/org/infinispan/configuration/cache/SyncConfiguration.java
danberindei: pasto :)

core/src/main/java/org/infinispan/remoting/transport/Transport.java
danberindei: I was thinking that I've protected myself against that by marking the method `@Experimental` when I first added it, but you're right, I should have also made it `default`.

pruivo: ah... I didn't see the annotation :)

core/src/main/java/org/infinispan/remoting/transport/impl/SingletonMapResponseCollector.java
danberindei: I thought I had renamed the class.

core/src/main/java/org/infinispan/remoting/transport/impl/SingletonMapResponseCollector.java
danberindei: It makes sense to remove the field in `MapResponseCollector` because it saves some memory, but here we don't allocate new instances for each request.

core/src/test/java/org/infinispan/remoting/rpc/RpcManagerTest.java
danberindei: I was trying to see what happens if I remove all the test names at one point, maybe I was influenced by that. Or maybe I just forgot :)

core/src/main/java/org/infinispan/remoting/transport/impl/SingletonMapResponseCollector.java
danberindei: Actually I did move it, so I'd keep 9.2

core/src/main/java/org/infinispan/remoting/transport/impl/VoidResponseCollector.java
danberindei: same ;)

pruivo:
@danberindei compile error:
```
infinispan/core/src/test/java/org/infinispan/remoting/transport/MockTransport.java:[32,49] cannot find symbol
[ERROR]   symbol:   class SyncMapResponseCollector
[ERROR]   location: package org.infinispan.remoting.transport.jgroups
```

danberindei:
Rebased!

pruivo:
@danberindei needs rebase again :(

danberindei:
Rebased again!

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
```diff
@@ -42,9 +119,14 @@
     *                   not change it. Any change in {@link RpcOptions} during a remote invocation can lead to
     *                   unpredictable behavior.
     * @return  a map of responses from each member contacted.
+    * @deprecated Since 9.2, please use {@link #invokeCommand(Collection, ReplicableCommand, ResponseCollector, RpcOptions)} instead.
     */
    Map<Address, Response> invokeRemotely(Collection<Address> recipients, ReplicableCommand rpc, RpcOptions options);
```
pruivo: you forgot to annotate the method as `@deprecated`. 
There are still usages in `LocalStreamManagerImpl` and `BiasManagerImpl`. Do you want to "fix" it in this PR or later?

pruivo: edit: and `XSiteAdminOperations`

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -605,6 +605,10 @@ public synchronized void addPendingAcks(Address[] waitFor) {
        */
       public synchronized void primaryResult(Address from, Object value, boolean success) {
          Collector<Response> collector = primaryCollectors.get(from);
+         if (trace) {
+            log.tracef("[Collector#%s] PutMap Primary ACK. Address=%s. TopologyId=%s (expected=%s).",
+                       id, from, topologyId, this.topologyId);
```
pruivo: topology id isn't needed.

core/src/main/java/org/infinispan/remoting/rpc/RpcManager.java
danberindei: Initially I intended to replace both the usages of `invokeRemotely()` and `invokeRemotelyAsync`. But now I see there are quite a few usages of `invokeRemotely()` remaining (or maybe added since I first created the PR), so I'm leaving them for a separate PR. 

I did replace the usages of `invokeRemotelyAsync()` in `LocalStreamManagerImpl`, `BiasManagerImpl`, and `TransactionDecodeContext`.

PR #5489, opened 2017-10-02
ISPN-8363 wildfly-modules missing required dependencies

ryanemerson:
--InfinispanExtensionIT test added

https://issues.jboss.org/browse/ISPN-8363


ryanemerson:
@danberindei It doesn't appear to be as there is a green build on CI and it passes every time locally.

PR #5492, opened 2017-10-03
ISPN-8368 java.lang.String cannot be cast to org.infinispan.commons.m…

gustavonalle:
…arshall.WrappedBytes with off-heap

https://issues.jboss.org/browse/ISPN-8368

### DIFF
--- a/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/AbstractRESTClientIT.java
+++ b/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/AbstractRESTClientIT.java
@@ -55,9 +55,13 @@ public abstract class AbstractRESTClientIT {
         }
     }
 
+    protected String getDefaultCache() {
+        return "default";
+    }
+
     @Before
     public void setUp() throws Exception {
-        rest = new RESTHelper();
+        rest = new RESTHelper(getDefaultCache());
         addRestServer();
 
         cleanUpEntries();
--- a/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
+++ b/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
@@ -35,17 +35,17 @@ public class RESTOffHeapClientIT extends AbstractRESTClientIT {
    }
 
    @Test
-   @Ignore
+   @Ignore("ISPN-8370")
    public void testPutDataTTLMaxIdleCombo1() throws Exception {
    }
 
    @Test
-   @Ignore
+   @Ignore("ISPN-8370")
    public void testPutDataTTLMaxIdleCombo2() throws Exception {
    }
 
    @Test
-   @Ignore
+   @Ignore("ISPN-8370")
    public void testPutDataWithMaxIdleTime() throws Exception {
    }
 }
--- a/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
+++ b/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
@@ -1,0 +1,51 @@
+package org.infinispan.server.test.client.rest;
+
+import org.infinispan.arquillian.core.InfinispanResource;
+import org.infinispan.arquillian.core.RemoteInfinispanServer;
+import org.infinispan.server.test.category.RESTSingleNode;
+import org.jboss.arquillian.junit.Arquillian;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+/**
+ * @since 9.2
+ */
+@RunWith(Arquillian.class)
+@Category({RESTSingleNode.class})
+public class RESTOffHeapClientIT extends AbstractRESTClientIT {
+
+   @InfinispanResource("container1")
+   RemoteInfinispanServer server1;
+
+   @Override
+   protected void addRestServer() {
+      rest.addServer(server1.getRESTEndpoint().getInetAddress().getHostName(), server1.getRESTEndpoint().getContextPath());
+   }
+
+   @Override
+   protected String getDefaultCache() {
+      return "offHeapCache";
+   }
+
+   @Test
+   @Ignore
+   public void testPutDataWithTimeToLive() throws Exception {
+   }
+
+   @Test
+   @Ignore
+   public void testPutDataTTLMaxIdleCombo1() throws Exception {
+   }
+
+   @Test
+   @Ignore
+   public void testPutDataTTLMaxIdleCombo2() throws Exception {
+   }
+
+   @Test
+   @Ignore
+   public void testPutDataWithMaxIdleTime() throws Exception {
+   }
+}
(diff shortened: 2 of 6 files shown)


server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
```diff
@@ -0,0 +1,51 @@
...
+   }
+
+   @Test
+   @Ignore
+   public void testPutDataTTLMaxIdleCombo2() throws Exception {
+   }
+
+   @Test
+   @Ignore
+   public void testPutDataWithMaxIdleTime() throws Exception {
```
gustavonalle: In the ```DefaultDataContainer```, expiration is checked when the entries are accessed, but for the ```OffHeapDataContainer``` those 4 tests don't remove the entry for some reason

danberindei: @gustavonalle @wburns Is this a bug?

gustavonalle: Created https://issues.jboss.org/browse/ISPN-8370 to handle this, as this PR is to solve only the CCE

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/AbstractRESTClientIT.java
```diff
@@ -55,9 +55,13 @@ public String getContent() {
         }
     }
 
+    protected String getDefaultCache() {
+        return "default";
+    }
+
     @Before
     public void setUp() throws Exception {
-        rest = new RESTHelper();
```
slaskawi: It seems this call has exactly the same result as previously. So perhaps we could either revert it to `rest = new RESTHelper()` or remove default constructor from `RESTHelper`?

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
```diff
@@ -0,0 +1,51 @@
...
+import org.infinispan.arquillian.core.RemoteInfinispanServer;
+import org.infinispan.server.test.category.RESTSingleNode;
+import org.jboss.arquillian.junit.Arquillian;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+/**
+ * @since 9.2
```
slaskawi: Empty stuff to be removed.

BTW, how about doing Off-Heap tests on functional level in REST module? Perhaps those tests would be faster?

server/rest/src/test/java/org/infinispan/rest/RestOperationsOffHeapTest.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.configuration.cache.StorageType;
+import org.testng.annotations.Test;
+
+@Test(groups = "functional", testName = "rest.RestOperationsOffHeapTest")
+public class RestOperationsOffHeapTest extends RestOperationsTest {
+
+   @Override
+   public ConfigurationBuilder getDefaultCacheBuilder() {
```
slaskawi: It is worth to run whole test suite (there are many tests in `RestOperationsTest`) just for one scenario? 

Maybe it's worth to write a single test, that would call PUT and check cache content?

server/rest/src/test/java/org/infinispan/rest/RestOperationsTest.java
```diff
@@ -37,10 +37,14 @@
    private HttpClient client;
```
slaskawi: If you agree with my previous comment, changes in this file can be reverted.

slaskawi:
@gustavonalle I'm not requesting changes since my comments are not critical. But I think implementing a single test in REST module (rather than running full operations test suite with off-heap cache) might be worth pursuing from execution time stand point.

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
gustavonalle: The PR has ```RestOperationsOffHeapTest``` for that.  IMO we need both, since usually the behavior of embedded servers is different

server/rest/src/test/java/org/infinispan/rest/RestOperationsOffHeapTest.java
gustavonalle: > It is worth to run whole test suite (there are many tests in RestOperationsTest) just for one scenario?

Definitely. During a simple put and get does not cover the whole spectrum of features, take for example ISPN-8370 that was just created

server/rest/src/test/java/org/infinispan/rest/RestOperationsTest.java
gustavonalle: I'd rather keep the changes. It does not make a difference in test execution time.

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
gustavonalle: "since 9.2" is not empty, it's important information about when the test was introduced :smile:

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/AbstractRESTClientIT.java
gustavonalle: The default ctor is used in several other places, and the effect of this change is to allow inheritors of ```AbstractRESTClientIT``` to override the cache name used in the test,

gustavonalle:
rebase

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
danberindei: @gustavonalle why not keep the `@since` tag and also add a short description? ;)

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
danberindei: Ok, but please use something like `@Category(Unstable.class) // See ISPN-8370` instead of `@Ignore`.

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
gustavonalle: @danberindei IMO the test name is self explanatory, WDYT?

gustavonalle:
@danberindei Added a new commit pointing to the JIRA for the expiration stuff

server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/RESTOffHeapClientIT.java
danberindei: @gustavonalle Ok, if it always fails `@Ignore` is just as good :)

danberindei:
@slaskawi @gustavonalle +1 to keep all the test methods, if a test doesn't insert thousands of keys and doesn't start/stop a WildFly server it's very hard for it to change the build time in a meaningful way.

danberindei:
Integrated, thanks Gustavo!

PR #5493, opened 2017-10-03
ISPN-7993 Encoding support on functional maps

karesti:
* Comments addressed
* Tx caches + Encoding not supported
* Added test with encoding on AdvancedCache
* Removed component registry in command's copy
* Caching decoded values in entries

### DIFF
--- a/core/src/main/java/org/infinispan/commands/CommandsFactoryImpl.java
+++ b/core/src/main/java/org/infinispan/commands/CommandsFactoryImpl.java
@@ -18,12 +18,16 @@ import javax.transaction.xa.Xid;
 
 import org.infinispan.Cache;
 import org.infinispan.commands.control.LockControlCommand;
+import org.infinispan.commands.functional.AbstractWriteKeyCommand;
+import org.infinispan.commands.functional.AbstractWriteManyCommand;
 import org.infinispan.commands.functional.ReadOnlyKeyCommand;
 import org.infinispan.commands.functional.ReadOnlyManyCommand;
 import org.infinispan.commands.functional.ReadWriteKeyCommand;
 import org.infinispan.commands.functional.ReadWriteKeyValueCommand;
 import org.infinispan.commands.functional.ReadWriteManyCommand;
 import org.infinispan.commands.functional.ReadWriteManyEntriesCommand;
+import org.infinispan.commands.functional.TxReadOnlyKeyCommand;
+import org.infinispan.commands.functional.TxReadOnlyManyCommand;
 import org.infinispan.commands.functional.WriteOnlyKeyCommand;
 import org.infinispan.commands.functional.WriteOnlyKeyValueCommand;
 import org.infinispan.commands.functional.WriteOnlyManyCommand;
@@ -87,6 +91,7 @@ import org.infinispan.context.InvocationContextFactory;
 import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.distribution.DistributionManager;
 import org.infinispan.distribution.group.impl.GroupManager;
+import org.infinispan.encoding.DataConversion;
 import org.infinispan.factories.ComponentRegistry;
 import org.infinispan.factories.KnownComponentNames;
 import org.infinispan.factories.annotations.ComponentName;
@@ -544,8 +549,32 @@ public class CommandsFactoryImpl implements CommandsFactory {
             InvalidateVersionsCommand invalidateVersionsCommand = (InvalidateVersionsCommand) c;
             invalidateVersionsCommand.init(dataContainer, orderedUpdatesManager, stateTransferLock, stateTransferManager);
             break;
+
+         // === Functional commands ====
+         case ReadOnlyKeyCommand.COMMAND_ID:
+         case TxReadOnlyKeyCommand.COMMAND_ID:
+            ((ReadOnlyKeyCommand) c).init(componentRegistry);
+            break;
+
+         case ReadOnlyManyCommand.COMMAND_ID:
+         case TxReadOnlyManyCommand.COMMAND_ID:
+            ((ReadOnlyManyCommand) c).init(componentRegistry);
+            break;
+
          case ReadWriteKeyCommand.COMMAND_ID:
-            ((ReadWriteKeyCommand) c).init(componentRegistry);
+         case ReadWriteKeyValueCommand.COMMAND_ID:
+         case WriteOnlyKeyCommand.COMMAND_ID:
+         case WriteOnlyKeyValueCommand.COMMAND_ID:
+            ((AbstractWriteKeyCommand) c).init(componentRegistry);
+            break;
+
+         case ReadWriteManyCommand.COMMAND_ID:
+         case ReadWriteManyEntriesCommand.COMMAND_ID:
+         case WriteOnlyManyCommand.COMMAND_ID:
+         case WriteOnlyManyEntriesCommand.COMMAND_ID:
+            ((AbstractWriteManyCommand) c).init(componentRegistry);
+            break;
+
          default:
             ModuleCommandInitializer mci = moduleCommandInitializers.get(c.getCommandId());
             if (mci != null) {
@@ -697,36 +726,36 @@ public class CommandsFactoryImpl implements CommandsFactory {
    }
 
    @Override
-   public <K, V, R> ReadOnlyKeyCommand<K, V, R> buildReadOnlyKeyCommand(K key, Function<ReadEntryView<K, V>, R> f, Params params) {
-      return new ReadOnlyKeyCommand<>(key, f, params);
+   public <K, V, R> ReadOnlyKeyCommand<K, V, R> buildReadOnlyKeyCommand(K key, Function<ReadEntryView<K, V>, R> f, Params params, DataConversion keyDataConversion, DataConversion valueDataConversion) {
+      return new ReadOnlyKeyCommand<>(key, f, params, keyDataConversion, valueDataConversion, componentRegistry);
    }
 
    @Override
-   public <K, V, R> ReadOnlyManyCommand<K, V, R> buildReadOnlyManyCommand(Collection<? extends K> keys, Function<ReadEntryView<K, V>, R> f, Params params) {
-      return new ReadOnlyManyCommand<>(keys, f, params);
+   public <K, V, R> ReadOnlyManyCommand<K, V, R> buildReadOnlyManyCommand(Collection<? extends K> keys, Function<ReadEntryView<K, V>, R> f, Params params, DataConversion keyDataConversion, DataConversion valueDataConversion) {
+      return new ReadOnlyManyCommand<>(keys, f, params, keyDataConversion, valueDataConversion, componentRegistry);
    }
 
    @Override
-   public <K, V, R> ReadWriteKeyValueCommand<K, V, R> buildReadWriteKeyValueCommand(
-         K key, V value, BiFunction<V, ReadWriteEntryView<K, V>, R> f, Params params) {
-      return new ReadWriteKeyValueCommand<>(key, value, f, generateUUID(transactional), getValueMatcher(f),
-            params);
+   public <K, V, R> ReadWriteKeyValueCommand<K, V, R> buildReadWriteKeyValueCommand(K key, V value, BiFunction<V, ReadWriteEntryView<K, V>, R> f,
+                                                                                    Params params, DataConversion keyDataConversion, DataConversion valueDataConversion) {
+      return new ReadWriteKeyValueCommand(key, value, f, generateUUID(transactional), getValueMatcher(f),
+            params, keyDataConversion, valueDataConversion, componentRegistry);
    }
 
    @Override
    public <K, V, R> ReadWriteKeyCommand<K, V, R> buildReadWriteKeyCommand(
-         K key, Function<ReadWriteEntryView<K, V>, R> f, Params params) {
-      return new ReadWriteKeyCommand<>(key, f, generateUUID(transactional), getValueMatcher(f), params, componentRegistry);
+         K key, Function<ReadWriteEntryView<K, V>, R> f, Params params, DataConversion keyDataConversion, DataConversion valueDataConversion) {
+      return new ReadWriteKeyCommand<>(key, f, generateUUID(transactional), getValueMatcher(f), params, keyDataConversion, valueDataConversion, componentRegistry);
    }
 
    @Override
-   public <K, V, R> ReadWriteManyCommand<K, V, R> buildReadWriteManyCommand(Collection<? extends K> keys, Function<ReadWriteEntryView<K, V>, R> f, Params params) {
-      return new ReadWriteManyCommand<>(keys, f, params, generateUUID(transactional));
+   public <K, V, R> ReadWriteManyCommand<K, V, R> buildReadWriteManyCommand(Collection<? extends K> keys, Function<ReadWriteEntryView<K, V>, R> f, Params params, DataConversion keyDataConversion, DataConversion valueDataConversion) {
+      return new ReadWriteManyCommand<>(keys, f, params, generateUUID(transactional), keyDataConversion, valueDataConversion, componentRegistry);
    }
 
    @Override
-   public <K, V, R> ReadWriteManyEntriesCommand<K, V, R> buildReadWriteManyEntriesCommand(Map<? extends K, ? extends V> entries, BiFunction<V, ReadWriteEntryView<K, V>, R> f, Params params) {
-      return new ReadWriteManyEntriesCommand<>(entries, f, params, generateUUID(transactional));
+   public <K, V, R> ReadWriteManyEntriesCommand<K, V, R> buildReadWriteManyEntriesCommand(Map<? extends K, ? extends V> entries, BiFunction<V, ReadWriteEntryView<K, V>, R> f, Params params, DataConversion keyDataConversion, DataConversion valueDataConversion) {
+      return new ReadWriteManyEntriesCommand<>(entries, f, params, generateUUID(transactional), keyDataConversion, valueDataConversion, componentRegistry);
    }
 
    @Override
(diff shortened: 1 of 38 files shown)


core/src/main/java/org/infinispan/commands/CommandsFactoryImpl.java
```diff
@@ -544,8 +547,43 @@ public void initializeReplicableCommand(ReplicableCommand c, boolean isRemote) {
             InvalidateVersionsCommand invalidateVersionsCommand = (InvalidateVersionsCommand) c;
             invalidateVersionsCommand.init(dataContainer, orderedUpdatesManager, stateTransferLock, stateTransferManager);
             break;
+         // === Functional commands ====
+         case ReadOnlyKeyCommand.COMMAND_ID:
+            ((ReadOnlyKeyCommand)c).init(componentRegistry);
```
rvansa: Could you squash these cases (casting to `AbstractWriteKeyCommand` or `AbstractWriteManyCommand`)?

core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyValueCommand.java
```diff
@@ -144,20 +147,32 @@ public LoadType loadType() {
...
+            .append(", keyDataConversion=").append(keyDataConversion)
+            .append(", valueDataConversion=").append(valueDataConversion)
+            .append("}")
+            .toString();
    }
 
    @Override
    public Mutation toMutation(K key) {
-      return new Mutations.ReadWriteWithValue<>(value, f);
+      V valueFromStorage = (V) valueDataConversion.fromStorage(value);
```
rvansa: Is the `valueFromStorage` guaranteed to be serializable? I guess that since normal commands do all conversions on originator, it's not necessary, but I am not sure what do we require with this compatibility 2.0 stuff. @gustavonalle ?
Could you add a test case when the stored value is and the converted value is not?

core/src/main/java/org/infinispan/commands/functional/WriteOnlyKeyCommand.java
```diff
@@ -44,6 +54,8 @@ public void writeTo(ObjectOutput output) throws IOException {
       Params.writeObject(output, params);
       output.writeLong(FlagBitSets.copyWithoutRemotableFlags(getFlagsBitSet()));
       CommandInvocationId.writeTo(output, commandInvocationId);
+      output.writeObject(keyDataConversion);
```
rvansa: IIUC the keyDataConversion is not used for write-only commands, right? Though, I'd like to add the `key()` method to the `WriteEntryView` eventually...

core/src/main/java/org/infinispan/functional/impl/EntryViews.java
```diff
@@ -34,81 +35,93 @@ private EntryViews() {
       // Cannot be instantiated, it's just a holder class
    }
 
+   public static <K, V> ReadEntryView<K, V> readOnly(CacheEntry<K, V> entry, DataConversion keyDataConversion, DataConversion valueDataConversion) {
+      return new EntryBackedReadOnlyView<>(entry, keyDataConversion, valueDataConversion);
+   }
+
    public static <K, V> ReadEntryView<K, V> readOnly(CacheEntry<K, V> entry) {
-      return new EntryBackedReadOnlyView<>(entry);
+      return new EntryBackedReadOnlyView<>(entry, null, null);
```
rvansa: Why are you keeping the conversion-less method?

core/src/main/java/org/infinispan/functional/impl/FunctionalMapFactory.java
```diff
@@ -0,0 +1,24 @@
+package org.infinispan.functional.impl;
+
+import org.infinispan.functional.FunctionalMap;
+
+/**
+ * Helper Factory to create FunctionalMaps
```
rvansa: Helper for whom, when it's in impl package? If it's for tests, put it under `test/`

core/src/main/java/org/infinispan/functional/impl/FunctionalMapImpl.java
```diff
@@ -41,7 +41,7 @@
 
    private static <K, V> long getFlagsBitSet(Cache<K, V> cache) {
       long flagsBitSet = 0;
-      for (;;) {
+      for (; ; ) {
```
rvansa: I wish we could use the good ol' `#define EVER ;;` and type `for(EVER) { ... }` :)

core/src/main/java/org/infinispan/functional/impl/ReadOnlyMapImpl.java
```diff
@@ -67,7 +69,7 @@ private ReadOnlyMapImpl(Params params, FunctionalMapImpl<K, V> functionalMap) {
    @Override
    public Traversable<ReadEntryView<K, V>> entries() {
       log.tracef("Invoked entries(%s)", params);
-      CloseableIterator<CacheEntry<K, V>> it = fmap.cache.cacheEntrySet().iterator();
+      Iterator<CacheEntry<K, V>> it = fmap.cache.cacheEntrySet().iterator();
```
rvansa: Oops, this is a risky spot; the Traversable is not guaranteed to go through all entries and noone's going to call it.close(). Could you file a JIRA and add TODO referencing the JIRA about a possible leak of `Closeable`? We might need to make `Traversable` `Closeable` as well...
@wburns Any chance I am panicking here?

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -321,8 +326,8 @@ private Object handleSecondPhaseCommand(TxInvocationContext ctx, TransactionBoun
    }
 
    protected void checkTxCommandResponses(Map<Address, Response> responseMap,
-         TransactionBoundaryCommand command, TxInvocationContext<LocalTransaction> context,
-         Collection<Address> recipients) {
+                                          TransactionBoundaryCommand command, TxInvocationContext<LocalTransaction> context,
```
rvansa: Could you please set up your IDE to not autoformat lines that you are not changing?

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -544,7 +550,8 @@ protected ReadOnlyKeyCommand remoteReadOnlyCommand(InvocationContext ctx, ReadOn
       return cf.thenRun(() -> {
          entryFactory.wrapEntryForWriting(ctx, key, false, true);
          MVCCEntry cacheEntry = (MVCCEntry) ctx.lookupEntry(key);
-         EntryView.ReadWriteEntryView readWriteEntryView = EntryViews.readWrite(cacheEntry);
+         // TODO: ISPN-8090 support full cache encoding in tx cache
+         EntryView.ReadWriteEntryView readWriteEntryView = EntryViews.readWrite(cacheEntry, null, null);
```
rvansa: Can't you call `command.getKeyDataConversion()` here?

core/src/test/java/org/infinispan/functional/FunctionalInMemoryTest.java
```diff
@@ -55,7 +58,8 @@ public void testWriteLoad(boolean isOwner, WriteMethod method) {
                assertEquals(view.get(), "value");
                return null;
             },
-            (view, nil) -> {}, getClass());
+            (view, nil) -> {
```
rvansa: c'mon, `{}` really does not need to be reformatted.

core/src/main/java/org/infinispan/commands/functional/WriteOnlyKeyCommand.java
karesti: Yes it is not, I was thinking about it just right now, that the dependency to the keyDataConverter may be removed. But if you want to add the key method, in that case both are necessary. I can leave that and add the method

core/src/main/java/org/infinispan/functional/impl/FunctionalMapFactory.java
karesti: At first there was more code than what i is right now. But I prefer 
`FunctionalMapFactory.readOnlyMap(functionalMap)`
that telling the user to use `ReadOnlyMapImpl.create(functionalMap);`
Like having a single entry point. But well, this is my opinion, and this class can be removed

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
karesti: ok. When I apply the infinispan formatter, sometimes formats lines that are not in the formatter way. I didn't know I could say (without just choosing myself the lines that I want to format) just to format modified lines.

core/src/test/java/org/infinispan/functional/FunctionalInMemoryTest.java
karesti: This si the formatter who did that

core/src/main/java/org/infinispan/commands/CommandsFactoryImpl.java
karesti: All the method is like that, you mean on the default part see if it's instanceOf AbstractWriteKeyCommand or the other to call the init method ?

core/src/main/java/org/infinispan/commands/functional/WriteOnlyKeyCommand.java
rvansa: Don't mix changes in one PR, unless necessary, please... just remove it here for now.

core/src/main/java/org/infinispan/functional/impl/FunctionalMapFactory.java
rvansa: I dislike `new ReadOnlyMapImpl(fmap)` as well - but there should be only one recommended way (for users, tests and internal code alike) to create the impl.

core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
rvansa: What do you mean by "apply the infinispan formatter"? Don't do that if it "fixes" too much - you can also format just selected text.

karesti: I don't do it all the time, but sometimes it happens by mistake. I'm going to see the changes I did too much for this case and repush

core/src/main/java/org/infinispan/commands/functional/WriteOnlyKeyCommand.java
karesti: Well, since it's a small change I don't see the harm of adding the key method. but ok, this is not a big deal for me.

gustavonalle:
Needs rebase to be able to trigger CI

core/src/main/java/org/infinispan/commands/functional/WriteOnlyKeyCommand.java
karesti: @rvansa after looking this, the keyconverter is in the abstract class that is used in several parts. I'm going to leave it as it is, considering key method will be added just after on WriteOnly entry views

karesti:
Build is green, but I rebased the branch again and did two minor modifications. Let's wait for the last build. If it's green, I would like to merge as it is @gustavonalle @rvansa

core/src/main/java/org/infinispan/commands/functional/AbstractWriteKeyCommand.java
```diff
@@ -6,22 +6,38 @@
...
 
    Params params;
    ValueMatcher valueMatcher;
    boolean successful = true;
+   DataConversion keyDataConversion;
+   DataConversion valueDataConversion;
 
    public AbstractWriteKeyCommand(K key, ValueMatcher valueMatcher,
-         CommandInvocationId id, Params params) {
+                                  CommandInvocationId id, Params params) {
```
rvansa: old constructor can be removed

core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyValueCommand.java
gustavonalle: One use case is, a non-serializable pojo (user type) that is encoded as a UTF-8 byte[ ] with some string representation of it.

rvansa: @karesti I think that the fromStorage conversion here is not right, `value` is of user type (it's not converted in ReadWriteMapImpl.eval`)

rvansa: Had we encoded the value in RWMI (and decoded when executing the command) we'd handle the case as well; we'd also keep the invariant that all commands carry only encoded values. At the cost of (potentially) not-needed conversions.
What path should we go, WDYT?

core/src/main/java/org/infinispan/functional/impl/ReadWriteMapImpl.java
```diff
@@ -57,7 +58,8 @@ private ReadWriteMapImpl(Params params, FunctionalMapImpl<K, V> functionalMap) {
    @Override
    public <R> CompletableFuture<R> eval(K key, V value, BiFunction<V, ReadWriteEntryView<K, V>, R> f) {
       log.tracef("Invoked eval(k=%s, v=%s, %s)", key, value, params);
-      ReadWriteKeyValueCommand cmd = fmap.commandsFactory.buildReadWriteKeyValueCommand(key, value, f, params);
+      Object keyEncoded = keyDataConversion.toStorage(key);
+      ReadWriteKeyValueCommand cmd = fmap.commandsFactory.buildReadWriteKeyValueCommand(keyEncoded, value, (BiFunction) f, params, keyDataConversion, valueDataConversion);
```
rvansa: On IRC we've discussed that it'll be better to encode the value as well (and decode it before applying the function ofc), because the user type here might be not serializable.

karesti:
@rvansa thanks for the help on the many command 💃

karesti:
tests failing

org.infinispan.client.hotrod.retry.CompleteShutdownDistRetryTest.testRetryAfterCompleteShutdown
org.infinispan.it.osgi.features.OSGiKarafFeaturesTest.testCleanInstall

core/src/main/java/org/infinispan/functional/impl/ReadOnlyMapImpl.java
wburns: No, @rvansa you are correct. Good catch.

In this case resources would be left open if the Traversable was not used or if a short circuited terminal operation terminated early (findAny, noneMatch, anyMatch, allMatch).

wburns: Also it seems that this creation of the Iterator is unneeded as we should be able to pass the Stream `fmap.cache.cacheEntrySet().stream()` directly in. Changing the spliterator Characteristics shouldn't matter as the underlying one is already CONCURRENT and we enforce no write updates in the map method below.

core/src/main/java/org/infinispan/functional/FunctionalMap.java
```diff
@@ -86,6 +86,15 @@
    ComponentStatus getStatus();
 
    /**
+    * Tells if the underlying cache is using encoding or not
+    *
+    * @return true if the underlying cache is encoded
+    */
+   default boolean isEncoded() {
```
wburns: Looks like this isn't implemented/referenced anywhere. Seems we can remove. And even if it was, I believe _all_ caches are encoded now by default.

core/src/main/java/org/infinispan/functional/impl/AbstractFunctionalMap.java
```diff
@@ -150,4 +160,18 @@ private Transaction getOngoingTransaction() {
          return cf;
       }
    }
+
+   protected Set encodeKeys(Set<? extends K> keys) {
+      return keys.stream().map(k -> keyDataConversion.toStorage(k)).collect(Collectors.toSet());
```
wburns: I wonder if we want to convert this to use something similar to https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/cache/impl/EncoderCache.java#L87

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
```diff
@@ -29,7 +29,7 @@
  * Lifecycle - <tt>EmbeddedCacheManager</tt>s have a lifecycle (it implements {@link Lifecycle}) and
  * the default constructors also call {@link #start()}.  Overloaded versions of the constructors are available, that do
  * not start the <tt>CacheManager</tt>, although it must be kept in mind that <tt>CacheManager</tt>s need to be started
- * before they can be used to create <tt>Cache</tt> instances.
+ * before they can be used to readWriteMap <tt>Cache</tt> instances.
```
wburns: Looks like an inadvertent change?

core/src/test/java/org/infinispan/functional/FunctionalStorageTypeTest.java
```diff
@@ -0,0 +1,29 @@
...
+   protected void configureCache(ConfigurationBuilder builder) {
+      builder.memory().storageType(storageType);
+      super.configureCache(builder);
+   }
+
+   public Object[] factory() {
+      return new Object[]{
+            new FunctionalStorageTypeTest().storageType(StorageType.OFF_HEAP),
+            new FunctionalStorageTypeTest().storageType(StorageType.BINARY),
+            new FunctionalStorageTypeTest().storageType(StorageType.OBJECT),
```
wburns: `OBJECT` shouldn't be needed as `FunctionalMapTest` already runs this.

karesti: ok !

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
karesti: yes, a bad refactoring I did and I thought everything was "restored" but missed that

core/src/main/java/org/infinispan/functional/FunctionalMap.java
karesti: yes, true. exactly. At first this was not the case but with @gustavonalle change this was not needed anymore. 
There is ticket to make encoding work properly with transactional caches too (it's not really implemented here) so I think I'm going to report your comments there and ensure they are done there if you agree

core/src/main/java/org/infinispan/functional/impl/ReadOnlyMapImpl.java
rvansa: @wburns Would a direct use of stream change anything about closeability?

wburns: No, we would just have to close the Stream instead. The second change is about just less wrapping going on :)

karesti: @wburns @rvansa do you prefer a new ticket for this, or the ticket for tx cache support in functional map can be enough ? I don't see the size of the impact,  don't know if I create a new ticket for this, or not ...

wburns: @rvansa created https://issues.jboss.org/browse/ISPN-8378

PR #5501, opened 2017-10-05
SUREFIRE-1426 Find all fork errors in the build output

danberindei:
The previous solution was missing "Error occurred in starting fork"
messages.


gustavonalle:

PR #5506, opened 2017-10-06
ISPN-8337 MarshallUtil.unmarshallArray returns Object[] for all zero-length arrays

rvansa:
https://issues.jboss.org/browse/ISPN-8377


core/src/main/java/org/infinispan/commands/tx/PrepareCommand.java
```diff
@@ -211,7 +211,7 @@ public void readFrom(ObjectInput input) throws IOException, ClassNotFoundExcepti
       super.readFrom(input);
       onePhaseCommit = input.readBoolean();
       retriedCommand = input.readBoolean();
-      modifications = MarshallUtil.unmarshallArray(input, WriteCommand[]::new);
+      modifications = MarshallUtil.unmarshallArray(input, WriteCommand::array);
```
danberindei: Are prepare commands with 0 modifications that common?

core/src/main/java/org/infinispan/xsite/statetransfer/XSiteStatePushCommand.java
```diff
@@ -76,7 +76,7 @@ public void writeTo(ObjectOutput output) throws IOException {
    @Override
    public void readFrom(ObjectInput input) throws IOException, ClassNotFoundException {
       timeoutMillis = input.readLong();
-      chunk = MarshallUtil.unmarshallArray(input, XSiteState[]::new);
+      chunk = MarshallUtil.unmarshallArray(input, XSiteState::array);
```
danberindei: Here too, I think 0-length arrays are quite rare (except maybe in tests that have little to no data in each segment).

core/src/main/java/org/infinispan/commands/tx/PrepareCommand.java
rvansa: Apparently not, otherwise we would be hitting the ClassCastException more often. I could remove that as premature optimization; I have just replaced all usages.

pruivo:
where did you catch the exception? I don't expect to see empty arrays in the lines you changes.

rvansa:
I've used the MarshallUtil.unmarshallArray in my wip, where an empty list would be common. Therefore I've replaced all existing uses.

danberindei:
Please rebase for Jenkins

rvansa:
It really does not make sense to optimize the zero-array when we're not hitting that case regularly. Removed & rebased.

tristantarrant:
Pushed to master

PR #5507, opened 2017-10-09
ISPN-8379 Support configuration wildcards

tristantarrant:
https://issues.jboss.org/browse/ISPN-8379


galderz:
Looks good, couple of things:

* What happens if multiple cache templates match a cache name requested by the user? We should specify that.
* The documentation needs to indicate what happens when a user requests a cache that does not exist and it matches an existing template.

tristantarrant:
@galderz now it throws an exception when there are multiple wildcard matches. I've also added additional information about retrieving undefined caches.

galderz:
Looks good! Remember to blog after alpha1 goes out ;)

PR #5510, opened 2017-10-10
ISPN-8385 Create a reproducer for retrieving entries via container

vjuranek:
More off-heap container tests, reproducer for ISPN-8385


wburns:
Closing for now. PR noted for future.

PR #5513, opened 2017-10-12
ISPN-8069 Clustered Lock first implementation

karesti:
https://issues.jboss.org/browse/ISPN-8069

* Non reentrant, node owner  (the others modes may come in another PR after)
* Configuration cache and xml will be done when other configurations are supported
* Documentation is in progress
* Race conditions may exist
* The underlying cache is a replicated cache for high consistency
* Split brain and nodes leaving is handled but should be improved yet


karesti:
@belaban your review is more than welcome ! 😄

lock/src/main/java/org/infinispan/concurrent/api/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -0,0 +1,203 @@
...
+      @CacheEntryModified
+      public void entryModified(CacheEntryModifiedEvent event) {
+         ClusteredLockValue value = (ClusteredLockValue) event.getValue();
+         if (value.getState() == ClusteredLockState.RELEASED) {
+            LockRequestHolder nextRequestor = null;
+            while (!pendingRequests.isEmpty() && (nextRequestor == null || nextRequestor.isDone()))
+               nextRequestor = pendingRequests.poll();
+
+            final LockRequestHolder requestor = nextRequestor;
+            clusteredLockManager.execute(() -> lock(requestor));
```
pruivo: executor can be remove once https://issues.jboss.org/browse/ISPN-8336 is fixed

pruivo: There is a race condition here with tryLock() with timer.
when a lock is released, the timer can be triggered and complete the future while the lock-request is in the network or in progress.
The lock is be blocked by this node in the cache and it can only be released from a unlock invocation of this node.

pruivo:
API suggestion
* Add `ClusterLock.isLocked()` //return true if the locked is acquired by some "owner"
* Add `ClusterLock.amILockOwner()` //returns true if the node/instance/whatever is the lock owner

basic functional tests:
* 2 cache managers trying to create a lock with different configs (only one should succeed)
* 2 cache managers retrieve the same lock (ensure it is the same lock)
* 2 cache managers, ensure the lock state is seen between them

other functional tests
* check if a joiner doesn't reset the lock state (neither it is lost)
* check if a leaver (not the lock owner) doesn't reset the lock state
* check if the lock owner leave, the lock is released

Partition tests
* lock stays in minority partition => no one can lock it/release it
* lock stays in both partition (one copy in each partition), minority can't update. in majority
    * if locked, only the owner can unlock it (if it is in the majority)
        * otherwise the lock stay locked until the merge
    * if the lock is free, the majority can lock it
* lock stays in majority partition => majority can progress, minority don't

Concurrency and stress test
* Something like `ConditionalOperationsConcurrentTest`. multiple threads try to acquire the lock (`tryLock()`) and only one of them can succeed.
    * it can be extended to remaining method `lock()` and `tryLock(time, timeunit)`

karesti:
@pruivo concerning partition tests, I'm using a replicated cache. Are they still necessary in this case (beside the test that if a node leaves and the lock was acquired by this node, then it has to be released)

lock/src/main/java/org/infinispan/lock/api/ClusteredLockManager.java
```diff
@@ -0,0 +1,88 @@
...
+
+   /**
+    * Releases - or unlocks - a {@link ClusteredLock} if such exists.
+    * This method is used when we just want to release the lock no matter who is holding it at a given time.
+    * Calling this method may cause concurrency issues and has to be used in exception situations.
+    *
+    * @param name, the name of the lock
+    * @return {@code true} if the lock has been released
+    */
+   CompletableFuture<Boolean> release(String name);
```
rvansa: Could we call this `break`, to make sure people don't use it as `unlock`?

pruivo:
@karesti yes they are. If you have CP enabled (assuming you allow configure the lock as AP or CP), it would be nice to test if the partition doesn't release the lock (otherwise you will violate the consistency :))

lock/src/main/java/org/infinispan/lock/api/ClusteredLockManager.java
karesti: ok

karesti:
@rvansa which was the alternative to use the randomUUID ? 
@pruivo I added tests, probably more should be added. I have put a TODO in the ViewChange listener that should be fixed for rentrant locks, to avoid the following situation : View Changes, unlock is sent for a node that has leaved the cluster. Meahhinwhile this node is back and reacquires the lock. So the unlock from the view change will unlock even if the lock was correctly reacquired.

Another thing that we should consider is releasing "lock" calls after an amount of time, for example, 10 minutes. So the lock won't starve forever. In case of split-brain, all the "lock" requests should be released  too (or auto-release them with the scheduler after 5 minutes waiting).

karesti:
@wburns I asked for a review, if you have time, of course 💃

karesti:
Tested in https://github.com/vert-x3/vertx-infinispan/pull/39

rvansa:
@karesti Just use ThreadLocalRandom.current().nextLong() to generate 2 random longs and then `UUID(long, long)` ctor

tristantarrant:
Use Util.threadLocalRandomUUID()

lock/src/main/java/org/infinispan/lock/api/ClusteredLockManager.java
```diff
@@ -0,0 +1,88 @@
...
+import org.infinispan.lock.api.impl.exceptions.InfinispanLockException;
+
+/**
+ * Provides the API to define, create and remove ClusteredLocks.
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+@Experimental
+public interface ClusteredLockManager {
```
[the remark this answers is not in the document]
karesti: Ok, I will move to commons those interfaces !

lock/src/main/java/org/infinispan/lock/api/ClusteredLockManager.java
karesti: I renamed to `forceRelease`

lock/src/main/java/org/infinispan/lock/api/impl/lock/ClusteredLockImpl.java
```diff
@@ -0,0 +1,359 @@
...
+      }
+
+      public void handleLockResult(Boolean result, Throwable ex) {
+         if (ex != null) {
+            log.trace("Exception on lock request " + this, ex);
+            request.completeExceptionally(handleException(ex));
+            return;
+         }
+
+         if (request == null) {
```
pruivo: bug: should `result == null`

lock/src/main/java/org/infinispan/lock/api/impl/lock/ClusteredLockImpl.java
```diff
@@ -0,0 +1,359 @@
...
+   public CompletableFuture<Boolean> tryLock() {
+      log.trace("tryLock called from " + originator);
+      CompletableFuture<Boolean> tryLockRequest = new CompletableFuture<>();
+      tryLock(new TryLockRequestHolder(originator, tryLockRequest, false), false);
+      return tryLockRequest;
+   }
+
+   @Override
+   public CompletableFuture<Boolean> tryLock(long time, TimeUnit unit) {
+      log.trace("tryLock with timeout (" + time + "," + unit + ") called from " + originator);
```
pruivo: shouldn't invoke `tryLock()` if `time <= 0`?

lock/src/test/java/org/infinispan/lock/api/ClusteredLockSplitBrainTest.java
```diff
@@ -0,0 +1,140 @@
...
+      ClusteredLock lock0 = clm0.get(LOCK_NAME);
+      ClusteredLock lock1 = clm1.get(LOCK_NAME);
+      ClusteredLock lock2 = clm2.get(LOCK_NAME);
+      ClusteredLock lock3 = clm3.get(LOCK_NAME);
+
+      splitCluster(new int[]{0, 1, 2}, new int[]{3});
+
+      await(lock0.lock()
+            .thenRun(() -> {
+               assertTrue("we locked with node A", true);
```
pruivo: weird `assertTrue(true)`

lock/src/test/java/org/infinispan/lock/api/ClusteredLockSplitBrainTest.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.concurrent.CompletionException;
+
+import org.infinispan.lock.exception.ClusteredLockException;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.partitionhandling.AvailabilityException;
+import org.infinispan.partitionhandling.BasePartitionHandlingTest;
+import org.testng.annotations.Test;
+
+@Test(groups = "functional", testName = "clusteredLock.ClusteredLockSplitBrainTest")
+public class ClusteredLockSplitBrainTest extends BasePartitionHandlingTest {
```
pruivo: there are missing tests with splitting the cluster with lock acquired. 
also, it would be nice to test the `tryLock()`s

lock/src/main/java/org/infinispan/lock/api/impl/entries/ClusteredLockValue.java
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      @Override
+      public Integer getId() {
+         return ExternalizerIds.CLUSTERED_LOCK_VALUE;
+      }
+
+      @Override
+      public void writeObject(ObjectOutput output, ClusteredLockValue object) throws IOException {
+         output.writeObject(object.requestId);
```
pruivo: `MarshallUtil.marshallString(object.requestId, output);`

lock/src/main/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManager.java
```diff
@@ -0,0 +1,152 @@
...
+   @Override
+   public boolean defineLock(String name) {
+      return defineLock(name, new ClusteredLockConfiguration());
+   }
+
+   @Override
+   public boolean defineLock(String name, ClusteredLockConfiguration configuration) {
+      // TODO: Configuration is not used because we don't support any other mode for now. For that : ISPN-8413
+      CacheHolder cacheHolder = extractCacheHolder(cacheHolderFuture);
+      cache = cacheHolder.getClusteredLockCache();
```
pruivo: can throw `NullPointerException`

lock/src/main/java/org/infinispan/lock/api/impl/lock/ClusteredLockImpl.java
karesti: Indeed ! I could not reproduce this case in tests haha thank you pointing this out.

lock/src/main/java/org/infinispan/lock/api/impl/functions/LockFunction.java
```diff
@@ -0,0 +1,72 @@
...
+      }
+
+      @Override
+      public Integer getId() {
+         return ExternalizerIds.LOCK_FUNCTION;
+      }
+
+      @Override
+      public void writeObject(ObjectOutput output, LockFunction object) throws IOException {
+         output.writeObject(object.requestId);
```
pruivo: `MarshallUtil.marshallString(object.requestId, output);`

lock/src/main/java/org/infinispan/lock/api/impl/lock/ClusteredLockImpl.java
karesti: yes, indeed.

lock/src/test/java/org/infinispan/lock/api/ClusteredLockSplitBrainTest.java
karesti: I didn't want to go to more tests without a check on this one first.

lock/src/test/java/org/infinispan/lock/api/ClusteredLockSplitBrainTest.java
karesti: This is just a way to document, but I can change it

karesti: I will do something smarter

lock/src/main/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManager.java
karesti: if the cacheHolder is null you mean ?

lock/src/main/java/org/infinispan/lock/api/impl/lock/ClusteredLockImpl.java
```diff
@@ -0,0 +1,359 @@
...
+
+      public TryLockRequestHolder(Object requestor, CompletableFuture<Boolean> request, boolean wait) {
+         super(requestor, request);
+         this.wait = wait;
+      }
+
+      @Override
+      protected void handle(Boolean result) {
+         if (!wait) {
+            log.trace("Wait is " + wait + "on request" + this);
```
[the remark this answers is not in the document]
karesti: ok

lock/src/main/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManager.java
karesti: Ok, I'm throwing an exception if the caches could not be started instead of returning null

lock/src/main/java/org/infinispan/lock/api/impl/functions/LockFunction.java
karesti: ok

lock/src/main/java/org/infinispan/lock/api/impl/entries/ClusteredLockValue.java
karesti: ok

karesti:
@pruivo Concerning the There is another concurrency issue with the tryLock() with timer.

This test fails:

   public void testTryLockWithTimeout11() throws Throwable {
      assertTrue(await(lock.tryLock()));
      CompletableFuture<Boolean> tryLock = lock.tryLock(1, TimeUnit.NANOSECONDS);
      System.out.println("try-lock result=" + await(tryLock));
      System.out.println("is-locked result=" + await(lock.isLocked()));
      System.out.println("is-locked-by-me result=" + await(lock.isLockedByMe()));
      assertTrue(await(tryLock));
      assertTrue(await(lock.isLocked()));
      assertTrue(await(lock.isLockedByMe()));
   }
the tryLock() reports false but the lock is acquired.

try-lock result=false
is-locked result=true
is-locked-by-me result=true

This is working properly. The lock is not rentrant. The lock is locked by the owner, but the second tryLock fails because it is already acquired

pruivo:
@karesti there is only one `tryLock()`

lock/src/main/java/org/infinispan/lock/api/impl/lock/ClusteredLockImpl.java
```diff
@@ -0,0 +1,359 @@
...
+            return;
+         }
+
+         if (request == null) {
+            log.trace("Result is null on request " + this);
+            request.completeExceptionally(new ClusteredLockException("Lock result is null, something is wrong"));
+            return;
+         }
+
+         if (request.isDone()) {
```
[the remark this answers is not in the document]
karesti: I have to check the status at the end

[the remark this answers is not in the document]
karesti: and unlock if needed

lock/src/main/java/org/infinispan/lock/api/impl/lock/ClusteredLockImpl.java
```diff
@@ -0,0 +1,359 @@
...
+      }
+
+      @Override
+      protected void handle(Boolean result) {
+         if (!wait) {
+            log.trace("Wait is " + wait + "on request" + this);
+            // The answer has to be returned without holding the CompletableFuture
+            request.complete(result);
+         } else if (result) {
+            if (request.isDone()) {
```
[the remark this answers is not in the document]
karesti: @pruivo I did another thing, your comment's are isolated in a commit

karesti:
@pruivo @tristantarrant reported changes, if accepted, I will rebase before merge

commons/src/main/java/org/infinispan/lock/api/ClusteredLockConfiguration.java
```diff
@@ -0,0 +1,41 @@
+package org.infinispan.lock.api;
+
+import org.infinispan.commons.util.Experimental;
+
+/**
+ * A Clustered Lock can be rentrant and there are different ownership levels.
```
pruivo: typo `rentrant` => `reentrant`

commons/src/main/java/org/infinispan/lock/api/ClusteredLockConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+ * <p>
+ * The only mode supported now is "non rentrant" locks for "nodes".
```
pruivo: typo `rentrant` => `reentrant`

lock/src/main/java/org/infinispan/lock/api/impl/lock/ClusteredLockImpl.java
```diff
@@ -0,0 +1,369 @@
...
+      readWriteMap.eval(lockKey, new LockFunction(requestHolder.requestId, requestHolder.requestor)).whenComplete((lockResult, ex) -> {
+         requestHolder.handleLockResult(lockResult, ex);
+      });
+   }
+
+   @Override
+   public CompletableFuture<Void> unlock() {
+      log.tracef("unlock called from %s", originator);
+      CompletableFuture<Void> unlockRequest = new CompletableFuture<>();
+      readWriteMap.eval(lockKey, new UnlockFunction(originator)).whenComplete((lockResult, ex) -> {
```
rvansa: This is not fluent, capturing another future... could you use `.exceptionally(t -> throw handleException(t))` instead of `.whenComplete` (maybe with `CompletionException` wrapping)?

commons/src/main/java/org/infinispan/lock/api/ClusteredLockConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @see <a href="http://infinispan.org/documentation/">Infinispan documentation</a>
+ * @since 9.2
+ */
+@Experimental
+public class ClusteredLockConfiguration {
+   private final OwnershipLevel ownershipLevel; // default NODE
+   private final boolean rentrant; // default false
```
pruivo: typo `rentrant` => `reentrant`

commons/src/main/java/org/infinispan/lock/api/ClusteredLockConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+
+   /**
+    * Default lock is non rentrant and the ownershop level is {@link OwnershipLevel#NODE}
```
pruivo: typo `rentrant` => `reentrant` 
typo `ownershop` => `ownership`

commons/src/main/java/org/infinispan/lock/api/ClusteredLockConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+    */
+   public ClusteredLockConfiguration() {
+      this.ownershipLevel = OwnershipLevel.NODE;
+      this.rentrant = false;
+   }
+
+   /**
+    * @return true if the lock is rentrant
```
pruivo: typo `rentrant` => `reentrant`
ps. maybe a find a replace will do the trick so I don't have to point them all :)

lock/src/main/java/org/infinispan/lock/api/impl/log/Log.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Range: 29001 - 30000
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+@MessageLogger(projectCode = "ISPN")
+public interface Log extends org.infinispan.util.logging.Log {
+   public static final String LOCK_DELETE_MSG = "The lock was deleted.";
```
rvansa: Other modules' logging code does not extract the messages into static vars.
Btw. name of the lock in error message could be useful for debugging.

lock/src/main/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManager.java
```diff
@@ -0,0 +1,152 @@
...
+      AdvancedCache<ClusteredLockKey, ClusteredLockValue> clusteredLockCache = cacheHolder.getClusteredLockCache();
+      return clusteredLockCache.removeAsync(new ClusteredLockKey(ByteString.fromString(name))).thenApply(value -> {
+         locks.remove(name);
+         return value != null;
+      });
+   }
+
+   @Override
+   public CompletableFuture<Boolean> forceRelease(String name) {
+      return CompletableFuture.supplyAsync(() -> {
```
rvansa: You could use functional API which is async by nature instead of executing this in FJP:
```
rwmap.eval(new ClusteredLockKey(ByteString.fromString(name)), view -> if (view.find().isPresent()) {
   boolean retval = view.find().getState() == ClusteredLockState.RELEASED;
   view.set(new ClusteredLockValue());
   return retval;
} else {
   return false;
});
```

lock/src/test/java/org/infinispan/lock/api/impl/lock/ClusteredLockImplTest.java
```diff
@@ -0,0 +1,181 @@
...
+   }
+
+   public void testTryLockWithZeroTimeout() throws Throwable {
+      assertTrue(await(lock.tryLock(0, TimeUnit.HOURS)));
+      assertFalse(await(lock.tryLock(0, TimeUnit.HOURS)));
+      assertTrue(await(lock.isLockedByMe()));
+      assertTrue(await(lock.isLocked()));
+   }
+
+   public void testTryLockWithNegatifTimeout() throws Throwable {
```
pruivo: typo `Negatif` => `Negative`

lock/src/test/java/org/infinispan/lock/api/ClusteredLockKillingNodesTest.java
```diff
@@ -0,0 +1,42 @@
...
+   }
+
+   @AfterClass(alwaysRun = true)
+   protected void destroyLock() {
+      ClusteredLockManager clusteredLockManager = clusteredLockManager(0);
+      clusteredLockManager.remove(LOCK_NAME);
+   }
+
+   @Test
+   public void testLockWithAquisitonAndKill() throws Throwable {
```
pruivo: typo `Aquisiton` => `Acquisition`

commons/src/main/java/org/infinispan/lock/api/ClusteredLockManager.java
```diff
@@ -0,0 +1,88 @@
...
+    * @param name, the lock name
+    * @return {@code true} if this lock is defined
+    */
+   boolean isDefined(String name);
+
+   /**
+    * Removes a {@link ClusteredLock} if such exists.
+    *
+    * @param name, the name of the lock
+    * @retun {@code true} if the lock is removed
```
pruivo: typo `retun` => `return`

lock/src/main/java/org/infinispan/lock/api/impl/ClusteredLockMetadataFileFinder.java
```diff
@@ -0,0 +1,19 @@
+package org.infinispan.lock.api.impl;
+
+import org.infinispan.factories.components.ModuleMetadataFileFinder;
+import org.kohsuke.MetaInfServices;
+
+/**
+ * Metatada file name, needed for dependency injection.
```
pruivo: typo: `Metatada` => `Metadata`

core/src/main/java/org/infinispan/util/UserRaisedFunctionalException.java
```diff
@@ -14,6 +14,10 @@ public UserRaisedFunctionalException(Throwable cause) {
       super(cause);
    }
 
+   public UserRaisedFunctionalException(String cause) {
```
pruivo: not used. can be removed.

lock/src/main/java/org/infinispan/lock/api/impl/ClusteredLockModuleLifecycle.java
```diff
@@ -0,0 +1,103 @@
...
+import org.kohsuke.MetaInfServices;
+
+/**
+ * Locks module configuration
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+@MetaInfServices(value = ModuleLifecycle.class)
+public class ClusteredLockModuleLifecycle extends AbstractModuleLifecycle {
```
pruivo: `AbstractModuleLifecycle` is deprecated! use `ModuleLifecycle` instead.

lock/src/main/java/org/infinispan/lock/api/impl/manager/CacheHolder.java
```diff
@@ -0,0 +1,19 @@
+package org.infinispan.lock.api.impl.manager;
+
+import org.infinispan.AdvancedCache;
+import org.infinispan.lock.api.impl.entries.ClusteredLockKey;
+import org.infinispan.lock.api.impl.entries.ClusteredLockValue;
+import org.infinispan.context.Flag;
+
+public class CacheHolder {
```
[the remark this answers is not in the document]
karesti: We are going to mutualise this code with counters after the merge. I inspire myself from the code having that in ming, and this CacheHolder was used in that way there,  even if it might have changed a bit since I did it. For this part, I will relook at it when we will have a 'commons' module for 'clustered data structures' : https://issues.jboss.org/browse/ISPN-8393

lock/src/main/java/org/infinispan/lock/api/impl/log/Log.java
pruivo: `public static final` is redundant in interfaces.

lock/src/main/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManager.java
```diff
@@ -0,0 +1,152 @@
...
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+public class EmbeddedClusteredLockManager implements ClusteredLockManager {
+
+   private static final long WAIT_CACHES_TIMEOUT = TimeUnit.SECONDS.toNanos(15);
+   private static final Log log = LogFactory.getLog(EmbeddedClusteredLockManager.class, Log.class);
+
+   private CompletableFuture<CacheHolder> cacheHolderFuture;
+   private ConcurrentHashMap<String, ClusteredLock> locks = new ConcurrentHashMap();
```
pruivo: `<>` missing

lock/src/main/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManager.java
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+public class EmbeddedClusteredLockManager implements ClusteredLockManager {
+
+   private static final long WAIT_CACHES_TIMEOUT = TimeUnit.SECONDS.toNanos(15);
+   private static final Log log = LogFactory.getLog(EmbeddedClusteredLockManager.class, Log.class);
+
+   private CompletableFuture<CacheHolder> cacheHolderFuture;
```
pruivo: can be final

lock/src/main/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManager.java
pruivo: can be final

lock/src/main/java/org/infinispan/lock/api/impl/entries/ClusteredLockValue.java
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Lock object inside the cache. Holds the lock owner, the lock request id and the status of the lock.
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+public class ClusteredLockValue {
+
+   public static final AdvancedExternalizer<ClusteredLockValue> EXTERNALIZER = new Externalizer();
+   private String requestId;
```
pruivo: all fields can be final

commons/src/main/java/org/infinispan/lock/api/ClusteredLock.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * ClusteredLock is a data structure used for concurrent programming between Infinispan instances in cluster mode.
+ *
+ * A typical usage idiom for {@link ClusteredLock#lock} will be :
+ *
+ * {@code
+ *
+ *  EmbeddedCache cm = ...;
+ *  ClusteredLockManager cclm = ClusteredLockManagerFactory.from(cm);
```
pruivo: what does `cclm` suppose to mean?

lock/src/test/java/org/infinispan/lock/api/ClusteredLockWith2NodesKillingOneTest.java
```diff
@@ -0,0 +1,56 @@
...
+   }
+
+   @AfterClass(alwaysRun = true)
+   protected void destroyLock() {
+      ClusteredLockManager clusteredLockManager = clusteredLockManager(1);
+      clusteredLockManager.remove(LOCK_NAME);
+   }
+
+   @Test
+   public void testLockWithAquisitonAndKill() throws Throwable {
```
pruivo: typo `Aquisiton` => `Acquisition`

lock/src/test/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManagerTest.java
```diff
@@ -0,0 +1,89 @@
...
+      assertEquals(OwnershipLevel.NODE, configuration.getOwnershipLevel());
+      assertFalse(configuration.isRentrant());
+   }
+
+   public void testGetConfigurationWithoutLockDefinition() {
+      ClusteredLockManager clusteredLockManager = clusteredLockManager(0);
+      expectException(ClusteredLockException.class, () -> clusteredLockManager.getConfiguration("notDefined"));
+   }
+
+   public void testIsDefinied() {
```
pruivo: typo `Definied` => `Defined`

lock/src/test/java/org/infinispan/lock/api/ClusteredLockTest.java
```diff
@@ -0,0 +1,210 @@
...
+      await(lock2Request
+            .exceptionally(e -> {
+               assertException(ClusteredLockException.class, e);
+               assertTrue(e.getMessage().contains(Log.LOCK_DELETE_MSG));
+               return null;
+            }));
+   }
+
+   @Test
+   public void testLockWithCountersInParalel() throws Throwable {
```
pruivo: typo `Paralel` => `Parallel`

lock/src/test/java/org/infinispan/lock/api/ClusteredLockTest.java
```diff
@@ -0,0 +1,210 @@
...
+      await(CompletableFuture.allOf(
+            lock.lock().thenRun(new Counter(counter, 1, 1000)).whenComplete((nil, t) -> lock.unlock()),
+            lock.lock().thenRun(new Counter(counter, -1, 1000)).whenComplete((nil, t) -> lock.unlock()),
+            lock.lock().thenRun(new Counter(counter, 2, 500)).whenComplete((nil, t) -> lock.unlock())));
+
+      assertEquals(counter.get(), 2);
+   }
+
+   @Test
+   public void testTryLockWithCountersInParalel() throws Throwable {
```
pruivo: typo `Paralel` => `Parallel`

lock/src/main/java/org/infinispan/lock/api/impl/lock/ClusteredLockImpl.java
```diff
@@ -0,0 +1,369 @@
...
+   @Listener
+   class ClusterChangeListener {
+
+      @ViewChanged
+      public void viewChange(ViewChangedEvent event) {
+         log.trace("viewChange event has been fired");
+         List<Address> newMembers = event.getNewMembers();
+         List<Address> oldMembers = event.getOldMembers();
+
+         if(newMembers.size() > 1 || (oldMembers.size() == 2 && newMembers.size() == 1)) {
```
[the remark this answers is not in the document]
karesti: It's because of the split brain vs one node is leaving. this part is a bit complicated now. 
The thing is that when I have a split brain or when a node just joins or reaves because of normal reasons, I get this event fired.
I just avoid in the case of an split brain where the partition is a single node, to fire unlocks for nothing. But if I have 2 nodes and one leaves, well, I want that other node to unlock even if the cluster is a single node now.
I can improve this, adding some more tests for that cases.

lock/src/main/java/org/infinispan/lock/api/impl/manager/CacheHolder.java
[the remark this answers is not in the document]
rvansa: I meant that you can already inline all uses of that wrapper.

lock/src/main/java/org/infinispan/lock/api/impl/functions/UnlockFunction.java
```diff
@@ -0,0 +1,86 @@
...
+      this.requestor = requestor;
+   }
+
+   @Override
+   public Void apply(EntryView.ReadWriteEntryView<ClusteredLockKey, ClusteredLockValue> entryView) {
+      ClusteredLockValue lockValue = entryView.find().orElseThrow(() -> log.lockDeleted());
+      boolean requestIdMatches = requestId == null || (lockValue.getRequestId() != null && lockValue.getRequestId().equals(requestId));
+      boolean ownerMatches = lockValue.getOwner() != null && lockValue.getOwner().equals(requestor);
+      if (requestIdMatches && ownerMatches) {
+         entryView.set(new ClusteredLockValue());
```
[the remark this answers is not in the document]
karesti: Because there is no way to know if a lock is already defined if I remove the lock. I want the lock to be there, and handle a state. Ok for the singleton point.

lock/src/main/java/org/infinispan/lock/api/impl/functions/IsLocked.java
```diff
@@ -0,0 +1,75 @@
...
+   }
+
+   public IsLocked(Object requestor) {
+      this.requestor = requestor;
+   }
+
+   @Override
+   public Boolean apply(EntryView.ReadWriteEntryView<ClusteredLockKey, ClusteredLockValue> entryView) {
+      ClusteredLockValue lock = entryView.find().orElseThrow(() -> log.lockDeleted());
+      Boolean result = Boolean.FALSE;
```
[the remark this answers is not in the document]
karesti: ok

lock/src/main/java/org/infinispan/lock/api/impl/lock/ClusteredLockImpl.java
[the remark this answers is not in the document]
rvansa: You should not try to detect split brains ahead, that's a task for other components.
I was asking about two, because with more owners it should work well even with e.g. 2 nodes crashing/leaving simultaneously. The cache is replicated after all.

lock/src/main/java/org/infinispan/lock/api/impl/entries/ClusteredLockValue.java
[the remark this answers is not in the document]
karesti: This is a protection for unlocks that is not used still because the problem is more something needed on reentrant locks. But I left it there because it helps for debug the different requests done by the same node.
The problem I imagined was : a node acquieres the lock. the node leaves. the unlock for that node is raised. meanwhile the node is back and reacquires the lock. the unlock happens. In case of reentrant locks, if the request that actually holds the lock is not the one that was supposed to be unlocked, we should not unlock. But this is not an issue for non reentrant locks, because if the unlock did not happen, and the node tries to lock, it won't be able to lock and will bloc.

lock/src/main/java/org/infinispan/lock/api/impl/entries/ClusteredLockState.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+public enum ClusteredLockState {
+   ACQUIRED,
+   RELEASED;
+
+   private static final ClusteredLockState[] CACHED_VALUES = ClusteredLockState.values();
+   public static final AdvancedExternalizer<ClusteredLockState> EXTERNALIZER = new ClusteredLockState.Externalizer();
```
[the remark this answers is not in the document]
karesti: Ok for the marshaller, but I'm not going to put boolean true/false, at least not for now if you don't mind

lock/src/main/java/org/infinispan/lock/api/impl/log/Log.java
karesti: yes, i forgot i was in a interface haha

lock/src/main/java/org/infinispan/lock/api/impl/ClusteredLockMetadataFileFinder.java
karesti: Metatada is much more fun Pedro ! 👍 thanks for all the typo reviews !

commons/src/main/java/org/infinispan/lock/api/ClusteredLock.java
karesti: clm

commons/src/main/java/org/infinispan/lock/api/ClusteredLockConfiguration.java
karesti: ok

commons/src/main/java/org/infinispan/lock/api/ClusteredLockConfiguration.java
karesti: ok

commons/src/main/java/org/infinispan/lock/api/ClusteredLockConfiguration.java
karesti: ok

commons/src/main/java/org/infinispan/lock/api/ClusteredLockConfiguration.java
karesti: ok

commons/src/main/java/org/infinispan/lock/api/ClusteredLockConfiguration.java
karesti: ok

commons/src/main/java/org/infinispan/lock/api/ClusteredLockManager.java
karesti: ok

core/src/main/java/org/infinispan/util/UserRaisedFunctionalException.java
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,212 @@
...
+failure.
+
+* `defineLock(String name, ClusteredLockConfiguration configuration)` :
+Defines a lock with the specific name and `ClusteredLockConfiguration`. It does not overwrite existing
+configurations. Returns true if successfully defined or false if the lock is already defined or any other
+failure.
+
+* `ClusteredLock get(String name)` :
+Get’s a `ClusteredLock` by it’s name. This method throws `ClusteredLockException` if the lock is not
+not defined. A call of {@link #defineLock} must be done at least once in the cluster. This method will return the
```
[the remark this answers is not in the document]
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,212 @@
...
+same lock object depending on the {@link OwnershipLevel}. If the `OwnershipLevel` is NODE, it wll return the same
```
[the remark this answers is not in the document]
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,212 @@
...
+instance per `ClusteredLockManager`.
+If the `OwnershipLevel` is `INSTANCE`, it wll return the new instance per call, so it may exist multiple instances that
+my acquire the lock per `ClusteredLockManager`.
+
+Currently, the only mode supported is NODE.
```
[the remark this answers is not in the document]
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,212 @@
...
+the lock is not not defined. A call of `defineLock` must be done at least once in the cluster.
+
+* `boolean isDefined(String name)` :
+Checks if a lock is already defined.
+
+* `CompletableFuture<Boolean> remove(String name)` :
+Removes a `ClusteredLock` if such exists.
+
+* `CompletableFuture<Boolean> forceRelease(String name)` :
+Releases - or unlocks - a `link ClusteredLock` if such exists.
```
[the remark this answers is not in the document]
karesti: ok

lock/src/main/java/org/infinispan/lock/api/impl/ClusteredLockModuleLifecycle.java
karesti: ok

lock/src/main/java/org/infinispan/lock/api/impl/entries/ClusteredLockValue.java
karesti: ok

lock/src/main/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManager.java
karesti: ok

lock/src/main/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManager.java
karesti: ok

lock/src/test/java/org/infinispan/lock/api/ClusteredLockKillingNodesTest.java
karesti: ok

lock/src/test/java/org/infinispan/lock/api/ClusteredLockTest.java
karesti: ok

lock/src/test/java/org/infinispan/lock/api/impl/lock/ClusteredLockImplTest.java
karesti: ok

lock/src/test/java/org/infinispan/lock/api/impl/manager/EmbeddedClusteredLockManagerTest.java
karesti: ok

karesti:
@pruivo I reported all. Intellij helps with typos, but somehow I did not pay enough attention to underlying green ! Thanks for pointing, I will make more attention next time

@rvansa I reported most of your comments too. I'm going to create a specific jira to properly treat all the state transfer / kill nodes situations, if it's too big to handle it in the other issue I have to support reentrancy. Same applies to exceptions messages and all that

karesti:
@pruivo @rvansa if the build passes and you do agree, I would like to rebase all the commits and someone to merge this, please

commons/src/main/java/org/infinispan/lock/api/ClusteredLock.java
```diff
@@ -0,0 +1,118 @@
...
+ * @since 9.2
+ */
+public interface ClusteredLock {
+
+   /**
+    * Acquires the lock. If the lock is not available then the {@link CompletableFuture} waits til the lock has been acquired.
+    * Currently, there is no maximum time specified for a lock request to fail, so this could cause thread starvation.
+    *
+    * @return a completed {@link CompletableFuture} when the lock is acquired
+    * @throws {@link ClusteredLockException} when the lock does not exist
```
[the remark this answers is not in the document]
karesti: ok

danberindei:
@karesti http://ci.infinispan.org/job/Infinispan/job/PR-5513/33/testReport/junit/org.infinispan.lock/ClusteredLockSplitBrainTest/testLockCreationWhenPartitionHappening_DIST_SYNC_/

karesti:
@danberindei yes, I know, split brain fails in jenkins. I have to check tests and code ... but I'm in another fight before that !!

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+In order to start using the clustered locks, you needs to add the dependency in your Maven `pom.xml` file:
+
+.pom.xml
+[source,xml]
+----
+<dependency>
+   <groupId>org.infinispan</groupId>
+   <artifactId>infinispan-clustered-lock</artifactId>
+   <version>...</version> <!-- 9.2.0.Final or higher -->
```
danberindei: @tristantarrant Maybe we should have a placeholder like `{infinispanversion}` but with the full version?

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+node can release the lock at a given time.
+
+* `INSTANCE` - not yet supported
+
+When a `ClusteredLock` is defined, this lock can be used from all the nodes in the infinispan cluster.
+When the ownership is INSTANCE type, this means that the owner of the lock is the actual instance we acquired when
+`ClusteredLockManager.get("lockName")` is called.
+
+This means that each time we get a `ClusteredLock` instance with the `ClusteredCacheManager`, this instance will be
+a new instance.
```
danberindei: Sorry I'm commenting so late, but I find the change in the `ClusteredLockManager.get(name)` behavior confusing.

Personally I like the vert.x API, where acquiring the lock gives you a `Lock` instance, and you can only release it through that `Lock` instance. Being able to release a lock that some other application component has acquired without any communication with that component sounds wrong to me.

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+Currently, only non reentrant locks are supported. This means that when two consecutive `lock` calls are sent for the same
+owner, the first call will acquire the lock if it's available, and the second call will block.
+
+
+=== The ClusteredLockManager interface.
+
+The `ClusteredLockManager` interface, *marked as experimental*, is the entry point to define, retrieve and remove a lock.
+It automatically listen to the creation of `EmbeddedCacheManager` and proceeds with the registration  of an
+instance of it per `EmbeddedCacheManager`.
```
danberindei: Does the interface really listen to events? And should the user care if the lock manager is created at the same time as the cache manager, or only on the first call to `EmbeddedClusteredLockManagerFactory.from(EmbeddedCacheManager)`?

lock/src/main/java/org/infinispan/lock/impl/ClusteredLockModuleLifecycle.java
```diff
@@ -0,0 +1,100 @@
...
+            .stateTransfer().fetchInMemoryState(true)
+            .partitionHandling().whenSplit(PartitionHandling.DENY_READ_WRITES)
+            .transaction().transactionMode(TransactionMode.NON_TRANSACTIONAL);
+
+      return builder.build();
+   }
+
+   private static CompletableFuture<CacheHolder> startCaches(EmbeddedCacheManager cacheManager) {
+      final CompletableFuture<CacheHolder> future = new CompletableFuture<>();
+      new Thread(() -> {
```
danberindei: Please set a thread name that includes the node name for easier debugging.

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: Don't worry, input is always welcome Dan ! So as I understand, the ownership level "NODE" is not ok and we should support "INSTANCE" level locking (rentrant or not rentrant) ?

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: This is not very clear, indeed, I have to rework this part.

lock/src/main/java/org/infinispan/lock/impl/ClusteredLockModuleLifecycle.java
karesti: Ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: About this point, @rvansa considered that we might only need "object" level (at leats for now). We discussed that briefly during the F2F

danberindei: First of all, I'm not enunciating rules here, I'm just thinking aloud :) But I think even with the vert.x API, node-ownership is still be useful if the lock is reentrant: that way multiple logical threads on the same node can acquire the lock concurrently, and only the final release actually allows another node to acquire it.

Thinking about it some more, maybe we should allow the application to use its own lock owner, as in `clusteredLock.lock(lockOwner)` or `clusteredLock.withOwner(lockOwner).lock()` (to match the not-yet-public API in Cache). That way, different application components that deal with a single request could call `get(name).lock(requestId)` without the need to explicitly pass a lock instance between them. Regular `lock()` would use the local node as the lock owner.

rvansa: @karesti It's you who came with the notion of node-level locks; I haven't seen that in Vert.X but the fact that the lock is not tied to a thread. To me a node-level ownership with non-reentrancy seems the same as instance-level. 

@danberindei Sounds good, but the owner should be provided when the lock is retrieved from manager rather than after that (to enforce const owner by api).

karesti: @rvansa it's not accurate to say that was me who came to that idea of ownership or reentrancy levels, but this is not the discussion here.
I think there is a misunderstanding here and probably is due to something I did not fully get

karesti: @danberindei concerning the "with owner" this is something I was considering too

commons/src/main/java/org/infinispan/lock/api/ClusteredLock.java
```diff
@@ -0,0 +1,112 @@
...
+ * }
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @see <a href="http://infinispan.org/documentation/">Infinispan documentation</a>
+ * @since 9.2
+ */
+public interface ClusteredLock {
+
+   /**
+    * Acquires the lock. If the lock is not available then the {@link CompletableFuture} waits til the lock has been acquired.
```
tristantarrant: s/til/until/

commons/src/main/java/org/infinispan/lock/api/ClusteredLock.java
```diff
@@ -0,0 +1,112 @@
...
+    * If the lock is not available then this method will return immediately with the {@link CompletableFuture} holding the value {@code false}.
+    *
+    * @return {@code CompletableFuture(true)} if the lock was acquired and {@code CompletableFuture(false)} otherwise
+    * @throws ClusteredLockException when the lock does not exist
+    */
+   CompletableFuture<Boolean> tryLock();
+
+   /**
+    * If the lock is available this method returns immediately with the {@link CompletableFuture} holding the value {@code true}.
+    * If the lock is not available then the {@link CompletableFuture} waits til :
```
tristantarrant: s/til/until/

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
+== Clustered Lock
+
+Clustered lock is a new building block introduced in Infinispan 9.2
+It is a lock distributed and shared among all nodes in the Infinispan cluster and currently provides a way to execute
```
tristantarrant: "It is a lock which is distributed and shared..."

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
tristantarrant: Indeed. @karesti please add a `<fullinfinispanversion>${project.version}</fullinfinispanversion>` to `documentation/pom.xml` in the `<attributes>` section so that we can substitute it correctly when docs are generated

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
tristantarrant: +1 for withOwner

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+This means that each time we get a `ClusteredLock` instance with the `ClusteredCacheManager`, this instance will be
+a new instance.
+This lock can be used to synchronize code between infinispan nodes and inside each infinispan node.
+The advantage of this lock is that only the instance that called 'lock' can release the lock.
+
+
+==== Reentrancy
+
+When a `ClusteredLock` is configured reentrant, this means the owner of the lock can reacquire the lock as many consecutive
```
tristantarrant: s/this means//

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+Currently, only non reentrant locks are supported. This means that when two consecutive `lock` calls are sent for the same
+owner, the first call will acquire the lock if it's available, and the second call will block.
+
+
+=== The ClusteredLockManager interface.
+
+The `ClusteredLockManager` interface, *marked as experimental*, is the entry point to define, retrieve and remove a lock.
+It automatically listen to the creation of `EmbeddedCacheManager` and proceeds with the registration  of an
+instance of it per `EmbeddedCacheManager`.
+It starts the caches needed to store the lock state.
```
tristantarrant: "It starts the internal caches needed..."

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+   boolean isDefined(String name);
+
+   CompletableFuture<Boolean> remove(String name);
+
+   CompletableFuture<Boolean> forceRelease(String name);
+}
+----
+
+* `defineLock` :
+Defines a lock with the specific name and the default `ClusteredLockConfiguration`. It does not overwrite
```
tristantarrant: s/specific/specified/

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+existing configurations.
+
+* `defineLock(String name, ClusteredLockConfiguration configuration)` :
+Defines a lock with the specific name and `ClusteredLockConfiguration`. It does not overwrite existing
```
tristantarrant: s/specific/specified/

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+configurations.
+
+* `ClusteredLock get(String name)` :
+Get’s a `ClusteredLock` by it’s name. A call of `defineLock` must be done at least once in the cluster.
+See ownership level section to understand the implications of `get` method call.
```
tristantarrant: Provide an anchor to the section you mention

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+Removes a `ClusteredLock` if such exists.
+
+* `CompletableFuture<Boolean> forceRelease(String name)` :
+Releases - or unlocks - a `ClusteredLock`, if such exists, no matter who is holding it at a given time.
+Calling this method may cause concurrency issues and has to be used in *exceptional situations*.
+
+
+=== The Clustered Lock
+
+`ClusteredLock` interface, *marked as experimental*, is the interface that implement the clustered locks.
```
tristantarrant: s/implement/implements/

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+   CompletableFuture<Boolean> isLocked();
+
+   CompletableFuture<Boolean> isLockedByMe();
+}
+
+----
+
+* `lock` :
+Acquires the lock. If the lock is not available then call blocks til the lock is acquired.
```
tristantarrant: s/til/until/

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+Currently, *there is no maximum time specified for a lock request to fail*, so this could cause thread starvation.
+
+* `tryLock`
+Acquires the lock only if it is free at the time of invocation, and returns `true` in that case. This method does not
+block (or wait) for any lock acquisition.
+
+* `tryLock(long time, TimeUnit unit)`
+If the lock is available this method returns immediately with `true`.
+If the lock is not available then the call waits til :
```
tristantarrant: s/til/until/

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+* `unlock`
+
+Releases the lock. Only the holder of the lock may release the lock.
+
+* `isLocked`
+Returns `true` when the lock is locked and `false` when the lock is released.
+
+* `isLockedByMe`
+Returns a `true` when the lock is owned by the caller and `false` when the lock is owned by someone else or it's released.
```
tristantarrant: s/Returns a/Returns/

lock/src/main/java/org/infinispan/lock/impl/ClusteredLockModuleLifecycle.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Locks module configuration
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+@MetaInfServices(value = ModuleLifecycle.class)
+public class ClusteredLockModuleLifecycle implements ModuleLifecycle {
+
+   public static final String CLUSTERED_LOCK_CACHE_NAME = "___clustered_locks";
```
tristantarrant: Please use org.infinispan.LOCKS

lock/src/test/java/org/infinispan/lock/ClusteredLockWith2NodesKillingOneTest.java
```diff
@@ -0,0 +1,59 @@
+package org.infinispan.lock;
+
+import static org.infinispan.functional.FunctionalTestUtils.await;
+import static org.infinispan.test.TestingUtil.killCacheManagers;
+import static org.testng.Assert.assertEquals;
```
tristantarrant: We tend to favor the AssertJUnit versions of these

karesti: Wrong import by mistake. Let me change and push

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: ok

commons/src/main/java/org/infinispan/lock/api/ClusteredLock.java
karesti: ok

commons/src/main/java/org/infinispan/lock/api/ClusteredLock.java
karesti: ok

documentation/src/main/asciidoc/user_guide/clustered_lock.adoc
karesti: but I don't want this value to be replaced each time. This is something from 9.2.0.Final. I did the same thing as in counters where it is said 
I will do this in a separate PR you want

karesti:
@tristantarrant I reported your modifications, minus the substitution of the version at least for now. if jenkins is ok, I would like to merge

karesti:
@tristantarrant jenkins seems ok !

tristantarrant:
Merged. Thanks !

karesti:
@tristantarrant thank you 😭

PR #5516, opened 2017-10-16
ISPN-8090 Functional commands don't support Data conversions In Transactional Caches

rvansa:
* Remove some generics in functional commands that don't make sense;
  the function generics no longer match to the stored keys and values
* Add DataConversion to Mutations
* Use DataConversion.writeTo/readFrom directly

https://issues.jboss.org/browse/ISPN-8090

### DIFF
--- a/core/src/main/java/org/infinispan/commands/functional/Mutation.java
+++ b/core/src/main/java/org/infinispan/commands/functional/Mutation.java
@@ -1,11 +1,13 @@
 package org.infinispan.commands.functional;
 
+import org.infinispan.commands.functional.functions.InjectableComponent;
+import org.infinispan.encoding.DataConversion;
 import org.infinispan.functional.EntryView;
 
 /**
  * Simplified version of functional command used for read-only operations after transactional modifications.
  */
-public interface Mutation<K, V, R> {
+public interface Mutation<K, V, R> extends InjectableComponent {
 
    /**
     * @return Internal identifier used for purposes of marshalling
@@ -14,8 +16,11 @@ public interface Mutation<K, V, R> {
 
    /**
     * Mutate the view
-    *
-    * @param view
-    */
+    *  @param view
+    * */
    R apply(EntryView.ReadWriteEntryView<K, V> view);
+
+   DataConversion keyDataConversion();
+
+   DataConversion valueDataConversion();
 }
--- a/core/src/main/java/org/infinispan/commands/functional/Mutations.java
+++ b/core/src/main/java/org/infinispan/commands/functional/Mutations.java
@@ -8,7 +8,12 @@ import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
+import org.infinispan.configuration.cache.Configuration;
+import org.infinispan.configuration.global.GlobalConfiguration;
+import org.infinispan.encoding.DataConversion;
+import org.infinispan.factories.ComponentRegistry;
 import org.infinispan.functional.EntryView;
+import org.infinispan.marshall.core.EncoderRegistry;
 
 /**
  * Helper class for marshalling, also hiding implementations of {@link Mutation} from the interface.
@@ -18,6 +23,9 @@ final class Mutations {
 
    // No need to occupy externalizer ids when we have a limited set of options
    static <K, V, R> void writeTo(ObjectOutput output, Mutation<K, V, R> mutation) throws IOException {
+      BaseMutation bm = (BaseMutation) mutation;
+      DataConversion.writeTo(output, bm.keyDataConversion);
+      DataConversion.writeTo(output, bm.valueDataConversion);
       byte type = mutation.type();
       output.writeByte(type);
       switch (type) {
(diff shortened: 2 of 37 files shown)


core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -1328,6 +1328,15 @@ private void processResponse(org.jgroups.Address src, byte[] buffer, int offset,
          requests.addResponse(requestId, address, response);
       } catch (Throwable t) {
          log.errorProcessingResponse(requestId, src);
+         Exception e = t instanceof Exception ? ((Exception) t) : new CacheException(t);
+         try {
+            marshaller.objectFromByteBuffer(buffer, offset, length);
+         } catch (IOException e1) {
+            e1.printStackTrace();  // TODO: Customise this generated block
```
karesti: Don't forget the TODOs here !

core/src/main/java/org/infinispan/commands/functional/Mutations.java
```diff
@@ -18,46 +20,72 @@ private Mutations() {}
...
       byte type = mutation.type();
       output.writeByte(type);
       switch (type) {
          case ReadWrite.TYPE:
             output.writeObject(((ReadWrite<K, V, ?>) mutation).f);
             break;
          case ReadWriteWithValue.TYPE:
-            output.writeObject(((ReadWriteWithValue<K, V, R>) mutation).value);
-            output.writeObject(((ReadWriteWithValue<K, V, R>) mutation).f);
+            ReadWriteWithValue<K, V, R> rwwv = (ReadWriteWithValue<K, V, R>) mutation;
```
karesti: Is there any way to refactor the code so we just need to do an if ? instead of checking every type to write f everytime etc ...

karesti:
@rvansa checkstyle seems ko

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
rvansa: Ouch, thanks - this was a debugging code ofc :-/

core/src/main/java/org/infinispan/commands/functional/Mutations.java
rvansa: Do you want me to remove the switch? I  could add `writeTo` method to each mutation, instead of calling `.type()`, but I prefer to keep the writeTo/readFrom symmetric (and readFrom cannot be refactored that way).

karesti: It's not a big deal for me, code seemed to me very similar in the switch so I suggested a change, but do whatever you think is better/homogenous for this case.

rvansa:
Wait a sec, I've noticed that I've omitted few places..

rvansa:
Ok, I've fixed several other generics (mostly the CommandFactory and xMapImpls)

wburns:
Restarting CI build as it CI itself blew chunks on last run.

core/src/main/java/org/infinispan/commands/functional/Mutation.java
```diff
@@ -1,11 +1,13 @@
...
 
+import org.infinispan.commands.functional.functions.InjectableComponent;
+import org.infinispan.encoding.DataConversion;
 import org.infinispan.functional.EntryView;
 
 /**
  * Simplified version of functional command used for read-only operations after transactional modifications.
  */
-public interface Mutation<K, V, R> {
+public interface Mutation<K, V, R> extends InjectableComponent {
```
wburns: Hrmm I am not a fan of having this at this level. `InjectableComponent` to me is more of an implementation specific thing.

Also should help perf a tiny little bit, since some implementations don't need injection to be called.

core/src/main/java/org/infinispan/commands/functional/TxReadOnlyKeyCommand.java
```diff
@@ -59,11 +59,22 @@ public void readFrom(ObjectInput input) throws IOException, ClassNotFoundExcepti
    }
 
    @Override
+   public void init(ComponentRegistry componentRegistry) {
+      super.init(componentRegistry);
+      // This may be called from parent's constructor when mutations are not initialized yet
```
wburns: Sorry what implications does this then have? It seems the `mutations` collection is not thread safe, so can we get `ConcurrentModificationException`?

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -1328,6 +1328,8 @@ private void processResponse(org.jgroups.Address src, byte[] buffer, int offset,
          requests.addResponse(requestId, address, response);
       } catch (Throwable t) {
          log.errorProcessingResponse(requestId, src);
+         Exception e = t instanceof Exception ? ((Exception) t) : new CacheException(t);
```
wburns: This change seems unrelated? What is this a fix for?

core/src/main/java/org/infinispan/commands/functional/Mutation.java
danberindei: @wburns I agree, but OTOH the `org.infinispan.commands.functional` package isn't `@public` either.

core/src/main/java/org/infinispan/commands/functional/TxReadOnlyKeyCommand.java
danberindei: @wburns I think it's just that on remote nodes the mutations list is initialized before `init()`, and on on the originator the mutations are added (already initialized) after `init()`. From a thread-safety perspective, AFAICT the collection is unique the command, and we never have 2 threads accessing the same command -- or we have a bug.

core/src/main/java/org/infinispan/commands/functional/Mutations.java
```diff
@@ -89,15 +122,21 @@ public byte type() {
 
       @Override
       public R apply(EntryView.ReadWriteEntryView<K, V> view) {
-         return f.apply(value, view);
+         return f.apply((V) valueDataConversion.fromStorage(value), view);
+      }
+
+      @Override
+      public void inject(ComponentRegistry registry) {
+         registry.wireDependencies(valueDataConversion);
```
danberindei: I only used `wireDependency` for `CacheTopologyControlCommand` because I knew those are pretty rare, I don't think it's efficient enough to use multiple times per operation.

core/src/main/java/org/infinispan/commands/functional/TxReadOnlyKeyCommand.java
rvansa: Mr. Berindei is correct, it's just about initialization order, we need the `if`.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
rvansa: I was seeing a testsuite stall due to a bug (during development) causing NPE. The bug has been fixed, so this is not technically needed...

rvansa:
Rebased and updated. @danberindei see `BaseMutation.inject`, is this what you were looking for?

rvansa:
@wburns I had to keep the `Mutation extends InjectableComponent` because the `inject` is called during command intialization (see `TxReadOnlyKeyCommand.init`). Otherwise I would have to add `@Inject` to mutation implementations and call `wireDependencies` on each.

pruivo:
@rvansa checkstyle fails!

galderz:
Needs rebase

rvansa:
Uuh, I completely forgot this is still open!

galderz:
Last CI build did not complete, fired off new build

tristantarrant:
Requeuing, but it seems the build hangs during the core testsuite.

rvansa:
@tristantarrant I've reproduced stuck build in counters, will check out. Seems that stuck builds are not a random issue anymore but sth worth investigation.

rvansa:
@tristantarrant OK, CI looks good now (besides failure in biased scattered cache which I should investigate).

galderz:
@rvansa Shall we keep PR open until you fix that? Or do you want to create a follow up JIRA for that?

rvansa:
@galderz No, that's completely unrelated, my comment was just a rant.

rvansa:
@galderz You seemed eager to merge this, anything missing?

galderz:
Actually no, I forgot about it before doing CR1.

PR #5517, opened 2017-10-17
ISPN-8410 Off-heap can crash JVM

danberindei:
https://issues.jboss.org/browse/ISPN-8410

* Trace allocated blocks and throw a Java exception
  before any illegal access.
* Extract unsafe access from (Bounded)OffHeapDataContainer.


danberindei:
@wburns The test passes now, but I don't know exactly why. I know https://github.com/infinispan/infinispan/pull/5517/files#diff-8e1bedfcc44a3b7ac91d755042840612L220 was always writing to offset 0, even with eviction, but even if I revert that I still can't get the test to fail :)

wburns:
There are a lot of test failures :(
http://ci.infinispan.org/job/Infinispan/job/PR-5517/1/

danberindei:
Updated, I had broken `metadataSize` in `OffHeapDataContainer.create()`.

wburns:
Tests all pass now, integrating.

tristantarrant:
We want this in 9.1.2 as well ?

danberindei:
Good point @tristantarrant. It cherry-picked cleanly on 9.1.x, so I pushed it myself.

PR #5520, opened 2017-10-17
ISPN-8414 Colored testsuite progress

tristantarrant:
https://issues.jboss.org/browse/ISPN-8414

### DIFF
--- a/commons-test/src/main/java/org/infinispan/commons/test/TestSuiteProgress.java
+++ b/commons-test/src/main/java/org/infinispan/commons/test/TestSuiteProgress.java
@@ -1,8 +1,14 @@
 package org.infinispan.commons.test;
 
 
+import static org.fusesource.jansi.Ansi.ansi;
+
+import java.io.FileDescriptor;
+import java.io.FileOutputStream;
+import java.io.PrintStream;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.fusesource.jansi.Ansi;
 import org.jboss.logging.Logger;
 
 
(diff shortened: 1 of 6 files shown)


tristantarrant:
Need to find a way to suppress colors when running in CI

commons-test/src/main/java/org/infinispan/commons/test/TestSuiteProgress.java
```diff
@@ -16,63 +22,77 @@
...
-   private static AtomicInteger succeeded = new AtomicInteger(0);
-   private static AtomicInteger skipped = new AtomicInteger(0);
+   private AtomicInteger failed = new AtomicInteger(0);
+   private AtomicInteger succeeded = new AtomicInteger(0);
+   private AtomicInteger skipped = new AtomicInteger(0);
+   private final PrintStream out;
+   private final boolean useColor;
+
+   public TestSuiteProgress() {
+      useColor = System.getenv("JENKINS_HOME") == null;
```
danberindei: I wonder if this is really necessary... jansi uses [`isatty`](http://man7.org/linux/man-pages/man3/isatty.3.html) to avoid writing the escape codes when output is redirected.

danberindei: Actually, I see that `isatty` check is in `AnsiConsole`, and you're not using it... still, when I run the build from my Python script (which uses `Popen` to parse the output and set the test count in the title), the colours just disappear.

tristantarrant:
Checking isatty is pointless, since we use surefire's fork which filters the output. In fact System.console() returns null when running the testsuite.
Can I have your Python script so I can check ?

danberindei:
@tristantarrant I forgot surefire is already doing one redirect... but that makes the fact that `./build.sh -pl commons | tee build.log` doesn't show any colors on the terminal or escapes in the file even more confusing. Maybe it's zsh being extra helpful...

You could avoid the Jenkins-specific check by using `-Dansi.strip`, like `AnsiConsole` does. And to skip the `H  ,` in IDEs, you could check for `infinispan.module-suffix`.

tristantarrant:
@danberindei I have made some changes:

- use -Djansi.strip (added to Jenkinsfile) to explicitly avoid colors 
- if output has been not been redirected, print directly without the magic `H  ,` prefix

danberindei:
Integrated, thanks Tristan!

PR #5524, opened 2017-10-19
Documentation - version update

rmacor:
A version update for documentation when it's used as a Wildfly extension.


ryanemerson:
@rmacor Good spot, thanks!

PR #5529, opened 2017-10-23
Transcoding

gustavonalle:
https://issues.jboss.org/browse/ISPN-7418
https://issues.jboss.org/browse/ISPN-7422
https://issues.jboss.org/browse/ISPN-8416

This PR contains most of https://github.com/infinispan/infinispan-designs/pull/9: 

* Caches can be configured with a media-type for keys and values
* Transcoders convert between different media-type on demand
* (extra) Remote query refactor to allow querying via Rest (the query endpoint will follow in another PR)

New features:

* It's possible to send a JSON document via Rest and get it stored as protobuf and indexed
* Hot Rod and Rest can interop using JSON and Protobuf, e.g., a protobuf encoded object can be written via Hot Rod, read as JSON via Rest or vice-versa: written as JSON via Rest and read using protobuf (and POJO) via Hot rod
* JSON and POJO interop: same as above, but storing annotated POJO in the cache (compat mode)


gustavonalle:
The 3 test failures don't seem related, and the full suite is always green in my env

gustavonalle:
Ok, now CI is also green

commons/src/main/java/org/infinispan/commons/dataconversion/MediaType.java
```diff
@@ -0,0 +1,231 @@
...
+      this.type = validate(type);
+      this.subType = validate(subtype);
+      this.typeSubtype = type + "/" + subtype;
+   }
+
+   public static MediaType fromString(String mediaType) {
+      return parse(mediaType);
+   }
+
+   public static MediaType parse(String str) {
```
tristantarrant: You could have used a javax.activation.MimeType which has all of the type, subtype and parameter parsing logic

core/src/main/java/org/infinispan/marshall/core/EncoderRegistryImpl.java
```diff
@@ -48,6 +55,28 @@ public void registerWrapper(Wrapper wrapper) {
...
+   }
+
+   @Override
+   public Transcoder getTranscoder(MediaType mediaType, MediaType another) {
+      Optional<Transcoder> transcoder = transcoders
+            .stream()
+            .filter(t -> t.supportsConversion(mediaType, another))
+            .findAny();
+      if (!transcoder.isPresent()) {
+         throw new EncodingException("Cannot find transcoder from " + mediaType + " to " + another);
```
tristantarrant: I believe some of these exception deserve to be given the jboss logging treatement

commons/src/main/java/org/infinispan/commons/dataconversion/MediaType.java
gustavonalle: I was using it initially but ditched. It'd involve messing with jboss modules to add it as a dependency, plus I didn't like the fact its constructor throws a checked exception.

core/src/main/java/org/infinispan/marshall/core/EncoderRegistryImpl.java
gustavonalle: right, I will externalize them all 👍

commons/src/main/java/org/infinispan/commons/dataconversion/MediaType.java
tristantarrant: javax.activation should be part of SE since 7, but +1 on avoiding the checked exception

gustavonalle: You're right, but this package is not exposed by default in Wildfly, needs explicit dependency

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/dataconversion/ProtostreamJsonTranscoder.java
```diff
@@ -0,0 +1,60 @@
...
+         } catch (IOException e) {
+            throw new CacheException(e);
+         }
+      } else {
+         if (destinationType.match(APPLICATION_PROTOSTREAM)) {
+            try {
+               Reader reader = new InputStreamReader(new ByteArrayInputStream((byte[]) content));
+               return ProtobufUtil.fromCanonicalJSON(ctx, reader);
+            } catch (IOException e) {
+               e.printStackTrace();
```
tristantarrant: Maybe rethrow it as a CacheException

tristantarrant:
Nice, just some very minor things

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/dataconversion/ProtostreamJsonTranscoder.java
gustavonalle: +1, forgot to amend this

gustavonalle:
ops, couple of failures added by last commit, fixing...

gustavonalle:
Updated (and rebased)

PR #5542, opened 2017-10-26
ISPN-8444 JPA cache store failed to deploy in OSGi

mgencur:
https://issues.jboss.org/browse/ISPN-8444

This was caused by the fact that the features.xml file for JPA cache store was not generated properly as the maven plugins use filtering and they couldn't see the javax.persistence property.

### DIFF
--- a/persistence/jpa/pom.xml
+++ b/persistence/jpa/pom.xml
@@ -121,6 +121,15 @@
                </execution>
             </executions>
          </plugin>
+         <plugin>
+            <groupId>org.apache.karaf.tooling</groupId>
+            <artifactId>karaf-maven-plugin</artifactId>
+            <configuration>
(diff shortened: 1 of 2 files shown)


danberindei:
@mgencur The `karaf-maven-plugin` should test `features.xml`, any idea why it didn't catch this problem?

mgencur:
@danberindei Because it doesn't run in the default profile:-D

mgencur:
Ok. I've moved the plugin to the default profile.

wburns:
Integrated as it passes locally. This should get things moving along quicker.

wburns:
Integrated into master, thanks @mgencur !

danberindei:
Thanks for fixing the profile @mgencur!

PR #5544, opened 2017-10-26
Off-heap statistics and compatibility tests

vjuranek:
https://issues.jboss.org/browse/ISPN-8435

### DIFF
--- a/core/src/test/java/org/infinispan/stats/SingleStatsTest.java
+++ b/core/src/test/java/org/infinispan/stats/SingleStatsTest.java
@@ -1,0 +1,116 @@
+package org.infinispan.stats;
+
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertTrue;
+
+import org.infinispan.Cache;
+import org.infinispan.configuration.cache.CacheMode;
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.configuration.cache.StorageType;
+import org.infinispan.eviction.EvictionType;
+import org.infinispan.persistence.dummy.DummyInMemoryStoreConfigurationBuilder;
+import org.infinispan.test.MultipleCacheManagersTest;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.Factory;
+import org.testng.annotations.Test;
+
+@Test(groups = "functional", testName = "stats.SingleStatsTest")
+public class SingleStatsTest extends MultipleCacheManagersTest {
+
+   protected final int EVICTION_MAX_ENTRIES = 3;
+   protected final int TOTAL_ENTRIES = 5;
+   protected StorageType storageType;
+   protected Cache cache;
+   protected Stats stats;
+
+   @Factory
+   public Object[] factory() {
+      return new Object[]{
+            new SingleStatsTest().withStorage(StorageType.BINARY),
+            new SingleStatsTest().withStorage(StorageType.OBJECT),
+            new SingleStatsTest().withStorage(StorageType.OFF_HEAP)
+      };
+   }
+
+   @Override
+   protected void createCacheManagers() throws Throwable {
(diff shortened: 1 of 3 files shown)


server/integration/testsuite/src/test/java/org/infinispan/server/test/cache/container/OffHeapCompatibilityIT.java
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Compatibility mode tests with off-heap data container
+ *
+ * @author vjuranek
+ * @since 9.2
+ */
+@RunWith(Arquillian.class)
+@WithRunningServer({@RunningServer(name = OffHeapCompatibilityIT.CONFIG_NAME)})
+public class OffHeapCompatibilityIT {
```
gustavonalle: OFF_HEAP and compatibility mode is (no pun intended) incompatible, IMO it should throw an error if someone configured the cache with both.

vjuranek: well, it doesn't, it throws exception later on, see [ISPN-8445](https://issues.jboss.org/browse/ISPN-8445). Also, AFAICT the incompatibility is not documented. 
Btw: why these two are not compatible, both store the data as raw bytes, so I'd expect that they can be used together.

gustavonalle: > Btw: why these two are not compatible

In order to interop between remote endpoints and embedded, compat mode internally relies on storing content in "unmarshalled form", so in theory even code running inside the server's JVM would be able to "read" objects without having to deal with marshalling. But OFF_HEAP has a hard requirement on storing byte[ ], so there's a conflict here.

>  both store the data as raw bytes, so I'd expect that they can be used together.

+100, an OFF_HEAP cache in the server should be able to work together with Rest. Have you tried to do so without using compat mode? It should work...

vjuranek: > OFF_HEAP cache in the server should be able to work together with Rest. Have you tried to do so without using compat mode? It should work...

yes, it works, test for it is actually [part](https://github.com/infinispan/infinispan/pull/5544/files#diff-60226fbe2e1fa9ef1e73352fa706a3e0R140) of this PR

vjuranek:
well, I removed test for compatibility mode with off-heap

core/src/test/java/org/infinispan/stats/SingleStatsTest.java
```diff
@@ -0,0 +1,115 @@
...
+   protected final int EVICTION_MAX_ENTRIES = 3;
+   protected final int TOTAL_ENTRIES = 5;
+   protected StorageType storageType;
+   protected Cache cache;
+   protected Stats stats;
+
+   @Factory
+   public Object[] factory() {
+      return new Object[]{
+            //new SingleStatsTest().withStorage(StorageType.OBJECT),
```
wburns: Do we want to reenable this?

vjuranek: yes, thanks! Added also `BINARY` for sure.

server/integration/testsuite/src/test/java/org/infinispan/server/test/cache/container/OffHeapCompatibilityIT.java
wburns: I have fixed the issue with OFF-HEAP and compatibility. The issue was that the memory configuration validate method was never invoked :(

https://github.com/infinispan/infinispan/pull/5590

wburns:
I would recommend rebasing to get fixes from master to better run CI.

vjuranek:
rebased

tristantarrant:
Requeued on CI

vjuranek:
failures are not related to this PR

wburns:
Integrated into master, thanks @vjuranek !

PR #5545, opened 2017-10-26
 ISPN-8446 Create tests for ConflictManager entry merge policies in server mode

jmarkos:
https://issues.jboss.org/browse/ISPN-8446

This PR depends on https://github.com/infinispan/infinispan/pull/5435 and only the last commit is relevant to this PR (the other commits are reviewed in the first PR).

The tests fail because of 2 reasons:
1. the custom merge policy class file is placed in ISPN_SERVER/standalone/deployments, but infinispan isn't scanning that directory when trying to find the custom policy, so it's a bug probably
2. the merge policies aren't called at all on partition merges, @ryanemerson already figured out why but fix is pending


ryanemerson:
Closing this PR. I will create a test for the server to ensure that custom policies are deployed as expected as part of a future PR that will contain other enhancements for partition handling

PR #5547, opened 2017-10-26
ISPN-8408 integrationtests-parent no longer users pom inheritance

ryanemerson:
The only downside to this fix is that we have to have the compiler settings explicit on the integration-tests pom.


wburns:

PR #5548, opened 2017-10-27
ISPN-7926 Add JMX support for Cluster Counters

pruivo:
https://issues.jboss.org/browse/ISPN-7926

### DIFF
--- a/commons/src/main/java/org/infinispan/counter/api/CounterManager.java
+++ b/commons/src/main/java/org/infinispan/counter/api/CounterManager.java
@@ -1,5 +1,7 @@
 package org.infinispan.counter.api;
 
+import java.util.Collection;
+
 /**
  * The {@link CounterManager} creates, defines and returns counters.
  * <p>
@@ -87,5 +89,12 @@ public interface CounterManager {
     * @param counterName The counter's name to remove.
     */
    void remove(String counterName);
+
+   /**
+    * Returns a {@link Collection} of defined counter names.
+    *
+    * @return a {@link Collection} of defined counter names.
+    */
+   Collection<String> getCounterNames();
 
 }
--- a/counter/src/main/java/org/infinispan/counter/impl/CounterModuleLifecycle.java
+++ b/counter/src/main/java/org/infinispan/counter/impl/CounterModuleLifecycle.java
@@ -38,6 +38,7 @@ import org.infinispan.counter.impl.strong.StrongCounterKey;
 import org.infinispan.counter.impl.weak.WeakCounterKey;
 import org.infinispan.factories.GlobalComponentRegistry;
 import org.infinispan.interceptors.impl.EntryWrappingInterceptor;
+import org.infinispan.jmx.CacheManagerJmxRegistration;
 import org.infinispan.lifecycle.ModuleLifecycle;
 import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.partitionhandling.PartitionHandling;
@@ -102,6 +103,10 @@ public class CounterModuleLifecycle implements ModuleLifecycle {
             counterManager = new EmbeddedCounterManager(future,
                   registry.getGlobalConfiguration().globalState().enabled());
             registry.registerComponent(counterManager, CounterManager.class);
+            //this start() is only invoked when the DefaultCacheManager.start() is invoked
+            //it is invoked here again to force it to check the managed global components
+            // and register them in the MBeanServer, if they are missing.
+            registry.getComponent(CacheManagerJmxRegistration.class).start(); //HACK!
          }
       }
    }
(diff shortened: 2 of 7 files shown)


commons/src/main/java/org/infinispan/counter/api/CounterManager.java
```diff
@@ -88,4 +90,11 @@
     */
    void remove(String counterName);
 
+   /**
+    * Returns a {@link Collection} of defined counter's name.
+    *
+    * @return a {@link Collection} of defined counter's name.
+    */
+   Collection<String> getCountersName();
```
danberindei: I think this should be `getCounterNames()`

counter/src/main/java/org/infinispan/counter/impl/CounterModuleLifecycle.java
```diff
@@ -102,6 +103,7 @@ private static void registerCounterManager(GlobalComponentRegistry registry, Com
             counterManager = new EmbeddedCounterManager(future,
                   registry.getGlobalConfiguration().globalState().enabled());
             registry.registerComponent(counterManager, CounterManager.class);
+            registry.getComponent(CacheManagerJmxRegistration.class).start(); //HACK!
```
danberindei: Looks like a hack indeed, but why is it needed?

pruivo: added the following comment
```
            //this start() is only invoked when the DefaultCacheManager.start() is invoked
            //it is invoked here again to force it to check the managed global components 
            // and register them in the MBeanServer, if they are missing.
```
let me know if it is clear or not

pruivo:
@danberindei updated! thanks!

commons/src/main/java/org/infinispan/counter/api/CounterManager.java
```diff
@@ -88,4 +90,11 @@
     */
    void remove(String counterName);
 
+   /**
+    * Returns a {@link Collection} of defined counter's name.
+    *
+    * @return a {@link Collection} of defined counter's name.
```
danberindei: I forgot to mention the javadoc! Should probably be "defined counter names".

counter/src/main/java/org/infinispan/counter/impl/CounterModuleLifecycle.java
```diff
@@ -102,6 +103,10 @@ private static void registerCounterManager(GlobalComponentRegistry registry, Com
             counterManager = new EmbeddedCounterManager(future,
                   registry.getGlobalConfiguration().globalState().enabled());
             registry.registerComponent(counterManager, CounterManager.class);
+            //this start() is only invoked when the DefaultCacheManager.start() is invoked
+            //it is invoked here again to force it to check the managed global components
+            // and register them in the MBeanServer, if they are missing.
+            registry.getComponent(CacheManagerJmxRegistration.class).start(); //HACK!
```
danberindei: WDYT about removing the automation and allowing each component to register on `start()` instead?

rvansa: @danberindei I don't follow, could you elaborate?

danberindei: @rvansa the component registry allows components to be created lazily, and these lazy components can't be exposed in JMX by `CacheJmxRegistration.start()` because they don't exist yet. So I was thinking of letting each MBean component call something like `CacheManagerJmxRegistration.registerMBean(this)` in their own `start()` method.

There are other options of course, e.g. we could also add listeners to the component registry for `CaheManagerJmxRegistration` to automatically register new components in JMX, or we could make sure `CacheManagerJmxRegistration.start()` runs last and throw an exception in `AbstractComponentRegistry` if a MBean component is added after`AbstractComponentRegistry.start()`.

pruivo: -1 to change the automated process, however...
+1 to add a `registerMbean` method! 

AFAIK, all the `core` component are registered before `CacheJmxRegistration.start())`, but in the remaining modules (counters, query, etc), it happens after the `CacheJmxRegistration.start()`. So, your last sentence won't work.

pruivo:
fixed the javadoc!

counter/src/main/java/org/infinispan/counter/impl/CounterModuleLifecycle.java
danberindei: @pruivo it could be that all the core components are registered, but I know for sure there isn't any check: the transport used to be started only with the first cache. 

For the other modules, I don't think `DefaultCacheManager` *has* to call `CacheJmxRegistration.start())` before all the modules are initialized, it's just how it used to work. Based on the ISPN-118 commit message I'd say stoppping the global component registry shouldn't deregister the cache manager MBean, but it should deregister all the other MBeans.

danberindei:
Integrated, thanks Pedro!

I created https://issues.jboss.org/browse/ISPN-8487 for further improvements on the JMX registration.

pruivo:
ok thanks @danberindei

PR #5549, opened 2017-10-27
 HHH-10677 Local Timestamp cache is not updated right after cache.put 

rvansa:
https://hibernate.atlassian.net/browse/HHH-10677


rvansa:
Random failure logged https://issues.jboss.org/browse/ISPN-8455 (unrelated).

PR #5554, opened 2017-10-31
ISPN-8376 Add DMR cluster counters management

vblagoje:
Master only, WIP. Do not integrate yet before @tristantarrant and @pruivo both approve.


vblagoje:
This is a working proposal for integrating counters to servers DMR. On configuration side ounters are located:
```
[domain@localhost:9990 counters=COUNTERS] pwd
/profile=clustered/subsystem=datagrid-infinispan/cache-container=clustered/counters=COUNTERS
[domain@localhost:9990 counters=COUNTERS] ls
strong-counter  weak-counter    

[domain@localhost:9990 counters=COUNTERS] 
```
All strong counter are under strong-counter nodes and weak-counters are under weak-counter.

Configuration of a strong counter can be added:

```
[domain@localhost:9990 counters=COUNTERS] ./strong-counter=a:add(name=a, storage=VOLATILE,type=BOUNDED_STRONG,initial-value=0, lower-bound=0, upper-bound=10)
{
    "outcome" => "success",
    "result" => undefined,
    "server-groups" => {"cluster" => {"host" => {"master" => {
        "server-one" => {"response" => {"outcome" => "success"}},
        "server-two" => {"response" => {"outcome" => "success"}}
    }}}}
}
[domain@localhost:9990 counters=COUNTERS] 
```
One can view that configuration:
```
[domain@localhost:9990 counters=COUNTERS] cd strong-counter=a
[domain@localhost:9990 strong-counter=a] ls
initial-value=0      lower-bound=0        name=a               storage=VOLATILE     type=BOUNDED_STRONG  upper-bound=10       

[domain@localhost:9990 strong-counter=a] 
```


While in runtime path it looks like this:
```
[domain@localhost:9990 strong-counter=a] pwd
/host=master/server=server-one/subsystem=datagrid-infinispan/cache-container=clustered/counters=COUNTERS/strong-counter=a
[domain@localhost:9990 strong-counter=a] ls
initial-value=0      lower-bound=0        name=a               storage=VOLATILE     type=BOUNDED_STRONG  upper-bound=10       value=0              

[domain@localhost:9990 strong-counter=a]
```

vblagoje:
I have also added an operation to reset counter:
```
[domain@localhost:9990 strong-counter=a] pwd
/profile=clustered/subsystem=datagrid-infinispan/cache-container=clustered/counters=COUNTERS/strong-counter=a
[domain@localhost:9990 strong-counter=a] ./:counter-reset()
{
    "outcome" => "success",
    "result" => undefined,
    "server-groups" => {"cluster" => {"host" => {"master" => {
        "server-one" => {"response" => {
            "outcome" => "success",
            "result" => undefined
        }},
        "server-two" => {"response" => {
            "outcome" => "success",
            "result" => undefined
        }}
    }}}}
}
[domain@localhost:9990 strong-counter=a]
```
We can also add an operation to set the counter, increment, decrement and so on

vblagoje:
After having slept over this design I think we should split counter DMR into counter configuration and counter instantiations.  Just like we do for caches. Thoughts @tristantarrant and @pruivo ?

pruivo:
@vblagoje I don't fully understand the DMR, but if you think it would be easier for an user to use the same scheme as in a cache, go for it! 
question: is the DMR node persisted? Counters can be configured to survive cluster restart while other don't. 
Isn't a parser needed for counters configured in the configuration file?

vblagoje:
@pruivo we can create caches in DMR just by referencing the name of configuration node, no parsing needed. For example:
```
[domain@localhost:9990 cache-container=clustered] ./distributed-cache=dist_example_cache:add(configuration=default)
```
Let's hear what @tristantarrant thinks about it.

vblagoje:
@pruivo After an even more thorough look through Counter API I am concluding that it is not possible to create more than one counter from a certain Configuration. So you, for now, we force one counter configuration - one counter instance mapping, is that correct?

pruivo:
The `CounterConfiguration` is immutable. You can use it to create multiple counters but each counter has to have an unique name.

vblagoje:
Right, I see it now @pruivo ! One would just call defineCounter multiple times with a different name and the same CounterConfiguration.

vblagoje:
@pruivo how do you anticipate users are going to deal with counters? Are they going to be creating a lot of them in their applications, say dozens, hundreds, thousands?

The dilemma I have is related to the above-mentioned splitting of the concepts and the mapping between counter configuration and counter instances. On one end of the possible spectrum use, if users are going to have many counters that have the same configuration then it makes sense to delineate the DMR concept of the counter configuration and its counter instance just like we do for caches and cache configuration templates. We deal with cache configurations as templates, one could create hundreds of caches from the same template. Similarly, we can do with counters. On the other end if users are going to create very few counters then it likely does not make much sense to separate counter configurations from its instance, they would have one to one mapping. For each new counter, users would just enter counter configuration and launch an instance of a corresponding counter.

The first approach saves resources and makes large counter instantiations easier while the second approach is easier to understand conceptually but is inefficient if we are going to have many counter instance. The implications of this decision are not just in DMR but in the console as well. So let's nail it right. Maybe I should ask for feedback on this on the dev list?

vblagoje:
@pruivo @tristantarrant closing this one and will open a new one

PR #5556, opened 2017-10-31
ISPN-7776 Clustered Cache State and Configuration

tristantarrant:
https://issues.jboss.org/browse/ISPN-7776

A few interesting things here:
- GlobalStateManager is now always added to the GCR (and not just when global state is enabled). Persistence, however, will only happen when it is enabled in the configuration.
- Added a "matches" method to AttributeSets/Configuraton so that we can tell if a configuration is "compatible", not necessarily identical
- A shared global cache ___config which is Scoped so that others can use too ( @pruivo & @karesti )
- Cache configurations are stored in the global cache "serialized" as XML.
- A common CacheContainerAdmin API between embedded and remote ( @galderz I'm interested in your opinion)
- Persistent caches are "saved" to a caches.xml in the global state persistent directory which acts as a configuration overlay. Concurrent changes to the file are prevented by the use of file locking and renaming.
- Joiners will receive and initialize the set of caches that have been "cluster-created". If they need to be persistent, they will add them to their own caches.xml


tristantarrant:
There is a recurring test failure in CacheManagerAdmin caused by this PR which results in huge amounts of messages queued up in UNICAST3. I'll try and fix it asap.

tristantarrant:
I have fixed the issue and also removed some changes to the server that slipped in (those will be in a separate PR).

commons/src/main/java/org/infinispan/commons/configuration/attributes/AttributeSet.java
```diff
@@ -269,6 +269,25 @@ public boolean equals(Object obj) {
...
+   public boolean matches(AttributeSet other) {
+      if (other.attributes.size() != attributes.size())
+         return false;
+      Iterator<Map.Entry<String, Attribute<?>>> i = attributes.entrySet().iterator();
+      while (i.hasNext()) {
+         Map.Entry<String, Attribute<?>> e = i.next();
+         String key = e.getKey();
+         Attribute<?> value = e.getValue();
+         if (value == null) {
+            if (!(other.attributes.get(key)==null && other.attributes.containsKey(key)))
```
wburns: I would think we would want to check `containsKey` first as I would think it would be much more likely to have no entry than to have keys with null values.

commons/src/main/java/org/infinispan/commons/configuration/attributes/AttributeSet.java
```diff
@@ -269,6 +269,25 @@ public boolean equals(Object obj) {
...
+               return false;
+         } else {
+            if (!value.matches(other.attributes.get(key)))
```
wburns: Hrmm should we check for null from `other.attributed.get(key)` ? It seems a lot of the `matches` implementations would throw `NullPointerException` if this happened.

commons/src/main/java/org/infinispan/commons/marshall/AdminFlagExternalizer.java
```diff
@@ -0,0 +1,31 @@
...
+import java.io.ObjectOutput;
+import java.util.Collections;
+import java.util.Set;
+
+import org.infinispan.commons.api.CacheContainerAdmin;
+
+public class AdminFlagExternalizer extends AbstractExternalizer<CacheContainerAdmin.AdminFlag> {
+   @Override
+   public void writeObject(ObjectOutput output, CacheContainerAdmin.AdminFlag flag) throws IOException {
+      output.writeUTF(flag.name());
```
wburns: Maybe we should use `ordinal` and `values()[offset]` instead of writing the entire name? And actually do you find the Externalizer for enums, are they worth it still?

core/src/main/java/org/infinispan/configuration/cache/Configuration.java
```diff
@@ -334,6 +335,11 @@ public boolean equals(Object obj) {
             return false;
       } else if (!lockingConfiguration.equals(other.lockingConfiguration))
          return false;
+      if (memoryConfiguration == null) {
```
wburns: Good catch :)

core/src/main/java/org/infinispan/configuration/cache/HashConfiguration.java
```diff
@@ -184,4 +185,9 @@ public int hashCode() {
       result = prime * result + ((attributes == null) ? 0 : attributes.hashCode());
       return result;
    }
+
+   @Override
+   public boolean matches(HashConfiguration other) {
```
wburns: Hrmm is this the only one that should override matches for the `Configuration` classes? How is this used btw?

core/src/main/java/org/infinispan/distribution/ch/impl/HashFunctionPartitioner.java
```diff
@@ -60,12 +63,7 @@ public boolean equals(Object o) {
          return true;
       if (o == null || getClass() != o.getClass())
          return false;
-
-      HashFunctionPartitioner that = (HashFunctionPartitioner) o;
-
-      if (numSegments != that.numSegments)
-         return false;
-      return Objects.equals(hashFunction, that.hashFunction);
+      return true;
```
wburns: So all `HashFunctionPartitioner`s are equal to each other? What about eh 3 different variables? Also the `hashCode` is broken as that uses 2 of the fields :(

core/src/main/java/org/infinispan/globalstate/GlobalConfigurationManager.java
```diff
@@ -0,0 +1,42 @@
...
+@Scope(Scopes.GLOBAL)
+public interface GlobalConfigurationManager {
+   String CONFIG_STATE_CACHE_NAME = "org.infinispan.CONFIG";
+   /**
+    * Defines a cluster-wide cache configuration
+    *  @param cacheName the name of the configuration
+    * @param configuration the configuration object
+    * @param flags
+    */
+   Configuration createCache(String cacheName, Configuration configuration, EnumSet<CacheContainerAdmin.AdminFlag> flags);
```
wburns: I would say to make these just `Set` instead of `EnumSet`.

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
```diff
@@ -0,0 +1,195 @@
...
+ * @since 9.2
+ */
+
+public class GlobalConfigurationManagerImpl implements GlobalConfigurationManager {
+   static Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
+   private GlobalConfiguration globalConfiguration;
+   private EmbeddedCacheManager cacheManager;
+   private Cache<ScopedState, Object> stateCache;
+   private GlobalConfigurationStateListener stateCacheListener;
+   private TimeService timeService;
```
wburns: Unused, did you mean to somewhere?

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
```diff
@@ -0,0 +1,195 @@
...
+            new ConfigurationBuilder().build(),
+            EnumSet.of(InternalCacheRegistry.Flag.GLOBAL));
+      parserRegistry = new ParserRegistry();
+      persistentCaches = new ConcurrentHashSet<>();
+   }
+
+   @PostStart
+   public void postStart() {
+      // Initialize caches which are present in the initial state
+      for (Map.Entry<ScopedState, Object> e : getStateCache().getAdvancedCache().entrySet()) {
```
wburns: Don't need call to `getAdvancedCache`.

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationStateListener.java
```diff
@@ -0,0 +1,37 @@
...
+   public void createCache(CacheEntryCreatedEvent<ScopedState, CacheState> event) {
+      String cacheName = event.getKey().getName();
+      CacheState cacheConfiguration = event.getCache().get(event.getKey());
+      gcm.localCreateCache(cacheName, cacheConfiguration);
+   }
+
+   @CacheEntryRemoved
+   public void removeCache(CacheEntryRemovedEvent<ScopedState, CacheState> event) {
+      String cacheName = event.getKey().getName();
+      gcm.log.debugf("Remove cache %s", cacheName);
```
wburns: Do we want a message for create as well?

core/src/main/java/org/infinispan/manager/CacheContainer.java
```diff
@@ -19,4 +20,13 @@
    @Override
    <K, V> Cache<K, V> getCache(String cacheName);
 
+   /**
+    * Provides access to administrative methods which affect the underlying cache container, such as cache creation and
+    * removal. If the underlying container is clustered or remote, the operations will affect
```
wburns: Think you missed finishing this sentence

core/src/main/java/org/infinispan/manager/CacheContainer.java
```diff
@@ -19,4 +20,13 @@
...
+    * @return
+    */
+   default CacheContainerAdmin administration() {
+      throw new UnsupportedOperationException();
```
wburns: So we lose the typing on `CacheContainerAdmin`, should we define that on this method or as a generic on `CacheContainer` at least?

core/src/main/java/org/infinispan/manager/DefaultCacheManagerAdmin.java
```diff
@@ -0,0 +1,57 @@
...
+
+   @Override
+   public <K, V> Cache<K, V> createCache(String cacheName, Configuration configuration) {
+      clusterConfigurationManager.createCache(cacheName, configuration, flags);
+      return cacheManager.getCache(cacheName);
+   }
+
+   @Override
+   public void createCache(String name, String template) {
+
```
wburns: Seems there should be something here :)

core/src/main/java/org/infinispan/manager/DefaultCacheManagerAdmin.java
```diff
@@ -0,0 +1,57 @@
...
+ * The default implementation of {@link EmbeddedCacheManagerAdmin}
+ *
+ * @author Tristan Tarrant
+ * @since 9.2
+ */
+
+public class DefaultCacheManagerAdmin implements EmbeddedCacheManagerAdmin {
+   private final EmbeddedCacheManager cacheManager;
+   private final GlobalConfigurationManager clusterConfigurationManager;
+   private final TimeService timeService;
```
wburns: This isn't used again.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -753,6 +753,8 @@ public void stop() {
       clusterView = new ClusterView(ClusterView.FINAL_VIEW_ID, Collections.emptyList(), null);
 
       requests = null;
+      address = null;
```
wburns: address was nulled out above. maybe move physicalAddress up above

core/src/test/java/org/infinispan/globalstate/AbstractGlobalStateRestartTest.java
```diff
@@ -73,6 +73,7 @@ private void createStatefulCacheManager(String id, boolean clear) {
       addClusterEnabledCacheManager(global, config);
    }
 
+
```
wburns: Unneeded

core/src/test/java/org/infinispan/jmx/ComponentsJmxRegistrationTest.java
```diff
@@ -48,13 +48,6 @@ public void tearDown() {
       cacheContainers.clear();
    }
 
-   public void testStopStartCM() throws Exception {
```
wburns: Does this test not work anymore? Do we disallow restarting a stopped `EmbeddedCacheManager`? If so we should make sure to document and add a simple test like this that asserts an exception etc.

core/src/test/java/org/infinispan/manager/CacheManagerTest.java
```diff
@@ -258,6 +258,7 @@ public void testCacheStopManagerStopFollowedByCacheOp() {
       }
    }
 
+   @Test(groups = "unstable")
```
wburns: Was this unstable before? Do we need a JIRA for it then?

etc/log4j2.xml
```diff
@@ -23,7 +23,7 @@
     <Logger name="org.jgroups" level="DEBUG"/>
 
     <Root level="DEBUG">
-      <AppenderRef ref="STDOUT" level="ERROR"/>
+      <AppenderRef ref="STDOUT" level="FATAL"/>
```
wburns: Unneeded

core/src/main/java/org/infinispan/configuration/cache/HashConfiguration.java
tristantarrant: No, in the future I plan to add more: this should be sufficient for now to demonstrate the capability

commons/src/main/java/org/infinispan/commons/configuration/attributes/Attribute.java
```diff
@@ -185,6 +185,22 @@ public boolean equals(Object obj) {
...
+   /**
+    * Compares this attribute to another attribute taking into account the {@link AttributeDefinition#isLocal()} flag.
+    * If the attribute is global, then this method will return true only if the values are identical.
+    * If the attribute is local, then this method will return true even if the values don't match.
+    * Essentially, this method only ensures that the attribute definitions are equals.
+    * @param other
+    * @return
+    */
+   public boolean matches(Attribute<?> other) {
+      if (!this.definition.equals(other.definition))
```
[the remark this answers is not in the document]
tristantarrant: AttributeDefinitions are final static, so this works, although it's probably better to implement equals

core/src/main/java/org/infinispan/globalstate/GlobalConfigurationManager.java
tristantarrant: Why ? It is easier to build an EnumSet using EnumSet.of(...). Alternatively we could use varargs

wburns: `EnumSet` implements `Set` already so it is just using the interface is all on the method.

core/src/main/java/org/infinispan/manager/DefaultCacheManagerAdmin.java
```diff
@@ -0,0 +1,57 @@
...
+import org.infinispan.util.TimeService;
+
+/**
+ * The default implementation of {@link EmbeddedCacheManagerAdmin}
+ *
+ * @author Tristan Tarrant
+ * @since 9.2
+ */
+
+public class DefaultCacheManagerAdmin implements EmbeddedCacheManagerAdmin {
```
pruivo: I don't fully understand how security is suppose to work, but shouldn't this be internal? 
I mean, what prevents a user of doing
```
DefaultCacheManagerAdmin myHack = new DefaultCacheManagerAdmin(mySecureCache.getCacheManager(), myDummyAuth, ....)
myHack.remove(...)
```

core/src/main/java/org/infinispan/manager/DefaultCacheManagerAdmin.java
```diff
@@ -0,0 +1,57 @@
...
+      this.cacheManager = cm;
+      this.authzHelper = authzHelper;
+      this.clusterConfigurationManager = cm.getGlobalComponentRegistry().getComponent(GlobalConfigurationManager.class);
+      this.timeService = cm.getGlobalComponentRegistry().getComponent(TimeService.class);
+      this.flags = flags;
+   }
+
+   @Override
+   public <K, V> Cache<K, V> createCache(String cacheName, Configuration configuration) {
+      clusterConfigurationManager.createCache(cacheName, configuration, flags);
```
pruivo: authorisation check?

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
```diff
@@ -206,6 +206,19 @@
    boolean cacheExists(String cacheName);
 
    /**
+    * Creates a cache on the local node using the supplied configuration. The cache may be clustered, but this
+    * method (or an equivalent combinatio of {@link #defineConfiguration(String, Configuration)} and
+    * {@link #getCache(String, boolean)}) needs to be invoked on all nodes.
```
pruivo: typo combinatio => combination

core/src/main/java/org/infinispan/manager/EmbeddedCacheManagerAdmin.java
```diff
@@ -0,0 +1,29 @@
...
+import org.infinispan.Cache;
+import org.infinispan.commons.api.CacheContainerAdmin;
+import org.infinispan.configuration.cache.Configuration;
+
+/**
+ * Cache manager operations which affect the whole cluster. An instance of this can be retrieved from
+ * {@link EmbeddedCacheManager#administration()}
+ *
+ * @author Tristan Tarrant
+ * @since 9.1
```
pruivo: 9.2

core/src/main/java/org/infinispan/globalstate/GlobalConfigurationManager.java
pruivo: or a collection

core/src/main/java/org/infinispan/globalstate/GlobalConfigurationManager.java
```diff
@@ -0,0 +1,42 @@
...
+    * @param cacheName the name of the cache
+    * @param flags
+    */
+   void removeCache(String cacheName, EnumSet<CacheContainerAdmin.AdminFlag> flags);
+
+   /**
+    * Returns the clustered configuration state cache
+    * @return the clustered configuration state cache
+    */
+   Cache<ScopedState, Object> getStateCache();
```
pruivo: -1 to expose the cache. if we need to change the internals, we need to deprecate/change API.

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationStateListener.java
pruivo: move the message to `gcm.localRemoveCache()`

core/src/main/java/org/infinispan/globalstate/GlobalConfigurationManager.java
tristantarrant: This is not public API anyway so we don't have to be picky :)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
```diff
@@ -344,7 +346,7 @@ public Marshaller getMarshaller() {
 
    public RemoteCacheManagerAdmin administration() {
       OperationsFactory operationsFactory = new OperationsFactory(transportFactory, codec, asyncExecutorService, configuration);
-      return new RemoteCacheManagerAdminImpl(operationsFactory);
+      return new RemoteCacheManagerAdminImpl(operationsFactory, EnumSet.noneOf(CacheContainerAdmin.AdminFlag.class));
```
pruivo: use NONE!

commons/src/main/java/org/infinispan/commons/api/BasicCache.java
```diff
@@ -110,7 +110,7 @@
    /**
     * An overloaded form of {@link #replace(Object, Object, Object)}, which takes in lifespan parameters.
     *
-    * @param key      key to use
+    * @param key w     key to use
```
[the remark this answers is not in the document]
tristantarrant: ok

commons/src/main/java/org/infinispan/commons/api/CacheContainerAdmin.java
```diff
@@ -0,0 +1,68 @@
...
+    * @since 9.2
+    */
+   enum AdminFlag {
+      /**
+       * If the operation affects configuration, make it persistent. If the server cannot honor this flag an error will
+       * be returned
+       */
+      PERSISTENT("persistent");
+
+      public static final EnumSet<AdminFlag> NONE = EnumSet.noneOf(AdminFlag.class);
```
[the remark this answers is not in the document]
tristantarrant: ok

commons/src/main/java/org/infinispan/commons/util/Util.java
```diff
@@ -1092,4 +1092,9 @@ public static String xmlify(String s) {
    public static String[] stringArray(int length) {
       return length == 0 ? EMPTY_STRING_ARRAY : new String[length];
    }
+
+   public static String basename(Path path, String ext) {
```
pruivo: all changes in this class can be reverted!

commons/src/main/java/org/infinispan/commons/marshall/AdminFlagExternalizer.java
tristantarrant: ok

core/src/main/java/org/infinispan/configuration/cache/Configuration.java
tristantarrant: :)

core/src/main/java/org/infinispan/globalstate/GlobalConfigurationManager.java
pruivo: why returning `Configuration`?

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
```diff
@@ -0,0 +1,195 @@
...
+      if (flags.contains(CacheContainerAdmin.AdminFlag.PERSISTENT) && !globalConfiguration.globalState().enabled())
+         throw log.globalStateDisabled();
+   }
+
+   @Override
+   public Configuration createCache(String cacheName, Configuration configuration, EnumSet<CacheContainerAdmin.AdminFlag> flags) {
+      validateFlags(flags);
+      try {
+         CacheState state = new CacheState(parserRegistry.serialize(cacheName, configuration), flags);
+         getStateCache().put(new ScopedState(CACHE_SCOPE, cacheName), state);
```
pruivo: shouldn't be `putIfAbsent()`?

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
```diff
@@ -0,0 +1,195 @@
...
+ * @author Tristan Tarrant
+ * @since 9.2
+ */
+
+public class GlobalConfigurationManagerImpl implements GlobalConfigurationManager {
+   static Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
+   private GlobalConfiguration globalConfiguration;
+   private EmbeddedCacheManager cacheManager;
+   private Cache<ScopedState, Object> stateCache;
+   private GlobalConfigurationStateListener stateCacheListener;
```
pruivo: I guess it can be removed... it is local only... or the `stop()` method is missing to remove the listener :)

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
tristantarrant: ok

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
tristantarrant: ok

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationStateListener.java
tristantarrant: yes

core/src/main/java/org/infinispan/manager/CacheContainer.java
tristantarrant: ok

core/src/main/java/org/infinispan/manager/CacheContainer.java
tristantarrant: ok

core/src/main/java/org/infinispan/manager/DefaultCacheManagerAdmin.java
tristantarrant: ok

core/src/main/java/org/infinispan/manager/DefaultCacheManagerAdmin.java
tristantarrant: oops :)

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
tristantarrant: ok

core/src/test/java/org/infinispan/globalstate/AbstractGlobalStateRestartTest.java
tristantarrant: ok

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
```diff
@@ -278,7 +291,9 @@
     * memory and in any backing cache store.
     *
     * @param cacheName name of cache to remove
+    * @deprecated obtain a {@link org.infinispan.commons.api.CacheContainerAdmin} instance using {@link #administration()} and invoke the {@link org.infinispan.commons.api.CacheContainerAdmin#removeCache(String)} method
     */
+   @Deprecated // since 9.2
    void removeCache(String cacheName);
```
pruivo: there are sill usage of this method in the server. I guess they can be removed/upgraded.

commons/src/main/java/org/infinispan/commons/configuration/attributes/AttributeDefinition.java
```diff
@@ -69,6 +74,10 @@ public boolean isAutoPersist() {
       return autoPersist;
    }
 
+   public boolean isLocal() {
```
pruivo: the IDE says that the result is always negated. maybe, instead of creating a `local` flag, a `global` flag is better :).

commons/src/main/java/org/infinispan/commons/configuration/attributes/AttributeSet.java
```diff
@@ -269,6 +269,25 @@ public boolean equals(Object obj) {
       return true;
    }
 
+   public boolean matches(AttributeSet other) {
+      if (other.attributes.size() != attributes.size())
+         return false;
+      Iterator<Map.Entry<String, Attribute<?>>> i = attributes.entrySet().iterator();
+      while (i.hasNext()) {
```
pruivo: foreach instead of iterator?

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
pruivo: is this loop really needed? 
the listener has `includeCurrentState = true`. I think it receives all the state and you don't need to iterate here first (except if you want some kind of sync??)

core/src/test/java/org/infinispan/manager/CacheManagerTest.java
```diff
@@ -363,12 +364,13 @@ public void testCacheManagerRestartReusingConfigurations() {
...
             EmbeddedCacheManager cm2 = cms[1];
+            TestingUtil.waitForNoRebalance(cm1.getCache(), cm2.getCache());
             Cache<Object, Object> c1 = cm1.getCache();
-            cm2.getCache();
 
             GlobalConfiguration globalCfg = cm1.getCacheManagerConfiguration();
             Configuration cfg = c1.getCacheConfiguration();
-            cm1.stop();
+            TestingUtil.killCacheManagers(cm1);
+            globalCfg = new GlobalConfigurationBuilder().read(globalCfg).build();
```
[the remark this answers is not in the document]
tristantarrant: I think I can go back to not copying now, since the fix is nulling the physicalAddress in JGroupsTransport

core/src/main/java/org/infinispan/manager/DefaultCacheManagerAdmin.java
tristantarrant: The constructor is package protected

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
```diff
@@ -524,27 +532,10 @@ public void run() {
...
-               future = cmd.invokeAsync().thenCompose(o -> remoteFuture);
-            } else {
-               future = cmd.invokeAsync();
-            }
-
-            future.get();
-         } catch (Throwable t) {
-            throw new CacheException("Error removing cache", t);
-         }
+      if (cacheManagerAdmin != null)
```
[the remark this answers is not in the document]
tristantarrant: ok

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
```diff
@@ -257,6 +256,7 @@ public DefaultCacheManager(GlobalConfiguration globalConfiguration, Configuratio
       this.stats = new CacheContainerStatsImpl(this);
       health = new HealthImpl(this);
       globalComponentRegistry.registerComponent(new HealthJMXExposerImpl(health), HealthJMXExposer.class);
+      this.cacheManagerAdmin = new DefaultCacheManagerAdmin(this, authzHelper, EnumSet.noneOf(CacheContainerAdmin.AdminFlag.class));
```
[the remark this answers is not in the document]
tristantarrant: Nah, I'll remove it.

core/src/main/java/org/infinispan/globalstate/impl/CacheState.java
```diff
@@ -0,0 +1,51 @@
...
+import org.infinispan.commons.api.CacheContainerAdmin;
+import org.infinispan.commons.marshall.SerializeWith;
+
+/**
+ *
+ *
+ * @author Tristan Tarrant
+ * @since 9.2
+ */
+@SerializeWith(CacheState.Externalizer.class)
```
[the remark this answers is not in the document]
tristantarrant: ok

core/src/main/java/org/infinispan/globalstate/GlobalConfigurationManager.java
tristantarrant: ok

commons/src/test/java/org/infinispan/commons/configuration/attributes/AttributeTest.java
```diff
@@ -131,6 +131,23 @@ public void attributeChanged(Attribute<Boolean> attribute, Boolean oldValue) {
...
+      setB.attribute("local").set("B");
+      setB.attribute("global").set("A");
+      assertTrue(setA.matches(setB));
+      assertFalse(setA.equals(setB));
+      setB.attribute("global").set("B");
+      assertFalse(setA.matches(setB));
+      assertFalse(setA.equals(setB));
+   }
+
    static class Holder<T> {
```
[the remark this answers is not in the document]
tristantarrant: ok

commons/src/main/java/org/infinispan/commons/configuration/attributes/CopyConstructorAttributeCopier.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.infinispan.commons.CacheException;
+
+/**
+ * CopyConstructorAttributeCopier. This {@link AttributeCopier} expects the attribute value class to have a copy constructor
+ *
+ * @author Tristan Tarrant
+ * @since 9.2
+ */
+public class CopyConstructorAttributeCopier<T> implements AttributeCopier<T> {
```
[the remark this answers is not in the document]
tristantarrant: yes

pruivo:
hmmm looks like the org.infinispan.manager.CacheManagerTest is running forever in my machine...

commons/src/main/java/org/infinispan/commons/api/CacheContainerAdmin.java
```diff
@@ -0,0 +1,68 @@
...
+    *                 will be used
+    */
+   void createCache(String name, String template);
+
+   /**
+    * Removes a cache from the cache container.
+    *
+    * @param name the name of the cache to remove
+    */
+   void removeCache(String name);
```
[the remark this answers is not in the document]
tristantarrant: Well, it depends on the configuration of the store (e.g. drop on exit for JDBC)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
tristantarrant: I've killed NONE

core/src/main/java/org/infinispan/manager/EmbeddedCacheManager.java
tristantarrant: I have removed the usage. Still there for abstract wrappers though.

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
```diff
@@ -0,0 +1,202 @@
...
+            new ConfigurationBuilder().build(),
+            EnumSet.of(InternalCacheRegistry.Flag.GLOBAL));
+      parserRegistry = new ParserRegistry();
+      persistentCaches = new ConcurrentHashSet<>();
+   }
+
+   @PostStart
+   public void postStart() {
+      // Initialize caches which are present in the initial state
+      for (Map.Entry<ScopedState, Object> e : getStateCache().entrySet()) {
```
tristantarrant: The loop here is because I want to ensure that the caches are created before we process the persistent stuff below. With the listener (even with includeCurrentState), this doesn't happen synchronously.

tristantarrant:
I've addressed all comments

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/AdminFlag.java
```diff
@@ -1,27 +1,35 @@
...
 
+import org.infinispan.commons.api.CacheContainerAdmin;
+
 /**
  * Flags which affect only administrative operations
  * @author Tristan Tarrant
  * @since 9.0
+ * @deprecated use {@link org.infinispan.commons.api.CacheContainerAdmin.AdminFlag} instead
  */
 public enum AdminFlag {
```
anistor: I'd also add ```@Deprecated``` here.

core/src/test/java/org/infinispan/jmx/ComponentsJmxRegistrationTest.java
tristantarrant: Re-enabled it

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManagerAdmin.java
```diff
@@ -3,14 +3,15 @@
...
+import org.infinispan.commons.api.CacheContainerAdmin;
 
 /**
  * Remote Administration operations
  *
  * @author Tristan Tarrant
  * @since 9.1
  */
-public interface RemoteCacheManagerAdmin {
+public interface RemoteCacheManagerAdmin extends CacheContainerAdmin<RemoteCacheManagerAdmin> {
```
anistor: There's another method here that deserves an ```@Override```.

tristantarrant:
I've somehow broken mbean unregistration that screws up the jsr107 tests. Working on it.

tristantarrant:
I've somehow broken mbean unregistration that screws up the jsr107 tests. Working on it.

commons/src/main/java/org/infinispan/commons/api/CacheContainerAdmin.java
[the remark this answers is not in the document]
tristantarrant: Actually, the behaviour is unchanged from before: stores are cleared on exit.

anistor:
@tristantarrant So is it ready to be merged then? I looked at the changes and I'm ok with it.

tristantarrant:
Waiting for CI. I'd really like to have this in for Beta1 so we can postpone the release to tomorrow if necessary

anistor:
OK.

anistor:
I noticed that some ```@Deprecated``` elements do not have a ```@deprecated``` javadoc tag and viceversa.

slaskawi:
I need to cancel this builds... I'll retrigger it after the update.

commons/src/main/java/org/infinispan/commons/api/CacheContainerAdmin.java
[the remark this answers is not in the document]
pruivo: so all the data is cleared. IMO, it should "say" / "warn" in the javadoc

commons/src/main/java/org/infinispan/commons/configuration/attributes/AttributeSet.java
pruivo: ... I guess not...

commons/src/main/java/org/infinispan/commons/configuration/attributes/CopyConstructorAttributeCopier.java
```diff
@@ -0,0 +1,32 @@
...
+   }
+
+   @Override
+   public T copyAttribute(T attribute) {
+      try {
+         Class<T> klass = (Class<T>) attribute.getClass();
+         Constructor<T> constructor = klass.getConstructor(klass);
+         return constructor.newInstance(attribute);
+      } catch (Exception e) {
+         throw new CacheException(e);
```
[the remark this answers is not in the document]
pruivo: ^

core/src/main/java/org/infinispan/configuration/cache/Configuration.java
```diff
@@ -376,4 +382,48 @@ public boolean equals(Object obj) {
          return false;
       return true;
    }
+
+   @Override
+   public boolean matches(Configuration other) {
```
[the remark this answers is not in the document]
pruivo: yes or no? :)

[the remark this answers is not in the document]
tristantarrant: I've fixed it since then :)

commons/src/main/java/org/infinispan/commons/configuration/attributes/CopyConstructorAttributeCopier.java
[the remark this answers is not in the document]
tristantarrant: Right

commons/src/main/java/org/infinispan/commons/configuration/attributes/AttributeSet.java
tristantarrant: Well, I want to exit the loop fast if there is a non-match.

commons/src/main/java/org/infinispan/commons/api/CacheContainerAdmin.java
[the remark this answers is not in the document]
tristantarrant: I'll add it.

core/src/main/java/org/infinispan/distribution/ch/impl/HashFunctionPartitioner.java
pruivo: @tristantarrant ^

commons/src/main/java/org/infinispan/commons/configuration/attributes/AttributeSet.java
pruivo: not the `forEach()`, this "for-each"
```
for (Map.Entry<String, Attribute<?>> e : attributes.entrySet()) {
```

core/src/main/java/org/infinispan/distribution/ch/impl/HashFunctionPartitioner.java
tristantarrant: I guess we should consider all fields of HashFunctionPartitioner as transient, since those are set at runtime from other parts of the configuration.

tristantarrant: Ok, I've solved with https://github.com/infinispan/infinispan/pull/5556/files#diff-f50db7fa1d8dc90862597d5393c6795bR64 and restored the equality.
Bear in mind that segmentSize is a field computed from numSegments, so doesn't affect equality or hashcode

tristantarrant:
Unfortunately CI has some catastrophic failures in the JCache stuff. I'm investigating.

tristantarrant:
2 unrelated failures

galderz:
> 2 unrelated failures

Are those unrelated failures being tracked? I don't see any JIRAs for those and not in spreadsheet, adding them. We really can't be slacking on appearing random failures :\

galderz:
Added [ISPN-8524](https://issues.jboss.org/browse/ISPN-8524) and [ISPN-8525](https://issues.jboss.org/browse/ISPN-8525) to track those unrelated random failures.

commons/src/main/java/org/infinispan/commons/api/CacheContainerAdmin.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.EnumSet;
+
+/**
+ * Administrative cache container operations.
+ *
+ * @author Tristan Tarrant
+ * @since 9.2
+ */
+
+public interface CacheContainerAdmin<C extends CacheContainerAdmin> {
```
galderz: We need async APIs too for Vert.x and many other reactive use cases.

tristantarrant: https://issues.jboss.org/browse/ISPN-8526

PR #5562, opened 2017-11-02
ISPN-8454 Off Heap can crash when using JMH test

wburns:
* Fixed issue with eviction removal to use proper pointer
* Also made sure address lookup doesn't have overflow
* Added new write and read test

https://issues.jboss.org/browse/ISPN-8454

### DIFF
--- a/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
@@ -259,8 +259,8 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
             try {
                InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(addressToRemove);
                passivator.passivate(ice);
-               // TODO: this reareads the object from memory again!
-               performRemove(addressToRemove, ice.getKey());
+               // TODO: this rereads the object from memory again!
+               performRemove(memoryLookup.getMemoryAddress(ice.getKey()), ice.getKey());
                evictionManager.onEntryEviction(Collections.singletonMap(ice.getKey(), ice));
             } finally {
                entryWriteLock.unlock();
@@ -308,11 +308,13 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
    private void moveToEnd(long lruNode) {
       if (lruNode != lastAddress) {
          long nextLruNode = OffHeapLruNode.getNext(lruNode);
+         assert nextLruNode != 0;
          if (lruNode == firstAddress) {
             OffHeapLruNode.setPrevious(nextLruNode, 0);
             firstAddress = nextLruNode;
          } else {
             long prevLruNode = OffHeapLruNode.getPrevious(lruNode);
+            assert prevLruNode != 0;
             OffHeapLruNode.setNext(prevLruNode, nextLruNode);
             OffHeapLruNode.setPrevious(nextLruNode, prevLruNode);
          }
--- a/core/src/main/java/org/infinispan/container/offheap/MemoryAddressHash.java
+++ b/core/src/main/java/org/infinispan/container/offheap/MemoryAddressHash.java
@@ -10,6 +10,7 @@ import sun.misc.Unsafe;
  */
 public class MemoryAddressHash {
    private static final Unsafe UNSAFE = UnsafeHolder.UNSAFE;
+   private static final OffHeapMemory MEMORY = OffHeapMemory.INSTANCE;
    private static final int MAXIMUM_CAPACITY = 1 << 30;
    private static final int HASH_BITS = 0x7fffffff; // usable bits of normal node hash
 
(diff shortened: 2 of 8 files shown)


wburns:
This will need to be ported to 9.1.x as well.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -259,8 +259,8 @@ private void ensureSize() {
             try {
                InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(addressToRemove);
                passivator.passivate(ice);
-               // TODO: this reareads the object from memory again!
-               performRemove(addressToRemove, ice.getKey());
+               // TODO: this rereads the object from memory again!
+               performRemove(memoryLookup.getMemoryAddress(ice.getKey()), ice.getKey());
```
wburns: This is the actual big bug fix. The problem was this was passing the address pointer to remove (which would be a fine assumption from the name as Dan mentioned). However this method required the address of the first node in the memory lookup. This is required since this is a forward linked list only and such when we remove an element we need its parent to mend the linked list. By not passing this the list would be corrupted which could cause all sorts of issues (especially if this memory location was reused by another entry - which was almost always the case)

core/src/main/java/org/infinispan/container/offheap/MemoryAddressHash.java
```diff
@@ -19,32 +20,32 @@
...
       this.pointerCount = nextPowerOfTwo(pointers);
       long bytes = ((long) pointerCount) << 3;
-      memory = UNSAFE.allocateMemory(bytes);
+      memory = OffHeapMemory.allocate(bytes);
       // Have to clear out bytes to make sure no bad stuff was read in
       UNSAFE.setMemory(memory, bytes, (byte) 0);
    }
 
-   private int findOffset(Object instance) {
+   private long findOffset(Object instance) {
```
danberindei: We need a stress test with > 500m entries ;)

wburns:
JMH test results btw https://gist.github.com/wburns/d3fdc9bc707ba296374c31d99f6e7a05

core/src/test/java/org/infinispan/container/offheap/KeyGenerator.java
```diff
@@ -0,0 +1,78 @@
...
+import org.infinispan.commons.marshall.WrappedByteArray;
+import org.infinispan.metadata.EmbeddedMetadata;
+import org.infinispan.metadata.Metadata;
+
+/**
+ * Generates keys and values to be used during the cache.
+ * These instances are pre-created to minimize impact
+ * during measurements.
+ * Strings are a popular choice, but trigger ad-hoc optimisations
+ * so we have options to test both with strings and other things.
```
danberindei: I'd like to see store as binary mentioned specifically wrt the ad-hoc optimizations (and any other places you know have optimizations for strings).

core/src/test/java/org/infinispan/container/offheap/OffHeapBoundedSingleNodeStressTest.java
```diff
@@ -1,30 +1,40 @@
...
 @Test(groups = "stress", testName = "container.offheap.OffHeapBoundedSingleNodeStressTest")
 public class OffHeapBoundedSingleNodeStressTest extends OffHeapMultiNodeStressTest {
 
    @Override protected void createCacheManagers() throws Throwable {
       ConfigurationBuilder dcc = getDefaultClusteredCacheConfig(CacheMode.LOCAL, false);
-      dcc.memory().storageType(StorageType.OFF_HEAP).evictionType(EvictionType.COUNT).size(2000);
+      dcc.memory().storageType(StorageType.OFF_HEAP).evictionType(EvictionType.COUNT)
+            .size(500)
+            // Set the address count to minimum to cause as many issues as possible!
+            .addressCount(1);
```
danberindei: I'd set it to 2, in the unlikely case an issue only appears when two buckets are modified concurrently.

wburns: Technically it will never be 1 or even 2 normally, I just put a low number. This will essentially become `roundUpToPowerOf2(#CPU) *2`

core/src/main/java/org/infinispan/container/offheap/MemoryAddressHash.java
```diff
@@ -19,32 +20,32 @@
    public MemoryAddressHash(int pointers) {
       this.pointerCount = nextPowerOfTwo(pointers);
       long bytes = ((long) pointerCount) << 3;
-      memory = UNSAFE.allocateMemory(bytes);
+      memory = OffHeapMemory.allocate(bytes);
```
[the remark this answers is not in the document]
wburns: I agree, I didn't like it either. I can make it consistent.

core/src/test/java/org/infinispan/container/offheap/OffHeapBoundedSingleNodeStressTest.java
danberindei: I forgot about that, I thought you could actually force the container to use a hash table with a single linked list :)

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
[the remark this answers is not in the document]
wburns: Hrmm not sure I understand in this context. We are using the precomputed hashCode stored in off heap when we create the entry at `fromMemory` call above.

core/src/main/java/org/infinispan/container/offheap/MemoryAddressHash.java
wburns: Have a machine to "lend" me :)

core/src/test/java/org/infinispan/container/offheap/KeyGenerator.java
wburns: Sorry this is a remnant of the JMH benchmark. Let me remove this jazz.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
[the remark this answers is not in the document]
wburns: I fixed the part in fromMemory when using a custom metadata that wasn't passing the hashCode.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -345,8 +345,8 @@ protected void entryRetrieved(long entryAddress) {
             size.decrementAndGet();
             return ice;
          }
-         prevAddress = address;
-         address = nextAddress;
+         prevAddress = bucketHeadAddress;
+         bucketHeadAddress = nextAddress;
```
danberindei: One more small complaint: once you change `bucketHeadAddress` it's no longer a the head of the list, so you should use a local variable for iteration.

wburns: Sure I can just assign it to address on the first line instead.

wburns:
Updated

PR #5565, opened 2017-11-03
ISPN-8486 Support home relative file store paths

galderz:
https://issues.jboss.org/browse/ISPN-8486

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
+++ b/core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
@@ -105,6 +105,7 @@ public class SingleFileStore<K, V> implements AdvancedLoadWriteStore<K, V> {
          if (location == null || location.trim().length() == 0)
             location = "Infinispan-SingleFileStore";
 
+         location = location.replaceFirst("^~", System.getProperty("user.home"));
          file = new File(location, ctx.getCache().getName() + ".dat");
          if (!file.exists()) {
             File dir = file.getParentFile();


danberindei:
Could you also mention this in the javadoc/user guide?

Personally I think this is too specific, I'd rather have the option of using system properties in the path.

tristantarrant:
This should also work for all paths (other stores, global state, etc)

danberindei:
@tristantarrant I meant this patch allows paths relative to the user home, but I'm not sure how common that use case is vs. making the path relative to a system property/environment variable.

tristantarrant:
We should include this https://github.com/infinispan/infinispan/blob/d46c658a872b9add4040badee80368c97475dbfd/commons/src/main/java/org/infinispan/commons/util/StringPropertyReplacer.java#L52 in the documentation explaining how property replacement works. Tilde expansion is also a nice addition, but to be perfect it must only expand it when it is followed by /. We cannot easily support full ~username expansion without some horrible trickery

galderz:
> Personally I think this is too specific, I'd rather have the option of using system properties in the path.

True, that'd be a better option indeed.

galderz:
> This should also work for all paths (other stores, global state, etc)

What global state path? I understand paths in other stores, but not sure I understand it in the context of global state.

tristantarrant:
Global state has two paths: a persistent one and a temporary one, so it would make sense to support tilde expansion there.

galderz:
Very good points made, I'll close for now and revisit once I have the opportunity to look into this again.

PR #5566, opened 2017-11-03
ISPN-8423 Remove metadata size integer in cases where we know the size

wburns:
https://issues.jboss.org/browse/ISPN-8423

This reduces off heap additional overhead from 17 bytes to 13 bytes in IMMORTAL no version/custom metadata. Expiration will now be 33 or 49 bytes overhead, compared to 37 or 53 before.


tristantarrant:
Please open another PR for 9.1.x

PR #5567, opened 2017-11-06
ISPN-8500 Don't return expired entries

vjuranek:
https://issues.jboss.org/browse/ISPN-8500

### DIFF
--- a/core/src/main/java/org/infinispan/container/DefaultDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/DefaultDataContainer.java
@@ -553,8 +553,9 @@ public class DefaultDataContainer<K, V> implements DataContainer<K, V> {
       if (action == null)
          throw new IllegalArgumentException("No action specified");
 
+      long now = timeService.wallClockTime();
       entries.forEach((K key, InternalCacheEntry<K, V> value) -> {
-         if (filter.accept(key)) {
+         if (filter.accept(key) && !value.isExpired(now)) {
             action.accept(key, value);
          }
       });
@@ -572,8 +573,9 @@ public class DefaultDataContainer<K, V> implements DataContainer<K, V> {
       if (action == null)
          throw new IllegalArgumentException("No action specified");
 
+      long now = timeService.wallClockTime();
       entries.forEach((K key, InternalCacheEntry<K, V> value) -> {
-         if (filter.accept(key, value.getValue(), value.getMetadata())) {
+         if (filter.accept(key, value.getValue(), value.getMetadata()) && !value.isExpired(now)) {
             action.accept(key, value);
          }
       });
--- a/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
@@ -33,6 +33,7 @@ import org.infinispan.util.logging.LogFactory;
 
 /**
  * Data Container implementation that stores entries in native memory (off-heap).
+ *
  * @author wburns
  * @since 9.0
  */
@@ -151,10 +152,11 @@ public class OffHeapDataContainer implements DataContainer<WrappedBytes, Wrapped
 
    protected InternalCacheEntry<WrappedBytes, WrappedBytes> performGet(long address, Object k, boolean peek) {
       WrappedBytes wrappedKey = toWrapper(k);
+      long now = timeService.wallClockTime();
       while (address != 0) {
          long nextAddress = offHeapEntryFactory.getNext(address);
          InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(address);
-         if (wrappedKey.equalsWrappedBytes(ice.getKey())) {
+         if (wrappedKey.equalsWrappedBytes(ice.getKey()) && !ice.isExpired(now)) {
             if (!peek) {
                entryRetrieved(address);
             }
@@ -183,9 +185,10 @@ public class OffHeapDataContainer implements DataContainer<WrappedBytes, Wrapped
    /**
     * Performs the actual put operation putting the new address into the memory lookups.  The write lock for the given
     * key <b>must</b> be held before calling this method.
-    * @param address the entry address of the first element in the lookup
+    *
+    * @param address    the entry address of the first element in the lookup
     * @param newAddress the address of the new entry
-    * @param key the key of the entry
+    * @param key        the key of the entry
     */
    protected void performPut(long address, long newAddress, WrappedBytes key) {
       // Have to start new linked node list
@@ -244,8 +247,9 @@ public class OffHeapDataContainer implements DataContainer<WrappedBytes, Wrapped
    }
 
    /**
-    * Invoked when an entry is about to be created.  The new address is fully addressable,
-    * The write lock will already be acquired for the given segment the key mapped to.
+    * Invoked when an entry is about to be created.  The new address is fully addressable, The write lock will already
+    * be acquired for the given segment the key mapped to.
+    *
     * @param newAddress the address just created that will be the new entry
     */
    protected void entryCreated(long newAddress) {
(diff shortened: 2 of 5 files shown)


core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
```diff
@@ -1,23 +1,36 @@
...
 public class ExpirationFunctionalTest extends SingleCacheManagerTest {
 
    protected final int SIZE = 10;
    protected ControlledTimeService timeService = new ControlledTimeService();
+   protected StorageType storage;
+
+   @Factory
+   public Object[] factory() {
+      return new Object[]{
+            new ExpirationFunctionalTest().withStorage(StorageType.OBJECT),
```
wburns: Might as well include BINARY just in case :)

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -154,7 +155,7 @@ protected void checkDeallocation() {
       while (address != 0) {
          long nextAddress = offHeapEntryFactory.getNext(address);
          InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(address);
-         if (wrappedKey.equalsWrappedBytes(ice.getKey())) {
+         if (wrappedKey.equalsWrappedBytes(ice.getKey()) && !ice.isExpired(timeService.wallClockTime())) {
```
wburns: There are a few other places this is need. I would advise renaming `entryStream` to `entryStreamIncludingExpired` and make a new `entryStream` that invokes `entryStreamIncludingExpired` that filters out expired entries. Then we can call the proper method. Currently only the `iteratorIncludingExpired` should have visibility to expired entries.

We also have to add an `isExpired` check in the `executeTask`, `containsKey` and `remove` methods as they all currently operate upon expired entries. For the `containsKey` we should add a method to the `OffHeapEntryFactory` class that just returns if the key is expired or not. Then it only needs to read the type to figure out what expiration variables to read instead of the entire object.

If you want I can handle this, but I figure you might enjoy doing this :)

vjuranek:
PR updated

vjuranek:
Looks like tests failures are not related to functionality itself (everything passes then test is run alone), but caused by running several instances of the test with different TestNG `@Factory` setup. Will look on it more tomorrow.

galderz:
Needs rebasing

vjuranek:
rebased

core/src/main/java/org/infinispan/container/offheap/OffHeapEntryFactoryImpl.java
```diff
@@ -359,6 +360,55 @@ public boolean equalsKey(long address, WrappedBytes wrappedBytes) {
...
+      // 8 bytes for linked pointer
+      int offset = evictionEnabled ? 24 : 8;
+
+      byte metadataType = MEMORY.getByte(address, offset);
+      offset += 1;
+      // hashCode
+      offset += 4;
+      // key length
+      int keyLenght = MEMORY.getInt(address, offset);
+      offset += 4;
```
wburns: This hasn't been updated with the new metadata changes I added. The metadata bytes could be present with a custom metadata.

https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/container/offheap/OffHeapEntryFactoryImpl.java#L246

core/src/main/java/org/infinispan/container/offheap/OffHeapEntryFactoryImpl.java
```diff
@@ -359,6 +360,55 @@ public boolean equalsKey(long address, WrappedBytes wrappedBytes) {
...
+      // 16 bytes for eviction if needed (optional)
+      // 8 bytes for linked pointer
+      int offset = evictionEnabled ? 24 : 8;
+
+      byte metadataType = MEMORY.getByte(address, offset);
+      offset += 1;
+      // hashCode
+      offset += 4;
+      // key length
+      int keyLenght = MEMORY.getInt(address, offset);
```
wburns: Typo in variable name.

core/src/main/java/org/infinispan/container/offheap/OffHeapEntryFactoryImpl.java
```diff
@@ -359,6 +360,55 @@ public boolean equalsKey(long address, WrappedBytes wrappedBytes) {
...
+      offset += 4;
+      // value length
+      offset += 4;
+      // key
+      offset += keyLenght;
+
+      byte[] metadataBytes;
+      long now = timeService.wallClockTime();
+      switch (metadataType) {
```
wburns: This isn't quite correct either as we could have VERSION or CUSTOM flag set on the type in which case this switch statement wouldn't work properly.

https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/container/offheap/OffHeapEntryFactoryImpl.java#L278

core/src/main/java/org/infinispan/container/offheap/OffHeapEntryFactoryImpl.java
```diff
@@ -359,6 +360,55 @@ public boolean equalsKey(long address, WrappedBytes wrappedBytes) {
...
+    * @param address the address of the entry's key to check
+    * @return {@code true} if the entry is expired, {@code false} otherwise
+    */
+   @Override
+   public boolean isExpired(long address) {
+      // 16 bytes for eviction if needed (optional)
+      // 8 bytes for linked pointer
+      int offset = evictionEnabled ? 24 : 8;
+
+      byte metadataType = MEMORY.getByte(address, offset);
```
wburns: In many cases we can just stop right after reading this. If the metadataType contains the IMMORTAL flag then we know it is not expired.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -286,7 +293,7 @@ public boolean containsKey(Object k) {
 
          while (address != 0) {
             long nextAddress = offHeapEntryFactory.getNext(address);
-            if (offHeapEntryFactory.equalsKey(address, wba)) {
+            if (offHeapEntryFactory.equalsKey(address, wba) && !offHeapEntryFactory.isExpired(address)) {
```
wburns: We should be able to short circuit this as we know that if the key equals then another can't equal.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -151,10 +152,11 @@ protected void checkDeallocation() {
 
    protected InternalCacheEntry<WrappedBytes, WrappedBytes> performGet(long address, Object k, boolean peek) {
       WrappedBytes wrappedKey = toWrapper(k);
+      long now = timeService.wallClockTime();
       while (address != 0) {
          long nextAddress = offHeapEntryFactory.getNext(address);
          InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(address);
-         if (wrappedKey.equalsWrappedBytes(ice.getKey())) {
+         if (wrappedKey.equalsWrappedBytes(ice.getKey()) && !ice.isExpired(now)) {
```
wburns: This can be short circuited as well so we don't have to iterate through all addresses.

core/src/main/java/org/infinispan/container/DefaultDataContainer.java
```diff
@@ -553,8 +553,9 @@ public void executeTask(final KeyFilter<? super K> filter, final BiConsumer<? su
       if (action == null)
          throw new IllegalArgumentException("No action specified");
 
+      long now = timeService.wallClockTime();
       entries.forEach((K key, InternalCacheEntry<K, V> value) -> {
-         if (filter.accept(key)) {
+         if (filter.accept(key) && !value.isExpired(now)) {
```
wburns: Good catch! I would say lets create another PR for the `DefaultDataContainer` as I see it is also broken for compute method as well. We should make sure to check all the other methods.

wburns:
Also as a side note this will conflict with https://github.com/infinispan/infinispan/pull/5583 :(

wburns:
I think I will just merge this into the other PR to resolve the conflicts. Leaving this one open until I do so.

wburns:
Closing this one in favor #5583

wburns:
Thanks again @vjuranek ~

PR #5568, opened 2017-11-06
ISPN-8489 Add compare-and-swap operation to StrongCounter

pruivo:
https://issues.jboss.org/browse/ISPN-8489

* compare-and-swap operation
* added documentation about remove()
* added documentation about getCounterNames()


counter/src/main/java/org/infinispan/counter/impl/strong/AbstractStrongCounter.java
```diff
@@ -124,7 +124,12 @@ public final String getName() {
 
    @Override
    public CompletableFuture<Boolean> compareAndSet(long expect, long update) {
-      return readWriteMap.eval(key, new CompareAndSetFunction<>(expect, update))
+      return compareAndSwap(expect, update).thenApply(value -> value == expect);
```
rvansa: Have you considered moving this to interface as a default method?

documentation/src/main/asciidoc/user_guide/counters.adoc
```diff
@@ -132,14 +132,19 @@ Default value is `Long.MAX_VALUE`.
...
 
+==== List counter names
+
+To list all the counters defined, the method `CounterManager.getCounterNames()` returns a collection of all counter
+names created cluster-wide.
+
 === The `CounterManager` interface.
 
-The `CounterManager` interface is the entry point to define and retrieve a counter.
+The `CounterManager` interface is the entry point to define, retrieve a counter and remove a counter.
```
rvansa: `define, retrieve and remove a counter` ?

documentation/src/main/asciidoc/user_guide/counters.adoc
```diff
@@ -157,46 +162,56 @@ EmbeddedCacheManager manager = ...;
 CounterManager counterManager = EmbeddedCounterManagerFactory.asCounterManager(manager);
 ----
 
+==== Remove a counter via CounterManager
+
+WARNING: use with caution.
+
+There is a difference between remove a counter via the `Strong/WeakCounter` interfaces and the `CounterManager`.
+The `CounterManager.remove(String)` removes the counter value from the cluster and removes all the listeners registered
```
rvansa: Please describe what a regular remove does when you already mention that there's a difference.

documentation/src/main/asciidoc/user_guide/counters.adoc
```diff
@@ -157,46 +162,56 @@ EmbeddedCacheManager manager = ...;
...
+NOTE: The counter is re-created if it is accessed after a removal.
+
 === The Counter
 
 A counter can be strong (`StrongCounter`) or weakly consistent (``WeakCounter``) and both is identified by a name.
 They have a specific interface but they share some logic, namely, both of them are asynchronous
 ( a `CompletableFuture` is returned by each operation), provide an update event and can be reset to its initial value.
 
+If you don't want to use the async API, there is a sync decorator that can be used.
+These are the `SyncStrongCounter` and the `SyncWeakCounter`.
```
rvansa: For discoverability it would be nice to add `SyncStrongCounter sync()` method to `StrongCounter`. Could be even default.

counter/src/main/java/org/infinispan/counter/impl/strong/AbstractStrongCounter.java
pruivo: totally forgot about default methods :+1:

documentation/src/main/asciidoc/user_guide/counters.adoc
pruivo: I'll have to move the `SyncStrongCounter` to commons. I guess it is ok.

pruivo:
@rvansa updated!

documentation/src/main/asciidoc/user_guide/counters.adoc
pruivo: done!

commons/src/main/java/org/infinispan/counter/api/SyncWeakCounter.java
```diff
@@ -1,11 +1,7 @@
-package org.infinispan.counter;
+package org.infinispan.counter.api;
```
rvansa: the same here

commons/src/main/java/org/infinispan/counter/api/Utils.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.infinispan.counter.exception.CounterException;
+
+/**
+ * Utility methods.
+ *
+ * @author Pedro Ruivo
+ * @since 9.2
+ */
+public final class Utils {
```
rvansa: This class should not be in a public package

commons/src/main/java/org/infinispan/counter/api/SyncStrongCounter.java
```diff
@@ -1,11 +1,7 @@
-package org.infinispan.counter;
+package org.infinispan.counter.api;
```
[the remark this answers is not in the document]
pruivo: :( you're right... but that is ugly

rvansa:
@pruivo Yep, keeping backwards compatibility is annoying. Don't forget to add removal as a task for Infinispan 10 https://issues.jboss.org/browse/ISPN-8070 when this gets in :)

Given that you made it an *interface* in commons... do you want to have second (identical) implementation in hotrod module?

pruivo:
updated! 
* added SyncWeak/StrongDecorator in commons (there are in private package and shouldn't be exposed)
* deprecated old ones
* added utility method in commons (await operation method)

rvansa:
@pruivo Sorry to grumble still, but why do you split that into interface and implementation? You could just have one class in commons in empty extending in counters. None of those classes should be final, so I don't see any issue.

pruivo:
I'm thinking about hot rod client. The client is synchronous (not sure if there are any plans to change that) and I'm supporting the strong/weak-counter interfaces with `CompletableFuture.supplyAsync()` methods. 

What I have in mind is to implement a Sync interfaces without paying the cost of creating `CompletableFuture`s. If you agree, I will remove the implementation from commons and un-deprecate the current implementation. 

Do you have any other suggestion?

rvansa:
@pruivo OK, I am thinking about the synchronicity of client as a temporary state, but you're right that the solution is not on the horizon yet (maybe I should put some muscle on that). So your proposal to actually have the non-trivial implementation in Hot Rod client makes sense.

pruivo:
updated!
* added default methods to the interface

pruivo:
updated!
* added default methods to the interface

pruivo:
@rvansa what should I do about the implementation in commons-module vs in counter-module?
should I leave it as it is now, assuming the hot rod client would be async and it will use the commons-module implementation...
or, should I remove the commons-module impl and keep only the counter-module?

rvansa:
@pruivo The implementation has no bounds to counter, therefore I would keep the common implementation (though, it's rather an **adapter** than a **decorator**) and remove only the methods in counter module:
`interface api.SyncStrongCounter <- class impl.SyncStrongCounterAdapter <- deprecated class SyncStrongCounter`

pruivo:
@rvansa updated! let me know if you more suggestions!

rvansa:
@pruivo Integrated, thanks for the patience!

PR #5570, opened 2017-11-07
ISPN-8376 Add DMR cluster counters management

vblagoje:
Preview only! Needs a thorough review by an expert in server DMR before integration.


vblagoje:
Please test it yourself and help me out to conclude this one. One would test it by building a server and starting it. Upon server start, add a counter configuration through CLI:
```
[domain@localhost:9990 /] cd /profile=clustered/subsystem=datagrid-infinispan/cache-container=clustered/configurations=CONFIGURATIONS/counters=COUNTERS
[domain@localhost:9990 counters=COUNTERS] ./strong-counter=c:add(name=c, initial-value=1, storage=VOLATILE, lower-bound=1, upper-bound=32,type=BOUNDED_STRONG)
{
    "outcome" => "success",
    "result" => undefined,
    "server-groups" => {"cluster" => {"host" => {"master" => {
        "server-one" => {"response" => {"outcome" => "success"}},
        "server-two" => {"response" => {"outcome" => "success"}}
    }}}}
}
[domain@localhost:9990 counters=COUNTERS] cd strong-counter=
[domain@localhost:9990 strong-counter] ls
c  

[domain@localhost:9990 strong-counter] cd c
[domain@localhost:9990 strong-counter=c] ls
initial-value=1      lower-bound=1        name=c               storage=VOLATILE     type=BOUNDED_STRONG  upper-bound=32       

[domain@localhost:9990 strong-counter=c] 
```
After adding counter configuration, in the above case configuration "c", you can then add a counter(s):
```
[domain@localhost:9990 counters=COUNTERS] ./strong-counter=a:add(configuration=c)
{
    "outcome" => "success",
    "result" => undefined,
    "server-groups" => {"cluster" => {"host" => {"master" => {
        "server-one" => {"response" => {"outcome" => "success"}},
        "server-two" => {"response" => {"outcome" => "success"}}
    }}}}
}
[domain@localhost:9990 counters=COUNTERS] ./strong-counter=b:add(configuration=c)
{
    "outcome" => "success",
    "result" => undefined,
    "server-groups" => {"cluster" => {"host" => {"master" => {
        "server-one" => {"response" => {"outcome" => "success"}},
        "server-two" => {"response" => {"outcome" => "success"}}
    }}}}
}
[domain@localhost:9990 counters=COUNTERS] pwd
/profile=clustered/subsystem=datagrid-infinispan/cache-container=clustered/counters=COUNTERS
[domain@localhost:9990 counters=COUNTERS] cd strong-counter=
[domain@localhost:9990 strong-counter] ls
a  b  

[domain@localhost:9990 strong-counter] cd a
[domain@localhost:9990 strong-counter=a] ls
configuration=c  

[domain@localhost:9990 strong-counter=a]
```
On the runtime side, a counter looks like this:
```
[domain@localhost:9990 /] cd /host=master/server=server-one/subsystem=datagrid-infinispan/cache-container=clustered/co
configurations  counters  
[domain@localhost:9990 /] cd /host=master/server=server-one/subsystem=datagrid-infinispan/cache-container=clustered/counters=COUNTERS
[domain@localhost:9990 counters=COUNTERS] cd strong-counter=
a  b  
[domain@localhost:9990 counters=COUNTERS] cd strong-counter=a
[domain@localhost:9990 strong-counter=a] ls
configuration=c  value=1          

[domain@localhost:9990 strong-counter=a] 
```
As you can see the value of the counter is 1. Let's increase is by 1.
```
[domain@localhost:9990 strong-counter=a] :counter-increase()
{
    "outcome" => "success",
    "result" => undefined
}
[domain@localhost:9990 strong-counter=a] ls
configuration=c  value=2          

[domain@localhost:9990 strong-counter=a] 
```

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerCountersConfigurationResource.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jboss.as.controller.services.path.ResolvePathHandler;
+
+/**
+ * CacheContainerCountersResource
+ *
+ * @author Vladimir Blagojevic 
+ * @since 9.2
+ */
+public class CacheContainerCountersConfigurationResource extends SimpleResourceDefinition {
+    static final PathElement PATH = PathElement.pathElement(ModelKeys.COUNTERS, ModelKeys.COUNTERS_NAME);
```
tristantarrant: You are using the same path for the configuration and for the concrete counters below. They should be separate.

tristantarrant:
This looks pretty good. It is however missing changes to the XSD as well as adding some tests to SubsystemParsingTestCase

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,253 @@
...
+    private static final OperationDefinition COUNTER_INCREASE = buildOperation("counter-increase").build();
+    private static final OperationDefinition COUNTER_DECREASE = buildOperation("counter-decrease").build();
+
+    private final boolean runtimeRegistration;
+    private final ResolvePathHandler resolvePathHandler;
+
+    public CounterResource(PathElement pathElement, ResourceDescriptionResolver descriptionResolver,
+            ResolvePathHandler resolvePathHandler, AbstractAddStepHandler addHandler,
+            OperationStepHandler removeHandler, boolean runtimeRegistration) {
+        super(pathElement, descriptionResolver, addHandler, removeHandler);
```
ryanemerson: The calls to `CounterResource` in both `Weak/StrongCounterResource` are exactly the same except for pathElement, so you can use the following constructor instead:

```java
public CounterResource(PathElement pathElement, ResourceDescriptionResolver descriptionResolver,
                           ResolvePathHandler resolvePathHandler, boolean runtimeRegistration) {
        super(pathElement, descriptionResolver, new CounterAddHandler(), new CounterRemoveHandler());
        this.resolvePathHandler = resolvePathHandler;
        this.runtimeRegistration = runtimeRegistration;
    }
```

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,253 @@
...
+    static final AttributeDefinition[] ATTRIBUTES = { CONFIGURATION };
+
+    // operations
+
+    private static final OperationDefinition COUNTER_RESET = buildOperation("counter-reset").build();
+    private static final OperationDefinition COUNTER_INCREASE = buildOperation("counter-increase").build();
+    private static final OperationDefinition COUNTER_DECREASE = buildOperation("counter-decrease").build();
+
+    private final boolean runtimeRegistration;
+    private final ResolvePathHandler resolvePathHandler;
```
ryanemerson: Unused variable `resolvePathHandler`

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
ryanemerson: Or you could go further and remove `Weak/StrongCounterResource` entirely and just create two `CounterResource` instances with the desired paths in `CacheContainerCountersResource`

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,253 @@
...
+
+    private static OperationFailedException counterManagerNotFound() {
+        return new OperationFailedException("CounterManager not found in server.");
+    }
+
+    public static class CounterRemoveCommand extends BaseCounterManagerCommand {
+        public static final CounterRemoveCommand INSTANCE = new CounterRemoveCommand();
+
+        private CounterRemoveCommand() {
+            super();
```
ryanemerson: Redundant constructor

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,253 @@
...
+            counterManager.remove(counterName);
+            return new ModelNode();
+        }
+    }
+
+    private static class CounterResetCommand extends BaseCounterManagerCommand {
+        private static final CounterResetCommand INSTANCE = new CounterResetCommand();
+
+        private CounterResetCommand() {
+            super();
```
ryanemerson: Redundant constructor

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
```diff
@@ -349,15 +358,190 @@ private void parseContainer(XMLExtendedStreamReader reader, PathAddress subsyste
...
+        for (int i = 0; i < reader.getAttributeCount(); i++) {
+            String value = reader.getAttributeValue(i);
+            Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
+            switch (attribute) {
+                case NUM_OWNERS: {
+                    //TODO
+                    break;
+                }
+                case RELIABILITY: {
+                    //TODO
```
ryanemerson: Don't forget this :smile:

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
```diff
@@ -349,15 +358,190 @@ private void parseContainer(XMLExtendedStreamReader reader, PathAddress subsyste
...
+            throws XMLStreamException {
+
+        PathAddress strongCounterAddress = countersConfigurationAddress;
+        ModelNode counter = Util.createAddOperation(strongCounterAddress);
+        for (int i = 0; i < reader.getAttributeCount(); i++) {
+            String value = reader.getAttributeValue(i);
+            Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
+            switch (attribute) {
+                case INITIAL_VALUE: {
+                    ((SimpleAttributeDefinition) StrongCounterConfigurationResource.INITIAL_VALUE)
```
ryanemerson: If you declared these as a `SimpleAttributeDefinition` in `CounterConfigurationResource` there is no need to cast

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/StrongCounterConfigurationAddHandler.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Add operation handler for /subsystem=infinispan/cache-container=clustered/counter=*
+ * 
+ * @author Vladimir Blagojevic
+ * 
+ */
+public class StrongCounterConfigurationAddHandler extends CounterConfigurationAddHandler {
+
+   StrongCounterConfigurationAddHandler() {
+      super();
```
ryanemerson: redundant

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/WeakCounterConfigurationAddHandler.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Add operation handler for /subsystem=infinispan/cache-container=clustered/counter=*
+ * 
+ * @author Vladimir Blagojevic
+ * 
+ */
+public class WeakCounterConfigurationAddHandler extends CounterConfigurationAddHandler {
+
+   WeakCounterConfigurationAddHandler() {
+      super();
```
ryanemerson: Redundant

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/StrongCounterConfigurationResource.java
```diff
@@ -0,0 +1,91 @@
...
+            resolvePathHandler, new StrongCounterConfigurationAddHandler(), new CounterConfigurationRemoveHandler(), runtimeRegistration);
+   }
+
+   @Override
+   public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
+      super.registerAttributes(resourceRegistration);
+      
+      final OperationStepHandler writeHandler = new ReloadRequiredWriteAttributeHandler(STRONG_ATTRIBUTES);
+      for (AttributeDefinition attr : STRONG_ATTRIBUTES) {
+          resourceRegistration.registerReadWriteAttribute(attr, null, writeHandler);
```
ryanemerson: We can reduce the repetition found in here, `CounterConfigurationResource` and `WeakCounterConfigurationResource` by following a similar approach to `BaseStoreConfigurationResource`

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerCountersConfigurationResource.java
vblagoje: Indeed, but I am adding them to different parent path. Configurations are at:
```
/profile=clustered/subsystem=datagrid-infinispan/cache-container=clustered/configurations=CONFIGURATIONS/counters=COUNTERS
```
while counter instances are at:
```
/profile=clustered/subsystem=datagrid-infinispan/cache-container=clustered/counters=COUNTERS
```
A bit confusing but a result of our discusssion on infinispan-dev. See details at http://lists.jboss.org/pipermail/infinispan-dev/2017-November/017917.html

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterAddHandler.java
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Add operation handler for /subsystem=infinispan/cache-container=clustered/counter=*
+ *
+ * @author Vladimir Blagojevic
+ *
+ */
+public class CounterAddHandler extends AbstractAddStepHandler {
+
+    CounterAddHandler() {
+        super();
```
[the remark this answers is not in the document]
ryanemerson: Sorry I mean the constructor that takes `AttributeDefinition... attributes`. So the code becomes

```java
CounterAddHandler() {
        super(CounterResource.ATTRIBUTES);
    }
```

You can then remove `populateModel` as this code already exists in `AbstractAddStepHandler::populateModel`

pruivo:
@ryanemerson the exception isn't because the DMR. I detected it when I was implementing the counter in hot rod. 
fix: https://github.com/infinispan/infinispan/pull/5552/files#diff-3d610fd97d822a830ccad8fba4c9da29R171

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterAddHandler.java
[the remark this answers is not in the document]
vblagoje: Right, that one works. Thanks man

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -142,7 +124,6 @@ private static OperationFailedException counterManagerNotFound() {
         public static final CounterRemoveCommand INSTANCE = new CounterRemoveCommand();
 
         private CounterRemoveCommand() {
```
ryanemerson: There is no need for this constructor, or any of the constructors in the classes that extend `BaseCounterManagerCommand`

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -78,16 +56,20 @@
     private static final OperationDefinition COUNTER_DECREASE = buildOperation("counter-decrease").build();
 
     private final boolean runtimeRegistration;
-    private final ResolvePathHandler resolvePathHandler;
 
     public CounterResource(PathElement pathElement, ResourceDescriptionResolver descriptionResolver,
```
ryanemerson: No need for this constructor anymore, just do the variable assignment and calls to super in the constructor:
 ```java
public CounterResource(PathElement pathElement, ResolvePathHandler resolvePathHandler,
            boolean runtimeRegistration) {
...
}
```

**Comment by vblagoje** [comment]
Thanks a bunch @ryanemerson Have not programmed in Java in a while and it shows :-)

**Comment by vblagoje** [comment]
@pruivo @tristantarrant @ryanemerson added cluster-wide num_owners and reliability runtime read attributes of counters. Pedro told me these are only read attributes and should not be changed at runtime. Please give it all another pass and let's integrate this one...if there are no more objections. Thank you guys

File: server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterConfigurationRemoveHandler.java [hunk 14]
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author Vladimir Blagojevic
+ */
+public class CounterConfigurationRemoveHandler extends AbstractRemoveStepHandler {
+
+   @Override
+   protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model)
+         throws OperationFailedException {
+      //CounterConfigurationResource.CounterRemoveCommand.INSTANCE.execute(context, operation);
+      //TODO
```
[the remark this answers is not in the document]
ryanemerson: This still needs implementing

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
ryanemerson: This still needs completing before we can merge

server/integration/infinispan/src/main/resources/schema/jboss-infinispan-core_9_2.xsd
```diff
@@ -371,6 +371,115 @@
         </xs:sequence>
     </xs:complexType>
 
+	<xs:complexType name="counters">
```
ryanemerson: Needs formatting to match the rest of the file

server/integration/infinispan/src/main/resources/schema/jboss-infinispan-core_9_2.xsd
```diff
@@ -371,6 +371,115 @@
...
+			<xs:element name="weak-counter" type="tns:weak-counter"
+				minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>
+						Defines strong counter
+					</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+		</xs:sequence>
+		<xs:attribute name="num-owners" type="xs:int" default="2"
```
ryanemerson: For the default values you should placeholders like other defaults in the file. For example `${Server.CacheContainer.start}`. To find the placeholder for the counter defaults, look in `target/defaults.properties` after making the server and you should see the placeholder text and value for `num-owners` and all other configured resources that have default values defined

ryanemerson: Although, looking at the `num-owners` resource, it seems you haven't set the default value via `setDefaultValue(...)`

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
ryanemerson: We should really remove all of these casts in this file as they aren't necessary if ^ action is taken

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerCountersResource.java
```diff
@@ -15,6 +20,30 @@
 public class CacheContainerCountersResource extends SimpleResourceDefinition {
     static final PathElement PATH = PathElement.pathElement(ModelKeys.COUNTERS, ModelKeys.COUNTERS_NAME);
 
+    //TODO Could not find these definitions anywhere in counter API
+    private static final String AVAILABLE = "AVAILABLE";
+    private static final String CONSISTENT = "CONSISTENT";
```
ryanemerson: These constants are defined in the following enum `org.infinispan.counter.configuration.Reliability`

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerCountersResource.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @author Vladimir Blagojevic
+ * @since 9.2
+ */
+public class CacheContainerCountersResource extends SimpleResourceDefinition {
+    static final PathElement PATH = PathElement.pathElement(ModelKeys.COUNTERS, ModelKeys.COUNTERS_NAME);
+
+    //TODO Could not find these definitions anywhere in counter API
+    private static final String AVAILABLE = "AVAILABLE";
+    private static final String CONSISTENT = "CONSISTENT";
```
ryanemerson: These constants are defined in the following enum `org.infinispan.counter.configuration.Reliability`

server/integration/infinispan/src/main/resources/schema/jboss-infinispan-core_9_2.xsd
vblagoje: @ryanemerson I put the placeholders in schema file, build the server, but then I do not see the values generated in target/default.properties?

vblagoje: How can this work, how can processing build program figure out the defaults for the placeholders?

vblagoje: Do I have to name placeholders according to some naming scheme?

ryanemerson: This is all done by the [defaults plugin](https://github.com/infinispan/infinispan/tree/master/plugins/infinispan-defaults-maven-plugin). During the build process, this plugin extracts the defined defaults of all infinispan `AttributeDefinition` or jboss `SimpleAttributeDefinition` and outputs them to the `target/defaults.properties` file. 

Property keys are the name of the class which contains the definition (minus Resource at the end), concatenated with the name of the attribute. Furthermore in the case of jboss resources, the key is pre-pended with `Server`. So if you update the NUM_OWNERS to the following:

```java
static final SimpleAttributeDefinition NUM_OWNERS = new SimpleAttributeDefinitionBuilder(ModelKeys.NUM_OWNERS,
            ModelType.LONG, false)
            .setXmlName(Attribute.NUM_OWNERS.getLocalName())
            .setDefaultValue(new ModelNode(2)) // You need to add this for a default value to be defined
            .setAllowExpression(false)
            .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
            .build();
```

Then the placeholder key, which you should use in the xsd will be `${Server.CacheContainerCounters.num_owners}`. 

FYI all of the successfully extracted defaults for the server will be located at `$INFINISPAN_ROOT/server/integration/infinispan/target/defaults.properties`

vblagoje: Wow, awesome! Thanks for the explanation @ryanemerson

vblagoje:
@ryanemerson @tristantarrant @pruivo Phew, I think this is now ready to go, it is updated, all tests pass, I created a default config with example counters. CounterService class deals with install of counters using dependencies (we need cache container running to define and instantiate counters). Boot sequence works, adding and removing counters and seeing their values at runtime. I have used the non-overengineered version of DMR proposal http://lists.jboss.org/pipermail/infinispan-dev/2017-November/017917.html We can always update the current proposal in the future if there is a real need for our users. Please have a look! Much appreciated your continued assistance!

vblagoje:
@ryanemerson @tristantarrant I discovered some subtle bugs while developing UI for counters. Give me until EOD before looking into this one.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CountersMetricsHandler.java
```diff
@@ -0,0 +1,90 @@
...
+    private CounterManagerConfiguration extractCounterManagerConfiguration(EmbeddedCacheManager cacheManager) {
+        GlobalComponentRegistry globalComponentRegistry = SecurityActions.getGlobalComponentRegistry(cacheManager);
+        CounterManagerConfiguration config = globalComponentRegistry.getGlobalConfiguration()
+                .module(CounterManagerConfiguration.class);
+        return config == null ? CounterManagerConfigurationBuilder.defaultConfiguration() : config;
+    }
+
+    public enum CounterConfigurationMetrics {
+
+        NUM_OF_OWNERS(MetricKeys.NUM_OF_OWNERS, ModelType.LONG),
```
pruivo: can't you reuse the attribute definition in `CacheContainerCountersResource`?

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterMetricsHandler.java
```diff
@@ -0,0 +1,88 @@
...
+                .getService(CacheContainerServiceName.CACHE_CONTAINER.getServiceName(cacheContainerName));
+
+        Long value = null;
+        if (controller != null) {
+            DefaultCacheContainer cacheManager = (DefaultCacheContainer) controller.getValue();
+            CounterManager counterManager = EmbeddedCounterManagerFactory.asCounterManager(cacheManager);
+            if (ModelKeys.STRONG_COUNTER.equals(counterType)) {
+                StrongCounter sc = counterManager.getStrongCounter(counterName);
+                try {
+                    value = sc.getValue().get();
```
pruivo: you can use the sync counter if you don't want to hanled checked exceptions:
```
sc.sync().getValue()
```

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,247 @@
...
+/**
+ * Resource description for the addressable resource
+ * /subsystem=infinispan/cache-container=X/counter=COUNTERS
+ *
+ * @author Vladimir Blagojevic
+ * @since 9.2
+ */
+public class CounterResource extends SimpleResourceDefinition {
+
+    //atributes
```
pruivo: typo :)

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerCountersResource.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * CacheContainerCountersResource
+ *
+ * @author Vladimir Blagojevic
+ * @since 9.2
+ */
+public class CacheContainerCountersResource extends SimpleResourceDefinition {
+    static final PathElement PATH = PathElement.pathElement(ModelKeys.COUNTERS, ModelKeys.COUNTERS_NAME);
+
+    //atributes
```
pruivo: typo :)

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,247 @@
...
+    // operations
+
+    private static final OperationDefinition COUNTER_RESET = buildOperation("counter-reset").build();
+    private static final OperationDefinition COUNTER_INCREASE = buildOperation("counter-increase").build();
+    private static final OperationDefinition COUNTER_DECREASE = buildOperation("counter-decrease").build();
+
+    private final boolean runtimeRegistration;
+
+    public CounterResource(PathElement pathElement, ResourceDescriptionResolver descriptionResolver,
+            ResolvePathHandler resolvePathHandler, AbstractAddStepHandler addHandler,
```
pruivo: unused `resolvePathHandler`. it can be removed.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,247 @@
...
+        @Override
+        protected ModelNode invoke(CounterManager counterManager, OperationContext context, ModelNode operation)
+                throws Exception {
+            final String counterName = counterName(context, operation);
+            final String counterType = counterType(context, operation);
+            if (counterManager.isDefined(counterName)) {
+                boolean isStrongCounter = ModelKeys.STRONG_COUNTER.equals(counterType);
+                if (isStrongCounter) {
+                    StrongCounter strongCounter = counterManager.getStrongCounter(counterName);
+                    strongCounter.reset();
```
pruivo: the `reset()` is asynchronous! you are returning while the operation is executed. I'm not sure if it is on purpose or not. If not, use the `sync()` method.
The same note for the increment/decrement

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,247 @@
...
+                    CounterDecreaseCommand.INSTANCE);
+        }
+    }
+
+    private static SimpleOperationDefinitionBuilder buildOperation(String name) {
+        return new SimpleOperationDefinitionBuilder(name, new InfinispanResourceDescriptionResolver(ModelKeys.COUNTERS))
+                .setRuntimeOnly();
+    }
+
+    private static PathElement counterElement(OperationContext context, ModelNode operation)
```
pruivo: unused `context`

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterService.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Override
+    public void start(StartContext context) {
+        EmbeddedCacheManager container = this.dependencies.getCacheContainer();
+        CounterManager counterManager = EmbeddedCounterManagerFactory.asCounterManager(container);
+
+        //define counter
+        counterManager.defineCounter(counterConfigurationName, counterConfiguration);
+
+        //but also instantiate the counter as well
```
pruivo: why do you need to instantiate?

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterService.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void stop(StopContext context) {
+        // remove and destroy the counter instance
+        EmbeddedCacheManager container = this.dependencies.getCacheContainer();
+        CounterManager counterManager = EmbeddedCounterManagerFactory.asCounterManager(container);
+        counterManager.remove(counterConfigurationName);
```
pruivo: `remove()` only removes the counter value. the counter is still defined. just put a comment to be aware in the future. an undefined operation is needed...

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
```diff
@@ -31,6 +31,7 @@
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
```
pruivo: unused import.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLWriter.java
```diff
@@ -204,7 +207,69 @@ private static void writeScheduledThreadPoolElements(Element element, ScheduledT
...
+         for (Property e : configurations.asPropertyList()) {
+            processStrongCounterConfiguration(writer, e.getValue());
+         }
+      }
+   }
+
+   private void processWeakCounterConfiguration(XMLExtendedStreamWriter writer, ModelNode weakConfiguration) throws XMLStreamException {
+      writer.writeStartElement(Element.WEAK_COUNTER.getLocalName());
+      this.writeRequired(writer, Attribute.NAME, weakConfiguration, ModelKeys.NAME);
+      this.writeRequired(writer, Attribute.INITIAL_VALUE, weakConfiguration, ModelKeys.INITIAL_VALUE);
```
pruivo: the only required attribute is the name. the remaining attributes are optional.

server/integration/infinispan/src/main/resources/org/jboss/as/clustering/infinispan/subsystem/LocalDescriptions.properties
```diff
@@ -639,3 +639,54 @@ datagrid-infinispan.cache-container.health.number-of-nodes=Number of nodes in th
...
+
+datagrid-infinispan.cache-container.counters.name=The counter's name
+datagrid-infinispan.cache-container.counters.type=The counter's type (weak, strong bounded or strong unbounded)
+datagrid-infinispan.cache-container.counters.storage=The counter's storage mode (volatile or persistent)
+datagrid-infinispan.cache-container.counters.initial-value=The counter's initial value
+datagrid-infinispan.cache-container.counters.upper-bound=The counter's upper-bound, if strong and bounded
+datagrid-infinispan.cache-container.counters.lower-bound=The counter's lower-bound, if strong and bounded
+datagrid-infinispan.cache-container.counters.concurrency-level=The counter's concurrency-level, if weak
+datagrid-infinispan.cache-container.counters.value=The counter's value
+datagrid-infinispan.cache-container.counters.counter-reset=Resets the counter value to 0
```
pruivo: it resets the counter to its inital value, not 0.

server/integration/infinispan/src/main/resources/org/jboss/as/clustering/infinispan/subsystem/LocalDescriptions.properties
```diff
@@ -639,3 +639,54 @@ datagrid-infinispan.cache-container.health.number-of-nodes=Number of nodes in th
...
+datagrid-infinispan.cache-container.counters.counter-increase=Increase the counter value by 1
+datagrid-infinispan.cache-container.counters.counter-decrease=Decrease the counter value by 1
+datagrid-infinispan.cache-container.counters.counter-set=Set counter value
```
pruivo: we don't have a set.

server/integration/infinispan/src/main/resources/org/jboss/as/clustering/infinispan/subsystem/LocalDescriptions.properties
```diff
@@ -639,3 +639,54 @@ datagrid-infinispan.cache-container.health.number-of-nodes=Number of nodes in th
...
+datagrid-infinispan.cache-container.counters.counter-set.initial-value=Set counter value
```
pruivo: this is the same as reset

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/StrongCounterResource.java
```diff
@@ -0,0 +1,57 @@
...
+import org.jboss.as.controller.registry.AttributeAccess;
+import org.jboss.as.controller.registry.ManagementResourceRegistration;
+import org.jboss.as.controller.services.path.ResolvePathHandler;
+import org.jboss.dmr.ModelType;
+
+/**
+ * Resource description for the addressable resource
+ * /subsystem=infinispan/cache-container=X/counter=*
+ *
+ * @author Pedro Ruivo
```
pruivo: I guess  it can be removed. The credits are all yours!

pruivo:
thanks a lot for your work @vblagoje !

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterMetricsHandler.java
vblagoje: @pruivo I can't find sync() method on StrongCounter?

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CountersMetricsHandler.java
vblagoje: I think I can't because these need to be set for runtime, the other are for non-runtime

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterMetricsHandler.java
pruivo: https://github.com/infinispan/infinispan/blob/master/commons/src/main/java/org/infinispan/counter/api/StrongCounter.java#L117

rebase you PR. it was introduced recently in master!

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLWriter.java
vblagoje: ok, thanks @pruivo

vblagoje:
Responded to @pruivo comments. Should be gtg now!

slaskawi:
Killing CI run because of the update. I will schedule it right after I'm done.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
```diff
@@ -349,15 +349,183 @@ private void parseContainer(XMLExtendedStreamReader reader, PathAddress subsyste
...
+                    WeakCounterResource.CONCURRENCY_LEVEL.parseAndSetParameter(value,
+                            counter, reader);
+                    break;
+                }
+                default: {
+                    throw ParseUtils.unexpectedAttribute(reader, i);
+                }
+            }
+        }
+        while (reader.hasNext() && (reader.nextTag() != XMLStreamConstants.END_ELEMENT)) {
```
pruivo: can be replace with an `if` since it never loops

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
```diff
@@ -349,15 +349,183 @@ private void parseContainer(XMLExtendedStreamReader reader, PathAddress subsyste
...
+                    StrongCounterResource.UPPER_BOUND.parseAndSetParameter(value, counter, reader);
+                }
+                break;
+            }
+            default: {
+                throw ParseUtils.unexpectedAttribute(reader, i);
+            }
+            }
+        }
+        while (reader.hasNext() && (reader.nextTag() != XMLStreamConstants.END_ELEMENT)) {
```
pruivo: can be replace with an `if` since it never loops

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,246 @@
...
+import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.ReloadRequiredWriteAttributeHandler;
+import org.jboss.as.controller.SimpleAttributeDefinition;
+import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
+import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
+import org.jboss.as.controller.SimpleResourceDefinition;
+import org.jboss.as.controller.descriptions.ResourceDescriptionResolver;
+import org.jboss.as.controller.registry.AttributeAccess;
+import org.jboss.as.controller.registry.ManagementResourceRegistration;
+import org.jboss.as.controller.services.path.ResolvePathHandler;
```
pruivo: unused import

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,246 @@
...
+
+    private static SimpleOperationDefinitionBuilder buildOperation(String name) {
+        return new SimpleOperationDefinitionBuilder(name, new InfinispanResourceDescriptionResolver(ModelKeys.COUNTERS))
+                .setRuntimeOnly();
+    }
+
+    private static PathElement counterElement(ModelNode operation)
+            throws OperationFailedException {
+        final PathAddress address = pathAddress(operation.require(OP_ADDR));
+        final PathElement counterElement = address.getElement(address.size() - 1);
```
pruivo: you don't need the `counterElement`: `return address.getElement(address.size() - 1);`

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterMetricsHandler.java
```diff
@@ -0,0 +1,110 @@
...
+                WeakCounter wc = counterManager.getWeakCounter(counterName);
+                value = wc.getValue();
+            }
+            result.set(value);
+        }
+        context.getResult().set(result);
+    }
+
+    public void registerMetrics(ManagementResourceRegistration container) {
+        for (CounterMetrics metric : CounterMetrics.values()) {
```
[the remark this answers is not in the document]
pruivo: I agree with @ryanemerson.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/WeakCounterAddHandler.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Add operation handler for /subsystem=infinispan/cache-container=clustered/counter=*
+ *
+ * @author Vladimir Blagojevic
+ *
+ */
+public class WeakCounterAddHandler extends CounterAddHandler {
+
+   @Override
+   protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model)
```
pruivo: can be removed. only invokes `super.*`

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterAddHandler.java
```diff
@@ -0,0 +1,161 @@
...
+        int size = counterAddress.size();
+        PathAddress subAddress = counterAddress.subAddress(size - 1, size);
+        return subAddress.getLastElement().getKey();
+    }
+
+    private PathAddress getCounterAddressFromOperation(ModelNode operation) {
+        return PathAddress.pathAddress(operation.get(OP_ADDR));
+    }
+
+    private Builder getBuilder(OperationContext context, ModelNode counter, String counterType)
```
pruivo: `context` never used and it can be removed. 
`OperationFailedException` never thrown and it can be removed.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+
+    private static SimpleOperationDefinitionBuilder buildOperation(String name) {
+        return new SimpleOperationDefinitionBuilder(name, new InfinispanResourceDescriptionResolver(ModelKeys.COUNTERS))
+                .setRuntimeOnly();
+    }
+
+    private static PathElement counterElement(ModelNode operation)
+            throws OperationFailedException {
```
pruivo: `OperationFailedException` never thrown and it can be removed

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
```diff
@@ -349,15 +349,183 @@ private void parseContainer(XMLExtendedStreamReader reader, PathAddress subsyste
...
+                    break;
+                }
+                default: {
+                    throw ParseUtils.unexpectedElement(reader);
+                }
+            }
+        }
+    }
+
+    private void parseStrongCounterElement(XMLExtendedStreamReader reader, Element element,
```
pruivo: 'element' never used and it can be removed.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
```diff
@@ -349,15 +349,183 @@ private void parseContainer(XMLExtendedStreamReader reader, PathAddress subsyste
...
                 default: {
                     throw ParseUtils.unexpectedElement(reader);
                 }
             }
         }
+        operations.put(strongCounterAddress, counter);
+
+    }
+
+    private void parseWeakCounterElement(XMLExtendedStreamReader reader, Element element,
```
pruivo: 'element' never used and it can be removed.

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/spi/service/CounterServiceNameFactory.java
```diff
@@ -0,0 +1,18 @@
+package org.infinispan.server.infinispan.spi.service;
+
+import org.jboss.msc.service.ServiceName;
+
+/**
+ * Factory for generating service names for services associated with a counter.
+ * @author Vladimir Blagojevic
+ */
+public interface CounterServiceNameFactory {
```
pruivo: I'm not sure but this interface isn't needed (IMO).
It is only implemented by enum `CounterServiceName.CONFIGURATION`

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerCountersResource.java
```diff
@@ -0,0 +1,72 @@
...
+    static final SimpleAttributeDefinition NUM_OWNERS = new SimpleAttributeDefinitionBuilder(ModelKeys.NUM_OWNERS,
+            ModelType.LONG, false)
+            .setXmlName(Attribute.NUM_OWNERS.getLocalName())
+            .setAllowExpression(false)
+            .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
+            .setDefaultValue(new ModelNode().set(CounterManagerConfigurationBuilder.defaultConfiguration().numOwners()))
+            .build();
+
+
+    static final AttributeDefinition[] ATTRIBUTES = { RELIABILITY, NUM_OWNERS };
```
pruivo: unused and can be removed.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerCountersResource.java
```diff
@@ -0,0 +1,72 @@
...
+
+
+    private final boolean runtimeRegistration;
+    private final ResolvePathHandler resolvePathHandler;
```
pruivo: unused and can be removed. Can be removed from the constructor as well.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,246 @@
...
+
+    static final SimpleAttributeDefinition INITIAL_VALUE = new SimpleAttributeDefinitionBuilder(ModelKeys.INITIAL_VALUE,
+          ModelType.LONG, true)
+          .setXmlName(Attribute.INITIAL_VALUE.getLocalName())
+          .setAllowExpression(false)
+          .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
+          .setDefaultValue(new ModelNode().set(0))
+          .build();
+
+    static final SimpleAttributeDefinition VALUE = new SimpleAttributeDefinitionBuilder(ModelKeys.VALUE,
```
pruivo: unused and can be removed.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,246 @@
...
+        super(pathElement, descriptionResolver, addHandler, removeHandler);
+        this.runtimeRegistration = runtimeRegistration;
+    }
+
+    @Override
+    public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
+        super.registerAttributes(resourceRegistration);
+        final OperationStepHandler writeHandler = new ReloadRequiredWriteAttributeHandler(ATTRIBUTES);
+        for (AttributeDefinition attr : ATTRIBUTES) {
+            resourceRegistration.registerReadWriteAttribute(attr, null, writeHandler);
```
pruivo: can you explain what is happening here? the counter-name, storage and initial-value can't be changed and they are set when you define a counter.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
pruivo: up ^

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    private static abstract class BaseCounterManagerCommand extends CacheContainerCommands {
+
+        BaseCounterManagerCommand() {
+            //path to container from counter address has two elements
+            super(2);
+        }
+
+        abstract ModelNode invoke(CounterManager counterManager, OperationContext context, ModelNode operation)
```
pruivo: `context` not used and it can be removed.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterMetricsHandler.java
```diff
@@ -0,0 +1,88 @@
...
+            DefaultCacheContainer cacheManager = (DefaultCacheContainer) controller.getValue();
+            CounterManager counterManager = EmbeddedCounterManagerFactory.asCounterManager(cacheManager);
+            if (ModelKeys.STRONG_COUNTER.equals(counterType)) {
+                StrongCounter sc = counterManager.getStrongCounter(counterName);
+                try {
+                    value = sc.getValue().get();
+                } catch (Exception e) {
+                    throw new OperationFailedException("Can not get value from counter " + sc, e);
+                }
+            } else if (ModelKeys.WEAK_COUNTER.equals(counterType)) {
```
pruivo: I'm not sure if it is worth to check for weak counter.

ryanemerson:
@vblagoje as well as @pruivo's feedback there are a few of my comments that still need addressing from previous reviews

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterService.java
pruivo: you don't need, except if you want to cache the implementation. the getStrong/WeakCounter doesn't interact with the cache, just instantiate the implementation

vblagoje: Let's cache it! That's what I intended from the beginning :-)

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/spi/service/CounterServiceNameFactory.java
vblagoje: It is, believe me! I did not want to use hacks in other service factories and mix apples and oranges. All it does it appends container name with counter name. And we need that for the uniqueness of counter. I could have hacked it with other service factories but it would be a kludge

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterService.java
vblagoje: I made a note

server/integration/infinispan/src/main/java/org/infinispan/server/infinispan/spi/service/CounterServiceNameFactory.java
ryanemerson: Why not just have: 
```java
public class CounterServiceName {
    public static ServiceName getServiceName(String container, String counter) {
        return CacheContainerServiceName.CACHE_CONTAINER.getServiceName(container).append(counterName)
    }
}
```
Or you could just add ^ method to `CounterService.java`

Either way there is no need for the interface.

vblagoje:
Ok, I thought I had to follow the pattern here..

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
pruivo: you forgot this comment ^

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterMetricsHandler.java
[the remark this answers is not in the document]
pruivo: @vblagoje any comments on this ^

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/StrongCounterResource.java
pruivo: don't forget this :)

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/StrongCounterResource.java
```diff
@@ -0,0 +1,54 @@
...
+   public static final PathElement PATH = PathElement.pathElement(ModelKeys.STRONG_COUNTER);
+
+   public StrongCounterResource(ResolvePathHandler resolvePathHandler, boolean runtimeRegistration) {
+      super(StrongCounterResource.PATH, new InfinispanResourceDescriptionResolver(ModelKeys.COUNTERS),
+            resolvePathHandler, new CounterAddHandler(), new CounterRemoveHandler(), runtimeRegistration);
+   }
+
+   @Override
+   public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
+      super.registerAttributes(resourceRegistration);
```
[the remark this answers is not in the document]
pruivo: :+1:

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/WeakCounterResource.java
```diff
@@ -0,0 +1,57 @@
...
+
+
+   WeakCounterResource(ResolvePathHandler resolvePathHandler, boolean runtimeRegistration) {
+      super(WeakCounterResource.PATH, new InfinispanResourceDescriptionResolver(ModelKeys.COUNTERS), resolvePathHandler,
+            new CounterAddHandler(), new CounterRemoveHandler(), runtimeRegistration);
+   }
+
+   @Override
+   public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
+      super.registerAttributes(resourceRegistration);
```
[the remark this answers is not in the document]
pruivo: :+1:

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerCountersResource.java
```diff
@@ -0,0 +1,65 @@
+package org.jboss.as.clustering.infinispan.subsystem;
+
+import org.infinispan.counter.configuration.CounterManagerConfigurationBuilder;
+import org.infinispan.counter.configuration.Reliability;
+import org.jboss.as.controller.AttributeDefinition;
```
pruivo: unused import

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
ryanemerson: Here you have specified the `readHandler` as null, is that correct?

My understanding from what @pruivo says is that the values are read only? If so, we can use `registerReadOnlyAttribute`

vblagoje: I tried it @pruivo and for some reason I get 
```
[domain@localhost:9990 counters=COUNTERS] cd strong-counter=c4
[domain@localhost:9990 strong-counter=c4] ls
WFLYCTL0063: Composite operation was rolled back
[domain@localhost:9990 strong-counter=c4] 
```

ryanemerson: @vblagoje What was the output on the server?

vblagoje: @ryanemerson , nothing, that's was weird

ryanemerson:
@vblagoje Ok looks good :+1:, can you squash the commits and re-push. Then we'll integrate after CI tests

pruivo:
```
[domain@localhost:9990 counters=COUNTERS] ./strong-counter=my-counter:add(initial-value=0,lower-bound=0,upper-bound=10)
{
    "outcome" => "failed",
    "failure-description" => {"domain-failure-description" => "WFLYCTL0155: name may not be null"},
    "rolled-back" => true
}
```
^ isn't the name "my-counter"?  why is it asking for it?

-----------------------------------------------------

```
[domain@localhost:9990 counters=COUNTERS] ./strong-counter=my-counter:add(initial-value=0,lower-bound=0,upper-bound=10,name=weird-counter)
{
    "outcome" => "failed",
    "failure-description" => {"domain-failure-description" => "WFLYCTL0172: storage is required"},
    "rolled-back" => true
}
```
^ only the counter name is mandatory. the remaining attributes are optional...

-------------------------------------------------

```
[domain@localhost:9990 strong-counter=my-counter] ls
initial-value=0     lower-bound=0       name=weird-counter  storage=PERSISTENT  upper-bound=10  
```
^ what is my counter name? my-counter or weird-counter?

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterAddHandler.java
```diff
@@ -0,0 +1,164 @@
...
+
+        String counterName = getCounterName(operation);
+        String containerName = getContainerName(operation);
+        String counterType = getCounterType(operation);
+
+        Builder b = getBuilder(containerModel, counterType);
+        processModelNode(context, containerModel, b);
+        String name = CounterResource.COUNTER_NAME.resolveModelAttribute(context, containerModel).asString();
+        if (!counterName.equals(name)) {
+            throw new OperationFailedException("Counter node name and node's name attribute should be the same");
```
vblagoje: @pruivo @remerson I added the check to make sure the name node and the attribute name are equal. No more inconsistent counters.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
pruivo: this is a no-go for me. the model is updated with the new values but the configuratio stays the same.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterService.java
pruivo: this is the cause of the persistent problem. you are deleting the counter from the persistent storage!

vblagoje: Ok @pruivo , so we remove only non-persistent? What is the right thing to do once stop is invoked?

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
vblagoje: @pruivo what exactly are you doing to update the model? I am doing this:

```
[domain@localhost:9990 counters=COUNTERS] pwd
/profile=clustered/subsystem=datagrid-infinispan/cache-container=clustered/counters=COUNTERS
[domain@localhost:9990 counters=COUNTERS] ./strong-counter=c89:add(
initial-value=  lower-bound=  name=  storage=  upper-bound=  
[domain@localhost:9990 counters=COUNTERS] ./strong-counter=c89:add(name=c89, initial-value=0, lower-bound=0, storage=VOLATILE,upper-bound=20)
{
    "outcome" => "success",
    "result" => undefined,
    "server-groups" => {"cluster" => {"host" => {"master" => {
        "server-one" => {"response" => {"outcome" => "success"}},
        "server-two" => {"response" => {"outcome" => "success"}}
    }}}}
}
[domain@localhost:9990 counters=COUNTERS] cd strong-counter=c
c1  c89  
[domain@localhost:9990 counters=COUNTERS] cd strong-counter=c89
/  :  
[domain@localhost:9990 counters=COUNTERS] cd strong-counter=c89
[domain@localhost:9990 strong-counter=c89] :write-attribute(name=upper-bound, value=30)
{
    "outcome" => "success",
    "result" => undefined,
    "server-groups" => {"cluster" => {"host" => {"master" => {
        "server-one" => {"response" => {
            "outcome" => "success",
            "response-headers" => {
                "operation-requires-reload" => true,
                "process-state" => "reload-required"
            }
        }},
        "server-two" => {"response" => {
            "outcome" => "success",
            "response-headers" => {
                "operation-requires-reload" => true,
                "process-state" => "reload-required"
            }
        }}
    }}}}
}
[domain@localhost:9990 strong-counter=c89] ls
initial-value=0   lower-bound=0     name=c89          storage=VOLATILE  upper-bound=30    

[domain@localhost:9990 strong-counter=c89]
```
It all works as expected, once server is rebooted, the new value is loaded into configuration.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterService.java
vblagoje: Ok, check this one @pruivo . You are right. Persistent counter,if not removed, preserves its value. So we simply do not remove counters from cache manager on stop?

vblagoje:
@pruivo @ryanemerson PERSISTENT counters are preserved through reboots now. Please have a look and let integrate this one if there are no more objections.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterService.java
pruivo: IMO, do nothing! if the counter is persistent, its value is stored in the file. if it is volatile, the reboot will clear its value from the memory.

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CounterResource.java
pruivo: the volatile counter is destroyed in the reboot. so you are creating a new counter. a persistent counter isn't and the configuration isn't updated...

pruivo: and there is a small issue. it destroys *all* the volatile counters!

vblagoje: But how can I circumvent this with current Counter API, not sure what I can do to handle this properly? If I destroy counter (with remove) then I can not save its value. If I don't destroy it (remove it) then I can not update its configuration on reboot...it is chicken-egg issue, right? Suggestions?

pruivo: that's why they should be read-only for now. when the `undefine()` is implemented, we can support to write the attributed!

vblagoje: Ok @pruivo will investigate that impl now. Thanks

vblagoje:
@pruivo @ryanemerson all attributes are read only. In CounterService#stop we do not do anything. Should be GTG now.

pruivo:
@vblagoje looks good to me.
@ryanemerson you can push the button if it is ok for you :+1:

ryanemerson:
I've initiated a build, lets wait for the CI results

vblagoje:
@remerson can you please do this as I have no idea what you are talking
about

ryanemerson:
@vblagoje Sure np, I've created [ISPN-8575](https://issues.jboss.org/browse/ISPN-8575). FYI wildfly-modules allows users to deploy the latest infinispan and server subsystems to a wildfly/eap instance

ryanemerson:
@vblagoje there are a lot of integration test failures that need fixing before we can merge

vblagoje:
@ryanemerson but they do not seem to be related to this commit? If yes, which tests are you referring to in CI run?

ryanemerson:
@vblagoje The test associated with wildfly-modules unfortunately:

```
org.infinispan.test.integration.as.InfinispanCoreIT.org.infinispan.test.integration.as.InfinispanCoreIT
org.infinispan.test.integration.as.InfinispanExtensionIT.org.infinispan.test.integration.as.InfinispanExtensionIT
org.infinispan.test.integration.as.InfinispanStoreJdbcIT.org.infinispan.test.integration.as.InfinispanStoreJdbcIT
org.infinispan.test.integration.as.InfinispanStoreJpaIT.org.infinispan.test.integration.as.InfinispanStoreJpaIT
org.infinispan.test.integration.as.InfinispanStoreRocksDBIT.org.infinispan.test.integration.as.InfinispanStoreRocksDBIT
org.infinispan.test.integration.as.InfinispanTreeIT.org.infinispan.test.integration.as.InfinispanTreeIT
org.infinispan.test.integration.as.cdi.DuplicatedDomainsCdiIT.org.infinispan.test.integration.as.cdi.DuplicatedDomainsCdiIT
org.infinispan.test.integration.as.cdi.GreetingCacheManagerIT.org.infinispan.test.integration.as.cdi.GreetingCacheManagerIT
org.infinispan.test.integration.as.cdi.GreetingServiceIT.org.infinispan.test.integration.as.cdi.GreetingServiceIT
org.infinispan.test.integration.as.client.HotRodClientIT.org.infinispan.test.integration.as.client.HotRodClientIT
org.infinispan.test.integration.as.client.HotRodQueryIT.org.infinispan.test.integration.as.client.HotRodQueryIT
org.infinispan.test.integration.as.query.DSLQueryIT.org.infinispan.test.integration.as.query.DSLQueryIT
org.infinispan.test.integration.as.query.ElasticsearchIndexManagerIT.org.infinispan.test.integration.as.query.ElasticsearchIndexManagerIT
org.infinispan.test.integration.as.query.InfinispanQueryIT.org.infinispan.test.integration.as.query.InfinispanQueryIT
org.infinispan.test.integration.as.remote.InfinispanRemoteWithQueryIT.org.infinispan.test.integration.as.remote.InfinispanRemoteWithQueryIT
```

As we don't want to break CI for everything else, I think you need to add counters to wildfly-modules in this PR. The steps are as follows:

1. Add counters dependency to `wildfly-modules/pom.xml`
2. Add entry in `wildfly-modules/build.xml`
```
<module-def name="org.infinispan.counter" slot="${infinispan.slot}">
            <maven-resource group="org.infinispan" artifact="infinispan-clustered-counter" />
        </module-def>
```
3. Create `wildfly-modules/src/main/org/infinispan/counter/main/module.xml with the following content (ignore the comment):

```
<?xml version="1.0" encoding="UTF-8"?>
<module xmlns="urn:jboss:module:1.3" name="org.infinispan.client.hotrod" slot="${slot}">
    <resources>
        <!-- Insert resources here -->
    </resources>

    <dependencies>
        <module name="javax.api"/>
        <module name="javax.transaction.api"/>
        <module name="org.infinispan"  slot="${slot}"/>
        <module name="org.jboss.logging"/>
    </dependencies>
</module>
```

From memory that should be all that is required

vblagoje:
Ok @ryanemerson , on it!

vblagoje:
Ok @ryanemerson updated it all compiles on my local machine. Give it a quick look at https://github.com/infinispan/infinispan/pull/5570/commits/25c8b201b827757c9b113c542d809f32371e3ec9 and lets wait for CI

wildfly-modules/src/main/resources/org/infinispan/counter/main/module.xml
```diff
@@ -0,0 +1,14 @@
...
+        <artifact name="${org.infinispan:infinispan-clustered-counter}"/>
+    </resources>
+
+    <dependencies>
+        <module name="javax.api"/>
+        <module name="javax.transaction.api"/>
+        <module name="org.infinispan"/>
+        <module name="org.jboss.logging"/>
+    </dependencies>
+</module>
```
ryanemerson: This content is wrong! It should be like I stated on the PR:

```
<?xml version="1.0" encoding="UTF-8"?>
<module xmlns="urn:jboss:module:1.3" name="org.infinispan.counter" slot="${slot}">
    <resources>
        <!-- Insert resources here -->
    </resources>

    <dependencies>
        <module name="javax.api"/>
        <module name="javax.transaction.api"/>
        <module name="org.infinispan"  slot="${slot}"/>
        <module name="org.jboss.logging"/>
    </dependencies>
</module>
```

vblagoje:
Ok @ryanemerson did not understand you. Should be ok now

ryanemerson:
@vblagoje Starting a new CI build, I'll merge if there are no related failures

vblagoje:
Thanks @remerson appreciate it man!

server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
```diff
@@ -122,6 +129,32 @@
          </cache-container>
       </replacement>
    </supplement>
+   <supplement name="counters">
```
ryanemerson: This supplement doesn't make sense and is what is causing the wildfly-module tests to fail. No transport is defined, but counters are clustered which results in [this error](https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/factories/RpcManagerFactory.java#L24) being thrown on server startup. AFAICT, this supplement isn't used anywhere either

server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
```diff
@@ -77,6 +77,13 @@
             <distributed-cache name="default" />
 
             <replicated-cache name="repl" configuration="replicated"/>
+
+            <counters>
+                <strong-counter name="c1" initial-value="0" storage="VOLATILE">
+                    <lower-bound value="0" />
+                </strong-counter>
+                <weak-counter name="c2" initial-value="0" storage="VOLATILE" concurrency-level="32" />
+            </counters>
```
ryanemerson: I'm pretty certain that we don't want to have two counters on the server's default clustered configuration. @tristantarrant any thoughts?

server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
vblagoje: Ok, I'll remove it @ryanemerson

server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
vblagoje: Why not @ryanemerson ? It is ok to have one strong and one weak counter IMHO.

ryanemerson: @vblagoje Because you are creating an additional distributed cache to store the counter values. This doesn't come for free in terms of resources locally or RPCs across the cluster.

vblagoje: @ryanemerson got it, @tristantarrant wdys?

server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
ryanemerson: Actually, the server throws ^ error whenever you run it standalone in `server/integration/build/target/...` so this isn't just a problem with `wildfly-modules`

server/integration/infinispan/src/main/resources/subsystem-templates/infinispan-core.xml
pruivo: actually, you are creating 2 additional caches - a replicated and a distributed - and they are created when the CacheManager is started.

tristantarrant: No predefined counters in the default configuration please. We should just have an empty counters container and that's it.
We could have some examples of counters in the example directory, but then we might also need a test to ensure that the example works as expected.

tristantarrant: Additionally, the three configurations (standalone.xml, clustered.xml and domain.xml) should be more or less "equivalent" (with the obvious limitation that standalone.xml cannot have clustered features).

ryanemerson:
@vblagoje The remaining test failures in `integrationtests/wildfly-modules` are because we have not declared a dependency on counters in the `org.infinispan.extension` module. I obviously missed step 4 in my instructions :see_no_evil:.

You need to add the following line to `wildfly-modules/src/main/resources/org/infinispan/extension/main/module.xml`:

`<module name="org.infinispan.counter" services="import" slot="${slot}"/>`

vblagoje:
Ok @ryanemerson I amended the PR. Let's see what CI says now!

ryanemerson:
@vblagoje the fun continues ... rebase required

vblagoje:
Should be gtg now @ryanemerson lets see CI, one more time?

ryanemerson:
Integrated, thanks @vblagoje

PR #5574, opened 2017-11-07
ISPN-8504 View change during cluster listener registration can throw

wburns:
SuspectException

* Only register listeners with addresses that have this cache
* Cache could be stopped concurrently, so ignore SuspectException

https://issues.jboss.org/browse/ISPN-8504


pruivo:
integrated! thanks @wburns!
Can you create a PR for 9.1 too?

pruivo:
I've created the PR myself https://github.com/infinispan/infinispan/pull/5575

PR #5575, opened 2017-11-08
[9.1.x] ISPN-8504 View change during cluster listener registration can throw …

pruivo:
…SuspectException

* Only register listeners with addresses that have this cache
* Cache could be stopped concurrently, so ignore SuspectException

https://issues.jboss.org/browse/ISPN-8504

backport from master. I'm just waiting for CI :)


wburns:
Integrated into 9.1.x, thanks @pruivo

PR #5576, opened 2017-11-08
ISPN-8385 Document how DataContainer can store entries in different

wburns:
format than provided to Cache

https://issues.jboss.org/browse/ISPN-8385


gustavonalle:
Integrated, thanks @wburns !

PR #5569, opened 2017-11-06
Backporting various off heap issues to 9.1.x

wburns:
https://issues.jboss.org/browse/ISPN-8409
https://issues.jboss.org/browse/ISPN-8419
https://issues.jboss.org/browse/ISPN-8481
https://issues.jboss.org/browse/ISPN-8423
https://issues.jboss.org/browse/ISPN-8454

PR #5577, opened 2017-11-08
ISPN-8421 Eviction loop creates entry from memory and remove call does

wburns:
same thing

* Don't create ice if not needed for various removes

https://issues.jboss.org/browse/ISPN-8421

### DIFF
--- a/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
@@ -260,7 +260,7 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
                InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(addressToRemove);
                passivator.passivate(ice);
                // TODO: this rereads the object from memory again!
-               performRemove(memoryLookup.getMemoryAddress(ice.getKey()), ice.getKey());
+               performRemove(memoryLookup.getMemoryAddress(ice.getKey()), ice.getKey(), false);
                evictionManager.onEntryEviction(Collections.singletonMap(ice.getKey(), ice));
             } finally {
                entryWriteLock.unlock();
--- a/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
@@ -259,8 +259,7 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
             try {
                InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(addressToRemove);
                passivator.passivate(ice);
-               // TODO: this rereads the object from memory again!
-               performRemove(memoryLookup.getMemoryAddress(ice.getKey()), ice.getKey(), false);
+               performRemove(memoryLookup.getMemoryAddress(ice.getKey()), addressToRemove, ice.getKey(), false);
                evictionManager.onEntryEviction(Collections.singletonMap(ice.getKey(), ice));
             } finally {
                entryWriteLock.unlock();
--- a/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
@@ -316,7 +316,7 @@ public class OffHeapDataContainer implements DataContainer<WrappedBytes, Wrapped
          if (address == 0) {
             return null;
          }
-         return performRemove(address, key);
+         return performRemove(address, key, true);
       } finally {
          lock.unlock();
       }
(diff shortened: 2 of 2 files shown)


wburns:
Initial results were promising when eviction was required when doing puts or compute > 20% throughput improvement.

https://issues.jboss.org/browse/ISPN-8421?focusedCommentId=13485790&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13485790

wburns:
This should cherry pick as is to 9.1.x as well.

wburns:
This should cherrypick into 9.1.x after https://github.com/infinispan/infinispan/pull/5569 is integrated.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -260,7 +260,7 @@ private void ensureSize() {
                InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(addressToRemove);
                passivator.passivate(ice);
                // TODO: this rereads the object from memory again!
```
anistor: @wburns Is this TODO still relevant?

anistor:
Looks good. The 2 CI failures are unrelated.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
wburns: Oops thought I removed that, let me remove just that line real quick

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -327,30 +327,40 @@ protected void entryRetrieved(long entryAddress) {
...
       long address = bucketHeadAddress;
+      InternalCacheEntry<WrappedBytes, WrappedBytes> ice = null;
 
       while (address != 0) {
          long nextAddress = offHeapEntryFactory.getNext(address);
-         InternalCacheEntry<WrappedBytes, WrappedBytes> ice = offHeapEntryFactory.fromMemory(address);
-         if (ice.getKey().equals(wba)) {
+         boolean removeThisAddress;
+         if (requireReturn) {
+            ice = offHeapEntryFactory.fromMemory(address);
```
danberindei: This could be done in if `removeThisAddress == true`.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -327,30 +327,40 @@ protected void entryRetrieved(long entryAddress) {
...
+         if (removeThisAddress) {
             entryRemoved(address);
             if (prevAddress != 0) {
                offHeapEntryFactory.setNext(prevAddress, nextAddress);
             } else {
                memoryLookup.putMemoryAddress(key, nextAddress);
             }
             size.decrementAndGet();
-            return ice;
+            break;
```
danberindei: Doesn't this mean if none of the entries in the bucket match the key, the method returns the last entry in the bucket?

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
wburns: This one is a tough one imo. Do we want to optimize the miss case or the replace case?  My thought was that normally there would be just 1 address in the lookup, in which case the replace case would be hit more often.

The inverse is that if we optimize for replace (as is right now) it is more costly when a miss occurs though.

wburns: I guess I am open to optimizing for the miss case, assuming that people have their address count not high enough.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
wburns: When `requireReturn` is true, yeah. Good catch. I can fix this by adding an else and assigning to null or if we change the other comment it would be fine.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
danberindei: So `offHeapEntryFactory.equalsKey(address, wba)` is slower than `ice.getKey().equals(wba)`?

danberindei: Actually in ensureSize()` you already have the `addressToRemove` pointer, so you could compare addresses instead of keys in that case.

wburns: No, my thought at the time (which as I think about it now was probably wrong). Would be that calling:

> offHeapEntryFactory.equalsKey(address, wba)
> ice = offHeapEntryFactory.fromMemory(address);
> return ice;

would be slower than 

> ice = offHeapEntryFactory.fromMemory(address);
> ice.getKey().equals(wba)
> return ice;

My mistaking thought at the time would be that CPU would be able to optimize the latter one, but thinking more the former actually has a better chance to be more optimized as the key could be smaller and fit in the cache between subsequent calls.

So I will change it as you suggested, Thanks! This will also fix the other comment.

wburns: >Actually in ensureSize()you already have theaddressToRemove` pointer, so you could compare addresses instead of keys in that case.

Hrmm that is a good point. You mean to add an additional argument to `performRemove` that takes the `bucketHeadAddress` and the `addressToRemove` and then we don't even have to do the equals check at all. That is even way better :+1:

danberindei: Ok, I thought you saw something in the profiler that suggests HotSpot can't optimize `equalsKey`.

wburns: Actually this can apply to all the write operations (evict, remove or put/compute that replace an entry 👍

wburns:
Adding additional changes for this that should bring additional performance to many use cases, but I have to test and verify with JMH so this PR will have to end up missing this build.

wburns:
Okay I added in the address to remove for eviction. I have created https://issues.jboss.org/browse/ISPN-8517 that should handle the other operations that I want to do.

wburns:
If this review is alright I still need to fixup the last commit though.

wburns:
Just running a quick JMH test with this branch compared to master

danberindei:
Looks good, running the core tests on my machine as well.

wburns:
JMH tests are better, but not as good as they once were. My guess is the other changes have made an impact as well, but still it is an improvement:

https://gist.github.com/wburns/ca2b35ded37e3dd685ac3c5845a9c146

PR #5579, opened 2017-11-09
ISPN-8235 Test behaviour of projection and aggregations with both Sto…

anistor:
…re.YES/NO with the sample 'bank.proto' schema

just a test.

https://issues.jboss.org/browse/ISPN-8235


gustavonalle:
Integrated, thanks @anistor

PR #5584, opened 2017-11-10
ISPN-8515 Allow @Inject on fields

rvansa:
https://issues.jboss.org/browse/ISPN-8515

Notice that I've removed > 1000 lines of code!

### DIFF
--- a/cli/cli-interpreter/src/main/java/org/infinispan/cli/interpreter/Interpreter.java
+++ b/cli/cli-interpreter/src/main/java/org/infinispan/cli/interpreter/Interpreter.java
@@ -45,23 +45,18 @@ public class Interpreter {
    private static final long DEFAULT_SESSION_REAPER_WAKEUP_INTERVAL = 60000l; // in millis
    private static final long DEFAULT_SESSION_TIMEOUT = 360000l; // in millis
 
-   private EmbeddedCacheManager cacheManager;
+   @Inject private EmbeddedCacheManager cacheManager;
+   @Inject private TimeService timeService;
+
    private ScheduledExecutorService executor;
    private long sessionReaperWakeupInterval = DEFAULT_SESSION_REAPER_WAKEUP_INTERVAL;
    private long sessionTimeout = DEFAULT_SESSION_TIMEOUT;
    private CodecRegistry codecRegistry;
-   private TimeService timeService;
 
    private final Map<String, Session> sessions = new ConcurrentHashMap<>();
    private ScheduledFuture<?> sessionReaperTask;
 
    public Interpreter() {
-   }
-
-   @Inject
-   public void initialize(final EmbeddedCacheManager cacheManager, TimeService timeService) {
-      this.cacheManager = cacheManager;
-      this.timeService = timeService;
    }
 
    @Start
--- a/commons/src/main/java/org/infinispan/commons/util/ReflectionUtil.java
+++ b/commons/src/main/java/org/infinispan/commons/util/ReflectionUtil.java
@@ -30,7 +30,6 @@ public class ReflectionUtil {
                                                       float[].class, double[].class, boolean[].class, char[].class};
    public static final Class<?>[] EMPTY_CLASS_ARRAY = new Class[0];
 
-
    /**
     * Returns a set of Methods that contain the given method annotation.  This includes all public, protected, package
     * and private methods, as well as those of superclasses.  Note that this does *not* include overridden methods.
@@ -60,6 +59,12 @@ public class ReflectionUtil {
             annotated.add(m);
       }
 
+      return annotated;
+   }
+
+   public static List<Field> getAllFields(Class<?> c, Class<? extends Annotation> annotationType) {
+      List<Field> annotated = new LinkedList<>();
+      inspectFieldsRecursively(c, annotated, annotationType);
       return annotated;
    }
 
@@ -128,6 +133,21 @@ public class ReflectionUtil {
       }
    }
 
+   private static void inspectFieldsRecursively(Class<?> c, List<Field> s, Class<? extends Annotation> annotationType) {
+      if (c == null || c.isInterface()) {
+         return;
+      }
+      for (Field f : c.getDeclaredFields()) {
+         if (f.isAnnotationPresent(annotationType)) {
+            s.add(f);
+         }
+      }
+
+      if (!c.equals(Object.class)) {
+         inspectFieldsRecursively(c.getSuperclass(), s, annotationType);
+      }
+   }
+
    /**
     * Tests whether a method has already been found, i.e., overridden.
     *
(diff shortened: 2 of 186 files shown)


galderz:
Build failed, checkstyle

galderz:
NICE :)

commons/src/main/java/org/infinispan/commons/util/ReflectionUtil.java
```diff
@@ -128,6 +133,21 @@ private static void inspectRecursively(Class<?> c, List<Method> s, Class<? exten
...
+      if (c == null || c.isInterface()) {
+         return;
+      }
+      for (Field f : c.getDeclaredFields()) {
+         if (f.isAnnotationPresent(annotationType)) {
+            s.add(f);
+         }
+      }
+
+      if (!c.equals(Object.class)) {
```
anistor: It's safe to use ```==``` here.

anistor:
@rvansa Looks nice. One thing I'll miss about the method injection is the ability to set a breakpoint and inspect the contents of the component registry at the injection time. Our dependency injection framework is far from perfect so I used this simple trick in order to debug various things. Can we still achieve something similar in an easy way?

rvansa:
@anistor I thought you could set a field breakpoint, but this does not seem to work :-/ Looks like you can either set a conditional breakpoint in `ReflectionUtil.setAccessibly` with `field.getName().equals("foo") && field.getDeclaringClass() == SomeClass.class` or temporarily add a method. I haven't disabled the `@Inject` on methods, and I haven't rewritten some methods that do non-trivial processing there (I've noticed that query-related modules often use @Inject methods rather than @Start, while I think that @Inject was intended purely for the assignment of references).

commons/src/main/java/org/infinispan/commons/util/ReflectionUtil.java
rvansa: True, should I change it in `inspectRecursively` above, too?

anistor:
Yes, the query modules, and maybe others misused the inject annotation. I vaguely remember somebody just fixed that, not me.

core/src/test/java/org/infinispan/test/TestingUtil.java
```diff
@@ -1772,4 +1774,16 @@ public static boolean isTriangleAlgorithm(CacheMode cacheMode, boolean transacti
    public static <K,V> Map.Entry<K,V> createMapEntry(K key, V value) {
       return new AbstractMap.SimpleEntry<>(key, value);
    }
+
+   public static void inject(Object instance, Object... components) {
```
wburns: Maybe just add a comment that this only works on fields with `@Inject` and doesn't invoke any methods with `@Inject`.

This is also much cleaner with fields since we can just override specific components :+1:

danberindei:
Looks nice! The only thing I'm worried about is memory usage: the component registry keeps the parsed component metadata in memory at all times, and I it's pretty big even with a single injection method per component. We need to make sure that we don't keep 100MB of component metadata in memory for each cache manager.

anistor:
The component metadata is my all time unfavourite piece of Infinispan. Some years ago I started a quick hack to make the comp metadata a generated class (using javassist), full of statics, rather than a serialized object. That approach also had the nice side effect of not wasting memory with useless duplication.

That branch was lost during a git repo corruption caused by a hdd crash. I think I must re-do that.

anistor:
Yes, it is out of scope of this PR. 
I was using javassist at compile time, not classload time, so there were no runtime dependencies.

rvansa:
@anistor Another thing I find very annoying is that when you update something related to metadata you have to compile from cmdline, as IDE (at least by default) doesn't generate the metadata. I would rather avoid keeping metadata completely if scanning was sub-second.

rvansa:
@danberindei I've made a simple test where I instantiate 2 or 10 distributed caches (on single node, though) and taken heap dump from those. It seems that this adds roughly 100k of objects, regardless of number of caches (the histogram diff showed similar numbers for both cases).

anistor:
@rvansa The compile-time generation of the metadata can be solved with a  javax.annotation.processing-style processor that is invoked by javac rather than the external custom-made tool that we use now. Before java 8 annotation processor were poorly supported by IDEs but now this should not be a problem.

danberindei:
@rvansa yeah, the metadata is only loaded once per cache manager. I tried the same experiment locally and the total retained size of `ComponentMetadataRepo` is 258k for me (vs 236k without your changes), much less than I was expecting.

I had this image of a heap dump with component metadata using 100s of MBs of heap, but I see now those weren't just the metadata of the currently running tests: some tests keep references to cache managers after they're done, and TestNG only discards the test instances at the very end.

One thing we could improve without changing the format is to call `intern()` on the `factories` and `componentMetadataMap` keys and values, but even that is probably too far from the scope of this PR.

core/src/test/java/org/infinispan/test/TestingUtil.java
danberindei: Injecting multiple named executors wouldn't work that well, maybe throw an exception when 2 or more components match the same field? The test could create a proper `ComponentRegistry` and call `wireDependencies` in those cases.

cli/cli-interpreter/src/main/java/org/infinispan/cli/interpreter/Interpreter.java
```diff
@@ -45,25 +45,20 @@
    private static final long DEFAULT_SESSION_REAPER_WAKEUP_INTERVAL = 60000l; // in millis
    private static final long DEFAULT_SESSION_TIMEOUT = 360000l; // in millis
 
-   private EmbeddedCacheManager cacheManager;
+   @Inject private EmbeddedCacheManager cacheManager;
+   @Inject private TimeService timeService;
```
danberindei: Don't know how I missed this before, but do we really want to put the annotation on the same line as the field?

rvansa: I like the declaration to be concise, since `@Inject` does not have any parameters. 8 characters including space don't justify newline.

I actually put the annotation on separate line when component name is set through another annotation, but not in the default case.

core/src/test/java/org/infinispan/test/TestingUtil.java
rvansa: @danberindei It's not bullet proof, I did not bother since this is only a testing helper. When you need it, you'll code that, and test that such util test code works.

rvansa: I've added the check for multiple matches.

galderz:
Needs rebase

rvansa:
@galderz @anistor @danberindei Rebased.

rvansa:
I've been also forced to provide a way to set named components using `TestingUtil.inject`: see `TestingUtil.named`

galderz:
1,531 failures...? Kicking off another CI build

rvansa:
@galderz CI was right, fixed & rerunning...

rvansa:
@galderz OK, now we're down to 7 failures (the usual suspects).

PR #5586, opened 2017-11-13
Ispn 7599/jolokia server part 2

slaskawi:
https://issues.jboss.org/browse/ISPN-7599

This is a second approach to this ticket. Previously this has been merged by @vblagoje  (https://github.com/infinispan/infinispan/pull/5267) and then reverted by @tristantarrant.

So at the moment this PR is more about how we want to implement this (see [this email](http://lists.jboss.org/pipermail/infinispan-dev/2017-November/017938.html) thread). Basically we have two approaches:

1. Put all the bits into main server distribution. This effectively means that this PR is valid.
2. Put all the bits into integration layers. This effectively means close this PR.

@tristantarrant @gustavonalle @ryanemerson @vblagoje - What's your call on this?

### DIFF
--- a/documentation/src/main/asciidoc/server_guide/management.adoc
+++ b/documentation/src/main/asciidoc/server_guide/management.adoc
@@ -65,3 +65,10 @@ In domain mode the port will be the port of the Remoting connector for the Serve
 </subsystem>
 
 ----
+
+=== Jolokia
+
+You can also expose JMX beans using link:https://jolokia.org/[Jolokia]. In order to do this, just run the server with
+additional parameter `--jmx`, for example: `./standalone.xml -c cloud.xml --jmx`. Jolokia configuration is stored in
+`agent.conf` file. It is possible to override this file by specifying it after `--jmx` parameter. For example:
+`./standalone.sh -c cloud.xml --jmx my-config.properties`.
--- a/server/integration/feature-pack/src/main/resources/content/bin/agent.conf
+++ b/server/integration/feature-pack/src/main/resources/content/bin/agent.conf
@@ -1,0 +1,1 @@
+jolokia=port=8778
--- a/server/integration/feature-pack/src/main/resources/content/bin/standalone.bat
+++ b/server/integration/feature-pack/src/main/resources/content/bin/standalone.bat
@@ -13,6 +13,7 @@ rem         standalone.bat --debug 9797
 rem By default debug mode is disable.
 set DEBUG_MODE=false
 set DEBUG_PORT=8787
+set USE_JMX_COLLECTORS=fals
 rem Set to all parameters by default
 set "SERVER_OPTS=%*"
 
@@ -43,6 +44,8 @@ if "%~1" == "" (
    goto MAIN
 ) else if "%~1" == "--debug" (
    goto READ-DEBUG-PORT
+) else if "%~1" == "--jmx" (
+   goto READ-JMX-CONFIG
 ) else if "%~1" == "-secmgr" (
    set SECMGR=true
 )
@@ -56,6 +59,18 @@ if not %DEBUG_ARG% == "" (
    if x%DEBUG_ARG:-=%==x%DEBUG_ARG% (
       shift
       set DEBUG_PORT=%DEBUG_ARG%
+   )
+   shift
+   goto READ-ARGS
+)
+
+:READ-JMX-CONFIG
+set "USE_JMX_CONNECTORS=true"
+set JMX_ARG="%2"
+if not %JMX_ARG% == "" (
+   if x%JMX_ARG:-=%==x%JMX_ARG% (
+      shift
+      set JMX_CONF=%JMX_ARG%
    )
    shift
    goto READ-ARGS
@@ -96,7 +111,6 @@ if exist "%STANDALONE_CONF%" (
    echo Config file not found "%STANDALONE_CONF%"
 )
 
-
 rem Set debug settings if not already set
 if "%DEBUG_MODE%" == "true" (
    echo "%JAVA_OPTS%" | findstr /I "\-agentlib:jdwp" > nul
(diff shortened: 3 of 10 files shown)


gustavonalle:
+1 for adding it to the server so everyone can benefit from it!

Wondering why it was reverted? From my understanding, it was because it was too late for 9.1...

slaskawi:
Yes, I think that was the reason... @tristantarrant Maybe you remember?

gustavonalle:
re-triggered CI

gustavonalle:
Can we have a test for this?

documentation/src/main/asciidoc/server_guide/management.adoc
```diff
@@ -65,3 +65,10 @@ In domain mode the port will be the port of the Remoting connector for the Serve
 </subsystem>
 
 ----
+
+=== Jolokia
+
+You can also expose JMX beans using link:https://jolokia.org/[Jolokia]. In order to do this, just run the server with
+additional parameter `--jmx`, for example: `./standalone.xml -c cloud.xml --jmx`. Jolokia configuration is stored in
+`agent.conf` file. It is possible to override this file by specifying it after `--jmx` parameter. For example:
+`./standalone.sh -c cloud.xml --jmx my-config.properties`.
```
gustavonalle: Please add a quick way to check if jolokia is up and running, via ```curl http://127.0.0.1:8778/jolokia/version```

server/integration/feature-pack/src/main/resources/content/bin/agent.conf
```diff
@@ -0,0 +1 @@
+jolokia=port=8778
```
gustavonalle: With this conf, jolokia will bind only to ```127.0.0.1``` which is a good default security-wise. Is this enough for Openshift to operate? 

Also, Wildfly has cmd line flag ```-b``` flag to control the binding of application level services, and ```-bmanagement``` to control binding of the admin console. Wondering if it's worth it to provide a switch to control the jolokia binding or piggyback in the ```-bmanagement``` one, as an easier way of getting it running quickly without providing a custom config via --jmx.

gustavonalle:
LGTM, I've added a few comments meant as optional improvements

documentation/src/main/asciidoc/server_guide/management.adoc
slaskawi: Jolokia spits a startup message to the logs... Do we need to mention additional check here?

server/integration/feature-pack/src/main/resources/content/bin/agent.conf
slaskawi: This file needs to be overridden for OpenShift :)

OpenShift needs different settings and certificates. Here's an example: https://github.com/infinispan/infinispan-openshift-templates/pull/6/files#diff-edf8fc0cedef8b07d126d61e0799548aR74

slaskawi:
@gustavonalle Unfortunately there is no test for it at the moment. I can add it if you want...

documentation/src/main/asciidoc/server_guide/management.adoc
gustavonalle: It won't hurt to add a simple check :)

gustavonalle:
@slaskawi yes, I'd like a test, something simple as checking the output of http://127.0.0.1:8778/jolokia/version is enough IMO.

tristantarrant:
Both standalone.sh and jolokia.sh specify `/bin/sh` but you are using bashisms (functions):

`./bin/standalone.sh --jmx
./bin/standalone.sh: 3: ./bin/jolokia.sh: function: not found
jboss-modules.jar
./bin/standalone.sh: 6: ./bin/jolokia.sh: Syntax error: "}" unexpected`

Also, you should add some information about the --jmx switch in the standalone.sh help

gustavonalle:
@tristantarrant good points, but I didn't cache the errors you pasted when running locally, and I suppose it also works via docker. How did you come accross them?

tristantarrant:
This fails when /bin/sh is a symbolic link to something which is not bash. For example, on Debian/Ubuntu, /bin/sh is provided by dash.
The jolokia.sh script can be implemented without a function anyway.

gustavonalle:
@tristantarrant gotcha! Nice catch

tristantarrant:
@slaskawi forgot about this ?

slaskawi:
@tristantarrant I'm not sure where to put the code to display help when typing `./standalone.sh help`.

However I put some comments into the file.

server/integration/feature-pack/src/main/resources/content/bin/standalone.bat
```diff
@@ -13,6 +13,7 @@ rem         standalone.bat --debug 9797
 rem By default debug mode is disable.
 set DEBUG_MODE=false
 set DEBUG_PORT=8787
+set USE_JMX_COLLECTORS=fals
```
tristantarrant: you meant false ?

slaskawi:
@gustavonalle Unfortunately I have bad info RT tests...

Arquillian constructs a classpath with JBoss Modules and fires it up (see [this piece of the code](https://github.com/wildfly/wildfly-arquillian/blob/master/container-managed/src/main/java/org/jboss/as/arquillian/container/managed/ManagedDeployableContainer.java#L74)). The idea behind adding Agent Bond was to wire it up in the scripts and try not to touch the internals. Unfortunately this made it almost impossible to test in `integrationtests` module. 

I think the best approach would be to setup a CI job in templates repo. This sounds like a better idea than writing OS-dependent tests in Infinispan repo....

gustavonalle:
@slaskawi no prob, we can add a test later

galderz:
Anything else to add here?

gustavonalle:
Merged, thanks @slaskawi !

PR #5587, opened 2017-11-13
ISPN-8529 Implement cache admin ops over the REST protocol

tristantarrant:
https://issues.jboss.org/browse/ISPN-8529

### DIFF
--- a/server/rest/src/main/java/org/infinispan/rest/JSONConstants.java
+++ b/server/rest/src/main/java/org/infinispan/rest/JSONConstants.java
@@ -14,4 +14,7 @@ public interface JSONConstants {
    String QUERY_STRING = "query";
    String TOTAL_RESULTS = "total_results";
    String TYPE = "_type";
+   String TEMPLATE = "template";
+   String INFINITY = "infinity";
+   String INFINITY_NOROOT = "infinity,noroot";
 }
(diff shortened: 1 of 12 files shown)


gustavonalle:
My comments:

- Although it is a limited cache creation feature, since you need to have a template pre-defined, it's better than not having any capability. I can see there's already an 'indexed' cache template in the server, so playing with query over Rest would be possible without touching the server config   👍 

- Not sure about WebDav and extra verbs: it doesn't feel very restful IMO. OTOH, there's not much we can do with the current API. 
Btw, the restful way would be to stick to HTTP verbs, and make use resources and collections of resources, like:
  - Create a cache:  POST  /rest/caches
  - Delete a cache:  DELETE /rest/caches/myCache
  - Create a template: POST  /rest/templates
  - Delete a template: DELETE /rest/templates/myTemplate
  - Create an entry: POST /rest/caches/myCache/1
  - Create a counter: POST /rest/counters
  - Get a counter value: GET /rest/counters/mycounter
  - Increment a counter: GET /rest/counters/mycounter?action=increment 
  - Search a cache: GET /rest/caches/myCache?action=search


**TL;DR:**

LGTM for 9.2. Going further, we need to redesign our Rest API, I will write up something and create a JIRA

server/rest/src/main/java/org/infinispan/rest/JSONConstants.java
```diff
@@ -14,4 +14,7 @@
    String QUERY_STRING = "query";
    String TOTAL_RESULTS = "total_results";
    String TYPE = "_type";
+   String TEMPLATE = "template";
```
gustavonalle: Technically those are not JSONConstants, so it'd be better to either rename this interface or move those 3 new values to another place

tristantarrant:
Ok, it seems like it would be best to use actions. I'll fix it shortly.

gustavonalle:
I created https://issues.jboss.org/browse/ISPN-8535 with a proposal of the new API, comments are welcome!

gustavonalle:
Obviously the new API is for Infinispan 10, as it breaks the current one ;)

tristantarrant:
Closing this for now

PR #5589, opened 2017-11-14
ISPN-8519 deleted not needed

karesti:
https://issues.jboss.org/browse/ISPN-8519


pruivo:
why is the test not needed?

karesti:
because there is already one with 3 nodes where I kill one and I have another test where I do an split brain with two minority partitions. Vert.x tests run now on 2 nodes killing one but this does not make too much sense and should be running with 3 nodes.

danberindei:
Integrated, thanks Katia!

pruivo:
you have a failure:
org.infinispan.lock.ClusteredLockSplitBrainTest.testAutoReleaseIfLockIsAcquiredFromAMinorityPartition[DIST_SYNC] (from infinispan-clustered-lock)
is it tracked?

karesti:
@pruivo 😭  nooooooooooooooo

PR #5590, opened 2017-11-14
ISPN-8534 Make off heap memory based eviction a better estimation

wburns:
* Add address count to memory based eviction count
* Align sizes of off heap to 8 bytes
* Also changed allocated count to LongAdder
* Fixed issue that MemoryConfiguration wasn't being validated

https://issues.jboss.org/browse/ISPN-8534
https://issues.jboss.org/browse/ISPN-8545
https://issues.jboss.org/browse/ISPN-8550

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/cache/MemoryConfigurationBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/cache/MemoryConfigurationBuilder.java
@@ -5,11 +5,13 @@ import static org.infinispan.configuration.cache.MemoryConfiguration.EVICTION_TY
 import static org.infinispan.configuration.cache.MemoryConfiguration.SIZE;
 import static org.infinispan.configuration.cache.MemoryConfiguration.STORAGE_TYPE;
 
-import org.infinispan.commons.CacheConfigurationException;
 import org.infinispan.commons.configuration.Builder;
 import org.infinispan.commons.configuration.attributes.AttributeSet;
+import org.infinispan.commons.logging.LogFactory;
 import org.infinispan.configuration.global.GlobalConfiguration;
+import org.infinispan.container.offheap.OffHeapDataContainer;
 import org.infinispan.eviction.EvictionType;
+import org.infinispan.util.logging.Log;
 
 /**
  * Controls the data container for the cache.
@@ -17,6 +19,8 @@ import org.infinispan.eviction.EvictionType;
  * @author William Burns
  */
 public class MemoryConfigurationBuilder extends AbstractConfigurationChildBuilder implements Builder<MemoryConfiguration> {
+
+   private static final Log log = LogFactory.getLog(MemoryConfigurationBuilder.class, Log.class);
 
    private AttributeSet attributes;
 
@@ -81,13 +85,24 @@ public class MemoryConfigurationBuilder extends AbstractConfigurationChildBuilde
    public void validate() {
       StorageType type = attributes.attribute(STORAGE_TYPE).get();
       if (type != StorageType.OBJECT && getBuilder().compatibility().isEnabled()) {
-         throw new CacheConfigurationException("Compatibility mode requires OBJECT storage type but was: " + type);
+         throw log.compatibilityModeOnlyCompatibleWithObjectStorage(type);
       }
       long size = attributes.attribute(SIZE).get();
       if (size > 0) {
          EvictionType evictionType = attributes.attribute(EVICTION_TYPE).get();
-         if (evictionType == EvictionType.MEMORY && type != StorageType.BINARY) {
-            throw new CacheConfigurationException("MEMORY based eviction requires binary but was : " + type);
+         if (evictionType == EvictionType.MEMORY) {
+            switch (type) {
+               case OBJECT:
+                  throw log.offHeapMemoryEvictionNotSupportedWithObject();
+               case OFF_HEAP:
+                  int addressCount = attributes.attribute(ADDRESS_COUNT).get();
+                  // Note this is cast to long as we have to multiply by 8 below which could overflow
+                  long actualAddressCount = OffHeapDataContainer.getActualAddressCount(addressCount);
+                  actualAddressCount *= 8;
+                  if (size < actualAddressCount) {
+                     throw log.offHeapMemoryEvictionSizeNotLargeEnoughForAddresses(size, actualAddressCount, addressCount);
+                  }
+            }
          }
       }
    }
--- a/core/src/main/java/org/infinispan/configuration/cache/MemoryConfigurationBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/cache/MemoryConfigurationBuilder.java
@@ -10,6 +10,7 @@ import org.infinispan.commons.configuration.attributes.AttributeSet;
 import org.infinispan.commons.logging.LogFactory;
 import org.infinispan.configuration.global.GlobalConfiguration;
 import org.infinispan.container.offheap.OffHeapDataContainer;
+import org.infinispan.container.offheap.UnpooledOffHeapMemoryAllocator;
 import org.infinispan.eviction.EvictionType;
 import org.infinispan.util.logging.Log;
 
@@ -97,8 +98,8 @@ public class MemoryConfigurationBuilder extends AbstractConfigurationChildBuilde
                case OFF_HEAP:
                   int addressCount = attributes.attribute(ADDRESS_COUNT).get();
                   // Note this is cast to long as we have to multiply by 8 below which could overflow
-                  long actualAddressCount = OffHeapDataContainer.getActualAddressCount(addressCount);
-                  actualAddressCount *= 8;
+                  long actualAddressCount = OffHeapDataContainer.getActualAddressCount(addressCount << 3);
+                  actualAddressCount = UnpooledOffHeapMemoryAllocator.estimateSizeOverhead(actualAddressCount);
                   if (size < actualAddressCount) {
                      throw log.offHeapMemoryEvictionSizeNotLargeEnoughForAddresses(size, actualAddressCount, addressCount);
                   }
--- a/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
@@ -44,7 +44,7 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
          // Use size of entry plus 16 for our LRU pointers
          sizeCalculator = i -> offHeapEntryFactory.getSize(i);
          // We have to make sure to count the address hash as part of our size
-         initialSize = memoryAddressCount << 3;
+         initialSize = UnpooledOffHeapMemoryAllocator.estimateSizeOverhead(memoryAddressCount << 3);
          currentSize = initialSize;
       }
       this.lruLock = new ReentrantLock();
--- a/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
@@ -28,6 +28,7 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
    protected final long maxSize;
    protected final Lock lruLock;
    protected final LongUnaryOperator sizeCalculator;
+   protected final long initialSize;
 
    protected long currentSize;
    protected long firstAddress;
@@ -38,9 +39,13 @@ public class BoundedOffHeapDataContainer extends OffHeapDataContainer {
       this.maxSize = maxSize;
       if (type == EvictionType.COUNT) {
          sizeCalculator = i -> 1;
+         initialSize = 0;
       } else {
          // Use size of entry plus 16 for our LRU pointers
          sizeCalculator = i -> offHeapEntryFactory.getSize(i);
+         // We have to make sure to count the address hash as part of our size
+         initialSize = memoryAddressCount << 3;
+         currentSize = initialSize;
       }
       this.lruLock = new ReentrantLock();
       firstAddress = 0;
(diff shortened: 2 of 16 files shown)


wburns:
This will need to go into 9.1.x as well.

wburns:
Fixed test failures. Also found a bug in clear with new sizing changes to add initial size to write on a clear.

core/src/main/java/org/infinispan/configuration/cache/MemoryConfigurationBuilder.java
```diff
@@ -86,8 +87,20 @@ public void validate() {
       long size = attributes.attribute(SIZE).get();
       if (size > 0) {
          EvictionType evictionType = attributes.attribute(EVICTION_TYPE).get();
-         if (evictionType == EvictionType.MEMORY && type != StorageType.BINARY) {
-            throw new CacheConfigurationException("MEMORY based eviction requires binary but was : " + type);
+         if (evictionType == EvictionType.MEMORY) {
+            switch (type) {
+               case OBJECT:
+                  throw new CacheConfigurationException("MEMORY based eviction is not supported with OBJECT storage");
```
tristantarrant: i18n

core/src/main/java/org/infinispan/configuration/cache/MemoryConfigurationBuilder.java
```diff
@@ -86,8 +87,20 @@ public void validate() {
...
+               case OFF_HEAP:
+                  int addressCount = attributes.attribute(ADDRESS_COUNT).get();
+                  // Note this is cast to long as we have to multiply by 8 below which could overflow
+                  long actualAddressCount = OffHeapDataContainer.findRequiredMemoryAddress(addressCount);
+                  actualAddressCount *= 8;
+                  if (size < actualAddressCount) {
+                     throw new CacheConfigurationException("MEMORY based OFF_HEAP eviction configured size " + size +
```
tristantarrant: i18n

core/src/main/java/org/infinispan/container/offheap/MemoryAddressHash.java
```diff
@@ -16,11 +16,13 @@
...
    private final int pointerCount;
+   private final OffHeapMemoryAllocator allocator;
 
-   public MemoryAddressHash(int pointers) {
+   public MemoryAddressHash(int pointers, OffHeapMemoryAllocator allocator) {
       this.pointerCount = nextPowerOfTwo(pointers);
       long bytes = ((long) pointerCount) << 3;
-      memory = MEMORY.allocate(bytes);
+      this.allocator = allocator;
+      memory = allocator.allocate(bytes);
```
tristantarrant: I think we should really check all invocations to `allocate()`: if 0 is returned in means that something went wrong.

wburns: Unsafe.cpp already throws an exception :)

http://hg.openjdk.java.net/jdk8/jdk8/hotspot/file/87ee5ee27509/src/share/vm/prims/unsafe.cpp#l595

tristantarrant: Damn, I read the implementation of os::malloc but then I forgot to unwind the stack :-)

core/src/test/java/org/infinispan/eviction/impl/MemoryBasedEvictionFunctionalTest.java
```diff
@@ -8,29 +8,52 @@
...
+      } else {
+         builder.memory().size(CACHE_SIZE + MemoryConfiguration.ADDRESS_COUNT.getDefaultValue() * 8);
+      }
       configure(builder);
       EmbeddedCacheManager cm = TestCacheManagerFactory.createCacheManager(builder);
       cache = cm.getCache();
       return cm;
    }
 
+   @Factory
```
wburns: Believe I need to remove this now.

server/rest/src/test/java/org/infinispan/rest/RestOperationsOffHeapTest.java
```diff
@@ -13,6 +13,8 @@
    public ConfigurationBuilder getDefaultCacheBuilder() {
       ConfigurationBuilder configurationBuilder = new ConfigurationBuilder();
       configurationBuilder.memory().storageType(StorageType.OFF_HEAP);
+      // Off heap and compatibility don't work together
+      configurationBuilder.compatibility().disable();
```
wburns: Also this didn't work (test still fails), have to look as to why.

core/src/test/java/org/infinispan/eviction/impl/MemoryBasedEvictionFunctionalTest.java
wburns: Nevermind had to add it to the subclass.

server/rest/src/test/java/org/infinispan/rest/RestOperationsOffHeapTest.java
wburns: Should be fixed as I made the super class not always enable compatibility.

wburns:
Fixed comments and tests should be fixed.

I also added fixes for https://issues.jboss.org/browse/ISPN-8545

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
```diff
@@ -38,9 +39,13 @@ public BoundedOffHeapDataContainer(int desiredSize, long maxSize, EvictionType t
       this.maxSize = maxSize;
       if (type == EvictionType.COUNT) {
          sizeCalculator = i -> 1;
+         initialSize = 0;
       } else {
          // Use size of entry plus 16 for our LRU pointers
          sizeCalculator = i -> offHeapEntryFactory.getSize(i);
+         // We have to make sure to count the address hash as part of our size
+         initialSize = memoryAddressCount * 8;
```
danberindei: I'd use `<< 3` to match `MemoryAddressHash`

core/src/main/java/org/infinispan/container/offheap/UnpooledOffHeapMemoryAllocator.java
```diff
@@ -18,7 +18,7 @@
    private static final Log log = LogFactory.getLog(UnpooledOffHeapMemoryAllocator.class, Log.class);
    private static final boolean trace = log.isTraceEnabled();
    private static final OffHeapMemory MEMORY = OffHeapMemory.INSTANCE;
-   private final AtomicLong amountAllocated = new AtomicLong();
+   private final LongAdder amountAllocated = new LongAdder();
```
danberindei: This is a good step, but I wonder if we could eliminate it completely when the data container is configured with MEMORY eviction.

core/src/main/java/org/infinispan/container/offheap/UnpooledOffHeapMemoryAllocator.java
```diff
@@ -44,16 +45,27 @@ public void deallocate(long memoryAddress) {
...
    }
 
    @Override
    public long getAllocatedAmount() {
-      return amountAllocated.get();
+      return amountAllocated.sum();
+   }
+
+   /**
+    * Returns a long that is aligned to 8 as most memory allocators will align to this size to more appropriately
```
danberindei: The memory allocator does not change alignment based on the requested size. In C you'd have to use `posix_memalign` to get an alignment different than 8.

core/src/test/java/org/infinispan/eviction/impl/MemoryBasedEvictionFunctionalTest.java
```diff
@@ -8,29 +8,52 @@
...
 import org.testng.annotations.Test;
 
 @Test(groups = "functional", testName = "eviction.MemoryBasedEvictionFunctionalTest")
 public class MemoryBasedEvictionFunctionalTest extends SingleCacheManagerTest {
 
    protected static final long CACHE_SIZE = 2000;
 
+   private StorageType storageType;
+
+   public MemoryBasedEvictionFunctionalTest storageType(StorageType storageType) {
```
danberindei: The test also need a `parameters()` implementation, otherwise the name is the same in the report and the second one is ignored.

server/rest/src/test/java/org/infinispan/rest/RestOperationsOffHeapTest.java
danberindei: Aren't there any test methods that require compatibility?

wburns: Sure they are at https://github.com/infinispan/infinispan/pull/5590/files#diff-56432ab419df420236bf86f0e4cea135R19

Off heap and compatibility is not a valid configuration together.

core/src/main/java/org/infinispan/container/offheap/BoundedOffHeapDataContainer.java
wburns: Sure.

core/src/main/java/org/infinispan/container/offheap/UnpooledOffHeapMemoryAllocator.java
wburns: That is possible yes. I would say we can do that later though.

core/src/main/java/org/infinispan/container/offheap/UnpooledOffHeapMemoryAllocator.java
wburns: I wasn't saying it would change based on size. This is just to account for the fact that the memory is aligned when we are counting.

core/src/test/java/org/infinispan/eviction/impl/MemoryBasedEvictionFunctionalTest.java
wburns: Okay

wburns:
@danberindei we could try that, maybe as a different PR though. I created https://issues.jboss.org/browse/ISPN-8550 to look into that.

wburns:
Updated

server/rest/src/test/java/org/infinispan/rest/RestOperationsOffHeapTest.java
danberindei: I was assuming that the base class enabled compatibility because some of its test methods actually needed compatibility, but if not that's fine.

core/src/main/java/org/infinispan/container/offheap/UnpooledOffHeapMemoryAllocator.java
danberindei: Maybe you should just rename the method to `roundUpTo8`? 

Alignment doesn't make sense for longs anyway, so `Round up the size to a multiple of 8 to account for most memory allocators aligning allocations to a multiple of 8` looks clearer to me.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -72,14 +73,21 @@ static int nextPowerOfTwo(int target) {
 
    public OffHeapDataContainer(int desiredSize) {
       lockCount = nextPowerOfTwo(Runtime.getRuntime().availableProcessors()) << 1;
+      memoryAddressCount = findRequiredMemoryAddress(desiredSize, lockCount);
+      // Unfortunately desired size directly correlates to lock size
+      locks = new StripedLock(lockCount);
+   }
+
+   public static int findRequiredMemoryAddress(int desiredSize) {
```
danberindei: I'd prefer something like `getActualAddressCount`

core/src/main/java/org/infinispan/configuration/cache/MemoryConfigurationBuilder.java
```diff
@@ -81,13 +85,24 @@ public MemoryConfigurationBuilder addressCount(int addressCount) {
...
+            switch (type) {
+               case OBJECT:
+                  throw log.offHeapMemoryEvictionNotSupportedWithObject();
+               case OFF_HEAP:
+                  int addressCount = attributes.attribute(ADDRESS_COUNT).get();
+                  // Note this is cast to long as we have to multiply by 8 below which could overflow
+                  long actualAddressCount = OffHeapDataContainer.findRequiredMemoryAddress(addressCount);
+                  actualAddressCount *= 8;
+                  if (size < actualAddressCount) {
+                     throw log.offHeapMemoryEvictionSizeNotLargeEnoughForAddresses(size, actualAddressCount, addressCount);
```
danberindei: I don't like the idea that a user could test a configuration on their machine, it would work fine, but then it would fail on a bigger machine. I'd rather make the number of locks smaller than optimal instead of making the address count bigger.

True, the configuration would be suboptimal, but at least it would work the same everywhere.

core/src/main/java/org/infinispan/container/offheap/UnpooledOffHeapMemoryAllocator.java
wburns: Sure

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
wburns: okay

core/src/main/java/org/infinispan/configuration/cache/MemoryConfigurationBuilder.java
wburns: I am not sold on this, personally. In most things you can't have the same exact settings unless the machines themselves have similar specs.

wburns:
Updated

wburns:
Updated to include 16 byte addition for ISPN-8550.

wburns:
Unfortunately that darn RestIT test is failing worse now :(

wburns:
Updated to change over to 16 byte alignment with 8 byte overhead.

wburns:
Passed CI tests.

tristantarrant:
Merged, thanks

PR #5592, opened 2017-11-18
ISPN-8542 IndexedField annotation in proto file causes java.lang.NumberFormatException

anistor:
https://issues.jboss.org/browse/ISPN-8542


anistor:
The test failures are not related. Looks good to me!

gustavonalle:
👍

PR #5594, opened 2017-11-20
ISPN-8543 Pluggable global configuration persistence

tristantarrant:
https://issues.jboss.org/browse/ISPN-8543

Note: this doesn't actually implement any other configuration persistence backends yet. Also it does not expose the configuration to the XML parsers.

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/configuration/attributes/Attribute.java
+++ b/commons/src/main/java/org/infinispan/commons/configuration/attributes/Attribute.java
@@ -2,6 +2,7 @@ package org.infinispan.commons.configuration.attributes;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
@@ -198,8 +199,13 @@ public final class Attribute<T> implements Cloneable, Matchable<Attribute<?>> {
          return false;
       if (!this.definition.equals(other.definition))
          return false;
-      if (this.definition.isGlobal())
-         return value.equals(other.value);
+      if (this.definition.isGlobal()) {
+         if (Matchable.class.isAssignableFrom(this.definition.getType())) {
+            return ((Matchable)value).matches(other.value);
+         } else {
+            return Objects.equals(value, other.value);
+         }
+      }
       return true;
    }
 
--- a/core/src/main/java/org/infinispan/configuration/cache/ClusteringConfiguration.java
+++ b/core/src/main/java/org/infinispan/configuration/cache/ClusteringConfiguration.java
@@ -137,6 +137,16 @@ public class ClusteringConfiguration implements Matchable<ClusteringConfiguratio
    }
 
    @Override
+   public boolean matches(ClusteringConfiguration other) {
+      return (attributes.matches(other.attributes) &&
+            hashConfiguration.matches(other.hashConfiguration) &&
+            l1Configuration.matches(other.l1Configuration) &&
+            partitionHandlingConfiguration.matches(other.partitionHandlingConfiguration) &&
+            stateTransferConfiguration.matches(other.stateTransferConfiguration) &&
+            syncConfiguration.matches(other.syncConfiguration));
+   }
+
+   @Override
    public String toString() {
       return "ClusteringConfiguration [hashConfiguration=" + hashConfiguration +
             ", l1Configuration=" + l1Configuration +
--- a/core/src/main/java/org/infinispan/configuration/global/GlobalStateConfiguration.java
+++ b/core/src/main/java/org/infinispan/configuration/global/GlobalStateConfiguration.java
@@ -3,6 +3,8 @@ package org.infinispan.configuration.global;
 import org.infinispan.commons.configuration.attributes.Attribute;
 import org.infinispan.commons.configuration.attributes.AttributeDefinition;
 import org.infinispan.commons.configuration.attributes.AttributeSet;
+import org.infinispan.globalstate.LocalConfigurationManager;
+import org.infinispan.globalstate.impl.EmbeddedLocalConfigurationManager;
 
 /**
  *
@@ -20,21 +22,26 @@ public class GlobalStateConfiguration {
    public static final AttributeDefinition<String> TEMPORARY_LOCATION = AttributeDefinition
          .builder("temporaryLocation", null, String.class)
             .initializer(() -> SecurityActions.getSystemProperty("java.io.tmpdir")).immutable().build();
+   public static final AttributeDefinition<LocalConfigurationManager> LOCAL_CONFIGURATION_MANAGER = AttributeDefinition
+         .builder("localConfigurationManager", null, LocalConfigurationManager.class)
+         .initializer(EmbeddedLocalConfigurationManager::new).immutable().build();
 
    public static AttributeSet attributeDefinitionSet() {
-      return new AttributeSet(GlobalStateConfiguration.class, ENABLED, PERSISTENT_LOCATION, TEMPORARY_LOCATION);
+      return new AttributeSet(GlobalStateConfiguration.class, ENABLED, PERSISTENT_LOCATION, TEMPORARY_LOCATION, LOCAL_CONFIGURATION_MANAGER);
    }
 
    private final AttributeSet attributes;
    private final Attribute<Boolean> enabled;
    private Attribute<String> persistentLocation;
    private Attribute<String> temporaryLocation;
+   private Attribute<LocalConfigurationManager> localConfigurationManager;
 
    public GlobalStateConfiguration(AttributeSet attributes) {
       this.attributes = attributes.checkProtection();
       this.enabled = attributes.attribute(ENABLED);
       this.persistentLocation = attributes.attribute(PERSISTENT_LOCATION);
       this.temporaryLocation = attributes.attribute(TEMPORARY_LOCATION);
+      this.localConfigurationManager = attributes.attribute(LOCAL_CONFIGURATION_MANAGER);
    }
 
    public boolean enabled() {
@@ -58,6 +65,13 @@ public class GlobalStateConfiguration {
       return temporaryLocation.get();
    }
 
+   /**
+    * Returns the @{@link LocalConfigurationManager}
+    */
+   public LocalConfigurationManager localConfigurationManager() {
+      return localConfigurationManager.get();
+   }
+
    public AttributeSet attributes() {
       return attributes;
    }
@@ -66,4 +80,6 @@ public class GlobalStateConfiguration {
    public String toString() {
       return "GlobalStateConfiguration [attributes=" + attributes + "]";
    }
+
+
 }
--- a/core/src/main/java/org/infinispan/globalstate/GlobalConfigurationManager.java
+++ b/core/src/main/java/org/infinispan/globalstate/GlobalConfigurationManager.java
@@ -2,7 +2,7 @@ package org.infinispan.globalstate;
 
 import java.util.EnumSet;
 
-import org.infinispan.commons.api.CacheContainerAdmin;
+import org.infinispan.commons.api.CacheContainerAdmin.AdminFlag;
 import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.factories.scopes.Scope;
 import org.infinispan.factories.scopes.Scopes;
@@ -24,7 +24,15 @@ public interface GlobalConfigurationManager {
     * @param configuration the configuration object
     * @param flags the flags to apply
     */
-   Configuration createCache(String cacheName, Configuration configuration, EnumSet<CacheContainerAdmin.AdminFlag> flags);
+   Configuration createCache(String cacheName, Configuration configuration, EnumSet<AdminFlag> flags);
+
+   /**
+    * Defines a cluster-wide cache configuration or retrieves an existing one
+    * @param cacheName the name of the configuration
+    * @param configuration the configuration object
+    * @param flags the flags to apply
+    */
+   Configuration getOrCreateCache(String cacheName, Configuration configuration, EnumSet<AdminFlag> flags);
 
    /**
     * Defines a cluster-wide cache configuration using the supplied template
@@ -32,12 +40,22 @@ public interface GlobalConfigurationManager {
     * @param template the template name to use
     * @param flags the flags to apply
     */
-   Configuration createCache(String cacheName, String template, EnumSet<CacheContainerAdmin.AdminFlag> flags);
+   Configuration createCache(String cacheName, String template, EnumSet<AdminFlag> flags);
+
+   /**
+    * Defines a cluster-wide cache configuration using the supplied template or retrieves an existing one
+    * @param cacheName the name of the configuration
+    * @param template the template name to use
+    * @param flags the flags to apply
+    */
+   Configuration getOrCreateCache(String cacheName, String template, EnumSet<AdminFlag> flags);
 
    /**
     * Removes a cluster-wide cache and its configuration
     * @param cacheName the name of the cache
     * @param flags
     */
-   void removeCache(String cacheName, EnumSet<CacheContainerAdmin.AdminFlag> flags);
+   void removeCache(String cacheName, EnumSet<AdminFlag> flags);
+
+
 }
--- a/core/src/main/java/org/infinispan/globalstate/LocalConfigurationManager.java
+++ b/core/src/main/java/org/infinispan/globalstate/LocalConfigurationManager.java
@@ -1,0 +1,52 @@
+package org.infinispan.globalstate;
+
+import java.util.EnumSet;
+import java.util.Map;
+
+import org.infinispan.commons.api.CacheContainerAdmin;
+import org.infinispan.configuration.cache.Configuration;
+import org.infinispan.manager.EmbeddedCacheManager;
+
+/**
+ * The {@link LocalConfigurationManager} is responsible for applying on each node the configuration changes initiated
+ * through the {@link org.infinispan.globalstate.GlobalConfigurationManager}.
+ *
+ * @author Tristan Tarrant
+ * @since 9.2
+ */
+
+public interface LocalConfigurationManager {
+   /**
+    * Initialization entry point for the {@link LocalConfigurationManager}
+    * @param embeddedCacheManager
+    * @param globalConfigurationManager
+    */
+   void initialize(EmbeddedCacheManager embeddedCacheManager, GlobalConfigurationManager globalConfigurationManager);
+   /**
+    * Checks whether this {@link LocalConfigurationManager} supports the supplied flags.
+    * A {@link org.infinispan.commons.CacheConfigurationException} will be thrown in case this cannot be done.
+    *
+    */
+   void validateFlags(EnumSet<CacheContainerAdmin.AdminFlag> flags);
+
+   /**
+    * Creates the cache using the supplied configuration and flags
+    *
+    * @param name the name of the cache to create
+    * @param configuration the {@link Configuration} to use
+    * @param flags the desired {@link org.infinispan.commons.api.CacheContainerAdmin.AdminFlag}s
+    */
+   void createCache(String name, Configuration configuration, EnumSet<CacheContainerAdmin.AdminFlag> flags);
+
+   /**
+    * Removes the specified cache.
+    *
+    * @param name the name of the cache to remove
+    */
+   void removeCache(String name);
+
+   /**
+    * Loads all persisted cache configurations
+    */
+   Map<String, Configuration> loadAll();
+}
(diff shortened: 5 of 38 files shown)


core/src/main/java/org/infinispan/globalstate/impl/CacheState.java
```diff
@@ -10,25 +10,32 @@
...
 
 
-   CacheState(String configuration, EnumSet<CacheContainerAdmin.AdminFlag> flags) {
+   CacheState(String template, String configuration, EnumSet<CacheContainerAdmin.AdminFlag> flags) {
+      this.template = template;
       this.configuration = configuration;
       this.flags = flags.clone();
    }
 
+   public String getTemplate() {
```
pruivo: where is the template used?

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
```diff
@@ -0,0 +1,167 @@
...
+ * The default implementation of {@link LocalConfigurationManager} for embedded scenarios.
+ *
+ * This component persists cache configurations to the {@link GlobalStateConfiguration#persistentLocation()} in a
+ * <pre>caches.xml</pre> file which is read on startup.
+ *
+ * @author Tristan Tarrant
+ * @since 9.2
+ */
+
+public class EmbeddedLocalConfigurationManager implements LocalConfigurationManager {
```
pruivo: for some reason, I was expecting 2 implementations. 
one with persistence supported, and another without it...

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
```diff
@@ -0,0 +1,167 @@
...
+         cacheManager.defineConfiguration(name, configuration);
+         log.debugf("Defined cache '%s' on '%s' using %s", name, cacheManager.getAddress(), configuration);
+      } else if (!existing.equals(configuration)) {
+         throw log.incompatibleClusterConfiguration(name, configuration, existing);
+      } else {
+         log.debugf("%s already has a cache %s with configuration %s", cacheManager.getAddress(), name, configuration);
+      }
+      if (flags.contains(CacheContainerAdmin.AdminFlag.PERSISTENT)) {
+         persistentCaches.add(name);
+         storeAll();
```
pruivo: shouldn't the store be invoked when `add()` is successful?

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
```diff
@@ -0,0 +1,167 @@
...
+   @Override
+   public void validateFlags(EnumSet<CacheContainerAdmin.AdminFlag> flags) {
+      if (flags.contains(CacheContainerAdmin.AdminFlag.PERSISTENT) && !globalConfiguration.globalState().enabled())
+         throw log.globalStateDisabled();
+   }
+
+   public void createCache(String name, Configuration configuration, EnumSet<CacheContainerAdmin.AdminFlag> flags) {
+      Configuration existing = cacheManager.getCacheConfiguration(name);
+      if (existing == null) {
+         cacheManager.defineConfiguration(name, configuration);
```
pruivo: should check the returned configuration is the same as `configuration` (or null)

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
```diff
@@ -0,0 +1,167 @@
...
+      // Ensure the cache is started
+      cacheManager.getCache(name);
+   }
+
+
+   public void removeCache(String name) {
+      log.debugf("Remove cache %s", name);
+      if (persistentCaches.contains(name)) {
+         persistentCaches.remove(name);
+         storeAll();
```
pruivo: can be invoked if the remove is successful.

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
```diff
@@ -0,0 +1,167 @@
...
+         for (String cacheName : persistentCaches) {
+            configurationMap.put(cacheName, cacheManager.getCacheConfiguration(cacheName));
+         }
+         try (FileOutputStream f = new FileOutputStream(temp)) {
+            parserRegistry.serialize(f, null, configurationMap);
+         }
+         FileLock lock = null;
+         try (FileOutputStream lockFile = new FileOutputStream(getPersistentFileLock())) {
+            lock = lockFile.getChannel().lock();
+            temp.renameTo(getPersistentFile());
```
pruivo: IMO, it should "warn" if the rename (delete below) fails.

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
```diff
@@ -41,17 +32,20 @@
 
 public class GlobalConfigurationManagerImpl implements GlobalConfigurationManager {
    private static Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
+
+   public static final String CACHE_SCOPE = "cache";
+
    private GlobalConfiguration globalConfiguration;
```
pruivo: unused

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
```diff
@@ -41,17 +32,20 @@
...
-   public static final String CACHE_SCOPE = "cache";
    private ParserRegistry parserRegistry;
-   private ConcurrentHashSet<String> persistentCaches;
+   private LocalConfigurationManager localConfigurationManager;
 
    @Inject
    public void inject(GlobalConfiguration globalConfiguration, EmbeddedCacheManager cacheManager) {
       this.globalConfiguration = globalConfiguration;
       this.cacheManager = cacheManager;
+      this.localConfigurationManager = globalConfiguration.globalState().localConfigurationManager();
```
pruivo: don't you want to create a factory for it? so it can be injected :)

core/src/main/java/org/infinispan/globalstate/LocalConfigurationManager.java
```diff
@@ -0,0 +1,51 @@
...
+    * Removes the specified cache.
+    *
+    * @param name the name of the cache to remove
+    */
+   void removeCache(String name);
+
+   /**
+    * Loads all persisted cache configurations and instantiates them
+    */
+   void loadAll();
```
pruivo: IMO, it should return a `Map<String, Configuration>` to avoid the cycli dependency `LocalConfigurationManager` <-> `GlobalConfigurationManager`

core/src/main/java/org/infinispan/globalstate/LocalConfigurationManager.java
```diff
@@ -0,0 +1,51 @@
...
+ * @since 9.2
+ */
+
+public interface LocalConfigurationManager {
+   /**
+    * Initialization entry point for the {@link LocalConfigurationManager}
+    * @param embeddedCacheManager
+    * @param globalConfigurationManager
+    */
+   void initialize(EmbeddedCacheManager embeddedCacheManager, GlobalConfigurationManager globalConfigurationManager);
```
pruivo: I'm not sure if it is needed since you can use `@Inject` annotation to set dependencies...

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
```diff
@@ -138,75 +99,38 @@ public Configuration createCache(String cacheName, String template, EnumSet<Cach
...
-         } finally {
-            if (lock != null && lock.isValid())
-               lock.release();
-            getPersistentFileLock().delete();
-         }
-
-      } catch (Exception e) {
-         throw log.errorPersistingGlobalConfiguration(e);
-      }
+   void removeCache(String name, CacheState state) {
```
pruivo: the `state` is not needed.

pruivo: name suggestion: `removeCacheLocally()` (or keep the old name)

core/src/main/java/org/infinispan/globalstate/impl/GlobalConfigurationManagerImpl.java
```diff
@@ -138,75 +99,38 @@ public Configuration createCache(String cacheName, String template, EnumSet<Cach
...
+      try {
+         CacheState state = new CacheState(template, parserRegistry.serialize(cacheName, configuration), flags);
+         getStateCache().putIfAbsent(new ScopedState(CACHE_SCOPE, cacheName), state);
+         return configuration;
+      } catch (Exception e) {
+         throw log.configurationSerializationFailed(cacheName, configuration, e);
+      }
+   }
+
+   void createCache(String name, CacheState state) {
```
pruivo: name suggestion: `createCacheLocally()` (or keep the old name)

core/src/main/java/org/infinispan/configuration/global/GlobalStateConfiguration.java
```diff
@@ -20,21 +22,26 @@
    public static final AttributeDefinition<String> TEMPORARY_LOCATION = AttributeDefinition
          .builder("temporaryLocation", null, String.class)
             .initializer(() -> SecurityActions.getSystemProperty("java.io.tmpdir")).immutable().build();
+   public static final AttributeDefinition<LocalConfigurationManager> LOCAL_CONFIGURATION_MANAGER = AttributeDefinition
+         .builder("localConfigurationManager", null, LocalConfigurationManager.class)
+         .initializer(() -> new EmbeddedLocalConfigurationManager()).immutable().build();
```
pruivo: you can use method reference `EmbeddedLocalConfigurationManager::new`
althoug, using the class name doesn't sound a bad idea either...

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
tristantarrant: I didn't feel like a simple ephemeral implementation was necessary...

core/src/main/java/org/infinispan/globalstate/LocalConfigurationManager.java
tristantarrant: Right, much better that way

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
tristantarrant: I'd prefer to throw an exception.

core/src/main/java/org/infinispan/globalstate/LocalConfigurationManager.java
tristantarrant: I'm not treating it as a component in the component registry, and explicitly initializing it in GlobalConfigurationManagerImpl.postStart()

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
tristantarrant: Not quite sure I understand what you mean here. I already handle the case where the "new" config and the existing ones are equals, in the the second else clause by printing a debug log.

tristantarrant:
@pruivo I've addressed the changes which you requested (unless otherwise noted).
I also had to introduce some additional changes in here:
- renamed the PERSISTENT flag to PERMANENT to avoid confusion with cache persistence (this is configuration persistence)
- Fix a logic problem in GlobalComponentRegistry PostStart handling (ISPN-8625)
- Unwrap ExecutionExceptions in the Hot Rod server to expose the real cause to clients (ISPN-8626)
- Add a getOrCreateCache() method to CacheContainerAdmin (ISPN-8627): I wanted to separate the commit but I've reworked some of the logic in the GlobalConfigurationManager code around it

core/src/test/java/org/infinispan/manager/CacheManagerAdminTest.java
```diff
@@ -30,22 +32,37 @@ public void testClusterCacheTest() {
...
-
+      assertEquals(cacheManagers.size(), cache.getAdvancedCache().getRpcManager().getMembers().size());
       checkConsistencyAcrossCluster("a", configuration);
 
       addClusterEnabledCacheManager();
-
       checkConsistencyAcrossCluster("a", configuration);
 
-      manager(1).administration().removeCache("a");
+      try {
```
pruivo: suggestion:
```
Exceptions.expectException(CacheConfigurationException.class,
            "Should not succeed in creating a cache with a non-existing template",
            () -> manager(0).administration().createCache("b", "nonExistingTemplate"));
```
also, there is a typo in `fail()` message (succeed)

core/src/test/java/org/infinispan/manager/CacheManagerAdminTest.java
```diff
@@ -30,22 +32,37 @@ public void testClusterCacheTest() {
...
+         manager(0).administration().createCache("b", "nonExistingTemplate");
+         fail("Should not succed in creating a cache with a non-existing template");
+      } catch (CacheConfigurationException e) {
+         // Expected
+      }
 
-      checkCacheExistenceAcrossCluster("a", false);
+      // attempt to create an existing cache
+      try {
```
pruivo: suggestion:
```
Exceptions.expectException(CacheConfigurationException.class,
            "Should not have allowed creation of existing cache",
            () -> manager(0).administration().createCache("a", configuration));
```

core/src/main/java/org/infinispan/globalstate/impl/CacheState.java
```diff
@@ -41,15 +48,17 @@ public String getConfiguration() {
...
       }
 
       @Override
       public CacheState readObject(ObjectInput input) throws IOException, ClassNotFoundException {
-         String configuration = input.readUTF();
+         String template = MarshallUtil.unmarshallString(input);
+         String configuration = MarshallUtil.unmarshallString(input);
          EnumSet<CacheContainerAdmin.AdminFlag> flags = (EnumSet<CacheContainerAdmin.AdminFlag>) input.readObject();
-         return new CacheState(configuration, flags == null ? EnumSet.noneOf(CacheContainerAdmin.AdminFlag.class) : flags);
+         return new CacheState(template, configuration, flags == null ? EnumSet.noneOf(CacheContainerAdmin.AdminFlag.class) : flags);
```
pruivo: I don't think `flags` can be `null`.

core/src/main/java/org/infinispan/manager/EmbeddedCacheManagerAdmin.java
```diff
@@ -26,4 +48,18 @@
     * @return the cache
     */
    <K, V> Cache<K, V> createCache(String name, Configuration configuration);
```
pruivo: I guess this method will also throw a `CacheException` if the cache already exists (missing in javadoc)

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
```diff
@@ -0,0 +1,170 @@
...
+ * @since 9.2
+ */
+
+public class EmbeddedLocalConfigurationManager implements LocalConfigurationManager {
+   private static Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
+   private ConcurrentHashSet<String> persistentCaches = new ConcurrentHashSet<>();
+   private EmbeddedCacheManager cacheManager;
+   private ParserRegistry parserRegistry;
+   private GlobalConfiguration globalConfiguration;
+   private GlobalConfigurationManager globalConfigurationManager;
```
pruivo: unused

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
pruivo: forget :)

core/src/main/java/org/infinispan/globalstate/impl/EmbeddedLocalConfigurationManager.java
pruivo: are you throw the exception in this PR or another? :)

tristantarrant:
@pruivo addressed all issues

core/src/test/java/org/infinispan/manager/CacheManagerAdminTest.java
```diff
@@ -4,11 +4,14 @@
 import static org.testng.AssertJUnit.assertFalse;
 import static org.testng.AssertJUnit.assertNotNull;
 import static org.testng.AssertJUnit.assertTrue;
+import static org.testng.AssertJUnit.fail;
```
pruivo: unused import

core/src/test/java/org/infinispan/configuration/serializer/ConfigurationSerializerTest.java
```diff
@@ -9,7 +9,7 @@
    @DataProvider(name = "configurationFiles")
    public Object[][] configurationFiles() {
       return new Object[][] {
-            {"configs/unified/7.0.xml"},
+            {"configs/unified/7.0.S"},
```
pruivo: oops!?

core/src/main/java/org/infinispan/globalstate/GlobalConfigurationManager.java
```diff
@@ -24,20 +24,38 @@
     * @param configuration the configuration object
     * @param flags the flags to apply
     */
-   Configuration createCache(String cacheName, Configuration configuration, EnumSet<CacheContainerAdmin.AdminFlag> flags);
+   Configuration createCache(String cacheName, Configuration configuration, EnumSet<AdminFlag> flags);
```
pruivo: to think (in another PR): the return values aren't used. probably they should return void.

tristantarrant: Yes, I'm trying to remember why I did this, and I think it's connected to the persistence of configuration into the server model, but I need to fetch that from the back of my brain.

pruivo: no worries :+1:

pruivo:
let's wait for CI before pushing the button :)

tristantarrant:
I had to revert the CacheState externalizer reader check for nullness on the EnumSet, as the EnumSet externalizer actually returns null for empty sets...

commons/src/main/java/org/infinispan/commons/configuration/attributes/Attribute.java
```diff
@@ -198,8 +198,15 @@ public boolean matches(Attribute<?> other) {
          return false;
       if (!this.definition.equals(other.definition))
          return false;
-      if (this.definition.isGlobal())
-         return value.equals(other.value);
+      if (this.definition.isGlobal()) {
+         if (Matchable.class.isAssignableFrom(this.definition.getType())) {
+            return ((Matchable)value).matches(other.value);
+         } else if (value != null) {
```
pruivo: suggestion
```
} else {
return Objects.equals(value, other.value);
} 
```

server/core/src/main/java/org/infinispan/server/core/admin/embeddedserver/CacheGetOrCreateTask.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Admin operation to create a cache
+ * Parameters:
+ * <ul>
+ *    <li><strong>name</strong> the name of the cache to create</li>
+ *    <li><strong>flags</strong> any flags, e.g. PERMANENT</li>
+ * </ul>
+ *
+ * @author Tristan Tarrant
+ * @since 9.1
```
pruivo: I guess I miss this: `@since 9.2`

core/src/main/java/org/infinispan/configuration/cache/ClusteringConfiguration.java
```diff
@@ -136,6 +136,41 @@ public AttributeSet attributes() {
...
    }
 
+   @Override
+   public boolean matches(ClusteringConfiguration other) {
+      if (attributes == null) {
+         if (other.attributes != null)
+            return false;
+      } else if (!attributes.matches(other.attributes))
+         return false;
+      if (hashConfiguration == null) {
```
pruivo: is it possible to the `*Configuration` to be `null`?

tristantarrant: I copied equals and s/equals/matches/ :)

pruivo: suggestion:
```
return hashConfiguration.matches(other.hashConfiguration) &&
          l1Configuration.matches(other.l1Configuration) &&
          ...
```

pruivo: ok

tristantarrant:
Done @pruivo

pruivo:
failures not related. integrated! thanks @tristantarrant

PR #5598, opened 2017-11-21
ISPN-7070

wburns:
CacheNotifierImplInitialTransferDistTest.testIterationBeganAndSegmentNotComplete
random failures

* Changed test to wait for both key and segment completion
* Fixed issue where iterator could be complete before segment was

https://issues.jboss.org/browse/ISPN-7070

### DIFF
--- a/core/src/main/java/org/infinispan/notifications/cachelistener/DistributedQueueingSegmentListener.java
+++ b/core/src/main/java/org/infinispan/notifications/cachelistener/DistributedQueueingSegmentListener.java
@@ -77,12 +77,15 @@ class DistributedQueueingSegmentListener<K, V> extends BaseQueueingSegmentListen
 
    @Override
    public void transferComplete() {
-      justCompletedSegments.forEach(completeSegment);
+      // Complete any segments that are still open - means we just didn't receive the completion yet
+      // Iterator is guaranteed to not complete until all segments are complete.
+      for (int i = 0; i < queues.length(); ++i) {
+         if (queues.get(i) != null) {
+            completeSegment(i);
+         }
+      }
       completed.set(true);
       notifiedKeys.clear();
-      for (int i = 0; i < queues.length(); ++i) {
-         queues.set(i, null);
-      }
    }
 
    public Object markKeyAsProcessing(K key) {
@@ -99,7 +102,7 @@ class DistributedQueueingSegmentListener<K, V> extends BaseQueueingSegmentListen
    }
 
    private void completeSegment(int segment) {
-      Queue<KeyValuePair<CacheEntryEvent<K, V>, ListenerInvocation<Event<K, V>>>> queue = queues.get(segment);
+      Queue<KeyValuePair<CacheEntryEvent<K, V>, ListenerInvocation<Event<K, V>>>> queue = queues.getAndSet(segment, null);
       if (queue != null) {
          if (trace) {
             log.tracef("Completed segment %s", segment);
@@ -108,7 +111,6 @@ class DistributedQueueingSegmentListener<K, V> extends BaseQueueingSegmentListen
             // The InitialTransferInvocation already did the converter if needed
             event.getValue().invoke(event.getKey());
          }
-         queues.set(segment, null);
       }
    }
 
--- a/core/src/main/java/org/infinispan/stream/impl/ClusterStreamManagerImpl.java
+++ b/core/src/main/java/org/infinispan/stream/impl/ClusterStreamManagerImpl.java
@@ -452,8 +452,8 @@ public class ClusterStreamManagerImpl<K> implements ClusterStreamManager<K> {
       }
 
       @Override
-      public void subscribe(Subscriber<? super V> s, Consumer<? super Supplier<PrimitiveIterator.OfInt>> onLostSegments,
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/infinispan/notifications/cachelistener/DistributedQueueingSegmentListener.java
```diff
@@ -77,12 +77,15 @@ public boolean handleEvent(EventWrapper<K, V, CacheEntryEvent<K, V>> wrapped, Li
 
    @Override
    public void transferComplete() {
-      justCompletedSegments.forEach(completeSegment);
-      completed.set(true);
-      notifiedKeys.clear();
+      // Complete any segments that are still open - means we just didn't receive the completion yet
```
danberindei: I'm having some trouble parsing this comment, could you elaborate on why we `completeSegment()` may not have been called for some of the segments?

core/src/main/java/org/infinispan/stream/impl/ClusterStreamManagerImpl.java
```diff
@@ -452,8 +452,8 @@ public void forgetOperation(Object id) {
       }
 
       @Override
-      public void subscribe(Subscriber<? super V> s, Consumer<? super Supplier<PrimitiveIterator.OfInt>> onLostSegments,
-            Consumer<? super Supplier<PrimitiveIterator.OfInt>> onSegmentsComplete) {
+      public void subscribe(Subscriber<? super V> s, Consumer<? super Supplier<PrimitiveIterator.OfInt>> onSegmentsComplete,
+            Consumer<? super Supplier<PrimitiveIterator.OfInt>> onLostSegments) {
```
danberindei: `RemoteIteratorPublisher` default subscribe method still uses the old parameter order (although neither supplier does anything).

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -720,41 +718,36 @@ public void valueAdded(Object key) {
...
          // added for a response before the segments are completed.
          // The valueIterated method can be invoked at any point however.
          // See ClusterStreamManagerImpl$ClusterStreamSubscription.sendRequest where the added and segments are called into
-         AtomicBoolean shouldNotify = new AtomicBoolean(true);
+         ByRef<Supplier<PrimitiveIterator.OfInt>> segmentsToNotify = new ByRef<>(segments);
          Object key = currentKey.get();
          if (key != null) {
             pendingSegments.compute(key, (k, v) -> {
-               // This check is if iterator caught up to us
+               // The iterator has consumed all the keys, so there is no reason to wait and just notify of segment
```
danberindei: s/wait and/wait:/

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -720,41 +718,36 @@ public void valueAdded(Object key) {
...
-               shouldNotify.set(false);
-               if (v == null) {
-                  return segments;
-               } else {
-                  // This is ugly, but we shouldn't hit this often (need a response with no entries and completed
-                  // segments and iterator that hasn't caught up)
-                  return () -> Stream.of(segments, v).flatMapToInt(s -> toStream(s.get())).iterator();
-               }
+               // We save previous segments if there were any. This should normally be null since if we iterate
+               // over an entry then it will null this. So this is mostly for segments that have no entries
```
danberindei: If all the segments that will ever be set in `segmentsToNotify` are empty, then I'd make that clearer. Something like "valueIterated notifies the completion of non-empty segments, but we need to notify the completion of empty segments here".

danberindei:
Unfortunately the test now fails in CI: http://ci.infinispan.org/job/Infinispan/job/PR-5598/1/

wburns:
Oops I was just testing the remove one as I was able to get that one to fail about 2% of the time :) Guessing I just need to tweak the test method a bit more.

core/src/main/java/org/infinispan/stream/impl/ClusterStreamManagerImpl.java
wburns: Yeah this wasn't a bug, just variables in wrong spots :) I can fix the other.

core/src/main/java/org/infinispan/notifications/cachelistener/DistributedQueueingSegmentListener.java
wburns: The reason it wouldn't be completed is that the completion is done on a different thread. I will see if I can tweak it to be done in a slightly different order to provide more clear ordering semantics.

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
wburns: Sure

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
wburns: Sure

wburns:
Updated and tests should be fixed. I haven't gotten into the ordering of completion to be before iterator, but will keep looking. However this should be able to integrated as is even if it doesn't get changed.

PR #5600, opened 2017-11-21
ISPN-6395 Unify clustered queries with non clustered queries

gustavonalle:
https://issues.jboss.org/browse/ISPN-6395

~Preview, just to gather feedback (from @anistor mainly) on the backwards compatible API changes.~

This PR introduces a ```IndexQueryQueryMode``` that specifies **how an indexed query is executed**. BROADCAST mean the query is sent to all nodes (and results aggregated), while ~CALLER~ FETCH (the default) executes the query directly (thus needs a distributed index to work).

Why is this relevant? Because with BROADCAST, each node can index its own data on its own index, so both indexing and querying is way more scalable than having a single global index handled by the ```InfinispanIndexManager```

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/query/RemoteQuery.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/query/RemoteQuery.java
@@ -13,6 +13,7 @@ import org.infinispan.client.hotrod.impl.operations.QueryOperation;
 import org.infinispan.protostream.ProtobufUtil;
 import org.infinispan.protostream.SerializationContext;
 import org.infinispan.protostream.WrappedMessage;
+import org.infinispan.query.dsl.IndexedQueryMode;
 import org.infinispan.query.dsl.QueryFactory;
 import org.infinispan.query.dsl.impl.BaseQuery;
 import org.infinispan.query.remote.client.QueryResponse;
@@ -25,22 +26,25 @@ public final class RemoteQuery extends BaseQuery {
 
    private final RemoteCacheImpl<?, ?> cache;
    private final SerializationContext serializationContext;
+   private final IndexedQueryMode indexedQueryMode;
 
    private List<?> results = null;
    private int totalResults;
 
    RemoteQuery(QueryFactory queryFactory, RemoteCacheImpl<?, ?> cache, SerializationContext serializationContext,
-               String queryString) {
+               String queryString, IndexedQueryMode indexQueryMode) {
       super(queryFactory, queryString);
       this.cache = cache;
       this.serializationContext = serializationContext;
+      this.indexedQueryMode = indexQueryMode;
    }
 
-   RemoteQuery(QueryFactory queryFactory, RemoteCacheImpl cache, SerializationContext serializationContext,
+   RemoteQuery(QueryFactory queryFactory, RemoteCacheImpl<?, ?> cache, SerializationContext serializationContext,
                String queryString, Map<String, Object> namedParameters, String[] projection, long startOffset, int maxResults) {
       super(queryFactory, queryString, namedParameters, projection, startOffset, maxResults);
       this.cache = cache;
       this.serializationContext = serializationContext;
+      this.indexedQueryMode = IndexedQueryMode.FETCH;
    }
 
    @Override
@@ -110,6 +114,10 @@ public final class RemoteQuery extends BaseQuery {
 
    public RemoteCache<?, ?> getCache() {
       return cache;
+   }
+
+   public IndexedQueryMode getIndexedQueryMode() {
+      return indexedQueryMode;
    }
 
    @Override
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/query/RemoteQueryStringTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/query/RemoteQueryStringTest.java
@@ -97,10 +97,14 @@ public class RemoteQueryStringTest extends QueryStringTest {
       return cache;
    }
 
+   protected int getNodesCount() {
+      return 1;
+   }
+
    @Override
    protected void createCacheManagers() throws Throwable {
       ConfigurationBuilder cfg = getConfigurationBuilder();
-      createClusteredCaches(1, cfg, true);
+      createClusteredCaches(getNodesCount(), cfg, true);
 
       cache = manager(0).getCache();
 
@@ -194,9 +198,7 @@ public class RemoteQueryStringTest extends QueryStringTest {
     */
    @Override
    public void testInstant1() throws Exception {
-      QueryFactory qf = getQueryFactory();
-
-      Query q = qf.create("from " + getModelFactory().getUserTypeName() + " u where u.creationDate = " + Instant.parse("2011-12-03T10:15:30Z").toEpochMilli());
+      Query q = createQueryFromString("from " + getModelFactory().getUserTypeName() + " u where u.creationDate = " + Instant.parse("2011-12-03T10:15:30Z").toEpochMilli());
 
       List<User> list = q.list();
       assertEquals(3, list.size());
@@ -208,18 +210,14 @@ public class RemoteQueryStringTest extends QueryStringTest {
     */
    @Override
    public void testInstant2() throws Exception {
-      QueryFactory qf = getQueryFactory();
-
-      Query q = qf.create("from " + getModelFactory().getUserTypeName() + " u where u.passwordExpirationDate = " + Instant.parse("2011-12-03T10:15:30Z").toEpochMilli());
+      Query q = createQueryFromString("from " + getModelFactory().getUserTypeName() + " u where u.passwordExpirationDate = " + Instant.parse("2011-12-03T10:15:30Z").toEpochMilli());
 
       List<User> list = q.list();
       assertEquals(3, list.size());
    }
 
    public void testCustomFieldAnalyzer() throws Exception {
-      QueryFactory qf = getQueryFactory();
-
-      Query q = qf.create("from sample_bank_account.AnalyzerTestEntity where f1:'test'");
+      Query q = createQueryFromString("from sample_bank_account.AnalyzerTestEntity where f1:'test'");
 
       List<AnalyzerTestEntity> list = q.list();
       assertEquals(2, list.size());
--- a/query-dsl/src/main/java/org/infinispan/query/dsl/IndexedQueryMode.java
+++ b/query-dsl/src/main/java/org/infinispan/query/dsl/IndexedQueryMode.java
@@ -1,0 +1,20 @@
+package org.infinispan.query.dsl;
+
+/**
+ * Defines the execution mode of an indexed query.
+ *
+ * @since 9.2
+ */
+public enum IndexedQueryMode {
+   /**
+    * Query is sent to all nodes, and results are combined before returning to the caller. This allows each node to have
+    * its own index, and the query will return the cluster wide results.
+    */
+   BROADCAST,
+
+   /**
+    * Query is executed locally in the caller. The whole index must be available in order to return full
+    * results, otherwise only results available at the caller's local index are returned.
+    */
+   FETCH
+}
--- a/query-dsl/src/main/java/org/infinispan/query/dsl/QueryFactory.java
+++ b/query-dsl/src/main/java/org/infinispan/query/dsl/QueryFactory.java
@@ -18,6 +18,12 @@ public interface QueryFactory {
    Query create(String queryString);
 
    /**
+    * Creates a Query based on an Ickle query string
+    * @param queryMode the {@link IndexedQueryMode} dictating the indexed query execution mode if applicable.
+    */
+   Query create(String queryString, IndexedQueryMode queryMode);
+
+   /**
     * Creates a QueryBuilder for the given entity type.
     *
     * @param entityType the Class of the entity
--- a/query/src/main/java/org/infinispan/query/QueryDefinition.java
+++ b/query/src/main/java/org/infinispan/query/QueryDefinition.java
@@ -1,0 +1,112 @@
+package org.infinispan.query;
+
+import java.util.Optional;
+
+import org.apache.lucene.search.Filter;
+import org.apache.lucene.search.Sort;
+import org.hibernate.search.filter.FullTextFilter;
+import org.hibernate.search.query.engine.spi.HSQuery;
+import org.infinispan.AdvancedCache;
+import org.infinispan.query.dsl.embedded.impl.EmbeddedQueryEngine;
+import org.infinispan.query.dsl.embedded.impl.HsQueryRequest;
+import org.infinispan.query.dsl.embedded.impl.QueryEngine;
+import org.infinispan.query.logging.Log;
+import org.infinispan.util.logging.LogFactory;
+
+/**
+ * Stores the query to be executed in a cache in either a String or {@link HSQuery} form together with pagination
+ * and sort information.
+ *
+ * @since 9.2
+ */
+public class QueryDefinition {
+
+   private static final Log log = LogFactory.getLog(QueryDefinition.class, Log.class);
+
+   private String queryString;
+   private HSQuery hsQuery;
+   private int maxResults = 100;
+   private int firstResult;
+
+   private transient Sort sort;
+
+   public QueryDefinition(String queryString) {
+      this.queryString = queryString;
+   }
+
+   public QueryDefinition(HSQuery hsQuery) {
+      this.hsQuery = hsQuery;
+   }
+
+   public Optional<String> getQueryString() {
+      return Optional.ofNullable(queryString);
+   }
+
+   public void initialize(AdvancedCache<?, ?> cache) {
+      if (hsQuery == null) {
+         QueryEngine queryEngine = cache.getComponentRegistry().getComponent(EmbeddedQueryEngine.class);
+         HsQueryRequest hsQueryRequest = queryEngine.createHsQuery(queryString);
+         this.hsQuery = hsQueryRequest.getHsQuery();
+         this.sort = hsQueryRequest.getSort();
+         hsQuery.firstResult(firstResult);
+         hsQuery.maxResults(maxResults);
+      }
+   }
+
+   public HSQuery getHsQuery() {
+      if (hsQuery == null) {
+         throw new IllegalStateException("The QueryDefinition has not been initialized, make sure to call initialize(...) first");
+      }
+      return hsQuery;
+   }
+
+   public int getMaxResults() {
+      return maxResults;
+   }
+
+   public void setMaxResults(int maxResults) {
+      this.maxResults = maxResults;
+      if (hsQuery != null) {
+         hsQuery.maxResults(maxResults);
+      }
+   }
+
+   public int getFirstResult() {
+      return firstResult;
+   }
+
+   public void setFirstResult(int firstResult) {
+      if (hsQuery != null) {
+         hsQuery.firstResult(firstResult);
+      }
+      this.firstResult = firstResult;
+   }
+
+   public Sort getSort() {
+      return sort;
+   }
+
+   public void setSort(Sort sort) {
+      if (queryString != null) {
+         throw log.sortNotSupportedWithQueryString();
+      }
+      hsQuery.sort(sort);
+      this.sort = sort;
+   }
+
+
+   public void filter(Filter filter) {
+      if (queryString != null) throw log.filterNotSupportedWithQueryString();
+      hsQuery.filter(filter);
+   }
+
+   public FullTextFilter enableFullTextFilter(String name) {
+      if (queryString != null) throw log.filterNotSupportedWithQueryString();
+      return hsQuery.enableFullTextFilter(name);
+   }
+
+   public void disableFullTextFilter(String name) {
+      if (queryString != null) throw log.filterNotSupportedWithQueryString();
+      hsQuery.disableFullTextFilter(name);
+   }
+}
--- a/query/src/main/java/org/infinispan/query/QueryDefinition.java
+++ b/query/src/main/java/org/infinispan/query/QueryDefinition.java
@@ -1,5 +1,7 @@
 package org.infinispan.query;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
@@ -35,6 +37,7 @@ public class QueryDefinition {
    private Set<String> sortableFields;
    private Class<?> indexedType;
 
+   private final Map<String, Object> namedParameters = new HashMap<>();
    private transient Sort sort;
 
    public QueryDefinition(String queryString) {
@@ -63,9 +66,9 @@ public class QueryDefinition {
          HsQueryRequest hsQueryRequest;
          if (indexedType != null && sortableFields != null) {
             IndexedTypeMap<CustomTypeMetadata> metadata = createMetadata();
-            hsQueryRequest = queryEngine.createHsQuery(queryString, metadata);
+            hsQueryRequest = queryEngine.createHsQuery(queryString, metadata, namedParameters);
          } else {
-            hsQueryRequest = queryEngine.createHsQuery(queryString, null);
+            hsQueryRequest = queryEngine.createHsQuery(queryString, null, namedParameters);
          }
          this.hsQuery = hsQueryRequest.getHsQuery();
          this.sort = hsQueryRequest.getSort();
@@ -91,6 +94,18 @@ public class QueryDefinition {
       if (hsQuery != null) {
          hsQuery.maxResults(maxResults);
       }
+   }
+
+   public void setNamedParameters(Map<String, Object> params) {
+      if (params == null) {
+         namedParameters.clear();
+      } else {
+         namedParameters.putAll(params);
+      }
+   }
+
+   public Map<String, Object> getNamedParameters() {
+      return namedParameters;
    }
 
    public int getFirstResult() {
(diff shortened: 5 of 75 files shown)


gustavonalle:
Updated

server/rest/src/test/java/org/infinispan/rest/search/SingleNodeLocalIndexTest.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+@Test(groups = "functional", testName = "rest.SingleNodeLocalIndexTest")
+public class SingleNodeLocalIndexTest extends BaseRestSearchTest {
+
+   @Override
+   ConfigurationBuilder getConfigBuilder() {
+      ConfigurationBuilder configurationBuilder = new ConfigurationBuilder();
+      configurationBuilder.clustering().cacheMode(CacheMode.LOCAL);
+      configurationBuilder.indexing().index(Index.ALL)
+            .addProperty("default.directory_provider", "ram");
```
tristantarrant: s/ram/local-heap/

query-dsl/src/main/java/org/infinispan/query/dsl/IndexedQueryMode.java
```diff
@@ -0,0 +1,20 @@
+package org.infinispan.query.dsl;
+
+/**
+ * Defines the execution mode of an indexed query.
+ *
+ * @since 9.2
+ */
+public enum IndexedQueryMode {
```
tristantarrant: Shouldn't we also have an AUTO mode which chooses the best strategy depending on the underlying indexing mode ?

gustavonalle: 👍  I was planning to add it later (another PR)

server/rest/src/test/java/org/infinispan/rest/search/SingleNodeLocalIndexTest.java
gustavonalle: Muscle memory :)

gustavonalle:
Updated again. I removed one commit which was just about refactoring the REST search search tests, and broadcast support for REST is not 100% yet.

gustavonalle:
updated one more time. Removed a "implements Serializable" left behind.

anistor:
I'll have a look again also.

gustavonalle:
CI seems unstable. Triggering another build

query-dsl/src/main/java/org/infinispan/query/dsl/QueryFactory.java
```diff
@@ -17,6 +17,12 @@
     */
    Query create(String queryString);
 
+   /**
+    * Creates a Query based on an Ickle query string
+    * @param queryMode the {@link IndexedQueryMode} dictating the indexed query execution mode if applicable.
+    */
+   Query create(String queryString, IndexedQueryMode queryMode);
```
anistor: I'm looking for the way to build a QueryBuilder-DSL based query and also specify IndexedQueryMode but could not find it. Maybe I'm missing something.

And another thing, I'm not sure whether we need to specify IndexedQueryMode at query creation time; maybe it's better to specify it at execution time.

gustavonalle: I did not touch the DSL, I though it'd be in maintenance mode only?

WRT IndexedQueryMode at execution time, I've chosen at creation time because:
* It's how it's working now: SearchManager.getQuery vs SearchManager.getClusteredQuery
* I tried to avoid having to change the API everywhere, i.e., ```query.list(mode), query.iterator(mode), query.getResultSize(mode)``` for each of the query types

gustavonalle:
CI is green!

query-dsl/src/main/java/org/infinispan/query/dsl/IndexedQueryMode.java
anistor: +1 for introducing an AUTO mode. I think that mode should become our default. Do we actually need a value called DEFAULT?

I think the one we now call DEFAULT should be named differently. We need to find a better name that describes in a word that: 'it is executed in a single phase in the caller'.  What is the opposite of clustered/distributed? Can't find a good name right now. :)

And then we could also add a DEFAULT enum value, for convenience, that is just an alias to one of the other constants (and warn users that the DEFAULT might be different in next major ispn version, as we might invent new query mechanisms). :)

gustavonalle: AUTO will come in a later PR, I possible can squeeze it in before the year ends.

Suggestions for renaming DEFAULT ? 

* SINGLE_PHASE
* SINGLE_STEP
* LOCAL
* NON_BROADCAST :)
* CALLER

query/src/main/java/org/infinispan/query/QueryDefinition.java
```diff
@@ -0,0 +1,104 @@
...
+
+   public int getMaxResults() {
+      return maxResults;
+   }
+
+   public int getFirstResult() {
+      return firstResult;
+   }
+
+   public void sort(Sort sort) {
```
anistor: setSort? to match with getSort.

query/src/main/java/org/infinispan/query/QueryDefinition.java
```diff
@@ -0,0 +1,104 @@
...
+         this.hsQuery = hsQueryRequest.getHsQuery();
+         this.sort = hsQueryRequest.getSort();
+         hsQuery.firstResult(firstResult);
+         hsQuery.maxResults(maxResults);
+         hsQuery.projection(hsQueryRequest.getProjections());
+      }
+      return hsQuery;
+   }
+
+   public void setMaxResults(int maxResults) {
```
anistor: I'd keep getter-setter pairs like getMaxResults/setMaxResults and getFirstResult/setFirstResult together without mising them with other methods like ```sort```.

gustavonalle:
Updated again. Exposed the ```IndexedQueryMode``` for Rest and Hot Rod. Did not change previous commits, so the reviews done so far are still valid!

query/src/main/java/org/infinispan/query/QueryDefinition.java
```diff
@@ -0,0 +1,104 @@
...
+      this.hsQuery = hsQuery;
+   }
+
+   public Optional<String> getQueryString() {
+      return Optional.ofNullable(queryString);
+   }
+
+   public HSQuery getHsQuery(AdvancedCache<?, ?> cache) {
+      if (hsQuery == null) {
+         QueryEngine queryEngine = cache.getComponentRegistry().getComponent(EmbeddedQueryEngine.class);
```
anistor: Would it be possible to pass the EmbeddedQueryEngine somehow to QueryDefinition instead of grabbing it from the component registry? Maybe in the ``` QueryDefinition(String queryString)``` constructor.

anistor:
Ok. I'm still looking here at some aspects. Please do not merge it yet.

query/src/main/java/org/infinispan/query/QueryDefinition.java
gustavonalle: Unfortunately no, the ```QueryDefinition``` is what is broadcast, so the serialization layers is responsible to construct it.

query-dsl/src/main/java/org/infinispan/query/dsl/IndexedQueryMode.java
gustavonalle: I've been thinking and all the names above have issues, apart for NON_BROADCAST, but this is silly. I'll go with FETCH which I believe reflects accurately  the fact that the query is tun locally and will read the index, potentially from remote nodes.

gustavonalle:
Addressed reviews, changed ```IndexedQueryMode.DEFAULT``` to ```IndexedQueryMode.FETCH``` and updated documentation.

query/src/main/java/org/infinispan/query/clustered/ClusteredCacheQueryImpl.java
```diff
@@ -39,35 +42,47 @@
...
    @Override
    public CacheQuery<E> firstResult(int firstResult) {
       this.firstResult = firstResult;
+      this.queryDefinition.setFirstResult(firstResult);
       return this;
    }
 
    @Override
    public CacheQuery<E> sort(Sort sort) {
       this.sort = sort;
```
anistor: Is the ```sort``` field used at all? I see it being set but don't see it used anywhere.

query/src/main/java/org/infinispan/query/QueryDefinition.java
```diff
@@ -0,0 +1,154 @@
...
+
+   public void setMaxResults(int maxResults) {
+      this.maxResults = maxResults;
+      if (hsQuery != null) {
+         hsQuery.maxResults(maxResults);
+      }
+   }
+
+   public void setNamedParameters(Map<String, Object> params) {
+      if (params != null) params.forEach(this.namedParameters::put);
```
anistor: Is forEach better that putAll? I suppose we also need to take care of the case when params are null.
```
   if (params == null) {
       namedParameters.clear();
   } else {
       namedParameters.putAll(params); 
   }
```

query/src/main/java/org/infinispan/query/dsl/embedded/impl/QueryEngine.java
```diff
@@ -687,10 +706,106 @@ protected IckleFilterAndConverter createFilter(String queryString, Map<String, O
...
+
+      if (log.isDebugEnabled()) {
+         log.debugf("Building Lucene query for : %s", queryString);
+      }
+
+      IckleParsingResult<TypeMetadata> parsingResult = parse(queryString);
+      if (parsingResult.hasGroupingOrAggregations()) {
+         throw log.groupAggregationsNotSupported();
+      }
+      LuceneQueryParsingResult luceneParsingResult = transformParsingResult(parsingResult, EMPTY_MAP);
```
anistor: Replacing EMPTY_MAP with emptyMap() spares of a generics warning.

query/src/main/java/org/infinispan/query/clustered/ClusteredCacheQueryImpl.java
gustavonalle: ```Sort``` is used in the ctor of ```DistributedIterator```

gustavonalle: My bad, was looking at the wrong branch. ```sort``` is unused, I'll remove it

query/src/main/java/org/infinispan/query/QueryDefinition.java
gustavonalle: ok

gustavonalle:
Updated

gustavonalle:
Rebased

query/src/main/java/org/infinispan/query/dsl/embedded/impl/DelegatingQuery.java
```diff
@@ -22,6 +23,7 @@
    private static final Log log = Logger.getMessageLogger(Log.class, DelegatingQuery.class.getName());
 
    private final QueryEngine<TypeMetadata> queryEngine;
+   private IndexedQueryMode queryMode;
```
anistor: queryMode can be final

query/src/main/java/org/infinispan/query/dsl/embedded/impl/DelegatingQuery.java
```diff
@@ -31,8 +33,13 @@
    private BaseQuery query;
 
    DelegatingQuery(QueryEngine<TypeMetadata> queryEngine, QueryFactory queryFactory, String queryString) {
```
anistor: This constructor is only used once in EmbeddedQueryFactory. I'd remove it.

query/src/main/java/org/infinispan/query/dsl/embedded/impl/EmbeddedLuceneQuery.java
```diff
@@ -34,16 +35,18 @@
     * The cached results, lazily evaluated.
     */
    private List<Object> results;
+   private IndexedQueryMode queryMode;
```
anistor: This can be final.

query/src/main/java/org/infinispan/query/dsl/embedded/impl/QueryEngine.java
```diff
@@ -706,12 +821,14 @@ protected IckleFilterAndConverter createFilter(String queryString, Map<String, O
...
       }
 
-      CacheQuery<?> cacheQuery = makeCacheQuery(ickleParsingResult, luceneQuery);
-      if (luceneParsingResult.getSort() != null) {
-         cacheQuery = cacheQuery.sort(luceneParsingResult.getSort());
-      }
-      if (luceneParsingResult.getProjections() != null) {
-         cacheQuery = cacheQuery.projection(luceneParsingResult.getProjections());
+      CacheQuery<?> cacheQuery = makeCacheQuery(ickleParsingResult, luceneQuery, queryMode, namedParameters);
+      if (queryMode != IndexedQueryMode.BROADCAST) {
```
anistor: Why no sorting and projection in broadcast mode?

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/RemoteQueryEngine.java
```diff
@@ -89,14 +90,19 @@ protected Query makeTypeQuery(Query query, String targetEntityName) {
...
+      final Set<String> sortableFields = indexingMetadata != null ? indexingMetadata.getSortableFields() : Collections.emptySet();
+      IndexedTypeMap<CustomTypeMetadata> queryMetadata = IndexedTypeMaps.singletonMapping(ProtobufValueWrapper.INDEXING_TYPE, () -> sortableFields);
+      RemoteQueryDefinition queryDefinition;
+      if (queryMode == IndexedQueryMode.BROADCAST) {
+         queryDefinition = new RemoteQueryDefinition(ickleParsingResult.getQueryString());
+      } else {
+         queryDefinition = new RemoteQueryDefinition(getSearchFactory().createHSQuery(luceneQuery, queryMetadata));
+      }
+      queryDefinition.setNamedParameters(namedParameters);
+      SearchManagerImplementor searchManager = getSearchManager().unwrap(SearchManagerImplementor.class);
```
anistor: I see ```getSearchManager().unwrap(SearchManagerImplementor.class)``` being called a few times. I think at this point it is better to modify QueryEngine.getSearchManager() to return SearchManagerImplementor and do the unwrapping only first time it is obtained.

gustavonalle: ok

query/src/main/java/org/infinispan/query/dsl/embedded/impl/EmbeddedLuceneQuery.java
gustavonalle: so ```final``` shall it be

query/src/main/java/org/infinispan/query/dsl/embedded/impl/QueryEngine.java
gustavonalle: Both are part of the query string in broadcast mode

anistor: Seems like something I would write a short comment about, so I can remember later :)

gustavonalle:
Updated

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/query/RemoteQuery.java
```diff
@@ -34,13 +36,23 @@
...
+
+   RemoteQuery(QueryFactory queryFactory, RemoteCacheImpl<?, ?> cache, SerializationContext serializationContext,
+               String queryString, IndexedQueryMode indexQueryMode) {
+      super(queryFactory, queryString);
+      this.cache = cache;
+      this.serializationContext = serializationContext;
+      this.indexedQueryMode = indexQueryMode;
    }
 
    RemoteQuery(QueryFactory queryFactory, RemoteCacheImpl cache, SerializationContext serializationContext,
```
anistor: It's a good opportunity to fix the cache param here; should be ```RemoteCacheImpl<?, ?>```

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/query/RemoteQuery.java
```diff
@@ -34,13 +36,23 @@
       super(queryFactory, queryString);
       this.cache = cache;
       this.serializationContext = serializationContext;
+      this.indexedQueryMode = IndexedQueryMode.FETCH;
+   }
+
+   RemoteQuery(QueryFactory queryFactory, RemoteCacheImpl<?, ?> cache, SerializationContext serializationContext,
```
anistor: This newly added constructor should be used in RemoteQueryFactory.create(String queryString), and the old constructor (without IndexedQueryMode) can be removed.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/query/RemoteQueryStringTest.java
```diff
@@ -97,10 +97,14 @@ protected ModelFactory getModelFactory() {
       return cache;
    }
 
+   protected int getNodesCount() {
```
anistor: I don't think this is used. The derived class overrides ```createCacheManagers``` anyway.

gustavonalle: I see, let me sort this out

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/query/RemoteQuery.java
gustavonalle: ok

query/src/main/java/org/infinispan/query/clustered/AbstractQueryDefinitionExternalizer.java
```diff
@@ -0,0 +1,71 @@
...
+public abstract class AbstractQueryDefinitionExternalizer<Q extends QueryDefinition> implements AdvancedExternalizer<Q> {
+
+   @Override
+   public void writeObject(ObjectOutput output, Q object) throws IOException {
+      if (object.getQueryString().isPresent()) {
+         output.writeBoolean(true);
+         output.writeUTF(object.getQueryString().get());
+      } else {
+         output.writeBoolean(false);
+         output.writeObject(object.getHsQuery(null));
```
anistor: I don't understand how this can work with ```null``` cache parameter.

gustavonalle: ops, let me clean this up

gustavonalle:
addressed latest reviews

query/src/main/java/org/infinispan/query/QueryDefinition.java
```diff
@@ -0,0 +1,161 @@
...
+         this.hsQuery = hsQueryRequest.getHsQuery();
+         this.sort = hsQueryRequest.getSort();
+         hsQuery.firstResult(firstResult);
+         hsQuery.maxResults(maxResults);
+         hsQuery.projection(hsQueryRequest.getProjections());
+      }
+   }
+
+   public HSQuery getHsQuery() {
+      return hsQuery;
```
anistor: Calling this method should throw an ```IllegalStateException``` if ```hsQuery``` is ```null```, ie ```initialize(...)``` was not called prior to this.

query/src/main/java/org/infinispan/query/impl/CacheQueryImpl.java
```diff
@@ -57,11 +60,26 @@ public CacheQueryImpl(Query luceneQuery, SearchIntegrator searchFactory, Advance
             cache, keyTransformationHandler);
    }
 
+   public CacheQueryImpl(Query luceneQuery, SearchIntegrator searchFactory, AdvancedCache<?, ?> cache,
```
anistor: One of the constructors in this class is no longer used.

server/rest/src/main/java/org/infinispan/rest/search/InfinispanSearchRequest.java
```diff
@@ -59,10 +61,12 @@ private QueryRequest getQueryRequest() throws IOException {
       if (request.method() == HttpMethod.GET) {
          String queryString = getParameterValue(QUERY_STRING);
```
anistor: The whole body of this ```if```statement would look better if extracted as a separate getQueryFromString method in the spirit of ```getQueryFromJSON```

query/src/main/java/org/infinispan/query/clustered/ClusteredQueryCommandType.java
```diff
@@ -55,11 +55,11 @@ public ClusteredQueryCommandWorker getNewInstance() {
 
    protected abstract ClusteredQueryCommandWorker getNewInstance();
 
-   public ClusteredQueryCommandWorker getCommand(Cache<?, ?> cache, HSQuery query, UUID lazyQueryId,
-            int docIndex) {
+   public ClusteredQueryCommandWorker getCommand(Cache<?, ?> cache, QueryDefinition queryDefinition, UUID lazyQueryId,
+                                                 int docIndex) {
       ClusteredQueryCommandWorker command = null;
```
anistor: Declaration and initialization of ```command``` can be done in same line.

query/src/main/java/org/infinispan/query/dsl/embedded/impl/QueryEngine.java
```diff
@@ -687,10 +706,108 @@ protected IckleFilterAndConverter createFilter(String queryString, Map<String, O
...
+   public <E> CacheQuery<E> buildCacheQuery(QueryDefinition queryDefinition, IndexedQueryMode indexedQueryMode, KeyTransformationHandler keyTransformationHandler, TimeoutExceptionFactory timeoutExceptionFactory, ExecutorService asyncExecutor, IndexedTypeMap<CustomTypeMetadata> indexedTypeMap) {
+      if (!isIndexed) {
+         throw log.cannotRunLuceneQueriesIfNotIndexed(cache.getName());
+      }
+      if (indexedQueryMode == IndexedQueryMode.BROADCAST) {
+         return new ClusteredCacheQueryImpl<>(queryDefinition, asyncExecutor, cache, keyTransformationHandler, indexedTypeMap);
+      } else {
+         queryDefinition.initialize(cache);
+         HSQuery hsQuery = queryDefinition.getHsQuery();
+         CacheQuery cacheQuery = new CacheQueryImpl<>(hsQuery, cache, keyTransformationHandler);
```
anistor: These two line can become ```return new CacheQueryImpl<E>(hsQuery, cache, keyTransformationHandler);``` to avoid the warning.

anistor:
I'm happy with this refactoring. I can still spot some small design issues that we can fix later.

The existence of RemoteQueryDefinition and HsQueryRequest seems to be a symptom of misplaced responsibility. It all starts with QueryDefinition.initialize, which IMO should actually be placed inside QueryEngine, not QueryDefinition. Doing that refactoring will remove the need for RemoteQueryDefinition, which now exists just to differentiate between embedded and remote case, but that differentiation can be done inside the query engine itself. Also, HsQueryRequest is just a data holder that carries the return value of QueryEngine.createHsQuery. If QueryDefinition.initialize is moved to QueryEngine we would also not need this anymore.

I did not think about it in detail but maybe we would also need to make QueryDefinition mutable for QueryEngine and immutable for external parties. In that case we can extract QueryDefinition as an immutable interface (exposing getters only) and it's implementation class could have package local setters accessible to QueryEngine only. 

But let's leave those improvements for another day. I'll merge this today as it as after you have applied the last 2-3 minor changes I suggested. Thanks @gustavonalle !

gustavonalle:
@anistor Created https://issues.jboss.org/browse/ISPN-8628 to further refactor it

anistor:

anistor:

query/src/main/java/org/infinispan/query/dsl/embedded/impl/QueryEngine.java
```diff
@@ -687,10 +706,107 @@ protected IckleFilterAndConverter createFilter(String queryString, Map<String, O
...
+
+   public HsQueryRequest createHsQuery(String queryString, IndexedTypeMap<CustomTypeMetadata> metadata, Map<String, Object> nameParameters) {
+      IckleParsingResult<TypeMetadata> parsingResult = parse(queryString);
+      if (parsingResult.hasGroupingOrAggregations()) {
+         throw log.groupAggregationsNotSupported();
+      }
+      LuceneQueryParsingResult luceneParsingResult = transformParsingResult(parsingResult, nameParameters);
+      org.apache.lucene.search.Query luceneQuery = makeTypeQuery(luceneParsingResult.getQuery(), luceneParsingResult.getTargetEntityName());
+      SearchIntegrator searchFactory = getSearchFactory();
+      HSQuery hsQuery = metadata == null ? searchFactory.createHSQuery(luceneQuery) : searchFactory.createHSQuery(luceneQuery);
```
anistor: I believe here you intended to write ```HSQuery hsQuery = metadata == null ? searchFactory.createHSQuery(luceneQuery) : searchFactory.createHSQuery(luceneQuery, metadata);```

anistor: Fixed this here: https://github.com/infinispan/infinispan/commit/29e92eec857e16eae487353b89745b71891c390a

PR #5602, opened 2017-11-22
ISPN-8538 TwoWaySplitAndMerge[DIST_SYNC] deadlock in conflict resolution

ryanemerson:
Deadlock caused by StateReceiverImpl::stop synchronizing on
StateReceiverImpl.this, but then attempting to cancel a
StateRequestImpl.SegmentRequest whoms cancel() method synchs on itself.
However, a SegmentRequest can also be cancelled by another thread
when a CompletableFuture completes exceptionally. This results in
Thread-A acquiring StateReceiverImpl.this in StateReceiverImpl::stop and
calling SegmentRequest::cancel, however Thread-B has called SegmentRequest::cancel and
is now waiting to synchronize on StateReceiverImpl.this, hence deadlock.

Also, we shouldn't call StateReceiverImpl::stop in the
DefaultConflictManager.ReplicaSpliterator::tryAdvance when a
Cancellation/Execution Exception occurs as failed requests are already
cancelled on an exception in StateReceiverImpl.

https://issues.jboss.org/browse/ISPN-8538

### DIFF
--- a/core/src/main/java/org/infinispan/conflict/impl/StateReceiverImpl.java
+++ b/core/src/main/java/org/infinispan/conflict/impl/StateReceiverImpl.java
@@ -31,6 +31,8 @@ import org.infinispan.statetransfer.InboundTransferTask;
 import org.infinispan.statetransfer.StateChunk;
 import org.infinispan.topology.CacheTopology;
 
+import net.jcip.annotations.GuardedBy;
+
 /**
  * @author Ryan Emerson
  * @since 9.1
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/infinispan/conflict/impl/StateReceiverImpl.java
```diff
@@ -161,26 +161,29 @@ InboundTransferTask createTransferTask(int segmentId, Address source, CacheTopol
...
-            return Collections.unmodifiableList(retVal);
+            synchronized (StateReceiverImpl.this) {
+               clear();
+               return Collections.unmodifiableList(retVal);
+            }
          });
          return future;
       }
 
+      // Should only be called when synchronized on StateReceiverImpl.this
```
danberindei: You can use `@GuardedBy("StateReceiverImpl.this")`, although I I'm not sure it this reference is correct and I don't think I ever got IntelliJ to give me a warning about missing synchronization based on those annotations.

ryanemerson:
Failure is not related https://issues.jboss.org/browse/ISPN-8525

danberindei:
Integrated, thanks Ryan!

I wasn't at all worried about `StaleLocksWithLockOnlyTxDuringStateTransferTest`, but `CacheManagerTest` is now constantly hanging on my machine (https://issues.jboss.org/browse/ISPN-8555).

PR #5608, opened 2017-11-27
[9.1.x] ISPN-8557 RESTful DELETE and PUT should always work

tristantarrant:
- reordered the test conditions so that metadata is only considered when etags are present

https://issues.jboss.org/browse/ISPN-8557

### DIFF
--- a/server/rest/src/main/java/org/infinispan/rest/operations/CacheOperations.java
+++ b/server/rest/src/main/java/org/infinispan/rest/operations/CacheOperations.java
@@ -206,26 +206,36 @@ public class CacheOperations {
 
          if (entry instanceof InternalCacheEntry) {
             InternalCacheEntry<String, Object> ice = (InternalCacheEntry<String, Object>) entry;
-            Metadata meta = entry.getMetadata();
-            if (meta instanceof MimeMetadata) {
-               String etag = calcETAG(ice.getValue(), ((MimeMetadata) meta).contentType());
-               Optional<String> clientEtag = request.getEtagIfNoneMatch();
-               if (clientEtag.map(t -> t.equals(etag)).orElse(true)) {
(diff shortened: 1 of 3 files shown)


gustavonalle:
@tristantarrant Puzzled why we didn't have a test for that (simple) case, can you add it?

tristantarrant:
@gustavonalle added a test

server/rest/src/main/java/org/infinispan/rest/operations/CacheOperations.java
```diff
@@ -206,21 +206,23 @@ public InfinispanResponse deleteCacheValue(InfinispanRequest request) throws Res
...
-                  response.status(HttpResponseStatus.OK);
-                  if (useAsync.isPresent() && useAsync.get()) {
-                     restCacheManager.getCache(cacheName).removeAsync(key);
+            Optional<String> clientEtag = request.getEtagIfNoneMatch();
+            if (clientEtag.isPresent()) {
+               Metadata meta = entry.getMetadata();
+               if (meta instanceof MimeMetadata) {
+                  String etag = calcETAG(ice.getValue(), ((MimeMetadata) meta).contentType());
+
+                  if (clientEtag.map(t -> t.equals(etag)).orElse(true)) {
```
danberindei: We're in the `clientEtag.isPresent()` branch, so this is equivalent to `clientEtag.get().equals(etag)`

tristantarrant: Fixed

tristantarrant:
For the TO State transfer test failures I've backported https://github.com/infinispan/infinispan/pull/5622

tristantarrant:
CI has 3 unrelated failures.

PR #5611, opened 2017-11-28
ISPN-8155

wburns:
SingleClusterExecutorTest.testExecutorTriConsumerTimeoutException random
failures

* Make sure timeout service doesn't capture state transfer as well

https://issues.jboss.org/browse/ISPN-8155

### DIFF
--- a/core/src/test/java/org/infinispan/manager/AllClusterExecutorTest.java
+++ b/core/src/test/java/org/infinispan/manager/AllClusterExecutorTest.java
@@ -393,15 +393,20 @@ public class AllClusterExecutorTest extends AbstractInfinispanTest {
                      KnownComponentNames.TIMEOUT_SCHEDULE_EXECUTOR, stpe, true);
             }
 
+            // These are to make sure the call to the scheduled executor doesn't overlap with others, so these should
+            // be unique
+            long crazyNumber = 84129912895471L;
+            TimeUnit unit = TimeUnit.DAYS;
+
             CompletableFuture<Void> future =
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/infinispan/manager/AllClusterExecutorTest.java
```diff
@@ -393,15 +393,20 @@ public void call() throws InterruptedException, ExecutionException, TimeoutExcep
...
             }
 
+            // These are to make sure the call to the scheduled executor doesn't overlap with others, so these should
+            // be unique
+            long crazyNumber = 84129912895471L;
+            TimeUnit unit = TimeUnit.DAYS;
+
             CompletableFuture<Void> future =
-                  executor(cm1).timeout(10, TimeUnit.MILLISECONDS).submitConsumer(m -> {
+                  executor(cm1).timeout(crazyNumber, unit).submitConsumer(m -> {
```
karesti: If the scheduled executor is a mock at this point, why is this overlapping with other tests ?

tristantarrant:
Merged. I have not seen any more failures locally (and it was failing very often)

core/src/test/java/org/infinispan/manager/AllClusterExecutorTest.java
wburns: It isn't other tests. The problem is that we started adding other internal caches and those use state transfer and will call into the mock. That is why this wasn't failing when I fixed this before :)

karesti: oh ok ! aniway, tristan has already merged 👍

PR #5605, opened 2017-11-24
ISPN-8001 HotRodCustomMarshallerIteratorIT fails randomly

anistor:
Fixing this was not easy... the principle of operation is clearly explained here:

https://www.youtube.com/watch?v=dg5xeesapQQ

https://issues.jboss.org/browse/ISPN-8001

PR #5604, opened 2017-11-24
ISPN-8001 HotRodCustomMarshallerIteratorIT fails randomly

anistor:
Fixing this was not easy... the principle of operation is clearly explained here: 

https://www.youtube.com/watch?v=dg5xeesapQQ

https://issues.jboss.org/browse/ISPN-8001

PR #5615, opened 2017-11-29
Undeploy jars properly after test suite termination

anistor:


gustavonalle:
We should add the same undeploy logic for \*CustomMarshaller\*IT

danberindei:
@anistor couldn't Arquillian delete the jars?

anistor:
I initially started cleaning up stuff that was NOT deployed using Arquillian. These obviously needed explicit cleaning. Then I continued with some Arquillian managed stuff. Seems it is not cleaned up, and not sure if it should be expected to be cleaned up by design.

anistor:
Closing this. These commits are now included in https://github.com/infinispan/infinispan/pull/5604 ,  https://github.com/infinispan/infinispan/pull/5605 and https://github.com/infinispan/infinispan/pull/5623

PR #5621, opened 2017-12-01
ISPN-8578 Do not log exceptions to the console unless they cause a te…

tristantarrant:
…st failure

https://issues.jboss.org/browse/ISPN-8578


tristantarrant:
This reduces the size of the console log to ~13MB instead of  > 20MB

ryanemerson:
Merged, thanks @tristantarrant

PR #5623, opened 2017-12-01
[90x] Server deployment cleanup 

anistor:
Remove those deployed jars after test suite completion.


tristantarrant:
I think we can safely ignore 9.0.x.
Aside from branches required for productization, we only support 9.1.x

anistor:
I made this PR to show the test works on 90x and was later broken on 91x. We do not need this any more.

PR #5624, opened 2017-12-05
JCache 1.1 compliance

galderz:
This PR won't pass CI because it relies on manually building JCache 1.1 artifacts locally. The spec leads do not release any artifacts until the spec is released :\

It includes:
* https://issues.jboss.org/browse/ISPN-8571
* https://issues.jboss.org/browse/ISPN-8579
* https://issues.jboss.org/browse/ISPN-8585


tristantarrant:
Maybe there is a way for us to build jcache SNAPSHOTS and push them to JBoss' Nexus

bom/pom.xml
```diff
@@ -86,7 +86,9 @@
       <version.hibernate.search>5.8.1.Final</version.hibernate.search>
       <version.hikaricp>2.4.6</version.hikaricp>
       <version.java>1.8</version.java>
-      <version.javax.cache>1.0.0</version.javax.cache>
+      <version.javax.cache>1.1.0-SNAPSHOT</version.javax.cache>
+      <version.cdi>1.2</version.cdi>
+      <version.jboss.marshalling>2.0.0.Beta3</version.jboss.marshalling>
```
tristantarrant: Redefining JBoss Marshalling ? It is defined as 2.0.2.Final a couple of lines below

bom/pom.xml
```diff
@@ -86,7 +86,9 @@
       <version.hibernate.search>5.8.1.Final</version.hibernate.search>
       <version.hikaricp>2.4.6</version.hikaricp>
       <version.java>1.8</version.java>
-      <version.javax.cache>1.0.0</version.javax.cache>
+      <version.javax.cache>1.1.0-SNAPSHOT</version.javax.cache>
+      <version.cdi>1.2</version.cdi>
```
tristantarrant: CDI is already defined at 1.2 a few lines above.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/ClientListenerNotifier.java
```diff
@@ -385,6 +386,10 @@ private ClientListenerInvocation(Object listener, Method method) {
       public void invoke(ClientEvent event) {
          try {
             method.invoke(listener, event);
+         } catch (InvocationTargetException e) {
+            log.errorInvokingListener(method, listener, e.getTargetException());
+            throw log.exceptionInvokingListener(
```
tristantarrant: Isn't it enough to throw the logging exception instead of also logging locally ?

core/src/main/java/org/infinispan/commands/write/RemoveCommand.java
```diff
@@ -103,7 +103,8 @@ public Object perform(InvocationContext ctx) throws Throwable {
    }
 
    public void notify(InvocationContext ctx, Object removedValue, Metadata removedMetadata, boolean isPre) {
-      notifier.notifyCacheEntryRemoved(key, removedValue, removedMetadata, isPre, ctx, this);
+      if (removedValue != null)
```
tristantarrant: @wburns You should see this

jcache/tck-runner/pom.xml
```diff
@@ -150,23 +150,23 @@
...
-                           <groupId>javax.cache</groupId>
-                           <artifactId>app-domain</artifactId>
-                           <version>${version.javax.cache.cache-tests}</version>
-                           <outputDirectory>${domain-lib-dir}</outputDirectory>
-                           <destFileName>${domain-jar}</destFileName>
-                        </artifactItem>
-                     </artifactItems>
-                  </configuration>
-               </execution>
+               <!--<execution>-->
```
tristantarrant: Removing is better than commenting. There is always git history

core/src/main/java/org/infinispan/commands/write/RemoveCommand.java
wburns: That is fine with me. Unfortunately it has been this way forever, and I have always wanted to change it, just never had a specific reason at the time.

wburns:
Seems fine to me, but maybe you want to squash some of those commits?

galderz:
Squashed and comments addressed. Note that we can't yet integrate this.

rvansa:
@galderz If you need an unpublished dependency, does it mean that you can't `git clone && mvn install` anymore? Or is this isolated in a non-default profile?

galderz:
> @galderz If you need an unpublished dependency, does it mean that you can't git clone && mvn install anymore? That would be a no-go for me. Could this be isolated into a non-default profile?

It's marked as PREVIEW. 

Can't be integrated till the powers-to-be allow us to publish snapshots or they publish artifacts.

galderz:
@rvansa I've added a `DO NOT INTEGRATE` label to it ;)

galderz:
JCache 1.1.0 has been approved and the artifacts for the API are [available](http://repo1.maven.org/maven2/javax/cache/cache-tests/)... However, TCK artifacts do not appear to be there yet.

tristantarrant:
The TCK is now there: http://repo1.maven.org/maven2/javax/cache/cache-tests/1.1.0/

jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheListenerAdapter.java
```diff
@@ -37,14 +37,14 @@ public void handleCacheEntryCreatedEvent(CacheEntryCreatedEvent<K, V> e) {
    public void handleCacheEntryModifiedEvent(CacheEntryModifiedEvent<K, V> e) {
       // JCache listeners notified only once, so do it after the event
       if (!e.isPre() && !e.isCreated())
-         notifier.notifyEntryUpdated(jcache, e.getKey(), e.getValue());
+         notifier.notifyEntryUpdated(jcache, e.getKey(), e.getValue(), e.getValue());
```
rvansa: Using current value with `isPre == false` is not correct, could you log a JIRA and add a reference here? I wonder why TCK did miss this.

jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheListenerAdapter.java
```diff
@@ -37,14 +37,14 @@ public void handleCacheEntryCreatedEvent(CacheEntryCreatedEvent<K, V> e) {
...
    }
 
    @CacheEntryRemoved
    public void handleCacheEntryRemovedEvent(CacheEntryRemovedEvent<K, V> e) {
       // JCache listeners notified only once, so do it after the event
       if (!e.isPre())
-         notifier.notifyEntryRemoved(jcache, e.getKey(), e.getOldValue());
+         notifier.notifyEntryRemoved(jcache, e.getKey(), e.getOldValue(), e.getOldValue());
```
rvansa: Current value should be `null`, shouldn't it?

rvansa:
@galderz @tristantarrant I took the liberty of updating the maven dependency and removed the labels.

galderz:
> @galderz @tristantarrant I took the liberty of updating the maven dependency and removed the labels.

I was not aware that you could add a commit of yours to someone else's branch. How did you do that? Can you simply commit on my branch and push it? :?

rvansa:
@galderz Yes, I got superpowers as a xmas present :)

Kidding aside, when you file a PR there's a small checkbox 'Allow maintainers to push commits', which is on by default.

galderz:
> @galderz Yes, I got superpowers as a xmas present :)

We always thought you were superhuman anyway ;)

> Kidding aside, when you file a PR there's a small checkbox 'Allow maintainers to push commits', which is on by default.

Nice tip!! 😎

jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheListenerAdapter.java
galderz: > I wonder why TCK did miss this.

Because the TCK only checks for a non-null value ;)

> Using current value with isPre == false is not correct.

Indeed, sending the current value as previous is not right. With the current embedded event API, getting the previous value would require handling isPre=true and somehow caching the value which would be messy. So, I guess you're referring to having a JIRA that adds a method to the event to retrieve the previous value independent of whether isPre or not?

rvansa: > Because the TCK only checks for a non-null value ;)
I guess we should file a PR to TCK...

> adds a method to the event to retrieve the previous value
Let's file a JIRA for flaw in our JCache impl ATM. You are describing a solution (probably the only viable). Caching the pre event is a no-go IMO.

jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheListenerAdapter.java
galderz: No, it's correct as is. As per the CacheEntryEvent javadoc for getValue():

    * Returns the same value as {@link #getOldValue()} for
    * {@link CacheEntryExpiredListener} and {@link CacheEntryRemovedListener}.

jcache/embedded/src/main/java/org/infinispan/jcache/embedded/JCacheListenerAdapter.java
galderz: > I guess we should file a PR to TCK...

I've submitted [an issue](https://github.com/jsr107/jsr107tck/issues/139) for now.

> Caching the pre event is a no-go IMO.

+1

galderz: > Let's file a JIRA for flaw in our JCache impl ATM

[ISPN-8652](https://issues.jboss.org/browse/ISPN-8652)

jcache/commons/src/main/java/org/infinispan/jcache/AbstractJCache.java
```diff
@@ -424,4 +424,13 @@ protected void removeCacheEntryListenerConfiguration(CacheEntryListenerConfigura
    protected boolean statisticsEnabled() {
       return getConfiguration(CompleteConfiguration.class).isStatisticsEnabled();
    }
+
+   @Override
+   public void close() {
+      notifier.close();
```
slaskawi: I believe notifier should also be `Closeable`.

jcache/commons/src/main/java/org/infinispan/jcache/Closeables.java
```diff
@@ -0,0 +1,40 @@
...
+import org.infinispan.jcache.logging.Log;
+
+import java.io.Closeable;
+import java.util.Set;
+
+class Closeables {
+
+   private static final Log log = LogFactory.getLog(Closeables.class, Log.class);
+
+   Closeables() {
```
slaskawi: You probably meant `private Closeables()` right? There is no need to instantiate this class. TBH, I think we could even get rid of this constructor, the class has default access after all...

jcache/commons/src/main/java/org/infinispan/jcache/Closeables.java
```diff
@@ -0,0 +1,40 @@
...
+   }
+
+   static void close(Object o) {
```
slaskawi: Out of curiosity - why this method doesn't take `Closeable` as parameter but `Object` and checks its class? If you want to make this very generic, you could take an object and check if it contains `close` method rather than doing type check.

jcache/commons/src/main/java/org/infinispan/jcache/logging/Log.java
```diff
@@ -101,6 +102,13 @@
    @Message(value = "The server management operation failed.", id = 21028)
    CacheException serverManagementOperationFailed(@Cause Throwable t);
 
+   @Message(value = "Cache manager is closed.", id = 21030)
```
slaskawi: How about `Cache Manager is already closed` (with already)?

galderz:
@rvansa I've pushed some commits to resolve the issues agreed.
@slaskawi I'll look at your comments next.

jcache/commons/src/main/java/org/infinispan/jcache/Closeables.java
galderz: Private constructor is what I was aiming for.

jcache/commons/src/main/java/org/infinispan/jcache/Closeables.java
galderz: If you look at how it's used, it's invoked on `ExpiryPolicy`, `CacheLoader` and `CacheWriter` which none of those extend `Closeable`, but the TCK expects that if implementations extend `Closeable`, they need to be closed. I think check for the `close` method would be a bit over the top of what's required here.

galderz:
@slaskawi I've pushed some commits to address agreed comments.

When you're all happy with this I'll squash the last few commits.

galderz:
I've squashed commits. Let's integrate this baby!

slaskawi:
Wow [16 failures](http://ci.infinispan.org/job/Infinispan/job/PR-5624/11/)?? Let me trigger this again...

galderz:
> Wow 16 failures?? Let me trigger this again...

None related to JCache

galderz:
Yeah, looks fine to me.

rvansa:
@galderz None of them related to JCache, but you have a (trivial) modification in server and core, too. Is there any other PR with the same failures?

tristantarrant:
The failures in the cache admin api might be connected to the listener changes. Let me verify

tristantarrant:
Running locally, I don't see those failures. Requeuing a build on CI

tristantarrant:
@galder, the Hot Rod failures are caused by this PR

galderz:
@tristantarrant Ok, I'll double check

galderz:
@tristantarrant Should be fixed now

rvansa:
CI looks almost happy, integrated. Thanks Galder.

PR #5626, opened 2017-12-07
ISPN-6879 Calculate (and expose) minimum number of nodes for data in

wburns:
Infinispan

* Added min num nodes to Stats classes

https://issues.jboss.org/browse/ISPN-6879


core/src/main/java/org/infinispan/interceptors/impl/CacheMgmtInterceptor.java
```diff
@@ -400,6 +404,52 @@ public long getOffHeapMemoryUsed() {
...
+      CacheTopology cacheTopology = cache.getDistributionManager().getCacheTopology();
+      if (mode.isInvalidation()) {
+         // Invalidation requires all as we don't know what data is installed on which
+         return cacheTopology.getMembers().size();
+      }
+      int numMembers = cacheTopology.getMembers().size();
+      int numOwners = clusteringConfiguration.hash().numOwners();
+      int minNodes;
+      if (mode.isScattered()) {
+         minNodes = numMembers - 1;
```
tristantarrant: Why are you special-casing scattered mode, which is a dist mode with 2 owners and would therefore be satisfied by the DIST branch below.

tristantarrant:
We want this exposed in the server DMR too:
https://github.com/infinispan/infinispan/blob/0446879eac18d68134e1f0c32062f361f3e75576/server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/ClusteredCacheMetricsHandler.java#L52

core/src/main/java/org/infinispan/interceptors/impl/CacheMgmtInterceptor.java
```diff
@@ -400,6 +404,52 @@ public long getOffHeapMemoryUsed() {
...
+      ClusteringConfiguration clusteringConfiguration = config.clustering();
+      CacheMode mode = clusteringConfiguration.cacheMode();
+      if (mode.isReplicated() || !mode.isClustered()) {
+         // Local and replicated only require the 1 node to keep the data
+         return 1;
+      }
+      CacheTopology cacheTopology = cache.getDistributionManager().getCacheTopology();
+      if (mode.isInvalidation()) {
+         // Invalidation requires all as we don't know what data is installed on which
+         return cacheTopology.getMembers().size();
```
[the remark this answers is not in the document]
wburns: What would you want to return though? It seemed simple enough to just say "Don't lose any nodes" :D

core/src/main/java/org/infinispan/interceptors/impl/CacheMgmtInterceptor.java
wburns: Scattered afaik doesn't use num owners afaik, so someone could put any value they wanted. So I just hard coded this to assume 2 in that case.

wburns:
Sure @tristantarrant. I can add it there as well. Sorry I was talking with @slaskawi and was more hopeful that we wouldn't be using DMR anymore :)

tristantarrant:
For what @slaskawi needs, JMX is sufficient, but we still need to think about wider use-cases. In particular the console obtains information through the DMR

core/src/main/java/org/infinispan/interceptors/impl/CacheMgmtInterceptor.java
wburns: Should we say 0? Also what about local then as well? In either case it could be in a state where it would say "don't ever shut down a node".

core/src/main/java/org/infinispan/interceptors/impl/CacheMgmtInterceptor.java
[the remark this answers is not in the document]
wburns: Should we say 0? Also what about local then as well? In either case it could be in a state where it would say "don't ever shut down a node".

wburns:
Updated with DMR changes. Let me squash before merge if it is alright though.

tristantarrant:
[domain@localhost:9990 /] cd /host=master/server=server-one/subsystem=datagrid-infinispan/cache-container=clustered/distributed-cache=default
[domain@localhost:9990 distributed-cache=default] ls
{"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:" => {"Operation step-2" => "WFLYCTL0158: Operation handler failed: java.util.MissingResourceException: Can't find resource for bundle java.util.PropertyResourceBundle, key datagrid-infinispan.clustered-cache.clusterwide-minimum-required-nodes"}}

wburns:
Yeah it was a copy pasta issue :(

https://github.com/infinispan/infinispan/pull/5626/files#diff-49a1065a2df11ed0b87768bb20af3300R250

wburns:
Fixed:

>  Running org.jboss.as.clustering.infinispan.subsystem.SubsystemParsingTestCase
>  Tests run: 70, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 10.914 sec - in 
> org.jboss.as.clustering.infinispan.subsystem.SubsystemParsingTestCase
>
>  Results :
>
>  Tests run: 70, Failures: 0, Errors: 0, Skipped: 0

alanfx:
Hmm, pasta! 🍝

wburns:
I loves me some pasta too :)

PR #5628, opened 2017-12-07
ISPN-8589 Cache @Stop methods should be invoked in privileged block

wburns:
https://issues.jboss.org/browse/ISPN-8589


tristantarrant:
Merged, thanks

PR #5631, opened 2017-12-08
ISPN-8604 Upgrade to JGroups 4.0.9.Final

danberindei:
https://issues.jboss.org/browse/ISPN-8604

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -189,33 +189,56 @@
       <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
       <testNGListeners>org.infinispan.commons.test.TestNGTestListener,org.infinispan.commons.test.PolarionJUnitXMLReporter</testNGListeners>
       <uberjar.deps.optional>true</uberjar.deps.optional>
-      <!-- Versions for dependencies -->
-      <version.ant>1.10.1</version.ant>
-      <version.ant-contrib>1.0b3</version.ant-contrib>
+
+      <!-- Java source/target version -->
+      <version.java>1.8</version.java>
+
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -248,7 +248,8 @@
       <version.jcipannotations>1.0</version.jcipannotations>
       <version.jetty.http.client>9.4.5.v20170502</version.jetty.http.client>
       <version.jgoodies.forms>1.0.7</version.jgoodies.forms>
-      <version.jgroups>4.0.8.Final</version.jgroups>
+      <!-- Make sure you also update the BOM -->
```
ryanemerson: I'm not sure this comment is necessary as a dependencies version should also be updated in both the bom and parent if they are located in both. Maybe we need to add a comment explaining this above all of the version properties in the pom.

tristantarrant: The only reason to redeclare it in the parent is if we need to use the property aside from the dependecy declaration, otherwise we should avoid it.

danberindei: @ryanemerson I only added the comment because I was surprised to see some properties duplicated in the parent and in the BOM, I don't remember having to modify the version in multiple places before.

ryanemerson: @danberindei before it wasn't necessary as properties were inherited from the bom, however we now import the bom which does not automatically import properties hence we have the duplication. In the case of jgroups, it is necessary as we require the jgroups version [here](https://github.com/infinispan/infinispan/blob/master/all/embedded/pom.xml#L331)

danberindei:
Updated @ryanemerson @tristantarrant!

PR #5636, opened 2017-12-11
ISPN-8319 Expose a constructor specifying a raw Caffeine weigher to subclasses.

pferraro:
For use by WildFly to override DefaultDataContainer behavior.

Please cherry-pick into 9.1.x.

### DIFF
--- a/core/src/main/java/org/infinispan/container/DefaultDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/DefaultDataContainer.java
@@ -49,6 +49,7 @@ import com.github.benmanes.caffeine.cache.CacheWriter;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.Policy;
 import com.github.benmanes.caffeine.cache.RemovalCause;
+import com.github.benmanes.caffeine.cache.Weigher;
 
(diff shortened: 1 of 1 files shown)


wburns:
I wonder would it not be sufficient if we instead just allowed a `EntrySizeCalculator<? super K, ? super InternalCacheEntry<K, V>>` argument constructor? I am thinking this should solve your issue as well right? And it wouldn't expose you to caffeine either. I am finding I need to do some refactoring for ISPN-8396 to allow for size calculation to be available from multiple classes, so I was thinking of doing this change.

pferraro:
@wburns Which interface we use doesn't really matter to me, i.e. if I'm extending DefaultDataContainer, Caffeine is already on my classpath.  Using the the EntrySizeCalculator interface is fine too - the goal was to expose a constructor that won't wrap the parameter with a CacheEntrySizeCalculator.

rvansa:
@wburns What's the state of this PR? Will you close that if you already propose another way?

wburns:
I think we can close it for now. The other PR I am going to introduce should handle this, but I will keep this in the back of my mind just in case.

pferraro:
@wburns That's fine, so long as there is a solution for 9.1.x as well.

PR #5638, opened 2017-12-12
ISPN-8570 Produce Hibernate 5.1 and 5.2 compatible 2LC providers

galderz:
https://issues.jboss.org/browse/ISPN-8570

The cache providers have been split into two modules:
* `org.infinispan:infinispan-hibernate-cache` for 5.2
* `org.infinispan:infinispan-hibernate-cache-v51` for 5.1

The cache-commons module keeps shared production and test code that's compatible with either version.

The work relies heavily on using ServiceLoader pattern to load version specific classes. This was much easier to implement than inheritance based approaches.

There are 3 big compatibility issues that drove the code changes:
* The switch from SessionImplementor (5.1) to SharedSessionContractImplementor (5.2) in Hibernate Region APIs.
* Refactoring of Hibernate transaction classes from `org.hibernate.resource.transaction` to `org.hibernate.resource.transaction.spi`
* Refactoring of Hibernate Query classes from `org.hibernate` to `org.hibernate.query`.


scottmarlow:
@galderz @Sanne @emmanuelbernard @sebersole will org.infinispan:infinispan-hibernate-cache also cover other ORM versions (e.g. 5.3, 6.0, 7.0, ...)?

Sanne:
@galderz thanks!

@scottmarlow we'll upgrade the targeted versions as necessary. For example I expect we'll soon target ORM 5.3 and drop 5.2, while maintaining the one compatible for ORM 5.1 a bit longer (as it's in WildFly).
Would be good to release at least one stable version before switching though, so that people using ORM 5.2 have an option.

galderz:
> @scottmarlow we'll upgrade the targeted versions as necessary. For example I expect we'll soon target ORM 5.3 and drop 5.2, while maintaining the one compatible for ORM 5.1 a bit longer (as it's in WildFly). Would be good to release at least one stable version before switching though, so that people using ORM 5.2 have an option.

So I guess the plan is: 
* Infinispan 9.2 would have Hibernate 5.2 and 5.1 providers.
* Infinispan 9.3 we'd move up to Hibernate 5.3 and 5.1 providers.

scottmarlow:
@galderz @Sanne thanks for the quick response!

I think that a tricky part will be knowing when to sync to Hibernate 6.0 and 5.1 providers, which is a bit in the future, so not a concern today but soon will be.  @Sanne proposed (WildFly) being able to sync with three ORM versions, perhaps Infinispan should consider that also when 6.0/7.0/later versions come into play.

We actually have been integrating with three Hibernate ORM versions in WildFly (when it makes sense), we just currently only include the Hibernate ORM jars for one version.

@pferraro any concerns from you, about potentially supporting WildFly second level caching with Hibernate 5.1 + 5.3 (JPA 2.2) + 6.0 (maybe next year)?  The idea is that the WildFly-Hibernate integration code ([JipiJapai impl](https://github.com/wildfly/wildfly/tree/master/jpa)) will extend this new Infinispan module that replaces the current (Hibernate ORM) hibernate-infinispan code.  I think that Hibernate may provide the Hibernate 5.3 implementation of https://github.com/wildfly/wildfly/tree/master/jpa.

sebersole:
So it's reasonable from your perspective to drop hibernate-infinispan in
5.3?

pferraro:
@scottmarlow No concerns from me.  This is already integrated as part of my Infinispan 9.1.x upgrade branch (currently sitting at the end of a long PR dependency queue). See: https://github.com/pferraro/wildfly/tree/infinispan-9.1

sebersole:
No concerns about dropping hibernate-infinispan as part of Hibernate 5.3?

pferraro:
@sebersole My integration branch already removes references to hibernate-infinispan (for 5.1, at least).  So, I don't have any concerns if this is dropped as part of Hibernate 5.3.

galderz:
Neither do I @sebersole

hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/entity/EntityRegionExtraAPITest.java
```diff
@@ -29,13 +30,14 @@
...
-		return environment.getEntityRegion( REGION_NAME, CACHE_DATA_DESCRIPTION).buildAccessStrategy( accessType );
+		return ((EntityRegion) environment.getEntityRegion( REGION_NAME, CACHE_DATA_DESCRIPTION))
+         .buildAccessStrategy( accessType );
 	}
 
 	@Test
 	@SuppressWarnings( {"UnnecessaryBoxing"})
 	public void testAfterInsert() {
-		boolean retval = accessStrategy.afterInsert(SESSION,	KEY, VALUE1, Integer.valueOf( 1 ));
+		boolean retval = testAccessStrategy.afterInsert(SESSION,	KEY, VALUE1, Integer.valueOf( 1 ));
```
rvansa: extra tabs?

hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/AbstractNonInvalidationTest.java
```diff
@@ -166,6 +168,10 @@ public void cleanup() throws Exception {
             log.info("Exception thrown: ", e);
             markRollbackOnly(s);
             return false;
+         } catch (OptimisticLockException e) {
```
rvansa: You could condense clause for all exceptions using `|`

hibernate/cache-v51/src/main/resources/META-INF/services/org.infinispan.hibernate.cache.commons.InternalRegionFactory
```diff
@@ -0,0 +1 @@
+org.infinispan.hibernate.cache.v51.InternalRegionFactoryImpl
```
rvansa: Could you use `@MetaInfServices` instead? (for all of the services files)

hibernate/cache-commons/src/main/java/org/infinispan/hibernate/cache/commons/access/AccessDelegate.java
```diff
@@ -20,7 +20,7 @@
  * @author Radim Vansa &lt;rvansa@redhat.com&gt;
  */
 public interface AccessDelegate {
-	Object get(SessionImplementor session, Object key, long txTimestamp) throws CacheException;
+	Object get(Object session, Object key, long txTimestamp) throws CacheException;
```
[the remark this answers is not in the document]
galderz: I did vaguely when I started with this, but given the niche use case for these changes (dealing with Hibernate's compatibility issues), I didn't think it was so important or worth the effort to spend time fighting generics ;)

rvansa:
@galderz There're couple of outstanding nitpicks, since CI is happy, please fix or dismiss them and remove the preview label.

galderz:
@rvansa Yeah, I'll fix those later today or tomorrow.

galderz:
@rvansa Just pushed a few commits to address the remaining comments above.

If you're happy I'll squash commits and we're ready to go.

rvansa:
@galderz Go for it.

PR #5646, opened 2017-12-15
ISPN-8619 Rewrite Hot Rod client using Netty

rvansa:
Preview at this moment since I need to rebase this on top of cluster counters.


galderz:
I certainly welcome this. I'll look at details tomorrow or after the break.

Is this compatible with Sebastian's plans around the Hot Rod client? He had planned to rewrite the client to do Hot Rod over HTTP 2.0. Deficiencies of current client had been noted [here](https://docs.google.com/document/d/1y-eLiZ6657qevHYlXXPI9VeY0sN2XIg-ykea9ZUVo4Q/edit#heading=h.tenj92huu8a1).

This should also be reviewed by @karesti and a member of the Vert.x team so that the client follow their required guidelines. Just a week ago I was having a discussion about this with @cescoffier.

tristantarrant:
@galderz the plan is to implement a single-port HTTP/1.1 server which can be "upgraded" to HTTP/2 or Hot Rod simply via an HTTP Upgrade request: https://en.wikipedia.org/wiki/HTTP/1.1_Upgrade_header
This would just mean that the Hot Rod client needs to recognize whether it is talking to an HTTP endpoint or a native one and send the Upgrade header in the first case.
Eventually, when we design Hot Rod 3, we can consider an alternative transport for it

rvansa:
@galderz While I have not consulted HTTP 2 with @slaskawi I don't see any issues here. Netty here is just a vehicle and I am using the 'basic API'; no fancy use of decoders etc. Netty has some support for HTTP 2 and I assume that changes during negotiation will affect the 'activation' part - see `o.i.c.h.impl.transport.netty.ChannelInitializer`. My plans to use Netty for async communication are mentioned in the 'deficiencies' document you've linked.

rvansa:
I've rewritten multimap and counter operations (counter listeners required non-trivial changes, though), so removing the 'Preview' label.

rvansa:
Rebased on top of #5654 and fixed OSGi issues.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/retry/DistributionRetryTest.java
```diff
@@ -114,14 +116,14 @@ private Object generateKeyAndShutdownServer() throws IOException, ClassNotFoundE
 
       remoteCache.put(key, "v");
       assertOnlyServerHit(getAddress(hotRodServer2));
-      TcpTransportFactory tcpTp = (TcpTransportFactory) ((InternalRemoteCacheManager) remoteCacheManager).getTransportFactory();
+      ChannelFactory channelFactory = ((InternalRemoteCacheManager) remoteCacheManager).getChannelFactory();
```
slaskawi: Looking how `ChannelFactory` is being used I think it may be beneficial to call it Netty (or Channel if you wish) Transport Factory.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/TransportFactory.java
```diff
@@ -1,95 +0,0 @@
...
-import org.infinispan.client.hotrod.impl.transport.tcp.TcpTransportFactory.ClusterSwitchStatus;
-import org.infinispan.commons.marshall.Marshaller;
-
-/**
- * Transport factory for building and managing {@link org.infinispan.client.hotrod.impl.transport.Transport} objects.
- *
- * @author Mircea.Markus@jboss.com
- * @since 4.1
- */
-public interface TransportFactory {
```
slaskawi: TBH, I'm not sure what to think about this. On one hand this class resides in `impl` package but on the other it is a sort of contract and removing it might break backward compatibility. It is especially important because we expose a property which allows to swap implementations.

My personal vote would be to keep this class but I wonder what others think about it.

//cc @tristantarrant

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
```diff
@@ -0,0 +1,666 @@
...
+         lock.writeLock().unlock();
+      }
+      pingServersIgnoreException();
+   }
+
+   private ChannelPool newPool(SocketAddress address) {
+      log.debugf("Creating new channel pool for %s", address);
+      Bootstrap bootstrap = new Bootstrap()
+            .group(eventLoopGroup)
+            .channel(NioSocketChannel.class)
```
slaskawi: We could probably use Epoll here (if it is supported). See how this is done in Hot Rod Server.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelInitializer.java
```diff
@@ -0,0 +1,154 @@
...
+               ssl.keyStorePassword(),
+               ssl.keyStoreCertificatePassword(),
+               ssl.keyAlias(),
+               ssl.trustStoreFileName(),
+               ssl.trustStoreType(),
+               ssl.trustStorePassword(),
+               ssl.protocol(),
+               configuration.classLoader());
+      }
+      SslContext nettySslContext = new JdkSslContext(jdkSslContext, true, ClientAuth.NONE);
```
slaskawi: I think there was a method called `SslContext#newClientContext` in Netty. Could we use it here?

slaskawi:
@galderz @tristantarrant @rvansa Hot Rod over HTTP/2 will be handled via Single Port (see [this design](https://github.com/infinispan/infinispan-designs/blob/master/Single_port.adoc)). The idea is to use either `TLS/ALPN` negotiation or `HTTP/1.1 Upgrade` header and switch existing TCP connection from REST to Hot Rod. As the matter of fact I would like to start looking into that next week.

Once the Single Port is done, we could start thinking about multi-protocol client implementation. Such an implementation could use pluggable transport and switch between them. The initial communication could always be sent using REST (`HTTP/1.1` or `HTTP/2+TLS/ALPN` - we support both currently) and then we could switch (if that's possible) to faster protocol. But.... that's the future... and Netty-based Hot Rod Client is a first step in that direction.

galderz:
Thanks @slaskawi @rvansa for clarifications. It's good that this PR is purely focused on changing the inner workings and keep the API as is. One step at the time :)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/TransportFactory.java
rvansa: @slaskawi so you suggest keeping just the interface around, though not using it anywhere? I wouldn't have a problem with that. Though, I think that a completely custom implementation is extremely unlikely. There could be a tweak impl inheriting from TcpTransportFactory but we'd broke that anyway.

Btw., check out https://issues.jboss.org/browse/ISPN-8528.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
rvansa: Yes, I've considered this but committed the smallest possible set of changes. Would you mind adding that as separate PR, or should I include that in this one?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelInitializer.java
rvansa: Probably yes, I've checked `SslUtils` in server and put the bits here. Should we use `SslContextBuilder` when we tie this to JDK impl anyway? (the method you mention is deprecated)

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/retry/DistributionRetryTest.java
rvansa: It's not creating any `Transport` instances, therefore Transport Factory does not match. NettyChannelFactory would be redundant.

tools/src/test/resources/6.0.xml
```diff
@@ -439,7 +439,7 @@
         <persistence passivation="false">
             <remoteStore xmlns="urn:infinispan:config:remote:6.0" fetchPersistentState="false" shared="true" preload="false" ignoreModifications="false"
                          purgeOnStartup="false" tcpNoDelay="true" pingOnStartup="true" balancingStrategy="org.infinispan.client.hotrod.impl.transport.tcp.RoundRobinBalancingStrategy"
-                         transportFactory="org.infinispan.client.hotrod.impl.transport.tcp.TcpTransportFactory" keySizeEstimate="32" valueSizeEstimate="512"
+                         transportFactory="org.infinispan.client.hotrod.impl.transport.netty.ChannelFactory" keySizeEstimate="32" valueSizeEstimate="512"
```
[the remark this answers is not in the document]
rvansa: Probably not. Let's see if that breaks any tests in CI.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/counter/operation/CompareAndSwapOperation.java
```diff
@@ -28,25 +29,28 @@
...
-   protected Long executeOperation(Transport transport) {
-      HeaderParams header = writeHeaderAndCounterName(transport, COUNTER_CAS_REQUEST);
-      transport.writeLong(expect);
-      transport.writeLong(update);
-      transport.flush();
+   protected void executeOperation(Channel channel) {
+      ByteBuf buf = getHeaderAndCounterNameBufferAndRead(channel, COUNTER_CAS_REQUEST, 16);
+      buf.writeLong(expect);
+      buf.writeLong(update);
+      channel.writeAndFlush(buf);
```
galderz: AFAIK, all calls to channel.write need to be done after checking if the channel is writable. If the server is overloaded, the channel might not be writable in which case the client needs to queue up the operations. By overriding `channelWritabilityChanged`, you get a notification when the channel is writable again and you can send the operations.

We had to deal with this when implementing remote events on the server. We had to add back-pressure to the server so that it would not overload the clients. I think the same would be required here, we don't want the clients overloading the server.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelPool.java
```diff
@@ -0,0 +1,273 @@
...
+import org.infinispan.client.hotrod.logging.LogFactory;
+
+import io.netty.channel.Channel;
+import io.netty.util.concurrent.EventExecutor;
+import io.netty.util.concurrent.ScheduledFuture;
+import io.netty.util.internal.PlatformDependent;
+
+/**
+ * This is a custom implementation of {@link io.netty.channel.Channel} pooling.
+ * Compared to {@link io.netty.channel.pool.ChannelPool} implementations in Netty it does not enforce
```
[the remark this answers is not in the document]
rvansa: To simplify synchronization, Netty enforces most operations on the channel to be executed in the event loop thread. It gets easier with that.
Writing to a channel from other threads is legal and safe, but through debugging I've found that it ends up in event loop thread after passing through pipeline, so the effect of this pool is partially cancelled (there's should be stil fewer allocations).

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/HintedReplayingDecoder.java
```diff
@@ -0,0 +1,139 @@
...
+import io.netty.handler.codec.DecoderException;
+import io.netty.util.Signal;
+
+/**
+ * Copy-paste of {@link io.netty.handler.codec.ReplayingDecoder} which is hinted to not attempt decoding unless enough
+ * bytes are read.
+ *
+ * The decoder does not expect pass different message up the pipeline, this is a terminal read operation.
+ */
+public abstract class HintedReplayingDecoder<S> extends ByteToMessageDecoder {
```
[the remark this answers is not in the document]
rvansa: I've used the state change wherever possible/appropriate, e.g. when multiple values are sent, these are unmarshalled and put into collection, updating the checkpoint. With big values, you need to have them in the buffer for one moment anyway. The hinted decoder just optimizes out some unmarshalling attempts that are deemed to fail.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/counter/operation/CompareAndSwapOperation.java
rvansa: Good point, I didn't know that. Since I have to apply backpressure to the application thread (which is running this code with the custom channel pool), is it possible to block on the `writeAndFlush` when the channel is not writable?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/HintedReplayingDecoder.java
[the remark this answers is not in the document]
wburns: Yeah I have to admit, this is odd why we have our own impl of `ReplayingDecoder` in this case. Why doesn't this just extend `ReplayingDecoder` from netty? It seems to have a lot of the same copied code.

The reason I no longer used `ReplayingDecoder` is solely due to performance. The netty team rewrote the HTTP decoder in this way and saw 177% throughput increase for decoding [1].

`ReplayingDecoder` is by far a lot easier to use that doing it by only extending `ByteToMessageDecoder`. This code is actually doing the former though. If we do change it to the latter we have to unfortunately check bounds at each point you read from the bytes. Fortunately it is easy to install an additional frame decoder that only allows 1 byte down at a time [2], which  finds a lot of issues in tests. Disregard the mentions of 2 :) - missed updating those before committing.

[1] https://github.com/netty/netty/commit/fda8808210c4f4f092a93da5929df43062c3aa6c#diff-bf7a179a319b29e0262d5badaf3770e0
[2] https://github.com/infinispan/infinispan/blob/master/server/hotrod/src/test/java/org/infinispan/server/hotrod/transport/SingleByteFrameDecoderChannelInitializer.java#L20

rvansa:
@slaskawi About the performance, I am not sure what exactly should we measure. Maximum throughput from a singlethreaded/n-threaded client to X servers? (so that the servers aren't the bottleneck). CPU utilization with some constant workload? I am pretty sure that this netty stuff will produce more garbage.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/HintedReplayingDecoder.java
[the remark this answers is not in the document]
rvansa: >  why we have our own impl of ReplayingDecoder

Should I be more verbose than "to not attempt decoding unless enough bytes are read"? The issue was found in `ClientEventsOOMTest` where each event has 40M: this arrives chunked in 512kB buffers (and initially shorter ones). The buffer is then read in marshalling about 60 times, every time just finding out that it's too short. The hinted decoder finds out that we're trying to read 40M chunk and does not pass it to marshalling until enough data are in the buffer.

The bounds checks are executed once you know how many bytes should be read next, and that is either just next byte (with variable-length number encoding) or we are reading an array and do the bounds check only once.

I am not opposing avoiding the exception-throwing approach, but I don't know if I could rewrite it more efficiently.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/impl/EventDispatcher.java
```diff
@@ -0,0 +1,91 @@
...
+      this.listenerId = listenerId;
+      this.channel = channel;
+      this.cacheName = cacheName;
+   }
+
+   @Override
+   protected void decode(ChannelHandlerContext ctx, ByteBuf buf, List<Object> out) throws Exception {
+      T event;
+      try {
+         event = readEvent(buf);
```
wburns: Normally with netty you would want to read the event and then send that down the pipeline. This way you can disconnect the reading and invocation of the event into different execution handler or w/e is needed. Maybe the code isn't as conducive to this though.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/HintedReplayingDecoder.java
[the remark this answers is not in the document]
wburns: Makes sense, and yes I wanted the more detailed explanation (I just glanced at the code) :) I now better understand the issue this is trying to solve now and is an interesting approach. TBH I kinda wonder if the netty guys might want to include changes like this into their `ReplayingDecoder` even.

And tbh looking at how little decoding there is going on, the non exception approach could be better, but I doubt it would be that much, since there are so few possible checkpoints. It would however complicate the code substantially. So I would say to keep going with this for now.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/impl/EventDispatcher.java
rvansa: The event type that is decoded is tied to the way the event is executed. Therefore I would have two decoding classes, and two executing classes, logically coupled but theoretically interchangeable. Could you tell what are the pros? (do you expect more executing classes for the event in the future?)

wburns: The hotrod server does this for a few different reasons.

1. Optionally installed authentication so we don't execute if that operation is blocked.
2. Optionally log details about the operation (including how long decoding and execution took)
2. Execution of some operations on a different thread. We don't want to run an operation that will block the netty thread as we could starve reads from the socket.

The last one seems like it might be important for this since we are invoking user code and that could be doing anything. If I remember correctly the events are all sent back through the same socket for all listeners for a single client.

rvansa: @wburns But when you push something further to the pipeline, you still do not specify who should handle the event, do you? It's still on the event loop thread. In this case the notifier is called in the implementation, and that can forward it to the threadpool.

wburns: Not sure who you is in this case :) The decoder doesn't really care who handles it, it just decodes. A later Handler would have to do something with it.

The hotrod server has 2 handlers later in the pipeline that do almost the same thing (LocalContextHandler and ContextHandler) [1]. The former immediately invokes the command and sends the response (for commands that don't block) and the latter submits this on an executor to be handled in a different thread. (note these guys don't send anything else down the pipeline except exceptions).

Also you can extend `SimpleChannelInboundHandler` from netty and type it to the expected object class. For example we do that at [2] but override the matcher to be a bit faster.

[1] https://github.com/infinispan/infinispan/blob/master/server/hotrod/src/main/java/org/infinispan/server/hotrod/transport/HotRodChannelInitializer.java#L49
[2] https://github.com/infinispan/infinispan/blob/master/server/hotrod/src/main/java/org/infinispan/server/hotrod/ContextHandler.java#L286

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/TransportFactory.java
slaskawi: What I was trying to say is that the new implementation should implement this interface IMO.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
slaskawi: Either way works for me. But I guess Epoll will have impact on performance so I would include it here if I were in your shoes.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelInitializer.java
slaskawi: I'm not entirely sure if we should tie to JDK implementation here. Sometimes you just want to trust all certificates and just encrypt the traffic. In that case you have no truststore and you can use OpenSSL implementation. BTW, while implementing Netty-based REST, I noticed that using OpenSSL implementation on the client side gave me huge performance boost. It's really worth exploring.

The Javadoc [here](https://github.com/netty/netty/blob/ab9f0a0fda23e9c254cca9b18f5c19090b6c63ef/handler/src/main/java/io/netty/handler/ssl/SslContext.java#L82) suggests that you should use `SslContextBuilder`. 

BTW, if you don't want to explore it here - just log a JIRA and assign it to me. I have some experience with Netty+SSL.

slaskawi:
@rvansa I believe you are better than me in performance testing. But just for start and to see where we are I would suggest testing two things - Radargun/PerfACK and compare allocation rates of both (the old and new) clients.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/impl/EventDispatcher.java
rvansa: @wburns OK, I still don't see the benefit but I can restructure the handlers as you suggest.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/TransportFactory.java
rvansa: @slaskawi What's the point of keeping the interface when we would have to change the method signatures completely?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
rvansa: Ok, no issues with that.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/TransportFactory.java
slaskawi: As I mentioned before, it depends how do we treat this class. If we decided this is a private API, we can remove it in a minor release. However, if we decided that this is a public API, we can not remove it in a minor release.

Personally, I would treat it as a private API since it resides in `impl` package. But I'm really curious if @tristantarrant agrees with me here (and with you, since you're trying to remove this class).

tristantarrant: TcpTransportFactory is an actual API, since it is exposed in the configuration, although I doubt there are any practical situations where one would supply a custom implementation (unless, as @rvansa says, it is an extension of the existing one).
Therefore, for API preservation's sake, please keep the interface around, but log a deprecation warning

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/counter/operation/CompareAndSwapOperation.java
galderz: I guess you're talking about sync calls (e.g. put(k, v)) that require some kind of blocking if they can't write? For async calls (e.g. putAsync) you'd not need to block anyway.

Anyway, I guess it might depend on which the the thread is. You definitely don't want to block a thread belonging to Netty. An application thread blocking might be fine, e.g. for sync calls.

@wburns Thoughts?

rvansa: @galderz Yes, the intention is to block only application threads. How else would you apply backpressure? Queuing the write request does not solve anything.
IMO it's OK to block `putAsync()` when the network is saturated. You're not waiting until the server responds, just until you write stuff down the line.
AFAIK netty threads should not call cache.put() unless this is invoked from a listener (we probably discourage that).

galderz: > IMO it's OK to block putAsync() when the network is saturated. You're not waiting until the server responds, just until you write stuff down the line.

To me blocking putAsync() does not make sense. Even if the network is saturated, it should return early. When the network is no longer saturated, the request can be send, the response received and complete the completable future that putAsync returns.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/event/impl/EventDispatcher.java
wburns: If you find no benefit, don't worry about it. I was just saying that is how normally netty recommends doing this kind of stuff, with a separate decode and handler step [1]

[1] http://netty.io/wiki/user-guide-for-4.x.html#speaking-in-pojo-instead-of-bytebuf

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/counter/operation/CompareAndSwapOperation.java
rvansa: @galderz Would you throw an exception in this case, then?

galderz: Not sure but don't think so. If anything, the client that waits for the CompletableFuture could get a timeout exception.

rvansa: @galderz To sum our IRC discussion:
* `writeAndFlush` does not block the writer if channel is not writable, it queues the write
* I am strongly convinced that another layer of queuing wouldn't be beneficial
* we should probably throw exception if the channel queues too much

However, as the current code does not allow requests to be multiplexed over channel, there's always at most one in-flight operation on given channel. Therefore I'll defer any checks for overused channel to https://issues.jboss.org/browse/ISPN-8620 where this could really happen. The situation as in server (sending too many events which the client is not handling) does not happen at this moment.

rvansa:
Rebase on top of #5671 and addressed comments. I'll try do grab some perf data on Monday.

rvansa:
run performance tests please
cs

rvansa:
I've ran benchmark using 2 clients and 2 servers, and the results are too good to be true: it shows 34% throughput increase with single thread per client and 230% with 100 threads!
Please checkout the results (incl. logs): https://www.dropbox.com/s/l6o6wt3cqh1d9cv/results.zip?dl=0
(note that despite saying infinispan90 or infinispan92 it's always 9.2.0-SNAPSHOT with or w/o the netty client).

tristantarrant:
There is a karaf validator failure related to netty's epoll

rvansa:
run performance tests please
cs

rvansa:
run performance tests please
cs

rvansa:
run performance tests please
cs
and don't mess it again

rvansa:
run performance tests please
cs

rvansa:
run performance tests please
cs dist

rvansa:
Ok, the perfack shows regression close to none. I'll try to run my test on bigger cluster and see why the heck did I get so good results.

diegolovison:
run performance tests please
cs-hotrod

diegolovison:
run performance tests please
cs-hotrod

diegolovison:
run performance tests please
cs-hotrod dist

rvansa:
The stuck build in cachestore-remote is a bug; the event loop thread is waiting in native `epollWait` and the request to create a new channel is never handled.

diegolovison:
run performance tests please
cs-hotrod dist reads

diegolovison:
run performance tests please
cs-hotrod dist reads

rvansa:
Ah, this was a tricky one. Cachestore was setting async executor factory to DefaultExecutorFactory (from core) which defaults to single thread. However the ChannelFactory still thought there can be 99 threads (default value in ConfigurationProperties) and therefore the event loop group was using 16 event loops that shared single thread. When we tried to create another connection through different event loop, the task was never executed and deadlocked.

diegolovison:
run performance tests please
cs-hotrod dist reads

diegolovison:
run performance tests please
cs-hotrod

diegolovison:
run performance tests please
cs-hotrod

galderz:
@rvansa Is this ready to go in? Build timed out so kicked off another build on CI

rvansa:
@galderz Just started to check it; I don't think that timeouts are at random now in CI. So probably not, unless the build finishes fine (which I doubt) :-/

gustavonalle:
I was trying to get one more test run for this PR but the whole CI is simply unresponsive? Anyone has any idea how to revive it?

gustavonalle:
Triggered one last CI round, then I'll merge it

gustavonalle:
I rebased and run the tests locally, apparently two tests started failing:
```
[ERROR] cacheCreateWithEmbeddedConfigurationTest(org.infinispan.client.hotrod.admin.RemoteCacheAdminTest)  Time elapsed: 0.029 s  <<< FAILURE!
[ERROR] cacheCreateWithXMLConfigurationTest(org.infinispan.client.hotrod.admin.RemoteCacheAdminTest)  Time elapsed: 0.078 s  <<< FAILURE!
```

Could you check @rvansa ?

rvansa:
@gustavonalle Trivial merge issue, fixed.

gustavonalle:
Merged! Thanks @rvansa !

PR #5649, opened 2017-12-18
ISPN-8638 Create @Deprecated RegionFactory classes for compatibility

pferraro:
https://issues.jboss.org/browse/ISPN-8638

Please cherry-pick into 9.1.x.

### DIFF
--- a/hibernate-cache/src/main/java/org/hibernate/cache/infinispan/InfinispanRegionFactory.java
+++ b/hibernate-cache/src/main/java/org/hibernate/cache/infinispan/InfinispanRegionFactory.java
@@ -1,0 +1,14 @@
+/*
+ * Hibernate, Relational Persistence for Idiomatic Java
+ *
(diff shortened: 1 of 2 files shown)


galderz:
Not sure this is going to work seamlessly because:

1. You'd have same package shared by different artifacts out there. Don't know if that's going to play ball with Java 9 rules.
2. Even if we can apply these changes, they can only apply to one of the supported Hibernate versions by Infinispan. IOW, we can't define this classes for say both Hibernate 5.2 and 5.1 artifacts. See [here](https://github.com/infinispan/infinispan/pull/5638) for more info.

pferraro:
@galderz Re: (1) I don't think this is much of an issue as the org.hibernate:hibernate-infinispan and org.infinispan:infinispan-hibernate-cache modules are, by definition, mutually exclusive.

Re: (2), it would only work if the Hibernate 5.1 and 5.2 artifacts were also mutually exclusive.

Regardless, it looks like this is being handled in Hibernate itself.

PR #5651, opened 2017-12-19
ISPN-8637 ReadAfterLostDataTest.testPutMap failure

pruivo:
https://issues.jboss.org/browse/ISPN-8637

### DIFF
--- a/core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
+++ b/core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
@@ -1,5 +1,7 @@
 package org.infinispan.util.concurrent;
 
+import static java.lang.String.format;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -85,17 +87,19 @@ public void inject(
       if (backupOwners.isEmpty()) {
          return new PrimaryOwnerOnlyCollector<>();
       }
-      SingleKeyCollector collector = new SingleKeyCollector(id, backupOwners, topologyId);
+      SingleKeyCollector<T> collector = new SingleKeyCollector<>(id, backupOwners, topologyId);
       BaseAckTarget prev = collectorMap.put(id, collector);
-      assert prev == null : prev.toString();
+      //is it possible the have a previous collector when the topology changes after the first collector is created
+      //in that case, the previous collector must have a lower topology id
+      assert prev == null || prev.topologyId < topologyId : format("replaced old collector '%s' by '%s'", prev, collector);
```
tristantarrant: So, the only "fix" is an additional assert to verify an assumption. How does this "fix" things ? What if the user runs with assertions disabled ?

pruivo: the code is correct but the assertion wasn't. the test was failing because of the assertion. 
if you disable the assertions, the test is fine (and the code as well)

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -385,6 +392,7 @@ protected void markReady() {
 
    private class BiasedKeyCollector<T> extends SingleKeyCollector<T> {
       private final Address primaryOwner;
+      @SuppressWarnings("MismatchedQueryAndUpdateOfCollection")
```
rvansa: Add a comment that `unsolicitedAcks.removeIf(backupOwners::remove)` does the query (since it iterates through the collection), this is a bug in IntelliJ.

danberindei:
@pruivo the fix looks good, but I wonder if it would be even better to always remove the collector before retrying. E.g. if B starts a put with `owners(k) = AB` and A crashes, the retry will happen on `B` any RPCs, so we have to make sure we don't leak the ack collector.

pruivo:
The leak can't happen.
If A is the primary, it sends a sync rpc. If the rpc fails, the collector is completed (with exception) and removed.
If A is the backup, the topology change event completes the collector is the backup list is empty. The completion removes the collector.

PR #5652, opened 2017-12-20
ISPN-8645 Server no longer inherits from jboss-parent

ryanemerson:
https://issues.jboss.org/browse/ISPN-8645

The first commit removing the use of `<groupId>${project.groupId}</groupId>` is strictly only necessary for `pom.xml` and `bom/pom.xml`, however I have removed this use from all pom files so that we have a consistent approach.  In the event that we ever want to change from "org.infinispan", modifying the `<groupId>..` declarations via search and replace is trivial.


pom.xml
```diff
@@ -184,6 +184,7 @@
       <module.metadata.isCoreModule>false</module.metadata.isCoreModule>
       <module.skipComponentMetaDataProcessing>true</module.skipComponentMetaDataProcessing>
       <module.skipTestComponentMetaDataProcessing>true</module.skipTestComponentMetaDataProcessing>
+      <module.skipMavenRemoteReource>false</module.skipMavenRemoteReource>
```
rvansa: looks like a typo `Reource`

rvansa:
@ryanemerson I miss some justification in the JIRA why the server should not inherit from jboss-parent.

Have you removed the `${project.groupId}` just because you do not like it or was there any further reason behind this?

ryanemerson:
@rvansa You're right, I should have added more information in the JIRA. I have now added a description to the issue. The intention is for the server to utilise infinispan-parent instead of jboss-parent in order to reduce some code duplication and reduce the number of places where dependency versions need to be specified (where possible). Furthermore, this refactoring helps with the long term goal of migrating from wildfly to wildlfy-core.

I removed `${project.groupId}` because of maven's inheritance behaviour. If we utilise `${project.groupId}` in `infinispan-parent`, then when we inherit in `infinispan-server-versions`, it utilises the `groupId` of the child, i.e. `org.infinispan.server`. So in the parent `dependencyManagement`, `org.infinispan:infinispan-core` becomes `org.infinispan.server:infinispan-core` which obviously doesn't exist.

tristantarrant:
Merged, thanks

PR #5653, opened 2017-12-20
ISPN-8646 PessimisticTxPartitionAndMergeDuringRollbackTest.testDegrad…

pruivo:
…edPartitionWithDiscard failures

https://issues.jboss.org/browse/ISPN-8646

### DIFF
--- a/core/src/test/java/org/infinispan/partitionhandling/BaseTxPartitionAndMergeTest.java
+++ b/core/src/test/java/org/infinispan/partitionhandling/BaseTxPartitionAndMergeTest.java
@@ -61,7 +61,8 @@ public abstract class BaseTxPartitionAndMergeTest extends BasePartitionHandlingT
       controlledInboundHandler.filter = filter;
    }
 
-   protected FilterCollection createFilters(String cacheName, boolean discard, Class<? extends CacheRpcCommand> commandClass, SplitMode splitMode) {
+   FilterCollection createFilters(String cacheName, boolean discard, Class<? extends CacheRpcCommand> commandClass,
+         SplitMode splitMode) {
       Collection<AwaitAndUnblock> collection = new ArrayList<>(2);
       if (splitMode == SplitMode.ORIGINATOR_ISOLATED) {
          if (discard) {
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/infinispan/partitionhandling/BaseTxPartitionAndMergeTest.java
```diff
@@ -358,5 +362,12 @@ public void await(long timeout, TimeUnit timeUnit) throws InterruptedException {
       public void unblock() {
          collection.forEach(BaseTxPartitionAndMergeTest.AwaitAndUnblock::unblock);
       }
+
+      public void disableDiscard() {
+         collection.stream()
+               .map(awaitAndUnblock -> awaitAndUnblock instanceof DiscardFilter ? (DiscardFilter) awaitAndUnblock : null)
```
rvansa: `.filter(DiscardFilter.class::isInstance).map(DiscardFilter.class::cast)`

core/src/test/java/org/infinispan/partitionhandling/BaseTxPartitionAndMergeTest.java
```diff
@@ -289,16 +291,18 @@ public void unblock() {
 
       private final Class<? extends CacheRpcCommand> aClass;
       private final ReclosableLatch notifier;
+      private volatile boolean discard;
 
       private DiscardFilter(Class<? extends CacheRpcCommand> aClass) {
          this.aClass = aClass;
          notifier = new ReclosableLatch(false);
+         discard = true; //discard everything by default. change if needed in the future.
```
danberindei: @rvansa I guess the key is that now everything is discarded by default, before a command would be discarded only when `!notifier.isOpened()`

core/src/test/java/org/infinispan/partitionhandling/BaseTxPartitionAndMergeTest.java
pruivo: :+1:

core/src/test/java/org/infinispan/partitionhandling/BaseTxPartitionAndMergeTest.java
pruivo: that was the problem. only the first command is was discarded. The test expects the command only to be executed during the partition merging. So, I discard everything (only a single operation is executed in the tests) and I stop discarding when I merge

core/src/test/java/org/infinispan/partitionhandling/BaseTxPartitionAndMergeTest.java
pruivo: I changed the name to `stopDiscard()`

galderz:
> Test Result (16 failures / +9)

Giving it another CI spin

PR #5654, opened 2018-01-02
ISPN-8649 Fine-grained Netty dependencies

rvansa:
https://issues.jboss.org/browse/ISPN-8649

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -648,13 +648,53 @@
          </dependency>
          <dependency>
             <groupId>io.netty</groupId>
-            <artifactId>netty-all</artifactId>
+            <artifactId>netty-buffer</artifactId>
+            <version>${version.netty}</version>
+         </dependency>
+         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-codec</artifactId>
+            <version>${version.netty}</version>
+         </dependency>
+         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-codec-http</artifactId>
+            <version>${version.netty}</version>
+         </dependency>
+         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-codec-http2</artifactId>
+            <version>${version.netty}</version>
+         </dependency>
+         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-common</artifactId>
+            <version>${version.netty}</version>
+         </dependency>
+         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-handler</artifactId>
+            <version>${version.netty}</version>
+         </dependency>
+         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-resolver</artifactId>
             <version>${version.netty}</version>
          </dependency>
          <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-tcnative-boringssl-static</artifactId>
             <version>${version.netty.tcp.native}</version>
+         </dependency>
+         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport</artifactId>
+            <version>${version.netty}</version>
+         </dependency>
+         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-native-epoll</artifactId>
+            <version>${version.netty}</version>
          </dependency>
          <dependency>
             <groupId>org.rocksdb</groupId>
--- a/persistence/rest/pom.xml
+++ b/persistence/rest/pom.xml
@@ -19,7 +19,11 @@
       </dependency>
       <dependency>
          <groupId>io.netty</groupId>
-         <artifactId>netty-all</artifactId>
+         <artifactId>netty-handler</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.netty</groupId>
+         <artifactId>netty-transport-native-epoll</artifactId>
       </dependency>
       <dependency>
          <groupId>${project.groupId}</groupId>
--- a/server/core/pom.xml
+++ b/server/core/pom.xml
@@ -21,7 +21,11 @@
    <dependencies>
       <dependency>
          <groupId>io.netty</groupId>
-         <artifactId>netty-all</artifactId>
+         <artifactId>netty-handler</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.netty</groupId>
+         <artifactId>netty-transport-native-epoll</artifactId>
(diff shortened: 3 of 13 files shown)


rvansa:
I've dropped the bump, late notice that ALPN hacks break the update.

slaskawi:
@rvansa Could you please tell me what is the advantage of using small Netty jars over one uber jar? 

Also, productizing Netty is a PITA. So I think we should have really strong arguments to replace uber jar with small jars. 

Finally, could you please explain, what kind of problems did you experience with ALPN hacks? Maybe I could help?

//cc @jsenko

rvansa:
@slaskawi I think that consuming uberjars in a project using dependency management is a bad practice in general.

I was experiencing a problem in infinispan-remote uberjar once I've included the dependency to netty-all, though it is not fixed by this PR. My problem was that without manual exclusion in https://github.com/infinispan/infinispan/blob/master/all/remote/pom.xml#L282 the Import-Package included some packages referenced through netty-parent.pom which were not satisfied. One of the dependencies was related to deprecated netty-transport-udt, which is included in netty-all. I've tried to get rid of this using fine-grained dependencies instead, but there were other dependencies I had to exclude in the `pom.xml` anyway (Btw., do you understand why we have the wildcard inclusion there?).

Is any other JBoss project consuming the netty uberjar? Why is the productization PITA?

Some of the `ALPNHack*` classes inherits from a class with method that was changed between 4.1.9 and 4.1.19, since it's probably not considered a public API. Can't remember which one, but it would require code change (I didn't want to have code changes in this PR).

gustavonalle:
@slaskawi +1 to remove all uber jars: both produced ones and consumed ones

gustavonalle:
> Is any other JBoss project consuming the netty uberjar? Why is the productization PITA?

yes, but it consumes a tainted version of the netty-all uber jar that does not contain certain classes, one more reason for us to stop using it altogether

slaskawi:
> I was experiencing a problem in infinispan-remote uberjar once I've included the dependency to netty-all, though it is not fixed by this PR. My problem was that without manual exclusion in https://github.com/infinispan/infinispan/blob/master/all/remote/pom.xml#L282 the Import-Package included some packages referenced through netty-parent.pom which were not satisfied. One of the dependencies was related to deprecated netty-transport-udt, which is included in netty-all. I've tried to get rid of this using fine-grained dependencies instead, but there were other dependencies I had to exclude in the pom.xml anyway.

Netty-all pom doesn't seem to have any dependencies ([link](http://central.maven.org/maven2/io/netty/netty-all/4.1.19.Final/netty-all-4.1.19.Final.pom)). I can also confirm it with `mvn dependency:tree` on 'router' module:

```
[INFO] |        \- com.google.code.gson:gson:jar:2.8.2:compile
[INFO] +- io.netty:netty-all:jar:4.1.9.Final:compile
[INFO] +- org.assertj:assertj-core:jar:3.4.1:test
```

Netty-all is also a bundle (in the pom there is a Felix Bundle Plugin) but I don't see any entries in the `MANIFEST.MF`. So in my opinion our own uber jar module is either incorrect or tries to figure out too much automatically. 

> Btw., do you understand why we have the wildcard inclusion there?

With `!`, we list all packages that *should not* be in `Import-Package` entry and then use `*` to include anything else. In other words, we black-list packages we don't want and include everything else. Looking how quickly we add new dependencies, maybe it's the right time to invert that - type only packages we want to expose and ban (`!*`) everything else.

It probably needs some more investigation but I dare to say that with those manual exclusions in Remote Uber Jar, Netty-all should work just fine as small jars (maybe you will need to add 1-2 more, I haven't examined the jar content).

> Is any other JBoss project consuming the netty uberjar? Why is the productization PITA?

I believe so (but @jsenko can probably confirm this). Netty contains also native libraries and building them in Koji/Brew is quite complex. 

> Some of the ALPNHack* classes inherits from a class with method that was changed between 4.1.9 and 4.1.19, since it's probably not considered a public API. Can't remember which one, but it would require code change (I didn't want to have code changes in this PR).

Yes, ALPN and SNI hacks integrate tightly with Netty internal bits. Thanks for the version pointers. I'll have a look at it. See [ISPN-8657](https://issues.jboss.org/browse/ISPN-8657).

slaskawi:
@gustavonalle Just to give you some more context - I'm trying to weigh pros and cons using small Netty jars over uber jar. 

The reasoning behind using fat jars is simplicity. After this change gets in, you will need to remember that HTTP/1.1 codec resides in `netty-codec-http` whereas HTTP/2 in `netty-codec-http2`. Whenever you want to turn the epoll on, you will be trying to recall `netty-transport-native-epoll`'s name. It's s very fragmented library. Of course there is a price for this "simplicity" - if you you're not using HTTP/2 at all, you will still get it from uber jar. There might be also other Netty's 3rd party dependencies that we are not aware of, but still, they cost us nothing because they won't be loaded in the runtime.

I probably forgot about ton of other arguments, but is it really worth to add those 237 lines of code and deal with fragments of Netty library rather than all-in-one solution?

rvansa:
@slaskawi 
> our own uber jar module is either incorrect or tries to figure out too much automatically.

Yes, I think that the wildcard is interpreted as 'any package that's referenced'. In my case the first import that was brought in was `com.barchart.udt` through one of the classes in `io.netty.channel.udt.nio`.

> maybe it's the right time to invert that - type only packages we want to expose and ban (!*) everything else

Thinking about that a bit more, I am not sure if this is the right approach. The goal of OSGi is to fail fast on missing dependency; this enforces us to either include the dependency or say 'this ain't gonna be needed in runtime, trust me'. So are we trying to protect ourselves or the user of the uberjar?

> Netty contains also native libraries and building them in Koji/Brew is quite complex

But we're building them anyway and then just repack it in `netty-all` anyway, don't we?

rvansa:
@slaskawi 
> It's s very fragmented library

*Modular* is the word you were seeking ;-)

FTR despite your arguments I am still more inclined towards 'cleaner' though not 'simpler' solution. For example for the Hot Rod client I don't want to include HTTP codecs (ok, until some of the single-port stuff gets in).

gustavonalle:
@slaskawi Uberjars should not be an artifact in maven or any dependency management system. By definition, they were created for people do hello words quickly without worrying about dependencies.

There are plenty of problems they introduce from the moment you start using them inside maven, please check for example https://github.com/netty/netty/issues/4671

> It's s very fragmented library

I'd call it properly modularized

> I probably forgot about ton of other arguments, but is it really worth to add those 237 lines of code and deal with fragments of Netty library rather than all-in-one solution?

IMO yes

tristantarrant:
Merged, thanks

PR #5655, opened 2018-01-03
A POC for replacing explicit null-checks by automatically generated

denis-zhdanov:
This *PR* illustrates an approach when explicit *null*-checks are replaced by those automatically generated during compilation based on source code annotations.  

Let's consider [ArrayMap. containsValue()](https://github.com/infinispan/infinispan/blob/master/commons/src/main/java/org/infinispan/commons/util/ArrayMap.java#L47) as an example:  

*Source code before*  

```java
public boolean containsValue(Object value) {
   Objects.requireNonNull(value);
   for (int i = 0; i < values.length; ++i) {
      Object v = values[i];
      if (v != null && v.equals(value)) return true;
   }
   return false;
}
```  

*Source code after*  

```java
public boolean containsValue(@Nonnull Object value) {
   for (int i = 0; i < values.length; ++i) {
      Object v = values[i];
      if (v != null && v.equals(value)) return true;
   }
   return false;
}
```  

Resulting bytecode looks like if it's compiled from the source below:  

```java
public boolean containsValue(@Nonnull Object value) {
   if (value == null) {
      throw new NullPointerException("value is mandatory");
   }
   for (int i = 0; i < values.length; ++i) {
      Object v = values[i];
      if (v != null && v.equals(value)) return true;
   }
   return false;
}
```  

That is achieved by the [Traute](http://traute.oss.harmonysoft.tech/) *javac* plugin. If the team likes the approach, I'm ok with applying it to the whole project in a current/new *PR*.

### DIFF
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -25,6 +25,11 @@
       </dependency>
 
       <dependency>
+         <groupId>com.google.code.findbugs</groupId>
+         <artifactId>jsr305</artifactId>
+      </dependency>
+
+      <dependency>
          <groupId>org.jboss.spec.javax.transaction</groupId>
          <artifactId>jboss-transaction-api_1.1_spec</artifactId>
          <scope>provided</scope>
--- a/commons/src/main/java/org/infinispan/commons/util/ArrayMap.java
+++ b/commons/src/main/java/org/infinispan/commons/util/ArrayMap.java
@@ -1,5 +1,6 @@
 package org.infinispan.commons.util;
 
+import javax.annotation.Nonnull;
 import java.util.AbstractCollection;
 import java.util.AbstractSet;
 import java.util.Arrays;
@@ -44,8 +45,7 @@ public abstract class ArrayMap<K, V> extends java.util.AbstractMap<K, V> {
    }
 
    @Override
-   public boolean containsValue(Object value) {
-      Objects.requireNonNull(value);
+   public boolean containsValue(@Nonnull Object value) {
       for (int i = 0; i < values.length; ++i) {
          Object v = values[i];
          if (v != null && v.equals(value)) return true;
--- a/pom.xml
+++ b/pom.xml
@@ -269,6 +269,7 @@
       <version.jgoodies.forms>1.0.7</version.jgoodies.forms>
       <version.jmh>1.12</version.jmh>
       <version.jsap>2.1</version.jsap>
+      <version.jsr305>3.0.2</version.jsr305>
       <version.jstl>1.2</version.jstl>
       <version.junit>4.12</version.junit>
       <version.karaf-maven-plugin>4.1.2</version.karaf-maven-plugin>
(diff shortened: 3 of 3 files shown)


rvansa:
@denis-zhdanov Out of curiosity, do you know how much longer would the build take? (Even dry run inspecting all methods).

Can Traute work with type annotations? Are the null checks omitted when you can statically verify that the variable contains non-null value?

denis-zhdanov:
Hey @rvansa,  

1. I didn't observe any build time change difference - the plugin is very lightweight and most of the time during compilation is spend to I/O anyway
2. *Traute* doesn't support type annotations at the moment and it doesn't attempt to check if *null* check is necessary, it just generates it based on the configured rules (e.g. annotated method parameters)
3. Yes, current *PR* targets replacing explicit checks with automatic but there additional benefits:  
    * the code is better documented when target method parameters/return values are marked by *NonNull*
    * it's easier to add new checks - just add new annotations
    * it's possible to use [non-null by default with nullable exceptions](http://traute.oss.harmonysoft.tech/core/javac/#72-notnullbydefault-annotations) approach to reduce boilerplate even more

galderz:
Does it work fine when compiling/running tests with typical IDEs? Developers are mostly using Eclipse or IntelliJ

denis-zhdanov:
Hey @galderz,

It's possible to configure *IntelliJ* to use the plugin for the compilation, I assume, it should be possible for *Eclipse* as well.  

I created corresponding tickets for the [eclipse](https://github.com/denis-zhdanov/traute/issues/90) and [intellij](https://github.com/denis-zhdanov/traute/issues/91) plugins

tristantarrant:
As an annotation processor, it should be supported automatically by both Eclipse and Idea, just like the JBoss Logging stuff. However I hope that the generated code is self-standing and doesn't require the jsr305 dependency at runtime (which can therefore be marked as optional)

denis-zhdanov:
@tristantarrant, it does require a runtime dependency because  the *javax.annotation.Nonnull* has *RUNTIME* retention.  

However, it's possible to use any other annotation with, say, *SOURCE* retention

tristantarrant:
runtime retention only means that using the various annotation discovery APIs at runtime will find them IFF they are present on the classpath. If they are missing, they simply won't be returned by those APIs. Obviously this is a problem if the generated code depends on their presence at runtime. I'll inspect the PR tomorrow

tristantarrant:
@denis-zhdanov 
- jsr305 annotations should really replace jcip in Infinispan
- the jsr305 dependency should be optional (the annotations are not needed at runtime)
- the annotation processor should be placed on the classpath as a "provided" scope dependency, otherwise the default APT discovery of javac is prevented (for example, with your PR jboss-logging-processor doesn't run).

@rvansa while you are right that the @NonNull annotations do not raise the confidence level, IDEs have built-in support for them for static analysis, highlighting problems at the caller (e.g. IDEA: settings/inspections/probable bugs/@NotNull/@Nullable problems)

tristantarrant:
I've pushed a branch with some of the fixes suggested above: https://github.com/tristantarrant/infinispan/commits/traute

rvansa:
@tristantarrant Will you include a test showing that the checks are in with the automatic discovery?

tristantarrant:
Yes, I will take over this PR. Thanks @denis-zhdanov

tristantarrant:
Closing this. I'll reopen it in early 9.3.x

PR #5660, opened 2018-01-08
ISPN-8647 Initialize all loggers statically

tristantarrant:
This prevents a potential deadlock in the Logger factory which synchronizes on the ClassLoader.

https://issues.jboss.org/browse/ISPN-8647

### DIFF
--- a/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
@@ -1,5 +1,6 @@
 package org.infinispan.container.offheap;
 
+import java.lang.invoke.MethodHandles;
 import java.util.AbstractCollection;
 import java.util.AbstractSet;
 import java.util.Collection;
@@ -39,8 +40,8 @@ import org.infinispan.util.logging.LogFactory;
  * @since 9.0
  */
 public class OffHeapDataContainer implements DataContainer<WrappedBytes, WrappedBytes> {
-   protected final Log log = LogFactory.getLog(getClass());
-   protected final boolean trace = log.isTraceEnabled();
+   private static final Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
+   protected final boolean trace = getLog().isTraceEnabled();
 
    protected final AtomicLong size = new AtomicLong();
    protected final int lockCount;
@@ -115,7 +116,7 @@ public class OffHeapDataContainer implements DataContainer<WrappedBytes, Wrapped
       locks.lockAll();
       try {
          if (size.get() != 0) {
-            log.warn("Container was not cleared before deallocating memory lookup tables!  Memory leak " +
+            getLog().warn("Container was not cleared before deallocating memory lookup tables!  Memory leak " +
                   "will have occurred!");
          }
          clear();
@@ -431,7 +432,7 @@ public class OffHeapDataContainer implements DataContainer<WrappedBytes, Wrapped
 
    protected void performClear() {
       if (trace) {
-         log.trace("Clearing off heap data");
+         getLog().trace("Clearing off heap data");
       }
       memoryLookup.toStreamRemoved().forEach(address -> {
          while (address != 0) {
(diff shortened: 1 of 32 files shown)


rvansa:
@tristantarrant While I agree with making the loggers static (despite losing info about inheriting class), I don't think that the output is equivalent to the old behaviour. Simple test tells me:
```java
import java.lang.invoke.MethodHandles;

public class LookupTest {
   static class A {
      static final Class<?> clazz = MethodHandles.lookup().lookupClass();

      public void foo() {
         System.out.println(clazz.getName());
      }
   }

   static class B extends A {
   }

   public static void main(String[] args) {
      new B().foo();
   }
}
```
and the output is `LookupTest$A`. If this is intended you don't have to use MethodHandles, you can simply state the class directly.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
```diff
@@ -39,7 +40,7 @@
  * @since 9.0
  */
 public class OffHeapDataContainer implements DataContainer<WrappedBytes, WrappedBytes> {
-   protected final Log log = LogFactory.getLog(getClass());
+   protected static final Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
```
wburns: Unfortunately before this was used to show trace messages from the appropriate implementation. I have had to fix this before and the way to do this is to instead add an abstract `getLog` method to the abstract class that is implemented by the extended classes to properly say where the messages are from. This way the extending class can still define the logger as static but the abstract class uses the appropriate logger.

core/src/main/java/org/infinispan/notifications/cachelistener/BaseQueueingSegmentListener.java
```diff
@@ -26,7 +27,7 @@
  * @since 7.0
  */
 abstract class BaseQueueingSegmentListener<K, V, E extends Event<K, V>> implements QueueingSegmentListener<K, V, E> {
-   protected final Log log = LogFactory.getLog(getClass());
+   protected static final Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
```
wburns: This can use the `getLog` approach as well.

core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
```diff
@@ -66,7 +67,7 @@
  * @param <S> The stream interface
  */
 public abstract class AbstractCacheStream<T, S extends BaseStream<T, S>, S2 extends S> implements BaseStream<T, S> {
-   protected final Log log = LogFactory.getLog(getClass());
+   protected static final Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
```
wburns: Actually this line can be completely removed (I missed removing it on a refactor). This is already using the `getLog` approach I mentioned on the off heap one.

core/src/main/java/org/infinispan/stream/impl/local/AbstractLocalCacheStream.java
```diff
@@ -19,7 +20,7 @@
  * stream is populated
  */
 public abstract class AbstractLocalCacheStream<T, S extends BaseStream<T, S>, S2 extends S> implements BaseStream<T, S> {
-   protected final Log log = LogFactory.getLog(getClass());
+   protected static final Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
```
wburns: This line can be removed - unused.

core/src/main/java/org/infinispan/container/offheap/OffHeapDataContainer.java
wburns: Check out https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java#L153

tristantarrant:
I have used the getLog() approach where inheritance matters.

wburns:

PR #5671, opened 2018-01-11
ISPN-8649 Add epoll binaries

rvansa:


gustavonalle:
Why does it need to be in several places, as opposed to be declared just in server-core and propagated transitively?

rvansa:
@gustavonalle I was merely following all locations of netty-all. I recall having compilation failures somewhere without the explicit dependencies, but I can try to remove them.

gustavonalle:
merged, thanks

PR #5675, opened 2018-01-11
Ispn 8396 interceptor out of memory

wburns:


wburns:
Still need to add documentation and clean up some TODOs.

galderz:
Needs rebasing

core/src/main/java/org/infinispan/configuration/cache/Configurations.java
```diff
@@ -19,11 +19,19 @@ private Configurations() {
    }
 
    public static boolean isOnePhaseCommit(Configuration cfg) {
+      // Otherwise pessimistic transactions will be one phase commit
+      if (cfg.memory().size() > 0 && cfg.memory().evictionType().isExceptionBased()) {
```
slaskawi: Perhaps you could extract this statement into a method. This way, the method name would explain everything and the comment above would be redundant.

core/src/main/java/org/infinispan/configuration/cache/Configurations.java
```diff
@@ -19,11 +19,19 @@ private Configurations() {
...
+         return false;
+      }
       return !cfg.clustering().cacheMode().isSynchronous() ||
             cfg.transaction().lockingMode() == LockingMode.PESSIMISTIC;
    }
 
    public static boolean isOnePhaseTotalOrderCommit(Configuration cfg) {
+      // Even total order needs 2 phase commit with this - isn't even worth it then?
```
slaskawi: Good point. Maybe we should spit out a warning to the user? Something like - "TotalOrder optimization will be disabled due to Blocking Eviction".

core/src/main/java/org/infinispan/container/entries/CacheEntrySizeCalculator.java
```diff
@@ -27,76 +30,81 @@ public CacheEntrySizeCalculator(EntrySizeCalculator<? super K, ? super V> calcul
...
+      } else {
+         metadata = null;
+      }
+      long keyValueMetadataSize = calculateSize(key, ice.getValue(), metadata);
+      return keyValueMetadataSize + noMetadataSize;
+   }
+
+   @Override
+   public long calculateSize(K key, V value, Metadata metadata) {
+      long objSize = calculator.calculateSize(key, value);
```
slaskawi: I was just wondering if we could hire a tool (or just copy some bits from it) that will calculate the size for us? Maybe this: http://openjdk.java.net/projects/code-tools/jol/?

core/src/main/java/org/infinispan/factories/SizeCalculatorFactory.java
```diff
@@ -0,0 +1,47 @@
...
+   public <T> T construct(Class<T> componentType) {
+      MemoryConfiguration memory = configuration.memory();
+      if (memory.evictionType().isMemoryBased()) {
+         StorageType type = memory.storageType();
+         switch (type) {
+            case BINARY:
+               return componentType.cast(CacheEntrySingleton.INSTANCE);
+            case OFF_HEAP:
+               return componentType.cast(componentRegistry.getComponent(OffHeapEntryFactory.class));
+            case OBJECT:
```
slaskawi: The case can be removed. It will always fall into `default`. Although *with* a case it's a bit more readable...

core/src/main/java/org/infinispan/container/entries/CacheEntrySizeCalculator.java
wburns: This could definitely be a different implementation of a SizeCalculator. At the time I used reflection and simple addition as it was substantially faster. Unfortunately this approach only supports primitives, primitive Wrappers, Strings and arrays.

One thing about jol is that it doesn't count nested objects. So this means we would have to invoke it at least 3 times (possibly 6) to estimate a CacheEntry (CacheEntry itself, key, value, Metadata, Version, Custom Metadata). This doesn't even take into account if the key or value have nested objects, which jol won't add in either.  Needless to say to support jol to be a valid estimation we would have to expose the Calculator to the user to implement and then tell us how to estimate use jol for their key and values as well.

core/src/main/java/org/infinispan/factories/SizeCalculatorFactory.java
wburns: Yeah I am aware, I more put it there so the comment explained why it fell into default is all.

core/src/main/java/org/infinispan/container/entries/CacheEntrySizeCalculator.java
wburns: Actually I might be wrong about the nested part, I will have to look into this at a later point.

core/src/main/java/org/infinispan/configuration/cache/Configurations.java
wburns: What would you recommend the method name to be then?

core/src/main/java/org/infinispan/configuration/cache/Configurations.java
wburns: I can add that to the validate method on the ConfigurationBuilder maybe. I guess it should be an INFO message or do you think WARN?

wburns:
Rebased and added method for Configurations class. I still have to address TODOs and documentation though.

galderz:
Needs rebasing again

wburns:
Yup I merged some of the changes into a different PR, since they weren't directly related :)

wburns:
Rebased and updated now that all dependent bugs are already in master. That took much longer than expected, now I can finally finish TODOs and documentation. Also for whatever reason I can't run a core test suite on my box on this branch or master even :( Hopefully CI has a better chance.

wburns:
CI ran fine, I don't what is wrong with my local box then :( Still needs cleanup changes though.

galderz:
Will, we're integrating this now. Please create a JIRA for any follow up cleanup changes and we can integrate that later on.

wburns:
Sounds good. I got quite a bit done last night, but I forgot how much more stuff was needed. I revamped how the configuration is done (reused EvictionStrategy) and added server changes for DMR. Also I was talking to Dan about how to change up the test to be faster, but that is going to take longer :(

I will be glad to have it separated though, just don't tout the changes too much yet :D

PR #5678, opened 2018-01-15
ISPN-8667 EncoderCache.withWrapping(Class) is not implemented

wburns:
https://issues.jboss.org/browse/ISPN-8667


tristantarrant:
Merged, thanks

PR #5684, opened 2018-01-16
Upgrade admin console to 9.2.0.CR1

vblagoje:
Master only. Upgrade console to 9.2.0.CR1 I ran the build and it picked up the new version. I also tested localhost:9990 after adding a user and it worked. GTG


galderz:
Thanks @vblagoje! :)

PR #5688, opened 2018-01-18
ISPN-8695 Create remote caches using the XML configuration

tristantarrant:
https://issues.jboss.org/browse/ISPN-8695

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManagerAdmin.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManagerAdmin.java
@@ -3,7 +3,9 @@ package org.infinispan.client.hotrod;
 import java.util.EnumSet;
 
 import org.infinispan.client.hotrod.exceptions.HotRodClientException;
+import org.infinispan.commons.api.BasicCache;
 import org.infinispan.commons.api.CacheContainerAdmin;
+import org.infinispan.commons.configuration.BasicConfiguration;
 
 /**
  * Remote Administration operations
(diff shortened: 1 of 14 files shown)


gustavonalle:
Build timeout, retriggering...

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManagerAdmin.java
```diff
@@ -25,6 +27,18 @@
    @Override
    <K, V> RemoteCache<K, V> createCache(String name, String template) throws HotRodClientException;
 
+   /**
+    * Creates a cache on the remote server cluster using the specified configuration
+    *
+    * @param name the name of the cache to create
+    * @param configuration a concrete configuration that will be sent to the server. The XML representation of this
+    *                      configuration must be in a format supported by the server.
```
gustavonalle: It's not clear what is the XSD of the configuration that must be passed in. Is it ```urn:infinispan:config:9.2``` (embedded)  or  ```urn:infinispan:server:core:9.2``` (server)? To avoid any doubt, I think we should mention in the javadoc

tristantarrant: I'd really like to have a common schema which is then xincluded into the two subschemas (embedded and server) which can then add their own idiosyncrasies. But at the moment this is the embedded schema.

gustavonalle: that's fine, IMO we should just mention that in the javadoc

tristantarrant:
@gustavonalle Updated the javadoc as requested

gustavonalle:
Merged, thanks @tristantarrant !

PR #5695, opened 2018-01-23
ISPN-8714 Don't send acks for async-triangle

pruivo:
https://issues.jboss.org/browse/ISPN-8714


pruivo:
@rvansa I don't think it is needed. If I'm not sending the ack and it is needed, some test would fail. If I send the ack and it isn't needed, it doesn't hurt :+1:

gustavonalle:
@pruivo Can you rebase to retrigger CI?

pruivo:
@gustavonalle sure

gustavonalle:
@pruivo There are some test failures. Some of them are on master, could you double check?

pruivo:
@gustavonalle I don't think they are related. I run the test suite locally a couple of times and I got no failures.

gustavonalle:
merged, thanks @pruivo !

PR #5698, opened 2018-01-25
ISPN-8721 rearrange constructor to avoid race condition

tomdearman:
rearrange constructor to avoid race condition leading to NPE in beginInvalidatingWithPFER

### DIFF
--- a/hibernate/cache-commons/src/main/java/org/infinispan/hibernate/cache/commons/access/PutFromLoadValidator.java
+++ b/hibernate/cache-commons/src/main/java/org/infinispan/hibernate/cache/commons/access/PutFromLoadValidator.java
@@ -158,6 +158,12 @@ public class PutFromLoadValidator {
 		else {
 			throw log.pendingPutsMustHaveMaxIdle();
 		}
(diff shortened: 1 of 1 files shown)


rvansa:
Thanks for the fix. It would be even better if you could add a test-case as well.

tomdearman:
I spent some time with trying to understand how to get NodeEnvironment test util class to work such that you could guarantee one fake environment starts up part way through whilst another already started tries to update its cache but I was unable to do that.  I suppose now I know where the problem is maybe something could be done directly with the PutFromLoadValidator class to directly test it.  But since it would still require that the constructor was partially completed, I am not sure how to go about that.  Any examples of how that is done elsewhere to test a similar race condition in infinispan?

Tom

galderz:
@rvansa I think it'd be hard to develop a unit test case for this, I think we should just integrate it.

rvansa:
@galderz Will you file a PR against ORM 5.1/5.2?

galderz:
@rvansa Fixed in cache-commons

PR #5707, opened 2018-01-29
ISPN-8708 Remote iteration with limited batch sends finished segments

wburns:
too early

* Use KeyWatchingListener for each publisher

https://issues.jboss.org/browse/ISPN-8708

### DIFF
--- a/core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
+++ b/core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
@@ -576,21 +576,23 @@ public class DistributedCacheStream<R> extends AbstractCacheStream<R, Stream<R>,
     * @param <S>
     */
    private class CompletionListenerRehashIterator<S> extends RehashIterator<S> {
-      private final KeyWatchingCompletionListener completionListener;
+      private final Consumer<? super Supplier<PrimitiveIterator.OfInt>> userListener;
+
+      private volatile CompletionRehashPublisherDecorator completionRehashPublisherDecorator;
 
       private CompletionListenerRehashIterator(Iterable<IntermediateOperation> intermediateOperations,
-            Consumer<? super Supplier<PrimitiveIterator.OfInt>> completionListener) {
+            Consumer<? super Supplier<PrimitiveIterator.OfInt>> userListener) {
          super(intermediateOperations);
-         this.completionListener = new KeyWatchingCompletionListener(completionListener);
+         this.userListener = userListener;
       }
 
       @Override
       protected S getNext() {
          S next = super.getNext();
          if (next != null) {
-            completionListener.valueIterated(next);
+            completionRehashPublisherDecorator.valueIterated(next);
          } else {
-            completionListener.completed();
+            completionRehashPublisherDecorator.complete();
          }
          return next;
       }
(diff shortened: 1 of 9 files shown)


rvansa:
@wburns I guess it wouldn't be easy to create a reproducer, would it?

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -651,109 +662,154 @@ public void close() {
...
+         this.completionListeners = Collections.synchronizedList(new ArrayList<>(4));
+         this.userListener = completionListener;
       }
 
       @Override
       protected S getNext() {
          S next = super.getNext();
          if (next != null) {
-            completionListener.valueIterated(next);
+            for (KeyWatchingCompletionListener keyWatchingCompletionListener : completionListeners) {
```
rvansa: is it guaranteed that all listeners will be initialized when this gets called?

wburns: Yes, they are always initialized the way the code works.

And actually if this changes it doesn't matter since the listeners only care about entries that were retrieved by their own publisher.

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -541,11 +534,29 @@ private IdentityPublisherDecorator() { }
                lostSegments.accept((Supplier<PrimitiveIterator.OfInt>) ourSegments::iterator);
             }
          });
-         return decorateBeforeReturn(convertedPublisher);
+         return iteratorOperation.handlePublisher(convertedPublisher, keyConsumer);
       }
+   }
+
+   private class RehashPublisherDecorator<S> extends AbstractRehashPublisherDecorator<S> {
```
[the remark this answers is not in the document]
wburns: Sure, I can do that. Unfortunately these classes will only make sense in the scope of this class, but for clarity I can move them out.

wburns:
I might be able to create a reproducer that uses a controlled Publisher, let me look into it a bit.

danberindei:
@wburns Which test?

wburns:
The test is https://github.com/infinispan/infinispan/blob/master/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/impl/iteration/MultiServerDistRemoteIteratorTest.java

I could usually reproduce about 40% of the time.

wburns:
I am almost done with a test that always reproduces it, unfortunately it is very verbose :(

wburns:
Updated and added test that failed on previous implementation, but new one passes.

rvansa:
Thanks, Will. I'll ask you, though, to create a `package-info.java` in `o.i.stream.impl` and document the big picture with reference to the concrete classes. The code has grown in complexity and `ClusterStreamManager` javadoc is not exhaustive.

wburns:
Yes I can agree some additional documentation would be helpful.

I am also finding that while using Producer/Subscriber is very flexible it is extremely hard to follow code as you don't know which is which most of the time :(

I was planning on writing a blog post about these changes in a very high level, I will try to get some actual documentation in code at the same time.

Thanks @rvansa

rvansa:
It's not a problem of the reactive APIs but also when e.g. using suppliers. I often had to trace a long chain to find out what is the actual value (e.g. if is given supplier of entry set providing just the local or remote entries etc.). Calling variables `supplier` then does not help much.

wburns:
Yeah I realized this as well. I unfortunately only renamed the _userListener_ in the CompletionRehashPublisherDecorator which helped a little, but unfortunately there are quite a few :)

PR #5708, opened 2018-01-30
ISPN-8732 Avoiding waiting indefinitely

gustavonalle:
https://issues.jboss.org/browse/ISPN-8732


karesti:
LGTM! So sometimes the test is failing just due to the lack of timeout on await or because there is another failure behind ? I guess we will see. 
Await is used in other tests too, but should be fine. I believed await had already a timeout !

gustavonalle:
I'd say it is probably waiting for some that will never happen, this PR will simply makes it fail faster.

gustavonalle:
I triggered another CI run

karesti:
@gustavonalle I'm going to merge and backport this to JDG, and we will see for the clusteredlocksplitbrain test after

PR #5710, opened 2018-01-30
ISPN-8701 IndexOutOfBoundsException in SIFS during shutdown

rvansa:
https://issues.jboss.org/browse/ISPN-8701

I started working on a reproducer but finding a sequence operations that would take the index to the desired state (see JIRA) seemed too complicated.


tristantarrant:
Would this be necessary for 9.1.x ?

rvansa:
@tristantarrant This is a low-priority bug, since the failure during stop just forces the index to be rebuilt when the store is started (startup delay). And a couple of errors in the log.

It affects all versions since 9.0.

gustavonalle:
Integrated into master and 9.1.x, thanks @rvansa !

PR #5712, opened 2018-01-30
ISPN-8670 Add a configuration storage for standalone servers

diegolovison:
Created a test case the verify if the configuration remains the same
after the server has been restarted when shared equals true.


galderz:
Looks good, thanks @diegolovison

PR #5713, opened 2018-01-31
DistSyncTxL1FuncTest.testGetBlockingAnotherGetCacheEntry random failures

danberindei:
https://issues.jboss.org/browse/ISPN-8742

getAsync() doesn't always run interceptor callbacks on a separate thread.
The remote get can finish before the calling thread calls thenAccept(),
especially when the test suite is limited to a single CPU.

The fix is trivial, figuring how the next test was affected took me much longer :)

### DIFF
--- a/core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
+++ b/core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
@@ -3,7 +3,6 @@ package org.infinispan.test;
 import static org.testng.AssertJUnit.assertTrue;
 import static org.testng.AssertJUnit.fail;
 
-import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.IdentityHashMap;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
```diff
@@ -57,7 +57,7 @@
       boolean isSatisfied() throws Exception;
    }
 
-   protected static final Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
+   protected final Log log = LogFactory.getLog(getClass());
```
wburns: Hrmm is this an issue with the other JIRA that changed this @tristantarrant ?

danberindei: @wburns The problem is that logs of tests use the inherited field, and their messages end up logged with the `AbstractInfinispanTest` category. I initially wanted to make it private static, but there were too many tests using it and I wouldn't have been able to sneak it into this PR that way :)

PR #5717, opened 2018-02-01
ISPN-8720 Implement test for REST with SSL configuration template

vjuranek:
https://issues.jboss.org/browse/ISPN-8720

### DIFF
--- a/server/integration/testsuite/src/test/resources/arquillian.xml
+++ b/server/integration/testsuite/src/test/resources/arquillian.xml
@@ -339,6 +339,20 @@
                 <property name="jmxDomain">${server.jmx.domain}</property>
             </configuration>
         </container>
+        <container qualifier="standalone-rest-ssl" mode="manual">
+            <configuration>
+                <property name="javaHome">${server.jvm}</property>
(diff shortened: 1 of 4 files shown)


server/integration/testsuite/src/test/resources/arquillian.xml
```diff
@@ -339,6 +339,20 @@
                 <property name="jmxDomain">${server.jmx.domain}</property>
             </configuration>
         </container>
+        <container qualifier="standalone-rest-ssl" mode="manual">
```
galderz: Hmmm, do we really need a new container for this? Couldn't the hotrod-ssl container contain configuration to access rest over ssl? The more containers we have, the longer it takes to test :|

vjuranek: Unfortunately no, this is a test for particular config template which we provide as part of ISPN distribution, i.e. each test in `ExampleConfigsIT` has to start its own container to test given template. We can either reduce number of templates we provide or skip testing of the templates (and especially later option I wouldn't recommend).

PR #5721, opened 2018-02-01
ISPN-8528 Move FailoverRequestBalancingStrategy to public package

rvansa:
https://issues.jboss.org/browse/ISPN-8528

Motivation for the change in configuration from class/instance -> factory: Currently the ChannelFactory uses single instance for all threads which leads to extensive locking as the instances of balancing strategy are not thread safe. In the future (once we remove the deprecated configuration method providing it with an instance) a thread-local strategy would be preferred.


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/Configuration.java
```diff
@@ -86,12 +85,34 @@ public ExecutorFactoryConfiguration asyncExecutorFactory() {
       return asyncExecutorFactory;
    }
 
-   public Class<? extends FailoverRequestBalancingStrategy> balancingStrategyClass() {
-      return balancingStrategyClass;
+   /**
+    * @deprecated since 9.3
```
tristantarrant: Do you want this merged after we fork ? Still, you should supply javadocs  on the replacement.

rvansa: My view of 'CR' releases is that PRs should get there only if 'needed' (bugfix etc.) or 'really really wanted'. This is neither, I just want to get this in strictly before 10 so that 10 can remove those ifaces. And I didn't expect this to be reviewed in time for Final :)
If you feel so, I can change these to 9.2, though.

rvansa:
@tristantarrant I didn't want to rush this in at the break of Final, but seems we're already postponing... Should I change the javadocs to 9.2 and let you merge this?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationChildBuilder.java
```diff
@@ -49,9 +50,18 @@
    /**
     * For replicated (vs distributed) Hot Rod server clusters, the client balances requests to the
     * servers according to this strategy.
+    *
+    * @deprecated since 9.3, use {@link #balancingStrategy(Supplier)} instead.
     */
+   @Deprecated
    ConfigurationBuilder balancingStrategy(FailoverRequestBalancingStrategy balancingStrategy);
```
galderz: This will break users as is since they'd expect to pass in a `org.infinispan.client.hotrod.impl.transport.tcp.FailoverRequestBalancingStrategy` here. This method should still take org.infinispan.client.hotrod.impl.transport.tcp.FailoverRequestBalancingStrategy, as fully qualified name, and then we could have a wrapper that delegates to the newer location.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/AbstractConfigurationChildBuilder.java
```diff
@@ -56,6 +57,11 @@ public ConfigurationBuilder balancingStrategy(FailoverRequestBalancingStrategy b
       return builder.balancingStrategy(balancingStrategy);
```
galderz: Same problem here, this old, deprecated method should take FQN of old `org.infinispan.client.hotrod.impl.transport.tcp.FailoverRequestBalancingStrategy`

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationChildBuilder.java
rvansa: @galderz `org.infinispan.client.hotrod.impl.transport.tcp.FailoverRequestBalancingStrategy` implements `org.infinispan.client.hotrod.FailoverRequestBalancingStrategy`, so this can be still safely passed in. Recompilation is needed as the method signature has changed but that's an allowed change (we don't guarantee binary compatibility).

galderz:
Hmmm, could [this failure](https://ci.infinispan.org/job/Infinispan/job/PR-5721/3/testReport/junit/org.infinispan.client.hotrod/DistTopologyChangeUnderLoadTest/testPutsSucceedWhileTopologyChanges/) be related to this? Doubt it but wondering...

rvansa:
@galderz Oops, no, this is rather a bug introduced in ISPN-8619 :-/

Do you think I should change the `@since` tags to 9.2 and get this in?

galderz:
> Do you think I should change the @since tags to 9.2 and get this in?

Yeah that'd be fine for me.

rvansa:
@galderz Updated.

rvansa:
And now I have to change the `@since` tags back...

rvansa:
`Timed out waiting for responses for request WDYT from @tristantarrant`

Rebased, changed `@deprecated since` versions...

PR #5722, opened 2018-02-01
ISPN-8700 Ensure that REST test threads include the test name

tristantarrant:
Fixes the fact that RestAccessLoggingTest occasionally fails when running concurrently with other tests


server/rest/src/test/java/org/infinispan/rest/AuthenticationTest.java
```diff
@@ -76,7 +76,7 @@ public void shouldRejectNotValidAuthorizationString() throws Exception {
 
       BasicAuthenticator basicAuthenticator = new BasicAuthenticator(securityDomainMock, "ApplicationRealm");
 
-      restServer = RestServerHelper.defaultRestServer().withAuthenticator(basicAuthenticator).start();
+      restServer = RestServerHelper.defaultRestServer().withAuthenticator(basicAuthenticator).start(this.getTestName());
```
slaskawi: How about getting current's thread name (which would be probably a Surefire fork) rather than grabbing test name explicitly. This way the start method wouldn't need to accept parameters.

tristantarrant: Well, this is a test helper, so I'm not bothered about parameters. I've aligned the naming to use the AbstractInfinispanTest strategy.

danberindei: +1 to keep the parameter. 

However, `getTestName()` returns the full class name, so I suggest using `TestResourceTracker.getCurrentTestShortName()` instead.

tristantarrant:
Fixed

gustavonalle:
pulling

gustavonalle:
Actually, this PR was run 3x by CI, and the failures were always high (> 40), could you check @tristantarrant ?

tristantarrant:
Weird that  a PR that only touches server/rest/ files has so many failures on core...

gustavonalle:
@tristantarrant Could you rebase? Let's wait for another CI run

tristantarrant:
rebased

gustavonalle:
restarted the build

PR #5724, opened 2018-02-01
ISPN-8751 Fix RestAccessLoggingTest

vjuranek:
https://issues.jboss.org/browse/ISPN-8751


tristantarrant:
@vjuranek I'll make necessary changes to https://github.com/infinispan/infinispan/pull/5722 and close this one
Notice however that in your Jira the log is wrong because the thread filter is not capturing the correct log.

PR #5725, opened 2018-02-02
ISPN-8740 Refactor backup-write commands (triangle)

pruivo:
https://issues.jboss.org/browse/ISPN-8740


core/src/main/java/org/infinispan/commands/triangle/BackupWriteCommand.java
```diff
@@ -0,0 +1,142 @@
...
+
+   void setCommonAttributesFromCommand(WriteCommand command) {
+      this.commandInvocationId = command.getCommandInvocationId();
+      this.topologyId = command.getTopologyId();
+      this.flags = command.getFlagsBitSet();
+   }
+
+   abstract WriteCommand createWriteCommand();
+
+   abstract Object anyKey();
```
rvansa: This method looks weird. I wonder if we should put the segmentId to constructor (and send it over the wire), or at least define the getSegmentId() separately for each command... Take a second thought about this, please.

danberindei:
@pruivo would you like this to go into 9.2.0.Final?

pruivo:
@rvansa updated:
* set the segment-id in the primary. removed the `KeyPartitioner` from the backup commands
* removed deprecated method usage:` CacheEntry.setValid()` 
* removed `writeTo()`/`readFrom()` from the abstract commands.

@danberindei yep :)

PR #5727, opened 2018-02-02
ISPN-8755 Incorrect local repository path in OSGi integration tests

danberindei:
https://issues.jboss.org/browse/ISPN-8755

### DIFF
--- a/client/hotrod-client/pom.xml
+++ b/client/hotrod-client/pom.xml
@@ -505,6 +505,15 @@
                </execution>
             </executions>
          </plugin>
+         <plugin>
+            <groupId>org.apache.karaf.tooling</groupId>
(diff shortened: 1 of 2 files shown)


danberindei:
Needs some more work... fixing the local repository location is not enough, because the integration tests need `org.infinispan.protostream:protostream:xml:features:<version>`, and that artifact is not downloaded by the build in the local repository.

karesti:
@danberindei LGTM, CI looks fine and I'm ready to click on the merge button if you are fine on that. or do you need additional reviews ?

danberindei:
@karesti yes, it's fine. It doesn't really improve anything in CI, since we're not using a custom repository location here.

PR #5728, opened 2018-02-05
[PREVIEW] Single port

slaskawi:
https://issues.jboss.org/browse/ISPN-8756
Design document: https://github.com/infinispan/infinispan-designs/blob/master/Single_port.adoc

This is a preview version of a Single Port functionality. Currently it handles only `HTTP/1.1 Upgrade` and `TLS/ALPN` types of "upgrade" for REST server. However the most of the refactoring has been done and now I need to plug Hot Rod Server in and implement wiring in the server modules. Feel free to comment.

TODO:
* [ ] Upgrade to Hot Rod using TLS/ALPN and `HTTP/1.1 Upgrade`
* [ ] Implement wiring in the endpoint
* [ ] Implement an integration test


server/core/src/main/java/org/infinispan/server/core/configuration/SslConfiguration.java
```diff
@@ -18,7 +18,7 @@
    private final boolean requireClientAuth;
    private final Map<String, SslEngineConfiguration> sniDomainsConfiguration;
 
-   SslConfiguration(boolean enabled, boolean requireClientAuth, Map<String, SslEngineConfiguration> sniDomainsConfiguration) {
+   public SslConfiguration(boolean enabled, boolean requireClientAuth, Map<String, SslEngineConfiguration> sniDomainsConfiguration) {
```
ryanemerson: Why does this need to be public? It doesn't appear to be used anywhere other than in `SslConfigurationBuilder`

server/core/src/main/java/org/infinispan/server/core/configuration/SslEngineConfiguration.java
```diff
@@ -21,7 +21,7 @@
    private final char[] trustStorePassword;
    private final char[] keyStoreCertificatePassword;
 
-   SslEngineConfiguration(String keyStoreFileName, String keyStoreType, char[] keyStorePassword, char[] keyStoreCertificatePassword, String keyAlias,
+   public SslEngineConfiguration(String keyStoreFileName, String keyStoreType, char[] keyStorePassword, char[] keyStoreCertificatePassword, String keyAlias,
```
ryanemerson: Does not need to be public

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
```diff
@@ -0,0 +1,139 @@
...
+import org.infinispan.configuration.global.GlobalConfiguration;
+import org.infinispan.distexec.DefaultExecutorService;
+import org.infinispan.distexec.DistributedCallable;
+import org.infinispan.distexec.DistributedExecutorService;
+import org.infinispan.jmx.JmxUtil;
+import org.infinispan.manager.EmbeddedCacheManager;
+
+import io.netty.channel.group.ChannelGroup;
+
+public class NettyTransportConnectionStats {
```
ryanemerson: Does this need to be public?

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
```diff
@@ -0,0 +1,139 @@
...
+   private final EmbeddedCacheManager cacheManager;
+   private final boolean isGlobalStatsEnabled;
+   private final ChannelGroup acceptedChannels;
+   private final String threadNamePrefix;
+   private final AtomicLong totalBytesWritten = new AtomicLong();
+   private final AtomicLong totalBytesRead = new AtomicLong();
+
+   public NettyTransportConnectionStats(EmbeddedCacheManager cacheManager, ChannelGroup acceptedChannels, String threadNamePrefix) {
+      this.cacheManager = cacheManager;
+      this.isGlobalStatsEnabled = cacheManager != null ? cacheManager.getCacheManagerConfiguration().globalJmxStatistics().enabled() : false;
```
ryanemerson: Can be simplified to `cacheManager != null && cacheManager.getCacheManagerConfiguration().globalJmxStatistics().enabled()`

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
```diff
@@ -0,0 +1,139 @@
...
+      if (cacheManager != null) {
+         org.infinispan.remoting.transport.Transport transport = cacheManager.getTransport();
+         return transport != null && transport.getMembers().size() > 1;
+      }
+      return false;
+   }
+
+   private int calculateGlobalConnections() {
+      Cache<Object, Object> cache = cacheManager.getCache();
+      DistributedExecutorService exec = new DefaultExecutorService(cache);
```
ryanemerson: I know you simply copied this from `NettyTransport`, but could we use `ClusterExecutor` instead as `DistributedExecutorService` is deprecated?

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
```diff
@@ -0,0 +1,139 @@
...
+            return (Integer) mbeanServer.getAttribute(transportMBeanName, "NumberOfLocalConnections");
+         } catch (MBeanException | AttributeNotFoundException | InstanceNotFoundException | ReflectionException |
+               MalformedObjectNameException e) {
+            throw new RuntimeException(e);
+         }
+      }
+   }
+
+   public Integer getNumberOfLocalConnections() {
+      return Integer.valueOf(acceptedChannels.size());
```
ryanemerson: Unnecessary boxing?

server/rest/src/main/java/org/infinispan/rest/Http11To2UpgradeHandler.java
```diff
@@ -27,19 +28,25 @@
...
    private static final int MAX_HEADER_SIZE = 8192;
 
-   private final RestServer restServer;
+   private final int maxContentLength;
+   private final ChannelHandler http11Handler;
+   private final ChannelHandler http20Handler;
+   private final boolean useAlpn;
 
-   Http11To2UpgradeHandler(RestServer restServer) {
+   public Http11To2UpgradeHandler(ChannelHandler http11Handler, ChannelHandler http20Handler, int maxContentLength, boolean useAlpn) {
```
ryanemerson: We could just `RestServerConfiguration` to set the class variables instead of explicitly passing `maxContentLength` and `useAlpn`?

So:
```java
new Http11To2UpgradeHandler(new Http11RequestHandler(server), new Http20RequestHandler(server), server.getConfiguration().maxContentLength(), server.getConfiguration().ssl().enabled())
```

becomes:
```java
new Http11To2UpgradeHandler(new Http11RequestHandler(server), new Http20RequestHandler(server), server.getConfiguration())
```

server/router/src/main/java/org/infinispan/server/router/configuration/builder/RouterConfigurationBuilder.java
```diff
@@ -1,17 +1,18 @@
...
-import org.infinispan.server.router.configuration.MultiTenantRouterConfiguration;
+import org.infinispan.server.router.configuration.RouterConfiguration;
 
 /**
  * Multi tenant router configuration builder.
  *
  * @author Sebastian Łaskawiec
  */
-public class MultiTenantRouterConfigurationBuilder implements MultiTenantConfigurationBuilderParent {
+public class RouterConfigurationBuilder implements ConfigurationBuilderParent {
```
ryanemerson: Do we have to worry about backwards compatibility here? cc @tristantarrant

server/core/src/main/java/org/infinispan/server/core/configuration/SslConfiguration.java
slaskawi: It's used as a part of a Single Port configuration. It turned out that `NettyTransport` takes `SslConfiguration` and `SslEngineConfiguration` and prepares `SslContext` based on that. So the easiest way to reuse `NettyTransport` is to feed it with a `SslConfiguration` instance.

Unfortunately using a `SslConfigurationBuilder` is a bit hard since it accepts `ProtocolServerConfigurationChildBuilder` in its constructor (and I don't have it at hand since a Single Port has different configuration root).

So here are the options I've been thinking about to solve this puzzle:
* Make `SslConfiguration` public [not very nice but short and simple].
* Add new constructor without `ProtocolServerConfigurationChildBuilder` to `SslConfigurationBuilder` [should work fine but you might get `NullPointerException`s on some method calls, which feels a bit wrong].
* Create interfaces for `SslConfiguration` and `SslEngineConfiguration` and implement them in Single Port Router [lots and lots of typing and duplicating code].
* Forget about `NettyTransport` and bootstrap Netty manually in a Single Port module [lots of duplicated code handling SSL and ALPN].

So I decided to take the shortest route even though it's a bit ugly....

slaskawi: But I'm absolutely open to some new options how to wire those bits up...

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
slaskawi: Good point.

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
slaskawi: Good point!

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
slaskawi: TBH, I don't know. Maybe @wburns will know what's the latest, greatest...

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
slaskawi: Good point!

server/rest/src/main/java/org/infinispan/rest/Http11To2UpgradeHandler.java
slaskawi: It seems like a good idea. I'll try it out and see how other bits wire up. If I won't discover any problems, I'll do that!

server/router/src/main/java/org/infinispan/server/router/configuration/builder/RouterConfigurationBuilder.java
slaskawi: I think it might be a good point. However this feature is so much tied to the server that I won't expect anybody to use it directly. Even though it's bending the rules, I would just document class signature changes in the upgrading guide and leave the new names.

server/core/src/main/java/org/infinispan/server/core/configuration/SslConfiguration.java
ryanemerson: At the minute your code utilises `SslConfigurationBuilder(null)` in the `SinglePortRouterBuilder` and `new SslConfiguration(..)` is never called outside of the builder. So this is a bit different to your description unless I am missing something. Do you intend to use the `SslConfiguration` constructor later?

ryanemerson: But I appreciate all of your points and can't think of a cleaner way, so if making SslConfiguration's constructor public is the least painful way, then so be it :+1:

slaskawi: Done

server/core/src/main/java/org/infinispan/server/core/configuration/SslEngineConfiguration.java
slaskawi: Done

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
slaskawi: Done

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
slaskawi: Done

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
slaskawi: Ok, so for now I'll leave it as is... Once I get a clarification from @wburns I'll consider changing this.

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
slaskawi: Done

server/rest/src/main/java/org/infinispan/rest/Http11To2UpgradeHandler.java
slaskawi: I somewhat reverted the changes I made here. At first I wanted to make REST protocol optional. But this doesn't make any sense since with `HTTP/1.1 Upgrade` and `TLS/ALPN` type of upgrades you should always have a fallback protocol. This effectively means that we should always have a REST Server as a fallback.

gustavonalle: Could you elaborate on "make REST protocol optional" and "always have a REST Server as a fallback"?

gustavonalle: It'd be really bad if REST clients such as curl, wget, etc, weren't able to talk to the server since REST is gone :)

slaskawi: ooh no, no.. I'm not removing anything ;)

There are two upgrade paths for the client. 

The first one is to "upgrade" existing connection to the new protocol. This method uses `HTTP 1.1/Upgrade` header. This of course requires some "existing connection", so we are assuming you have a client that is already talking to the HTTP Server. This probably requires a REST Server (I could imagine writing a simple HTTP handler which is only capable of upgrading it's connection but using a real REST Server makes more sense to me).

The second method is using `TLS/ALPN`. During TLS handshake the client and the server negotiates connection protocol. There is a change (due to OpenSSL versions mismatch for example) that this negotiation is not properly understood by the server. In such case, you're using a fallback protocol which is `HTTP 1.1`. Again I can imagine using a mock server or something similar, but using a real REST Server makes more sense.

So to sum it up - technically it is possible to use negotiation mechanism without a REST Server. Practically, it doesn't make much sense (at least to me).

slaskawi:
Rebased and comments addressed. All features are in, now I'll be focusing on the boring part (server wiring and e2e tests). Feel free to drop me more comments.

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
galderz: `DistributedExecutorService` is deprecated, so this is a good time to move it over to `ClusterExecutor` :)

server/router/src/main/java/org/infinispan/server/router/configuration/builder/RouterConfigurationBuilder.java
galderz: This code runs in the server so don't think there's a need to keep backwards compatibility. Users don't normally have to touch this. The only users it could potentially affect is those that create an embedded server, but this is very much niche use case.

slaskawi:
> DistributedExecutorService is deprecated, so this is a good time to move it over to ClusterExecutor :)

@galderz Unfortunately it doesn't seem to be a no-brainer task. The APIs are different and require different approach and I can see no tests around it. I created [ISPN-8873](https://issues.jboss.org/browse/ISPN-8873) to address this.

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
wburns: Yeah we should move this over to `ClusterExecutor`. You should be able to use something like: https://gist.github.com/wburns/0ebd63d53c9384c3ca021baac70641f6

slaskawi: +1, but as I explained on the bottom of the page, I would prefer to implement it in a separate PR. I created ISPN-8873 to track this.

tristantarrant:
There are many tests for the ClusterExecutor: https://github.com/infinispan/infinispan/tree/master/core/src/test/java/org/infinispan/manager
Since this is a rewrite of the code, I'd prefer to see s/DistributedExecutor/ClusterExecutor/ done here rather than in a subsequent PR.

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
```diff
@@ -0,0 +1,139 @@
...
+      @Override
+      public void setEnvironment(Cache<Object, Object> cache, Set<Object> inputKeys) {
+         this.cache = cache;
+      }
+
+      @Override
+      public Integer call() throws Exception {
+         GlobalConfiguration globalCfg = cache.getCacheManager().getCacheManagerConfiguration();
+         String jmxDomain = globalCfg.globalJmxStatistics().domain();
+         MBeanServer mbeanServer = JmxUtil.lookupMBeanServer(globalCfg);
```
tristantarrant: We should be doing this directly without jumping through a local JMX call...

slaskawi: Could you please show me how?

slaskawi:
@tristantarrant I really, really, (really) don't think that's a good idea. There are little to none tests around this so this PR will already be gigantic (believe me). I you really insist I can do it (being more precise incorporate ISPN-8873 into this PR), but I won't be able to test all the scenarios... That's the price for writing no tests unfortunately - it's bloody hard to refactor some parts of the system.

slaskawi:
@tristantarrant @galderz @wburns @ryanemerson Ok, I finally managed to integrate everything together... This PR is ready to be reviewed.

slaskawi:
Rebased. None of the failures occurred on my local box.

server/rest/src/test/java/org/infinispan/rest/http2/Http11ResponseHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private final Queue<FullHttpResponse> responses = new LinkedBlockingQueue<>();
+
+    private final Semaphore semaphore = new Semaphore(1);
+
+    @Override
+    protected void channelRead0(ChannelHandlerContext ctx, FullHttpResponse msg) throws Exception {
+        responses.add(msg);
+        semaphore.release();
+        System.out.println("Released");
```
galderz: These needs cleaning?

slaskawi:
Rebased.

@galderz @ryanemerson - May I ask for your review guys?

ryanemerson:
@slaskawi Yep! I'll make sure I go through it before the end of the week

server/core/src/main/java/org/infinispan/server/core/transport/NettyTransportConnectionStats.java
galderz: Please create a follow up JIRA for this

server/rest/src/test/java/org/infinispan/rest/http2/Http11ResponseHandler.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public Queue<FullHttpResponse> getResponses() {
+        try {
+            System.out.println("Getting responses");
```
tristantarrant: Another system.out

server/rest/src/test/java/org/infinispan/rest/http2/Http11ResponseHandler.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public void sendRequest(FullHttpRequest request, SslContext sslContext, Channel channel) {
+        HttpScheme scheme = sslContext != null ? HttpScheme.HTTPS : HttpScheme.HTTP;
+        request.headers().set(HttpConversionUtil.ExtensionHeaderNames.SCHEME.text(), scheme.name());
+        request.headers().set(HttpHeaderNames.CONNECTION, HttpHeaderValues.CLOSE);
+        request.headers().set(HttpHeaderNames.ACCEPT_ENCODING, HttpHeaderValues.TEXT_PLAIN);
+
+        channel.writeAndFlush(request);
+        semaphore.acquireUninterruptibly();
+        System.out.println("Acquired");
```
tristantarrant: And another

server/rest/src/test/java/org/infinispan/rest/http2/NettyHttpClient.java
```diff
@@ -0,0 +1,85 @@
...
+      b.group(workerGroup);
+      b.channel(NioSocketChannel.class);
+      b.option(ChannelOption.SO_KEEPALIVE, true);
+
+      b.remoteAddress(host, port);
+      b.handler(initializer);
+
+      // Start the client.
+      channel = b.connect().syncUninterruptibly().channel();
+      System.out.println("Connected to [" + host + ':' + port + ']');
```
tristantarrant: And another system.out

galderz:
I've created [ISPN-9251](https://issues.jboss.org/browse/ISPN-9251) and [ISPN-9252](https://issues.jboss.org/browse/ISPN-9252) for the other leftovers here.

slaskawi:
Thanks @galderz

PR #5729, opened 2018-02-05
ISPN-8741 - Temporary fix for unlock/retrylocks

karesti:
* Don't retry tryLocks while unlock due to topology change is not over
* Adds more trace logs


lock/src/main/java/org/infinispan/lock/impl/functions/UnlockFunction.java
```diff
@@ -12,52 +12,74 @@
...
+   public UnlockFunction(String requestId, Set<Object> requestors) {
       this.requestId = requestId;
-      this.requestor = requestor;
+      this.requestors = requestors;
    }
 
    @Override
-   public Void apply(EntryView.ReadWriteEntryView<ClusteredLockKey, ClusteredLockValue> entryView) {
+   public Boolean apply(EntryView.ReadWriteEntryView<ClusteredLockKey, ClusteredLockValue> entryView) {
+      log.tracef("unlock request by reqId [%s] requestors %s", requestId, requestors);
```
wburns: Might as well add the `static final boolean isTraceEnabled` so we don't always invoke these when tracing is disabled.

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -108,8 +114,26 @@ public void handleLockResult(Boolean result, Throwable ex) {
...
+   public class UnlockRequestHolder extends RequestHolder<Void> {
+
+      public UnlockRequestHolder(Object requestor, CompletableFuture<Void> request) {
+         super(requestor, request);
+      }
+
+      @Override
+      protected void handle(Boolean result) {
+         // Unlock result
+
```
wburns: Was there supposed to be something here?

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -170,19 +202,25 @@ protected void handle(Boolean result) {
...
                // Meanwhile another request for this owner might have locked it successfully and we don't want to unlock in that case
-               unlock(requestId, requestor);
+               unlock(requestId, Collections.singleton(requestor));
             }
-         } else if(!isScheduled){
+         } else if (!isScheduled) {
+            log.tracef("Schedule for expiration %s", this);
             // If the lock was not acquired, then schedule a complete false for the given timeout
             isScheduled = true;
             clusteredLockManager.schedule(() -> request.complete(false), time, unit);
```
wburns: We should log a JIRA to clean up old scheduled tasks when the request is complete as well.

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -59,7 +62,9 @@
    private final EmbeddedClusteredLockManager clusteredLockManager;
    private final FunctionalMap.ReadWriteMap<ClusteredLockKey, ClusteredLockValue> readWriteMap;
    private final Queue<RequestHolder> pendingRequests;
+   private final Queue<UnlockRequestHolder> pendingUnlocks;
```
wburns: Is this needed? I thought that unlocks were fine during a rehash event?

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -221,34 +255,101 @@ public void entryRemoved(CacheEntryRemovedEvent event) {
...
+       * @param possibleOwners
+       */
+      private void forceUnlockForLeavingMembers(Set<Object> possibleOwners) {
+         clusteredLockManager.execute(() -> {
+                  log.tracef("LOCK[%s] Call force unlock for %s from %s ", getName(), possibleOwners, originator);
+                  viewChangeUnlockHappening.set(true);
+                  unlock(null, possibleOwners)
+                        .whenComplete((unlockResult, ex) -> {
+                           log.tracef("LOCK[%s] Force unlock call completed for %s from %s ", getName(), possibleOwners, originator);
+                           viewChangeUnlockHappening.set(false);
```
wburns: I am wondering if a boolean is sufficient. If we have 2 nodes leave consecutively, but not at the same time I think we could get into a state where lock acquisitions could occur. If so we could use an AtomicInteger instead and only fire pending requests if it was zero after decrementing.

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -221,34 +255,101 @@ public void entryRemoved(CacheEntryRemovedEvent event) {
...
+                           if (ex == null) {
+                              log.tracef("LOCK[%s] Force unlock result %b for %s from %s ", getName(), unlockResult, possibleOwners, originator);
+                           } else {
+                              log.errorf(ex, "LOCK[%s] Unlock failed from node %s", getName(), originator);
```
wburns: This should be a `Log` entry.

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -264,8 +365,13 @@ private void lock(RequestHolder<Void> requestHolder) {
          return;
 
       pendingRequests.offer(requestHolder);
```
wburns: I don't see where pendingRequests is removed if the lock is acquire below in the eval method?

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
```diff
@@ -59,118 +79,128 @@ public void testLockUseAfterPartitionWithoutMajority() throws Throwable {
...
       ClusteredLock lock1 = clm1.get(LOCK_NAME);
       ClusteredLock lock2 = clm2.get(LOCK_NAME);
       ClusteredLock lock3 = clm3.get(LOCK_NAME);
+      ClusteredLock lock4 = clm4.get(LOCK_NAME);
+      ClusteredLock lock5 = clm5.get(LOCK_NAME);
+
+      splitCluster(new int[]{0, 1, 2, 3}, new int[]{4, 5});
+
+      asList(lock0, lock1, lock2, lock3).forEach(lock -> {
+         assertTrue("Lock acquisition should be true " + lock, await(lock.tryLock(29, TimeUnit.SECONDS)
```
wburns: 29 seconds is a fun number :)

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
```diff
@@ -59,118 +79,128 @@ public void testLockUseAfterPartitionWithoutMajority() throws Throwable {
...
+      CompletableFuture<Boolean> tryLock2Result = lock2.tryLock(1, TimeUnit.SECONDS);
+
+      assertTrue("Just one of the locks has to work", await(tryLock1Result) ^ await(tryLock2Result));
+
+      assertFailureFromMinorityPartition(lock0);
    }
 
    @Test
-   public void testAutoReleaseIfLockIsAcquiredFromAMinorityPartition() throws Throwable {
+   public void testTryLocksBeforeSplitBrain() throws Throwable {
```
wburns: So does this tests always pass?

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -221,34 +255,101 @@ public void entryRemoved(CacheEntryRemovedEvent event) {
...
+            log.tracef("LOCK[%s] Nothing to do, all nodes are present %s", getName(), originator);
+            return;
+         }
+
+         if (leavingNodes.size() >= newMembers.size()) {
+            log.tracef("LOCK[%s] Nothing to do, we are on a minority partition notification on %s", getName(), originator);
+            return;
+         }
+
+         //TODO : Not sure this can happen
```
wburns: This should be removed. This node address can't not be in the cluster view.

lock/src/main/java/org/infinispan/lock/impl/functions/UnlockFunction.java
wburns: Should do that across all the classes.

karesti: I have a Jira to add lock names everywhere. I can handle that there, wtyt ?
ISPN-8769

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: I experienced some issues with unlock too, but not 100% anymore

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: I'm wondering still if I need to hold unlocks too. I would like to just trace the unlock result

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: Is already done ! https://issues.jboss.org/browse/ISPN-8757

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: right, nice catch

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
karesti: It fails a lot on the topology change issue. I want to work on that with @danberindei

lock/src/main/java/org/infinispan/lock/impl/functions/UnlockFunction.java
wburns: sure

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: ok

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: it is removed on the listener, done requests are removed and skipped while seeking for the next requests to be treated. Do you think it is worth to remove it from the pending requests too to avoid memory issues ?

karesti:
@wburns I reported your last comments. could you have a look ? the cleanup of pending requests, could be handled in the cleanup scheduler issue

lock/src/main/java/org/infinispan/lock/impl/functions/UnlockFunction.java
danberindei: +1, but it should be `trace` instead of `isTraceEnabled`.

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -221,34 +255,95 @@ public void entryRemoved(CacheEntryRemovedEvent event) {
...
+      log.tracef("LOCK[%s] Unlock pending requests size[%d] in %s", getName(), pendingUnlocks.size(), originator);
+      while (!pendingUnlocks.isEmpty() && !isChangeViewUnlockInProgress()) {
+         UnlockRequestHolder unlockRequestHolder = pendingUnlocks.poll();
+         unlock(unlockRequestHolder);
+      }
+   }
+
+   private void retryPendingRequests() {
+      if (isChangeViewUnlockInProgress()) {
+         log.tracef("LOCK[%s] Hold pending requests while view change unlock is happening in %s", getName(), originator);
```
danberindei: Why is this necessary? Could we have a request from a leaver that is executed during (or even after) the view change unlock?

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -221,34 +255,95 @@ public void entryRemoved(CacheEntryRemovedEvent event) {
...
+                           log.tracef("LOCK[%s] viewChangeUnlockHappening value in %s ", getName(), viewChangeUnlockValueAfterUnlock, originator);
+                           if (ex == null) {
+                              log.tracef("LOCK[%s] Force unlock result %b for %s from %s ", getName(), unlockResult, possibleOwners, originator);
+                           } else {
+                              log.error(ex, log.unlockFailed(getName(), getOriginator()));
+                              // TODO: handle the exception. Retry ? End all the pending requests in this lock ?
+                           }
+
+                           retryPendingUnlocks();
+                           retryPendingRequests();
```
danberindei: I understand unlocking already acquires the lock for the new requestor, so is this really necessary?

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -290,23 +390,48 @@ private void tryLock(TryLockRequestHolder requestHolder) {
...
    }
 
    @Override
    public CompletableFuture<Void> unlock() {
       log.tracef("unlock called from %s", originator);
       CompletableFuture<Void> unlockRequest = new CompletableFuture<>();
-      readWriteMap.eval(lockKey, new UnlockFunction(originator)).whenComplete((lockResult, ex) -> {
+
+      if (isChangeViewUnlockInProgress()) {
+         pendingUnlocks.offer(new UnlockRequestHolder(originator, unlockRequest));
```
danberindei: What happens if `forceUnlockForLeavingMembers` decrements `viewChangeUnlockHappening` *and* calls `retryPendingUnlocks()` between lines 407 and 408?

lock/src/main/java/org/infinispan/lock/impl/log/Log.java
```diff
@@ -13,7 +13,16 @@
 @MessageLogger(projectCode = "ISPN")
 public interface Log extends org.infinispan.util.logging.Log {
    String LOCK_DELETE_MSG = "The lock was deleted.";
+   String NODE_LEFT_MSG = "The node has left the cluster.";
+   String UNLOCK_FAILED_MSG = "LOCK[%s] Unlock failed from node %s";
```
danberindei: Why are these constants? Are you thinking about reusing them somewhere else?

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
```diff
@@ -1,40 +1,60 @@
...
-      splitCluster(new int[]{0, 1}, new int[]{2, 3});
+      splitCluster(new int[]{0, 1, 2}, new int[]{3, 4, 5});
 
-      eventuallyEquals(2, () -> advancedCache(0).getRpcManager().getTransport().getMembers().size());
-      eventuallyEquals(2, () -> advancedCache(1).getRpcManager().getTransport().getMembers().size());
-      eventuallyEquals(2, () -> advancedCache(2).getRpcManager().getTransport().getMembers().size());
-      eventuallyEquals(2, () -> advancedCache(3).getRpcManager().getTransport().getMembers().size());
+      for (int i = 0; i < 6; i++) {
+         int finalI = i;
+         eventuallyEquals(3, () -> advancedCache(finalI).getRpcManager().getTransport().getMembers().size());
```
danberindei: This is basically what `TestingUtil.blockUntilViewReceived(long, Cache...)` does.

karesti:
@danberindei any additional comment ?

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
```diff
@@ -59,118 +79,132 @@ public void testLockUseAfterPartitionWithoutMajority() throws Throwable {
...
+      ClusteredLock lock4 = clm4.get(LOCK_NAME);
+      ClusteredLock lock5 = clm5.get(LOCK_NAME);
+
+      splitCluster(new int[]{0, 1, 2}, new int[]{3, 4, 5});
+
+      asList(lock0, lock1, lock2, lock3, lock4, lock5).forEach(lock -> {
+         assertNotNull(lock);
+         assertFalse(await(lock.tryLock(100, TimeUnit.MILLISECONDS).exceptionally(ex -> {
+            assertException(ClusteredLockException.class, ex);
+            assertException(AvailabilityException.class, ex.getCause());
```
danberindei: Could use `assertException(ClusteredLockException.class, AvailabilityException.class, ex)`, or even better`expectExecutionException(ClusteredLockException.class, AvailabilityException.class, lock.tryLock(100, MILLISECONDS)` if you expect the call to fail.

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
```diff
@@ -59,118 +79,132 @@ public void testLockUseAfterPartitionWithoutMajority() throws Throwable {
...
+
+   private void assertTryLock(ClusteredLock lock) {
+      assertTrue("Lock acquisition should be true " + lock, await(lock.tryLock(29, TimeUnit.SECONDS)
+            .thenApply(tryLockRequest -> {
+               if (tryLockRequest) {
+                  await(lock.unlock());
+               }
+               return tryLockRequest;
+            }).exceptionally(ex -> {
+               fail("Should not be failing from majority partition " + lock + " " + ex.getMessage());
```
danberindei: I don't think `exceptionally()` is needed when you expect the call to succeed, `await()` will rethrow the exception anyway.

lock/src/main/java/org/infinispan/lock/impl/log/Log.java
karesti: I used some in tests

lock/src/main/java/org/infinispan/lock/impl/functions/UnlockFunction.java
karesti: 👍

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
```diff
@@ -221,34 +255,95 @@ public void entryRemoved(CacheEntryRemovedEvent event) {
...
+
+         if (leavingNodes.isEmpty()) {
+            log.tracef("LOCK[%s] Nothing to do, all nodes are present %s", getName(), originator);
+            return;
+         }
+
+         if (leavingNodes.size() >= newMembers.size() && oldMembers.size() > 2) {
+            // If the oldMembers size is greater than 2, we do nothing because the other nodes will handle
+            // If the cluster was formed by 2 members and one leaves, we should not enter here
+            log.tracef("LOCK[%s] Nothing to do, we are on a minority partition notification on %s", getName(), originator);
```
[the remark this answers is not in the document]
karesti: @danberindei so we migth get notified several times for the same split brain ?

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: this is what we discussed before, holding requests to wait the unlock over.

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: retryPendingRequests oui, unlocks no

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: retryPendingUnlocks removed

lock/src/main/java/org/infinispan/lock/impl/log/Log.java
karesti: I used some of them on testing

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
wburns: I was more wondering so we don't try to retry a lock accidentally.

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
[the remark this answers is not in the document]
danberindei: Yes, we can get several view changes for the same network split event. Failure detection is always best-effort, and when with a very lossy/unstable network you could get any combination of view change events you can think of :)

But FD and FD_SOCK have each node verify a single other node (the next node in the members list, I think). So if the network splits between AB and CD, A and B won't even think of verifying the availability of D until they install a view [ABD].

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
danberindei: Ok, I'd just like some comments with the scenario that this is trying to fix.

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
danberindei: Doesn't `retryPendingUnlocks()` unlocking the locks trigger `LockReleasedListener`? Or is the problem that `LockReleasedListener` isn't notified reliably?

lock/src/main/java/org/infinispan/lock/impl/log/Log.java
danberindei: My bad, I was searching for references in the production scope in IntelliJ :)

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: well, I think today the risk is doing more than one rpc call for the same request, indeed.
The listener notification can arrive before the rpc call finishes, so we can retry the same request. I need to check in the listener that the request that acquired the lock is not retried

lock/src/main/java/org/infinispan/lock/impl/lock/ClusteredLockImpl.java
karesti: ok

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
```diff
@@ -1,40 +1,63 @@
...
+      splitCluster(new int[]{0, 1, 2}, new int[]{3, 4, 5});
+
 
-      eventuallyEquals(2, () -> advancedCache(0).getRpcManager().getTransport().getMembers().size());
-      eventuallyEquals(2, () -> advancedCache(1).getRpcManager().getTransport().getMembers().size());
-      eventuallyEquals(2, () -> advancedCache(2).getRpcManager().getTransport().getMembers().size());
-      eventuallyEquals(2, () -> advancedCache(3).getRpcManager().getTransport().getMembers().size());
+      for (int i = 0; i < 6; i++) {
+         blockUntilViewReceived(advancedCache(i), 3);
+//         eventuallyEquals(3, () -> advancedCache(finalI).getRpcManager().getTransport().getMembers().size());
```
danberindei: Actually I meant you could replace the entire loop with `blockUntilViewReceived(cache(0), cache(1), cache(2)); blockUntilViewReceived(cache(3), cache(4), cache(5)`.
But then I looked into how `splitCluster()` works and I don't think any wait is necessary, it installs the GMS view synchronously.

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
karesti: I added this more in case it fails to have an explanation of where it failed. Does this deserve to be removed then?

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
karesti: ok

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
karesti: ok

karesti:
@danberindei I reported your commits. if CI is ok, I would like to rebase the commits, wait for CI and merge

karesti:
@danberindei @wburns build stopped due to memory in the wildfly module on build 20. And now it sais there is no workspace. 
Any idea on how can I fix this pb and make this PR built and merged?

rvansa:
@karesti This happened to some of my PRs as well, @tristantarrant has deleted the workspace manually. So we probably need to do the same for all open PRs.

wburns:
Changes seem fine to me, however the build still isn't in good shape yet :( Not your fault @karesti, I don't think at least :)

We should make sure to fix up the remaining issues in a minor release though.

karesti:
@wburns yes, master is failing too on this 😢

karesti:
build finished with 9 tests in error that are unrelated

karesti:
@danberindei @wburns if you don't see anything additional, can this PR be merged please ? 🙏

PR #5739, opened 2018-02-07
ISPN-8778 Create HeartBeatCommand

pruivo:
https://issues.jboss.org/browse/ISPN-8778

### DIFF
--- a/core/src/main/java/org/infinispan/remoting/inboundhandler/GlobalInboundInvocationHandler.java
+++ b/core/src/main/java/org/infinispan/remoting/inboundhandler/GlobalInboundInvocationHandler.java
@@ -22,6 +22,7 @@ import org.infinispan.remoting.responses.ExceptionResponse;
 import org.infinispan.remoting.responses.Response;
 import org.infinispan.remoting.responses.SuccessfulResponse;
 import org.infinispan.remoting.transport.Address;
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/infinispan/remoting/inboundhandler/GlobalInboundInvocationHandler.java
```diff
@@ -67,7 +68,9 @@ private static ExceptionResponse exceptionHandlingCommand(Throwable throwable) {
    public void handleFromCluster(Address origin, ReplicableCommand command, Reply reply, DeliverOrder order) {
       command.setOrigin(origin);
       try {
-         if (command instanceof CacheRpcCommand) {
+         if (command.getCommandId() == HeartBeatCommand.COMMAND_ID) {
+            reply.reply(null);
```
karesti: I don't like this reply null, seems like an error when reading, but I don't know if I have any other suggestion or alternative to offer

pruivo: it is ok. it is an empty reply. the `DefaultResponseGenerator` does the same trick to avoid serializing a Response :)

tristantarrant:
Merged, thanks

PR #5742, opened 2018-02-09
ISPN-8620 Execute operations concurrently over the same connection

rvansa:
https://issues.jboss.org/browse/ISPN-8620


rvansa:
run performance tests please
cs-hotrod
[UDP 1kB]

rvansa:
run performance tests please
cs-hotrod

rvansa:
Trying out 10kB entries yet...

There are few test failures, the most troublesome is the ClientEventsOOMETest that shows a regression (perf-related - 40GB of events can't be pushed to client within 1 minute) despite the perf-related part of that did not change. I'll be working on that more.

rvansa:
Uuuh, 10kB don't look good at all. I probably need to add some priority queuing into ChannelPool to make sure the sockets are used 'just right'.

rvansa:
run performance tests please
cs-hotrod

rvansa:
Good news, the ClientEventsOOMETest did not show a performance regression; master (nor pre-netty) does not impose timeout on AddClientListenerOperation. Fixed.

rvansa:
run performance tests please
cs-hotrod

slaskawi:
@rvansa Could you please explain what is the main goal here? 

IMO reducing the number of connections isn't worth the effort unless there's a performance gain. 

The only scenario where this makes *a lot* of sense is a `BASIC` mode in HR client.

Could you please share your opinion on that with us?

rvansa:
@slaskawi Performance improvements are one of the goals, though not the main one.

Assume that you have a medium cluster, say 16 nodes. On client side you can't limit the total number of connections*, so you'll limit the number of connections per server to keep, say, at most 1600 concurrent connections - so each server can execute only 100 concurrent operations (from one client). 

In the asynchronous programming model this is not enough, you want to paralelize as much as you can; that's why we're multiplexing on precious resources (connections).

Ofc I don't accept regression, but the 1kB perf ack has shown 10% speedup for reads which is worth it alone. 10kB perf ack showed regression; this can be dealt with.

* Bbecause such algorithm would close connections and open new ones too fast, which would lead to running out of ports as the closed ones would be left in time_wait for another 30 seconds.

rvansa:
run performance tests please
cs-hotrod

slaskawi:
> On client side you can't limit the total number of connections (1)

I though we can by using `ConnectionPoolConfiguration` class. There are some interesting attributes like `maxTotal` and `maxActive` which I though were created for limiting the number of concurrent connections. 

> In the asynchronous programming model this is not enough, you want to paralelize as much as you can; that's why we're multiplexing on precious resources (connections).

I'm really interested in the perf results here. I personally wouldn't guess that the number of connections could be bottleneck here. Let's wait until you fix the regression and then have a look at the results.

BTW, may I kindly ask you to do some tests with `BASIC` client mode? My intuition tells me that we might get a really nice boost there... (BTW, simple JMH benchmark is absolutely fine but let's make it after you finish fighting with perf regression).

rvansa:
@slaskawi We have the `maxTotal` in place since we had blindly exposed what C3P0 provided. But using that led to (1) and and therefore I have deprecated that option with rewrite to Netty, and it is ignored since. The explanation is in the javadoc, too.

What's the point of running on level-1 intelligence? Just to have some estimate for OpenShift deployment?

slaskawi:
> We have the maxTotal in place since we had blindly exposed what C3P0 provided. But using that led to (1) and and therefore I have deprecated that option with rewrite to Netty, and it is ignored since. The explanation is in the javadoc, too.

Thanks for the explanation... 

> What's the point of running on level-1 intelligence? Just to have some estimate for OpenShift deployment?

Precisely!

rvansa:
The perf regression is still there :-/

rvansa:
run performance tests please
cs-hotrod

rvansa:
run performance tests please
cs-hotrod

rvansa:
run performance tests please
cs-hotrod
[UDP 1kB]

rvansa:
run performance tests please
cs-hotrod

rvansa:
run performance tests please
cs-hotrod
[UDP 1kB]

rvansa:
run performance tests please
cs-hotrod

diegolovison:
run performance tests please
cs-hotrod

rvansa:
@galderz This is out of scope of this PR; events still use the connection exclusively. I will address events in ISPN-8621 (already working on).
I need to deliver stuff in chunks, people tend to keep 5+k-line PRs in the queue...

rvansa:
Hmm, `org.infinispan.server.test.client.hotrod.HotRodRemoteStreamingIT.testBasicFunctionality` failure is probably related.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/HeaderDecoder.java
```diff
@@ -1,79 +1,152 @@
...
+   public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+      if (evt instanceof ChannelPoolCloseEvent) {
+         closing = true;
+         allCompleteFuture().whenComplete((nil, throwable) -> {
+            ctx.channel().close();
+         });
+      } else if (evt instanceof IdleStateEvent) {
+         // If we have incomplete operations this channel is not idle!
+         if (!incomplete.isEmpty()) {
+            return;
```
wburns: Curious, should we possibly complete exceptionally the operations saying there was a timeout or something? Or this should never happen if there are any incomplete?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/GetStreamOperation.java
```diff
@@ -75,9 +68,13 @@ public ChannelInputStream decodePayload(ByteBuf buf, short status) {
          int totalLength = ByteBufUtil.readVInt(buf);
          VersionedMetadataImpl versionedMetadata = new VersionedMetadataImpl(creation, lifespan, lastUsed, maxIdle, version);
 
-         ChannelInputStream stream = new ChannelInputStream(versionedMetadata, () -> channelFactory.releaseChannel(channel), totalLength);
-         channel.pipeline().addLast(stream);
-         return stream;
+         ChannelInputStream stream = new ChannelInputStream(versionedMetadata,  () -> {
```
wburns: Nitpick: braces are not needed

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/HeaderDecoder.java
rvansa: No, we shouldn't do that; Idle timeout can be lower than socket timeout (which was defining how long the operation can take with synchronous operations and is used for timeout for each operation). If there wasn't this check the channel would be closed and the operations would complete exceptionally as you suggest, but we don't want to fail the operation just because the server is taking somewhat longer. There's even a test that lowers the idle timeout ridiculously.

rvansa:
@wburns Testsuite looks good, anything else?

rvansa:
@wburns Any reason this is still in the queue?

wburns:
Overlooked - so many PRs :D We are fine with this being in 9.2.1 then?

rvansa:
@wburns IIRC there are no API/protocol changes.

wburns:

PR #5744, opened 2018-02-09
ISPN-8634 Wildfly 11 fails to start with Infinispan 9.x modules insta…

ryanemerson:
…lled

https://issues.jboss.org/browse/ISPN-8634

### DIFF
--- a/integrationtests/wildfly-modules/pom.xml
+++ b/integrationtests/wildfly-modules/pom.xml
@@ -143,7 +143,6 @@
       <dependency>
          <groupId>org.infinispan.server</groupId>
          <artifactId>infinispan-server-infinispan</artifactId>
-         <version>${project.version}</version>
          <scope>test</scope>
       </dependency>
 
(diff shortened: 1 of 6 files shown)


danberindei:
@ryanemerson the discussion in JIRA comments mentions the `org.jboss.sasl` module, but this PR doesn't do anything about it. Maybe the JIRA reference is wrong?

danberindei:
Also, could we expand the `integrationtests/wildfly-modules` to cover the problem?

ryanemerson:
@danberindei The JIRA is slightly behind as it was created before we updated to wildfly 11. The reason that jboss-sasl was missing before, was because wildfly 11 is based upon on Elytron and does not include the jboss-sasl module anymore. Now the issue is that the wildfly-modules were not updated to reflect the CounterManager depdencies as stated in the server feature-pack, resulting in wildfly 11 deployment still failing. 

I have extended InfinispanExtensionIT so that it will fail if either the hotrod or endpoint connectors fail to load for any reason.

ryanemerson:
Failures don't appear related

tristantarrant:
Merged, thanks

PR #5746, opened 2018-02-09
ISPN-8792 EntryWrappingInterceptorDoesNotBlockTest random failures

rvansa:
https://issues.jboss.org/browse/ISPN-8792

### DIFF
--- a/core/src/test/java/org/infinispan/tx/EntryWrappingInterceptorDoesNotBlockTest.java
+++ b/core/src/test/java/org/infinispan/tx/EntryWrappingInterceptorDoesNotBlockTest.java
@@ -50,6 +50,11 @@ import org.infinispan.util.concurrent.TimeoutException;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
+/**
+ * We make node 2 a backup owner for segment 0, but we block state transfer so that it doesn't have any entries.
+ * We verify that when the prepare command is invoked remotely on node 2, a remote get is sent to nodes 0 and 1,
+ * and that the remote thread is not blocked while waiting for the remote get responses.
+ */
 @Test(groups = "functional", testName = "tx.EntryWrappingInterceptorDoesNotBlockTest")
 @CleanupAfterMethod
 public class EntryWrappingInterceptorDoesNotBlockTest extends MultipleCacheManagersTest {
@@ -110,7 +115,6 @@ public class EntryWrappingInterceptorDoesNotBlockTest extends MultipleCacheManag
 
    protected void test(int expectRemoteGets, BiFunction<MagicKey, Integer, Object> operation, MagicKey... keys) throws Exception {
       ControlledRpcManager crm0 = ControlledRpcManager.replaceRpcManager(cache(0));
-      ControlledRpcManager crm1 = ControlledRpcManager.replaceRpcManager(cache(1));
       ControlledRpcManager crm2 = ControlledRpcManager.replaceRpcManager(cache(2));
       CountDownLatch topologyChangeLatch = new CountDownLatch(2);
       cache(0).addListener(new TopologyChangeListener(topologyChangeLatch));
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/infinispan/tx/EntryWrappingInterceptorDoesNotBlockTest.java
```diff
@@ -142,11 +143,16 @@ protected void test(int expectRemoteGets, BiFunction<MagicKey, Integer, Object>
...
 
       // the node should load all moving keys
+      List<ControlledRpcManager.SentRequest> remoteGets = new ArrayList<>();
       for (int i = 0; i < expectRemoteGets; i++) {
-         crm2.expectCommand(ClusteredGetCommand.class).send().receiveAll();
+         remoteGets.add(crm2.expectCommand(ClusteredGetCommand.class).send());
       }
 
       prepareExpectingInterceptor.await();
+      // we could get an exception if the remote get finished before the prepare finishes
```
danberindei: why?

rvansa: Because the PrepareExpectingInterceptor expects that the invocationstage is receives is not complete. If the remote get is completed before the interceptor gets the stage, the stage can get completed, too, and an assertion fires.

It was not complicated to get the failure when running the test locally, try yourselves on master.

danberindei: @rvansa I couldn't reproduce it locally, that's why I'm asking :)

I would expect the comment to explain why the invocation stage for the prepare command depends on whether the previous remote gets have finished. The JIRA should also have more information, or at least have a stack trace.

rvansa: Ok, I was sending the fix Friday midnight so I didn't bother :) I'll fill description in the JIRA.

danberindei: In my experience the late night fixes need more comments for the others to understand, not the other way around :)

rvansa: @danberindei I've added description to the JIRA, integration is just two clicks away!

danberindei: +1 for the JIRA description, but I still don't think the java comment explains anything... there's no reason why the comments in the code couldn't explain what the test is trying to achieve and how it wants to do it.

rvansa:
@danberindei Are the comments more understandable now?

tristantarrant:
Quite a few failures on EntryWrappingInterceptorDoesNotBlockTest https://ci.infinispan.org/job/Infinispan/job/PR-5746/2/#showFailuresLink

rvansa:
I've sent a group of well trained monkeys to investigate these...

danberindei:
@rvansa Sorry for being so picky, but now I think there's too much detail :)

I think what made the test harder for me to understand the first (and second) time I read it was that I got the node where PrepareExpectingInterceptor runs wrong (0 instead of 2), so I was expecting it to verify how the responses to the PrepareCommand are handled. The fact that I didn't remember exactly when the remote get commands are sent and that the comment talked about "the node" didn't help :)

Instead of piecemeal comments on every line, I would have liked to have a "mission statement" at the start of the test instead to go back to when things stopped making sense, something like this:

```
We make node 2 a backup owner for segment 0, but we block state transfer so that it doesn't have any entries. 
We verify that when the prepare command is invoked remotely on node 2, a remote get is sent to nodes 0 and 1, and that the remote thread is not blocked while waiting for the remote get responses.
```

rvansa:
@danberindei I am OK with adding the mission statement but I won't remove any comments unless wrong.

Still looking into the failures.

rvansa:
The monkeys did a good job, let's verify in CI.

wburns:
There are some checkstyle errors now.

rvansa:
Removed unused imports... It would be handy if CI sent a mail when the PR cannot be built.

wburns:

PR #5747, opened 2018-02-11
Switched some orElse to orElseGet

softarn:
Some orElse didn't use constants but created objects. Didn't seem to cause any problems however so this is just a small optimization to not create objects unless they are actually needed.

### DIFF
--- a/server/hotrod/src/main/java/org/infinispan/server/hotrod/iteration/DefaultIterationManager.java
+++ b/server/hotrod/src/main/java/org/infinispan/server/hotrod/iteration/DefaultIterationManager.java
@@ -178,7 +178,7 @@ public class DefaultIterationManager implements IterationManager {
    }
 
    private Object[] unmarshallParams(byte[][] params, Object factory) {
-      Marshaller m = marshaller.orElse(MarshallerBuilder.genericFromInstance(Optional.of(factory)));
+      Marshaller m = marshaller.orElseGet(() -> MarshallerBuilder.genericFromInstance(Optional.of(factory)));
       try {
          Object[] objectParams = new Object[params.length];
          int i = 0;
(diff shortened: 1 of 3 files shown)


rvansa:
ok to test

rvansa:
@softarn Thanks for the contribution.

In general it could make sense to avoid lazy retrieval since the lambda is capturing so it does an allocation (and also uses about 1.5kB in metaspace) but in these cases it seems to avoid some allocations, so the net effect should be beneficial.

softarn:
That's interesting, thanks! Do you have any good source of where or how to learn about the more in depth stuff about Java? Like the allocation you mention. 

I'll fix the failing build.

rvansa:
@softarn Not really a single resource, rather an experience that allocations matter. While allocations themselves are super fast, garbage collection can become a bottleneck under heavy load.

If you want to do some reading about performance, I can recommend reading some of these:
https://shipilev.net/jvm-anatomy-park/
http://psy-lob-saw.blogspot.cz/
https://mechanical-sympathy.blogspot.cz/

danberindei:
In these cases the code with `orElse()` just looks wrong: there's no reason why `MarshallerBuilder.fromClass()` should create an instance with the ClassLoader constructor *and* with the no-args constructor *and* with `genericFromInstance()`.

I doubt there are enough invocations to affect performance in any way, but we could switch to old-school for/if if you think we've finally reached the point where we have too many lambda-generated classes in metaspace...

rvansa:
Integrated, thanks.

softarn:
I was away for a while, thanks for the interesting discussion!

For me it just looks weird to have anything else than a variable or a constant in the `orElse` method. Might be easy to put some logging or something that hits hard on the performance without knowing that the method is called even though the returned cable isn't used.

Thanks for the links. I'll read up on this a bit more!

PR #5753, opened 2018-02-13
ISPN-8813 Avoid NPE during BaseJPAFilterIndexingServiceProvider.stop()

anistor:
https://issues.jboss.org/browse/ISPN-8813

There are numerous NPEs in BaseJPAFilterIndexingServiceProvider.stop() in CI after previous tests have failed. So this will hopefully help reduce the noise.

### DIFF
--- a/query/src/main/java/org/infinispan/query/dsl/embedded/impl/BaseJPAFilterIndexingServiceProvider.java
+++ b/query/src/main/java/org/infinispan/query/dsl/embedded/impl/BaseJPAFilterIndexingServiceProvider.java
@@ -12,7 +12,6 @@ import java.util.concurrent.ConcurrentMap;
 import org.infinispan.encoding.DataConversion;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.interceptors.locking.ClusteringDependentLogic;
-import org.infinispan.marshall.core.EncoderRegistry;
 import org.infinispan.notifications.Listener;
 import org.infinispan.notifications.cachelistener.CacheEntryListenerInvocation;
 import org.infinispan.notifications.cachelistener.CacheNotifier;
@@ -53,7 +52,8 @@ public abstract class BaseJPAFilterIndexingServiceProvider implements FilterInde
    private ClusteringDependentLogic clusteringDependentLogic;
 
    @Inject
-   protected void injectDependencies(CacheNotifier cacheNotifier, ClusteringDependentLogic clusteringDependentLogic, EncoderRegistry encoderRegistry) {
+   @SuppressWarnings("unused")
+   protected void injectDependencies(CacheNotifier cacheNotifier, ClusteringDependentLogic clusteringDependentLogic) {
       this.cacheNotifier = (CacheNotifierImpl) cacheNotifier;
       this.clusteringDependentLogic = clusteringDependentLogic;
    }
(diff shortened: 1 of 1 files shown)


query/src/main/java/org/infinispan/query/dsl/embedded/impl/BaseJPAFilterIndexingServiceProvider.java
```diff
@@ -65,16 +65,18 @@ public void start() {
...
+         cacheNotifier.getListenerCollectionForAnnotation(CacheEntryActivated.class).removeAll(invocations);
+         cacheNotifier.getListenerCollectionForAnnotation(CacheEntryCreated.class).removeAll(invocations);
+         cacheNotifier.getListenerCollectionForAnnotation(CacheEntryInvalidated.class).removeAll(invocations);
+         cacheNotifier.getListenerCollectionForAnnotation(CacheEntryLoaded.class).removeAll(invocations);
+         cacheNotifier.getListenerCollectionForAnnotation(CacheEntryModified.class).removeAll(invocations);
+         cacheNotifier.getListenerCollectionForAnnotation(CacheEntryPassivated.class).removeAll(invocations);
+         cacheNotifier.getListenerCollectionForAnnotation(CacheEntryRemoved.class).removeAll(invocations);
+         cacheNotifier.getListenerCollectionForAnnotation(CacheEntryVisited.class).removeAll(invocations);
+         cacheNotifier.getListenerCollectionForAnnotation(CacheEntriesEvicted.class).removeAll(invocations);
+         cacheNotifier.getListenerCollectionForAnnotation(CacheEntryExpired.class).removeAll(invocations);
```
danberindei: Couldn't you use `CacheNotifier.removeListener()` instead?

query/src/main/java/org/infinispan/query/dsl/embedded/impl/BaseJPAFilterIndexingServiceProvider.java
```diff
@@ -65,16 +65,18 @@ public void start() {
...
-      cacheNotifier.getListenerCollectionForAnnotation(CacheEntryCreated.class).removeAll(invocations);
-      cacheNotifier.getListenerCollectionForAnnotation(CacheEntryInvalidated.class).removeAll(invocations);
-      cacheNotifier.getListenerCollectionForAnnotation(CacheEntryLoaded.class).removeAll(invocations);
-      cacheNotifier.getListenerCollectionForAnnotation(CacheEntryModified.class).removeAll(invocations);
-      cacheNotifier.getListenerCollectionForAnnotation(CacheEntryPassivated.class).removeAll(invocations);
-      cacheNotifier.getListenerCollectionForAnnotation(CacheEntryRemoved.class).removeAll(invocations);
-      cacheNotifier.getListenerCollectionForAnnotation(CacheEntryVisited.class).removeAll(invocations);
-      cacheNotifier.getListenerCollectionForAnnotation(CacheEntriesEvicted.class).removeAll(invocations);
-      cacheNotifier.getListenerCollectionForAnnotation(CacheEntryExpired.class).removeAll(invocations);
+      if (cacheNotifier != null) {
```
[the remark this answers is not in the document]
anistor: They are started here https://github.com/anistor/infinispan/blob/1c26fe427df0422467dfa8de73929feb1eba2890/core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifierImpl.java#L247 and stopped here https://github.com/anistor/infinispan/blob/1c26fe427df0422467dfa8de73929feb1eba2890/core/src/main/java/org/infinispan/notifications/cachelistener/CacheNotifierImpl.java#L259

If one of them fails due to dependencies or fails during start it cause a complete failure for the remaining providers which do not get to be injected or started at all. And all of them still get stop() invoked later. So yes, that should be fixed in CacheNotifierImpl rather than BaseJPAFilterIndexingServiceProvider .

query/src/main/java/org/infinispan/query/dsl/embedded/impl/BaseJPAFilterIndexingServiceProvider.java
```diff
@@ -53,7 +52,8 @@
    private ClusteringDependentLogic clusteringDependentLogic;
 
    @Inject
-   protected void injectDependencies(CacheNotifier cacheNotifier, ClusteringDependentLogic clusteringDependentLogic, EncoderRegistry encoderRegistry) {
+   @SuppressWarnings("unused")
```
danberindei: This shouldn't be necessary, you can configure IntelliJ to consider all methods with the `@Inject` annotation as used.

gustavonalle:
Forgot about this one @anistor ?

ryanemerson:
@anistor What's the current state of this PR, are any more changes required? The latest failures are not related

anistor:
This does not seem to fix ISPN-8813, only reduces the likelyhood of the NPE. And does not fully avoid it. And the real issue is that I could not find the REAL cause of the NPE. So let's postpone this. I'll close the PR and move the jira to 9.4

anistor:
Reopened here: https://github.com/infinispan/infinispan/pull/6331

PR #5758, opened 2018-02-15
Ispn 8790 distributed iterator performance tweaks

wburns:
https://issues.jboss.org/browse/ISPN-8790
https://issues.jboss.org/browse/ISPN-8806

Most of these changes are about trying to remove as many wrappers and needless invocations when using an iterator. Also a lot of places using a spliterator will give better performance than using an iterator.

The other big change is that subscriptions are now performed on the async executor. This way the user thread doesn't incur the cost of marshalling and sending a request and can go back to returning entries much faster.

There was also a bug found in requesting a new batch that could cause it to request 2x batch size on accident and is fixed in the ISPN-8806 commit.

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/Closeables.java
+++ b/commons/src/main/java/org/infinispan/commons/util/Closeables.java
@@ -38,6 +38,9 @@ public class Closeables {
     * @return A spliterator that does nothing when closed
     */
    public static <T> CloseableSpliterator<T> spliterator(Spliterator<T> spliterator) {
+      if (spliterator instanceof CloseableSpliterator) {
+         return (CloseableSpliterator<T>) spliterator;
+      }
       return new SpliteratorAsCloseableSpliterator<>(spliterator);
    }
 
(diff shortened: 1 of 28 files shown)


wburns:
Performance table can be found at https://docs.google.com/spreadsheets/d/18v8e6vG-4aX8Pk-ihB3p1H5cQvsSQlFMhQSKxPhSLf0/edit#gid=0

gustavonalle:
I observed the Spark Connector suite, that uses remote iteration a lot, has a perf degradation starting with 9.2.0.CR2, I opened https://issues.jboss.org/browse/ISPN-8822 to track it.

I test with this PR and the regression is still there. I'm about to embark in a bissect journey to see which commit caused it, but since the connector relies heavily in the remote iterator, though you may have some ideas

wburns:
This PR unfortunately only deals with remote distributed iterators. The spark connector uses the local segment only iteration [1]. There may have been something broken there possibly. The code should be mostly the same though, I wonder if something simple broke it.

I can run the test we have in core to make sure that is still working properly real quick.

[1] https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java#L634

gustavonalle:
indeed, thanks @wburns. I'll let you know when I find the commit that caused it.

wburns:
I can confirm that the local segment only iteration is still working as it should. In case if you are curious about the issue that Gustavo mentioned it can be found at https://issues.jboss.org/browse/ISPN-8822

tristantarrant:
I assume the expiration test failures are unrelated ?

wburns:
Not necessarily. Let me look at them. I thought I had a clean core run last time... weird

wburns:
Okay yeah it is random, there is already https://issues.jboss.org/browse/ISPN-8602 for it.

tristantarrant:
Merged, thanks

PR #5763, opened 2018-02-15
ISPN-8829 LockedStream doesn't override BaseCacheStream methods

wburns:
* Added methods to interface

https://issues.jboss.org/browse/ISPN-8829

### DIFF
--- a/core/src/main/java/org/infinispan/LockedStream.java
+++ b/core/src/main/java/org/infinispan/LockedStream.java
@@ -3,6 +3,7 @@ package org.infinispan;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Set;
 import java.util.Spliterator;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/infinispan/LockedStream.java
```diff
@@ -135,6 +136,42 @@ default void forEach(SerializableBiConsumer<Cache<K, V>, ? super CacheEntry<K, V
       return invokeAll((BiFunction<Cache<K, V>, ? super CacheEntry<K, V>, R>) biFunction);
    }
 
+   /**
+    * {@inheritDoc}
+    */
+   @Override
+   LockedStream sequentialDistribution();
```
tristantarrant: All return types need to be in the generic form LockedStream<K, V> to preserve type information in chained invocations

wburns: Oops I didn't mean to forget those 👍

wburns:
Fixed and also updated the old methods that had it incorrect as well.

tristantarrant:
Looks good, waiting for CI to complete

tristantarrant:
Merged, thanks

PR #5765, opened 2018-02-16
ISPN-8822 Performance degradation when running Spark jobs

gustavonalle:
https://issues.jboss.org/browse/ISPN-8822

Some numbers from my local environment:

| Method| 9.2.0.CR1 (ms) | 9.2.0.CR2 (ms)| 
| --- |---| ---|
| ingestion|6873|25323|
| iteration|377|2570|

The iteration method is a single client/single server iterating through all entries and closing the remote cache manager (like the Spark connector does). As explained in the JIRA, CR2 adds a 2s constant delay in closing the cache manager, so this should account for most of the discrepancy

With relation to data ingestion, was not expecting such a difference...

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
@@ -24,6 +24,7 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -210,7 +211,7 @@ public class ChannelFactory {
    public void destroy() {
       try {
          channelPoolMap.values().forEach(ChannelPool::close);
-         eventLoopGroup.shutdownGracefully().sync().getNow();
+         eventLoopGroup.shutdownGracefully(0, 0, TimeUnit.MILLISECONDS).get();
          executorService.shutdownNow();
       } catch (Exception e) {
          log.warn("Exception while shutting down the connection pool.", e);
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/stress/IterationStressTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/stress/IterationStressTest.java
@@ -1,0 +1,106 @@
+package org.infinispan.client.hotrod.stress;
+
+import org.infinispan.client.hotrod.RemoteCache;
+import org.infinispan.client.hotrod.RemoteCacheManager;
+import org.infinispan.client.hotrod.test.SingleHotRodServerTest;
+import org.infinispan.commons.util.CloseableIterator;
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.test.fwk.TestCacheManagerFactory;
+import org.testng.annotations.Test;
+
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.IntStream;
+
+import static org.infinispan.server.hotrod.test.HotRodTestingUtil.hotRodCacheConfiguration;
+
+/**
+ * Simple stress test for remote iteration of entries.
+ * By default, it runs the server and the client in the same VM; when the sys prop -DserverHost=server
+ * is present it will connect to that server instead.
+ */
+@Test(groups = "stress", testName = "org.infinispan.client.hotrod.stress.IterationStressTest")
+public class IterationStressTest extends SingleHotRodServerTest {
+
+    private static final int NUM_ENTRIES = 500_000;
+    private static final String SERVER_HOST = "serverHost";
+    public static final int THREADS = Runtime.getRuntime().availableProcessors();
+
+    private RemoteCache<Object, Object> remoteCache;
+
+    @Override
+    protected void setup() throws Exception {
+        String serverHost = System.getProperty(SERVER_HOST);
+        org.infinispan.client.hotrod.configuration.ConfigurationBuilder builder =
+                new org.infinispan.client.hotrod.configuration.ConfigurationBuilder();
+
+        if (serverHost == null) {
+            cacheManager = createCacheManager();
+            hotrodServer = createHotRodServer();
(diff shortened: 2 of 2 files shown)


tristantarrant:
Can the time be divided per-op ?

wburns:
I was able to confirm these numbers on my local box

insertions took 38.8s vs 7.8s !!!
iteration took 279 ms vs 288 ms
closing manager took 2.2s vs 1 ms !!!

vjuranek:
@wburns so you observe such performance drop also in PUTs? Could you be more specific about the test? I wonder why we didn't observe any perf. drop in [perf. test corresponding to offending PR](http://perfrepo.mw.lab.eng.bos.redhat.com/reports/tableComparisonReport/9867). If the delay is in closing RCM, it make sense it wasn't visible in the comparison, but any drop in PUT should be visible there (otherwise it would imply there's some very wrong the tests).

wburns:
I agree @vjuranek, I am very curious as to what is going on as well. @gustavonalle was looking at it more closely. Sll the inserts is doing is 

`
IntStream.range(0, NUM_ENTRIES).parallel().forEach(i -> remoteCache.put(i, i));
`

So in my case this will use 8 threads (4 core HT) in parallel doing these inserts. I noticed that for CR1 the entire CPU was only about half utilized but with master it was almost 100% the entire time.

I haven't looked but does the 10 and 100 thread write tests are performed on the same RemoteCache instance or do they use separate ones?

tristantarrant:
What happens if you use a pre-Netty Hot Rod client ?

gustavonalle:
@tristantarrant the pre-netty HR client is CR1

gustavonalle:
There is something about using FJPool in the test that interferes with the netty HR client making it a lot slower. Replacing the data ingestion by simply using a fixed threadpool executor is much faster (< 10s)

wburns:
Yeah I can confirm looking at this that something is going on with the ForkJoinPool. In CR1 it only ever uses 8 threads, however in master it was using 31 or more threads. My guess is that something in netty is using a ManagedBlocker, which is causing additional fork join threads to be spawned.

wburns:
It looks like this is caused by the use of `CompletableFuture.get()` - which our netty client uses for all of its blocking methods. Internally when a blocking get is performed in a FJP, CompletableFuture will use a MangedBlocker to spawn a new thread in that FJP to ensure that the CPU is used fully. It does a good job with that as I saw it was almost 100% constantly. The problem is causes so many threads to be started and restarted and context switching that it takes up a ton of extra time.

wburns:
This essentially means instead of invoking on 8 threads it will run as many concurrent puts until 1 finishes (in my box it was able to do over 30 each time). If you had a dedicated box for the client you you probably could have gotten thousands of writes going concurrently :)

rvansa:
@wburns Great findings! I am not sure how should we deal with that, though.
1) Detect that we're running in FJP and trick it to block?
2) Change the test to use non-FJP executor?
3) Limit the concurrency internally (only in FJP/as a configuration setting)?

gustavonalle:
> 2. Change the test to use non-FJP executor?

Or alternatively change it to use non-blocking variants (```putAsync```)

galderz:
> It looks like this is caused by the use of CompletableFuture.get() - which our netty client uses for all of its blocking methods. Internally when a blocking get is performed in a FJP, CompletableFuture will use a MangedBlocker to spawn a new thread in that FJP to ensure that the CPU is used fully. It does a good job with that as I saw it was almost 100% constantly. The problem is causes so many threads to be started and stopped and context switching that it takes up a ton of extra time.

Just a note: I might be wrong but this is one of the reasons why @cescoffier and co are not keen on `CompletableFuture`s. In hindsight, we should have made async methods returning 1 result return `CompletionStage`, which is an interface giving us more flexibility. With `CompletionStage` you can't call get() but I guess you could always mimic it somehow.

rvansa:
@galderz So what are the other cons? (link?)
I don't think that `CompletionStage` alone would work; we would need to provide an extension to get the synchronous variant in a fluent API.

gustavonalle:
Upated, removed the FJPool and removed the delay in the RCM.close(). 

New numbers:

| Method| 9.2.0.CR1 (ms) |  ISPN-8822/spark_perf (ms)| 
| --- |---| ---|
| ingestion|6120|7687|
| size|46|45|
| iteration|331|364|
| close|1|5|

gustavonalle:
Failures not related

galderz:
> @galderz So what are the other cons? (link?)

Dunno, @cescoffier or @tsegismont would know more.

> I don't think that CompletionStage alone would work; we would need to provide an extension to get the synchronous variant in a fluent API.

Not sure what you mean exactly by extension. You mean inheritance or composition? Or none in particular? 

Having separate blocking/non-blocking APIs probably makes sense. Normally you use one or the other, I don't see them being mixed up by in the same project. E.g. if I'm a vert.x user, I exclusively care about non-blocking APIs.

Even if this is not true, you can always have a helper static method to take a `CompletionStage` and wait for it, without the need to extend the interface :).

cescoffier:
One of the issues of CompletableFuture is the behavior when the parallelization level is < 2 (basically in containers). In this case, it does spawn a new thread for every task with the risk to explode the container limit (and get your container killed).

rvansa:
@cescoffier So the problem is not CF itself but rather FJP, right? I've had issues with it in the past (memory leaks in threadlocals in FJ threads), so thread explosion might be another one.
The takeaway for me is always providing the `CF...Async` methods with an executor (to not use FJP accidentally). Parallel streams might be more complex to get right. Ofc we can't change it in the libraries.

danberindei:
@rvansa there's another issue with CF.get()/CF.join(), and the reason we have `CompletableFutures.await()`: it spins for a while using `ThreadLocalRandom` before blocking.

I also prefer using `CompletionStage` both internally and in APIs, if only to discourage blocking on it. OTOH callers who want to block will probably default to using `toCompletableFuture()` instead of findin `CompletableFutures.await()`, which means they'll spin (and use `ManagedBlocker`) anyway.

gustavonalle:
@rvansa Even not using the FJPool in the test, there is still a ~20% slowdown when doing multi-threaded PUT compared to CR1, is that expected?

rvansa:
@gustavonalle No. I'll grab some JFRs from the test...

cescoffier:
@rvansa yes, if you always pass an executor service, you are safe (better not to use too many threads, but well this is another issue).

Also, notice that `CompletableFuture` has a few drawbacks/limitations:
* it's a Future / Promise API. It is not well suited for streams. But I believe you are fine in your PR. It may be a bit limited around continuous queries
* `CompletableFuture.get` and `CompletableFuture.join` are blocking. Not sure you want to let your user access to these methods (your call)

galderz:
> it's a Future / Promise API. It is not well suited for streams. But I believe you are fine in your PR. It may be a bit limited around continuous queries.

Cont. queries is the territory of Flowable, we wouldn't use CF for that.

gustavonalle:
This PR can be reviewed and merged "as is", since it removes the 2s wait when closing the cache manager that hurts small spark jobs.  We can carry on the perf work in separate JIRAs

tristantarrant:
Merged

PR #5769, opened 2018-02-19
ISPN-8820 Don't allow lower-bound equals to upper-bound

pruivo:
https://issues.jboss.org/browse/ISPN-8820


tristantarrant:
Merged, thanks

PR #5770, opened 2018-02-19
ISPN-8843 Workaround for the IDE compiler

gustavonalle:
https://issues.jboss.org/browse/ISPN-8843

### DIFF
--- a/server/core/pom.xml
+++ b/server/core/pom.xml
@@ -23,6 +23,22 @@
          <groupId>io.netty</groupId>
          <artifactId>netty-handler</artifactId>
       </dependency>
+      <!-- Workaround for ISPN-8843 -->
+      <dependency>
+         <groupId>org.conscrypt</groupId>
+         <artifactId>conscrypt-openjdk</artifactId>
+         <version>${version.netty-conscrypt-optional}</version>
+         <classifier>${os.detected.classifier}</classifier>
+         <optional>true</optional>
+      </dependency>
+      <!-- Workaround for ISPN-8843 -->
+      <dependency>
+         <groupId>io.netty</groupId>
+         <artifactId>netty-tcnative</artifactId>
+         <version>${version.netty-tcnative-boringssl-static}</version>
+         <classifier>${os.detected.classifier}</classifier>
+         <optional>true</optional>
+      </dependency>
       <dependency>
          <groupId>io.netty</groupId>
          <artifactId>netty-transport-native-epoll</artifactId>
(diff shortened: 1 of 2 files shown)


gustavonalle:
@karesti Kind of what you suggested earlier, but making it optional and adding a build extension to obtain the os

karesti:
@gustavonalle it works, for me should be merged til we find another solution
@slaskawi any additional thoughts ?

gustavonalle:
CI is fine

server/core/pom.xml
```diff
@@ -23,6 +23,22 @@
          <groupId>io.netty</groupId>
          <artifactId>netty-handler</artifactId>
       </dependency>
+      <!-- Workaround for ISPN-8843 -->
+      <dependency>
+         <groupId>org.conscrypt</groupId>
+         <artifactId>conscrypt-openjdk</artifactId>
+         <version>${version.netty-conscrypt-optional}</version>
+         <classifier>${os.detected.classifier}</classifier>
```
rvansa: Shouldn't this be in test scope?

gustavonalle: let me try it and see it Intellij is happy

slaskawi: It doesn't really matter since this is an `<optional>true</optional>` dependency.

server/core/pom.xml
```diff
@@ -59,6 +75,14 @@
    </dependencies>
 
    <build>
+      <!-- Workaround for ISPN-8843 -->
+      <extensions>
+         <extension>
+            <groupId>kr.motd.maven</groupId>
+            <artifactId>os-maven-plugin</artifactId>
+            <version>1.4.1.Final</version>
```
slaskawi: Version should probably be in the BOM or parent pom.

server/core/pom.xml
gustavonalle: it doesn't work

gustavonalle:
> Let's make a deal than - let's wait a couple of days (a week?)

IMO we should integrate this now and revert when (if) they fix it.

server/core/pom.xml
gustavonalle: Let's not make this workaround official :)

gustavonalle:
>  But I do understand that you guys want to import everything into IntelliJ and just run some tests...

I didn't know you could import only certain modules in Intellij. I usually just go to File -> Open and it materializes the maven/sbt/gradle model

slaskawi:
@gustavonalle Yes you can! Just open whatever project you need (or start with an empty one), bring up Maven window and point `client/hotrod-client/pom.xml`. This way you will import only one module. I've always been working that way - IntelliJ indexing is killing me so I open up only those modules that I'm working on.

karesti:
@slaskawi @gustavonalle I'm with Gustavo. We merge and revert after. It's not a big deal this modification in the pom and the problem is super annoying

server/core/pom.xml
gustavonalle: done

galderz:
Merged, this is a roadblock for all of us. Create follow up JIRA(s) for any leftover work.

karesti:
@galderz thank you !!!

PR #5774, opened 2018-02-20
ISPN-8847 createCache is blocking, should wait for return

galderz:


galderz:
Pushed a fix

wburns:

PR #5775, opened 2018-02-20
ISPN-8854 NullPointerException in SingleFileStore

pruivo:
https://issues.jboss.org/browse/ISPN-8854

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
+++ b/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
@@ -14,7 +14,6 @@ import static org.infinispan.persistence.manager.PersistenceManager.AccessMode.B
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -86,6 +85,8 @@ import org.infinispan.util.concurrent.WithinThreadExecutor;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
+import net.jcip.annotations.GuardedBy;
+
 public class PersistenceManagerImpl implements PersistenceManager {
 
    private static final Log log = LogFactory.getLog(PersistenceManagerImpl.class);
@@ -101,26 +102,29 @@ public class PersistenceManagerImpl implements PersistenceManager {
    @Inject private ByteBufferFactory byteBufferFactory;
    @Inject private MarshalledEntryFactory marshalledEntryFactory;
    @Inject private CacheStoreFactoryRegistry cacheStoreFactoryRegistry;
-   @Inject private ExpirationManager expirationManager;
+   @Inject private ExpirationManager<Object, Object> expirationManager;
 
+   @GuardedBy("storesMutex")
    private final List<CacheLoader> loaders = new ArrayList<>();
+   @GuardedBy("storesMutex")
    private final List<CacheWriter> nonTxWriters = new ArrayList<>();
+   @GuardedBy("storesMutex")
    private final List<TransactionalCacheWriter> txWriters = new ArrayList<>();
    private final ReadWriteLock storesMutex = new ReentrantReadWriteLock();
    private final Map<Object, StoreConfiguration> configMap = new HashMap<>();
-   private AdvancedPurgeListener advancedListener;
+   private AdvancedPurgeListener<Object, Object> advancedListener;
 
    /**
     * making it volatile as it might change after @Start, so it needs the visibility.
     */
-   volatile boolean enabled;
+   private volatile boolean enabled;
    private volatile boolean clearOnStop;
    private boolean preloaded;
 
    @Override
-   @Start(priority = 10)
+   @Start()
    public void start() {
-      advancedListener = new AdvancedPurgeListener(expirationManager);
+      advancedListener = new AdvancedPurgeListener<>(expirationManager);
       preloaded = false;
       enabled = configuration.persistence().usingStores();
       if (!enabled)
@@ -131,9 +135,10 @@ public class PersistenceManagerImpl implements PersistenceManager {
          if (transactionManager != null) {
             xaTx = transactionManager.suspend();
          }
+         storesMutex.readLock().lock();
          try {
 
-            Set undelegated = new HashSet();//black magic to make sure the store start only gets invoked once
+            Set<Lifecycle> undelegated = new HashSet<>();//black magic to make sure the store start only gets invoked once
 
             Consumer<CacheWriter> startWriter = writer -> {
                writer.start();
@@ -169,6 +174,7 @@ public class PersistenceManagerImpl implements PersistenceManager {
             if (xaTx != null) {
                transactionManager.resume(xaTx);
             }
+            storesMutex.readLock().unlock();
          }
       } catch (Exception e) {
          throw new CacheException("Unable to start cache loaders", e);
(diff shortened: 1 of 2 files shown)


danberindei:
@pruivo Please add some more information in JIRA/PR/code comments to make reviewing easier

pruivo:
@danberindei more than this:
```
//simulates the scenario where, concurrently, the cache is stopping and handling a topology update.
```

danberindei:
@pruivo why would handling a topology update while stopping cause problems?

pruivo:
@danberindei because the SFS nullify all the fields on stop but the topology classes still can handle topology updates. in that case, it tries to "remote-stale-data" from the store that will fail because the in memory keys are null, throwing the NPE (as you can see in the stack trace)

@tristantarrant to be honest, I'm not sure how many topology the node can receive after cache.stop() :)

danberindei:
I like the new test much better @pruivo, because the store doesn't/shouldn't know anything about topology updates. 

However, I think it also makes it obvious that we need a test where `stop()` is called while `process()` is iterating through the store. Probably all stores need a test like that, but I concede that a generic store test is out of scope ;)

pruivo:
@danberindei do you want the test in this PR? it may increase its size :)

pruivo:
Updated all the stores. It isn't perfect but should work most of the cases :)

wburns:
You guys just like proving me wrong when I think no one is going to change the stores before my PR https://github.com/infinispan/infinispan/pull/5794 :D

persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
```diff
@@ -325,6 +329,9 @@ private long timeHeaderToSeconds(String header) {
 
    @Override
    public void process(KeyFilter keyFilter, final CacheLoaderTask cacheLoaderTask, Executor executor, boolean loadValue, boolean loadMetadata) {
+      if (!started) {
```
wburns: I don't think this is sufficient to protect the initial key retrieval as it will be propagated to the caller.

persistence/soft-index/src/main/java/org/infinispan/persistence/sifs/SoftIndexFileStore.java
```diff
@@ -443,6 +443,9 @@ private ByteBuffer toBuffer(byte[] array) {
    }
 
    private void forEachOnDisk(boolean readMetadata, boolean readValues, EntryFunctor functor, IntConsumer fileFunctor) throws PersistenceException {
+      if (!started) {
```
wburns: I  have a feeling this one isn't going to be saved either as it nulls out all the values.

persistence/rest/src/main/java/org/infinispan/persistence/rest/RestStore.java
wburns: The test that was added only catches things after it starts producing entries but not before.

wburns: I kinda wonder if we just want to add a semaphore to this, SFS and SIFS instead (like rocks and level db). We could even have a wrapper class that guarantees orderly shutdown (and apply this to user added custom stores as well).

core/src/main/java/org/infinispan/persistence/file/SingleFileStore.java
```diff
@@ -141,13 +149,17 @@ public void start() {
    public void stop() {
       try {
          if (channel != null) {
-            log.tracef("Stopping store %s, size = %d, file size = %d", ctx.getCache().getName(), entries.size(), channel.size());
+            log.tracef("Stopping store %s, size = %d, file size = %d", ctx.getCache().getName(), size(), channel.size());
 
             // reset state
             channel.close();
             channel = null;
```
wburns: Technically it should be possible to get a NPE from the channel being null in the load method (or see it in a closed state).

wburns:
Talking with @pruivo more he found that what is missing is actually a write lock acquisition at
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java#L196

This should make all stores good then.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -178,38 +185,45 @@ public void start() {
...
-            CacheWriter actual = undelegate(writer);
-            actual.stop();
-            undelegated.add(actual);
-         } else {
-            undelegated.add(writer);
+      storesMutex.writeLock().lock();
+      try {
+         // If needed, clear the persistent store before stopping
+         if (clearOnStop) {
+            clearAllStores(AccessMode.BOTH);
```
wburns: Unfortunately this method acquires the read lock which will deadlock. Maybe move the write lock acquisition to just after?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -278,7 +297,16 @@ public void disableStore(String storeType) {
             storesMutex.writeLock().unlock();
          }
 
-         if (loaders.isEmpty() && nonTxWriters.isEmpty() && txWriters.isEmpty()) {
+         boolean noMoreStores;
+
+         storesMutex.readLock().lock();
+         try {
+            noMoreStores = loaders.isEmpty() && nonTxWriters.isEmpty() && txWriters.isEmpty();
```
wburns: Hrmm might as well do this in the write lock above?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -601,16 +634,32 @@ public void setClearOnStop(boolean clearOnStop) {
       this.clearOnStop = clearOnStop;
    }
 
+   //TODO should we make a copy for all getters?
```
wburns: Technically these locks below do nothing at all and we could still get `ConcurrentModificationException`s etc. I would say make a copy like you mentioned. These should be invoked so infrequently it doesn't matter about perf.

core/src/test/java/org/infinispan/persistence/PersistenceManagerTest.java
```diff
@@ -0,0 +1,88 @@
...
+            (marshalledEntry, taskContext) -> {
+               before.countDown();
+               after.await();
+               count.incrementAndGet();
+            },
+            true, true));
+      before.await(30, TimeUnit.SECONDS);
+      Future<Void> stopFuture = fork(persistenceManager::stop);
+      //stop is unable to proceed while the process isn't finish.
+      Exceptions.expectException(TimeoutException.class, () -> stopFuture.get(15, TimeUnit.SECONDS));
```
wburns: 15 seconds seems awfully long.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
pruivo: deadlock where? you can acquire the read-lock with the write-lock acquired.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
pruivo: :+1:

core/src/test/java/org/infinispan/persistence/PersistenceManagerTest.java
pruivo: 10 seconds is fine?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
wburns: Whoops mixed up order. Disregard.

core/src/test/java/org/infinispan/persistence/PersistenceManagerTest.java
wburns: I would say maybe a few hundred milliseconds at most? This means this test will always take 15 seconds to complete currently.

pruivo:
@wburns updated!

wburns:

PR #5776, opened 2018-02-21
ISPN-8856 Exception eviction off heap size calculation always assumes

wburns:
eviction enabled

https://issues.jboss.org/browse/ISPN-8856


tristantarrant:
Merged thanks

PR #5779, opened 2018-02-21
ISPN-8862 BulkOperationsTest reported parameter values change in every run

danberindei:
https://issues.jboss.org/browse/ISPN-8858


pruivo:
looks good to me. integrated! thanks @danberindei

alanfx:
Hey @danberindei, is it possible to get this back-ported to the product too? Changing the parameters means that every time this test case is executed, it looks like a new one which messes with our Polarion reporting. Thank you!

PR #5781, opened 2018-02-22
ISPN-8867 - REST server fails to convert between text formats and octet-stream

diegolovison:


server/integration/testsuite/src/test/java/org/infinispan/server/test/client/rest/AbstractRESTClientIT.java
```diff
@@ -488,4 +488,73 @@ public void testKeyIncludingSlashURLEncoded() throws Exception {
       assertNotNull(get.getHeaders("Last-Modified")[0].getValue());
       assertEquals(APPLICATION_OCTET_STREAM_TYPE, get.getHeaders("Content-Type")[0].getValue());
    }
+
+   @Test
+   public void testIntegerKeysXmlToPlainValues() throws Exception {
```
slaskawi: IMO, this is the wrong place to do it. How about implementing those tests inside `server/rest` module. They execute really fast there.

gustavonalle: Almost totally agree with you, but I think at least one representative test should go in arquillian. I've seen countless features not working in real servers but perfectly fine in embedded hot rod and rest tests.

gustavonalle:
Thanks @diegolovison, I adapted your tests and moved them to rest module in another PR https://github.com/infinispan/infinispan/pull/5790

PR #5785, opened 2018-02-22
ISPN-8853 Registering interposed synchronization fails in async operation with implicit transaction

rvansa:
https://issues.jboss.org/browse/ISPN-8853

I am not providing a test case because TSR is not used in core, but `org.infinispan.server.test.task.DistributedServerTaskIT.shouldPutNewValueInRemoteCache` should not fail with NPE with this fix in.


tristantarrant:
Merged, thanks

PR #5786, opened 2018-02-22
ISPN-8859 DistTopologyChangeUnderLoadSingleOwnerTest takes too long

rvansa:
https://issues.jboss.org/browse/ISPN-8859

### DIFF
--- a/core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
+++ b/core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
@@ -132,6 +132,7 @@ public class DefaultCacheManager implements EmbeddedCacheManager {
    private static final Log log = LogFactory.getLog(DefaultCacheManager.class);
 
    private final ConcurrentMap<String, CompletableFuture<Cache<?, ?>>> caches = CollectionFactory.makeConcurrentMap();
+   private final ConcurrentMap<String, Cache<?, ?>> wiredCaches = CollectionFactory.makeConcurrentMap();
    private final GlobalComponentRegistry globalComponentRegistry;
    private final AuthorizationHelper authzHelper;
    private final DependencyGraph<String> cacheDependencyGraph = new DependencyGraph<>();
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
```diff
@@ -678,7 +679,13 @@ public void start() {
    private void terminate(String cacheName) {
       CompletableFuture<Cache<?, ?>> cacheFuture = this.caches.get(cacheName);
       if (cacheFuture != null) {
```
tristantarrant: Shouldn't we also remove caches from wiredCaches when they are being removed ?

rvansa: The caches are not removed AFAIK. I was about to ask if this isn't a memory leak :)

tristantarrant:
https://issues.jboss.org/browse/ISPN-7191

pruivo:
I guess this changes aren't needed once https://github.com/infinispan/infinispan/pull/5822 is integrated.

core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
danberindei: @pruivo I think @rvansa was saying that we're not removing the caches from the `caches` map either, so one extra reference won't hurt. I have no idea what's happening now if I call `removeCache("x"); getCache("x")`, but I'd modify `removeCache()` to remove the cache from both maps.

danberindei:
I don't think it's safe to call `cache.stop()` while `cache.start()` is still running.

@pruivo if #5822 really fixed the problem then I suggest we close this PR.

pruivo:
@danberindei it doesn't fix it. it just don't star the counter's cache. But there are others caches that may cause problems (e.g. locks)

danberindei:
@pruivo I was thinking of fixing this with https://issues.jboss.org/browse/ISPN-9112, but @anistor's comments and @tristantarrant's experience with listeners on the CONFIG cache pretty much convinced me that changing the default is not feasible at this time.

Still, running `component.start()` and `component.stop()` in parallel makes me uncomfortable, so at the very least I would like to wrap lifecycle method invocations in `synchronized (Component)` blocks.

rvansa:
@danberindei @pruivo Guys let me know if you want this (with minor modifications) in or not. I could fix it next week, but only if it's not going to rot in the queue for much longer.

rvansa:
Apparently there are users unhappy about the current state: https://developer.jboss.org/message/982869

danberindei:
@rvansa Are you sure the forum question is related?

Anyway, I'm still not a fan of running `ComponentRegistry.start()` and `ComponentRegistry.stop()` in parallel. Looking again at the issue description I think if you remove the entry from the `caches` map on cache stop you might be able to skip the `cacheFuture.join()` call in `DCM.undefineConfiguration()`. 

For `terminate()`, I think the best option would be to interrupt the thread that's starting the cache before trying to stop any components. We could try to run `cache.start()` in a `WithinThreadExecutor` and put that future in the `DCM.caches` map.

rvansa:
@danberindei Mea culpa, not reading thoroughly I've mistaken it for another thread revival...

I was checking when was the removal from `caches` dropped and found that the issue I am dealing with here was probably fixed in https://issues.jboss.org/browse/ISPN-8918

I have no reproducer as the counters cache has been fixed as well, so I'll close this PR.

PR #5789, opened 2018-02-22
ISPN-7683 a test 

anistor:


anistor:
This is forwardported from product :)

anistor:
retest this

anistor:
retest this please

rvansa:
Speaking about assertions, I recently came over [AssertJ](http://joel-costigliola.github.io/assertj/) and I must admit that I really like the fluent syntax. The existing support wouldn't cover all cases (e.g. our `eventually*` assertions would need to be implemented) and I wonder why exceptions require `assertThatThrownBy`, but I still like it.

danberindei:
To keep going off-topic, I also like the idea of `AssertJ`, but in order to convince everyone I think someone needs to do a POC (e.g. with `BaseDistFunctionalTest` and its subclasses) to show how complex assertions like `assertOnAllCachesAndOwnership()` would be easier to write and to reuse with a fluent syntax.

WRT `assertThatThrownBy` I think they expect specialized methods like `assertThatIOException()` to be used instead.

server/hotrod/src/test/java/org/infinispan/server/hotrod/test/HotRodServerRegisterUnregisterTopologyCacheTest.java
```diff
@@ -0,0 +1,73 @@
...
+   public void testServerRegisterTopologyCacheOnStart(HotRodServer server, Set<String> internalCaches) {
+      String topologyCacheName = server.getConfiguration().topologyCacheName();
+      assertTrue(internalCaches.contains(topologyCacheName));
+   }
+
+   @Test(dataProvider = "getServers")
+   public void testServerUnregisterTopologyCacheOnStop(HotRodServer server, Set<String> internalCaches) {
+      String topologyCacheName = server.getConfiguration().topologyCacheName();
+      killServer(server);
+      assertFalse(internalCaches.contains(topologyCacheName));
```
danberindei: Why is this assertion after killing the server? The `internalCaches` set is computed only once at the beginning of the test class.

Also, `testServerRegisterTopologyCacheOnStart` cannot run after `testServerUnregisterTopologyCacheOnStop`, so I suggest merging them and `getServers()` into a single method.

slaskawi:
We are already using AssertJ - https://github.com/infinispan/infinispan/blob/master/server/rest/pom.xml#L67.

+1 to make it standard since they abstract the way JUnit and TestNG `assertEquals` are structured. In JUnit you have the expected value as first parameter but in TestNG it's the second one ;)

RT eventually, you can always write custom assertions (see the AssertJ docs).

danberindei:
@slaskawi We already have a solution for the `assertEquals` parameter order problem: never use `org.testng.Assert`, always use `org.testng.AssertJUnit` instead :)

server/hotrod/src/test/java/org/infinispan/server/hotrod/test/HotRodServerRegisterUnregisterTopologyCacheTest.java
anistor: ```internalCaches``` set is live and will reflect the current set of internal caches! It is not computed at the beginning.

anistor: Feel free to change the order of those tests and see how they can run in any order.

rvansa:
...about always:
```
18:54:48@rvansa infinispan$ grep -R -I . -e 'import static org.testng.Assert[^J]' | wc -l
405
```

anistor:
Rebased, and simplified the test so the only test param is the server instance. That should avoid confusion in the future.

danberindei:
@rvansa the rule was always there, but maybe not everyone was aware of it :)

anistor:
I've simplified the test to not use data providers and make it all happen in a single test method. Should  be good now.

danberindei:
Integrated, thanks Adrian!

PR #5790, opened 2018-02-23
ISPN-8867 REST server fails to convert between text formats and octet…

gustavonalle:
…-stream

https://issues.jboss.org/browse/ISPN-8867


server/rest/src/main/java/org/infinispan/rest/dataconversion/XMLTranscoder.java
```diff
@@ -27,7 +34,8 @@
  */
 public class XMLTranscoder extends OneToManyTranscoder {
 
-   protected final static Log logger = LogFactory.getLog(JSONOutputPrinter.class, Log.class);
+   protected final static Log logger = LogFactory.getLog(XMLTranscoder.class, Log.class);
+   private static final SAXParserFactory SAXFACTORY = SAXParserFactory.newInstance();
 
    private static class XStreamHolder {
       static final XStream XStream = new XStream();
```
tristantarrant: Wondering whether we could use the StaxDriver instead of SAX

PR #5791, opened 2018-02-23
ISPN-8870 Remove logger inheritance to reduce class metadata (more)

tristantarrant:
- JCache loggers
- Removed some more unused methods

https://issues.jboss.org/browse/ISPN-8870


Sanne:
merged! thanks

PR #5796, opened 2018-02-26
ISPN-8180 ISPN-8145 Functional API enhancements

rvansa:
* Add key to WriteEntryView
* eval(): pass arbitrary argument

https://issues.jboss.org/browse/ISPN-8180

### DIFF
--- a/core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyValueCommand.java
+++ b/core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyValueCommand.java
@@ -28,17 +28,17 @@ import org.infinispan.metadata.Metadata;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
-public final class ReadWriteKeyValueCommand<K, V, R> extends AbstractWriteKeyCommand<K, V> {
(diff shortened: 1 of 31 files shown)


core/src/main/java/org/infinispan/commands/functional/ReadWriteKeyValueCommand.java
```diff
@@ -28,25 +28,25 @@
...
 
-   private Object value;
-   private BiFunction<V, ReadWriteEntryView<K, V>, R> f;
+   private Object argument;
+   private BiFunction<T, ReadWriteEntryView<K, V>, R> f;
    private Object prevValue;
    private Metadata prevMetadata;
 
-   public ReadWriteKeyValueCommand(Object key, Object value, BiFunction<V, ReadWriteEntryView<K, V>, R> f,
+   public ReadWriteKeyValueCommand(Object key, Object argument, BiFunction<T, ReadWriteEntryView<K, V>, R> f,
```
karesti: why are you renaming value to argument ?

rvansa: Because once we've unlinked type of value-in-cache from the argument we're passing, it makes no sense to call this 'value'. It's rather 'delta', or generally anything.

PR #5799, opened 2018-02-27
ISPN-8888 Wrong warn message for eviction "EvictionConfigurationBuilder

wburns:
ISPN000368: Memory approximation calculation for eviction is unsupported
for the 'OpenJDK 64-Bit Server VM' Java VM"

* Don't warn when OpenJDK VM is used

Wrong warn message for eviction "EvictionConfigurationBuilder ISPN000368: Memory approximation calculation for eviction is unsupported for the 'OpenJDK 64-Bit Server VM' Java VM"


anistor:
Integrated in master. Thanks @wburns !

PR #5805, opened 2018-03-01
ISPN-8897 Indexes from different caches are mixed-up in server mode

anistor:
https://issues.jboss.org/browse/ISPN-8897

### DIFF
--- a/remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProgrammaticSearchMappingProviderImpl.java
+++ b/remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProgrammaticSearchMappingProviderImpl.java
@@ -18,10 +18,13 @@ import org.kohsuke.MetaInfServices;
 @SuppressWarnings("unused")
 public final class ProgrammaticSearchMappingProviderImpl implements ProgrammaticSearchMappingProvider {
 
+   public static final String INDEX_NAME_SUFFIX = "_protobuf";
+
    @Override
    public void defineMappings(Cache cache, SearchMapping searchMapping) {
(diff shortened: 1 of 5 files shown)


remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProgrammaticSearchMappingProviderImpl.java
```diff
@@ -22,6 +22,7 @@
    public void defineMappings(Cache cache, SearchMapping searchMapping) {
       searchMapping.entity(ProtobufValueWrapper.class)
             .indexed()
+            .indexName(cache.getName() + "--" + ProtobufValueWrapperAnalyzerDiscriminator.class.getName())
```
gustavonalle: Why not simply ``` .indexName(cache.getName())```? 
The index name is stored in each of entries of the lucene caches, so I'm a bit worried about large strings

anistor: I just wanted to add some prefix or suffix to help distinguish it from possible other indexes that are user defined. A cache name usually comes from the user's domain so it is a potential conflict source. A shorter ```_protobuf``` suffix could work too.

anistor:
Updated.

anistor:
retest this please

anistor:
retest this please

anistor:
I've added the test from here https://github.com/infinispan/infinispan/pull/5809 to see what CI thinks about this combo.

gustavonalle:
Merged, thanks @anistor !

PR #5807, opened 2018-03-01
Updating doc to show correct method

jmcshane:
The encryption() method appears to have been replaced by ssl()


galderz:
Looks good, thanks @jmcshane :)

PR #5810, opened 2018-03-02
[91x] ISPN-8897 Indexes from different caches are mixed-up in server mode

anistor:
https://issues.jboss.org/browse/ISPN-8897

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/serializing/Serializer.java
+++ b/core/src/main/java/org/infinispan/configuration/serializing/Serializer.java
@@ -442,14 +442,16 @@ public class Serializer extends AbstractStoreSerializer implements Configuration
          writer.writeStartElement(Element.CUSTOM_INTERCEPTORS);
          for (InterceptorConfiguration interceptor : customInterceptors.interceptors()) {
             AttributeSet attributes = interceptor.attributes();
-            writer.writeStartElement(Element.INTERCEPTOR);
-            attributes.write(writer, InterceptorConfiguration.INTERCEPTOR_CLASS, Attribute.CLASS);
-            attributes.write(writer, InterceptorConfiguration.AFTER, Attribute.AFTER);
-            attributes.write(writer, InterceptorConfiguration.BEFORE, Attribute.BEFORE);
-            attributes.write(writer, InterceptorConfiguration.INDEX, Attribute.INDEX);
-            attributes.write(writer, InterceptorConfiguration.POSITION, Attribute.POSITION);
-            writeTypedProperties(writer, interceptor.properties());
-            writer.writeEndElement();
+            if (!attributes.attribute(InterceptorConfiguration.INTERCEPTOR_CLASS).isNull()) {
+               writer.writeStartElement(Element.INTERCEPTOR);
+               attributes.write(writer, InterceptorConfiguration.INTERCEPTOR_CLASS, Attribute.CLASS);
+               attributes.write(writer, InterceptorConfiguration.AFTER, Attribute.AFTER);
+               attributes.write(writer, InterceptorConfiguration.BEFORE, Attribute.BEFORE);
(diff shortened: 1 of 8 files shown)


anistor:
@tristantarrant This is a very serious issue, so I backported it to 91x and 82x. De we need also 90x and 81x ? I suppose we don't...

tristantarrant:
8.1.x is now abandoned. 8.2.x is supported for EAP 7.1, but not for query
and indexing.
We only need 9.1 and 9.2.

gustavonalle:
Build went OOM, re-triggering

anistor:
CI looks okish, except https://ci.infinispan.org/job/Infinispan/job/PR-5810/2/testReport/junit/org.infinispan.client.hotrod.admin/CacheAdminTest/cacheReindexTest/

So lets have another run.

anistor:
retest this please

wburns:
Still aborted, retrying.

gustavonalle:
org.infinispan.client.hotrod.admin.CacheAdminTest.cacheReindexTest is failing and apparently only in this PR

anistor:
There's a hidden problem with cache creation based on a template in 9.1.x that is only visible now. The new cache created here (https://github.com/infinispan/infinispan/blob/eb78dc0ceda1b4abd61a08e3f0d1c38f47c5d98a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/admin/CacheAdminTest.java#L81) gets to share the custom interceptors with the template. This is seriously wrong, especially for indexed caches. The QueryInterceptor should not be shared because it keeps a reference to the SearchIntegrator, which is cache-specific, so this leads to sharing of the SearchIntegrator and indexes too. This SearchIntegrator mixup was hidden by yet another mixup, ISPN-8897, so the test expectations were met, apparently, but fixing ISPN-8897 reveals the ugly truth.

The cache creation op in 9.2.x is implemented differently so it does not have this interceptor sharing issue and consequently the fix did not generate such side effects on 9.2.x

To demonstrate my claim I made a temporary 'fix' to avoid sharing of the QueryInterceptor here: https://github.com/infinispan/infinispan/pull/5810/files#diff-669495cf61374f2bc93b487ee484219aR59

core/src/main/java/org/infinispan/configuration/serializing/Serializer.java
```diff
@@ -442,14 +442,16 @@ private void writeCustomInterceptors(XMLExtendedStreamWriter writer, Configurati
...
             AttributeSet attributes = interceptor.attributes();
-            writer.writeStartElement(Element.INTERCEPTOR);
-            attributes.write(writer, InterceptorConfiguration.INTERCEPTOR_CLASS, Attribute.CLASS);
-            attributes.write(writer, InterceptorConfiguration.AFTER, Attribute.AFTER);
-            attributes.write(writer, InterceptorConfiguration.BEFORE, Attribute.BEFORE);
-            attributes.write(writer, InterceptorConfiguration.INDEX, Attribute.INDEX);
-            attributes.write(writer, InterceptorConfiguration.POSITION, Attribute.POSITION);
-            writeTypedProperties(writer, interceptor.properties());
-            writer.writeEndElement();
+            if (!attributes.attribute(InterceptorConfiguration.INTERCEPTOR_CLASS).isNull()) {
```
anistor: This change is not actually needed because in 9.1.x cache creation based on template does not use the Serializer as it happens on 92x.

anistor:
So the real question here is why is the template cache started at all? If it were not started, there would be no QueryInterceptor created by LifecycleManager, so no custom interceptor sharing issue when creating newCache based on the config of template cache.

tristantarrant:
I think the safest choice is to just tell users to upgrade to 9.2.

anistor:
Agreed.

PR #5812, opened 2018-03-02
ISPN-8907 Prevent CacheManager start when global state is corrupt

tristantarrant:
- perform validation of the global state file on startup

https://issues.jboss.org/browse/ISPN-8907


pruivo:
looks good to me. Integrated! thanks @tristantarrant

PR #5817, opened 2018-03-05
ISPN-8918 When a cache is stopped and restarted, its configuration re…

rhusar:
…ference within component registry is not refreshed with defineConfiguration/undefineConfiguration

When a cache is stopped and restarted, its configuration reference within component registry is not refreshed with defineConfiguration/undefineConfiguration.
The problem is that org/infinispan/factories/ComponentRegistry.java:77 is only called once the cache is being wired thus registerComponent(configuration, Configuration.class); is the only place where the configuration is put into the registry.
The best solution is to remove the cache in TERMINATED state once configuration is undefined thus establishing a clear terminal lifecycle.

Jira
https://issues.jboss.org/browse/ISPN-8918

Workaround is possible in wildfly, so not urgent.

### DIFF
--- a/core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
+++ b/core/src/main/java/org/infinispan/manager/DefaultCacheManager.java
@@ -403,6 +403,7 @@ public class DefaultCacheManager implements EmbeddedCacheManager {
             }
          }
          configurationManager.removeConfiguration(configurationName);
+         this.getGlobalComponentRegistry().removeCache(configurationName);
       }
    }
 


rhusar:
Hey @tristantarrant since you did the intial implementation for undefining caches (https://issues.jboss.org/browse/ISPN-5033) you might want to take a look. Thx!

galderz:
@rhusar, looks good. What implementation of ISPN-5033? You mean ISPN-8918?

rhusar:
Too late @pferraro, too late :-)

rhusar:
@galderz I did mean the original author of underfineConfiguration, so yes https://issues.jboss.org/browse/ISPN-5033 -- thanks for the merge.

PR #5818, opened 2018-03-06
ISPN-8901 Add merge policy NONE and how to deploy EntryMerpolicy to a server instance

wfink:
Adding the MergePolicy.NONE
Add chapter to explain how a custome EntryMergePolicy can be added to the server instance

https://issues.jboss.org/browse/ISPN-8901

### DIFF
--- a/documentation/src/main/asciidoc/user_guide/clustering.adoc
+++ b/documentation/src/main/asciidoc/user_guide/clustering.adoc
@@ -823,10 +823,11 @@ consistency is sacrificed in the presence of partition(s). Below is a list of th
 
 | ALLOW_READS
 | Allows reads for a given key if it exists in this partition, but only allows writes if this partition contains all owners of a segment.
-| Availability
+This is still a consistent approach because some entries are readable if available in this partition, but from a client application perspective it is not deterministic.
+| Consistency
 
 | ALLOW_READ_WRITES
-| Allow entries on each partition to diverge, with conflicts resolved during merge.
+| Allow entries on each partition to diverge, with conflict resolution attempted upon the partitions merging.
 | Availability
 |===
 
@@ -1013,6 +1014,13 @@ Currently Infinispan provides the following implementations of EntryMergePolicy:
 
 | MergePolicies.REMOVE_ALL
 | Always remove a key from the cache when a conflict is detected.
+
+| MergePolicies.NONE
+| No attempt is made to resolve conflicts. Instead, entries hosted on the minority partition are removed and the nodes in this partition do not hold any data until the rebalance starts. Note, this behaviour is equivalent to prior Infinispan versions which did not support conflict resolution.
+Note, in this case all changes made to entries hosted on the minority partition are lost, but once the rebalance has finished all entries will be consistent.
+
+| Fully qualified class name
+| The custom implementation for merge will be used <<partition-handling-custom-merge-policy, Custom merge policy>>
 |===
 
 ==== Usage
@@ -1070,8 +1078,9 @@ dcc.clustering().partitionHandling()
                     .mergePolicy(MergePolicies.PREFERRED_ALWAYS);
 ----
 
-It's also possible to provide custom implementations of the EntryMergePolicy:
-
+[[partition-handling-custom-merge-policy]]
+===== Implement a custom merge policy
+It's also possible to provide custom implementations of the EntryMergePolicy
 [source,xml]
 ----
 <distributed-cache name="the-default-cache">
@@ -1086,6 +1095,29 @@ dcc.clustering().partitionHandling()
                     .whenSplit(PartitionHandling.ALLOW_READ_WRITES)
                     .mergePolicy(new CustomMergePolicy());
 ----
+[source,java]
+----
+public class CustomMergePolicy implements EntryMergePolicy<String, String> {
+
+   @Override
+   public CacheEntry<String, String> merge(CacheEntry<String, String> preferredEntry, List<CacheEntry<String, String>> otherEntries) {
+      // decide which entry should be used
+      
+      return the_solved_CacheEntry;
+   }
+----
+
+===== Deploy custom merge policies to a Infinispan server instance
+To utilise a custom EntryMergePolicy implementation on the server, it's necessary for the implementation class(es) to be deployed to the server. This is accomplished by utilising the java service-provider convention and packaging the class files in a jar which has a META-INF/services/org.infinispan.conflict.EntryMergePolicy file containing the fully qualified class name of the EntryMergePolicy implementation.
+[source]
+----
+# list all necessary implementations of EntryMergePolicy with the full qualified name
+org.example.CustomMergePolicy
+----
+
+In order for a Custom merge policy to be utilised on the server, compatibility mode must be enabled if your policies semantics require access to the stored Key/Value objects. This is because without compatibility mode, cache entries in the server are stored in a marshalled format and the Key/Value objects returned to your policy would be instances of WrappedByteArray.
+However, if the custom policy only depends on the metadata associated with a cache entry, then compatibility mode is not required and should be avoided (unless needed for other reasons) due to the additional performance cost of marshalling data per request.
+Finally, compatibility mode is never required if one of the provided merge policies is used.
 
 [[partition-handling-monitoring]]
 ==== Monitoring and administration


documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1013,6 +1013,12 @@ Currently Infinispan provides the following implementations of EntryMergePolicy:
 
 | MergePolicies.REMOVE_ALL
 | Always remove a key from the cache when a conflict is detected.
+
+| MergePolicies.NONE
+| There is no action for conflicting entries. Entries are deleted on nodes which are not (longer) and owner. Copied from the primary owner to owners which does not have it, and not changed at all for nodes which are still owner and have the entry present no matter what value.
```
ryanemerson: I think this would be better as:

> No attempt is made to resolve conflicts. Instead, entries hosted on the minority partition are removed and the nodes in this partition do not hold any data until the rebalance starts. Note, this behaviour is equivalent to prior Infinispan versions which did not support conflict resolution.

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1013,6 +1013,12 @@ Currently Infinispan provides the following implementations of EntryMergePolicy:
...
+
+| Full qualified class name
+|A custom implementation of merge policy
```
ryanemerson: Is this section unfinished or included by mistake? It does not make sense on its own.

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1086,6 +1092,15 @@ dcc.clustering().partitionHandling()
                     .whenSplit(PartitionHandling.ALLOW_READ_WRITES)
                     .mergePolicy(new CustomMergePolicy());
 ----
+==== Deploy custom merge policies to a Infinispan server instance
+If custom merge-policy is configured inside of the server configuration the classes need to be available whithin all servers of this cluster.
```
ryanemerson: How about:

> To utilise a custom EntryMergePolicy implementation on the server, it's necessary for the implementation class(es) to be deployed to the server. This is accomplished by utilising the java service-provider convention and packaging the class files in a jar which has a `META-INF/services/org.infinispan.conflict.EntryMergePolicy` file containing the fully qualified class name of the `EntryMergePolicy` implementation.

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1086,6 +1092,15 @@ dcc.clustering().partitionHandling()
...
+For each policy an implementation of the EntryMergePolicy interface is needed.
```
ryanemerson: I think this is already covered in the [Merge Policies section](http://infinispan.org/docs/9.2.x/user_guide/user_guide.html#merge-policies).

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1013,6 +1014,14 @@ Currently Infinispan provides the following implementations of EntryMergePolicy:
 
 | MergePolicies.REMOVE_ALL
 | Always remove a key from the cache when a conflict is detected.
+
+| MergePolicies.NONE
+| No attempt is made to resolve conflicts. Instead, entries hosted on the minority partition are removed and the nodes in this partition do not hold any data until the rebalance starts. Note, this behaviour is equivalent to prior Infinispan versions which did not support conflict resolution.
+Note, in this case all changes made to one (the part with less nodes) side of the split are lost. But the state after rebalancing will be consistent.
```
ryanemerson: I think we can combine the last two sentences like so: "In this case all changes made to entries hosted on the minority partition are lost, but once the rebalance has finished all entries will be consistent"

We have already referred to the minority partition, so we should do so again.

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -823,10 +823,11 @@ consistency is sacrificed in the presence of partition(s). Below is a list of th
 
 | ALLOW_READS
 | Allows reads for a given key if it exists in this partition, but only allows writes if this partition contains all owners of a segment.
-| Availability
+This is still a consistent approach because some entries are readable if available in this partition, but from a client application perspective it is not deterministic.
+| Consistency
 
 | ALLOW_READ_WRITES
-| Allow entries on each partition to diverge, with conflicts resolved during merge.
+| Allow entries on each partition to diverge, with the possibility of conflicts resolved during merge.
```
ryanemerson: "Allow entries on each partition to diverge, with conflict resolution attempted upon the partitions merging."

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1013,6 +1014,14 @@ Currently Infinispan provides the following implementations of EntryMergePolicy:
 
 | MergePolicies.REMOVE_ALL
 | Always remove a key from the cache when a conflict is detected.
+
+| MergePolicies.NONE
+| No attempt is made to resolve conflicts. Instead, entries hosted on the minority partition are removed and the nodes in this partition do not hold any data until the rebalance starts. Note, this behaviour is equivalent to prior Infinispan versions which did not support conflict resolution.
+Note, in this case all changes made to one (the part with less nodes) side of the split are lost. But the state after rebalancing will be consistent.
+
+| Full qualified class name
```
ryanemerson: Fully

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1013,6 +1014,14 @@ Currently Infinispan provides the following implementations of EntryMergePolicy:
...
+|Instead of setting a included MergePolicy a custom merge policy implementation can be added by set the full qualified class name of the custom implementation for the org.infinispan.conflict.EntryMergePolicy<K,V>.
+TODO: Note that implementation need to know what kind of Objects are provided, this is different in embedded or client/server mode. See XXXXXXXXXXX
```
ryanemerson: I don't think this section belongs in this table, as this table details the EntryMergePolicy implementations provided by Infinispan. I think a separate section detailing custom implementations would be better.

ryanemerson: Actually, this is already shown in the "Configuring partition handling" section.  We could just explicity state in the existing section that you must provide the FQN of the implementation.

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1086,6 +1095,13 @@ dcc.clustering().partitionHandling()
                     .whenSplit(PartitionHandling.ALLOW_READ_WRITES)
                     .mergePolicy(new CustomMergePolicy());
 ----
+==== Deploy custom merge policies to a Infinispan server instance
+To utilise a custom EntryMergePolicy implementation on the server, it's necessary for the implementation class(es) to be deployed to the server. This is accomplished by utilising the java service-provider convention and packaging the class files in a jar which has a META-INF/services/org.infinispan.conflict.EntryMergePolicy file containing the fully qualified class name of the EntryMergePolicy implementation.
+[source]
+----
+# list all necessary implementations of EntryMergePolicy with the full qualified name
+org.example.CustomEntryMergePolicy
+----
```
ryanemerson: We should state that currently it's only possible for this to work with Compatibility mode enabled.

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1086,6 +1095,28 @@ dcc.clustering().partitionHandling()
...
+===== Deploy custom merge policies to a Infinispan server instance
+To utilise a custom EntryMergePolicy implementation on the server, it's necessary for the implementation class(es) to be deployed to the server. This is accomplished by utilising the java service-provider convention and packaging the class files in a jar which has a META-INF/services/org.infinispan.conflict.EntryMergePolicy file containing the fully qualified class name of the EntryMergePolicy implementation.
+[source]
+----
+# list all necessary implementations of EntryMergePolicy with the full qualified name
+org.example.CustomMergePolicy
+----
+
+Consider that implementation need to know what kind of Objects are provided and how the serialization is set when deploying in client server mode without compatibility enabled, this is different to embedded mode.
+If compatibility mode is enabled the CacheEntry is stored in its unmarshalled form but this will have a performance impact because the entry need to be marshalled for every transfer over the network.
```
ryanemerson: I think this section regarding compatibility can be expanded, how about:

"In order for a Custom merge policy to be utilised on the server, compatibility mode must be enabled if your policies semantics require access to the stored Key/Value objects. This is because without compatibility mode, cache entries in the server are stored in a marshalled format and the Key/Value objects returned to your policy would be instances of `WrappedByteArray`.  However, if your custom policy only depends on the metadata associated with a cache entry, then compatibility mode is not required and should be avoided (unless needed for other reasons) due to the additional performance cost of marshalling data per request. Finally, compatibility mode is never required if you're using one of the provided merge policies."

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1013,6 +1014,13 @@ Currently Infinispan provides the following implementations of EntryMergePolicy:
 
 | MergePolicies.REMOVE_ALL
 | Always remove a key from the cache when a conflict is detected.
+
+| MergePolicies.NONE
+| No attempt is made to resolve conflicts. Instead, entries hosted on the minority partition are removed and the nodes in this partition do not hold any data until the rebalance starts. Note, this behaviour is equivalent to prior Infinispan versions which did not support conflict resolution.
+Note, In this case all changes made to entries hosted on the minority partition are lost, but once the rebalance has finished all entries will be consistent.
```
ryanemerson: "In" should be lower case.

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -1086,6 +1095,30 @@ dcc.clustering().partitionHandling()
...
+===== Deploy custom merge policies to a Infinispan server instance
+To utilise a custom EntryMergePolicy implementation on the server, it's necessary for the implementation class(es) to be deployed to the server. This is accomplished by utilising the java service-provider convention and packaging the class files in a jar which has a META-INF/services/org.infinispan.conflict.EntryMergePolicy file containing the fully qualified class name of the EntryMergePolicy implementation.
+[source]
+----
+# list all necessary implementations of EntryMergePolicy with the full qualified name
+org.example.CustomMergePolicy
+----
+
+In order for a Custom merge policy to be utilised on the server, compatibility mode must be enabled if your policies semantics require access to the stored Key/Value objects. This is because without compatibility mode, cache entries in the server are stored in a marshalled format and the Key/Value objects returned to your policy would be instances of WrappedByteArray.
+Without compatibility mode the marshaller need to packaged with the merge policy and the merge policy implementation depends on the marshalling configuration which can be real complicated.
```
ryanemerson: I think we should remove this sentence for now as it will probably confuse all but the most advanced users. Furthermore, this sentence contradicts the previous assertion that "compatibility mode must be enabled", so if we do keep it then we must also adapt the rest of the paragraph accordingly. 

From a technical perspective another added complication is that any custom user classes would also have to be packaged with the merge policy if unmarshalling were to occur in the merge policy. Personally I don't think we should be advocating this until we have a more user-friendly offering.

ryanemerson:
Thanks @wfink

PR #5819, opened 2018-03-06
ISPN-8912 SoftIndexFileStore does not fail gracefully if max node sizee exceeds 32767

rvansa:
* validate the configuration
* use shorts where the value should be limited to short
* use i18d logging
* minor code cleanup based on IDE suggestions

https://issues.jboss.org/browse/ISPN-8912

### DIFF
--- a/persistence/soft-index/src/main/java/org/infinispan/persistence/sifs/Index.java
+++ b/persistence/soft-index/src/main/java/org/infinispan/persistence/sifs/Index.java
@@ -17,7 +17,6 @@ import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.infinispan.util.TimeService;
-import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
 /**
@@ -28,11 +27,11 @@ import org.infinispan.util.logging.LogFactory;
  * @author Radim Vansa &lt;rvansa@redhat.com&gt;
  */
 class Index {
-   private static final Log log = LogFactory.getLog(Index.class);
+   private static final Log log = LogFactory.getLog(Index.class, Log.class);
    private static final boolean trace = log.isTraceEnabled();
    private static final int GRACEFULLY = 0x512ACEF0;
    private static final int DIRTY = 0xD112770C;
-   protected static final int INDEX_FILE_HEADER_SIZE = 30;
+   private static final int INDEX_FILE_HEADER_SIZE = 30;
 
    private final String indexDir;
    private final FileProvider fileProvider;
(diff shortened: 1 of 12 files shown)


persistence/soft-index/src/main/java/org/infinispan/persistence/sifs/Index.java
```diff
@@ -342,17 +328,18 @@ private void shutdown() throws IOException {
...
          buffer.putInt(0, freeBlocks.size());
          write(indexFile, buffer);
-         for (Map.Entry<Integer, List<IndexSpace>> entry : freeBlocks.entrySet()) {
+         for (Map.Entry<Short, List<IndexSpace>> entry : freeBlocks.entrySet()) {
             List<IndexSpace> list = entry.getValue();
             int requiredSize = 8 + list.size() * 10;
             buffer = buffer.capacity() < requiredSize ? ByteBuffer.allocate(requiredSize) : buffer;
             buffer.position(0);
             buffer.limit(requiredSize);
+            // TODO: change this to short
```
ryanemerson: Did you forget about this?

persistence/soft-index/src/main/java/org/infinispan/persistence/sifs/Index.java
```diff
@@ -389,7 +376,9 @@ private void loadFreeBlocks(long freeBlocksOffset) throws IOException {
             if (!read(indexFile, buffer)) {
                throw new IOException("Cannot read free blocks lists!");
             }
+            // TODO: change this to short
```
ryanemerson: ?

persistence/soft-index/src/main/java/org/infinispan/persistence/sifs/IndexNode.java
```diff
@@ -861,8 +841,10 @@ private static int compare(byte[] first, byte[] second) {
...
 
-   private int headerLength() {
-      return INNER_NODE_HEADER_SIZE + prefix.length;
+   private short headerLength() {
+      int headerLength = INNER_NODE_HEADER_SIZE + prefix.length;
+      assert headerLength <= Short.MAX_VALUE;
+      return (short) headerLength;
    }
 
    private int contentLength() {
```
ryanemerson: Should be `short`

persistence/soft-index/src/main/java/org/infinispan/persistence/sifs/SoftIndexFileStore.java
```diff
@@ -241,7 +240,8 @@ public void stop() {
          indexQueue = null;
          storeQueue = null;
       } catch (InterruptedException e) {
-         throw new PersistenceException("Cannot stop cache store", e);
+         Thread.currentThread().interrupt();
```
wburns: I wouldn't normally expect to reinterpret the thread when throwing an exception that stating that we were interrupted. I find I would normally reset the status and return early or throw the exception, but not both.

persistence/soft-index/src/main/java/org/infinispan/persistence/sifs/Index.java
rvansa: No; I didn't want to change binary format in some micro - the plan is to add some version numbers both to index and to data files, and then I'll do the change to binary format as well.

persistence/soft-index/src/main/java/org/infinispan/persistence/sifs/SoftIndexFileStore.java
rvansa: OK, I have thought the best practice is to set thread status to interrupted if the thrown exception is not directly `InterruptedException`.

wburns: Yeah I was thinking it was a different exception. Seems good.

wburns:

PR #5821, opened 2018-03-06
ISPN-8896 Fix HTTP/2 support

slaskawi:
https://issues.jboss.org/browse/ISPN-8896

### DIFF
--- a/server/rest/src/test/java/org/infinispan/rest/Http2Test.java
+++ b/server/rest/src/test/java/org/infinispan/rest/Http2Test.java
@@ -41,6 +41,7 @@ import io.netty.util.CharsetUtil;
  *
  * @author Sebastian Łaskawiec
  */
+@Test(groups = "functional", testName = "rest.Http2Test")
 public final class Http2Test extends AbstractInfinispanTest {
 
     public static final String KEY_STORE_PATH = Http2Test.class.getClassLoader().getResource("./default_client_truststore.jks").getPath();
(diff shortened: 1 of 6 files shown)


server/rest/src/test/java/org/infinispan/rest/Http2Test.java
```diff
@@ -41,6 +41,7 @@
  *
  * @author Sebastian Łaskawiec
  */
+@Test(groups = "functional", testName = "rest.Http2Test")
```
slaskawi: This is silly... why TestNG doesn't pick up classes ending with `test`, just like JUnit? :imp:

slaskawi:
None of the failures is related.

gustavonalle:
@slaskawi Can we come up with a test for the HTTP/2 support? It was broken as we did not realize it :disappointed: 

Maybe pick a non-netty client? There is a couple aroung, e.g.  http://unrestful.io/2015/10/10/http2-java-client-examples.html

slaskawi:
> @slaskawi Can we come up with a test for the HTTP/2 support? It was broken and we did not realize it disappointed

It was actually my fault. I didn't annotate the test with `@Test`... It's so silly and confusing that this works out of the box for JUnit and doesn't for TestNG...

gustavonalle:
@slaskawi right, I can see that  ```Http2Test``` is failing on master when running from the IDE  :+1:

gustavonalle:
merged, thanks!

PR #5823, opened 2018-03-06
ISPN-8916 Enable cross origin requests for the admin console

gustavonalle:
https://issues.jboss.org/browse/ISPN-8916

### DIFF
--- a/server/rest/src/main/java/org/infinispan/rest/Http11To2UpgradeHandler.java
+++ b/server/rest/src/main/java/org/infinispan/rest/Http11To2UpgradeHandler.java
@@ -1,11 +1,24 @@
 package org.infinispan.rest;
 
+import static io.netty.handler.codec.http.HttpMethod.DELETE;
+import static io.netty.handler.codec.http.HttpMethod.GET;
+import static io.netty.handler.codec.http.HttpMethod.HEAD;
+import static io.netty.handler.codec.http.HttpMethod.OPTIONS;
+import static io.netty.handler.codec.http.HttpMethod.POST;
+import static io.netty.handler.codec.http.HttpMethod.PUT;
+
+import java.util.List;
+
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPipeline;
+import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpObjectAggregator;
 import io.netty.handler.codec.http.HttpServerCodec;
 import io.netty.handler.codec.http.HttpServerUpgradeHandler;
+import io.netty.handler.codec.http.cors.CorsConfig;
+import io.netty.handler.codec.http.cors.CorsConfigBuilder;
+import io.netty.handler.codec.http.cors.CorsHandler;
 import io.netty.handler.codec.http2.DefaultHttp2Connection;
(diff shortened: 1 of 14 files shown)


gustavonalle:
@ryanemerson Agreed. The only thing to be reviewed is that in this PR I enabled CORS for [all origins and methods](https://github.com/infinispan/infinispan/pull/5823/files#diff-d7f2ec1b10cf76bae14da59040b9bdb1R46)
Ideally we should maybe enabled for localhost/127.0.0.1 by default and offer a way to configure other domains

server/rest/src/main/java/org/infinispan/rest/Http11To2UpgradeHandler.java
```diff
@@ -61,6 +75,7 @@ private void configureHttp2(ChannelPipeline pipeline) {
    private void configureHttp1(ChannelPipeline pipeline) {
       final HttpServerCodec httpCodec = new HttpServerCodec(MAX_INITIAL_LINE_SIZE, MAX_HEADER_SIZE, restServer.getConfiguration().maxContentLength());
       pipeline.addLast(httpCodec);
+      pipeline.addLast(new CorsHandler(corsConfig));
```
slaskawi: Is there any reason why we can't inline it here? I believe we don't need to reference.

And could you please confirm if it works with HTTP/2?

gustavonalle: Inline what exactly?

Regarding HTTP/2, I have no idea. AFAICT, ```XMLHttpRequest``` uses HTTP/1.x, in which circumstances they use HTTP/2 so that I can test?

gustavonalle: In any case, added a task to https://issues.jboss.org/browse/ISPN-8915 to investigate that.

slaskawi: > inline what exactly?

Something like this:

```
pipeline.addLast(CorsConfigBuilder.forAnyOrigin().allowedRequestMethods(GET, POST, PUT, DELETE, HEAD, OPTIONS).build());
```

> Regarding HTTP/2, I have no idea. AFAICT, XMLHttpRequest uses HTTP/1.x, in which circumstances they use HTTP/2 so that I can test?

Is this feature connected only to `XMLHttpRequest`? If so, does it always use `HTTP/1.x`? I briefly looked through the [Wiki page](https://en.wikipedia.org/wiki/XMLHttpRequest) and couldn't find any information that this is `HTTP/1.x` specific...

gustavonalle: The ```corsConfig``` needs to be created once, inlining it would cause unnecessary allocations every time a channel is created.

ryanemerson:
@gustavonalle I think the default should just expose the host of the server (could be localhost), but with the configured management port. That should be sufficient for the console to function and it would mean that the user can use the console without additional configuration.

+1 on allowing users to configure domains, but I don't think that has to be part of this PR.

vblagoje:
Hey guys, I'll start testing now and I'll report my findings EOD.

vblagoje:
@gustavonalle verified the latest change and everything works as expected

PR #5824, opened 2018-03-07
ISPN-8921 KeyStream stream operations use entrySet on remote nodes

wburns:
* Now keySet operations use key set on remote nodes
* Clean up keySet/entrySet/values wrappers for interceptors

https://issues.jboss.org/browse/ISPN-8921
https://issues.jboss.org/browse/ISPN-8923
https://issues.jboss.org/browse/ISPN-8924
https://issues.jboss.org/browse/ISPN-8880
https://issues.jboss.org/browse/ISPN-8926

The main changes of this class revolve around the [variable](https://github.com/infinispan/infinispan/pull/5824/files#diff-83112cef2238f7c65212a112e94aeb8aR75) `entryStream` which dictate if the stream operation contains entries or not. This is determined [by whether](https://github.com/infinispan/infinispan/pull/5824/files#diff-be6f435a8edfd997f62d9a90ad59582bR109) the `toKeyFunction` is not null (entry) or is null (key).

The whole `Original` type found in code is to help with typing to prevent passing in the wrong types. But this type is normally just CacheEntry or Object depending on if it is a entryStream or keyStream.  Most of these can be pretty much ignored, including all the ForEach classes. Later at some point I hope to enhance the typing on the IntermediateOperations and that would allow typing all the way down.

I also removed the old `*IteratorOperation` classes as they are no longer used since we use the new reactive streams iterator implementation.


wburns:
To be honest this JIRA has a lot of noise and very little substantive changes.

The big change is that the DistributedCacheStream now has a toKeyFilter that based on if it is not null will cause the remote node to fire off a key stream or an entry stream.

I also cleaned up some code in EncoderCache/PrefetchInterceptor so they didn't delegate calls to the cache and instead let them push down to the child CacheSet (which should help for local caches). I believe the EntrySet and KeySet Command wrappers can also ignore the cache, but I haven't changed those yet.

I will try to get some perf #'s difference with doing a size method on a DIST cache tomorrow.

wburns:
A couple test failures could be related: https://ci.infinispan.org/job/Infinispan/job/PR-5824/6/#showFailuresLink

Looking...

wburns:
Added a couple more bug fixes that were exposed when moving to keySet (ISPN-8880 and ISPN-8926). @gustavonalle the latter was the cause of my problems :)

wburns:
The 4 test failures https://ci.infinispan.org/job/Infinispan/job/PR-5824/9/#showFailuresLink seem unrelated.

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -106,35 +106,11 @@
     * @param distributedBatchSize default size of distributed batches
     * @param executor executor to be used for certain operations that require async processing (ie. iterator)
     */
```
ryanemerson: We should probably update the javadoc to explain the purpose of `toKeyFunction` and maybe explain how it's required for tracking duplicates etc.

core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
```diff
@@ -64,24 +65,24 @@
  * @param <T> The type returned by the stream
  * @param <S> The stream interface
  */
-public abstract class AbstractCacheStream<T, S extends BaseStream<T, S>, S2 extends S> implements BaseStream<T, S> {
+public abstract class AbstractCacheStream<Original, T, S extends BaseStream<T, S>, S2 extends S> implements BaseStream<T, S> {
```
ryanemerson: Should we add `@param <Original>` to the javadoc?

wburns: Done

wburns:
Added comments, rebased and updated PR description to help a little more :)

tristantarrant:
Excellent work, merged !

PR #5825, opened 2018-03-09
ISPN-8927 Do not hardcode HTTP header and status constants in REST tests

anistor:
https://issues.jboss.org/browse/ISPN-8927


wburns:

PR #5827, opened 2018-03-09
ISPN-8769 trace logs improvements

karesti:
https://issues.jboss.org/browse/ISPN-8769

* Adds isTraceEnabled method call
* Adds lock name in every log


wburns:

PR #5828, opened 2018-03-09
ISPN-8930 RestOperationsTest random test failures

wburns:
RestOperationsTest#shouldPutImmortalEntryWithZeroTtlAndIdleTime random
failure

* Increased expiration times so they shouldn't expire with thread delay

https://issues.jboss.org/browse/ISPN-8930


wburns:
I also changed it to clear the data container instead of the cache. This is much faster for off heap making the test run in 1/4 of the time it used to for me.

karesti:
@wburns RestOperationsTest did not  fail on the build but the org.infinispan.rest.RestOperationsOffHeapTest. shouldPutImmortalEntryWithZeroTtlAndIdleTime
failed
 org.junit.ComparisonFailure: expected:<[45190]L> but was:<[1859446]L>

wburns:
Hrmm, I think this actually might have revealed a bug in off heap, I will have to look closer in a little bit.

wburns:
I have fixed up the mix up in off heap, hopefully no tests were relying on the old behavior :D

PR #5829, opened 2018-03-12
ISPN-8757 Avoid memory leaks on scheduler

karesti:
https://issues.jboss.org/browse/ISPN-8757

### DIFF
--- a/lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
+++ b/lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
@@ -1,0 +1,128 @@
+package org.infinispan.lock.impl.lock;
+
+import java.lang.invoke.MethodHandles;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.infinispan.commons.logging.LogFactory;
+import org.infinispan.lock.impl.log.Log;
+
+/**
+ * This class holds the logic to schedule/abort requests that need to be completed at a given time.
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+public class RequestExpirationScheduler {
+
+   private static final Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass(), Log.class);
+   private static final boolean trace = log.isTraceEnabled();
+
+   private final ScheduledExecutorService scheduledExecutorService;
+   private final ConcurrentMap<String, ScheduledRequest> scheduledRequests = new ConcurrentHashMap<>();
+
+   public RequestExpirationScheduler(ScheduledExecutorService scheduledExecutorService) {
+      this.scheduledExecutorService = scheduledExecutorService;
+   }
+
+   class ScheduledRequest {
+      private CompletableFuture<Boolean> request;
+      private ScheduledFuture<?> scheduledFuture;
+
+      public ScheduledRequest(CompletableFuture<Boolean> request, ScheduledFuture<?> scheduledFuture) {
+         this.request = request;
+         this.scheduledFuture = scheduledFuture;
+      }
+
+      public CompletableFuture<Boolean> getRequest() {
+         return request;
+      }
+
+      public ScheduledFuture<?> getScheduledFuture() {
+         return scheduledFuture;
+      }
+   }
+
+   /**
+    * Schedules a request for completion
+    *
+    * @param requestId, the unique identifier if the request
+    * @param request,   the request
+    * @param time,      time expressed in long
+    * @param unit,      {@link TimeUnit}
+    */
+   public void scheduleForCompletion(String requestId, CompletableFuture<Boolean> request, long time, TimeUnit unit) {
+      if (request.isDone()) {
+         if (trace) {
+            log.tracef("Request[%s] is not scheduled because is already done", requestId);
+         }
+         return;
+      }
+
+      if (scheduledRequests.containsKey(requestId)) {
+         String message = String.format("Request[%s] is not scheduled because it is already scheduled", requestId);
+         log.error(message);
+         throw new IllegalStateException(message);
+      }
+
+      if (trace) {
+         log.tracef("Request[%s] being scheduled to be completed in [%d, %s]", requestId, time, unit);
+      }
+
+      ScheduledFuture<?> scheduledFuture = scheduledExecutorService.schedule(() -> {
+         request.complete(false);
+         scheduledRequests.remove(requestId);
+      }, time, unit);
+
+      scheduledRequests.putIfAbsent(requestId, new ScheduledRequest(request, scheduledFuture));
+   }
+
+   /**
+    * Aborts the scheduled request if the request is already completed
+    *
+    * @param requestId, unique identifier of the request
+    */
+   public void abortScheduling(String requestId) {
+      abortScheduling(requestId, false);
+   }
+
+   /**
+    * Aborts the scheduled request. If force is true, it will abort even if the request is not completed
+    *
+    * @param requestId, unique identifier of the request
+    * @param force,     force abort
+    */
+   public void abortScheduling(String requestId, boolean force) {
+      if (trace) {
+         log.tracef("Request[%s] abort scheduling", requestId);
+      }
+      ScheduledRequest scheduledRequest = scheduledRequests.get(requestId);
+      if (scheduledRequest != null && (scheduledRequest.request.isDone() || force)) {
+         scheduledRequest.scheduledFuture.cancel(false);
+         scheduledRequests.remove(requestId);
+      }
+   }
+
+   /**
+    * Returns the size of the currently scheduled requests
+    *
+    * @return the number of requests that are pending on the scheduler
+    */
+   public int countScheduledRequests() {
+      return scheduledRequests.size();
+   }
+
+   /**
+    * Get scheduled request reference by id if such exist
+    *
+    * @param requestId, the id of the scheduled request
+    * @return {@link ScheduledRequest} the request
+    */
+   public ScheduledRequest get(String requestId) {
+      return scheduledRequests.get(requestId);
+   }
+}
--- a/lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
+++ b/lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
@@ -1,0 +1,136 @@
+package org.infinispan.lock.impl.lock;
+
+import static org.infinispan.test.Exceptions.expectException;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertFalse;
+import static org.testng.AssertJUnit.assertTrue;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.mockito.ArgumentCaptor;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+@Test(groups = "functional", testName = "clusteredLock.RequestExpirationSchedulerTest")
+public class RequestExpirationSchedulerTest {
+
+   private RequestExpirationScheduler expirationScheduler;
+   private ScheduledExecutorService scheduledExecutorService = mock(ScheduledExecutorService.class);
+   private ScheduledFuture scheduledFutureMock;
+   private CompletableFuture<Boolean> booleanCompletableFuture;
+
+   @BeforeMethod
+   public void createRequestExpirationScheduler() {
+      expirationScheduler = new RequestExpirationScheduler(scheduledExecutorService);
+      scheduledFutureMock = mock(ScheduledFuture.class);
+      booleanCompletableFuture = new CompletableFuture();
+      when(scheduledExecutorService.schedule(any(Runnable.class), anyLong(), any(TimeUnit.class))).thenReturn(scheduledFutureMock);
+   }
+
+   @Test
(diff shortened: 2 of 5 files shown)


karesti:
test failures are unrelated

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
```diff
@@ -0,0 +1,123 @@
...
+   public void createRequestExpirationScheduler() {
+      EmbeddedClusteredLockManager eclm = (EmbeddedClusteredLockManager) cm;
+      expirationScheduler = new RequestExpirationScheduler(eclm.getScheduledExecutorService());
+   }
+
+   @Test
+   public void testScheduleForCompletion() throws Exception {
+      CompletableFuture<Boolean> request = new CompletableFuture();
+
+      expirationScheduler.scheduleForCompletion("123", request, 10, TimeUnit.MILLISECONDS);
```
wburns: I wouldn't trust these tests with such low timeouts. I can guarantee some of these will fail at some point in the future. I would instead use Mockito to mock out the `ScheduledExecutorService` and verify invocations on it. This way the test should be 100% deterministic and should be pretty quick not having to wait for executions at all.

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * This class holds the logic to schedule/abort requests that need to be completed at a given time.
+ *
+ * @author Katia Aresti, karesti@redhat.com
+ * @since 9.2
+ */
+public class RequestExpirationScheduler {
+
+   private static final Log log = LogFactory.getLog(RequestExpirationScheduler.class, Log.class);
```
wburns: You can just use `MethodHandles.lookup().lookupClass()` instead of coding the class name.

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
```diff
@@ -0,0 +1,123 @@
...
+   private static final boolean trace = log.isTraceEnabled();
+
+   private final ScheduledExecutorService scheduledExecutorService;
+   private final ConcurrentHashMap<String, ScheduledRequest> scheduledRequests = new ConcurrentHashMap<>();
```
wburns: Just use the interface `ConcurrentMap`

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
```diff
@@ -0,0 +1,123 @@
...
+
+   public RequestExpirationScheduler(ScheduledExecutorService scheduledExecutorService) {
+      this.scheduledExecutorService = scheduledExecutorService;
+   }
+
+   class ScheduledRequest {
+      private String requestId;
```
wburns: This variable seems superfluous.

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
```diff
@@ -0,0 +1,123 @@
...
+      }
+
+      if (scheduledRequests.containsKey(requestId)) {
+         if(trace) {
+            log.tracef("Request[%s] is not scheduled because it is already scheduled", requestId);
+         }
+         return;
+      }
+
+      if(trace) {
```
wburns: For some reason all the `if (trace)` calls are missing a space :)

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
```diff
@@ -0,0 +1,123 @@
...
+    * Aborts the scheduled request
+    *
+    * @param requestId, unique identifier of the request
+    */
+   public void abortScheduling(String requestId) {
+      if(trace) {
+         log.tracef("Request[%s] abort scheduling", requestId);
+      }
+      ScheduledRequest scheduledRequest = scheduledRequests.get(requestId);
+      if (scheduledRequest != null && scheduledRequest.request.isDone()) {
```
wburns: We shouldn't have to care if the request is done or not, just cancel the future.

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
```diff
@@ -0,0 +1,123 @@
...
+      if(trace) {
+         log.tracef("Request[%s] being scheduled to be completed in [%d, %s]", requestId, time, unit);
+      }
+
+      ScheduledFuture<?> scheduledFuture = scheduledExecutorService.schedule(() -> {
+         request.complete(false);
+         scheduledRequests.remove(requestId);
+      }, time, unit);
+
+      scheduledRequests.putIfAbsent(requestId, new ScheduledRequest(requestId, request, scheduledFuture));
```
wburns: Hrmm this seems odd to me. I wonder what should we do if we have multiple requests for the same request Id ? I would think we should probably set one to exception maybe?

karesti: we should't have more than one requestId, this should be unique. if not, it's a bug for me

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
karesti: haha spaces, I forgot to repass the formatter xD

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
karesti: I'm using it on tests

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
karesti: agree

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
karesti: I considered using mockito and verify at first but I reconsidered to test cancel status etc.... but ok  to use mockito if you think it will probably fail in the furure

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
karesti: ok

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
wburns: Which is unneeded as well. All the test is doing is verifying that the name is the same after already using the name to retrieve the request. We already retrieved a non null value, we can assume it is tied to the name we used to retrieve it :)

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
wburns: Then we should probably throw an exception or set an exception on a future or something here.

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
wburns: We still test cancellation :) It is just that you invoke it (like the scheduled executor would have) But you get to control when it is invoked. Our code should assume that scheduled executor works properly and this way we don't have to have weird timeouts or wait for such things to happen.

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
karesti: ok 🗡

karesti:
@wburns I addressed the comments in a separate commit, if you validate I amend and we can merge. If CI is fine, of course

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
```diff
@@ -0,0 +1,128 @@
...
+    */
+   public void scheduleForCompletion(String requestId, CompletableFuture<Boolean> request, long time, TimeUnit unit) {
+      if (request.isDone()) {
+         if (trace) {
+            log.tracef("Request[%s] is not scheduled because is already done", requestId);
+         }
+         return;
+      }
+
+      if (scheduledRequests.containsKey(requestId)) {
```
wburns: Ah yes this is the check to make sure there wasn't duplicates. Like I mentioned this seems like a bug to me if we have submitted more than 1 request for the same id without the other completing first.

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
```diff
@@ -0,0 +1,128 @@
...
+    * Aborts the scheduled request
+    *
+    * @param requestId, unique identifier of the request
+    */
+   public void abortScheduling(String requestId) {
+      if (trace) {
+         log.tracef("Request[%s] abort scheduling", requestId);
+      }
+      ScheduledRequest scheduledRequest = scheduledRequests.get(requestId);
+      if (scheduledRequest != null && scheduledRequest.request.isDone()) {
```
wburns: I am not sure if you saw my last comment about this, but I am fine if you don't remove it (more of a best practices kinda thing)

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
```diff
@@ -0,0 +1,115 @@
...
+      expirationScheduler = new RequestExpirationScheduler(scheduledExecutorService);
+      scheduledFutureMock = mock(ScheduledFuture.class);
+      booleanCompletableFuture = new CompletableFuture();
+      when(scheduledExecutorService.schedule(any(Runnable.class), anyLong(), any(TimeUnit.class))).thenReturn(scheduledFutureMock);
+   }
+
+   @Test
+   public void testRunScheduling() throws Exception {
+
+      Runnable runnable = expirationScheduler.runScheduling("123", booleanCompletableFuture, 10, TimeUnit.MILLISECONDS);
```
wburns: I was thinking you could use an argument captor instead. This way you don't have to rewrite your class to have a new method.  You also verify the scheduled executor is called this way. Here is an example: https://github.com/infinispan/infinispan/blob/master/core/src/test/java/org/infinispan/manager/AllClusterExecutorTest.java#L403

karesti: I really hate them haha we abused in the past with them in other projects xD  but why not for this case

lock/src/main/java/org/infinispan/lock/impl/lock/RequestExpirationScheduler.java
karesti: ok

karesti:
@wburns I just pushed a commit on the argument captor. If you are ok on the final result, I amend the commits

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
wburns: Ah well argument captors are great for what they are useful for. And this case is pretty much the case they are designed for.

I am curious as to how you "abused" them because they only have a specific use case normally.

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
```diff
@@ -0,0 +1,121 @@
...
+   }
+
+   @Test
+   public void testAbortSchedulingShouldNotWorkIfRequestIsNotCompleted() throws Exception {
+
+      expirationScheduler.scheduleForCompletion("123", booleanCompletableFuture, 42, TimeUnit.SECONDS);
+      RequestExpirationScheduler.ScheduledRequest scheduledRequest = expirationScheduler.get("123");
+      assertEquals(1, expirationScheduler.countScheduledRequests());
+
+      expirationScheduler.abortScheduling("123");
```
wburns: This test to me seems like a bug if I look at it just from the viewpoint of a service, since if I call `abortScheduling` I would expect it to abort the scheduled service, like what I said in the previous comment. Just saying :)

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+      assertTrue(booleanCompletableFuture.isDone());
+      assertFalse(booleanCompletableFuture.get());
+      assertEquals(0, expirationScheduler.countScheduledRequests());
+   }
+
+   @Test
+   public void testScheduleForCompletionAddsASingleRequestById() throws Exception {
+
+      expirationScheduler.scheduleForCompletion("123", booleanCompletableFuture, 100, TimeUnit.SECONDS);
```
wburns: This also seems like a bug that if we schedule 2 requests on the same id (like my comment on the impl). I would normally expect some kind of error condition as there is no way to then timeout the second or third requests as well.

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
karesti: ok, I will change that

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
karesti: I agree, that it does not fail and it should throw an exception

karesti:
@wburns i amended the commits and made other changes concerning your two points.

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
wburns: I see that you added an additional argument, which is never used :P Not what I was thinking, but like I said I am fine with it.

wburns:

lock/src/test/java/org/infinispan/lock/impl/lock/RequestExpirationSchedulerTest.java
karesti: I agreed with you on the point of not being well documented and API side. I'm know is not used now, but I'm leaving soon for some months and I was wondering if somebody else might need to check and change this after. So this is why I added both ways for now. thanks for the merge!!!

PR #5835, opened 2018-03-13
ISPN-8935 AbstractFileLookup.lookupFileStrict does not work for jar scheme

dmurat:


tristantarrant:
ok to test

tristantarrant:
test this please

jcache/embedded/src/test/java/org/infinispan/jcache/JCacheConfigurationTest.java
```diff
@@ -40,18 +53,96 @@ public void testJCacheManagerWherePathContainsFileSchemaAndAbsolutePath() throws
...
+         if (originalClassLoader != null) {
+            Thread.currentThread().setContextClassLoader(originalClassLoader);
+         }
+      }
+   }
+
+   private void createJar(File jarFile, String existingResourceToAddInJar, String targetPathInJar) throws IOException, URISyntaxException {
+      Manifest manifest = new Manifest();
+      manifest.getMainAttributes().put(Attributes.Name.MANIFEST_VERSION, "1.0");
+      JarOutputStream sampleJarOutputStream = new JarOutputStream(new FileOutputStream(jarFile), manifest);
```
tristantarrant: For the test, the jar file should be created in the tmp directory. See org.infinispan.test.TestingUtil#tmpDirectory(java.lang.Class<?>).

dmurat: Ok, now test tmp folder is used.

tristantarrant:
Squashed and pushed to both master and 9.1.x. Thanks

sammyhk:
Suspected the fix is not correct, considering uri `jar:file:/path/to/my/jar.jar!/infinispan.xml`, [AbstractFileLookup.java#L69](https://github.com/infinispan/infinispan/blob/5d3d2f1c0a3c41d60158d2b771d43abb3d4a5233/commons/src/main/java/org/infinispan/commons/util/AbstractFileLookup.java#L69) extracted `insideJarFilePath = "/infinispan.xml"`. As `getAsInputStreamFromClassLoader(...)` implementation use `ClassLoader.getResourceAsStream(...)` to lookup the resource, any resource starting with "/" actually will always return null.

PR #5838, opened 2018-03-13
ISPN-8938 CacheNotifier should use the cache member and not the view …

pruivo:
…members

https://issues.jboss.org/browse/ISPN-8938


wburns:

PR #5841, opened 2018-03-14
ISPN-8942 Joiners may miss registered listener

pruivo:
https://issues.jboss.org/browse/ISPN-8942


wburns:

PR #5842, opened 2018-03-14
ISPN-8945 ClusterListenerDist(Tx)AddListener random failures

wburns:
* Increased timeouts as it encompasses node add and state transfer

https://issues.jboss.org/browse/ISPN-8945


danberindei:
@wburns I'm proposing https://github.com/infinispan/infinispan/pull/5843 instead

tristantarrant:
So we have to choose whose PR we like best ? :smile:

danberindei:
Something like that, yeah :)

Will was telling me a long time ago that he saw the tests go much faster for him with only 5 concurrent tests, but I always assumed it was related to GC. I initially opened my PR to look at GC activity, but that turned out to be irrelevant, so I changed the number of concurrent threads to 8 instead and it seems to fix the outrageous delays (10 seconds to start a node with 1 regular cache and 1 internal cache really is outrageous).

wburns:
Going from 20m to 13m is a nice improvement, we can close this one :) We can always come back if we find it still occurs.

PR #5843, opened 2018-03-14
Reduce test suite parallelism

danberindei:
https://issues.jboss.org/browse/ISPN-8945

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,7 +176,7 @@
       <infinispan.core.schema.version>${infinispan.base.version}</infinispan.core.schema.version>
       <infinispan.module-suffix>-${project.artifactId}</infinispan.module-suffix>
       <infinispan.test.jgroups.protocol>tcp</infinispan.test.jgroups.protocol>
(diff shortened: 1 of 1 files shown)


danberindei:
@wburns The GC logs don't show sow many pauses as I was expecting, I see at most 2 pauses per second. And the only pauses > 100ms are 2 explicit `System.gc()` calls from `ThreadLocalLeakTest`, each close to a full second but long after `ClusterListenerDistTxAddListenerTest` finished, so I don't think GC activity is an issue here.

Still, you were probably on to something with the number of concurrent tests... with the new internal caches there's a lot more work and less waiting in all those tests. The debug-only core log file is 845MB, only 5 times less than the trace log I got on my laptop, so I'm going to reduce the surefire parallelism and see if it makes a difference.

danberindei:
@wburns with 8 concurrent tests the core test suite takes only 13 mins, and there are fewer test failures, so let's make this change and assume it resolves the failure, even if we know exactly what was causing it.

wburns:
Integrated into master, thanks @danberindei ! Maybe I can run the core test suite a few extra times a day now, haha.

PR #5845, opened 2018-03-15
ISPN-8949 Clear from data container doesn't remove expired entries

wburns:
* Make sure to included expired entries in iteration

https://issues.jboss.org/browse/ISPN-8949

### DIFF
--- a/core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
+++ b/core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
@@ -135,4 +135,16 @@ public class ExpirationFunctionalTest extends SingleCacheManagerTest {
                (k, ice) -> { throw new RuntimeException("No task should be executed on expired entry"); });
       }
    }
+
+   public void testExpiredEntriesCleared() {
+      for (int i = 0; i < 2; i++) {
+         cache.put("key-" + i, "value-" + i,-1, null, i, TimeUnit.MILLISECONDS);
+      }
+
+      // This should expire approximately half of the entries
+      timeService.advance(1);
+
+      cache.clear();
+      assertEquals(0, cache.getAdvancedCache().getDataContainer().sizeIncludingExpired());
+   }
 }
--- a/core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
+++ b/core/src/main/java/org/infinispan/interceptors/locking/ClusteringDependentLogic.java
@@ -179,7 +179,7 @@ public interface ClusteringDependentLogic {
 
       private void commitClearCommand(DataContainer<Object, Object> dataContainer, ClearCacheEntry<Object, Object> cacheEntry,
                                       InvocationContext context, FlagAffectedCommand command) {
-         Iterator<InternalCacheEntry<Object, Object>> iterator = dataContainer.iterator();
+         Iterator<InternalCacheEntry<Object, Object>> iterator = dataContainer.iteratorIncludingExpired();
 
          while (iterator.hasNext()) {
             InternalCacheEntry entry = iterator.next();


core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
```diff
@@ -135,4 +135,16 @@ public void testExpirationMaxIdleInExec() throws Exception {
...
       }
    }
+
+   public void testExpiredEntriesCleared() {
+      for (int i = 0; i < SIZE; i++) {
+         cache.put("key-" + i, "value-" + i,-1, null, i, TimeUnit.MILLISECONDS);
+      }
+
+      // This should expire approximately half of the entries
+      timeService.advance(SIZE / 2);
```
danberindei: Do we really need to test with more than 1 entry?

danberindei:
Maybe we should remove the expired entries from the data container, but skip the listener notification?

core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
wburns: I wanted to test both expired entries and entries that can expire being removed. I can do 2 if you want and it still holds that.

wburns:
We can't skip the listener notification. We have to either notify them of expiration or removal. It is much easier to just notify them of removal since we are removing everything anyways, so I stuck with that.

wburns:
Updated to reduce inserts to 2 instead of 10.

karesti:
@wburns LGTM

karesti:
@danberindei @wburns the org.infinispan.manager.CacheManagerAdminSharedPermanentTest.testClusterCacheTest

is not related, right ?

danberindei:
@wburns I think it has the opposite problem now (entry expires and is purged from the data container between `iterator.next()` and `dataContainer.remove()`, triggering both notifications), but it's probably not that important.

wburns:
@danberindei The iterator.next doesn't cause an expiration event to occur. Only a `Cache.get` or purge thread can cause the expiration, iteration/streams etc don't cause expiration events. So the chance of having both at the same time on the same key is ridiculously low. And to be honest the clear is done non transactionally so you can have duplicate/incorrect events with write operation.

core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
```diff
@@ -135,4 +135,16 @@ public void testExpirationMaxIdleInExec() throws Exception {
                (k, ice) -> { throw new RuntimeException("No task should be executed on expired entry"); });
       }
    }
+
+   public void testExpiredEntriesCleared() {
+      for (int i = 0; i < 2; i++) {
+         cache.put("key-" + i, "value-" + i,-1, null, i, TimeUnit.MILLISECONDS);
+      }
+
+      // This should expire approximately half of the entries
```
danberindei: This isn't true any more. I'd rather remove the loop, and if you want one expired an one non-expired entry insert each of them explicitly. If having all entries expired is enough, then 1 entry is all you need.

danberindei:
@wburns could you open a JIRA for the test failure @karesti asked about? I don't see it failing in master, but I don't think it's related to this change either.

core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
wburns: What isn't true? The fact that it isn't approximate? It will expire 1 of the 2 entries always now.

You are really stuck on this 1 entry thing :D I am just curious, but why? Did you want me to change all the other tests to 1 as well, instead of 10?

wburns:
Yes the test is not even remotely related (guessing it it is a sign that the given node hasn't joined the cache view yet, but is in the JGroups view).

wburns:
Done https://issues.jboss.org/browse/ISPN-8966

core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
danberindei: I was thinking expiry <= now for both of them so both are expired. But you're right, only `InternalMetadataImpl` uses `<=`, regular entries use `<`, so the 2nd entry would still be alive.

I'm not against having multiple keys, I just see a for loop and I try to understand its purpose. Expiration is completely deterministic, so I try to figure out if there's another source of non-determinism in the test. And my memory is only getting shorter, so I do the same thing every time I read the PR :)

And now the history bit... a long time ago tests used lots of keys to counter the randomness of the CH, hoping that one of the keys will land on the "proper" nodes after a topology change. Nowadays we have `ControlledConsistentHashFactory`, making things much more deterministic, so the vast majority of tests would work just fine with a single key. But we still have lots of tests using 10 or 20 or 100 keys without any comment to explain how that particular number was chosen.

wburns:
Also your memory is a bit fuzzy :) https://issues.jboss.org/browse/ISPN-8617 was created as well - most likely the exact same cause.

core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
wburns: Okay I understand, I will try to keep that in mind for the future. To me a clear should at least remove a few entries to be somewhat tested :)

karesti:
integrated thanks @wburns !

core/src/test/java/org/infinispan/expiration/impl/ExpirationFunctionalTest.java
danberindei: That would be true if clear had to do something different for the first (or the last) entry, but here code coverage is exactly the same with 1 or 2 entries. The data container does have one live and one expired entry, but the loop makes it harder to understand that, even with the comment. So to me it looks like the test is trying very hard to be "interesting", when it should be trying to be as bland as possible.

Reminds me of when Mircea wanted to use my `Sequencer` stuff in a test. He was commenting on how it makes the test look very fancy, and I knew immediately that I was on the wrong path. The test still uses `Sequencer` though, I haven't had the time to make it simpler yet :D

wburns: I had written a whole response on here, but we are obviously way past the point, so I removed it. I will just end with if you are comparing this with `Sequencer` then ......

wburns: But in regards to the loop, I agree. It is much more clear to have 2 inserts instead of a loop. I am not sure why I didn't change it at the time, probably just because it was easier to change a couple characters :)

danberindei: I only mentioned `Sequencer` because I thought it was a funny sidenote, please don't take it to mean it's in any way comparable with a for loop :)

PR #5846, opened 2018-03-16
ISPN-8958 NPE in JGroupsTransport.send(..) when stopping

rhusar:
Jira
https://issues.jboss.org/browse/ISPN-8958

Mostly just reducing log noise.


ryanemerson:
@rhusar thanks!

PR #5848, opened 2018-03-16
ISPN-8959 Infinispan occasionally fails to locate an entry in a cache store when using a bounded data container

pferraro:
https://issues.jboss.org/browse/ISPN-8959

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/PersistenceUtil.java
+++ b/core/src/main/java/org/infinispan/persistence/PersistenceUtil.java
@@ -88,7 +88,8 @@ public class PersistenceUtil {
             return oldEntry; //no changes in container
          }
 
-         MarshalledEntry loaded = loadAndCheckExpiration(persistenceManager, k, ctx, timeService);
+         // Load using key from command
+         MarshalledEntry<K, V> loaded = loadAndCheckExpiration(persistenceManager, key, ctx, timeService);
          if (loaded == null) {
             if (isLoaded != null) {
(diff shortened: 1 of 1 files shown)


wburns:
Changes are harmless enough to integrate. The thing I worry about is what other ramifications this could have on other operations as well. I need to look into this more.

danberindei:
@wburns looks like we need a `DefaultDataContainerTest`...

wburns:
@ben-manes that was my conclusion as well. We 100% do not invoke weak keys, so I am not sure exactly what @pferraro was seeing. Unfortunately he wasn't able to make a simple reproducer. So for now I have just integrated his trivial fix (shouldn't cause other issues) until we can look into it further. 

@pferraro Maybe if you could add more tracing in ISPN directly so when the error occurs we can get more information, would be best.

pferraro:
@ben-manes I am definitely not use weak keys either (since we upgraded to Infinispan 9.2, we no longer build our own Caffeine instance, but even this, did not use weak keys), and the keys in question are immutable.  However, some doubt has crept in, so let me double check that our hashCodes are not somehow getting mangled following serialization/deserialization.

What I can say is, when I was able to trigger the issue, the assertion line in the linked jira reported the following:
"Expected key type org.wildfly.clustering.ejb.group.InfinispanBeanGroupKey but compute action uses java.lang.Object"

@wburns I'll file a separate PR with the requested trace logging.

PR #5850, opened 2018-03-20
Hibernate-cache: Fix deprecation warnings from XML

erikkemperman:
I am seeing deprecation warnings that I believe are coming from these XML files.


erikkemperman:
Sorry, I submitted this prematurely -- guess it wasn't coming from this XML after all.

PR #5852, opened 2018-03-20
Add debug logging when ISPN-8959 is detected.

pferraro:
https://issues.jboss.org/browse/ISPN-8959

@wburns As requested by https://github.com/infinispan/infinispan/pull/5848#issuecomment-374579388

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -372,6 +372,11 @@
          </dependency>
          <dependency>
             <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-hibernate-cache-spi</artifactId>
+            <version>${project.version}</version>
+         </dependency>
+         <dependency>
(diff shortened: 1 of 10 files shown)


wburns:
I meant more that you could try to get more info in your issue that reproduces it Paul. But there is no reason we can't run this in CI and see if it finds anything. Although I believe this equality check isn't quite right as we could have a serialized key that will always fail this check.

pferraro:
@wburns This is the same logging I used to find the issue.  I assumed you wanted to run with this logging in your CI.
> Although I believe this equality check isn't quite right as we could have a serialized key that will always fail this check.
When is this the case?  When StorageType is binary or off-heap?  Are there other scenarios?

danberindei:
So what's the conclusion @pferraro @wburns ?

PR #5854, opened 2018-03-20
ISPN-7066 NullPointerException fix

pruivo:
The previous fix was incomplete.

https://issues.jboss.org/browse/ISPN-7066


slaskawi:
BTW having a test for it would be nice...

PR #5856, opened 2018-03-20
ISPN-8868 exposes JMX methods

karesti:
https://issues.jboss.org/browse/ISPN-8868

* force release
* remove
* isDefined
* isLocked

### DIFF
--- a/lock/src/main/java/org/infinispan/lock/impl/ClusteredLockModuleLifecycle.java
+++ b/lock/src/main/java/org/infinispan/lock/impl/ClusteredLockModuleLifecycle.java
@@ -11,6 +11,7 @@ import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.configuration.global.GlobalConfiguration;
 import org.infinispan.factories.GlobalComponentRegistry;
+import org.infinispan.jmx.CacheManagerJmxRegistration;
 import org.infinispan.lifecycle.ModuleLifecycle;
 import org.infinispan.lock.api.ClusteredLockManager;
 import org.infinispan.lock.impl.entries.ClusteredLockKey;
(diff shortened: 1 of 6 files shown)


lock/src/main/java/org/infinispan/lock/impl/ClusteredLockModuleLifecycle.java
```diff
@@ -93,6 +94,10 @@ private static void registerClusteredLockManager(GlobalComponentRegistry registr
          if (counterManager == null || !(counterManager instanceof EmbeddedClusteredLockManager)) {
```
pruivo: `counterManager`?

karesti: this is copy/paste

lock/src/main/java/org/infinispan/lock/impl/manager/EmbeddedClusteredLockManager.java
```diff
@@ -96,14 +118,25 @@ public ClusteredLockConfiguration getConfiguration(String name) {
...
+      if (trace) {
+         log.tracef("LOCK[%s] isDefined has been called", name);
+      }
       if (cache == null) {
          cache = extractCacheHolder(cacheHolderFuture).getClusteredLockCache();
       }
       return cache.containsKey(new ClusteredLockKey(ByteString.fromString(name)));
    }
 
+   @ManagedOperation(
```
pruivo: you can't reuse this methods. 
`CompletableFuture` isn't serializable and the operation will fail (in JConsole at least) with `NotSerializableException`

pruivo:
since you are exposing some operation, don't you want to expose the remaining one:
* is-locked
* configuration
* get lock names (?)

karesti:
@pruivo yeah, maybe. I did the minimum asked but why not

lock/src/main/java/org/infinispan/lock/impl/manager/EmbeddedClusteredLockManager.java
karesti: ok, I need to make them sync ?

karesti:
@pruivo I did the modifications and added a two additional methods that might be useful

karesti:
@pruivo test failures unrelated ...

pruivo:
integrated! thanks @karesti !

PR #5857, opened 2018-03-20
Conflict resolution not initiated if node rejoins with same topology

ryanemerson:
https://issues.jboss.org/browse/ISPN-8903
https://issues.jboss.org/browse/ISPN-8954

### DIFF
--- a/server/hotrod/src/test/java/org/infinispan/server/hotrod/HotRodMergeTest.java
+++ b/server/hotrod/src/test/java/org/infinispan/server/hotrod/HotRodMergeTest.java
@@ -13,6 +13,7 @@ import static org.infinispan.server.hotrod.test.HotRodTestingUtil.startHotRodSer
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 import org.infinispan.configuration.cache.CacheMode;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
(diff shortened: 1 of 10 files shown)


danberindei:
@ryanemerson I've been working on a `PreferAvailabilityStrategy` unit test for https://github.com/infinispan/infinispan/commit/07560797d619d9f2af2fa8c4c4db6223cea0201f, but that doesn't touch conflict resolution. Maybe you could add something based on that? I'd rather not have to run the HR client tests all the time ;)

server/hotrod/src/test/java/org/infinispan/server/hotrod/HotRodMergeTest.java
```diff
@@ -92,12 +94,15 @@ public void testNewTopologySentAfterOverlappingMerge() {
       int initialTopology = advancedCache(0).getRpcManager().getTopologyId();
       expectCompleteTopology(client, initialTopology);
       PartitionDescriptor p1 = new PartitionDescriptor(0);
+      // isolatePartitions will always result in a CR fail as Node 0 tries to contact Node 1 in order to receive segments
+      // which is not possible as all messages received by Node 1 from Node 0 are discarded by the DISCARD protocol.
+      // Therefore, it is necessary for the state transfer timeout to be < then the timeout utilised by TestingUtil::waitForNoRebalance
```
danberindei: Maybe I'm not reading this right, but it sounds like a serious limitation of conflict resolution.

If the cluster splits, each partition may start its own conflict resolution, and it should only wait for nodes that are still in the same partition (i.e. in the same JGroups cluster view).

server/hotrod/src/test/java/org/infinispan/server/hotrod/HotRodMergeTest.java
```diff
@@ -67,7 +68,8 @@ protected void destroy() {
    @Override
    protected void createCacheManagers() {
       ConfigurationBuilder dcc = hotRodCacheConfiguration(new ConfigurationBuilder());
-      dcc.clustering().cacheMode(cacheMode).hash().numOwners(1);
+      dcc.clustering().cacheMode(cacheMode).hash().numOwners(1)
+            .stateTransfer().timeout(1, TimeUnit.SECONDS); // Must be less than timeout used in TestingUtil::waitForNoRebalance
```
danberindei: IMO the comment would look better on the previous line

ryanemerson:
Updated with tests. The only change to non-test code is that the CacheTopology `.equals` and `.hashcode` implementation now takes into account the topology phase.

danberindei:
Thanks Ryan, integrated!

PR #5787, opened 2018-02-22
ISPN-8621 Share connection for events

rvansa:
https://issues.jboss.org/browse/ISPN-8621

Set to preview as this depends on #5742

PR #5858, opened 2018-03-20
ISPN-8981 Generate Hot Rod parser automatically

rvansa:
https://issues.jboss.org/browse/ISPN-8981

Still preview as it depends on #5787 Note that the last commit removes almost 1000 lines of code.

Notable points are the meta-grammar [1] and Hot Rod grammar [2] itself.
[1] https://github.com/infinispan/infinispan/pull/5858/files#diff-aa04dd5a2a8820bb34632ad1d9abdf68
[2] https://github.com/infinispan/infinispan/pull/5858/files#diff-5cecd3e1f4a0ac9a7b6ad32a06e0d64d


rvansa:
run performance tests please
cs-hotrod
[UDP 1kB]

diegolovison:
The results seem promising to me but I will check why it is being compared with a snapshot from 01-30-2018

rvansa:
@diegolovison Thanks for starting the 10kB run - I've updated the execution numbers in the comparison and it's not as promising anymore.

rvansa:
run performance tests please
cs-hotrod
[UDP 1kB]

rvansa:
run performance tests please
cs-hotrod
[UDP 1kB]

rvansa:
run performance tests please
cs-hotrod
[UDP 1kB]

rvansa:
run performance tests please
cs-hotrod
[UDP 1kB]

rvansa:
The failures in CI look like some quota on GitHub usage is reached (therefore I've retried); I can access the problematic URL without issue.

rvansa:
run performance tests please
cs-hotrod
[UDP 1kB]

galderz:
Just had a quick look. I like the idea but I'm worried about this:

```
* The syntax of grammar is intentionally close to ANTLRv4 syntax despite it's neither
* a subset nor limited to that.
```

A couple of things worry me here:

1.  Who is going to maintain this custom stuff. 

2.  Long term usability. For this to be really useful IMO, you'd have to be able to use it to generate client code as well. Generating server code is easy since it's only one language as target. Generating client code is much harder and more interesting for us. I'd rather we spend the cycles in something that helps us with both client and server, e.g. gRPC (not limited to it, but a strong candidate. @rigazilla is doing some research here).

Looking at the JIRA:

> Manual parsing of Hot Rod protocol, invoking recursive methods that return {{Optional}}s or {{Optional<Optional<...>}}s seems to generate a lot of garbage.

^ Using Optional like this is probably over the top. I've always be in favour of exposing APIs with Optionals rather than returning null values, but internally we should strive for less garbage. There's no need to come up with a grammar to fix this ;)

anistor:
The concept of using a FSM is very nice, but as Galder pointed out I'm too a bit concerned about the added effort of maintaining the tooling, especially because the tool is just for a single use :).
Reminds me of our dependency injection framework, which looked good on paper initially but has become a bit of a drag.

rvansa:
About maintenance: I would like to donate that part to Netty, since it's reusable. 

I don't think that client code generator would be *that* useful: writing request down is dead-simple and you know what response is expected. To me it seems that the exception-based approach utilized in `ReplayingDecoder` works well on client side.

The server is different in this: we have higher pressure to not generate any garbage and the handling contains a big operation switch. I was told that there was the experience with ReplayingDecoder that showed that it does not work well there, so I was looking for an always-forward alternative.

The optional hell was the result of natural hierarchic parsing. We needed (at least) three states of return values; no state is even better though. That's why I went with grammar.

@rigazilla Does gRPC offer any netty integration?

galderz:
> @rigazilla Does gRPC offer any netty integration?

AFAIK, gRPC uses netty underneath

galderz:
> I don't think that client code generator would be that useful: writing request down is dead-simple and you know what response is expected.

I disagree, it's not as simple as you think. It sounds simple but does take a while to get it right. I have implemented a Node.js client and @rigazilla has implemented the C/C++/C# stuff, so we have a fairly large experience doing this and it's not really that trivial IMO.

Client generators cannot do everything but can at least help with the request/response and streaming parts. You'd still have to implement things on top of the client like near cache, handling of listeners...etc, but if a solid, proven, library can help us with the client code generation, it would allow us to more quickly churn clients, add operations more quickly and in general evolve the client/server architecture faster.

rvansa:
@galderz OK, generating clients for other languages certainly is useful. What I meant to say is that developing tooling to generate our *Java* (Netty) client wouldn't be useful (and we don't want to maintain generators to different languages).

+1 to gRPC clients if these can be as efficient as the hand-written one.

Btw. I've filed https://github.com/netty/netty/issues/7835 couple of days ago, but I haven't seen Norman on IRC since (I've noticed he had some LinkedIn updates so he's probably busy).

rvansa:
@galderz OK, so Norman considers this rather 'extra' that should not go to the main codebase. That leaves me with three options:
A) start a new project, do a release and just use that as a dependency
B) keep the code in Infinispan
C) kill the PR completely

gustavonalle:
Parsing the Hot Rod protocol is extremely tedious so I like the idea of having it automatically generated :+1: 

The part I am not so sure is generating it from an ANTLR-like grammar, I'd prefer the protocol definition to be an [IDL](https://en.wikipedia.org/wiki/Interface_description_language), like protobuf.

rvansa:
@gustavonalle Do you think we could retrofit a protobuf IDL to Hot Rod? (I doubt so, @anistor ?) If I were to rewrite the generator I'd use something standard.
Can any IDL you know embed the request header well into the message?
Also, I don't want to create any *request object*, that's an unnecessary allocation.

rigazilla:
@rvansa @gustavonalle I don't think that protobuf or grpc are proper tools to specify Hotrod intended as a binary protocol: protobuf is about specifying objects that need to be streamed and grpc specify procedure call.
Protobuf has it's own bits format and can't be forced to produce hotrod stream (even sending the first magic byte is a trouble).

Probably the best tool to specify Hotrod bits is the grammar, but in this case I can't evaluate if the result worth the effort.

As @galderz said it would be great if the resulting architecture could support "choose your language" developers on the client side also.

My experiments with grpc [here](https://github.com/rigazilla/ispn-grpc/) are addressing that point but my approach is to specify Hotrod at the api level and leave the grpc/protobuf framework generated the bitstream.

rigazilla:
@rvansa @galderz yep I'm quite sure gRPC has netty inside, but we are still talking about Java only :)

rvansa:
@rigazilla @galderz For clients, we need something existing, because we're not going to maintain Pascal client code generator. Is there anything that could fit Hot Rod? Because if there is not the discussion about this PR (server generator) and client generators is irrelevant.

gustavonalle:
@rigazilla I've seen a few cases around where protobuf is used to define a binary protocol, one of such is http://pulsar.incubator.apache.org/docs/v1.21.0-incubating/project/BinaryProtocol/ which is a messaging system. I think in terms of complexity it's roughly comparable to Hot Rod, worth taking a look.

They mention lack of framing in protobuf, so they add their own custom header. 

Btw, Hot Rod is already using protobuf for the [query operation ](https://github.com/infinispan/infinispan/blob/master/remote-query/remote-query-client/src/main/resources/org/infinispan/query/remote/client/query.proto) :)

rigazilla:
@rvansa we probably need a definition for "fit Hot Rod", but I think I get your point. Streaming the bits is boring but is not the heavy part of a client dev, so having the parser for free from the grammar surely helps but I don't think it makes a big difference.
As @gustavonalle said I would be more interested in a IDL spec for the client/server communication but, since I consider this incompatible with hot rod grammar, this is probably OT  here.

rvansa:
@tristantarrant So... the verdict? Or any more opinions needed?

galderz:
> @tristantarrant So... the verdict? Or any more opinions needed?

Before deciding whether to integrate this or not, I'd like the approach of generating a Protobuf IDL for Hot Rod protocol to be investigated fully. IIRC, numbers in Hot Rod are encoded using same encoding as Protobuf but I can't say 100% sure. If we can really get that working, I feel it'd be a better approach.

rigazilla:
> Couldn't a Hot Rod request, in all its entirety, be seen as an object that needs to be streamed/sent?

@galderz my opinion on this is that protobuf can't be easily forced to produce a valid hotrod message. I think is impossible, actually.

Every protobuf field is streamed with three info: the value, the type and the id.
Look at the example [here](https://developers.google.com/protocol-buffers/docs/encoding).

The main problem I see is that every basic field specified in our HotRod Request message must be preceded by the protobuf prefix containing the type and the id info: this is by design.

So you're right that hotrod and protobuf uses the same variable size integer format, but let's try to send the Message ID value 0x01: the HotRod bit stream is just the 0x01 byte. With protobuf we must define a field in our message, something like:
    int32 messageid = 2;

this generates the protobuf bit stream 0x10 0x01.

We're even in troubles if we just try to provide a protobuf definition for the magic byte 0xA0, we can't sent 0xA0 as first byte there must be the protobuf preamble.

I can't see a working approach to fit HotRod into a protobuf message, but if someone want to try I can provide my support if needed.

More in general I think it's hard to find an IDL or RPC framework that let us specify the underlying bitstream, but maybe I'm wrong on this.

My opinion is that @rvansa approach is correct, the grammar is the right way to specify a protocol at the binary level.

But I can't say whether the generated code must be merged or not.

rvansa:
Another 'standard' IDL is Apache Avro, which does not use the fields in the binary stream. I have no experience with that, though.

galderz:
@rigazilla Thanks for the detailed explanation. 

I have more twist on this: what about retrofitting existing Hot Rod protocol so that a protobuf IDL can be defined for it? What I mean by retrofiting is: morphing the protocol (release a new version of it) adding the necessary fields so that it could be defined via a protobuf IDL? Thoughts?

rvansa:
@galderz That could make sense if you were to drop support for 1.x and 2.x soonish. But we want to keep parsing them, so while the new protocol could use more performant parser, we'd have to keep the old parsing codebase.

galderz:
I don't we'll be dropping those soon, it'd depend on product lifecycles really to decide when older protocols get removed.

rigazilla:
@galderz 

> about retrofitting existing Hot Rod protocol 

the easiest solution I can propose for this is just enlist the hotrod fields in a .proto file, something like this:

    message quasi-hr {
        fixed32 magic = 1;
        int64 messageId = 2;
        fixed32 version = 3;
        fixed32 opcode = 4;
        ....
    }

Basically we got an extra byte preamble for every field, 4 bytes instead of 1 byte field (my example with the magic byte in the previous comment was bad, protobuf sends at least 4 bytes) ... and other differences.
Things will probably gets more tricky in defining the operation's specific data.

This could be useful if we need to generate parser for client in different language (assuming that we provide the server with a quasi-hr/hotrod filter), but honestly I would prefer a clean IDL with operations and arguments well defined, in this way instead we have everything melted in the message.

galderz:
I'd also prefer a clean IDL moving forward. FYI, @rigazilla has been looking a potential GRPC based Hot Rod version for the future. 

So, is adding all this parsing complexity worth the effort at this point? I know there's some performance gains to be had, but I'm worried about having to maintain this extra parsing code.

My personal gut feeling is that we should address the performance or memory consumption issues alone and don't spend more time on the parsing side, since having our own parsed Hot Rod protocol is not something we'd want going forward.

galderz:
What is everyone else thinking? I think this might need agreeing in the dev list too.

anistor:
It is not possible to use the protobuf wire protocol to encode messages compatible with the current version of the Hot Rod protocol. Our approach for defining the Hot Rod encoding is based on a fixed order of fields, with fixed field widths and offsets, and the eventual possibility that some fields are optional by convention - their presence being indicated by the value of a marker/discriminator field that was previously read from the data stream. Our approach has the benefit of packing the data as tightly as possible but suffers a bit on the flexibility side. But even that is debatable because we do not change the protocol every day.

Protobuf uses a diametrically opposed approach: field order is flexible (read: not guaranteed!), all fields can be optional without having to resort to the marker field gimmick. This is very flexible but comes at a cost: all fields are prepended with a _tag_. This is a fixed width int that encodes the schema number of the following field and its wire type. From the wire type we are able to tell the length of the field also.  So a decoder could skip it if it does not understand what field number 777 means (because it uses an older schema -- schema evolution baby!!).  This flexibility is superb for user payload. The users typically don't know what they want. They change the schema of their entities mid-project and want older data to survive :) But is not so badly needed for protocols which are supposed to be more stable. And it comes at a cost. The presence of that _tag_ that guarantees this flexibility has the potential to double the size of some small messages and the lack of fixed field ordering impacts speed of decoding and is not terribly memory efficient because we now have to also keep track of which fields were 'seen' and which were not.

So if somebody tells you he has a protobuf library that can output messages compatible with our current version of the Hot Rod protocol spec, it's just fake news, move along :).

Can we make a future version (completely backward incompatible) of the Hot Rod protocol using protobuf, yes we can but it's another story. First we need to ask ourselves why.

Can we make the current Hot Rod use a protobuf 'IDL' for generate the parser? Yes, in theory, but we would be only borrowing the IDL syntax, so in practice we would have to recreate from scratch a tool to process it, because whatever comes with your favourite protobuf library will not generate a suitable encoder. Do we want one more tool? :)

tristantarrant:
This is the decoder as generated by this PR: https://gist.github.com/tristantarrant/5249f40e37e8518c89d89a0cb4b4912d

anistor:
The generated code looks quite unfriendly for maintaining it manually or debugging.

tristantarrant:
Just like the query parser

anistor:
True, but the query parser is somebody else’s fault :) Blame ANTLR and move to something else. 
Here the situation is a bit different as we have to maintain the generator too :)

rvansa:
@anistor So unfriendly in what way, and how can we improve that? The parser is a stateful machine. So it's got states and a big switch. Each state in the switch has comment on the beginning describing where in the grammar you are (the grammar is non-recursive).
Hand-written parsers often use recursion but that's exactly the thing I wanted to avoid, in order to be able to stop and continue anywhere.
Actually one thing that is unfriendly is the multi-level switch, but that's necessary because JVM does not like too long methods (hard limit is 64k bytes, but optimization suffers even on smaller methods).

danberindei:
@anistor I don't think it's *that* unfriendly, although I mentioned earlier on Zulip that the inconsistent spaces around the brackets (and other punctuation) hurt my eyes :)

:+1: for the switch from me, although instead of splitting the cases into 4 switches I'd rather have a single big switch with a separate method for each state. Not sure if that's even possible with your generator @rvansa, but at the very least I'd like this sub-switch to be generated in a separate method: https://gist.github.com/tristantarrant/5249f40e37e8518c89d89a0cb4b4912d#file-hotroddecoder-java-L274

:-1: to check the readerIndex explicitly in the switch, I feel it's a higher layer that should only deal with intrinsics. Instead of

    pos = buf.readerIndex();
    hr2x_string = org.infinispan.server.hotrod.Intrinsics.string(buf);
    if (buf.readerIndex() == pos) return false;

I'd like something like

    hr2x_string = org.infinispan.server.hotrod.Intrinsics.string(buf, unfinishedCallback);
    if (unfinished()) return false;

Sorry if I'm completely missing what's possible and what you were trying to do with the generator @rvansa :)

danberindei:
One more question @rvansa, maybe maintenance would be easier if the generated decoder was committed to git and every PR that changes the grammar would also include the changes to the generated code?

rvansa:
@danberindei The hierarchy of state switches vs. the opcode switch are two completely orthogonal concepts. As for the former: even if each case consists only of
```
case X:
   return runStateX();
```
we could reach the bytecode limits. Moreover, such structure could turn problematic for the fall-through cases.

I could move the opcode switch to a separate method, if you prefer that variant.

I won't bother being smarter with spaces in the generated code.

About reader index: intrinsics are user code. The generated parser needs to record the position before invoking the intrinsic. We could either record it directly or call `buf.markReaderIndex()`. I don't recall why I used separate variable, but I don't think it matters.
And I don't want to introduce any extra layers.

rvansa:
@danberindei About committing the generated code: You're not supposed to read the generated code. It is debuggable, but it's not up to reviewer to check the state transitions. Note that with each change the states get renumbered.

rvansa:
Rebased. Generated parser is here: https://gist.github.com/rvansa/898af856c1b389b9a62a7fc7cc8fe3f5

rvansa:
run performance tests please
cs-hotrod
[UDP 1kB]

rvansa:
@diegolovison Don't worry about the failure, there are some test failures as well.

danberindei:
@rvansa I can't debug code without reading it, sorry :)

I really want us to see the parser generator as something to be improved in the future, not as something that we should never mess with because it's good enough. The cleaner the generated code looks now and the clearer the purpose of each generated line, the more someone will be willing to work on the generator and grammar definition. 

True, one could spend way too much time tracking down the last whitespace, but a line like this really doesn't inspire confidence:

    @ Override protected HotRodHeader getHeader (){

rvansa:
@danberindei I wasn't talking about debugging the code, but about reviewing.

server/hotrod/src/main/resources/hotrod.g
```diff
@@ -0,0 +1,343 @@
+// Hot Rod 2.x protocol
```
anistor: The *.g file extension is picked up by antlr plugin and highlights the whole file as invalid. Can we use a different extension to avoid the clash?

anistor:
> About committing the generated code: You're not supposed to read the generated code. It is debuggable, but it's not up to reviewer to check the state transitions. Note that with each change the states get renumbered.
@rvansa As @danberindei alread said before, and I'm going to repeat it, we do need to be able to read the generated code and understand it. No, I'm not suggesting to commit it in git and ever edit it manually. But I would like to be able to see how my changes to the input grammar affect the output. And would also like to have stable state numbers so I can diff with a previous run and see those changes.

danberindei:
@anistor stable state numbers would be nice for "debugging" grammar/generator changes, but I wouldn't block this PR for them. I'm more interested in making the code easier to understand, so if I decide in a couple months that I want stable state numbers or some other feature I won't be put off by the fact that I don't understand the generated code.

@rvansa you may have other expectations, but for me reviewing a change to a big switch in a regular Java file doesn't sound harder than reviewing a change to grammar file in a language that I already forgot :)

bom/pom.xml
```diff
@@ -294,6 +294,11 @@
             <artifactId>infinispan-server-core</artifactId>
             <version>${project.version}</version>
          </dependency>
+         <dependency>
+            <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-server-automaton-maven-plugin</artifactId>
```
danberindei: I'm not sure we need it in the BOM, maybe only in the parent.

server/automaton-maven-plugin/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+        <version>9.4.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.infinispan</groupId>
+    <artifactId>infinispan-server-automaton-maven-plugin</artifactId>
+    <packaging>maven-plugin</packaging>
+    <version>9.4.0-SNAPSHOT</version>
+    <name>Automaton Maven Plugin</name>
+    <description>This plugin should eventually be moved to Netty codebase</description>
```
danberindei: Not really a description :)

anistor:
It's not just nice @danberindei. Reproducible builds are mandatory, so I'd say it's a blocker. The algorithm that generates the FSM out of the grammar is deterministic. So stable state numbers should be trivial to do right now.

rvansa:
@anistor The numbers are stable as long as the grammar does not change, but once you add some rule in the middle it gets renumbered.
I'll use `.gr` extension to not cause inconvenience to you.
I am working on some pretty printing right now to ease of Dan's beautiful sore eyes...

rvansa:
I've fixed the client-hotrod failures and improved the formatting in the decoder. I've also moved the `HashMap` and `ArrayList` allocations to `BaseDecoder.allocXXX` so that we use `Collections.emptyXXX` if there are 0 params or entries or whatever.

server/automaton-maven-plugin/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+        <version>9.4.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.infinispan</groupId>
+    <artifactId>infinispan-server-automaton-maven-plugin</artifactId>
+    <packaging>maven-plugin</packaging>
+    <version>9.4.0-SNAPSHOT</version>
+    <name>Automaton Maven Plugin</name>
+    <description>Generates a parser Netty's ByteBuf for provided grammar</description>
```
anistor: Almost makes sense :))

server/automaton-maven-plugin/src/main/java/org/infinispan/server/automaton/generator/Action.java
```diff
@@ -0,0 +1,130 @@
...
+            array = "[";
+         }
+         return action.get(1) + array;
+      }
+
+      if (action.size() == 1) {
+         String token = action.get(0);
+         if (token.equals("true") || token.equals("false")) {
+            return "boolean";
+         } else if (token.matches("[0-9]*L")) {
```
anistor: small L suffix for longs should also work in java.

server/automaton-maven-plugin/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <parent>
+        <groupId>org.infinispan</groupId>
+        <artifactId>infinispan-server-parent</artifactId>
+        <version>9.4.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.infinispan</groupId>
+    <artifactId>infinispan-server-automaton-maven-plugin</artifactId>
```
anistor: Could we name this "infinispan-server-protocol-parser-generator-maven-plugin" and generally avoid the 'automaton' name (which is vague) and call these things 'protocol-parser' ?

rvansa: I don't insist on the current name, but 'protocol-parser' sounds even more vague to me :) I'd like to stress the fact that it uses a state automaton (strictly speaking it's not FSM, though). Maybe `infinispan-server-automaton-protocol-parser-generator-maven-plugin` but I wouldn't want to type that on command line :)

anistor: Protocol parsing is what it does. The automaton is just an implementation detail, that we do not need to stress, imho.

anistor: So this generates the server side of the protocol parser. Any chance to extend it to also generate the client side?

rvansa: @anistor https://github.com/infinispan/infinispan/pull/5858#issuecomment-378212962

> I don't think that client code generator would be that useful: writing request down is dead-simple and you know what response is expected. To me it seems that the exception-based approach utilized in ReplayingDecoder works well on client side.

Unless we introduce some very complex responses I don't think it's worth it.

Anyway, if I am renaming this I could omit the `server` and call it `infinispan-protocol-parser-generator-maven-plugin`, putting this out of the server directory.

danberindei: +1 to move it out of the server

I had to enable `SingleByteFrameDecoderChannelInitializer` in all the tests to weed out replay bugs in the server. I see the client tests don't use it, so there may be some replay bugs lurking, even with the responses being simpler than the requests. I'm not particularly fond of `SingleByteFrameDecoderChannelInitializer`, because it generates lots of extra trace logs, so being able to test replay in one specific test and ignore in the others is a big plus of the parser generator for me, for the client and the server.

rvansa:
Ok guys, moved/renamed.

anistor:
I see the grammar file ends up in the root of final jar.

anistor:
Oh, I see you moved it to top level, but I think we have our maven plugins under 'plugins'

rvansa:
Oh, I didn not notice that we already have that - I have checked only tools (and that's a module).
Moved to plugins/ and excluded the grammar file from final jar.

anistor:
Thanks @rvansa ! I'm happy with it now.

danberindei:
Push the button @anistor!

gustavonalle:
@rvansa @anistor @danberindei there are test failures

tristantarrant:
Those LinkageErrors with hibernate search  / lucene.... @ryanemerson

server/hotrod/pom.xml
```diff
@@ -199,6 +208,50 @@
...
+            <executions>
+               <execution>
+                  <id>add-source</id>
+                  <phase>generate-sources</phase>
+                  <goals>
+                     <goal>add-source</goal>
+                  </goals>
+                  <configuration>
+                     <sources>
+                        <source>${project.build.directory}/generated-sources/java/</source>
```
danberindei: I think you can do the same thing build-helper-maven-plugin does in your plugin:

https://github.com/mojohaus/build-helper-maven-plugin/blob/master/src/main/java/org/codehaus/mojo/buildhelper/AddSourceMojo.java#L64

gustavonalle:
The failures are not related apparently

server/hotrod/pom.xml
```diff
@@ -199,6 +208,50 @@
...
+            <version>${project.version}</version>
+            <configuration>
+               <files>
+                  <file>${basedir}/src/main/resources/hotrod.gr</file>
+               </files>
+            </configuration>
+            <executions>
+               <execution>
+                  <goals>
+                     <goal>generate</goal>
```
danberindei: I wonder if we can specify a phase here so that the plugin isn't needed for `mvn clean`. I usually run the build through a script that does `mvn clean` first, and it's annoying when it stops working because I don't have the plugin in my local repo (e.g. after a version change)

server/hotrod/pom.xml
```diff
@@ -19,10 +19,15 @@
...
    <dependencies>
-       <dependency>
+      <dependency>
          <groupId>org.infinispan</groupId>
          <artifactId>infinispan-server-core</artifactId>
       </dependency>
+      <dependency>
+         <groupId>org.infinispan</groupId>
+         <artifactId>infinispan-protocol-parser-generator-maven-plugin</artifactId>
+         <scope>provided</scope>
```
danberindei: I don't think this dependency is needed

server/hotrod/pom.xml
danberindei: This worked:

               <execution>
                  <goals>
                     <goal>generate</goal>
                  </goals>
                  <phase>generate-sources</phase>
               </execution>

server/hotrod/pom.xml
rvansa: I think I need it when I run `mvn install -pl :infinispan-server-hotrod -am` to get it built even w/o having the plugin in local repo.

danberindei: The dependency in the plugins section should be enough, in fact this worked for me the first time after adding the phase: `mvn clean && mvn install -am -pl integrationtests/as-lucene-directory`

danberindei:
Build worked, I'm not going to wait for the tests now :)
Thanks for your perseverence, Radim!

PR #5859, opened 2018-03-20
ISPN-8946 & ISPN-8970 Cache (entrySet/keySet/values) cleanup

wburns:
https://issues.jboss.org/browse/ISPN-8946
https://issues.jboss.org/browse/ISPN-8970

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/CloseableSpliteratorMapper.java
+++ b/commons/src/main/java/org/infinispan/commons/util/CloseableSpliteratorMapper.java
@@ -4,7 +4,7 @@ import java.util.Spliterator;
 import java.util.function.Function;
 
 /**
- * A spliterator that maps each value to the output of the Function that is also closeable. If the underlying
+ * A spliterator that maps each value to the output of the Function and is also closeable. If the underlying
  * spliterator is closeable, it will also close it
  * @author wburns
  * @since 9.0
--- a/commons/src/main/java/org/infinispan/commons/util/CloseableSpliteratorMapper.java
+++ b/commons/src/main/java/org/infinispan/commons/util/CloseableSpliteratorMapper.java
@@ -1,22 +1,32 @@
 package org.infinispan.commons.util;
 
+import java.util.Spliterator;
 import java.util.function.Function;
 
 /**
- * A spliterator that maps each value to the output of the Function that is also closeable.
+ * A spliterator that maps each value to the output of the Function that is also closeable. If the underlying
+ * spliterator is closeable, it will also close it
  * @author wburns
  * @since 9.0
  */
 public class CloseableSpliteratorMapper<E, S> extends SpliteratorMapper<E, S> implements CloseableSpliterator<S> {
-   private final CloseableSpliterator<? extends E> spliterator;
-
-   public CloseableSpliteratorMapper(CloseableSpliterator<E> spliterator, Function<? super E, ? extends S> mapper) {
+   public CloseableSpliteratorMapper(Spliterator<E> spliterator, Function<? super E, ? extends S> mapper) {
       super(spliterator, mapper);
-      this.spliterator = spliterator;
+   }
+
+   @Override
+   public Spliterator<S> trySplit() {
+      Spliterator<E> split = spliterator.trySplit();
+      if (split != null) {
+         return new CloseableSpliteratorMapper<>(split, mapper);
+      }
+      return null;
    }
 
    @Override
    public void close() {
-      spliterator.close();
+      if (spliterator instanceof CloseableSpliterator) {
+         ((CloseableSpliterator) spliterator).close();
+      }
    }
 }
--- a/core/src/main/java/org/infinispan/cache/impl/EncoderCache.java
+++ b/core/src/main/java/org/infinispan/cache/impl/EncoderCache.java
@@ -1,8 +1,7 @@
 package org.infinispan.cache.impl;
 
 import java.lang.annotation.Annotation;
-import java.util.AbstractCollection;
-import java.util.AbstractSet;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -22,16 +21,12 @@ import org.infinispan.AdvancedCache;
 import org.infinispan.Cache;
 import org.infinispan.CacheCollection;
 import org.infinispan.CacheSet;
-import org.infinispan.CacheStream;
 import org.infinispan.commons.dataconversion.Encoder;
 import org.infinispan.commons.dataconversion.IdentityEncoder;
 import org.infinispan.commons.dataconversion.IdentityWrapper;
 import org.infinispan.commons.dataconversion.MediaType;
 import org.infinispan.commons.dataconversion.Wrapper;
-import org.infinispan.commons.util.CloseableIterator;
-import org.infinispan.commons.util.CloseableIteratorMapper;
-import org.infinispan.commons.util.CloseableSpliterator;
-import org.infinispan.commons.util.CloseableSpliteratorMapper;
+import org.infinispan.commons.util.InjectiveFunction;
 import org.infinispan.compat.BiFunctionMapper;
 import org.infinispan.compat.FunctionMapper;
 import org.infinispan.container.InternalEntryFactory;
@@ -46,7 +41,8 @@ import org.infinispan.metadata.Metadata;
 import org.infinispan.notifications.cachelistener.ListenerHolder;
 import org.infinispan.notifications.cachelistener.filter.CacheEventConverter;
 import org.infinispan.notifications.cachelistener.filter.CacheEventFilter;
-import org.infinispan.util.AbstractDelegatingCacheStream;
+import org.infinispan.util.WriteableCacheCollectionMapper;
+import org.infinispan.util.WriteableCacheSetMapper;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
@@ -170,88 +166,6 @@ public class EncoderCache<K, V> extends AbstractDelegatingAdvancedCache<K, V> {
          entryMap.put(unwrappedKey, entryToPut);
       });
       return entryMap;
-   }
-
-   private class EncodedCacheStream<R> extends AbstractDelegatingCacheStream<R> {
-
-      public EncodedCacheStream(CacheStream<R> stream) {
-         super(stream);
-      }
-
-      @Override
-      public AbstractDelegatingCacheStream<R> filterKeys(Set<?> keys) {
-         return super.filterKeys(encodeKeysForWrite(keys));
-      }
-   }
-
-   /**
-    * This only extends AbstractSet so it falls back to iterator in case if we didn't implement a method
-    */
-   private class EncodedKeySet extends AbstractSet<K> implements CacheSet<K> {
-      private final CacheSet<K> actualCollection;
-      private final EncoderKeyMapper keyMapper = new EncoderKeyMapper(keyDataConversion);
-
-      EncodedKeySet(CacheSet<K> actualCollection) {
-         this.actualCollection = actualCollection;
-      }
-
-      @Override
-      public CacheStream<K> stream() {
-         return new EncodedCacheStream<>(actualCollection.stream().map(keyMapper));
-      }
-
-      @Override
-      public CacheStream<K> parallelStream() {
-         return new EncodedCacheStream<>(actualCollection.parallelStream().map(keyMapper));
-      }
-
-      @Override
-      public CloseableIterator<K> iterator() {
-         return new CloseableIteratorMapper<>(actualCollection.iterator(), EncoderCache.this::keyFromStorage);
-      }
-
-      @Override
-      public int size() {
-         return actualCollection.size();
-      }
-
-      @Override
-      public CloseableSpliterator<K> spliterator() {
-         return new CloseableSpliteratorMapper<>(actualCollection.spliterator(), EncoderCache.this::keyFromStorage);
-      }
-
-      @Override
-      public boolean contains(Object o) {
-         return actualCollection.contains(keyToStorage(o));
-      }
-
-      @Override
-      public boolean remove(Object o) {
-         return actualCollection.remove(keyToStorage(o));
-      }
-
-      // These methods should be delegated to actual collection as they can be faster
-
-      @Override
-      public boolean removeAll(Collection<?> c) {
-         return actualCollection.removeAll(buildKeyCollection(c));
-      }
-
-      @Override
-      public boolean retainAll(Collection<?> c) {
-         return actualCollection.retainAll(buildKeyCollection(c));
-      }
-
-      Collection<K> buildKeyCollection(Collection<?> c) {
-         Collection<K> toRemove = new ArrayList<>(c.size());
-         for (Object obj : c) {
-            K entry = keyToStorage(obj);
-            if (entry != null) {
-               toRemove.add(entry);
-            }
-         }
-         return toRemove;
-      }
    }
 
    @Override
@@ -556,7 +470,10 @@ public class EncoderCache<K, V> extends AbstractDelegatingAdvancedCache<K, V> {
 
    @Override
    public CacheSet<CacheEntry<K, V>> cacheEntrySet() {
-      return new EncoderEntrySet(super.cacheEntrySet());
+      EncoderEntryMapper<K, V, CacheEntry<K, V>> cacheEntryMapper = EncoderEntryMapper.newCacheEntryMapper(
+            keyDataConversion, valueDataConversion);
+      return new WriteableCacheSetMapper<>(super.cacheEntrySet(), cacheEntryMapper,
+            e -> new CacheEntryWrapper<>(cacheEntryMapper.apply(e), e), this::toCacheEntry, this::keyToStorage);
    }
 
    @Override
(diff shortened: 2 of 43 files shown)


wburns:
First commit:

1. EncoderCache changes now so that all operations go through a mapper only and delegate EVERYTHING to the underlying cache. All multi object operations also make mapped collections with no copies (ie. removeAll, retainAll, removeIf). Also encoder cache uses InjectiveFunctions for its various map operations (which will be useful if we ever implement streams using Spliterator characteristics enhancements)
2. CacheLoaderInterceptor / PrefetchInterceptor / TxInterceptor - writes will always delegate to the cache immediately (ie. clear / remove / iterator.remove) - reads will almost always delegate to lower interceptor adding functionality on top (ie. add store entries, scattered cache logic or tx entries)
3. (EntrySet|KeySet)Commands delegate all writes to the cache directly and reads are always to the data container
4. Various Closeable classes now take a non Closeable instance (ie. Iterator) but will close it if it is Closeable when the Closeable class is closed. This allows for less wrapping. https://github.com/infinispan/infinispan/pull/5859/files#diff-e40453e0ade0934f49c80ea182d50d1fR22
5. Fixed bug in Simple cache that iterator remove wasn't working in value collection https://github.com/infinispan/infinispan/pull/5859/files#diff-16d7fd1202fded0dea1d5c8a80fbd5e1R1738
6. Fixed bug in cache loader that computeIfAbsent set a null value if compute returned null. https://github.com/infinispan/infinispan/pull/5859/files#diff-abcc696dab7dcd8af469785e238eb605R233
7. Fixed bug in spliterator mapper that splits were not mapped. https://github.com/infinispan/infinispan/pull/5859/files#diff-202639316b539ccced62713a5b1d2204R18
8. Iterator methods will only wrap the iterator if it is the first interceptor to do so using REMOTE_ITERATION flag https://github.com/infinispan/infinispan/pull/5859/files#diff-3a4b4e02f4977873b3ca49ac218c7c85R69 Technically this flag has been repurposed for sign of showing don't wrap iterator if present.
9. SimpleCache now use the data container spliterator directly, allowing for better parallelism than using iterator method. https://github.com/infinispan/infinispan/pull/5859/files#diff-16d7fd1202fded0dea1d5c8a80fbd5e1R1768

Unfortunately many interceptors still use iterator instead of spliterator for stream (CacheLoaderInterceptor and TxInterceptor for example) which reduce parallelism a bit. This might be fixable for 9.3, but way too much work for 9.2.

Second commit:

Stream iterator method will no longer support remove operation. There are no implementations in standard JDK that provide this functionality. Ours semi worked, but not really. It is a remnant of the way Cache.entrySet.iterator worked. This removes a few layers of wrapping for a stream, which are really unneeded.

core/src/main/java/org/infinispan/cache/impl/EncoderCache.java
```diff
@@ -789,170 +706,96 @@ public void forEach(BiConsumer<? super K, ? super V> action) {
...
+               CacheEntry returned = (CacheEntry) o;
+               return convertEntry(newKey, newValue, returned);
+            } else {
+               return entryFactory.create(newKey, newValue, (Metadata) null);
+            }
+         }
+         if (entry instanceof CacheEntry) {
+            return (CacheEntry<K, V>) entry;
+         } else {
+            return entryFactory.create(entry.getKey(), entry.getValue(), (Metadata) null);
```
karesti: we can use key and value variables here, right ?

core/src/main/java/org/infinispan/commands/read/AbstractCloseableIteratorCollection.java
```diff
@@ -16,6 +16,9 @@
  * methods such as {@link Collection#contains(Object)}, {@link Collection#remove(Object)} and
  * {@link org.infinispan.commons.util.CloseableIteratorCollection#iterator()}.  The {@link Collection#add(Object)} by default will throw an
  * {@link java.lang.UnsupportedOperationException}.
+ * <p>
+ * @implNote Any implementor should always use the underlying {@link Cache} instance whe performing writes on this
```
karesti: when

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
```diff
@@ -498,76 +584,114 @@ long calculateTimeoutSeconds() {
...
+         return new KeyStreamSupplier<>(cache, getSegmentMapper(cache), () -> StreamSupport.stream(spliterator(), false));
       }
 
       @Override
-      public int size() {
-         long size = stream().count();
-         if (size > Integer.MAX_VALUE) {
-            return Integer.MAX_VALUE;
+      public boolean contains(Object o) {
+         boolean contains = super.contains(o);
```
karesti: if o is null, should no we just return false ?

core/src/main/java/org/infinispan/util/WriteableCacheSetMapper.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Collection;
+
+import org.infinispan.CacheSet;
+import org.infinispan.commons.util.InjectiveFunction;
+
+/**
+ * A writeable cache set mapper that also has constant time operations for things such as
+ * {@link Collection#contains(Object)} if the underlying Set does. Also implements the Set interface.
+ * <p>
+ * This set should be used for cases when a simple transformation of a element to another is all that is
```
karesti: an element ?

core/src/test/java/org/infinispan/persistence/CacheAPIPeristenceTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.infinispan.persistence.dummy.DummyInMemoryStoreConfigurationBuilder;
+import org.mockito.Mockito;
+
+/**
+ * Test that ensure that when persistence is used with an always empty data container that various operations
+ * are properly supported
+ * @author wburns
+ * @since 9.2
+ */
+public class CacheAPIPeristenceTest extends APINonTxTest {
```
karesti: CacheAPIPersistenceTest

core/src/test/java/org/infinispan/persistence/CacheAPIPeristenceTest.java
```diff
@@ -0,0 +1,51 @@
...
+   @Override
+   protected void configure(ConfigurationBuilder builder) {
+      // We use a mocked container, so nothing is written
+      DataContainer mockContainer = Mockito.mock(DataContainer.class);
```
karesti: I personally like adding static imports to avoir Mockito. noise

core/src/main/java/org/infinispan/commands/read/AbstractCloseableIteratorCollection.java
wburns: 👍

core/src/test/java/org/infinispan/persistence/CacheAPIPeristenceTest.java
wburns: 👍

core/src/test/java/org/infinispan/persistence/CacheAPIPeristenceTest.java
wburns: Yeah I teeter back and forth. I will do that though.

wburns:
Also there are 2 test failures related to compatibility mode that I am talking to @gustavonalle about.

wburns:
Pushed a commit to test something, most likely it will cause a bunch of CI failures, but maybe we can get lucky.

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
wburns: Sure, in the Collection API it is valid, but we don't allow null keys.

wburns:
Address comments and tests should be fixed. Waiting for CI to confirm.

wburns:
Test failures are unrelated.

gustavonalle:
LGTM, @danberindei might want to look as well since this PR originated form one of his comments?

core/src/main/java/org/infinispan/context/impl/FlagBitSets.java
```diff
@@ -65,4 +65,18 @@ public static Flag extractStateTransferFlag(InvocationContext ctx, FlagAffectedC
...
+    * Retrieves the values for the given flags and then sets those flags on the command. This can be useful
+    * if subsequent users of this command know that an upstream operation has already handled a given flag
+    * @param command the command to get the flags from and then set
+    * @param flagsToSet the flags to get and set
+    * @return the flags that were set
+    */
+   public static long getAndSetFlags(FlagAffectedCommand command, long flagsToSet) {
+      long originalFlags = command.getFlagsBitSet();
+      long setFlags = originalFlags & flagsToSet;
+      command.setFlagsBitSet(originalFlags | flagsToSet);
```
danberindei: I would expect a method named `getAndSet` to set all the flags and to return all the previous flags, without any change. A more appropriate name would be `getFlagsIntersectionAndSetFlagsUnion` ;)

I'd also prefer having it as a default method in `FlagAffectedCommand`, and using `EnumUtil` methods instead of bit operations.

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
```diff
@@ -483,12 +498,83 @@ public void disableStore(@Parameter(name = "storeType", description = "Fully qua
       persistenceManager.disableStore(storeType);
    }
 
-   private class WrappedEntrySet extends AbstractDelegatingEntryCacheSet<K, V> {
-      private final CacheSet<CacheEntry<K, V>> entrySet;
+   private ToIntFunction<Object> getSegmentMapper(Cache<?, ?> cache) {
+      DistributionManager dm = cache.getAdvancedCache().getDistributionManager();
+      if (dm != null) {
+         return dm.getCacheTopology()::getSegment;
```
danberindei: `CacheLoaderInterceptor` can inject the `KeyPartitioner` directly.

BTW, the `KeyPartitioner` is customizable, but we don't say in the javadoc whether what key format it's supposed to support. That might be because we didn't have any guarantee that it would only be called on the storage-formatted keys at the time we added it :)

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
```diff
@@ -498,76 +584,120 @@ long calculateTimeoutSeconds() {
          return Math.round(Math.log1p(size) * 10);
       }
 
+      private Map.Entry<K, V> toEntry(Object obj) {
+         if (obj instanceof Map.Entry) {
+            return (Map.Entry) obj;
+         } else {
+            return null;
```
danberindei: The cast would work with`null` as well, what kind of objects are we filtering out here?

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
```diff
@@ -301,19 +332,53 @@ public int size() {
...
+      return super.visitKeySetCommand(ctx, command);
    }
 
    @Override
    public Object visitEntrySetCommand(InvocationContext ctx, EntrySetCommand command) throws Throwable {
       enlistIfNeeded(ctx);
-      return invokeNextThenApply(ctx, command, (rCtx, rCommand, rv) -> {
-         if (rCtx.isInTxScope()) {
+      if (ctx.isInTxScope()) {
+         // We just set it, unfortunately we always wrap even if this flag was set
```
danberindei: I don't expect anyone to benchmark calls to `cache.entrySet()` without actually iterating through the set, so I don't think an extra wrapper instance matters.

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
```diff
@@ -341,10 +406,18 @@ public int size() {
                   }
                   return (int) size;
                }
+
+               private Map.Entry<K, V> toEntry(Object obj) {
+                  if (obj instanceof Map.Entry) {
+                     return (Map.Entry) obj;
+                  } else {
+                     return null;
```
danberindei: Again, do we expect non-null non-entry objects here?

core/src/main/java/org/infinispan/util/CacheStreamMapper.java
```diff
@@ -0,0 +1,25 @@
...
+import org.infinispan.CacheStream;
+import org.infinispan.commons.util.InjectiveFunction;
+
+/**
+ * CacheStream that allows for mapping the filtered keys to different keys. Note that the function provided
+ * must be a {@link InjectiveFunction} guaranteeing that keys are distinct when mapped.
+ * @author wburns
+ * @since 9.2
+ */
+class CacheStreamMapper<R> extends AbstractDelegatingCacheStream<R> {
```
danberindei: I'd rename it to `MapperCacheStream`. And maybe check that `keyMapper instanceof InjectiveFunction`?

core/src/test/java/org/infinispan/persistence/CacheAPIPersistenceTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.infinispan.container.InternalEntryFactoryImpl;
+import org.infinispan.persistence.dummy.DummyInMemoryStoreConfigurationBuilder;
+
+/**
+ * Test that ensure that when persistence is used with an always empty data container that various operations
+ * are properly supported
+ * @author wburns
+ * @since 9.2
+ */
+public class CacheAPIPersistenceTest extends APINonTxTest {
```
danberindei: I'd rename it to `APINonTxPersistenceTest` to match the base class. Or maybe rename the base class?

core/src/test/java/org/infinispan/stream/BaseStreamIteratorTest.java
```diff
@@ -273,14 +278,17 @@ public Object visitRemoveCommand(InvocationContext ctx, RemoveCommand command) t
...
-              Flag.SKIP_CACHE_STORE).entrySet().stream().iterator(); it.hasNext();) {
-         Map.Entry<Object, Object> entry = it.next();
-         Object key = entry.getKey();
-         assertEquals(values.get(key), entry.getValue());
-         it.remove();
-      }
-
-      assertEquals(0, cache.size());
+      Iterator<Map.Entry<Object, Object>> it = cache(0, CACHE_NAME).getAdvancedCache().withFlags(
+              Flag.SKIP_CACHE_STORE)
```
danberindei: You can use `advancedCache(0, CACHE_NAME)` and fit it on one line

commons/src/main/java/org/infinispan/commons/util/CloseableSpliteratorMapper.java
```diff
@@ -1,22 +1,32 @@
 package org.infinispan.commons.util;
 
+import java.util.Spliterator;
 import java.util.function.Function;
 
 /**
- * A spliterator that maps each value to the output of the Function that is also closeable.
+ * A spliterator that maps each value to the output of the Function that is also closeable. If the underlying
```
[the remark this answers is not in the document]
wburns: Sure

core/src/main/java/org/infinispan/util/WriteableCacheSetMapper.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A writeable cache set mapper that also has constant time operations for things such as
+ * {@link Collection#contains(Object)} if the underlying Set does. Also implements the Set interface.
+ * <p>
+ * This set should be used for cases when a simple transformation of an element to another is all that is
+ * needed by the underlying set.
+ * <p>
+ * This implementation is basically identical to {@link WriteableCacheCollectionMapper} except that this class
+ * also implements {@link java.util.Set}.
```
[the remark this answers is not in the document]
wburns: Okay

core/src/main/java/org/infinispan/util/WriteableCacheCollectionMapper.java
```diff
@@ -0,0 +1,96 @@
...
+      this.fromNewTypeFunction = fromNewTypeFunction;
+   }
+
+   @Override
+   public CloseableIterator<R> iterator() {
+      return new CloseableIteratorMapper<>(realCollection.iterator(), toNewTypeIteratorFunction);
+   }
+
+   @Override
+   public boolean contains(Object o) {
```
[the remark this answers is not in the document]
wburns: Unfortunately it can't implement it as it would need the function to convert from target type to storage type (ie. R to E). I can document this on `CacheCollectionMapper` as well as it currently is only on `CollectionMapper`

core/src/main/java/org/infinispan/cache/impl/EncoderCache.java
```diff
@@ -789,170 +706,96 @@ public void forEach(BiConsumer<? super K, ? super V> action) {
...
-                  V value = entry.getValue();
-                  V valueFromStorage = valueFromStorage(value);
-                  if (keyFromStorage != key || valueFromStorage != value) {
-                     return convertEntry(keyFromStorage, valueFromStorage, entry);
-                  }
-                  return entry;
-               });
-      }
+   private class EntryWrapper<A, B> implements Entry<A, B> {
+      private final Entry<A, B> previousEntry;
```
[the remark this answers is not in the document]
wburns: Okay, that was just how it was before :) But I like the new name better.

core/src/main/java/org/infinispan/util/WriteableCacheCollectionMapper.java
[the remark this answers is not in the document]
danberindei: Then maybe remove `CacheCollectionMapper`? I don't see it used anywhere else.

core/src/main/java/org/infinispan/cache/impl/EncoderEntryMapper.java
```diff
@@ -4,29 +4,43 @@
...
+      InjectiveFunction<T, T> {
    @Inject private transient InternalEntryFactory entryFactory;
 
    private final DataConversion keyDataConversion;
    private final DataConversion valueDataConversion;
 
-   public EncoderEntryMapper(DataConversion keyDataConversion, DataConversion valueDataConversion) {
+   public static <K, V> EncoderEntryMapper<K, V, Map.Entry<K, V>> newEntryMapper(DataConversion keyDataConversion,
+         DataConversion valueDataConversion) {
+      return new EncoderEntryMapper<>(keyDataConversion, valueDataConversion);
```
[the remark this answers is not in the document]
wburns: Good point. I can move that into the static method factory.

core/src/main/java/org/infinispan/cache/impl/EncoderEntryMapper.java
```diff
@@ -4,29 +4,43 @@
...
+import org.infinispan.metadata.Metadata;
 import org.infinispan.util.function.RemovableFunction;
 
 /**
  * {@link java.util.function.Function} that uses an encoder to converter entries from the configured storage format to
  * the requested format.
  */
-public class EncoderEntryMapper<K, V> implements RemovableFunction<CacheEntry<K, V>, CacheEntry<K, V>> {
+public class EncoderEntryMapper<K, V, T extends Map.Entry<K, V>> implements RemovableFunction<T, T>,
+      InjectiveFunction<T, T> {
```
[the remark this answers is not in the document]
wburns: Actually after these changes there shouldn't be any need for it. I will deprecate it.

core/src/main/java/org/infinispan/cache/impl/EncoderCache.java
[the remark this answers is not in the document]
danberindei: Well, if it's not always an entry in storage format, then definitely don't rename it to `storageEntry`... I thought it's always the storage-format entry, although it's not very clear to me why updating the storage-format entry is necessary.

core/src/main/java/org/infinispan/cache/impl/EncoderEntryMapper.java
```diff
@@ -44,9 +58,13 @@ private Object decode(Object o, Encoder encoder) {
...
-   public CacheEntry<K, V> apply(CacheEntry<K, V> e) {
-      boolean keyFilterable = keyDataConversion.isStorageFormatFilterable();
-      boolean valueFilterable = valueDataConversion.isStorageFormatFilterable();
+   public T apply(T e) {
+      return apply(e, false);
+   }
+
+   public T apply(T e, boolean ignoreStorageFormatFilterable) {
+      boolean keyFilterable = !ignoreStorageFormatFilterable && keyDataConversion.isStorageFormatFilterable();
+      boolean valueFilterable = !ignoreStorageFormatFilterable && valueDataConversion.isStorageFormatFilterable();
```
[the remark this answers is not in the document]
wburns: Yeah I really hate these variables. I was talking with @gustavonalle for a few hours last week about this. To be honest, this is something that needs to be cleaned up after we remove compatibility mode. I tried to clean it up in my PR here a bit more, but failed (so I just left it as is - with minimal interaction).

core/src/main/java/org/infinispan/cache/impl/SimpleCacheImpl.java
```diff
@@ -1611,8 +1612,8 @@ public void clear() {
 
       @Override
       public CloseableSpliterator<Entry<K, V>> spliterator() {
-         return Closeables.spliterator(iterator(), dataContainer.sizeIncludingExpired(),
-               Spliterator.CONCURRENT | Spliterator.NONNULL | Spliterator.DISTINCT);
+         // Cast to raw since we need to convert from ICE to Entry
+         return Closeables.spliterator((Spliterator) dataContainer.spliterator());
```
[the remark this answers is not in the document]
wburns: Unfortunately that won't avoid the cast either - since the compiler doesn't know the type (could be Spliterator<Object> for example).

core/src/main/java/org/infinispan/cache/impl/EncoderEntryMapper.java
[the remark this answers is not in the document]
danberindei: Yeah, sometimes it's easier to clean up than it is to explain how it's working now ;)

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
wburns: Thanks, forgot to change this one from before.

core/src/main/java/org/infinispan/util/WriteableCacheCollectionMapper.java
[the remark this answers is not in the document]
wburns: Technically I can't remove it as it was there before. I also have to add back in the previous constructor without the `keyFilterMapper`.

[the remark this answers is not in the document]
wburns: I will deprecate it though.

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
wburns: 👍

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
wburns: 👍

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
wburns: Actually it has to be like this. The problem is the user can pass anything since method declaration is `Object`

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
wburns: It does actually. And actually looking at this closer, this is highlighted because there is a bug in `TxInterceptor` that it isn't properly wrapping the iterator to support `remove` operation. Normally this would cause an extra wrapper per iterator instance.

wburns: I have another test to add it looks like.

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
wburns: See above comment, but it has to do this :(

core/src/main/java/org/infinispan/util/CacheStreamMapper.java
wburns: Sure about name. Not sure what you mean about the second point, the `keyMapper` has to be an instanceof `InjectiveFunction` as it is declared in the constructor.

core/src/test/java/org/infinispan/persistence/CacheAPIPersistenceTest.java
wburns: Okay.

core/src/test/java/org/infinispan/stream/BaseStreamIteratorTest.java
wburns: Ah wow I never noticed that method 👍

core/src/main/java/org/infinispan/util/CacheStreamMapper.java
wburns: I must admit `MapperCacheStream` sounds a bit odd to me. Maybe `MappedCacheStream` ?

core/src/main/java/org/infinispan/util/WriteableCacheCollectionMapper.java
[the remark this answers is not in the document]
danberindei: There's no `@public` in the `org.infinispan.util.package-info`...

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
danberindei: I just meant there's no need for "unfortunately" in the comment, we need to wrap it for correctness so we do it.

core/src/main/java/org/infinispan/cache/impl/EncoderCache.java
```diff
@@ -471,7 +471,7 @@ public void putAll(Map<? extends K, ? extends V> map, Metadata metadata) {
    @Override
    public CacheSet<CacheEntry<K, V>> cacheEntrySet() {
       EncoderEntryMapper<K, V, CacheEntry<K, V>> cacheEntryMapper = EncoderEntryMapper.newCacheEntryMapper(
-            keyDataConversion, valueDataConversion);
+            keyDataConversion, valueDataConversion, componentRegistry);
```
danberindei: I assumed you were going to pass `entryFactory` here

wburns: I can do that instead, sure. I saw the other Inject method and forgot those were already handled by EncoderCache.

danberindei:
Squashed and integrated, thanks Will!

PR #5860, opened 2018-03-21
ISPN-8974 Nullify marshaller/unmarshaller is instance cache too big

galderz:


commons/src/main/java/org/infinispan/commons/marshall/jboss/ExtendedRiverUnmarshaller.java
```diff
@@ -46,4 +48,11 @@ public void finish() throws IOException {
       }
    }
 
+   int instanceCacheResizeCount() {
+      List instanceCache = (List) ReflectionUtil
+         .getValue(this, "instanceCache");
+
+      return instanceCache.size();
```
danberindei: I think `size()` will always return 0 at this point, because the instance cache is cleared between invocations. OTOH `ArrayList` has a `trimToSize()` method, maybe we could use that and keep using the same `RiverUnmarshaller` instance.

commons/src/main/java/org/infinispan/commons/marshall/jboss/ExtendedRiverMarshaller.java
```diff
@@ -35,4 +36,12 @@ void setCloseListener(RiverCloseListener listener) {
       this.listener = listener;
    }
 
+   int instanceCacheResizeCount() {
+      Object instanceCache = ReflectionUtil
+         .getValue(this, "instanceCache");
+
+      return (int) ReflectionUtil
+         .getValue(instanceCache, "resizeCount");
```
danberindei: I know `resizeCount` is the actual capacity, but I'd rather use `keys.length` (and call the method `instanceCacheCapacity`) instead because we care about memory usage here.

galderz:
@danberindei Addressed both comments

danberindei:
Wow, lots of test failures on 8.2.x! At least the core non-xsite tests look good on my machine, so I'm integrating it :)

danberindei:
Interesting enough, I only got one test failure running the xsite tests on my machine, but pretty much all of them are failing in CI.

PR #5863, opened 2018-03-22
ISPN-8978 Remove usage of deprecated GMS#install_view_locally_first

rhusar:
Jira
https://issues.jboss.org/browse/ISPN-8978

GMS#install_view_locally_first was deprecated and is never used but does not issue a warning message (JGRP-2258).

This should be removed from the stack configuration.

Cross-check
https://github.com/belaban/JGroups/blob/master/src/org/jgroups/protocols/pbcast/GMS.java#L107


tristantarrant:
Merged, thanks

PR #5866, opened 2018-03-23
ISPN-8962 PreferAvailabilityStrategy: Rely less on the stable topology

danberindei:
https://issues.jboss.org/browse/ISPN-8962
https://issues.jboss.org/browse/ISPN-8993


danberindei:
Updated to recognize more situations when multiple topologies represent the same partition and rebased on top of Ryan's ISPN-8903 changes.

danberindei:
Added a separate commit for https://issues.jboss.org/browse/ISPN-8993

tristantarrant:
I assume https://ci.infinispan.org/job/Infinispan/job/PR-5866/5/testReport/junit/org.infinispan.server.hotrod/HotRodMergeTest/testNewTopologySentAfterOverlappingMerge_DIST_SYNC_/ is unrelated ?

danberindei:
I forgot about that test, it's definitely related.

danberindei:
@tristantarrant `HotRodMergeTest` was failing because I had changed the way the preferred topology is selected, and the new pick required an extra rebalance. I pushed some more changes to avoid the extra rebalance if possible. I also changed `PreferConsistencyStrategy` to make it use the read CH for conflict resolution (for rebalance phases `READ_NEW` or `READ_ALL`).

I was also getting a different error when running locally and I tracked it to the other caches defined by the HR server that also have conflict resolution enabled (___protobuf_metadata, org.infinispan.CONFIG, __global_tx_table__, org.infinispan.COUNTER), because 2 test methods use the same servers and the 2nd test wasn't waiting for all those caches to finish conflict resolution. I've created https://issues.jboss.org/browse/ISPN-9009 to test this scenario properly.

tristantarrant:
is https://ci.infinispan.org/job/Infinispan/job/PR-5866/8/testReport/junit/org.infinispan.functional/FunctionalDistributionTest/testDistributionFromPrimaryOwner/ related ?

danberindei:
@tristantarrant almost certainly not related, https://issues.jboss.org/browse/ISPN-7098

core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
```diff
@@ -502,7 +502,7 @@ private void recoverClusterStatus(int newViewId, final List<Address> clusterMemb
       CountDownLatch latch = new CountDownLatch(responsesByCache.size());
       LimitedExecutor cs = new LimitedExecutor("Merge-" + newViewId, stateTransferExecutor, maxThreads);
       for (final Map.Entry<String, Map<Address, CacheStatusResponse>> e : responsesByCache.entrySet()) {
-         CacheJoinInfo joinInfo = e.getValue().values().stream().findAny().get().getCacheJoinInfo();
+         CacheJoinInfo joinInfo = e.getValue().values().iterator().next().getCacheJoinInfo();
```
tristantarrant: I wish ordered collections had a first() method that encapsulates iterator().next()

danberindei: I don't think it's that bad, `first()` would probably return an `Optional` like the one in `Stream` so you'd need an extra method call anyway. The only advantage would be replacing the `NoSuchElementException` with our own, but then that would be more verbose, and we assume at least one response anyway.

tristantarrant:
Looks good to me. I actually understood it :smile:

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -792,11 +793,17 @@ public synchronized void startQueuedRebalance() {
...
       boolean rebalance = false;
+      boolean updateStableTopology = false;
       if (rebalanceType == RebalanceType.NONE) {
          updateTopology = true;
       } else if (balancedCH.equals(currentCH)) {
          log.tracef("The balanced CH is the same as the current CH, not rebalancing");
          updateTopology = cacheTopology.getPendingCH() != null;
+         // After a cluster view change that leaves only 1 node, we don't need either a topology update or a rebalance
+         // but we must still update the stable topology
+         updateStableTopology =
```
tristantarrant: Wondering if it makes sense to have methods on CacheTopology that encapsulate some of the logic (I haven't checked about repeated uses).
e.g cacheTopology.needsStableTopologyUpdate(stableTopology)

danberindei: I'd like to avoid adding methods to `CacheTopology` that the application shouldn't care about, because it's exposed in `DistributionManager` and that makes changes harder.

For 9.3 I'd actually like to move internals like `ClusterCacheStatus` and `ClusterTopologyManagerImpl` away from using `CacheTopology`, so I can deprecate currentCH/pendingCH/unionCH/persistentUUIDs. The internal-only replacement would be a good location for such methods, though.

core/src/test/java/org/infinispan/conflict/impl/BaseMergePolicyTest.java
```diff
@@ -144,45 +143,24 @@ public void testPartitionMergePolicy() {
...
-         if (maxTopology == null || maxTopology.getMembers().size() < topology.getMembers().size()) {
-            maxTopology = topology;
-            maxTopologySender = response.getKey();
-         }
-      }
-      if (trace) log.tracef("getCacheFromPreferredPartition: partition=%s", maxTopology != null ? maxTopology.getCurrentCH().getMembers() : null);
-      return maxTopologySender;
+      Map<Address, CacheStatusResponse> statusResponses =
+         Arrays.stream(caches).map(c -> new KeyValuePair<>(address(c), getCacheStatus(c)))
+               .collect(Collectors.toMap(KeyValuePair::getKey, KeyValuePair::getValue));
```
wburns: Can be just: `Arrays.stream(caches).collect(Collectors.toMap(this::address, this::getCacheStatus));`

core/src/test/java/org/infinispan/partitionhandling/impl/PreferAvailabilityStrategyTest.java
```diff
@@ -53,162 +73,199 @@
...
-      return new Object[][] {{true},{false}};
+   @Factory(dataProvider = "conflictResolutionProvider")
+   public PreferAvailabilityStrategyTest(ConflictResolution conflicts) {
+      this.conflicts = conflicts;
    }
 
    @BeforeMethod(alwaysRun = true)
    public void setup() {
+      mockitoSession = Mockito.mockitoSession()
+                              .strictness(Strictness.STRICT_STUBS)
```
wburns: The default is `STRICT_STUBS` in 2.7. Can you clarify what this is doing for us?

core/src/test/java/org/infinispan/topology/ClusterCacheStatusTest.java
```diff
@@ -0,0 +1,147 @@
...
+   private static final Address C = new TestAddress(3, "C");
+
+   private ClusterCacheStatus status;
+   private ClusterTopologyManager topologyManager;
+   private MockitoSession mockitoSession;
+   private Transport transport;
+
+   @BeforeMethod(alwaysRun = true)
+   public void setup() {
+      mockitoSession = Mockito.mockitoSession().initMocks(this).strictness(Strictness.STRICT_STUBS).startMocking();
```
wburns: Same thing with this one. Also I don't see any fields annotated with `@Mock` for `initMocks`?

core/src/test/java/org/infinispan/topology/TestClusterCacheStatus.java
```diff
@@ -0,0 +1,245 @@
...
+                                   topology.getActualMembers(), persistentUUIDs(topology.getActualMembers()));
+   }
+
+   /**
+    * The topologies included in the status responses do not have a union CH, so
+    * {@link CacheTopology#getReadConsistentHash()} doesn't work.
+    */
+   public ConsistentHash readConsistentHash() {
+      switch (topology.getPhase()) {
+         case NO_REBALANCE:
```
wburns: replace this with `AvailabilityStrategy.readConsistentHash`?

core/src/main/java/org/infinispan/partitionhandling/impl/PreferAvailabilityStrategy.java
```diff
@@ -93,166 +68,283 @@ public void onClusterViewChange(AvailabilityStrategyContext context, List<Addres
...
+                  log.tracef("Cache %s ignoring compatible old topology from %s: %s", cacheName, p.senders, p.topology);
+                  fold = true;
+               } else {
+                  if (trace)
+                     log.tracef("Cache %s partition of %s overlaps with partition of %s but possibly holds extra entries",
+                                cacheName, p.senders, referencePartition.senders);
+               }
+            }
+            if (fold) {
+               referencePartition.senders.addAll(p.senders);
```
wburns: I kinda wonder if this would be simpler if the outer i loop was doing `for (int i = 0; i < partitions.size(); ++i)` and we removed the `referencePartition` instead?

core/src/main/java/org/infinispan/partitionhandling/impl/PreferAvailabilityStrategy.java
```diff
@@ -93,166 +68,283 @@ public void onClusterViewChange(AvailabilityStrategyContext context, List<Addres
...
+         p.stableTopology = response.getStableTopology();
+         // The topologies used here do not have a union CH, so CacheTopology.getReadConsistentHash() doesn't work
+         p.readCH = readConsistentHash(topology, response.getCacheJoinInfo().getConsistentHashFactory());
+         p.actualMembers = topology.getActualMembers();
+         p.senders.add(sender);
+         partitions.add(p);
+      }
+
+      // Sort partitions in reverse topology id order to simplify the algorithm
+      partitions.sort((p1, p2) -> {
```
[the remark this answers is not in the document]
danberindei: Yeah, I had a much complicated comparator but I realized it didn't matter.

core/src/test/java/org/infinispan/conflict/impl/BaseMergePolicyTest.java
danberindei: Here too, I needed a `LinkedHashMap` at some point but it's too much now

core/src/test/java/org/infinispan/partitionhandling/impl/PreferAvailabilityStrategyTest.java
danberindei: I thought it was only going to be the default in 3.0, I'll remove it

core/src/main/java/org/infinispan/partitionhandling/impl/PreferAvailabilityStrategy.java
```diff
@@ -93,166 +68,283 @@ public void onClusterViewChange(AvailabilityStrategyContext context, List<Addres
...
+
       // Then start a rebalance with the merged members
       context.queueRebalance(newMembers);
    }
 
-   private boolean attemptConflictResolution(AvailabilityStrategyContext context, int maxTopologyId,
-                                             CacheTopology maxTopology, Set<Address> possibleOwners) {
-      if (!context.resolveConflictsOnMerge() || possibleOwners.isEmpty() || possibleOwners.size() == 1)
-         return false;
+   private Partition selectPreferredPartition(List<Partition> partitions) {
```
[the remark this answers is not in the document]
danberindei: Done

core/src/test/java/org/infinispan/partitionhandling/impl/PreferAvailabilityStrategyTest.java
danberindei: The explicit `MockitoSession` seems to help by allowing me to skip `verify()` for method calls already expected with `when()`. I don't know why, as I thought that was a feature of STRICT_SUBS, but I'll keep it for now.

core/src/main/java/org/infinispan/partitionhandling/impl/PreferAvailabilityStrategy.java
danberindei: Done

core/src/test/java/org/infinispan/partitionhandling/impl/PreferAvailabilityStrategyTest.java
wburns: Oh because you are using `noMoreInteracations`. Yeah I guess the Mock must use WARN by default in this case, but the Session default is STRICT_SUBS - kinda odd.

wburns:

PR #5867, opened 2018-03-23
ISPN-8917 Make CORS configurable

gustavonalle:
https://issues.jboss.org/browse/ISPN-8917


server/rest/src/main/java/org/infinispan/rest/configuration/RestServerConfiguration.java
```diff
@@ -49,8 +52,16 @@ public int maxContentLength() {
       return maxContentLength;
    }
 
+   /**
+    * @deprecated Use {@link #getCorsRules()} instead.
+    */
+   @Deprecated
```
ryanemerson: Doesn't this mean that we can't include this in 9.2.1.Final?

gustavonalle: This should not be @deprecated, as it was added after 9.2. Force of habit :smile:

tristantarrant:
Merged, thanks

PR #5868, opened 2018-03-26
ISPN-8983 Invalid Ickle query causes server exception

gustavonalle:
https://issues.jboss.org/browse/ISPN-8983

### DIFF
--- a/object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/parser/QueryRendererDelegateImpl.java
+++ b/object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/parser/QueryRendererDelegateImpl.java
@@ -355,10 +355,10 @@ final class QueryRendererDelegateImpl<TypeMetadata> implements QueryRendererDele
       if (property.isEmpty()) {
          throw log.getPredicatesOnEntityAliasNotAllowedException(propertyPath.asStringPath());
       }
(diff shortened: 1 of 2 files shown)


object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/parser/QueryRendererDelegateImpl.java
```diff
@@ -355,10 +355,11 @@ public void predicateFullTextTerm(String term, String fuzzyFlop) {
       if (property.isEmpty()) {
          throw log.getPredicatesOnEntityAliasNotAllowedException(propertyPath.asStringPath());
       }
-      String comparisonValue = (String) parameterValue(term);
+      checkAnalyzed(property,true);
```
anistor: There's ```checkAnalyzed(property,true);``` inside the if statement already. Why do we need it here too?

gustavonalle: Where?

gustavonalle: Ok, saw it

gustavonalle:
@anistor Updated

anistor:
Merged. Thanks @gustavonalle !

PR #5871, opened 2018-03-27
ISPN-8991 fixes testLockUseAfterPartitionWithoutMajority random failure

karesti:
https://issues.jboss.org/browse/ISPN-8991

The modification I did avoid errors, but there is a bug related being corrected by Dan concerning replicated caches

### DIFF
--- a/lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
+++ b/lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
@@ -13,6 +13,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.TimeUnit;
 
+import org.infinispan.configuration.cache.CacheMode;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.lock.api.ClusteredLock;
 import org.infinispan.lock.api.ClusteredLockManager;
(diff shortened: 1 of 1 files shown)


lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
```diff
@@ -30,16 +31,13 @@
...
    }
 
    @Override
    protected void createCacheManagers() throws Throwable {
       ConfigurationBuilder dcc = cacheConfiguration();
-      dcc.clustering().cacheMode(cacheMode).partitionHandling().whenSplit(partitionHandling).mergePolicy(mergePolicy);
-      if (biasAcquisition != null) {
-         dcc.clustering().biasAcquisition(biasAcquisition);
-      }
+      dcc.clustering().cacheMode(cacheMode).partitionHandling().whenSplit(partitionHandling);
```
danberindei: I'd remove the partition handling stuff as well, and hard-code the cache mode. That way you can set `this.cacheMode = null` in the constructor, and `REPL_SYNC` won't appear in the test reports (because the default cache isn't used at all).

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
```diff
@@ -88,11 +86,13 @@ public void testLockUseAfterPartitionWithoutMajority() throws Throwable {
       ClusteredLock lock5 = clm5.get(LOCK_NAME);
 
       splitCluster(new int[]{0, 1, 2}, new int[]{3, 4, 5});
+      partition(0).assertDegradedMode();
+      partition(1).assertDegradedMode();
```
danberindei: Please add a comment here, something like "Wait for degraded topologies to work around ISPN-9008"

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
karesti: when I remove partition handling, test fails

karesti: ok for the rest

lock/src/test/java/org/infinispan/lock/ClusteredLockSplitBrainTest.java
karesti: ok

danberindei:
Integrated, thanks Katia!

PR #5872, opened 2018-03-27
ISPN-8995 Use mockito-core instead of mockito-all

tristantarrant:
https://issues.jboss.org/browse/ISPN-8995


wburns:

PR #5873, opened 2018-03-27
ISPN-8494 Clear is leaking transaction with Batching

pruivo:
https://issues.jboss.org/browse/ISPN-8494

reviewer notes:
* changes are in CacheImpl and BatchingInterceptor
* tests are in AbstractBatchTest
* the remaining changes are cleanups to the tests

### DIFF
--- a/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
+++ b/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
@@ -590,7 +590,7 @@ public class CacheImpl<K, V> implements AdvancedCache<K, V> {
    }
 
    private void transactionalRemoveGroup(String groupName, long explicitFlagsBitSet) {
-      final boolean onGoingTransaction = getOngoingTransaction() != null;
+      final boolean onGoingTransaction = getOngoingTransaction(true) != null;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -1863,8 +1863,8 @@ public V putIfAbsent(K key, V value, Metadata metadata) {
       return putAsync(key, value, metadata, EnumUtil.EMPTY_BIT_SET, getInvocationContextWithImplicitTransaction(false, 1));
    }
 
-   private Transaction suspendOngoingTransactionIfExists() {
-      final Transaction tx = getOngoingTransaction();
+   private Transaction suspendOngoingTransactionIfExists(boolean skipBatching) {
```
danberindei: I don't think this is (or should be) ever called with `skipBatching == false`, so I suggest removing the parameter.

I would also suggest reversing the parameter for `getOngoingTransaction`, into `includeBatchTx` (or even `includeSuspendedBatchTx`).

pruivo:
@danberindei updated!

danberindei:
Integrated, thanks Pedro!

PR #5877, opened 2018-03-27
ISPN-8994 ClusterHealth.getHealthStatus() improvement

karesti:
https://issues.jboss.org/browse/ISPN-8994

### DIFF
--- a/core/src/main/java/org/infinispan/health/impl/CacheHealthImpl.java
+++ b/core/src/main/java/org/infinispan/health/impl/CacheHealthImpl.java
@@ -1,6 +1,5 @@
 package org.infinispan.health.impl;
 
-import org.infinispan.AdvancedCache;
 import org.infinispan.Cache;
 import org.infinispan.distribution.DistributionManager;
 import org.infinispan.health.CacheHealth;
@@ -9,36 +8,36 @@ import org.infinispan.partitionhandling.AvailabilityMode;
 
 public class CacheHealthImpl implements CacheHealth {
 
-    private final AdvancedCache<?, ?> cache;
+   private final Cache<?, ?> cache;
 
-    public CacheHealthImpl(Cache<?, ?> cache) {
-        this.cache = cache.getAdvancedCache();
-    }
+   public CacheHealthImpl(Cache<?, ?> cache) {
+      this.cache = cache;
+   }
 
-    @Override
-    public String getCacheName() {
-        return cache.getName();
-    }
+   @Override
+   public String getCacheName() {
+      return cache.getName();
+   }
 
-    @Override
-    public HealthStatus getStatus() {
-        if (!isComponentHealthy() || cache.getAvailability() == AvailabilityMode.DEGRADED_MODE) {
-            return HealthStatus.UNHEALTHY;
-        }
-        DistributionManager distributionManager = SecurityActions.getDistributionManager(cache);
-        if (distributionManager != null && distributionManager.isRehashInProgress()) {
-            return HealthStatus.REBALANCING;
-        }
-        return HealthStatus.HEALTHY;
-    }
+   @Override
+   public HealthStatus getStatus() {
+      if (!isComponentHealthy() || cache.getAdvancedCache().getAvailability() == AvailabilityMode.DEGRADED_MODE) {
+         return HealthStatus.UNHEALTHY;
+      }
+      DistributionManager distributionManager = SecurityActions.getDistributionManager(cache);
+      if (distributionManager != null && distributionManager.isRehashInProgress()) {
+         return HealthStatus.REBALANCING;
+      }
+      return HealthStatus.HEALTHY;
+   }
 
-    private boolean isComponentHealthy() {
-        switch (cache.getStatus()) {
-            case INSTANTIATED:
-            case RUNNING:
-                return true;
-            default:
-                return false;
-        }
-    }
+   private boolean isComponentHealthy() {
+      switch (cache.getStatus()) {
+         case INSTANTIATED:
+         case RUNNING:
+            return true;
+         default:
+            return false;
+      }
+   }
 }
--- a/core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
+++ b/core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
@@ -2,53 +2,61 @@ package org.infinispan.health.impl;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.infinispan.health.ClusterHealth;
 import org.infinispan.health.HealthStatus;
 import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.registry.InternalCacheRegistry;
 
 public class ClusterHealthImpl implements ClusterHealth {
 
-    private final EmbeddedCacheManager cacheManager;
+   private final EmbeddedCacheManager cacheManager;
+   private final InternalCacheRegistry internalCacheRegistry;
 
-    public ClusterHealthImpl(EmbeddedCacheManager cacheManager) {
-        this.cacheManager = cacheManager;
-    }
+   public ClusterHealthImpl(EmbeddedCacheManager cacheManager) {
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/infinispan/health/impl/CacheHealthImpl.java
```diff
@@ -9,36 +8,36 @@
...
-        }
-        DistributionManager distributionManager = SecurityActions.getDistributionManager(cache);
-        if (distributionManager != null && distributionManager.isRehashInProgress()) {
-            return HealthStatus.REBALANCING;
-        }
-        return HealthStatus.HEALTHY;
-    }
+   @Override
+   public HealthStatus getStatus() {
+      if (!isComponentHealthy() || cache.getAdvancedCache().getAvailability() == AvailabilityMode.DEGRADED_MODE) {
```
slaskawi: Why do we need `cache.getAdvancedCache()` here?

core/src/main/java/org/infinispan/health/impl/CacheHealthImpl.java
```diff
@@ -9,36 +8,36 @@
...
-        switch (cache.getStatus()) {
-            case INSTANTIATED:
-            case RUNNING:
-                return true;
-            default:
-                return false;
-        }
-    }
+   private boolean isComponentHealthy() {
+      switch (cache.getAdvancedCache().getStatus()) {
```
slaskawi: And here?

core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
```diff
@@ -9,46 +9,56 @@
...
+   public ClusterHealthImpl(EmbeddedCacheManager cacheManager) {
+      this.cacheManager = cacheManager;
+      internalCacheRegistry = cacheManager.getGlobalComponentRegistry().getComponent(InternalCacheRegistry.class);
+   }
+
+   @Override
+   public HealthStatus getHealthStatus() {
+      HealthStatus globalHealthStatus = HealthStatus.HEALTHY;
+
+      Set<HealthStatus> userCacheHealthStatus = getHealthStatus(cacheManager.getCacheNames());
```
slaskawi: I think the logic below would be much easier if you created a map of Cache names / Health Status and then just check if any of those caches is unhealthy or rebalancing.

core/src/test/java/org/infinispan/health/ClusterHealthImplTest.java
```diff
@@ -1,107 +1,181 @@
...
-        DefaultCacheManager cacheManagerWithNullTransport = mock(DefaultCacheManager.class);
-        clusterHealth = new ClusterHealthImpl(cacheManagerWithNullTransport);
-
-        //when
-        List<String> nodeNames = clusterHealth.getNodeNames();
-
-        //then
-        assertEquals(nodeNames.size(), 0);
-    }
+   private static final String INTERNAL_CACHE_NAME = "internal_cache";
```
slaskawi: In very much against what happened here...

We introduced a change in the Health Check logic and in the same Pull Request we rewrote this whole test suite. Some tests have been preserved (though with changed name) and some were wiped out (like `testReturningNodeNamesWithNullTransport` which has been discovered long after implementing this functionality and it appears when using CLI on the server with local caches). In my opinion this is one of the easiest ways to introduce new regressions, which won't be found by the CI because of changed test suite.

Since we extended the health check functionality, I would prefer to add some new tests to this test suite (instead of rewriting it). This way we would be sure that we didn't add any new regressions and at the same time we would test new functionality. If for some reason this test suite can be optimized, please do so, but in a separate Pull Request so that we are 100% we didn't break anything.

core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
pruivo: I had another idea in mind. just concat the streams like
```
Stream.concat(cacheManager.getCacheNames().stream(), internalCacheRegistry.getInternalCacheNames().stream())
.map(....
```

core/src/test/java/org/infinispan/health/ClusterHealthImplTest.java
```diff
@@ -1,107 +1,181 @@
...
+   }
+
+   public void testGetNumberOfNodesWithNullTransport() throws Exception {
+      ClusterHealth clusterHealth = new ClusterHealthImpl(mockedCacheManager);
+
+      assertEquals(1, clusterHealth.getNumberOfNodes());
+   }
+
+   public void testGetNodeNamesWithNullTransport() throws Exception {
+      ClusterHealth clusterHealth = new ClusterHealthImpl(mockedCacheManager);
```
karesti: @slaskawi I rewrote the testReturningNodeNamesWithNullTransport

core/src/test/java/org/infinispan/health/ClusterHealthImplTest.java
karesti: I payed much attention not removing any test, I just changed to AssertJUnit and renamed.
The original test was mixing thins in a way where tests where getting overlapped with the mocked cache manager. The healty status was testing nothing neither, because even when I deleted lines on the code and test was failing

karesti: In order to add the fucntionality, I needed to rework on the test

core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
karesti: Ok

core/src/main/java/org/infinispan/health/impl/CacheHealthImpl.java
karesti: this is something I wanted to discuss on the PR.
the CacheHealthImpl has originally advancedCache instead of cache
When I looked to the code of `SecurityActions.getDistributionManager(cache);`

the code does again `getAdvancedCache`
so we where doing `getAdvancedCache` and again `getAdvancedCache`

I used mockito for the rebalancing cover. That's how I realised.

core/src/main/java/org/infinispan/health/impl/CacheHealthImpl.java
karesti: same reason as before, it was `advancedCache

core/src/test/java/org/infinispan/health/ClusterHealthImplTest.java
karesti: moreover, unhealthy or rebalancing status was not tested at all.

karesti: I first covered the code and after on TDD style, I added tests for internal caches that proved that my new code was failing til I added more core to correct it. Caches where getting created calling this method even if we did not want to create any cache

core/src/test/java/org/infinispan/health/ClusterHealthImplTest.java
slaskawi: Ok, got it. Thanks!

karesti:
@pruivo any other comment ?

core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java
```diff
@@ -5,50 +5,57 @@
...
+      internalCacheRegistry = cacheManager.getGlobalComponentRegistry().getComponent(InternalCacheRegistry.class);
+   }
+
+   @Override
+   public HealthStatus getHealthStatus() {
+      HealthStatus globalHealthStatus = HealthStatus.HEALTHY;
+
+      Set<HealthStatus> healthStatuses = Stream.concat(cacheManager.getCacheNames().stream(), internalCacheRegistry.getInternalCacheNames().stream())
+            .map(cacheName -> cacheManager.getCache(cacheName, false))
+            .filter(cache -> cache != null)
```
pruivo: just a minor thing: it can be replaced by methods reference
```
.filter(Objects::nonNull)
.map(CacheHealthImpl::new)
.map(CacheHealthImpl::getStatus)
```

karesti: done

pruivo:
Integrated! thanks @karesti !

PR #5878, opened 2018-03-28
ISPN-9013 Partition Handling enabled toggle should set merge-policy=NONE

danberindei:
https://issues.jboss.org/browse/ISPN-9013

Added deprecation javadoc and fix log message.

### DIFF
--- a/core/src/main/java/org/infinispan/util/logging/Log.java
+++ b/core/src/main/java/org/infinispan/util/logging/Log.java
@@ -1591,7 +1591,7 @@ public interface Log extends BasicLogger {
    IllegalStateException getConflictsStateTransferInProgress(String cacheName);
 
    @LogMessage(level = WARN)
-   @Message(value = "The partition handling 'enable' attribute has been deprecated. Please update your configuration file to use the 'type' attribute instead", id = 464)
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/util/logging/Log.java
```diff
@@ -1591,7 +1591,7 @@ TimeoutException coordinatorTimeoutWaitingForView(int expectedViewId, int curren
    IllegalStateException getConflictsStateTransferInProgress(String cacheName);
 
    @LogMessage(level = WARN)
-   @Message(value = "The partition handling 'enable' attribute has been deprecated. Please update your configuration file to use the 'type' attribute instead", id = 464)
+   @Message(value = "The partition handling 'enable' attribute has been deprecated. Please update your configuration to use 'when-split' instead", id = 464)
```
tristantarrant: s/enable/enabled/

danberindei:
Fixed @tristantarrant

wburns:
I am fine with the changes, but thinking about this more. If we are doing this for backwards compatibility shouldn't the default be `NONE` no matter what? @tristantarrant what do you think?

tristantarrant:
I think we should have better defaults win over blind backwards compatibility. As long as a user can configure the old behaviour, I'm fine with it.

PR #5879, opened 2018-03-28
ISPN-8911 Copy dependencies for the distribution only for specific modules

tristantarrant:
https://issues.jboss.org/browse/ISPN-8911

### DIFF
--- a/demos/gui/pom.xml
+++ b/demos/gui/pom.xml
@@ -13,6 +13,10 @@
    <name>Infinispan GUI Demo</name>
    <description>Infinispan - GUI Demo</description>
 
+   <properties>
+      <module.skipCopyDependencies>false</module.skipCopyDependencies>
+   </properties>
+
    <dependencies>
       <dependency>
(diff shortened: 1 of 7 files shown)


tristantarrant:
Hints for testing: mvn -Pdistribution clean install 
Then look at server/integration/target/distribution/infinispan-server-x.y.z-bin.zip: it should not contain a dependency directory.
Then look at distribution/target/distribution and examine the all and minimal distributions. Try to run the GUI demo and check that the lib directory contains dependencies for the supplied modules.

ryanemerson:
@tristantarrant The changes look good to me, however when I attempt to run the gui demo I get the following error: `Error: Could not find or load main class org.infinispan.Version`, however this also exists on the master branch. I have created https://issues.jboss.org/browse/ISPN-9055 to track the issue.

ryanemerson:
Thanks @tristantarrant, the above error was not really an issue as you explained.

PR #5883, opened 2018-03-29
ISPN-8994 fix cleanAfterMethod

karesti:
The cleanup is not properly done and if the order of the tests changez, we can have unhealthy caches and user created caches from other tests


rvansa:
Integrated, thanks.

PR #5886, opened 2018-03-30
ISPN-8728 ExceptionEvictionTest.testSizeCorrectWithStateTransfer random

wburns:
failures

* Use eventually since no way to wait for stale entries to be removed

https://issues.jboss.org/browse/ISPN-8728


tristantarrant:
Merged, thanks

PR #5887, opened 2018-04-01
[do not integrate] jdg1729

anistor:
A test for infinispan extension exercising more api than just the cache api.
=> We get NoClassDefFoundError unless we export "org.infinispan"


anistor:
@ryanemerson Should we or should we not export "org.infinispan"? :)

PR #5888, opened 2018-04-02
Update rocksdb

schernolyas:
The PR contains:
1) update version of RocksDB
2) use inner property of RocksDB for count keys (see https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ)

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -89,7 +89,7 @@
       <version.osgi>4.3.1</version.osgi>
       <version.protostream>4.2.0.CR2</version.protostream>
       <version.reactivestreams>1.0.1</version.reactivestreams>
-      <version.rocksdb>5.8.0</version.rocksdb>
+      <version.rocksdb>5.11.3</version.rocksdb>
       <version.rxjava>2.1.3</version.rxjava>
       <version.slf4j-jboss-logging>1.1.0.Final</version.slf4j-jboss-logging>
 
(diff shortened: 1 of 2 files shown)


ryanemerson:
test this please

ryanemerson:
@schernolyas We can't upgrade the RocksDB minor version until the 9.3 release.

Also, I'm  a little concerned about how reliable `rocksdb.estimate-num-keys` will be for our needs considering that it's very likely that the same key will be overwritten many times.  From their docs:

> RocksDB can estimate number of keys through DB property “rocksdb.estimate-num-keys”. Note this estimation can be far off when there are merge operators, existing keys overwritten, or deleting non-existing keys.

schernolyas:
Hi @ryanemerson !

Why you can't update version of RocksDB in current release ? I think, I need lead own version of Infinispan :-(

About the property. Now Infinispan count all keys :-(  It is too long operation and it blocks usage of Web/CLI .

wburns:
Also another problem is that this method doesn't take into account expired entries. Although we could check the expiredDB and expiryEntryQueue and if they are empty, invoking this method may be okay. Although as Ryan said it is still an estimate.

schernolyas:
Hi colleagues!

I understood. Thanks!

wburns:
I am not necessarily opposed to this change, however I don't think it will be beneficial until 9.3 when we can use size more appropriately. We could try to push some support into 9.2 as well even.

However we would have to see how much of an approximation this is. If it is just off by very little and for a very short time period I would be open to allowing it.

PR #5889, opened 2018-04-02
ISPN-9030 Avoid anonymous classes inside lambdas

danberindei:
https://issues.jboss.org/browse/ISPN-9030

Only changing TxInterceptor and HibernateSearchPropertyHelper for now because that's where we saw the errors.

I've been trying to find all the lambdas that contain an inner class but I've hit a bug in structural search ([IDEA-189344](https://youtrack.jetbrains.com/issue/IDEA-189344)), and I have a feeling the 2 IntelliJ bugs are related.

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
@@ -279,71 +279,7 @@ public class TxInterceptor<K, V> extends DDAsyncInterceptor implements JmxStatis
          command.setFlagsBitSet(FlagBitSets.REMOTE_ITERATION);
          return invokeNextThenApply(ctx, command, (rCtx, rCommand, rv) -> {
             CacheSet<K> set = (CacheSet<K>) rv;
-            return new AbstractDelegatingKeyCacheSet<K, V>(Caches.getCacheWithFlags(cache, (FlagAffectedCommand) rCommand), set) {
-               @Override
-               public CloseableIterator<K> iterator() {
-                  if (isRemoteIteration) {
-                     return innerIterator();
-                  }
-                  // Need to support remove of iterator
-                  return new RemovableCloseableIterator<>(innerIterator(), cache::remove);
-               }
-
-               CloseableIterator<K> innerIterator() {
-                  return new TransactionAwareKeyCloseableIterator<>(super.iterator(),
-                        (TxInvocationContext<LocalTransaction>) rCtx, cache);
-               }
-
-               @Override
-               public void clear() {
-                  cache.clear();
-               }
-
-               @Override
-               public boolean contains(Object o) {
-                  boolean contained = false;
-                  if (o != null) {
-                     CacheEntry contextEntry = rCtx.lookupEntry(o);
-                     if (contextEntry == null) {
-                        contained = super.contains(o);
-                     } else {
-                        contained = !contextEntry.isRemoved();
-                     }
-                  }
-                  return contained;
-               }
-
-               @Override
-               public boolean remove(Object o) {
-                  boolean removed = false;
-                  if (o != null) {
-                     removed = cache.remove(o) != null;
-                  }
-                  return removed;
-               }
-
-               @Override
-               public CloseableSpliterator<K> spliterator() {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
```diff
@@ -858,4 +717,168 @@ protected E getNextFromIterator() {
          return returnedValue;
       }
    }
+
+   private class TxKeyCacheSet extends AbstractDelegatingKeyCacheSet<K, V> {
+      private final boolean isRemoteIteration;
+      private final InvocationContext rCtx;
+
+      public TxKeyCacheSet(VisitableCommand rCommand, CacheSet<K> set, boolean isRemoteIteration,
```
anistor: no need for ```public```, package local is good.

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
```diff
@@ -858,4 +717,168 @@ protected E getNextFromIterator() {
...
+         return (int) size;
+      }
+   }
+
+   private class TxEntryCacheSet extends AbstractDelegatingEntryCacheSet<K, V> {
+
+      private final InvocationContext rCtx;
+      private final boolean isRemoteIteration;
+
+      public TxEntryCacheSet(VisitableCommand rCommand, CacheSet<CacheEntry<K, V>> set, InvocationContext rCtx,
```
anistor: no need for ```public```, package local is good.

query/src/main/java/org/infinispan/query/dsl/embedded/impl/HibernateSearchPropertyHelper.java
```diff
@@ -383,4 +357,37 @@ private ResolvedProperty resolveProperty(EntityIndexBinding entityIndexBinding,
       }
       return null;
    }
+
+   private class HibernateSearchMetadata implements IndexedFieldProvider.FieldIndexingMetadata {
+      private final EntityIndexBinding entityIndexBinding;
+
+      public HibernateSearchMetadata(EntityIndexBinding entityIndexBinding) {
```
anistor: no need for ```public```, package local is good.

anistor:
If it's just these two, ok, let's fix them. If there are more ... I would not bother. There are workarounds.

query/src/main/java/org/infinispan/query/dsl/embedded/impl/HibernateSearchPropertyHelper.java
```diff
@@ -383,4 +357,37 @@ private ResolvedProperty resolveProperty(EntityIndexBinding entityIndexBinding,
       }
       return null;
    }
+
+   private class HibernateSearchMetadata implements IndexedFieldProvider.FieldIndexingMetadata {
```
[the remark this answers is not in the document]
danberindei: I wasn't sure whether to take this comment seriously, but I made the change :)

danberindei:
@anistor I always got the error in one of these 2 classes, but even if it was 20 classes I think it would still be worth it: I spend at least 30 mins every time fixing the compilation error every time it appears.

anistor:
30?? I spend about 3 mins when it happens. I think we need to move to another IDE....

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
```diff
@@ -858,4 +717,168 @@ protected E getNextFromIterator() {
          return returnedValue;
       }
    }
+
+   private class TxKeyCacheSet extends AbstractDelegatingKeyCacheSet<K, V> {
+      private final boolean isRemoteIteration;
+      private final InvocationContext rCtx;
+
+      TxKeyCacheSet(VisitableCommand rCommand, CacheSet<K> set, boolean isRemoteIteration,
```
anistor: I'd like ```isRemoteIteration``` to be the last arg so it looks identical to ```TxEntryCacheSet``` constructor. Let me do this refactoring myself and merge.

anistor:

PR #5894, opened 2018-04-04
JDBC Oracle & DB2 Fixes

ryanemerson:
https://issues.jboss.org/browse/ISPN-9040
https://issues.jboss.org/browse/ISPN-9041

### DIFF
--- a/persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/table/management/DB2TableManager.java
+++ b/persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/table/management/DB2TableManager.java
@@ -22,6 +22,15 @@ class DB2TableManager extends AbstractTableManager {
    }
 
    @Override
+   public String getInsertRowSql() {
+      if (insertRowSql == null) {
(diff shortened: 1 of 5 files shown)


persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/table/management/DB2TableManager.java
```diff
@@ -21,6 +21,15 @@
       super(connectionFactory, config, metaData, LOG);
    }
 
+   @Override
+   public String getInsertRowSql() {
+      if (insertRowSql == null) {
+         insertRowSql = String.format("INSERT INTO %s (%s,%s,%s) VALUES (?,?,?)", getTableName(),
+               config.idColumnName(), config.timestampColumnName(), config.dataColumnName());
```
pruivo: shouldn't the insert have the same order as upsert?

ryanemerson: It is, the only difference is the order of the `String.format(...)` params not the set values.

pruivo:
integrated! thanks @ryanemerson !

PR #5896, opened 2018-04-04
ISPN-8719 KeySet.(iterator|spliterator|stream) not compatible with

wburns:
versions before 9.2

* Added iterator methods to Codecs
* Put appropriate methods on each Codec version

https://issues.jboss.org/browse/ISPN-8719

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/ProtocolVersion.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/ProtocolVersion.java
@@ -14,21 +14,28 @@ import java.util.stream.Collectors;
  */
 public enum ProtocolVersion {
 
-   PROTOCOL_VERSION_27(2, 7),
-   PROTOCOL_VERSION_26(2, 6),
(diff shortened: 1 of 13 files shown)


wburns:
I have added tests and they all seem to work except for ProtocolVersion.2.3 where for some reason the result is getting jumbled and the messageId is found to be 0 causing a NPE. I am not sure where the bug is as the server side seems to be sending the correct bytes back. I need to look into it further still.

wburns:
So the cause with version 2.3 is due to the fact that the retrieveEntries in newer clients always uses the 2.4 protocol even when 2.3 is set. Also newer clients can't communicate to any server of version 2.2 or older in general. So that means no client since 8.1 could ever talk to any version older than that.

wburns:
Also fixed a 2.5+ client talking to 2.4 since it was always passing meta but that is not returned in 2.4. This would make reading entries return incorrect values and/or possibly hang.

tristantarrant:
This looks great. We will eventually need some integration tests for testing against combinations of servers/clients.

PR #5899, opened 2018-04-06
ISPN-9025 Transaction leak when API invoked with invalid arguments

pruivo:
https://issues.jboss.org/browse/ISPN-9025

### DIFF
--- a/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
+++ b/core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
@@ -129,6 +129,7 @@ import org.infinispan.transaction.xa.TransactionXaAdapter;
 import org.infinispan.transaction.xa.XaTransactionTable;
 import org.infinispan.util.concurrent.CompletableFutures;
 import org.infinispan.util.concurrent.locks.LockManager;
+import org.infinispan.util.concurrent.locks.RemoteLockCommand;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
 
@@ -174,6 +175,8 @@ public class CacheImpl<K, V> implements AdvancedCache<K, V> {
    private volatile boolean stopping = false;
    private boolean transactional;
    private boolean batchingEnabled;
+   private final ContextBuilder contextBuilder = this::getInvocationContextWithImplicitTransaction;
+   private final ContextBuilder pferContextBuilder = this::putForExternalReadContext;
 
    public CacheImpl(String name) {
       this.name = name;
@@ -285,17 +288,15 @@ public class CacheImpl<K, V> implements AdvancedCache<K, V> {
    }
 
    private V computeInternal(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction, boolean computeIfPresent, Metadata metadata, long flags) {
-      return computeInternal(key, remappingFunction, computeIfPresent, metadata, flags, getInvocationContextWithImplicitTransaction(false, 1));
+      return computeInternal(key, remappingFunction, computeIfPresent, metadata, flags, contextBuilder);
    }
 
-   V computeInternal(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction, boolean computeIfPresent, Metadata metadata, long flags, InvocationContext ctx) {
+   V computeInternal(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction, boolean computeIfPresent,
+         Metadata metadata, long flags, ContextBuilder contextBuilder) {
       assertKeyNotNull(key);
       assertFunctionNotNull(remappingFunction);
       ComputeCommand command = commandsFactory.buildComputeCommand(key, remappingFunction, computeIfPresent, metadata, flags);
-      if (ctx.getLockOwner() == null) {
-         ctx.setLockOwner(command.getKeyLockOwner());
-      }
-      return (V) executeCommandAndCommitIfNeeded(ctx, command);
+      return executeCommandAndCommitIfNeeded(contextBuilder, command, 1);
    }
 
    @Override
@@ -306,24 +307,21 @@ public class CacheImpl<K, V> implements AdvancedCache<K, V> {
    @Override
    public V computeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction, Metadata metadata) {
       return computeIfAbsentInternal(key, mappingFunction, metadata, addUnsafeFlags(EnumUtil.EMPTY_BIT_SET),
-            getInvocationContextWithImplicitTransaction(false, 1));
+            contextBuilder);
    }
 
-   V computeIfAbsentInternal(K key, Function<? super K, ? extends V> mappingFunction, Metadata metadata, long flags, InvocationContext ctx) {
+   V computeIfAbsentInternal(K key, Function<? super K, ? extends V> mappingFunction, Metadata metadata, long flags,
+         ContextBuilder contextBuilder) {
       assertKeyNotNull(key);
       assertFunctionNotNull(mappingFunction);
       ComputeIfAbsentCommand command = commandsFactory.buildComputeIfAbsentCommand(key, mappingFunction, metadata, flags);
-      if (ctx.getLockOwner() == null) {
-         ctx.setLockOwner(command.getKeyLockOwner());
-      }
-      return (V) executeCommandAndCommitIfNeeded(ctx, command);
+      return executeCommandAndCommitIfNeeded(contextBuilder, command, 1);
    }
 
    @Override
    public V merge(K key, V value, BiFunction<? super V, ? super V, ? extends V> remappingFunction) {
-      return mergeInternal(key, value, remappingFunction,
-            defaultMetadata, addUnsafeFlags(EnumUtil.EMPTY_BIT_SET),
-            getInvocationContextWithImplicitTransaction(false, 1));
+      return mergeInternal(key, value, remappingFunction, defaultMetadata, addUnsafeFlags(EnumUtil.EMPTY_BIT_SET),
+            contextBuilder);
    }
 
    @Override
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -285,17 +286,15 @@ private V compute(K key, BiFunction<? super K, ? super V, ? extends V> remapping
    }
 
    private V computeInternal(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction, boolean computeIfPresent, Metadata metadata, long flags) {
-      return computeInternal(key, remappingFunction, computeIfPresent, metadata, flags, getInvocationContextWithImplicitTransaction(false, 1));
+      return computeInternal(key, remappingFunction, computeIfPresent, metadata, flags, this::singleKeyContextWithImplicitTransaction);
```
rvansa: Please cache the method reference - this is a capturing method handle => allocation.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -622,42 +607,34 @@ private void nonTransactionalRemoveGroup(String groupName, long explicitFlags) {
          //executed. If the context is already populated, it throws a ClassCastException because the wrapForRemove is
          //not invoked.
          assertKeyNotNull(key);
-         InvocationContext ctx = getInvocationContextWithImplicitTransaction(false, 1);
-         RemoveCommand command = createRemoveCommand(key, removeFlags, ctx);
+         InvocationContext ctx = singleKeyContextWithImplicitTransaction();
+         RemoveCommand command = createRemoveCommand(key, removeFlags);
+         checkLockOwner(ctx, command);
```
rvansa: I am a bit confused that `executeCommandAndCommitIfNeeded(ctx, command)` does not set the lock owner and `executeCommandAndCommitIfNeeded(contextBuilder, command)` does - please move the `checkLockOwner` to the former.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -1143,7 +1135,6 @@ public AuthorizationManager getAuthorizationManager() {
 
    @Override
    public AdvancedCache<K, V> lockAs(Object lockOwner) {
-      Objects.nonNull(lockOwner);
```
rvansa: Why? DecoratedCache does not do the null check.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -1411,32 +1390,29 @@ public final void putAll(Map<? extends K, ? extends V> map, long lifespan, TimeU
...
       explicitFlags = EnumUtil.mergeBitSets(explicitFlags, FlagBitSets.IGNORE_RETURN_VALUES);
-      PutMapCommand command = createPutAllCommand(map, metadata, explicitFlags, ctx);
-      executeCommandAndCommitIfNeeded(ctx, command);
+      PutMapCommand command = createPutAllCommand(map, metadata, explicitFlags);
+      executeCommandAndCommitIfNeeded(contextBuilder, command);
    }
 
    public final Map<K, V> getAndPutAll(Map<? extends K, ? extends V> map) {
-      return getAndPutAll(map, defaultMetadata, EnumUtil.EMPTY_BIT_SET, getInvocationContextWithImplicitTransaction(false, map.size()));
+      return getAndPutAll(map, defaultMetadata, EnumUtil.EMPTY_BIT_SET, () -> getInvocationContextWithImplicitTransaction(map.size()));
```
rvansa: I wonder if sizing the lookupEntries in context justifies this allocation enough - I would rather drop it and create multi-key context all the time (and if you want to optimize the single-key case you could supply the proper method handle right here - though I think it's an unnecessary optimization).

core/src/main/java/org/infinispan/cache/impl/EncoderCache.java
```diff
@@ -860,4 +856,16 @@ public void addListener(Object listener) {
       initState(cache, this);
       return cache;
    }
+
+   private BiFunctionMapper wrapBiFunction(BiFunction<?, ?, ?> biFunction) {
```
rvansa: The compute/merge functions must be non-null according to `ConcurrentMap` javadoc.

core/src/test/java/org/infinispan/api/APINonTxTest.java
```diff
@@ -622,7 +665,7 @@ public void testMerge() throws Exception {
 
       // remove if null value after remapping
       cache.merge("A", "C", (oldValue, newValue) -> null);
-      assertEquals(null, cache.get("A"));
+      assertNull(cache.get("A"));
```
rvansa: I prefer `assertEquals(null, ` since the error message contains the value; `assertNull` just fails.

pruivo: ok

core/src/main/java/org/infinispan/cache/impl/EncoderCache.java
pruivo: that's why I fixed. we were wrapping the null function and it only fails later in the code. now, it doesn't wrap and the superclass checks for null.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
pruivo: I did some refactorization. now the context builder receives the key count as parameter

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
pruivo: `nonNull()` is a no-op since it only returns `true` if the `lockOwner` is `null`. No logic changes here.

However, I'm not sure if it can or not be null, but the `DecoratedCache` always check the field for `null`. Example:
```
if (lockOwner != null) {
     ctx.setLockOwner(lockOwner);
}
```

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
pruivo: removed executeCommandAndCommitIfNeeded(ctx, command). it was only used in one place anyway

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
pruivo: I'm not a big fan of caching this kind of stuff. It is memory always used, independently if the cache is used or not. and the cache already has a lot of fields. 

only if you insist, I'll change it :)

rvansa: I dislike caching method references as well (ideally this should be done by our clever JVM), but when the con is an allocation on each method call, I think that such ugly optimization makes sense. @danberindei wdyt?

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
rvansa: Eh, right, I've mistaken it for `Objects.requireNonNull`. It seems that calling this with `null` parameter is clearly unintentional, the decorated cache does not decorate anything, then.

pruivo: ok. adding requireNonNull

core/src/main/java/org/infinispan/cache/impl/EncoderCache.java
pruivo: I would like to keep it consistency with the remaining methods (put, replace, etc). The `keyToStorage()` and `valueToStorage()` pass `null` to the superclass too.

rvansa:
@danberindei In case you missed the previous question... Pedro prefers using `this::someMethod` as an argument to some method calls. Will you convince him to cache the method handle?

rvansa:
Otherwise this PR looks good.

rvansa:
@danberindei This PR is still waiting for your attention.

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
danberindei: @pruivo, I agree with Radim: if storing one lambda in a `CacheImpl` field is too much, then creating a new lambda instance on every invocation is much much worse.

I see in https://github.com/infinispan/infinispan/pull/5458 I only added 2 fields, but I have no idea now why I closed that PR.

danberindei:
I also suggest adding this as well: https://github.com/infinispan/infinispan/pull/5458/files#diff-439b4a1c3c5af3b0206440bb243e742cR63

danberindei:
run performance tests please

pruivo:
@danberindei @rvansa updated as requested :(

core/src/main/java/org/infinispan/cache/impl/CacheImpl.java
```diff
@@ -174,9 +175,13 @@
    private volatile boolean stopping = false;
    private boolean transactional;
    private boolean batchingEnabled;
+   private final ContextBuilder contextBuilder;
+   private final ContextBuilder pferContextBuilder;
 
    public CacheImpl(String name) {
       this.name = name;
+      this.contextBuilder = this::getInvocationContextWithImplicitTransaction;
```
rvansa: I would prefer to initialize these inline in field declaration rather than constructor... If I am not nitpicking too much...

core/src/main/java/org/infinispan/cache/impl/DecoratedCache.java
```diff
@@ -50,6 +50,7 @@
    private final long flags;
    private final Object lockOwner;
    private final CacheImpl<K, V> cacheImplementation;
+   private final CacheImpl.ContextBuilder contextBuilder;
```
rvansa: the same here

pruivo:
@rvansa done! don't forget to let CI do a full run :)

danberindei:
run performance tests please

rvansa:
I don't think the EntityCollectionInvalidationTest failure is related, despite I really dislike to see such failure. Integrating.

danberindei:
Looks like performance didn't change too much by caching the creator function. And the -2.95% throughput on repl non-tx writes with 100 threads is a bit worrisome...

rvansa:
Uh, I haven't noticed that one. Could you guess the cause @pruivo ?

I wouldn't expect a noticeable difference for a one lambda allocation, that hides easily into the noise.

pruivo:
IMO, it is noise. If it was something serious, it would affect all the writes and the dist-writes with 100 threads are fine.

danberindei:
run performance tests please

rvansa:
@danberindei The noise might not be in the run we trigger here but also in the baseline run (being too good).

PR #5902, opened 2018-04-06
Conflict Resolution Fixes 

ryanemerson:
https://issues.jboss.org/browse/ISPN-9014
https://issues.jboss.org/browse/ISPN-9089
https://issues.jboss.org/browse/ISPN-9103
https://issues.jboss.org/browse/ISPN-9032

### DIFF
--- a/core/src/main/java/org/infinispan/conflict/impl/StateReceiverImpl.java
+++ b/core/src/main/java/org/infinispan/conflict/impl/StateReceiverImpl.java
@@ -2,6 +2,7 @@ package org.infinispan.conflict.impl;
 
 import static org.infinispan.factories.KnownComponentNames.STATE_TRANSFER_EXECUTOR;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -10,7 +11,6 @@ import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
-import java.util.stream.Collectors;
 
 import org.infinispan.Cache;
 import org.infinispan.commands.CommandsFactory;
@@ -140,8 +140,17 @@ public class StateReceiverImpl<K, V> implements StateReceiver<K, V> {
 
       synchronized CompletableFuture<List<Map<Address, CacheEntry<K, V>>>> requestState() {
          assert future == null;
-         if (trace) log.tracef("Cache %s attempting to receive replicas for segment %s from %s with topology %s",
-               cacheName, segmentId, replicaHosts, topology);
+         if (trace) log.tracef("Cache %s attempting to receive replicas for segment %s from %s with topologyId=%s, timeout=%d",
+               cacheName, segmentId, replicaHosts, topology.getTopologyId(), timeout);
+
+         future = new CompletableFuture<>();
+         future.whenComplete((v, t) -> {
+            if (t != null) {
+               if (trace) log.tracef("Cache %s segment request(s) cancelled due to exception=%s", cacheName, t);
+               // If an exception has occurred, possibly a CancellationException, we must must cancel all ongoing transfers
+               cancel(t);
+            }
+         });
 
          for (final Address replica : replicaHosts) {
             if (replica.equals(rpcManager.getAddress())) {
@@ -151,18 +160,21 @@ public class StateReceiverImpl<K, V> implements StateReceiver<K, V> {
                      addKeyToReplicaMap(replica, entry);
                   }
                });
+               // numOwner == 1, then we cannot rely on receiveState to complete the future
+               if (replicaHosts.size() == 1) {
+                  completeRequest();
+               }
             } else {
                final InboundTransferTask transferTask = createTransferTask(segmentId, replica, topology, timeout);
                transferTaskMap.put(replica, transferTask);
 
-               stateReceiverExecutor.executeAsync(() -> {
+               stateReceiverExecutor.execute(() -> {
                   // If the transferTaskMap does not contain an entry for this replica, then it must have been cancelled
                   // before this request was executed..
                   if (!transferTaskMap.containsKey(replica))
-                     return null;
+                     return;
 
-                  CompletableFuture<Void> transferStarted = transferTask.requestSegments();
-                  return transferStarted.exceptionally(throwable -> {
+                  transferTask.requestSegments().exceptionally(throwable -> {
                      if (trace) log.tracef(throwable, "Cache %s exception when processing InboundTransferTask", cacheName);
                      cancel(throwable);
                      return null;
@@ -170,7 +182,6 @@ public class StateReceiverImpl<K, V> implements StateReceiver<K, V> {
                });
             }
          }
-         future = new CompletableFuture<>();
          return future;
       }
 
@@ -205,18 +216,16 @@ public class StateReceiverImpl<K, V> implements StateReceiver<K, V> {
                transferTaskMap.remove(sender);
 
                if (transferTaskMap.isEmpty()) {
-                  List<Map<Address, CacheEntry<K, V>>> retVal = keyReplicaMap.entrySet().stream()
-                        .map(Map.Entry::getValue)
-                        .collect(Collectors.toList());
-                  clear();
-
-                  future.complete(Collections.unmodifiableList(retVal));
+                  completeRequest();
                }
             }
          }
       }
 
       synchronized void cancel(Throwable throwable) {
+         if (future.isDone())
+            return;
+
          log.debugf(throwable, "Cache %s cancelling request for segment %s", cacheName, segmentId);
          transferTaskMap.forEach((address, inboundTransferTask) -> inboundTransferTask.cancel());
          if (throwable != null) {
@@ -225,6 +234,12 @@ public class StateReceiverImpl<K, V> implements StateReceiver<K, V> {
             future.cancel(true);
          }
          clear();
+      }
+
+      synchronized void completeRequest() {
+         List<Map<Address, CacheEntry<K, V>>> retVal = new ArrayList<>(keyReplicaMap.values());
+         clear();
+         future.complete(Collections.unmodifiableList(retVal));
       }
 
       void addKeyToReplicaMap(Address address, CacheEntry<K, V> ice) {
--- a/core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
+++ b/core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
@@ -11,12 +11,15 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.infinispan.AdvancedCache;
 import org.infinispan.commons.CacheException;
 import org.infinispan.commons.util.Immutables;
 import org.infinispan.conflict.ConflictManagerFactory;
-import org.infinispan.conflict.impl.DefaultConflictManager;
+import org.infinispan.conflict.impl.InternalConflictManager;
 import org.infinispan.distribution.ch.ConsistentHash;
 import org.infinispan.distribution.ch.ConsistentHashFactory;
 import org.infinispan.globalstate.ScopedPersistentState;
@@ -27,6 +30,7 @@ import org.infinispan.partitionhandling.impl.AvailabilityStrategy;
 import org.infinispan.partitionhandling.impl.AvailabilityStrategyContext;
 import org.infinispan.remoting.transport.Address;
 import org.infinispan.remoting.transport.Transport;
+import org.infinispan.remoting.transport.jgroups.SuspectException;
 import org.infinispan.statetransfer.RebalanceType;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
@@ -75,6 +79,7 @@ public class ClusterCacheStatus implements AvailabilityStrategyContext {
    private volatile List<Address> queuedRebalanceMembers;
    private volatile boolean rebalancingEnabled = true;
    private volatile boolean rebalanceInProgress = false;
+   private volatile ConflictResolution conflictResolution;
 
    private RebalanceConfirmationCollector rebalanceConfirmationCollector;
    private ComponentStatus status;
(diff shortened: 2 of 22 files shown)


core/src/test/java/org/infinispan/conflict/impl/CrashedNodeDuringConflictResolutionTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.infinispan.remoting.inboundhandler.Reply;
+import org.infinispan.statetransfer.StateRequestCommand;
+import org.infinispan.test.TestingUtil;
+import org.infinispan.test.fwk.TransportFlags;
+
+
+/**
+ * 1. Partition cluster
+ * 2. When coordinator sends InboundTransferTask for segment crash a node from that segment
+ * 3. CR should timeout and the rebalance should occur as normal
```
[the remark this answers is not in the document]
ryanemerson: The test only takes ~10 seconds as that's what I set the state transfer timeout to in the test :smiley: 

But I agree,  cancelling the CR would be better.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -899,11 +901,82 @@ public boolean resolveConflictsOnMerge() {
...
+         return false;
+      }
+
+      CacheTopology t = queuedConflictTopology;
+      ConsistentHashFactory chf = getJoinInfo().getConsistentHashFactory();
+      ConsistentHash newHash = chf.updateMembers(t.getCurrentCH(), members, capacityFactors);
+
+      if (trace) log.tracef("Cache %s restarting conflict resolution with new consistent hash %s", cacheName, newHash);
+
+      queuedConflictTopology = new CacheTopology(t.getTopologyId() + 1, t.getRebalanceId() + 1,
```
ryanemerson: The topologyId needs to take into account the currentTopology Id, I'll push the change.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -908,4 +910,73 @@ public ConsistentHash calculateConflictHash(ConsistentHash preferredHash, Set<Co
...
+         // Check membership, if node removed then we must cancel the CR and start again below
+         // If currentTopology members size is one, cancel CR as it will be re-triggered on future merge
+         conflictFuture.cancel(true);
+      }
+
+      AdvancedCache<?, ?> cache = cacheManager.getCache(cacheName).getAdvancedCache();
+      InternalConflictManager<?, ?> conflictManager = (InternalConflictManager) ConflictManagerFactory.get(cache);
+      conflictFuture = conflictManager.resolveConflicts(queuedConflictTopology);
+      // Will execute using the StateTransferExecutor on completion
+      conflictFuture.thenRun(this::completeConflictResolution);
```
danberindei: I'd like to run `completeConflictResolution`, either to restart conflict resolution or to start a rebalance, even if `ConflictManagerFactory.get()` failed, or if `resolveConflicts` throws an exception.

If I understand correctly you don't restart CR immediately now, you only do it if there's a new graceful leave/cluster view? If so, that would leave CR in limbo if it times out, as it did in Diego's test.

core/src/test/java/org/infinispan/topology/TestClusterCacheStatus.java
```diff
@@ -135,7 +135,7 @@ public void startConflictResolution(ConsistentHash conflictCH, Address... mergeM
 
    public void startConflictResolution(ConsistentHash conflictCH, List<Address> mergeMembers) {
       topology = new CacheTopology(topology.getTopologyId() + 1, topology.getRebalanceId() + 1,
-                                   readConsistentHash(), conflictCH, conflictCH,
+                                   conflictCH, conflictCH, conflictCH,
```
danberindei: Ok, so removing only the union CH doesn't work, but now that I got you to use the conflict CH for reading, how about we only set the `currentCH` and change `CacheTopology` to use it for everything during `CONFLICT_RESOLUTION`?

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -736,6 +732,12 @@ public synchronized boolean doLeave(Address leaver) throws Exception {
    }
 
    public synchronized void startQueuedRebalance() {
+      // We cannot start rebalance until queued CR is complete
+      if (queuedConflictTopology != null) {
+         log.tracef("Postponing rebalance for cache %s as conflict resolution is in progress", cacheName);
+         return;
```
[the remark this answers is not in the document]
ryanemerson: I think we need `onConflictResolutionEnd` as different actions are taken by the two strategies.

core/src/test/java/org/infinispan/topology/TestClusterCacheStatus.java
ryanemerson: I was mistaken here, this still needs to be the readconsistent hash after all as this is how we determine the preferred partition in the ConflictManager.

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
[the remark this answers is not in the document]
danberindei: I think the actions are pretty much the same now, because in `PreferConsistencyStrategy` the availability mode should already be available and the list of members should be correct. In fact, if `updateMembersAndRebalance` has to do more than just queue the rebalance there, I'd argue that the `PreferAvailabilityStrategy` should be modified to do the same.

core/src/test/java/org/infinispan/topology/TestClusterCacheStatus.java
danberindei: That would undo the ISPN-9032 fix... maybe we could set the preferred read CH as the pendingCH for internal use, but use the currentCH (i.e. conflict CH) for both reading and writing?

Or we could just leave ISPN-9032 out of this PR and change the test to block the installation of the CONFLICT_RESOLUTION topology :)

ryanemerson: What I meant about still needing the currentCH was that atm we call `getCurrentCH` to determine the preferred partition in the conflict manager. But this is hacky anyway, so I'm going to change `resolveConlicts(topology)` to `resolveConflicts(topology, preferredOwners)` and then we can just use the `currentCH` as you suggested.

danberindei: Ok, got it now

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -899,11 +901,116 @@ public boolean resolveConflictsOnMerge() {
...
+         // Check membership, if node removed then we must cancel the CR and start again below
+         // If currentTopology members size is one, cancel CR as it will be re-triggered on future merge
+         conflictFuture.cancel(true);
+      }
+
+      AdvancedCache<?, ?> cache = cacheManager.getCache(cacheName).getAdvancedCache();
+      InternalConflictManager<?, ?> conflictManager = (InternalConflictManager) ConflictManagerFactory.get(cache);
+      conflictFuture = conflictManager.resolveConflicts(queuedConflictResolution.conflictTopology, queuedConflictResolution.preferredNodes);
+      // Will execute using the StateTransferExecutor on completion
+      conflictFuture.thenRun(this::completeConflictResolution);
```
danberindei: I think we also need `completeConflictResolution` to run if CR finishes with a `TimeoutException` (Diego's scenario) or basically anything other than a `SuspectException`.

I was thinking that we may need special handling for `InterruptedException`, too, but then I remembered `CompletableFuture.cancel()` doesn't do anything, so now I'm thinking that `restartConflictResolution()` may need another way to interrupt CR if necessary. TBH I'm not sure interrupting is necessary anyway, as a node leaving would also trigger a `SuspectException` and stop CR.

ryanemerson: "CompletableFuture.cancel()" still forces the CompletableFuture to complete with a `CancellationException`, which triggers https://github.com/ryanemerson/infinispan/blob/a2d29dff5e799e7215a0d45524f2baae24dc4886/core/src/main/java/org/infinispan/conflict/impl/DefaultConflictManager.java#L251 that then cancels any ongoing segment requests. Or am I missing something?

danberindei: `cancel()` only completes the CF instance you're calling it on. If you want to call the callback added by `resolveConflicts()` with a `CancellationException` you'd have to cancel the initial future, not the result of `future.whenComplete(...)`, but returning the initial future and adding another callback on that would again complicate things by not guaranteeing an execution order.

ryanemerson:
@danberindei Updated so that CR restarts if a non SuspectException Throwable is thrown during a segment request. During these changes it was necessary to change the approach of `ClusterCacheStatus`, so we now make use of the `ConflictResolution` class, instead of `ConflictData`, to control CR request and retries.

core/src/main/java/org/infinispan/conflict/impl/StateReceiverImpl.java
```diff
@@ -227,6 +234,15 @@ synchronized void cancel(Throwable throwable) {
          clear();
       }
 
+      synchronized void completeRequest() {
+         List<Map<Address, CacheEntry<K, V>>> retVal = keyReplicaMap.entrySet().stream()
+               .map(Map.Entry::getValue)
+               .collect(Collectors.toList());
```
danberindei: Could we get away with returning `keyReplicaMap.values()` here?

core/src/test/java/org/infinispan/conflict/impl/CrashedNodeDuringConflictResolutionTest.java
```diff
@@ -0,0 +1,201 @@
...
+
+      for (String key : ALL_KEYS) {
+         assertCacheGet(key, PARTITION_0_VAL, p0.getNodes());
+         assertCacheGet(key, PARTITION_1_VAL, p1.getNodes());
+      }
+
+      fork(() -> partition(0).merge(partition(1), false));
+
+      try {
+         future.get(60, TimeUnit.SECONDS);
```
danberindei: The point I was trying to make about refactoring `CompleteFutureOnStateRequestHandler` is that we should do as much as possible on the test thread, so we have better control over when the node is killed. So I had something more like this in mind:

      partition(0).merge(partition(1), false);
      blockedStateRequest.get(10, TimeUnit.SECONDS);

      if (trace) log.trace("crashCacheManager(2)");
      TestingUtil.crashCacheManagers(manager(2));

core/src/test/java/org/infinispan/conflict/impl/CrashedNodeDuringConflictResolutionTest.java
```diff
@@ -0,0 +1,201 @@
...
+               if (trace) log.trace("crashCacheManager(2)");
+               TestingUtil.crashCacheManagers(manager(2));
+            });
+
+      for (String key : ALL_KEYS) {
+         assertCacheGet(key, PARTITION_0_VAL, p0.getNodes());
+         assertCacheGet(key, PARTITION_1_VAL, p1.getNodes());
+      }
+
+      fork(() -> partition(0).merge(partition(1), false));
```
danberindei: I'd rather avoid the fork here, but I wanted to mention that it's always a good idea to wait for any forked thread you start before the test finishes.

core/src/test/java/org/infinispan/conflict/impl/CrashedNodeDuringConflictResolutionTest.java
```diff
@@ -0,0 +1,201 @@
...
+
+      // Once the JGroups view has been updated to remove manager(index), then the CR should be restarted when the
+      // coordinator continues to recover the cluster state
+      TestingUtil.waitForNoRebalance(cache(0), cache(1), cache(3));
+   }
+
+   @Override
+   protected void afterConflictResolutionAndMerge() {
+      ConflictManager cm = conflictManager(0);
+      assert !cm.isConflictResolutionInProgress();
```
danberindei: `assertFalse`

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -899,11 +904,156 @@ public boolean resolveConflictsOnMerge() {
...
+      }
+   }
+
+   private synchronized void completeConflictResolution() {
+      if (trace) log.tracef("Cache %s conflict resolution future complete", cacheName);
+      // CR is only queued for PreferConsistencyStrategy when a merge it is determined that the newAvailabilityMode will be AVAILABLE
+      // therefore if this method is called we know that the partition must be set to AVAILABLE
+      availabilityMode = AvailabilityMode.AVAILABLE;
+
+      // Install a NO_REBAlANCE topology with pendingCh == null to signal conflict resolution has finished
```
danberindei: Typo: l should be L

core/src/test/java/org/infinispan/conflict/impl/BaseMergePolicyTest.java
```diff
@@ -77,7 +83,7 @@ protected BaseMergePolicyTest(CacheMode cacheMode, String description, Availabil
 
 
    protected void beforeSplit() {
-      conflictKey = new MagicKey(cache(p0.node(0)), cache(p1.node(0)));
+      conflictKey = numberOfOwners > 1? new MagicKey(cache(p0.node(0)), cache(p1.node(0))) : new MagicKey(cache(p0.node(0)));
```
danberindei: Whitespace

danberindei:
I added a few minor comments, but unfortunately I'm also seeing a new query test failure:

https://ci.infinispan.org/job/Infinispan/job/PR-5902/14/testReport/junit/org.infinispan.query.blackbox/ClusteredCacheWithAffinityIndexManagerTest/testConditionalRemoveFromBackup/

core/src/test/java/org/infinispan/test/TestingUtil.java
```diff
@@ -354,15 +354,15 @@ public static void waitForNoRebalance(Cache... caches) {
             boolean chContainsAllMembers;
             boolean currentChIsBalanced;
             if (cacheTopology != null) {
-               rebalanceInProgress = cacheTopology.getPendingCH() != null;
+               rebalanceInProgress = cacheTopology.getPendingCH() != null || cacheTopology.getPhase() == CacheTopology.Phase.CONFLICT_RESOLUTION;
```
danberindei: I think we could use `cacheTopology.getPhase() != NO_REBALANCE`

ryanemerson: If it wasn't clear from the change, the change to `currentCH.locateOwnersForSegment(i).size() != actualNumOwners` was necessary as the CR+Rebalance topology update goes as follows:

1.CONFLICT_RESOLUTION
2.NO_REBALANCE
3.READ_OLD_WRITE_ALL

So some of the tests were failing because `waitForNoRebalance` was returning during stage 2.

danberindei: I actually missed that change, I was only focusing on the pending CH test, but looks ok to me.

I'm not sure why I used `<` there, but I don't see any reason to keep using it now.

danberindei:
I found quite a few other `ClusteredCacheWithAffinityIndexManagerTest/testConditionalRemoveFromBackup` failures in CI (`ssh ci -C "find /var/lib/jenkins/jobs/Infinispan/branches -name build.xml | xargs sudo grep 'ClusteredCacheWithAffinityIndexManagerTest'"`), and it almost always fails on my machine (https://issues.jboss.org/browse/ISPN-8160), so I think it's safe to ignore it.

danberindei:
Integrated, thanks Ryan!

PR #5897, opened 2018-04-05
ISPN-9000 Added email notify on failure

rigazilla:
This enables email notification when a build fails.

https://issues.jboss.org/browse/ISPN-9000

PR #5904, opened 2018-04-09
ISPN-8997 Deploy artifacts

slaskawi:
https://issues.jboss.org/browse/ISPN-8997

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -126,6 +126,11 @@ pipeline {
                         body: '${DEFAULT_CONTENT}'
                     }
                 }
+                if (!env.BRANCH_NAME.startsWith('PR-')) {
+                    configFileProvider([configFile(fileId: 'maven-settings-with-deploy-snapshot', variable: 'MAVEN_SETTINGS')]) {
+                        sh "$MAVEN_HOME/bin/mvn deploy:deploy -B -V -e -s $MAVEN_SETTINGS -Dmaven.test.failure.ignore=true -Dansi.strip -DskipTests=true"
+                    }
+                }
             }
         }
     }


slaskawi:
@tristantarrant Ready for review.

Jenkinsfile
```diff
@@ -70,6 +70,20 @@ pipeline {
                 sh 'find . -name "hs_err_*" -exec echo {} \\; -exec grep "^# " {} \\;'
             }
         }
+
+        stage('Deploy') {
+            when {
+                allOf {
+                    branch 'master'
+                    equals expected: 0, actual: currentBuild.rawBuild.getAction(AbstractTestResultAction.class).failCount
```
danberindei: I think this is way too specific, the build could fail before it tries to run any test.
Why not use a `failure` block like https://github.com/infinispan/infinispan/pull/5897/files?

slaskawi: I agree. @rigazilla did a pretty nice job by extracting `failure` and `success` scenarios in https://github.com/infinispan/infinispan/pull/5897. I guess I should put this there...

Jenkinsfile
```diff
@@ -125,6 +125,11 @@ pipeline {
                         body: '${DEFAULT_CONTENT}'
                     }
                 }
+                if (!env.BRANCH_NAME.startsWith('PR-')) {
+                    configFileProvider([configFile(fileId: 'maven-settings-with-deploy-snapshot', variable: 'MAVEN_SETTINGS')]) {
+                        sh "$MAVEN_HOME/bin/mvn deploy -B -V -e -s $MAVEN_SETTINGS -Dmaven.test.failure.ignore=true -Dansi.strip -DskipTests=true"
```
danberindei: I suggest using `mvn deploy:deploy` so that we don't compile or run the tests again.

rvansa:
@slaskawi Is this PR still valid?

danberindei:
@rvansa I was hoping to get a build without failures to actually test the deployment before integrating, but that's still pretty hard to obtain. I triggered one more build today, but I'll integrate it regardless of the result and wait for a green build on master to test it.

danberindei:
The last build had some test failures as well, so I integrated it and we'll check for problems in the master builds :)

PR #5905, opened 2018-04-09
ISPN-8962 Add one more merge scenario

danberindei:
The issue is already fixed, just adding one more test method.


ryanemerson:
Thanks @danberindei

PR #5907, opened 2018-04-10
ISPN-9045 Failure to change 'protocolVersion' in RemoteStoreConfigura…

rhusar:
…tion

Jira
https://issues.jboss.org/browse/ISPN-9045


tristantarrant:
Well, it does work if you use "PROTOCOL_VERSION_27", but it ain't pretty. Can you please also add a test to https://github.com/infinispan/infinispan/blob/f07383259a1168fccb1845c18dc018a73f2524c0/persistence/remote/src/test/java/org/infinispan/persistence/remote/configuration/ConfigurationTest.java#L10 ?
Ideally we should be able to use the same property names that the Hot Rod client ConfigurationBuilder supports... but that is for another time

tristantarrant:
Added a test and merged as 59d291ee79d31b2e35cbe41dae4487a73e23fdd9

rhusar:
Thanks @tristantarrant for handling this (I was on PTO yesterday).

PR #5909, opened 2018-04-10
ISPN-9008 RpcManager.invokeCommandOnAll ignores cache member missing from cluster view

danberindei:
https://issues.jboss.org/browse/ISPN-9008

I added an `ImmutableHopscotchHashSet` mainly because I didn't want to have both a `unmodifiableSet()` wrapper and `HashSet` wrapping `HashMap`. Considering addresses are random numbers I don't think hopscotch hashing per se can help with performance, but maybe we could do something to optimize `containsAll` @rvansa?

### DIFF
--- a/core/src/test/java/org/infinispan/replication/ForceSyncAsyncFlagsTest.java
+++ b/core/src/test/java/org/infinispan/replication/ForceSyncAsyncFlagsTest.java
@@ -6,8 +6,6 @@ import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 
-import java.util.concurrent.TimeUnit;
-
 import org.infinispan.AdvancedCache;
 import org.infinispan.commands.ReplicableCommand;
 import org.infinispan.configuration.cache.CacheMode;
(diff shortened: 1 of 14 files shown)


danberindei:
Rebased. I ended up using `ImmutableHopscotchHashSet` only in JGroupsTransport, and there's no usage of `containsAll` any more.

tristantarrant:
I believe https://ci.infinispan.org/job/Infinispan/job/PR-5909/4/testReport/junit/org.infinispan.jmx/RpcManagerMBeanTest/testSuccessRatio/ is related

danberindei:
Indeed, I updated the test to use `MagicKey` for more determinism.

danberindei:
There was one more test failure caused by a conflict with the ISPN-9062 fix, it's solved now.

rvansa:
@danberindei I don't see how you'd like to optimize `containsAll`... Somehow sorting the hashes and scanning ranges in the map?

core/src/test/java/org/infinispan/replication/ForceSyncAsyncFlagsTest.java
```diff
@@ -55,9 +52,7 @@ public void testForceAsyncFlagUsage() throws Exception {
 
       // check that the replication call was sync
       cache1.put("k", "v");
-      verify(mockTransport)
-            .invokeCommandOnAll(any(ReplicableCommand.class), any(ResponseCollector.class), any(DeliverOrder.class),
-                                anyLong(), any(TimeUnit.class));
+      verify(mockTransport).invokeCommandOnAll(any(), any(), any(), any(), anyLong(), any());
```
rvansa: why have you removed the classes?

danberindei: I just find it more readable when the method call fits into one line, and the test only cares about the name of the method being called (`invokeCommandOnAll` vs. `sendToAll`).

rvansa:
LGTM, integrated.

PR #5910, opened 2018-04-10
ISPN-9057 JGroupsTransport.invokeCommand should return CompletionStage

danberindei:
https://issues.jboss.org/browse/ISPN-9057


tristantarrant:
Merged

PR #5914, opened 2018-04-11
ISPN-9062 JGroupsTransport should only send messages to nodes in the …

danberindei:
…cluster view

https://issues.jboss.org/browse/ISPN-9062

* Only check the view if Request.onNewView removed a target
* Move Request to the impl package, it's too low-level to expose
  in the Transport/RpcManager API


danberindei:
@wburns I added an assert in the existing `JGroupsTransportTest`, I don't think it's worth doing more.

wburns:

PR #5915, opened 2018-04-11
Fix broken link to JavaDocs

tristantarrant:


danberindei:
@tristantarrant the link works now, but links from the javadoc to `ConfigurationStorage` and `LocalConfigurationStorage` don't work because the `o.i.globalstate` package is not `@public`.

PR #5917, opened 2018-04-12
ISPN-9069 REST store should depend on the REST server only for tests

tristantarrant:
https://issues.jboss.org/browse/ISPN-9069

This shrinks the all distribution from 92MB to 78MB


gustavonalle:
Integrated, thanks @tristantarrant !

PR #5929, opened 2018-04-19
ISPN-9091 Determine available CPUs taking CGroup quotas into account

tristantarrant:
https://issues.jboss.org/browse/ISPN-9091

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/ProcessorInfo.java
+++ b/commons/src/main/java/org/infinispan/commons/util/ProcessorInfo.java
@@ -1,0 +1,144 @@
+package org.infinispan.commons.util;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Locale;
+import java.util.stream.Stream;
+
+/**
+ * Provides general information about the processors on this host.
+ *
+ * @author Jason T. Greene
+ * @author Tristan Tarrant
+ */
+public class ProcessorInfo {
+   private static final byte[] BITS = new byte[]{0, 1, 1, 2, 1, 2, 2, 3, 1, 2, 2, 3, 2, 3, 3, 4};
+   private static final String CPUS_ALLOWED = "Cpus_allowed:";
+   private static final String CGROUP_STATUS_FILE = "/proc/cgroups";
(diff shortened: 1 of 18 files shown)


tristantarrant:
This now uses the same logic adopted by Java 10+ taking into account:
* a system property (infinispan.activeprocessorcount)
* available CPUs as set by taskset
* CGroup CPU shares
* CGroup quota/period allocation

tristantarrant:
Hints for testing:
Run a docker container with the some quota/period configuration and each of the JDKs:
e.g.
`docker run -it --cpu-quota=120000 --cpu-period=10000 [openjdk:8-jdk|openjdk:10-jdk|openjdk:11-jdk]`
copy infinispan-commons.jar to the container:
`docker cp commons/target/infinispan-commons-9.3.0-SNAPSHOT.jar inspiring_mestorf:/infinispan-commons.jar`
From inside the container, run the following:
`java -classpath infinispan-commons.jar org.infinispan.commons.util.ProcessorInfo`
Try different combinations of cpu-quota/cpu-period

ryanemerson:
@tristantarrant It appears to be working as expected :+1:. On Jdk >= 10 should we not just return `Runtime.getRuntime().availableProcessors();` in `ProcessorInfo::determineProcessors()`?

tristantarrant:
Yes, but the multi-release jar work  is going into my JDK 10+ PR. Once this is in I will adapt https://github.com/infinispan/infinispan/pull/5940 to do just that

commons/src/main/java/org/infinispan/commons/util/ProcessorInfo.java
```diff
@@ -0,0 +1,144 @@
...
+      if (System.getSecurityManager() != null) {
+         return AccessController.doPrivileged((PrivilegedAction<Integer>) () -> Integer.valueOf(determineProcessors())).intValue();
+      }
+
+      return determineProcessors();
+   }
+
+   private static int determineProcessors() {
+      int javaProcs = Runtime.getRuntime().availableProcessors();
+      int userProcs = Integer.getInteger("infinispan.activeprocessorcount", javaProcs);
```
wburns: Shouldn't we always use the value returned from the setting "infinispan.activeprocessorcount" ?

tristantarrant: it does so only if it is less than the available processors.

wburns: Okay, from the name I assumed it would be use this setting always type of thing. Just side note but we non linux would always use this value if it is set btw.

PR #5932, opened 2018-04-20
ISPN-9093 Samples from query docs should not be affected by branding

anistor:
* replace {brandname}IndexManager back to InfinispanIndexManager in query docs
* avoid using {brandname} in query samples

https://issues.jboss.org/browse/ISPN-9093


anistor:
updated!

PR #5935, opened 2018-04-23
ISPN-9087 Timeout during put operation when a node is blocked

danberindei:
https://issues.jboss.org/browse/ISPN-9087

* Replace FD_ALL with FD_ALL2 and change the timeout from 60s to 10s.
* Keep the default cache replication timeout at 15s
* Increase the default backup site timeout from 10s to 15s

### DIFF
--- a/cloud/src/main/resources/default-configs/default-jgroups-ec2.xml
+++ b/cloud/src/main/resources/default-configs/default-jgroups-ec2.xml
@@ -24,11 +24,12 @@
            max_interval="30000" 
    />
    <FD_SOCK />
-   <FD_ALL timeout="60000" 
-           interval="15000" 
-           timeout_check_interval="5000" 
+   <!-- Suspect node `timeout` to `timeout + timeout_check_interval` millis after the last heartbeat -->
+   <FD_ALL timeout="10000"
+           interval="2000"
+           timeout_check_interval="1000"
    />
-   <VERIFY_SUSPECT timeout="5000" />
+   <VERIFY_SUSPECT timeout="1000"/>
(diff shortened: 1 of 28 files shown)


rvansa:
I find these values quite short; with CMS and recommended heap settings (32 GB) I recall seeing 10 second STW pauses.

Also, if you have some constraints in mind (e.g. FD always applying before replication timeout) please add them into configuration validation, with a WARN notice to the users.

danberindei:
@rvansa these are just defaults. Users are free to change them if they have bigger GC pauses, but at least if they changed the FD timeout they won't be surprised when regular write operations also time out.

I would like to enforce this in the configuration validation, but:
1. We don't always have access to the JGroups configuration, e.g. in the server the channel is created by a different subsystem and we only get a `ForkChannel`
1. It's not always obvious what's the maximum time it could take for a node to be removed from the view after becomes unresponsive. E.g. even after changing the default I'm still not sure if `FD_ALL2` always removes the node in `timeout` millis or if it can sometimes take `2 * timeout` millis...

danberindei:
Bela confirmed that `FD_ALL2` may suspect a node only `2 * timeout` millis after the last heartbeat, so I went back to `FD_ALL` and I added a comment in the config files.

ryanemerson:
@danberindei The changes LGTM, but the commit message in https://github.com/infinispan/infinispan/pull/5935/commits/0ebceb8daa82302db49ae9db51710aabb295e84b needs updating to reflect the latest changes.

danberindei:
Updated @ryanemerson

PR #5937, opened 2018-04-24
ISPN-9094 ArrayIndexOutOfBoundsException on server using scattered cache

rvansa:
* Replaced disfunctional CASes with read-write locking

https://issues.jboss.org/browse/ISPN-9094

### DIFF
--- a/core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
+++ b/core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
@@ -16,7 +16,9 @@ import java.util.concurrent.atomic.AtomicIntegerArray;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicLongArray;
 import java.util.concurrent.atomic.AtomicReferenceArray;
-import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.infinispan.commands.CommandsFactory;
 import org.infinispan.commands.write.InvalidateVersionsCommand;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/infinispan/scattered/ControlledScatteredVersionManager.java
```diff
@@ -28,8 +28,14 @@ public static void flush(boolean expectRemoval, ControlledScatteredVersionManage
       for (;;) {
          boolean done = true;
          for (int i = 0; i < svms.length; ++i) {
-            if (regularCounters[i] >= svms[i].regularCounter.get()) done = false;
-            if (expectRemoval && removeCounters[i] >= svms[i].removeCounter.get()) done = false;
+            if (regularCounters[i] >= svms[i].regularCounter.get()) {
```
wburns: Maybe we should put `svms[i].regularCounter.get()` into a field variable? Same with `removeCounter` below? Or are these cases not standard?

core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
```diff
@@ -462,36 +462,39 @@ protected void regularInvalidationFinished(Object[] keys, int[] topologyIds, lon
...
+      final ConcurrentMap<K, InvalidationInfo> removedKeys;
+      Lock writeLock = removedKeysLock.writeLock();
+      writeLock.lock();
+      try {
+         removedKeys = this.removedKeys;
+         this.removedKeys = new ConcurrentHashMap<>(invalidationBatchSize);
+      } finally {
+         writeLock.unlock();
       }
+      executorService.execute(() -> {
```
wburns: Maybe optionally invoke this only if `removedKeys.size() > 0` and then this method can return if it is removing invalidations? Is it possible to check `isEmpty` in `startFlush` and see it is empty and then get an insert afterwards?

rvansa: Do you mean moving the checks from lines 203 and 469 here? Yes, I can do that..

core/src/test/java/org/infinispan/scattered/ControlledScatteredVersionManager.java
rvansa: OK, though this is just a test class...

wburns: Sorry, overlooked this was in src/test, you can ignore.

core/src/main/java/org/infinispan/scattered/impl/ScatteredVersionManagerImpl.java
wburns: Yeah this way you can't add to the ConcurrentMap in between since they are swapped. This could prevent possible missed `removedKeys`, but I am not sure if that was possible before.

rvansa: @wburns Not sure if I follow: if you add to `removedKeys` after the `.isEmpty` check all that happens is that the invalidations are sent a bit later. Since you swap with a write-lock and add with a read-lock, the situation where you'd add a key to removedKeys and this would be dropped without sending is not possible, IMO.

wburns: Yeah looking closer the method I was more worried about was `startFlush` which I didn't realize was only for testing. And as you mentioned it doesn't replace the map either, so it should be fine.

wburns:

PR #5947, opened 2018-04-29
ISPN-9111 Add any enabled backup sites to protobuf metadata cache

galderz:
https://issues.jboss.org/browse/ISPN-9111

* So that metadata cache information can be replicated accross sites.

### DIFF
--- a/remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerImpl.java
+++ b/remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerImpl.java
@@ -8,13 +8,19 @@ import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Stream;
 
 import javax.management.MBeanException;
 import javax.management.ObjectName;
 
 import org.infinispan.Cache;
 import org.infinispan.commons.CacheException;
+import org.infinispan.commons.api.BasicCacheContainer;
 import org.infinispan.commons.dataconversion.IdentityEncoder;
+import org.infinispan.commons.logging.LogFactory;
+import org.infinispan.configuration.cache.BackupConfiguration;
+import org.infinispan.configuration.cache.BackupFailurePolicy;
 import org.infinispan.configuration.cache.CacheMode;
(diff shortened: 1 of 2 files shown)


galderz:
Is state transfer for xsite still manual? If you have protobuf metadata cache replicated across sites, would you still need to call push state on xsite state admin to push that data to other sites when they come up? For such specific internal cache, it'd be good for this to be done internally :|

tristantarrant:
Ideally, internal caches state transfer should be prioritised over user
caches so that we don't fall in a situation where protobuf payloads are
replicated before their schemas.

galderz:
@tristantarrant Indeed! Due to timing issues like that one that I feel x-site state transfer for internal caches should be automated. For the time being I'm checking with Clement et al to see if manual state transfer is viable, but I suspect we could still encounter timing issues.

I'll look into the feasibility of automated x-site state transfer for protobuf metadata cache tomorrow.

slaskawi:
TBH, I really like the first approach @tristantarrant proposed. Internal caches configuration should be additive, so that we could add x-site replication or some other feature.

galderz:
> TBH, I really like the first approach @tristantarrant proposed. Internal caches configuration should be additive, so that we could add x-site replication or some other feature.

I agree. It does require a lot more thought and it has considerable complications: 

* I don't know what's the state of our configuration, but in the past it used to be hard to separate between those configuration options that had been defined in the XML versus the default values.
* You need to determine which combination of additive options and internal default options are incompatible, and/or which ones can/cannot be overridden.

Once that's in place, I'd happy for the code in this PR to be replaced.

tristantarrant:
All *Configuration objects wrap Attributes which can tell you if they've been modified:
https://github.com/infinispan/infinispan/blob/master/commons/src/main/java/org/infinispan/commons/configuration/attributes/Attribute.java#L76

danberindei:
I'm not sure we can make this completely automatic, I believe the user can always create a new cache that has a completely new backup site. Not sure how adding a new site would work at the JGroups level, but that's nothing stopping it at the Infinispan level.

I'd rather move the list of backup sites to the global configuration as backup configuration templates, so individual cache configurations could only pick one or more of those global sites (but maybe with a different config). Then the protobuf metadata cache would just back up to all the globally defined backup sites, with the template config. And I think having a single "offline" status for the backup site across all caches would also be an improvement.

PR #5950, opened 2018-05-02
ISPN-8982 Remove ORM 5.2 support

rvansa:
I've forgotten to commit removal of 5.2 module.


danberindei:
@rvansa For the ORM neophytes amongst us, could you explain here or in the JIRA why you're removing 5.2 and keeping 5.1?

rvansa:
@danberindei ORM 5.1 is used in some versions of WF so we want to keep that updates with bugfixes etc. 5.2 is not, so we're saving resources by dropping that completely.
Regrettably we can't drop 5.2 support in 9.2 micro, so I had to have an impl for 9.2.2.

PR #5954, opened 2018-05-03
ISPN-9059 Use functional commands for 2LC

rvansa:
https://issues.jboss.org/browse/ISPN-9059

Replacing the nasty overrides of `CallInterceptor` with functional commands.

Depends on #5950


hibernate/cache-commons/src/main/java/org/infinispan/hibernate/cache/commons/access/NonStrictAccessDelegate.java
```diff
@@ -35,17 +39,18 @@
 
 	private final InfinispanDataRegion region;
 	private final AdvancedCache cache;
-	private final AdvancedCache writeCache;
-	private final AdvancedCache putFromLoadCache;
+	private final FunctionalMap.ReadWriteMap<Object, Object> writeMap;
```
galderz: ~Shouldn't this be WriteOnlyMap? We're only writing, we don't care about previous value.~

Ah, these need to be RW so that the injectable component can check previous value? Does it apply to all uses?

hibernate/cache-commons/src/main/java/org/infinispan/hibernate/cache/commons/util/VersionedEntry.java
```diff
@@ -56,27 +71,105 @@ public String toString() {
...
+			oldVersion = findVersion(oldValue);
 		}
 
-		@Override
-		public Object convert(Object key, Object value, Metadata metadata) {
-			if (value instanceof VersionedEntry) {
-				return ((VersionedEntry) value).getValue();
+		if (oldVersion == null) {
+			assert oldValue == null || oldTimestamp != Long.MIN_VALUE : oldValue;
+			if (timestamp <= oldTimestamp) {
```
galderz: ~I'm not sure this is a good use of asserts. You're doing an `if` call which only matters if the asserts are enabled. I'd favour asserts not to be used and used normal exception reporting for consistency with the rest of the codebase.~

Actually, I see the if matters from a normal logic perspective. We hardly use `assert` throughout our code base. It worries me to start using them without some clear guidelines on when to use it, as well as some documentation for our users to enable them?

hibernate/cache-commons/src/main/java/org/infinispan/hibernate/cache/commons/access/NonStrictAccessDelegate.java
rvansa: Yes, we always merge values, because order in which the commands are applied is not guaranteed. That's why we need the version from Hibernate.

hibernate/cache-commons/src/main/java/org/infinispan/hibernate/cache/commons/util/VersionedEntry.java
rvansa: Users are not supposed to enable them, the asserts are intended to check for unexpected states in the testsuite.

galderz:
All good @rvansa. Rebase the PR and I'll integrate it

rvansa:
@galderz Rebased.

galderz:
I've just squashed PR for ISPN-9194, apologies for the delay. Once that's in, I'll take one of the CI machines and run the testsuite continuously for a day to make sure we don't have any more random failures. Once that's good we'll integrate this PR.

PR #5957, opened 2018-05-03
ISPN-9002 LocalModeNoPassivationTest.testValuesWithEvictedEntries random

wburns:
failures

* Fixed issue with retrieval during close

https://issues.jboss.org/browse/ISPN-9002

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/util/PersistenceManagerCloseableSupplier.java
+++ b/core/src/main/java/org/infinispan/persistence/util/PersistenceManagerCloseableSupplier.java
@@ -1,5 +1,6 @@
 package org.infinispan.persistence.util;
 
+import java.lang.invoke.MethodHandles;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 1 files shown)


wburns:
This could occur during the following window

1. iterator thread finds nothing in queue and was about to acquire lock
2. store thread adds entry to queue and then acquires lock and closes
3. iterator thread retrieves lock and then sees that it was closed returning null.

core/src/main/java/org/infinispan/persistence/util/PersistenceManagerCloseableSupplier.java
```diff
@@ -126,7 +132,7 @@ public void processEntry(MarshalledEntry<K, V> marshalledEntry, AdvancedCacheLoa
          closeLock.lock();
          try {
             // If is possible that someone inserted a value and then acquired the close lock - thus we must recheck
-            if (closed || (entry = queue.poll()) != null) {
+            if ((entry = queue.poll()) != null || closed) {
```
danberindei: I stared for 5 minutes at this code yesterday without realizing what switching the evaluation order would do...
I saw it when I looked again today, but may I suggest not mixing complex `if` conditions and assignments? :)

PR #5961, opened 2018-05-04
ISPN-9111 x-site configuration for protobuf metadata cache

galderz:
This is just a prototype. If we're happy with the approach a different interceptor should be created and injected into protobuf metadata cache.

Ideally we'd use SYNC x-site configuration but that's not fully working as expected. The reason for that is that you'd not want protobuf metadata updates to be applied partially cluster wide. However, SYNC with FAIL backup policy has a number of issues. So, SYNC with WARN is the best we can do maybe now.

On top of that, lazy get functionality has been implemented so that if a site protobuf lookup fails to find anything, this lookup can be pushed to other sites. This is needed because there's no state transfer in place for x-site data. Alternatively we could have tried to push state when a new node joins the x-site cluster, but at the point when this is known, the channel is not set up yet.

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/xsite/PessimisticBackupInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/xsite/PessimisticBackupInterceptor.java
@@ -1,7 +1,16 @@
 package org.infinispan.interceptors.xsite;
 
+import org.infinispan.commands.VisitableCommand;
+import org.infinispan.commands.read.GetKeyValueCommand;
 import org.infinispan.commands.tx.CommitCommand;
+import org.infinispan.container.DataContainer;
+import org.infinispan.context.InvocationContext;
 import org.infinispan.context.impl.TxInvocationContext;
+import org.infinispan.factories.ComponentRegistry;
+import org.infinispan.factories.annotations.Inject;
+import org.infinispan.interceptors.InvocationSuccessFunction;
+import org.infinispan.metadata.Metadata;
+import org.infinispan.remoting.transport.BackupResponse;
 
 /**
  * Handles x-site data backups for pessimistic transactional caches.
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/infinispan/interceptors/xsite/PessimisticBackupInterceptor.java
```diff
@@ -11,10 +20,55 @@
...
+   private Object handleLocalGetReturn(InvocationContext ctx, VisitableCommand rCommand, Object rv) throws Throwable {
+      if (rv == null) {
+         final GetKeyValueCommand getCmd = (GetKeyValueCommand) rCommand;
+         BackupResponse backupResponse =
+            backupSender.backupGet(getCmd);
+
+         backupSender.processResponses(backupResponse, rCommand);
+         final Object value = getValue(backupResponse);
+
+         if (value != null)
```
rvansa: You should also wrap the result in context: `entryFactory.wrapExternalEntry`

core/src/main/java/org/infinispan/interceptors/xsite/PessimisticBackupInterceptor.java
```diff
@@ -11,10 +20,55 @@
...
+      }
+   }
+
+   private Object handleLocalGetReturn(InvocationContext ctx, VisitableCommand rCommand, Object rv) throws Throwable {
+      if (rv == null) {
+         final GetKeyValueCommand getCmd = (GetKeyValueCommand) rCommand;
+         BackupResponse backupResponse =
+            backupSender.backupGet(getCmd);
+
+         backupSender.processResponses(backupResponse, rCommand);
```
rvansa: this looks like a sync invocation in our async stack :-/

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsBackupResponse.java
```diff
@@ -51,6 +53,7 @@ public JGroupsBackupResponse(Map<XSiteBackup, Future<Response>> syncBackupCalls,
    public void waitForBackupToFinish() throws Exception {
       long deductFromTimeout = timeService.timeDuration(sendTimeNanos, MILLISECONDS);
       errors = new HashMap<>(syncBackupCalls.size());
+      values = new HashMap<>(syncBackupCalls.size());
```
rvansa: I wonder if it's necessary to allocate these maps even if there are no errors/values.

remote-query/remote-query-server/src/test/java/org/infinispan/query/remote/impl/xsite/ProtobufMetadataXSiteStateTransferTest.java
```diff
@@ -0,0 +1,82 @@
...
+      return getDefaultClusteredCacheConfig(cacheMode);
+   }
+
+   @Override
+   protected ConfigurationBuilder getLonActiveConfig() {
+      return getNycActiveConfig();
+   }
+
+   public void testMetadataStateTransfer() {
+//      stopNYC();
```
rvansa: remove comments

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerImpl.java
```diff
@@ -120,6 +127,31 @@ private ConfigurationBuilder getProtobufMetadataCacheConfig() {
...
+
+      return cfg;
+   }
+
+   private ConfigurationBuilder addMetadataSiteBackups(
+         ConfigurationBuilder cfg, EmbeddedCacheManager cacheManager) {
+      final List<String> allCacheNames = new ArrayList<>(cacheManager.getCacheConfigurationNames());
+      allCacheNames.add(BasicCacheContainer.DEFAULT_CACHE_NAME);
+
+      allCacheNames.stream()
```
rvansa: What happens when the remote site does not have protobuf cache configured? (is this possible?)

galderz:
Closing, this was just a mere prototype with many holes. A wider rethink is needed for this and related JIRAs.

PR #5964, opened 2018-05-07
ISPN-9075 Leverage CacheTransactionSynchronization

rvansa:
https://issues.jboss.org/browse/ISPN-9075

Depends on #5954


galderz:
Needs rebase

rvansa:
@galderz Done.

galderz:
This is already included in #6010. Shall we just focus on that PR?

hibernate/cache-commons/src/main/java/org/infinispan/hibernate/cache/commons/access/LockingInterceptor.java
```diff
@@ -42,6 +46,17 @@
...
+         // See CompletableFunction javadoc for explanation
+         if (rCommand instanceof ReadWriteKeyCommand) {
+            Function function = ((ReadWriteKeyCommand) rCommand).getFunction();
+            if (function instanceof CompletableFunction) {
+               ((CompletableFunction) function).markComplete();
+            }
+         }
+         // Similar to CompletableFunction above, signals that the command has been applied for non-functional commands
+         FlagAffectedCommand flagCmd = (FlagAffectedCommand) rCommand;
+         flagCmd.setFlagsBitSet(flagCmd.getFlagsBitSet() & ~FlagBitSets.FORCE_WRITE_LOCK);
```
pruivo: shouldn't the flags to be updated even if the return value isn't a `CompletableFuture`?

hibernate/cache-v53/src/main/java/org/infinispan/hibernate/cache/v53/impl/NonTxInvalidationCacheAccessDelegate.java
```diff
@@ -0,0 +1,50 @@
...
+
+   @Override
+   protected void invoke(Object session, InvocationContext ctx, RemoveCommand command) {
+      CompletableFuture<Object> future = invoker.invokeAsync(ctx, command);
+      // LockingInterceptor removes this flag when the command is applied and we only wait for RPC
+      if (!command.hasAnyFlag(FlagBitSets.FORCE_WRITE_LOCK)) {
+         Sync sync = (Sync)((SharedSessionContractImplementor) session).getCacheTransactionSynchronization();
+         sync.registerBeforeCommit(future);
+      } else {
+         log.trace("Removal was not applied immediatelly, waiting.");
```
pruivo: typo: `immediately`

hibernate/cache-v53/src/main/java/org/infinispan/hibernate/cache/v53/impl/NonTxInvalidationCacheAccessDelegate.java
```diff
@@ -0,0 +1,50 @@
...
+         log.tracef("Registering synchronization on transaction in %s, cache %s: %s", command.getKeyLockOwner(), cache.getName(), command.getKey());
+      }
+      sync.registerAfterCommit(new InvalidationInvocation(nonTxPutFromLoadInterceptor, command.getKey(), command.getKeyLockOwner()));
+   }
+
+   @Override
+   protected void invoke(Object session, InvocationContext ctx, RemoveCommand command) {
+      CompletableFuture<Object> future = invoker.invokeAsync(ctx, command);
+      // LockingInterceptor removes this flag when the command is applied and we only wait for RPC
+      if (!command.hasAnyFlag(FlagBitSets.FORCE_WRITE_LOCK)) {
```
pruivo: I'm not sure if this works. The flag is removed when the command is completed and, if I understand completely, it only happens after the rpc is completed.

pruivo: similar comment for `TombstoneAccessDelegate`

hibernate/cache-v53/src/main/java/org/infinispan/hibernate/cache/v53/impl/NonTxInvalidationCacheAccessDelegate.java
pruivo: similar comment for `TombstoneAccessDelegate`

pruivo:
@rvansa needs rebase again :(

galderz:
@pruivo Did you see my comment above? We've not split our comments between two PRs :\

pruivo:
@galderz I miss that particular comment. :) 
you could have close this PR to avoid this mistake. I'll move my comments to the other PR later :+1:

rvansa:
@galderz No, these are separate changes, but based on each other. You should have commented just on changes in the commits related to the PR - let's get this in first.

hibernate/cache-v53/src/main/java/org/infinispan/hibernate/cache/v53/impl/FutureUpdateInvocation.java
```diff
@@ -0,0 +1,74 @@
...
+      this.region = region;
+      this.sessionTimestamp = sessionTimestamp;
+   }
+
+   public UUID getUuid() {
+      return uuid;
+   }
+
+   @Override
+   public CompletableFuture<Void> invoke(boolean success) {
```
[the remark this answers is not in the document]
rvansa: 1) I think it's safe to use `success = false` when we've encountered problems - we're wipe out the entry and let it be loaded from DB.
2) The CF consumption looks really incorrect. I really miss some `thenCompose` variant with exceptional handling :-/

hibernate/cache-v53/src/main/java/org/infinispan/hibernate/cache/v53/impl/NonTxInvalidationCacheAccessDelegate.java
rvansa: The flag is removed when the command is applied (to DC) and the key is unlocked. At that point the command is not completed yet, the completion is waiting for the RPC.

This is an optimization based on the assumption that the command does not get blocked by locking and gets applied to DC in this thread, so when we get the future, it's already written to DC. In this case the `FWL` is unset and we don't have to block, as here we don't need to wait until the RPC completes: all we care about is that it is locally applied. We'll have to eventually wait until the RPC completes (int the synchronization), but we can proceed with other commands.

If we happen to be unlucky and have to wait for the lock, we get the future before the command is applied. We can't wait specifically for the local application, so we wait for the RPC too (and this gets completed only after local application).

hibernate/cache-v53/src/main/java/org/infinispan/hibernate/cache/v53/impl/FutureUpdateInvocation.java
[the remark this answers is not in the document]
danberindei: 2. use `InvocationStage` instead :)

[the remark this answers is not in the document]
rvansa: @danberindei Hmm, are we trying to push `IS` as the ultimate `CF` replacement for all internal workings?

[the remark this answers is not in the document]
danberindei: Of course, here you don't have a context and a command, so `InvocationStage` would do more bad than good. But I just couldn't help myself :)

hibernate/cache-v53/src/main/java/org/infinispan/hibernate/cache/v53/impl/NonTxInvalidationCacheAccessDelegate.java
danberindei: I didn't really follow the discussion, but this sounds like something that would be useful in a comment :)

pruivo: what cache mode is used? INVALIDATION?
I think the optimized path isn't hit in repl or dist mode (except if you are the primary owner)

in any case, I would prefer a better future proof solution. If we introduce the flag removal in another interceptor, you can start observing some weird behaviour. 

but it is up to you :+1: (I guess you could convert the regular operation to function operation?)

rvansa: The lock is acquired on all owners; so in this case we acquire the lock on originator and then the command is sent to all other owners, the order in which we apply invalidations doesn't matter so we don't need to go through primary owner. Locking interceptor expect the CF from NonTxII and waits for that after releasing the lock.

I agree that this is not future proof... I could consider using functional remove as the solution there is still hackish but less prone to be broken. Any 'proper' (design-wise) solution would increase size of the command for the general execution (-1) or introduce new flags that will reflect a 'status' of the command.

galderz: @rvansa I think it'd be good to have a design or how-it-works document in the 2LC module that explains all this in detail. The 2LC module works quite differently to the rest of the codebase, so it's common to see these kind of questions. In essence, I think we should cleanup/update explanations [here](https://github.com/infinispan/infinispan-designs/blob/master/Infinispan-Hibernate-Second-Level-Cache-improvements.md) and add it to the module itself as part of the readme or similar doc. I don't think this should be in the user guide but it's useful for contributors. Now that the 2LC modules are part of the Infinispan codebase, there are more eyes on this :)

rvansa:
Extending CF from FutureUpdateInvocation was rather too clever - I did not realize that there are some instanceof checks in `Sync` and this got it broken. I've added assert to prevent that issue in the future.

I've also noticed that FUS and FUI used `UUID.randomUUID()` call. Fixed.

About general design docs: I agree. Do I need to add it as part of this PR?

pruivo:
@rvansa if you add the design docs to https://github.com/infinispan/infinispan-designs, it isn't in the scope of this PR
@galderz overall, looks good to me. It your call! :+1:

rvansa:
https://github.com/infinispan/infinispan-designs/blob/master/Second-Level-Cache.md

galderz:
Thx @rvansa for updates, having a last look at the PR.

danberindei:
@rvansa Sending an email to infinispan-dev quoting lines from github sounds weird, but maybe it's weird enough to work...

The new section is not exactly what I had in mind, the functional commands are still mentioned only in passing. I would like a bullet list of entry types + commands used + functions used with functional commands, each with a short description, some examples of how an update would work, and references to the previous implementation only as footnotes.

Think about the reader as someone who doesn't know/care how 2LC used to work in the past, because that's almost certainly how the next maintainer of 2LC will approach it :)

danberindei:
@rvansa Yeah, maybe a package javadoc would be a better place for what I had in mind. But even for a design doc, IMO it reads too much like "you already know how 2LC works, here's the story behind that".

galderz:
@danberindei At a high level, how it works is explained in the [user guide](http://infinispan.org/docs/stable/user_guide/user_guide.html#jpa_hibernate_2l_cache). @rvansa's doc is trying to explain how things work underneath. We should fill any holes or missing info that you or others find in either of those though.

danberindei:
@galderz the user guide doesn't mention functional commands either...

danberindei:
@rvansa I'm reading "The regular ordering is not necessary because we have only few types of updates:" and I have no idea what's the regular ordering or how those updates are being applied.

rvansa:
@danberindei Two paragraphs above (second paragraph of *Cache modes* introduction) I talk about ordering in vanilla Infinispan. Is the *regular* word confusing you? (I could use *normal*, *classic*, or even *primary-owner-induced*).

danberindei:
@rvansa I'm reading ""The regular ordering is not necessary" but also "2LC does not change anything about how functional commands work". Which is true?

rvansa:
@danberindei Oookay, the first one :) The distribution change applies to all writes so "2LC does not change anything about how functional commands *specifically* work". But I admit you got me here :)

PR #5965, opened 2018-05-08
 ISPN-7977 NullPointerException during shutdown

danberindei:
https://issues.jboss.org/browse/ISPN-7977
https://issues.jboss.org/browse/ISPN-9060
https://issues.jboss.org/browse/ISPN-8782
https://issues.jboss.org/browse/ISPN-9067
https://issues.jboss.org/browse/ISPN-9127

Please don't close ISPN-9127, this commit is specific to `StateProviderImpl` but I hope to write a generic solution soon.


core/src/main/java/org/infinispan/topology/LocalTopologyManagerImpl.java
```diff
@@ -106,15 +106,18 @@ public void start() {
       latestStatusResponseViewId = transport.getViewId();
    }
 
-   // Need to stop before the JGroupsTransport
-   @Stop(priority = 9)
+   // Need to stop after ClusterTopologyManagerImpl and before the JGroupsTransport
+   @Stop(priority = 110)
```
tristantarrant: We should really have proper dependency graphs to avoid these "conjured" numbers

danberindei:
Updated to add more null checks in `GlobalMarshaller`

PR #5972, opened 2018-05-10
ISPN-9108 Hot Rod Lightweight Transaction

pruivo:
* Synchronization mode in the client

https://issues.jboss.org/browse/ISPN-9108


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/TransactionalRemoteCache.java
```diff
@@ -0,0 +1,23 @@
...
+ * TransactionalRemoteCache#size()} aren't transactional in a way they don't interact with the transaction's data (keys,
+ * values).
+ * <p>
+ * {@link TransactionalRemoteCache#containsValue(Object)} is a special case where a key with the specific value is
+ * marked as read fpr the transaction.
+ *
+ * @author Pedro Ruivo
+ * @since 9.3
+ */
+public interface TransactionalRemoteCache<K, V> extends RemoteCache<K, V> {
```
tristantarrant: I don't like the asymmetry this creates with embedded caches, where getTransactionManager() is a member of AdvancedCache.
Why not create a common TransactionalCache interface with the method and have both AdvancedCache and TransactionalRemoteCache implement it.
Additionally, maybe we should take the opportunity to return introduce an AsyncTransactionManager where commit and rollback return CompletableFutures

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
+:tx: link:{javaeedocroot}/javax/transaction/Transaction.html[Transaction]
```
tristantarrant: I didn't know about aliases. Nice.

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+:tx-mode: link:{javadocroot}/org/infinispan/client/hotrod/configuration/TransactionMode.html[TransactionMode]
+:cb: link:{javadocroot}/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.html[ConfigurationBuilder]
+:gtml: link:{javadocroot}/org/infinispan/client/hotrod/transaction/lookup/GenericTransactionManagerLookup.html[GenericTransactionManagerLookup]
+:rtml: link:{javadocroot}/org/infinispan/client/hotrod/transaction/lookup/RemoteTransactionManagerLookup.html[RemoteTransactionManagerLookup]
+:rtm: link:{javadocroot}/org/infinispan/client/hotrod/transaction/manager/RemoteTransactionManager.html[RemoteTransactionManager]
+:trc: link:{javadocroot}/org/infinispan/client/hotrod/TransactionalRemoteCache.html[TransactionalRemoteCache]
+
+==== Hot Rod Transaction
+
+The following article provide details about how to configure the Hot Rod client to be a participant in a
```
tristantarrant: This section provides details about how to configure and use the Hot Rod client to be a participant in a JTA {tx}.

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+In order to participate in a transaction, the client needs to know the {tm} to interact with and
+how it is going to participate in the {tx} (as {sync} or {xa}).
+
+INFO: The transactions are optimistic in a way that the write locks are acquired at prepare time.
+
+===== Configuration
+
+When creating the {rcm}, it is possible to set the default {tm} and {tx-mode} to use by the {rc}.
+However, it is possible to set a different {tm} and {tx-mode} for each cache.
+This will be shown later in this article.
```
tristantarrant: s/article/section/

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+The code snippet below show only the configuration available for transactional caches:
```
tristantarrant: s/show/shows/

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+org.infinispan.client.hotrod.configuration.ConfigurationBuilder cb = new org.infinispan.client.hotrod.configuration.ConfigurationBuilder();
+//other client configuration parameters
+cb.transaction().transactionManagerLookup(GenericTransactionManagerLookup.getInstance());
+cb.transaction().transactionMode(TransactionMode.NON_XA);
+RemoteCacheManager rmc = new RemoteCacheManager(cb.build());
+----
+
+For a complete reference to the available configuration option please refer to the {cb}'s javadoc.
+
+It is also possible to configure the Java Hot Rod client using an properties file, e.g.:
```
tristantarrant: s/an/a/

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+[source,java]
+----
+org.infinispan.client.hotrod.configuration.ConfigurationBuilder cb = new org.infinispan.client.hotrod.configuration.ConfigurationBuilder();
+//other client configuration parameters
+cb.transaction().transactionManagerLookup(GenericTransactionManagerLookup.getInstance());
+cb.transaction().transactionMode(TransactionMode.NON_XA);
+RemoteCacheManager rmc = new RemoteCacheManager(cb.build());
+----
+
+For a complete reference to the available configuration option please refer to the {cb}'s javadoc.
```
tristantarrant: s/option/options/

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+----
+# other client configuration properties
+infinispan.client.hotrod.transaction.transaction_manager_lookup = org.infinispan.client.hotrod.transaction.lookup.GenericTransactionManagerLookup
+infinispan.client.hotrod.transaction.transaction_mode = NON_XA
+----
+
+====== The `TransactionManagerLookup` interface
+
+This interface is the entry point to fetch the {tm} to use.
+There are a couple of implementation available and you can implement this interface to add your own {tm}.
```
tristantarrant: s/implementation/implementations/

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+The available implementation are:
```
tristantarrant: s/implementation/implementations/

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+* {gtml}:
+This is a lookup class that locate {tm} in the most popular Java EE application servers.
```
tristantarrant: {tm}s

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+If no {tm} can be found, it defaults on the {rtm}.
```
tristantarrant: s/on/to/

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+* {rtml}:
+This provides with a basic {tm} which should only be used when no other implementation is available.
```
tristantarrant: s/with//

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+concurrent transactions and recovery.
+
+====== The `TransactionMode`
+
+The {tx-mode} tells how the {rc} is going to interact with the {tm}.
+The possible modes are:
+
+* NON_XA: The {rc} interacts via {sync}.
+* NON_DURABLE_XA: The {rc} interacts via {xa} but without recovery capabilities enabled.
+* FULL_XA: Same as above byt with recovery available.
```
tristantarrant: s/byt/but/

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod_transaction.adoc
```diff
@@ -0,0 +1,176 @@
...
+TransactionManager tm = cache.getTransactionManager();
+
+//just a simple transaction.
+tm.begin();
+cache.put("k1", "v1");
+System.out.println("K1 value is " + cache.get("k1"));
+tm.commit();
+----
+
+//recovery not yet implemented.
```
tristantarrant: Remove this line and amend the paragraph above where you list the possible tx modes to indicate which ones work and which ones don't yet.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.java
```diff
@@ -323,15 +323,17 @@
...
-   private final List<ClusterConfigurationBuilder> clusters = new ArrayList<ClusterConfigurationBuilder>();
+   private final List<ClusterConfigurationBuilder> clusters = new ArrayList<>();
 
    public ConfigurationBuilder() {
       this.classLoader = new WeakReference<>(Thread.currentThread().getContextClassLoader());
       this.connectionPool = new ConnectionPoolConfigurationBuilder(this);
       this.asyncExecutorFactory = new ExecutorFactoryConfigurationBuilder(this);
       this.security = new SecurityConfigurationBuilder(this);
       this.nearCache = new NearCacheConfigurationBuilder(this);
+      transaction = new TransactionConfigurationBuilder(this);
```
tristantarrant: for symmetry either qualify the field with this. or remove the others :smile:

tristantarrant:
Also a number of java.lang.NoClassDefFoundError: javax/transaction/RollbackException Caused by: java.lang.ClassNotFoundException: javax.transaction.RollbackException

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/TransactionalRemoteCacheImpl.java
```diff
@@ -0,0 +1,303 @@
...
+   }
+
+   @Override
+   public CompletableFuture<Boolean> removeWithVersionAsync(K key, long version) {
+      TransactionContext<K, V> txContext = getTransactionContext();
+      return txContext == null ?
+            super.removeWithVersionAsync(key, version) :
+            txContext.compute(key,
+                  entry -> removeEntryIfSameVersion(entry, version),
+                  super::getWithMetadata);
```
rvansa: uncached method reference...

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/TransactionalRemoteCacheImpl.java
```diff
@@ -0,0 +1,303 @@
...
+      TransactionContext<K, V> txContext = getTransactionContext();
+      if (txContext == null) {
+         return super.removeAsync(key);
+      }
+      if (forceReturnValue) {
+         //noinspection unchecked
+         return txContext.compute((K) key, this::removeEntry, super::getWithMetadata);
+      } else {
+         //noinspection unchecked
+         return txContext.compute((K) key, this::removeEntry);
```
rvansa: uncached...

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/TransactionalRemoteCacheImpl.java
```diff
@@ -0,0 +1,303 @@
...
+      assertRemoteCacheManagerIsStarted();
+      Transaction tx = getRunningTransaction();
+      return tx == null ? null : transactionTable.enlist(this, tx);
+   }
+
+   private Transaction getRunningTransaction() {
+      try {
+         return transactionManager.getTransaction();
+      } catch (SystemException e) {
+         return null;
```
rvansa: log exception?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/entry/ControlByte.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @author Pedro Ruivo
+ * @since 9.3
+ */
+public enum ControlByte {
+   NOT_READ(0x1),
+   NON_EXISTING(0x2),
+   REMOVE_OP(0x4);
+
+   private final byte bit;
```
rvansa: bitSet?

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/tx/util/TransactionSetup.java
```diff
@@ -0,0 +1,58 @@
...
+      } else {
+         //use JBossTM as default (as in core)
+         System.out.println("Transaction manager used: JBossTM");
+         JBossStandaloneJTAManagerLookup tmLookup = new JBossStandaloneJTAManagerLookup();
+         tmLookup.init(new GlobalConfigurationBuilder().build());
+         lookup = tmLookup;
+      }
+   }
+
+   public static void ammendJTA(ConfigurationBuilder builder) {
```
rvansa: typo (double `m`)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
```diff
@@ -202,6 +209,21 @@ public RemoteCacheManager() {
       return createRemoteCache("", forceReturnValue);
    }
 
+   @Override
+   public <K, V> TransactionalRemoteCache<K, V> getTransactionalCache(String cacheName,
```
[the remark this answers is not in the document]
rvansa: ... above in the sense that I'd omit the TransactionalRemoteCacheImpl altogether.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/PrepareTransactionOperation.java
```diff
@@ -0,0 +1,81 @@
...
+   @Override
+   protected void executeOperation(Channel channel) {
+      retry = false;
+      scheduleRead(channel);
+      ByteBuf buf = channel.alloc().buffer(estimateSize());
+      codec.writeHeader(buf, header);
+      writeXid(buf, xid);
+      buf.writeByte(onePhaseCommit ? 1 : 0);
+      writeVInt(buf, modifications.size());
+      modifications.forEach(m -> m.writeTo(buf, codec));
```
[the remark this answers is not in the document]
tristantarrant: @rvansa is that true for all lambdas or only for capturing ones ?

pruivo:
@rvansa it uses JBossTM as default. And the pessimistic/optimistic tests are already done in server. no need to duplicate them again.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/CompleteTransactionOperation.java
```diff
@@ -0,0 +1,57 @@
...
+      ByteBufUtil.writeXid(buf, xid);
+      channel.writeAndFlush(buf);
+   }
+
+   @Override
+   public void acceptResponse(ByteBuf buf, short status, HeaderDecoder decoder) {
+      if (status == NO_ERROR_STATUS) {
+         complete(buf.readInt());
+      } else {
+         complete(XAResource.XA_OK);
```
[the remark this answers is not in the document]
pruivo: no. any unexpected status (or exception in server) should be handled by the codec. 
I'll prefer to return the rollback code instead of complete exceptionally.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/PrepareTransactionOperation.java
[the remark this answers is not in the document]
pruivo: I'm always forgetting that you and Dan don't like lambdas, methods reference, stream()... anything else? :)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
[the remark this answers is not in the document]
pruivo: the problem is the transaction manager is always there (it defaults to RemoteTM, to keep it symmetric with embedded) but we don't have any per-cache configuration in the client. and I don't want to add it.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/SyncModeTransactionTable.java
```diff
@@ -0,0 +1,238 @@
...
+      }
+      if (trace) {
+         log.tracef("Registered synchronization for transaction %s. Sync=%s", transaction, adapter);
+      }
+      return adapter;
+   }
+
+   private static class SynchronizationAdapter implements Synchronization {
+
+      private final Map<String, TransactionContext<?, ?>> registeredCaches = new ConcurrentSkipListMap<>();
```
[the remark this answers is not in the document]
pruivo: do you mean 1 Sync for all caches vs 1 Sync for each cache? 

* first, it avoids deadlocks if you have multiple caches accessing the same keys. Since the Sync can be invoked by any order, having a single Sync for all caches makes sure deadlocks doesn't occur.

* second, if we want, we can run commit/rollback phase in parallel :D

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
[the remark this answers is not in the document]
rvansa: Okay, then use `tm.getStatus()` in any call. We don't have a TransactionalCache interface and tx impl in embedded mode either, and there was a strong call for parity.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/PrepareTransactionOperation.java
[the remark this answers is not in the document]
rvansa: @tristantarrant Metaspace is consumed by both capturing and non-capturing lambdas, even by method handles. It uses about 1.5 kB of metaspace for each of the above in the code (there are some fixes in OpenJDK to bring that down to 1 kB or so).
Capturing lambdas are different story: these allocate data on heap, so increase GC pressure rather than just occupying static memory.

@pruivo Portuguese... :) I love lambdas, but the readability improvements has to justify the metaspace costs. Simple `forEach` that can be replaced with `for (... : ...)` does not IMO. And I dislike the need to cache this::foo references, but when it's on the hot path...

rvansa:
@pruivo [pessimistic/optimistic] Good enough for me... Btw. big kudos that you didn't require server-side changes when working on the client, apparently the server was well tested.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/SyncModeTransactionTable.java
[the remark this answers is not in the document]
rvansa: I meant `ConcurrentSkipListMap` and `LinkedList` rather than `ConcurrentHashMap` and `ArrayList`.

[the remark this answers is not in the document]
pruivo: The ConcurrentSkipListMap ensures the values() returns the TxContext by the natural order for the keys of the map. 
If you use the ConcurrentHashMap, then values() can return the TxContext by any order and it may create deadlocks.

The LinkedList is because I don't know at priory the number of caches involved.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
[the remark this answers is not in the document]
pruivo: That won't work , unfortunately. As an example, you can have non-tx cache invocations inside a transaction:

```
tm.begin()
tx-cache.put(k,v);
non-tx-cache.put(k,v); <= in this case, tm.getStatus() will return an active transaction
```

The way we do it in embedded is to have a null TM for non-tx caches. But we ended with a lot of `if (transactionManager != null)`. I would prefer to split the cache implementation in embedded than merge remove the TransactionalRemoteCache...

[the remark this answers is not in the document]
rvansa: @pruivo Since you don't have any per-cache configuration there'd be nothing like a non-transactional remote cache. If user wants to do a non-transactional operation when he's in the middle of an active transaction, he'd have to suspend the running tx and resume is afterwards - it's not so much of a hassle.

rvansa:
@pruivo I've realized that I miss the documentation how do the HR transactions interact with server side transactions:

* I suppose that server-side caches that are used for TxRemoteCache have to be configured as transactional. If this does not hold, does the operation fail upon ping, or during first prepare? If the latter I'd do a noop transaction instead of ping.
* Does a non-tx RemoteCache connected to transactional server-side cache do autocommit transactions?
* Can we have a FULL_XA on server while NON_XA on client? Is that really a NON_XA, then? Does NON_DURABLE_XA/FULL_XA distinction make sense at all when we can recover on server-side? (Synchronization vs. XAResource distinction makes sense at least in reporting, where failed commit in Synchronization does not fail the rest of the tx).

pruivo:
@rvansa you're right. I forgot to document the server changes

* the cache needs to be transactional and with repeatable-read. the check happens during prepare (we will abort the tx) because, AFAIK, the ping on startup is optional (?)
* you can disable auto-commit in server. so,all non-tx request will happen in auto commit transactions
* it doesn't matter because the server transaction isn't JTA compliant. I'm gonna need to implement recovery and probably we'll need some internal tooling for recovering in the server (and here I'm thinking the case where a transaction is prepared and the client crashes)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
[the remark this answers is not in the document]
rvansa: @pruivo Old problem? You'll enlist all caches as either sync or xa, depending on the global configuration. I don't say that you shouldn't have some `getCache(...)` overloads, though I wonder how useful these are when you can create another RCM.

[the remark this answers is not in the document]
pruivo:
```
When creating the {rcm}, it is possible to set the default {tm} and {tx-mode} to use by the {rc}.
However, it is possible to set a different {tm} and {tx-mode} for each cache.
```
that's why I have the getTxCache methods so the user can change the TM or the TxMode per cache.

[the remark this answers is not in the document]
rvansa: As a user, I don't care too much about sync/xa, I want to call `getCache("foo")` without any overloads and get an instance that I can later use for transactions.
I don't want to care about extra interface and rewrite all signatures in my code from `RemoteCache` to `TransactionalRemoteCache`, just to be able to get the TM.
Least of all, the `Transactional` in `getTransactionalCache` is superfluous - the cache is a concept, it depends on the way how I use it.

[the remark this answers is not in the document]
pruivo: I disagree. 
IMO, the user should be able to choose to which `TransactionManager` he wants to interact with and how the cache should interact with. It is the same idea as `forceReturnValue` and `getCache(cacheName, forceReturnValues)`.  It isn't fair to make the user create N `RemoteCacheManager` to use the caches differently. 
It also keeps symmetric with embedded mode.

I guess will need a 3rd opinion otherwise we won't move forward. @tristantarrant or @danberindei, any ideas?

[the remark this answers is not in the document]
rvansa: @pruivo I was never suggesting to remove the option, overloads are completely fine. I just said that as long as global configuration is set correctly, the user should not be forced to use anything but `getCache(name)`.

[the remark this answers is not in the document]
rvansa: In my view the `getTransactionalCache` methods are introducing the asymmetry.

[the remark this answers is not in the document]
danberindei: @pruivo I think `getCache(name, forceReturnValues)` makes sense because there's no configuration behind that, either on the client or in the server. If you need some extra configuration for a cache to work as a tx cache, then there's no point in having an overload with an extra parameter that fails when the user calls it with the wrong parameter.

TBH I haven't read most of the PR yet so I don't know where you stand :)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
```diff
@@ -411,6 +432,63 @@ public ChannelFactory getChannelFactory() {
...
+   }
+
+   private TransactionMode getTransactionMode(TransactionMode override) {
+      return override == null ? configuration.transaction().transactionMode() : override;
+   }
+
+   private TransactionTable getTransactionTable(TransactionMode transactionMode) {
+      switch (transactionMode) {
+         case NON_XA:
+            return syncTransactionTable;
```
danberindei: `NON_XA` means `useSynchronization(true)`, which hides all commit errors from the user. Couldn't we support at least `NON_DURABLE_XA`?

commons/src/main/java/org/infinispan/commons/tx/TransactionImpl.java
```diff
@@ -343,7 +343,7 @@ public void runCommit(boolean forceRollback)
 
    @Override
    public String toString() {
-      return "EmbeddedTransaction{" +
+      return "TransactionImp{" +
```
danberindei: Missing `l`

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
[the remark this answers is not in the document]
pruivo: > there's no configuration behind that

@danberindei what do you mean by that? https://github.com/infinispan/infinispan/blob/master/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.java#L450

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
pruivo: I split the work by JIRA. this is the first one. then I'll add XA and finally recovery

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/entry/ControlByte.java
```diff
@@ -0,0 +1,58 @@
...
+      }
+      return builder.append("]").toString();
+   }
+
+   /**
+    * Sets {@code this} flag to the {@code bitSet}.
+    *
+    * @return The new bit set.
+    */
+   public byte set(byte bitSet) {
```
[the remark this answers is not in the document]
pruivo: what is your suggestion?

pruivo:
@rvansa about the ping-on-start, I'll have to add a new ping operation. 
is it ok to wait for the next major release and change the current ping format instead of creating a new operation?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/RemoteCacheManager.java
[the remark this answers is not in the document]
danberindei: I forgot that it's in the configuration as well, but my point was that you can enable `forceReturnValues` just by calling `getCache(name, true)`, without having anything about it in the configuration.

pruivo:
updated!

tristantarrant:
Merged. Thanks !

PR #5975, opened 2018-05-14
ISPN-9130 Scattered cache with bias sends ack too early

rvansa:
https://issues.jboss.org/browse/ISPN-9130


core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -258,12 +242,31 @@ private BaseAckTarget(long id, int topologyId) {
          this.timeoutTask = timeoutExecutor.schedule(this, timeoutNanoSeconds, TimeUnit.NANOSECONDS);
       }
 
+      /**
+       * Invoked when the future is completed, it must cleanup all task related to this collector.
```
danberindei: Since the future is not `timeoutTask`, the comment is a bit ambiguous. I suggest adding a protected method `cleanupAfterFuture(CF)` and call that from the subclasses so you can reference that in the javadoc and so subclasses don't have to know whether it's a `BiConsumer` or a `BiFunction`.

I also suggest extracting the entire body in a separate method, e.g. `cleanup()`: with a name like that, it wouldn't need as much javadoc.

core/src/main/java/org/infinispan/interceptors/distribution/BiasedScatteredDistributionInterceptor.java
```diff
@@ -187,21 +194,21 @@ protected void manyWriteResponse(InvocationContext ctx, WriteCommand cmd, Object
...
+            }));
          }
       }
-      Address[] waitForAddresses = waitFor == null ? null : waitFor.toArray(new Address[waitFor.size()]);
-      rpcManager.sendTo(ctx.getOrigin(), cf.buildPrimaryAckCommand(cmd.getCommandInvocationId().getId(),
-            cmd.isSuccessful(), returnValue, waitForAddresses), DeliverOrder.NONE);
+      if (waitFor == null) {
+         return returnValue;
+      } else {
+         return new BiasRevocationResponse(returnValue, waitFor.toArray(new Address[waitFor.size()]));
```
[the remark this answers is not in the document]
rvansa: Great post, thanks! And yes, I miss a nice way for waiting for a bunch of futures, too...

About revocation: we don't need to invoke the `RevokeBiasCommand` only when `revocation == null`. The `shouldRevoke` might be mis-named: it tells you if you can revoke the command now, or if you need to schedule this using `handleCompose`. Note that the list of all addresses is assembled from both `revocation.biased()` and the futures.
I'll rename it to `canRevokeNow` for better clarity.

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
rvansa: With the ResponseCollector approach I'll revert the BiConsumer / BiFunction change, so this won't be needed.

rvansa:
@danberindei Rebased && addressed your comments

core/src/main/java/org/infinispan/interceptors/distribution/BiasedScatteredDistributionInterceptor.java
```diff
@@ -38,18 +41,23 @@ public void inject(CommandAckCollector collector, BiasManager biasManager) {
...
+      super.singleWriteOnRemotePrimary(target, command, collector);
+      return collector.getFuture();
    }
 
    @Override
-   protected CompletionStage<Map<Address, Response>> manyWriteOnRemotePrimary(Address target, WriteCommand command) {
-      Collector<Response> collector = commandAckCollector.createMultiTargetCollector(command.getCommandInvocationId().getId(), target, command.getTopologyId());
-      rpcManager.sendTo(target, command, DeliverOrder.NONE);
-      return collector.getFuture().thenApply(value -> Collections.singletonMap(target, value));
+   protected CompletionStage<ValidResponse> manyWriteOnRemotePrimary(Address target, WriteCommand command, ResponseCollector<ValidResponse> unused, MultiTargetCollector multiTargetCollector) {
```
danberindei: I think the unused parameters complicate things a bit, I'd rather inline the call to `super` so only the `multiTargetCollector` needs to be passed in. Same in `singleWriteOnRemotePrimary`. WDYT?

core/src/main/java/org/infinispan/util/concurrent/CommandAckCollector.java
```diff
@@ -520,37 +488,18 @@ public String toString() {
...
-      public void primaryException(Throwable throwable) {
-         throw new UnsupportedOperationException();
-      }
-
-      @Override
-      public void primaryResult(T result, boolean success) {
-         throw new UnsupportedOperationException();
-      }
-   }
+   private static class CombiningCollector implements BiasedCollector, Function<Void, CompletableFuture<ValidResponse>> {
```
danberindei: I'd like a bit of javadoc here explaining what is being combined, at a first glance it looks like it just wraps a `MultiTargetCollectorImpl`

rvansa:
Done; I've renamed the `CombiningCollector` to `SingleTargetCollectorImpl`.

danberindei:
Integrated, thanks Radim!

PR #5978, opened 2018-05-14
 ISPN-8944 Maven profile traceTests no longer works

danberindei:
https://issues.jboss.org/browse/ISPN-8944
https://issues.jboss.org/browse/ISPN-9190

* Fix the log4j2 config in the traceTests profile
* Split trace log into one file per test


danberindei:
Added a custom purge policy.
Updated `bin/filter_failed_logs.sh` to take a failures log as an argument and move/rename the logs of the failed tests with a prefix.

danberindei:
@tristantarrant I found a way to change the `GZipOutputStream` compression level to 3, which should be 2x faster according to http://java-performance.info/performance-general-compression/. The file size should be ~ 30% bigger than with level 6, but fortunately I found that changing `syncFlush` from true to false more than made up for that.

I also did some experiments with no compression and with LZ4, and the run time was about the same (actually a bit higher with LZ4, but I only ran it once). So I think it's best to stick with `gzip` for now.

PR #5979, opened 2018-05-15
Unquarantine some tests

tristantarrant:


tristantarrant:
4 runs in CI and none of the re-enabled tests have failed.

core/src/test/java/org/infinispan/partitionhandling/ScatteredSplitAndMergeTest.java
```diff
@@ -56,7 +56,6 @@ public void testSplitAndMerge4() throws Exception {
          new PartitionDescriptor(DEGRADED_MODE, 3));
    }
 
-   @Test(groups = "unstable", description = "ISPN-8431")
```
danberindei: I'm not sure this is fixed for scattered caches, but let's keep the JIRA open and let's see.

danberindei:
Integrated, thanks Tristan!

PR #5980, opened 2018-05-15
ISPN-9150 RemoteCache additional tests

wburns:
* Added tests for the forEach method on collections
* Make sure tests were testing remote size as well

https://issues.jboss.org/browse/ISPN-9150


wburns:
Note this just adds an additional test and tightens up the assertions for a couple others. This doesn't close the issue itself as I am unsure how to reproduce the bug.

PR #5985, opened 2018-05-16
Write-behind Fault Tolerance

ryanemerson:
https://issues.jboss.org/browse/ISPN-6677
https://issues.jboss.org/browse/ISPN-8008

~~I still need to adapt the existing Async tests to test both failSilently enabled and disabled.~~
Having looked into the tests in greater detail, I'm not sure this will provide any benefit as the existing tests do not focus on faults in the underlying store.

I still need to to implement `isAvailable` on the appropriate store implementations.


tristantarrant:
I think it is quite harmless and probably good enough. So please go ahead and add the others.

tristantarrant:
One failure is certainly caused by this PR: https://ci.infinispan.org/job/Infinispan/job/PR-5985/8/testReport/junit/org.infinispan.persistence.remote.upgrade/HotRodUpgradeWithStoreTest/testSynchronization/

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -115,8 +126,33 @@ public void start() {
...
-      coordinator = coordinatorThreadFactory.newThread(new AsyncStoreCoordinator());
+
+      coordinator = coordinatorThreadFactory.newThread(new AsyncStoreCoordinator(asyncConfiguration.failSilently()));
       coordinator.start();
+
+      // If no errors on start, then start observing the status of the underlying store. We monitor the availability of
+      // the delegate cache store here, opposed to the PersistenceManager, as it's possible for AsyncCacheWriter to be
+      // availabile even if the delegate isn't, this occurs when the modification queue is still able to queue requests
+      long interval = ctx.getCache().getCacheConfiguration().persistence().availabilityInterval();
+      Observable.interval(interval, TimeUnit.MILLISECONDS, Schedulers.from(ctx.getExecutor()))
```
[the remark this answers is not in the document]
wburns: Also is this even needed? Doesn't the `PersistenceManagerImpl` do this? Can't we just update the `availability` when the manager queries it?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -750,10 +780,18 @@ private void initializeLoader(StoreConfiguration cfg, CacheLoader loader, Initia
          storesMutex.writeLock().lock();
          try {
             loaders.add(loader);
+
+            long interval = configuration.persistence().availabilityInterval();
+            ConnectableObservable<StoreAvailability> observable = Observable.interval(interval, TimeUnit.MILLISECONDS, Schedulers.from(persistenceExecutor))
+                  .map(t -> new StoreAvailability(loader, loader.isAvailable()))
+                  .doOnError(err -> log.debugf("Error encountered when calling isAvailable on %s: %s", loader, err))
```
wburns: Maybe we want to do http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#onErrorReturn-io.reactivex.functions.Function- and print the error and change it to a not available status `new StoreAvailability(loader, false)`?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -750,10 +780,18 @@ private void initializeLoader(StoreConfiguration cfg, CacheLoader loader, Initia
...
+                  .retry() // Swallow any exceptions so that a call is made to isAvailable on the next interval
+                  .publish();
```
wburns: I am not sold on using a `ConnectableObservable`. We only have a single subscriber and we don't ever invoke `connect. You should be able to just remove this line.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -137,39 +148,39 @@ public void start() {
...
-                  }
-               }
-            }
+            Set<Lifecycle> undelegated = new HashSet<>();
+            nonTxWriters.forEach(w -> startWriter(w, undelegated));
+            txWriters.forEach(w -> startWriter(w, undelegated));
+            loaders.forEach(l -> startLoader(l, undelegated));
+
+            // Observe all status checks and update store statuses when they are received
+            List<Observable<StoreAvailability>> availabilityChecks = storeStatuses.values().stream().map(s -> s.observable).collect(Collectors.toList());
```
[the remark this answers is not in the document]
wburns: Also it seems a bit odd that we may have stores firing their status checks at different times.

Is there any reason we don't implement this like this?

```
Observable.interval(interval, TimeUnit.MILLISECONDS, Schedulers.from(persistenceExecutor))
   .flatMapIterabl(ignore -> storeStatuses.keySet())
   .map(obj -> ... cast to writer/loader and invoke isAvailable ...)
   .onErrorReturn(e -> log error and return false)
   .retry()
   .publish(... current method tweaked...)
```
Then we should retain the `Disposable` returned from `publish` method so we can stop this all when the persistence manager is stopped or when all stores have been disabled (ie. `disableStore`)

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -242,14 +300,34 @@ public void run() {
                final boolean shouldStop;
                stateLock.readLock();
                try {
+                  if (!failSilently) {
+                     availabilityLock.lock();
+                     try {
+                        // If the delegate is unavailable, relinquish the readLock and await for the delegate to become available
+                        if (!delegateAvailable) {
+                           stateLock.readUnlock();
```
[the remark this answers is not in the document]
ryanemerson: I didn't want to keep the `stateLock` locked while we await the `availability` condition. But I have updated the code so that the `!failSilently` block now occurs before the stateLock's try/finally statement. That way the `stateLock` is only unlocked once when we continue the loop.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -395,76 +421,67 @@ public void purgeExpired() {
...
       } catch (Exception e) {
          log.exceptionPurgingDataContainer(e);
       }
    }
 
    @Override
    public void clearAllStores(AccessMode mode) {
-      storesMutex.readLock().lock();
-      try {
+      runnableWithReadLockAndAvailabilityCheck(() -> {
```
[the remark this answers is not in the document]
ryanemerson: On reflection I think it's best to just remove the methods and stick with acquiring the lock and calling `checkStoreAvailability` in each method. The potential for a performance regression is too high for such little benefit.

ryanemerson:
@wburns As discussed on Zulip, I decided to simplify things and just remove the use of Observables etc and just utilise a `ScheduledExecutorService` for the periodic polling. The `StringBasedStoreInvalidationCacheSyncIT` on [build #11](https://ci.infinispan.org/job/Infinispan/job/PR-5985/11/) is not reproducable locally for me, but I have initiated another build to make sure it's not related.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -115,8 +124,38 @@ public void start() {
...
+      try {
+         available = actual.isAvailable();
+      } catch (Throwable t) {
+         log.debugf("Error encountered when calling isAvailable on %s: %s", actual, t);
+      }
+      availabilityLock.lock();
+      try {
+         if (available != delegateAvailable) {
+            if (available) {
+               delegateAvailable = true;
```
wburns: I would say just assign `delegateAvailable = available` and remove the else block.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -115,8 +124,38 @@ public void start() {
...
+
+      coordinator = coordinatorThreadFactory.newThread(new AsyncStoreCoordinator(asyncConfiguration.failSilently()));
       coordinator.start();
+
+      pollStoreAvailability();
+      // If no errors on start, then start observing the status of the underlying store. We monitor the availability of
+      // the delegate cache store here, opposed to the PersistenceManager, as it's possible for AsyncCacheWriter to be
+      // available even if the delegate isn't, this occurs when the modification queue is still able to queue requests
+      long interval = ctx.getCache().getCacheConfiguration().persistence().availabilityInterval();
+      ctx.getExecutor().scheduleAtFixedRate(this::pollStoreAvailability, interval, interval, TimeUnit.MILLISECONDS);
```
wburns: We should retain a reference to the `Future` so we can cancel it when the `stop` method is invoked.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -355,16 +438,50 @@ public void run() {
...
                if (log.isDebugEnabled())
                   log.debug("Failed to process async modifications", e);
+
+               if (!failSilently) {
+                  try {
+                     // Wait for availabilityInterval time to ensure that before the next attempt the delegate's availability
+                     // flag will have been updated and availability.await will be reached
+                     Thread.sleep(configuration.availabilityInterval());
+                  } catch (InterruptedException ie) {
+                     Thread.currentThread().interrupt();
```
wburns: I wonder if we should just break out of the entire loop if we are interrupted?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -232,6 +239,21 @@ public boolean isEnabled() {
       return enabled;
    }
 
+   private void checkStoreAvailability() {
+      if (!enabled) return;
+
+      if (!isAvailable()) {
```
wburns: Couldn't this just check `unavailableException` if it is null or not? Seems we wouldn't need the `managerAvailable` variable as well then.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
wburns: I am still not sure why the `AsyncCacheWriter` needs to schedule this? Couldn't it just check the availability when the `PersistenceManager` calls `isAvailable`?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -135,52 +141,53 @@ public void start() {
...
-               if (l instanceof DelegatingCacheLoader) {
-                  CacheLoader actual = undelegate(l);
-                  if (!undelegated.contains(actual)) {
-                     actual.start();
-                  }
-               }
-            }
+            // Now schedule the availability check
+            long interval = configuration.persistence().availabilityInterval();
+            persistenceExecutor.scheduleAtFixedRate(this::pollStoreAvailability, interval, interval, TimeUnit.MILLISECONDS);
```
wburns: We should retain a reference to the `Future` so we can stop it when the `stop` method is invoked or when all cache stores have been disabled since there is no way to add a new store to a running system.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -135,52 +141,53 @@ public void start() {
...
+      try {
+         for (StoreStatus status : storeStatuses.values()) {
+            if (!status.availabilityChanged())
+               continue;
+
+            cacheNotifier.notifyPersistenceAvailabilityChanged(status.availability);
+            if (!status.availability) {
+               unavailableException = new StoreUnavailableException(String.format("Store %s is unavailable", status.store));
+               managerAvailable = false;
+               return;
```
wburns: Should we check all of the stores availability, even when one fails?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -786,7 +817,67 @@ private CacheLoader undelegate(CacheLoader l) {
...
       return (w instanceof DelegatingCacheWriter) ? ((DelegatingCacheWriter)w).undelegate() : w;
+   }
 
+   private void waitForConnectionInterval() {
+      int connectionInterval = configuration.persistence().connectionInterval();
+      if (connectionInterval > 0) {
+         try {
+            Thread.sleep(connectionInterval);
+         } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
wburns: Same thing here, maybe we should return if it waited full time or was interrupted. I would think we would want to end early if we were interrupted and not do all the connection attempts.

core/src/main/java/org/infinispan/configuration/cache/PersistenceConfiguration.java
```diff
@@ -13,17 +13,26 @@
...
+   public static final AttributeDefinition<Integer> AVAILABILITY_INTERVAL = AttributeDefinition.builder("availabilityInterval", 1000).immutable().build();
+   public static final AttributeDefinition<Integer> CONNECTION_ATTEMPTS = AttributeDefinition.builder("connectionAttempts", 10).immutable().build();
+   public static final AttributeDefinition<Integer> CONNECTION_INTERVAL = AttributeDefinition.builder("connectionInterval", 50).immutable().build();
    static AttributeSet attributeDefinitionSet() {
-      return new AttributeSet(PersistenceConfiguration.class, PASSIVATION);
+      return new AttributeSet(PersistenceConfiguration.class, PASSIVATION, AVAILABILITY_INTERVAL, CONNECTION_ATTEMPTS, CONNECTION_INTERVAL);
    }
 
    private final Attribute<Boolean> passivation;
+   private final Attribute<Integer> availabilityInterval;
```
wburns: This should probably be a long I would think.

core/src/main/java/org/infinispan/configuration/cache/PersistenceConfiguration.java
```diff
@@ -13,17 +13,26 @@
...
+   private final Attribute<Integer> connectionAttempts;
+   private final Attribute<Integer> connectionInterval;
```
wburns: This should probably be a long I would think.

core/src/test/java/org/infinispan/persistence/PersistenceManagerAvailabilityTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.testng.annotations.Test;
+
+@CleanupAfterMethod
+@Test(testName = "persistence.PersistenceManagerAvailabilityTest", groups = "functional")
+public class PersistenceManagerAvailabilityTest extends AbstractInfinispanTest {
+
+   private Cache<Object, Object> createManagerAndGetCache(int startFailures) {
+      GlobalConfiguration globalConfiguration = new GlobalConfigurationBuilder().globalJmxStatistics().allowDuplicateDomains(true).build();
+      ConfigurationBuilder cb = TestCacheManagerFactory.getDefaultCacheConfiguration(false);
+      Configuration config = cb.persistence()
```
wburns: We should lower the availabilityInterval to a smaller number or mock the `ScheduledExecutorService` to run the operation whenever we want in the same thread.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -144,6 +191,14 @@ public void stop() {
       }
    }
 
+   @Override
+   public boolean isAvailable() {
+      // Available if failSilently || actual == available || actual != available and queue has capacity
+      // Worst case, writeBatch comes in before !Available is declared by Async, in which case it will wait on
+      // writeLock until the stateLock is reset when the queue is finally flushed
+      return !stopped && (asyncConfiguration.failSilently() || delegateAvailable || stateLock.hasCapacity());
```
wburns: It seems updating the `delegateAvailable` every time this is invoked would be remedy the issue of having another timer task?

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
ryanemerson: Is that achieved by the break statement below? I used break instead of return so that the error message would still be executed.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
ryanemerson: Probably ... I was thinking that there is no need to check all stores availability if one is unavailable, as that is enough for the PersistenceManager to become unavailable. However, I guess it makes more sense to keep track of all store statuses so that the PM can become available sooner in the event of the unavailable store being removed from the PM.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
ryanemerson: You're totally right, I have just over complicated things :confused:. Originally the availability interval was going to be implemented on a per store basis and I think that skewed my thinking.

core/src/main/java/org/infinispan/configuration/cache/PersistenceConfiguration.java
ryanemerson: Isn't Integer.MAX_VALUE equivalent to over 3 weeks in milliseconds?

core/src/main/java/org/infinispan/configuration/cache/PersistenceConfiguration.java
ryanemerson: Isn't Integer.MAX_VALUE equivalent to over 3 weeks in milliseconds?

wburns: I guess it is fine as an `int`. I am just used to any field being a time always being a `long`.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -355,16 +432,49 @@ public void run() {
...
                if (log.isDebugEnabled())
                   log.debug("Failed to process async modifications", e);
+
+               if (!failSilently) {
+                  try {
+                     // Wait for availabilityInterval time to ensure that before the next attempt the delegate's availability
+                     // flag will have been updated and availability.await will be reached
+                     Thread.sleep(configuration.availabilityInterval());
+                  } catch (InterruptedException ie) {
+                     break;
```
wburns: Missed resetting the interrupt flag.

ryanemerson: I thought that's what you told me to remove? lol

previously it was:
```
break;
```

wburns: Ah I must have missed what it was before, sorry for the confusion.

wburns:
Looks like this test failure is related https://ci.infinispan.org/view/Infinispan/job/Infinispan/view/change-requests/job/PR-5985/14/testReport/junit/org.infinispan.persistence.remote.upgrade/HotRodUpgradeWithStoreTest/testSynchronization/

ryanemerson:
@wburns Fix pushed

wburns:
I will verify the test manually real quick.

wburns:
Test passed for me [INFO] Tests run: 75, Failures: 0, Errors: 0, Skipped: 0

wburns:

PR #5989, opened 2018-05-17
ISPN-9128 RehashWithSharedStoreTest.testRehashes random failures

rvansa:
https://issues.jboss.org/browse/ISPN-9128

Includes https://issues.jboss.org/browse/ISPN-9095 as well. 

The fix is just one liner :)

### DIFF
--- a/core/src/main/java/org/infinispan/partitionhandling/impl/AvailabilityStrategy.java
+++ b/core/src/main/java/org/infinispan/partitionhandling/impl/AvailabilityStrategy.java
@@ -26,13 +26,17 @@ import org.infinispan.topology.CacheTopology;
 public interface AvailabilityStrategy {
    /**
     * Compute the read consistent hash for a topology with a {@code null} union consistent hash.
+    * Originally a copy of {@link CacheTopology#getReadConsistentHash()} but differs in case of scattered cache.
     */
-   static ConsistentHash readConsistentHash(CacheTopology topology, ConsistentHashFactory chFactory) {
+   static ConsistentHash ownersConsistentHash(CacheTopology topology, ConsistentHashFactory chFactory) {
       switch (topology.getPhase()) {
          case NO_REBALANCE:
             return topology.getCurrentCH();
          case TRANSITORY:
-            return topology.getPendingCH();
+            // This is used to determine nodes that own the entries. In scattered cache (which uses transitory topology)
+            // the pendingCH is used for reading but the nodes in there are not guaranteed to have the data yet.
+            // CurrentCH should be safe - the nodes either have the data or the owner is unknown.
+            return topology.getCurrentCH();
          case CONFLICT_RESOLUTION:
          case READ_OLD_WRITE_ALL:
             return topology.getCurrentCH();
--- a/core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
+++ b/core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
@@ -499,6 +499,11 @@ public class ScatteredStateConsumerImpl extends StateConsumerImpl {
       for (int i = 0; i < keys.size(); ++i) {
          Object key = keys.get(i);
          InternalCacheValue icv = values[i];
+         if (icv == null) {
+            // The entry got lost in the meantime - this can happen when the container is cleared concurrently to processing
+            // the GetAllCommand. We'll just avoid NPEs here: data is lost as > 1 nodes have left.
+            continue;
+         }
          PutKeyValueCommand put = commandsFactory.buildPutKeyValueCommand(key, icv.getValue(),
                keyPartitioner.getSegment(key), icv.getMetadata(), STATE_TRANSFER_FLAGS);
          try {
--- a/core/src/test/java/org/infinispan/remoting/transport/ControlledTransport.java
+++ b/core/src/test/java/org/infinispan/remoting/transport/ControlledTransport.java
@@ -1,16 +1,20 @@
 package org.infinispan.remoting.transport;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.TimeUnit;
+import static org.infinispan.test.TestingUtil.wrapGlobalComponent;
+import static org.testng.AssertJUnit.assertTrue;
 
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+import org.infinispan.Cache;
 import org.infinispan.commands.ReplicableCommand;
 import org.infinispan.commands.remote.SingleRpcCommand;
 import org.infinispan.remoting.responses.Response;
 import org.infinispan.util.concurrent.ReclosableLatch;
+import org.infinispan.util.logging.Log;
+import org.infinispan.util.logging.LogFactory;
 import org.infinispan.xsite.XSiteReplicateCommand;
 
 /**
@@ -21,15 +25,21 @@ import org.infinispan.xsite.XSiteReplicateCommand;
  * @since 7.0
  */
 public class ControlledTransport extends AbstractDelegatingTransport {
+   private static final Log log = LogFactory.getLog(ControlledTransport.class);
+   private static final Predicate<ReplicableCommand> NEVER = cmd -> false;
 
    private final ReclosableLatch replicationLatch = new ReclosableLatch(true);
    private final ReclosableLatch blockingLatch = new ReclosableLatch(true);
-   private volatile Set<Class> blockBeforeFilter = Collections.emptySet();
-   private volatile Set<Class> blockAfterFilter = Collections.emptySet();
-   private volatile Set<Class> failFilter = Collections.emptySet();
+   private volatile Predicate<ReplicableCommand> blockBeforeFilter = NEVER;
+   private volatile Predicate<ReplicableCommand> blockAfterFilter = NEVER;
+   private volatile Predicate<ReplicableCommand> failFilter = NEVER;
 
-   public ControlledTransport(Transport realOne) {
+   private ControlledTransport(Transport realOne) {
       super(realOne);
+   }
+
+   public static ControlledTransport replace(Cache<?, ?> cache) {
+      return wrapGlobalComponent(cache.getCacheManager(), Transport.class, ControlledTransport::new, true);
    }
 
    @Override
@@ -38,38 +48,50 @@ public class ControlledTransport extends AbstractDelegatingTransport {
    }
 
    public void failFor(Class... filter) {
-      this.failFilter = new HashSet<>(Arrays.asList(filter));
+      failFor(classListToFilter(filter));
+   }
+
+   private void failFor(Predicate<ReplicableCommand> filter) {
+      this.failFilter = filter;
       blockingLatch.open();
    }
 
    public void stopFailing() {
-      this.failFilter = Collections.emptySet();
+      this.failFilter = NEVER;
       blockingLatch.open();
    }
 
    public void blockBefore(Class... filter) {
-      this.blockBeforeFilter = new HashSet<>(Arrays.asList(filter));
+      blockBefore(classListToFilter(filter));
+   }
+
+   public void blockBefore(Predicate<ReplicableCommand> filter) {
+      this.blockBeforeFilter = filter;
       replicationLatch.close();
       blockingLatch.close();
    }
 
    public void blockAfter(Class... filter) {
-      this.blockAfterFilter = new HashSet<>(Arrays.asList(filter));
+      blockAfter(classListToFilter(filter));
+   }
+
+   public void blockAfter(Predicate<ReplicableCommand> filter) {
+      this.blockAfterFilter = filter;
       replicationLatch.close();
       blockingLatch.close();
    }
 
    public void stopBlocking() {
-      getLog().tracef("Stop blocking commands");
-      blockBeforeFilter = Collections.emptySet();
-      blockAfterFilter = Collections.emptySet();
+      log.tracef("Stop blocking commands");
+      blockBeforeFilter = NEVER;
+      blockAfterFilter = NEVER;
       replicationLatch.open();
       blockingLatch.open();
    }
 
    public void waitForCommandToBlock() throws InterruptedException {
-      getLog().tracef("Waiting for at least one command to block");
-      blockingLatch.await(30, TimeUnit.SECONDS);
+      log.tracef("Waiting for at least one command to block");
+      assertTrue(blockingLatch.await(30, TimeUnit.SECONDS));
    }
 
    public boolean waitForCommandToBlock(long time, TimeUnit unit) throws InterruptedException {
@@ -77,7 +99,7 @@ public class ControlledTransport extends AbstractDelegatingTransport {
    }
 
    public void failIfNeeded(ReplicableCommand rpcCommand) {
-      if (failFilter.contains(getActualClass(rpcCommand))) {
+      if (failFilter.test(rpcCommand)) {
          throw new IllegalStateException("Induced failure!");
       }
    }
@@ -90,21 +112,21 @@ public class ControlledTransport extends AbstractDelegatingTransport {
       waitForReplicationLatch(rpcCommand, blockAfterFilter);
    }
 
-   protected void waitForReplicationLatch(ReplicableCommand rpcCommand, Set<Class> filter) {
-      Class cmdClass = getActualClass(rpcCommand);
-      if (!filter.contains(cmdClass)) {
+   protected void waitForReplicationLatch(ReplicableCommand rpcCommand, Predicate<ReplicableCommand> filter) {
+      if (!filter.test(rpcCommand)) {
+         log.tracef("Not blocking command %s", rpcCommand);
          return;
       }
 
       try {
          if (!blockingLatch.isOpened()) {
-            getLog().debugf("Replication trigger called, releasing any waiters for command to block.");
+            log.debugf("Replication trigger called, releasing any waiters for command to block.");
             blockingLatch.open();
          }
 
-         getLog().debugf("Replication trigger called, waiting for latch to open.");
-         replicationLatch.await(30, TimeUnit.SECONDS);
-         getLog().trace("Replication latch opened, continuing.");
+         log.debugf("Replication trigger called, waiting for latch to open.");
+         assertTrue(replicationLatch.await(30, TimeUnit.SECONDS));
+         log.trace("Replication latch opened, continuing.");
       } catch (Exception e) {
          throw new RuntimeException("Unexpected exception!", e);
       }
@@ -132,6 +154,13 @@ public class ControlledTransport extends AbstractDelegatingTransport {
    protected BackupResponse afterBackupRemotely(ReplicableCommand command, BackupResponse response) {
       waitAfter(command);
       return response;
+   }
+
+   private Predicate<ReplicableCommand> classListToFilter(Class[] filter) {
+      return cmd -> {
+         Class<?> actualClass = getActualClass(cmd);
+         return Stream.of(filter).anyMatch(clazz -> clazz.isAssignableFrom(actualClass));
+      };
    }
 
    private Class getActualClass(ReplicableCommand rpcCommand) {
--- a/core/src/test/java/org/infinispan/scattered/statetransfer/CoordinatorStopTest.java
+++ b/core/src/test/java/org/infinispan/scattered/statetransfer/CoordinatorStopTest.java
@@ -1,0 +1,216 @@
+package org.infinispan.scattered.statetransfer;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Arrays;
+import java.util.concurrent.BrokenBarrierException;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import org.infinispan.commands.read.GetKeyValueCommand;
+import org.infinispan.commands.remote.ClusteredGetCommand;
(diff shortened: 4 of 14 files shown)


core/src/main/java/org/infinispan/partitionhandling/impl/AvailabilityStrategy.java
```diff
@@ -32,7 +32,10 @@ static ConsistentHash readConsistentHash(CacheTopology topology, ConsistentHashF
          case NO_REBALANCE:
             return topology.getCurrentCH();
          case TRANSITORY:
-            return topology.getPendingCH();
+            // This is used to determine nodes that own the entries. In scattered cache (which uses transitory topology)
+            // the pendingCH is used for reading but the nodes in there are not guaranteed to have the data yet.
+            // CurrentCH should be safe - the nodes either have the data or the owner is unknown.
+            return topology.getCurrentCH();
```
danberindei: I meant this method to be a mirror of `CacheTopology.getReadConsistentHash()`, only without using the `unionCH` field (which is `null` on the `ClusterTopologyManagerImpl` side). Since this is breaking the symmetry, we should change the method name as well.

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
```diff
@@ -503,6 +503,11 @@ private void applyValues(Address address, List<Object> keys, Response response)
       for (int i = 0; i < keys.size(); ++i) {
          Object key = keys.get(i);
          InternalCacheValue icv = values[i];
+         if (icv == null) {
+            // The entry got lost in the meantime - this can happen when the cache is stopped concurrently to processing
+            // the GetAllCommand. We'll just avoid NPEs here: data is lost as > 1 nodes have left.
```
danberindei: Shouldn't we get an exception if the cache is being stopped?

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
```diff
@@ -0,0 +1,166 @@
...
+import org.infinispan.util.BlockingLocalTopologyManager;
+import org.infinispan.util.BlockingLocalTopologyManager.BlockedTopology;
+import org.infinispan.util.ControlledConsistentHashFactory;
+import org.infinispan.util.ControlledRpcManager;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.Test;
+
+@Test(groups = "functional", testName = "scattered.statetransfer.CoordStopTest")
+@CleanupAfterMethod
+public class CoordStopTest extends MultipleCacheManagersTest {
```
danberindei: We actually call it "coordinator" ;)

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
```diff
@@ -0,0 +1,166 @@
...
+      // Therefore we have to delay the view.
+      GlobalConfigurationBuilder gcb = new GlobalConfigurationBuilder();
+      gcb.transport().transport(new DelayedViewJGroupsTransport(viewLatch));
+      addClusterEnabledCacheManager(gcb, cb);
+
+      assertTrue(cache(0).getCacheManager().isCoordinator());
+      // start other caches
+      cache(1);
+      cache(2);
+      waitForClusterToForm();
```
danberindei: `waitForClusterToForm()` already starts the default caches.

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+      // segment 0 will be moved to cache(2). Since we've lost coord cache(1) now -> 0 and cache(2) -> 1
+      chf.setOwnerIndexes(new int[][] { {1}, {0}, {1} });
+   }
+
+   @AfterMethod
+   @Override
+   protected void clearContent() throws Throwable {
+      super.clearContent();
+      viewLatch = new CountDownLatch(1);
```
danberindei: Is this really needed?

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
```diff
@@ -0,0 +1,166 @@
...
+   }
+
+   // Reproducer for ISPN-9128
+   public void testCoordLeaves() throws InterruptedException, ExecutionException {
+      MagicKey key = new MagicKey(cache(0));
+      cache(1).put(key, "value");
+      int stableTopologyId = cache(1).getAdvancedCache().getDistributionManager().getCacheTopology().getTopologyId();
+      BlockingLocalTopologyManager ltm2 = BlockingLocalTopologyManager.replaceTopologyManager(manager(2), cache(2).getName());
+      ControlledRpcManager rpcManager2 = ControlledRpcManager.replaceRpcManager(cache(2));
+      rpcManager2.excludeCommands(cmd -> cmd instanceof StateResponseCommand, CoordStopTest::isCancelStateTransfer);
```
danberindei: IMO it would be simpler to handle the response commands explicitly, then you can guarantee there are no cancel commands.

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+      log.infof("Current members are: %s", cache(1).getCacheManager().getMembers());
+      assertTrue(cache(1).getCacheManager().isCoordinator());
+
+      // topology + 1 is the one that just omits the leaving node
+      BlockedTopology t1 = ltm2.expectTopologyUpdate(CacheTopology.Phase.NO_REBALANCE, stableTopologyId + 1);
+      assertEquals(2, t1.getCacheTopology().getActualMembers().size());
+      assertEquals(null, t1.getCacheTopology().getPendingCH());
+      for (Address member : t1.getCacheTopology().getMembers()) {
+         assertEquals("Topology: " + t1.getCacheTopology(), 1, t1.getCacheTopology().getCurrentCH().getSegmentsForOwner(member).size());
```
danberindei: Since you know the owner indexes, you can assert them directly (maybe even extract this to a method):

      assertEquals(emptyList(), t1.getCacheTopology().getCurrentCH().locateOwnersForSegment(0));
      assertEquals(singletonList(address(1)), t1.getCacheTopology().getCurrentCH().locateOwnersForSegment(1));
      assertEquals(singletonList(address(2)), t1.getCacheTopology().getCurrentCH().locateOwnersForSegment(2));

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
```diff
@@ -0,0 +1,166 @@
...
+      }
+      t1.unblock();
+
+      // topology + 2 has TRANSITORY phase and all segments have an owner in pendingCH
+      BlockedTopology t2 = ltm2.expectTopologyUpdate(CacheTopology.Phase.TRANSITORY, stableTopologyId + 2);
+      assertEquals(2, t2.getCacheTopology().getActualMembers().size());
+      assertNotNull(t2.getCacheTopology().getPendingCH());
+      for (int segment = 0; segment < t2.getCacheTopology().getPendingCH().getNumSegments(); ++segment) {
+         assertEquals("Topology: " + t2.getCacheTopology(), 1, t2.getCacheTopology().getPendingCH().locateOwnersForSegment(segment).size());
```
danberindei: Same here, this assert is too generic considering we know who the owners are.

danberindei:
I got a failure when running the whole test suite:
```
[ERROR] testCoordLeaves[bias=NEVER](org.infinispan.scattered.statetransfer.CoordStopTest)  Time elapsed: 0.053 s  <<< FAILURE!
java.lang.AssertionError: expected:<10> but was:<11>
	at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:245)
	at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:252)
	at org.infinispan.util.BlockingLocalTopologyManager.expectTopologyUpdate(BlockingLocalTopologyManager.java:102)
	at org.infinispan.scattered.statetransfer.CoordStopTest.testCoordLeaves(CoordStopTest.java:127)
```

When running in the IDE, I didn't get a failure, but I got some errors in the log:
```
09:13:51,898 ERROR (testng-test) [BlockingLocalTopologyManager] Stopped blocking topology updates, but there are 1 blocked updates in the queue: [Event{type=REBALANCE_START, topologyId=11, viewId=3}]
09:14:11,875 ERROR (remote-thread-test-NodeF-p18-t6) [BlockingLocalTopologyManager] Failed waiting for test to unblock REBALANCE_START 11 on test-NodeF-15557
java.util.concurrent.TimeoutException: null
	at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1771) ~[?:1.8.0_152]
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1915) ~[?:1.8.0_152]
	at org.infinispan.util.BlockingLocalTopologyManager$Event.awaitUnblock(BlockingLocalTopologyManager.java:247) ~[test-classes/:?]
	at org.infinispan.util.BlockingLocalTopologyManager.beforeHandleRebalance(BlockingLocalTopologyManager.java:204) ~[test-classes/:?]
	at org.infinispan.util.AbstractControlledLocalTopologyManager.handleRebalance(AbstractControlledLocalTopologyManager.java:64) ~[classes/:?]
```
[pr_rvansa_ISPN-9128_2018-05-22_CoordStopTest-infinispan-core.log.gz](https://github.com/infinispan/infinispan/files/2025249/pr_rvansa_ISPN-9128_2018-05-22_CoordStopTest-infinispan-core.log.gz)
[pr_rvansa_ISPN-9128_2018-05-22_ScatteredSyncStoreSharedTest-infinispan-core.log.gz](https://github.com/infinispan/infinispan/files/2025250/pr_rvansa_ISPN-9128_2018-05-22_ScatteredSyncStoreSharedTest-infinispan-core.log.gz)

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
rvansa: Not when the command enters interceptor stack and the cache is stopped only after that.

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
rvansa: What do you mean by handling explicitly? Overriding RpcManager on my own?

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
danberindei: If it can happen for our internal `GetAllCommand` it can probably happen for the application's `GetAllCommands` as well, so I'd change the code to throw an `IllegalLifecycleStateException` if possible.

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
danberindei: No, calling `ControlledRpcManager.expectCommand()`

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
rvansa: On the stopping receiver this topology check was ignored because the GetAllCommand has `SKIP_OWNERSHIP_CHECK` during the state transfer - this node was *not* the owner in previous topology (so it can't respond according to CH) but it has responded during the keys transfer as having the last version of the key. However as it's being stopped the data-container has been cleared concurrently and it fetches null data.
I agree that this could be improved by another type of signalization but this is out of scope of this PR.
On the originator, I don't see why you should throw an exception instead of processing the other data. Would you be ok with issuing a WARN log here?

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
rvansa: IIUC the ControlledRpcManager requires the commands to be send in order, and cancel commands depend on the timing - I would probably need to add additional synchronization to enforce them. Not that it wouldn't be possible, but I simply don't care about them.

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
danberindei: You're probably right that this is out of the scope of this PR, we could handle ISPN-9095 separately.

> However as it's being stopped the data-container has been cleared concurrently and it fetches null data.

When a cache shuts down while a remote command is running I would expect `BaseBlockingRunnable` to replace the command's response with a `CacheNotFoundResponse`. `DataContainer.clear()` has priority 999 and `PerCacheInboundInvocationHandler.stop()` has priority 10, so IMO the `GetAllCommand` should return either the proper value or a CNFR. BTW, non-scattered caches also rely on CNFR to know that a remote node is no longer an owner and the originator should ask another node, so if the CNFR logic is broken it's even more important to fix it.

> On the originator, I don't see why you should throw an exception instead of processing the other data. Would you be ok with issuing a WARN log here?

I'd fix the remote node to properly reply with a CNFR when it's shutting down, and log an error in `ScatteredStateConsumerImpl` if a value was invalid. If SSCI can retry that key and fetch the value from another node we could make it a warning instead.

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
danberindei: IMO the order of the cancel commands is pretty straightforward: if you intercept the response commands and you wait for their responses, all the segments will be received and there will be no need for a cancel command.

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
rvansa: Oh, I was wrong... I thought that the test is already stopping the cache after test failure but in fact the `TestingUtil` is clearing the data container directly. So this was a test problem.

In this case I don't understand why the exception happened in https://issues.jboss.org/browse/ISPN-9095

core/src/test/java/org/infinispan/scattered/statetransfer/CoordStopTest.java
rvansa: Personal question: are you Dan or Daniel on your birth certificate? :)

danberindei: Dan :smile: 

I know some of our test names are ridiculously long, but I don't think `CoordinatorStopTest` would be one of them ;)

rvansa:
Okay, the first failure is an oversight in the test (I thought that `ControlledRpcManager.stopBlocking` will send anything that's in the queue, but it does not).
The second failure looks like an instance of https://issues.jboss.org/browse/ISPN-9130 fixed in #5975

core/src/test/java/org/infinispan/util/ControlledRpcManager.java
```diff
@@ -60,7 +62,7 @@
    private final AtomicInteger count = new AtomicInteger(1);
 
    private volatile boolean stopped = false;
-   private final Set<Class<? extends ReplicableCommand>> excludedCommands =
+   private final Set<Predicate<ReplicableCommand>> excludedCommands =
```
danberindei: Please revert, I'd really like to keep the logic on the main thread

core/src/test/java/org/infinispan/scattered/statetransfer/CoordinatorStopTest.java
```diff
@@ -0,0 +1,168 @@
...
+      addClusterEnabledCacheManager(gcb, cb);
+
+      assertTrue(cache(0).getCacheManager().isCoordinator());
+      // start other caches
+      cache(1);
+      cache(2);
+      waitForClusterToForm();
+
+      // segment 0 will be moved to cache(2). Since we've lost coord cache(1) now -> 0 and cache(2) -> 1
+      chf.setOwnerIndexes(new int[][] { {1}, {0}, {1} });
```
danberindei: Nitpicking, but I'd like this in the test before the node is killed

core/src/test/java/org/infinispan/scattered/statetransfer/CoordinatorStopTest.java
```diff
@@ -0,0 +1,168 @@
...
+      assertNotNull(t2.getCacheTopology().getPendingCH());
+      assertOwners(t2, false, 0, address(2));
+      assertOwners(t2, false, 1, address(1));
+      assertOwners(t2, false, 2, address(2));
+      t2.unblock();
+
+      // Let the rebalance begin
+      rpcManager2.expectCommand(StateRequestCommand.class,
+            request -> assertEquals(StateRequestCommand.Type.CONFIRM_REVOKED_SEGMENTS, request.getType()))
+            .send().receiveAll();
```
danberindei: It's weird that I wrote `expectCommand` to only allow access to the command in the consumer, but I'm not going to argue with myself now :)

core/src/main/java/org/infinispan/scattered/impl/ScatteredStateConsumerImpl.java
danberindei: Ok, then do we still need this change?

rvansa:
I've kept the `icv == null` as the test code clearing containers was not fixed in this PR.

core/src/test/java/org/infinispan/scattered/statetransfer/CoordinatorStopTest.java
```diff
@@ -0,0 +1,168 @@
...
+      ControlledRpcManager.BlockedRequest cancelStateTransfer = rpcManager2.expectCommand(StateRequestCommand.class,
+            request -> assertEquals(StateRequestCommand.Type.CANCEL_STATE_TRANSFER, request.getType()));
+      cancelStateTransfer.sendWithoutResponses();
+
+      // Wait until topology + 3 is installed
+      assertTrue(topologyLatch.await(10, TimeUnit.SECONDS));
+
+      // unblock outdated rebalance
+      rpcManager2.stopBlocking();
+      keyTransferRequest.send().receiveAll();
```
danberindei: This is the kind of thing I meant in the chat... the fact that you call `stopBlocking()` before completing the blocked request makes me think that the test absolutely needs another command to be invoked remotely, but you'd rather not say what command :)

rvansa: I need to verify if this works (I didn't ran the test manually after last update and there was an oversight I had to fix); I've used similar construct in other test and seems it got broken as receiveAll schedules the continuation to the rpcManager's executor.

danberindei: I think I understand now. In my mind, the only alternative would have been to stop blocking at the end of the test, and stopping the executor wouldn't matter there because all the commands are already done.

danberindei:
Sorry @rvansa, now the test is failing:

https://ci.infinispan.org/job/Infinispan/job/PR-5989/5/testReport/junit/org.infinispan.scattered.statetransfer/CoordinatorStopTest/testCoordinatorLeaves_bias_NEVER_/

rvansa:
@danberindei I know, debugging ATM :-/ One of the issues is that the CH_UPDATE from old coord and REBALANCE_START from new coord can come in arbitrary order, I need to set the order.

Actually leaving coord sending CH updates is pretty PITA in many cases, this is one aspect of ST that should be eventually revised (I am not sure where we're ATM with the graceful leave).

danberindei:
Yes, this and many other aspects should be revised, but I haven't made much progress -- too many random failures :(

danberindei:
I integrated Will's data container changes and now there's a conflict as well :/

rvansa:
@danberindei I made the tests as deterministic as I could... let's hope it passes this time.

danberindei:
Ah, I forgot about this again... @rvansa, the test is still failing in Jenkins, but for some reason it doesn't appear upfront in the list of failed tests:

https://ci.infinispan.org/job/Infinispan/job/PR-5989/18/testReport/

rvansa:
No worries in this case; I've fixed some of the stuff but it's not 100% as the CI shows... I have been lucky enough to reproduce this after couple hundred invocations, so debugging...

rvansa:
@danberindei *sigh* one liner bug need 200+ lines of test. Now I've ran the test 2200 times without error, let's see if CI can give me one green run!

danberindei:
@rvansa I'd like us to write more unit tests and fewer functional tests for this kind of issue, but we need a lot of work in order to get there... With the huge lists of dependencies our components have I'm afraid a unit test may be even longer.

rvansa:
@danberindei Or we fully cover Infinispan in TLA+, verify that and just add a nice Java code generator... :)

rvansa:
@danberindei Seems the tests have passed now; only some RemoteCounter failure.

core/src/test/java/org/infinispan/remoting/transport/ControlledTransport.java
```diff
@@ -21,63 +25,81 @@
...
       replicationLatch.close();
       blockingLatch.close();
    }
 
    public void blockAfter(Class... filter) {
-      this.blockAfterFilter = new HashSet<>(Arrays.asList(filter));
+      blockAfter(classListToFilter(filter));
+   }
+
+   public void blockAfter(Predicate<ReplicableCommand> filter) {
```
danberindei: Without any further info on what caused the latest failures, I'd say it's this... the more complex the filter gets, the more likely it is to exclude the wrong command and crash the whole Rube Goldberg machine.

True, it's painful to enumerate all the commands that state transfer now needs like I'm trying to do with `MockTransport`/`MockLocalTopologyManager`, but IMO those can easily the sequence of commands can be extracted into reusable methods, and where the "business logic" is too complex for the test to predict the commands, then perhaps the logic should be simplified. I'm integrating the PR, but I trust you'll remember what you did to debug the test by the next time it fails :)

danberindei:
Integrated, thanks Radim!

danberindei:
@rvansa should I keep https://issues.jboss.org/browse/ISPN-9095 open?

rvansa:
I think you should close that; with the fix in the server won't emit the NPE and won't get stuck. Maybe there's an underlying problem, but it will manifest differently.

PR #5991, opened 2018-05-18
ISPN-9176 Hot Rod Lightweight Transaction (XaResource)

pruivo:
https://issues.jboss.org/browse/ISPN-9176

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
@@ -1,0 +1,360 @@
+package org.infinispan.client.hotrod.impl.transaction;
+
+import java.util.Collection;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.function.Function;
+
+import javax.transaction.RollbackException;
+import javax.transaction.SystemException;
+import javax.transaction.Transaction;
+import javax.transaction.xa.XAException;
+import javax.transaction.xa.XAResource;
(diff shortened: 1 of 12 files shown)


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
```diff
@@ -0,0 +1,370 @@
...
+         if (trace) {
+            log.tracef("Registering remote cache '%s' for transaction xid=%s", remoteCache.getName(), currentXid);
+         }
+         return new TransactionContext<>(remoteCache.keyMarshaller(), remoteCache.valueMarshaller(),
+               remoteCache.getOperationsFactory(), remoteCache.getName());
+      }
+
+      private void cleanup() {
+         registeredTransactions.remove(transaction);
+         registeredCaches.clear();
```
rvansa: Do we need to clear/null the fields? I'd expect that this adapter gets forgotten and gced...

pruivo: I think we do. 
Although we don't have recovery in place, the TM can keep this resource for a while if another resource may need some recovery.
Also, (I need to investigate), when recovery is in place, the instance may reincarnate as another transaction and we need to repopulate the `preparedCaches` and `currentXid`.

pruivo:
updated! and no failures!?

rvansa:
Poor CI, almost 2 hours running the tests to no avail... :) Integrating, thanks.

danberindei:
@pruivo do we have any tests w/ servers joining/crashing? We could add some split partition tests, if you're missing test failures ;)

danberindei:
@pruivo Jokes aside, I'd like to see some tests in the client as well. Not at the level we have in embedded, after all lots of embedded PH tests are quarantined because they were unstable, but still I think *some* coverage would be useful.

PR #5999, opened 2018-05-21
Updating all Hibernate libraries, related integration tests

Sanne:
- https://issues.jboss.org/browse/ISPN-9192


Sanne:
Right, I'm aware of those.

With WF coincidentally happening to have the same versions as we need it needs special treatment - but it's test code only so we'll notice at the next version upgrade.

anistor:
Merged. Thanks  @Sanne !

PR #5951, opened 2018-05-02
ISPN-7420 Hot Rod enhancements for transcoding

gustavonalle:
https://issues.jboss.org/browse/ISPN-7420

Adds similar capabilities to the REST endpoint when it comes to send and receive data in different formats given by a ```MediaType```. 

The API changes can be seen [here](https://github.com/infinispan/infinispan/compare/master...gustavonalle:HotRod-Transcoding?expand=1#diff-7de81536b3e502192f30f16d1e833eaeR108).

PR #6003, opened 2018-05-23
ISPN-9201 Do not forget to apply Hot Rod flags to the cache before pe…

tristantarrant:
https://issues.jboss.org/browse/ISPN-9201


tristantarrant:
@gustavonalle this was introduced by https://github.com/infinispan/infinispan/pull/5951

gustavonalle:
> @gustavonalle this was introduced by #5951 

ok, strange though that if it was applying the client flags, it affect only 1 test in the whole suite?

tristantarrant:
Indeed. Apparently we do not have tests that verify flags on the first op
on a cache.

gustavonalle:
@tristantarrant right, makes sense, I though it was never applying the flags :)

gustavonalle:

PR #6005, opened 2018-05-23
ISPN-9204 Single node JgroupsTransport invoke* methods that take a

wburns:
collection leak memory

* Add check to only send command if not already complete

https://issues.jboss.org/browse/ISPN-9204

### DIFF
--- a/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
+++ b/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
@@ -837,6 +837,10 @@ public class JGroupsTransport implements Transport {
       Address excludedTarget = deliverOrder == DeliverOrder.TOTAL ? null : getAddress();
       MultiTargetRequest<T> request =
             new MultiTargetRequest<>(collector, requestId, requests, targets, excludedTarget);
+      // Request may be completed due to exclusion of target nodes, so only send it if it isn't complete
+      if (request.isDone()) {
+         return request;
+      }
       try {
          addRequest(request);
          boolean checkView = request.onNewView(clusterView.getMembersSet());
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
```diff
@@ -837,16 +837,19 @@ public final void checkTotalOrderSupported() {
...
-         boolean checkView = request.onNewView(clusterView.getMembersSet());
-         sendCommand(targets, command, requestId, deliverOrder, isRsvpCommand(command), checkView);
-      } catch (Throwable t) {
-         request.cancel(true);
-         throw t;
-      }
-      if (timeout > 0) {
-         request.setTimeout(timeoutExecutor, timeout, unit);
+      // Request may be completed due to exclusion of target nodes, so only send it if it isn't complete
+      if (!request.isDone()) {
```
danberindei: I would suggest `if (request.isDone() return request;` to avoid indenting the rest of the method.

core/src/test/java/org/infinispan/remoting/rpc/RpcManagerTest.java
```diff
@@ -207,7 +207,7 @@ public void testInvokeCommands() throws Exception {
       RpcManager rpcManager0 = cache(0).getAdvancedCache().getRpcManager();
 
       Exceptions.expectException(IllegalArgumentException.class, () -> {
-         rpcManager0.invokeCommands(Arrays.asList(address(0)), a -> command, MapResponseCollector.validOnly(),
+         rpcManager0.invokeCommands(Arrays.asList(address(1)), a -> command, MapResponseCollector.validOnly(),
```
danberindei: The topology id check is done in `RpcManagerImpl`, so I think this change isn't necessary.

core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsTransport.java
wburns: Sure.

core/src/test/java/org/infinispan/remoting/rpc/RpcManagerTest.java
wburns: Hrmm it failed without this change.

danberindei: I see now, the validation is done by wrapping the command generator, so it actually runs inside `JGroupsTransport.invokeCommands`.

PR #6011, opened 2018-05-25
ISPN-9178 EntityCollectionInvalidationTest tests failing randomly

galderz:
https://issues.jboss.org/browse/ISPN-9178

### DIFF
--- a/hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/cluster/EntityCollectionInvalidationTest.java
+++ b/hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/cluster/EntityCollectionInvalidationTest.java
@@ -44,8 +44,9 @@ import org.infinispan.notifications.cachelistener.annotation.CacheEntryVisited;
 import org.infinispan.notifications.cachelistener.event.CacheEntryVisitedEvent;
 import org.infinispan.test.hibernate.cache.commons.util.TestRegionFactory;
 import org.infinispan.util.ControlledTimeService;
-import org.junit.Ignore;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TestName;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
@@ -72,6 +73,10 @@ public class EntityCollectionInvalidationTest extends DualNodeTest {
 	private MyListener localListener, remoteListener;
 	private SessionFactory localFactory, remoteFactory;
 	private final ControlledTimeService timeService = new ControlledTimeService();
+
+   @Rule
+   public TestName name = new TestName();
+
 
 	@Override
 	public List<Object[]> getParameters() {
@@ -128,8 +133,9 @@ public class EntityCollectionInvalidationTest extends DualNodeTest {
 	}
 
 	@Test
-	@Ignore("ISPN-9175")
 	public void testAll() throws Exception {
+      log.infof(name.getMethodName());
+
 		assertEmptyCaches();
 		assertTrue( remoteListener.isEmpty() );
 		assertTrue( localListener.isEmpty() );
@@ -144,6 +150,14 @@ public class EntityCollectionInvalidationTest extends DualNodeTest {
 		// and therefore the putFromLoad in getCustomer could be considered stale if executed too soon.
 		timeService.advance(1);
 
+      CountDownLatch remoteCollectionLoadLatch = null;
+      if (!cacheMode.isInvalidation()) {
+         remoteCollectionLoadLatch = new CountDownLatch(1);
+         ExpectingInterceptor.get(remoteCollectionCache)
+            .when((ctx, cmd) -> cmd instanceof PutKeyValueCommand)
+            .countDown(remoteCollectionLoadLatch);
+      }
+
 		log.debug( "Find node 0" );
 		// This actually brings the collection into the cache
 		getCustomer( ids.customerId, localFactory );
@@ -157,6 +171,12 @@ public class EntityCollectionInvalidationTest extends DualNodeTest {
 		// Check the read came from the cache
 		log.debug( "Check cache 0" );
 		assertLoadedFromCache( localListener, ids.customerId, ids.contactIds );
+
+      if (remoteCollectionLoadLatch != null) {
+         log.debug( "Wait for remote collection put from load to complete" );
+         assertTrue(remoteCollectionLoadLatch.await(2, TimeUnit.SECONDS));
+         ExpectingInterceptor.cleanup(remoteCollectionCache);
+      }
 
 		log.debug( "Find node 1" );
 		// This actually brings the collection into the cache since invalidation is in use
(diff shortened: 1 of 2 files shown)


hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/cluster/EntityCollectionInvalidationTest.java
```diff
@@ -144,6 +150,14 @@ public void testAll() throws Exception {
 		// and therefore the putFromLoad in getCustomer could be considered stale if executed too soon.
 		timeService.advance(1);
 
+      CountDownLatch remoteCollectionLoadLatch = null;
+      if (!cacheMode.isInvalidation() && (mode.equals("nonstrict") || mode.equals("read-write"))) {
```
rvansa: shouldn't a non-invalidation mode imply either nonstrict or read-write?

hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/cluster/EntityCollectionInvalidationTest.java
```diff
@@ -158,6 +172,12 @@ public void testAll() throws Exception {
 		log.debug( "Check cache 0" );
 		assertLoadedFromCache( localListener, ids.customerId, ids.contactIds );
 
+      if (remoteCollectionLoadLatch != null) {
+         log.debug( "Wait for remote collection put from load to complete" );
```
rvansa: The remote cache should be updated from the next `getCustomer( ids.customerId, remoteFactory );` call; we don't have to rely on the putFromLoad from localCache.
Uncontended putFromLoad should update data container on the node where it is invoked before returning, so that further queries from the same thread are already handled from the cache.
Therefore I don't see why this waiting is needed (documenting in source would be preferred) unless we use DIST mode and remote node is not an owner.

hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/cluster/EntityCollectionInvalidationTest.java
galderz: Read only?

rvansa: That's not tested in this test (see `getParameters`). Besides, putFromLoads happen in RO exactly the same way.

hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/cluster/EntityCollectionInvalidationTest.java
galderz: > The remote cache should be updated from the next getCustomer( ids.customerId, remoteFactory ); call; we don't have to rely on the putFromLoad from localCache.

There's nothing that guarantees that.

The local cache sends a remote async PFER to remote, and remote executes a local PFER, one of them will win but there's no guarantee which one. The one coming from local factory could be slow, keep the locks and the local put on remote factory could silently fail with 0ms lock timeout.

So, which one will win? The remote PFER coming from local factory (ASYNC), or the local PFER coming from remote factory (LOCAL)? If the latter wins, no need for the latch. If the former wins we need the latch because it's ASYNC.

rvansa: @galderz Ouch, I see the problem now. And I am afraid that the problem is in the code, not in the test. Because I still think it's a good idea that after you read data from DB and there's no conflicting invalidation it ends up in the cache. The fact that other node read and putFromLoaded that data, too, shouldn't prevent you from being able to read cached data (regardless who actually updated them).

So, my first thought was delaying the write until the lock gets released, but canceling the putFromLoad afterwards. #5954 Adds `LockingMode.TRY_LOCK` as an equivalent of `Flag.ZERO_LOCK_ACQUISITION_TIMEOUT`, so we could add another one that would have such behaviour (implementation wouldn't be complex, naming would). Compared to `TRY_LOCK` it would have a very narrow use-case, though.
The simplest solution is dropping the TRY_LOCK/ZAT semantics completely.

I understand that this goes against the policy of 'never blocking on reads' but given that we don't hold the lock during RPC anymore, it's affordable.

galderz: > Because I still think it's a good idea that after you read data from DB and there's no conflicting invalidation it ends up in the cache. The fact that other node read and putFromLoaded that data, too, shouldn't prevent you from being able to read cached data (regardless who actually updated them).

Hmmm, why would you want to do that? Do you have any specific use cases where that must happen? With the current code there's an eventually part to that. So, you will read your writes, eventually ;). That might be find for a cache under concurrent put from loads.

I know it might feel a little bit non obvious, but if we get performance gains from it, I'd be happy with it. We could educate the user too.

Btw, now that you understand the problem, this wasn't failing 3-6 months ago. Are you aware of any changes that have made this surface?

> So, my first thought was delaying the write until the lock gets released, but canceling the putFromLoad afterwards. #5954 Adds LockingMode.TRY_LOCK as an equivalent of Flag.ZERO_LOCK_ACQUISITION_TIMEOUT, so we could add another one that would have such behaviour (implementation wouldn't be complex, naming would). Compared to TRY_LOCK it would have a very narrow use-case, though. The simplest solution is dropping the TRY_LOCK/ZAT semantics completely. I understand that this goes against the policy of 'never blocking on reads' but given that we don't hold the lock during RPC anymore, it's affordable.

I don't object to that, as long as the performance does not degraded. I think it would since gets would need to acquire locks now :|

rvansa: > Hmmm, why would you want to do that? Do you have any specific use cases where that must happen? With the current code there's an eventually part to that. So, you will read your writes, eventually ;). That might be find for a cache under concurrent put from loads.

There have been issues when this was assumed in the tests (both in our testsuite and in WF) and the putFromLoad didn't take place because of tx start time matching last invalidation time. That's a bit different because there has been no concurrency, and it's just a test after all, but...

OK, I understand that it's a weak ground to drop the non-blocking semantics or complicating code (especialy for including a special locking mode).

> Btw, now that you understand the problem, this wasn't failing 3-6 months ago. Are you aware of any changes that have made this surface?

Nope, this hasn't changed. I have some changes in #5954 but these are not in yet. You know how it goes with races...

> I don't object to that, as long as the performance does not degraded. I think it would definitely affect perf since gets would need to acquire locks now though :(

Perf impact of changes in 2LC is often non-obvious - I've found myself guessing wrong couple of times. For #5964 I was running some perf tests locally but we should add 2LC to nightly regression suite.

galderz:
Yeah, let me do so...

galderz:
Actually, let me add documentation first because this is something that might surprise users. I'll add that tomorrow.

galderz:
@rvansa Added docu and squashed the commits, can you have a look?

rvansa:
Thanks, Galder! The discussion almost got longer than the PR itself :)

PR #6019, opened 2018-05-28
ISPN-9216 Importing @Immutable from nashorn instead of net.jcip break…

Sanne:
…s CounterValue on JDK10

 - https://issues.jboss.org/browse/ISPN-9216


tristantarrant:
Merged, thanks

PR #6024, opened 2018-05-30
ISPN-9224 Reduce ExceptionEvictionTest combinations

wburns:
* Comment out OFFHEAP/BINARY cluster mode options

https://issues.jboss.org/browse/ISPN-9224

### DIFF
--- a/core/src/test/java/org/infinispan/eviction/impl/ExceptionEvictionTest.java
+++ b/core/src/test/java/org/infinispan/eviction/impl/ExceptionEvictionTest.java
@@ -152,6 +152,7 @@ public class ExceptionEvictionTest extends MultipleCacheManagersTest {
             memoryConfigurationBuilder.evictionType(EvictionType.MEMORY).size(convertAmountForStorage(SIZE) + 16);
             break;
          case OFF_HEAP:
+            memoryConfigurationBuilder.addressCount(1 << 7);
             // Each entry takes up 63 bytes total for our tests, however tests that add expiration require 16 more
             memoryConfigurationBuilder.evictionType(EvictionType.MEMORY).size(24 +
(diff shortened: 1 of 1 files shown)


danberindei:
@wburns, I've looked a little more into this and I think it would be better to keep all the configurations for the base test if we split off the expiration, state transfer, and overflow on scale down test methods into separate classes that only deal with OBJECT storage (or BINARY, or OFF_HEAP if you think it would be more interesting).

There's also a small inefficiency in `clearContent()`, which invokes `cache.clear()` on all the caches but without `CACHE_MODE_LOCAL`. Just setting the interceptor field saved me almost 1 minute (w/ trace logging enabled). Weird, because it's 1 call instead of 3, calling just `cache(0).clear()` got me almost exactly in the middle, with 30s saved vs. master.

So even without the OffHeapMemory logging, the extra data container iterations seem quite expensive. I wonder if we could make it faster, maybe we could implement `forEach()` in `OffHeapDataContainer` to avoid the stream overhead?

We could also skip the data container iteration in `ClearCommand` and `AbstractClusterDependentLogic` if there are no remove listeners, but I'm not sure how I'd like a remove-specific method in `CacheNotifier`, I'd like a more generic solution there for interceptors to also skip loading the previous value when possible. Maybe `CacheImpl` could set the `SKIP_CACHE_LISTENERS` flag on the command when there are no listeners registered at command start? Or maybe we'll even change `clear()` to not notify listeners at all...

danberindei:
Forgot another option, maybe we could use a smaller `addressCount` in tests and/or when we know the size limit?

wburns:
Reducing the address count to say `1 << 7` cut the time to a third. So I can just change that.

wburns:
Updated.

tristantarrant:
Merged

PR #6026, opened 2018-05-30
ISPN-9225 ANTLR 3.5.2

anistor:
https://issues.jboss.org/browse/ISPN-9225

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -69,7 +69,7 @@
       <version.java>1.8</version.java>
 
       <!-- Dependency versions duplicated in the parent POM -->
-      <version.antlr>3.4</version.antlr>
+      <version.antlr>3.5.2</version.antlr>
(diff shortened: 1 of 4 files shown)


anistor:
Compilation failed with 3.5.2, but I found out the cause. Seems that the 3.4 we were using before did not generate code using generics. An then for whatever reason 3.5.2 generates code with List<Object> instead of List<Tree>. I diffed the parser generated with 3.4 and 3.5.2 and it seems these + some exception handling are the only differences so I hope we are not breaking anything (more than it already is) while upgrading. This upgrade seems logical also because we previoulsly processed the grammar using 3.4 but used the 3.5.2 runtime :) Not sure how that happened but I think it's safer to have both 3.5.2.

anistor:
Updated commit message and rebased. Ready now.

tristantarrant:
Merged, thanks

PR #6028, opened 2018-05-31
ISPN-9218 Wrap InterruptedException in HotRodClientException

galderz:
* When catching InterruptedException, since we cannot send the
exception as is, also restore its interrupt status.


tristantarrant:
Merged, thanks

PR #6030, opened 2018-06-01
ISPN-9194 EntityRegionAccessStrategyTest.testRemoveAll failing randomly

galderz:
https://issues.jboss.org/browse/ISPN-9194

Please find details about the actual issue in the JIRA. The problem was that a `FutureUpdate` put was leaking from one test to other, resulting in the latches in the latter test passing when they shouldn't.

The fix contains:

* Wait for the `FutureUpdate` put at the end of testRemove.
* When checking for put latches, make sure the key is the expected one.
* Logging to make it easier to diagnose problems in the future.


rvansa:
@galderz OK, could you squash?

galderz:
@rvansa Done

rvansa:
Merged, thanks!

PR #6032, opened 2018-06-03
ISPN-9241 NPE when creating caches via HotRod on non-clustered servers

gustavonalle:
https://issues.jboss.org/browse/ISPN-9241


anistor:
Code looks good but CI says no. Let's spin the wheel again.

PR #6033, opened 2018-06-04
ISPN-9151 Pass in ClientListener overrides as parameters

galderz:
* Overriding or alternative to annotated POJO values.

### DIFF
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerParamsOverrideTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerParamsOverrideTest.java
@@ -1,0 +1,117 @@
+package org.infinispan.client.hotrod.event;
+
+import io.netty.buffer.ByteBuf;
+import org.infinispan.client.hotrod.RemoteCache;
+import org.infinispan.client.hotrod.RemoteCacheManager;
+import org.infinispan.client.hotrod.annotation.ClientListener;
+import org.infinispan.client.hotrod.exceptions.HotRodClientException;
+import org.infinispan.client.hotrod.impl.RemoteCacheImpl;
+import org.infinispan.client.hotrod.impl.operations.OperationsFactory;
+import org.infinispan.client.hotrod.impl.protocol.Codec28;
+import org.infinispan.client.hotrod.test.SingleHotRodServerTest;
+import org.testng.annotations.Test;
+
+import java.lang.annotation.Annotation;
+import java.util.Deque;
+import java.util.LinkedList;
+
+import static org.infinispan.test.TestingUtil.replaceField;
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertTrue;
+
+@Test(groups = "functional", testName = "client.hotrod.event.ClientListenerParamsOverrideTest")
+public class ClientListenerParamsOverrideTest extends SingleHotRodServerTest {
+
+   private TestCodec testCodec;
+
+   @Override
+   protected RemoteCacheManager getRemoteCacheManager() {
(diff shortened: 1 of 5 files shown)


client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerParamsOverrideTest.java
```diff
@@ -0,0 +1,117 @@
...
+         assertEquals("b", listener.converterFactoryName());
+         assertTrue(listener.useRawData());
+         assertTrue(listener.includeCurrentState());
+      }
+   }
+
+   public void testClientListenerParameter() {
+      final RemoteCache<?, ?> cache = remoteCacheManager.getCache();
+      try {
+         ClientListener listenerParams = new ClientListener() {
```
anistor: This does not implement hashcode and equals as per java.lang.annotation.Annotation contract. Especially hashcode.
Sure that does not harm us but it just leaves me the impression we are stretching this annotation thing more than we should.

anistor:
Galder, I think this area of client listeners API needs a bit of thoughtful refactoring because it started to become a bit unfriendly right after the addition of factory and converter params years ago. Before adding more to it I would reconsider a bit. I know this fix solves the problem you were initially facing, but let's start to think a bit about the future of the api.

I was never a big fan of using annotations (which are very static by nature) for things that need dynamic configuration. I'm not against annotations were they fit well but this does not seem to always be the case here. So this fix-up with more annotations looks even more odd to me. Yes you can be clever and instantiate them but it makes the whole thing verbose and feels a bit abusive. And besides being verbose, it is also complex. If you really want to stick to the annotation contract then you need to implement hashcode properly using the predefined algorithm which is not Object.hashCode (https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/Annotation.html#hashCode--). Sure, we do not use hashCode, but that's just another sign we are stretching things :)

When we have too many parameters to pass in a method call, and we do have a ton of them for listener registration, many being optional, we can use the good old tried and loved pattern called fluent builder dsl.
```
cache.createClientListener(myListenerInstance).
     .withFilterFactory("no-more-spam-please")
     .withFilterFactoryParams(new Object[] {"foo", "bar", 3.14})
     .withConverter("give-me-the-bytes-leave-out-the-floats")
     .withConverterParams(new Object[] {200})
     .includeCurrentState().
     .addListener();
```

I see this approach more natural and we can keep the current addClientListener method as deprecated and superseded by this. And we can also continue to use the filter metadata that was already added to the listener instance in the form of an annotation as default metadata (if present) that can be overridden by calling the methods in the registration dsl.

anistor:
This might also be an opportunity to add named parameters in the form of a map besides the existing positional paramereter(Object[])

danberindei:
:+1: for dynamic listeners without annotations @anistor, both for the client and for embedded caches. The fluent example looks really nice.

:-1: to mix annotations and external metadata though

wburns:
I have to admit I am would love to move to a more fluent approach that @anistor showed here. There are already so many different methods on the `Listenable`, `FilteringListenable` and `ClassLoaderAwareListenable` classes (embedded use case) that it is already quite out of hand.

galderz:
https://issues.jboss.org/browse/ISPN-9312

anistor:
Good, but then can we mark this new API method as 'experimental' and javadoc it is a temporary solution and subject to quick removal when ISPN-9312 is done? I'd like to not have to keep around duplicate stuff, we already have a bunch.

anistor:
BTW, is the OpenWhisk thing that urgent that we need to rush this for 9.3.0.Final?

galderz:
Never mind guys, found a workaround https://gist.github.com/galderz/eb5994eea7d253a2e0fc031b1929b174

🤦‍♂️

PR #6034, opened 2018-06-04
ISPN-9251 Removed System.out calls

slaskawi:
https://issues.jboss.org/browse/ISPN-9251

I tried to remove only the `System.out` calls which seemed unnecessary preserving all of them from Perf tests and some of them from the Test Utils.


tasks/manager/src/test/java/org/infinispan/tasks/SecurityTaskManagerTest.java
```diff
@@ -91,7 +91,6 @@ public Void run() {
 
     @Test(dataProvider = "principleProvider")
```
anistor: Hey, ```principleProvider``` seems misspelled :)

galderz: Not relevant to this PR. Create a jira if needed ;)

anistor: No, I'll just fix it myself and push it when nobody's looking ;)

PR #6035, opened 2018-06-04
ISPN-9254 Single Port documentation

slaskawi:
https://issues.jboss.org/browse/ISPN-9254

### DIFF
--- a/documentation/src/main/asciidoc/server_guide/server_guide.asciidoc
+++ b/documentation/src/main/asciidoc/server_guide/server_guide.asciidoc
@@ -186,3 +186,34 @@ The next step is to add a new `router-connector` endpoint and configure how othe
 
 With the following configuration, Hot Rod clients will access `hotrod-1` connector when using SNI Host Name "hotrod-1".
 REST clients will need to use the following URL to access "rest-1" connector - `https://<server_ip>:<server_port>/rest/rest-1`.
+
+== Single Port
+
+Single Port features allows to expose multiple protocols over the same TCP port.
+This approach is very convenient in Cloud Environments such as OpenShift or Kubernetes where the application
+(a Pod) is hidden behind the frontend proxy (a Route or an Ingress).
+
+For protocol switching, Infinispan Server uses HTTP/1.1 Upgrade header or TLS/ALPN.
+
+NOTE: The initial implementation supports only HTTP/1.1, HTTP/2 and Hot Rod protocols.
+
+==== Single Port router
+
+Internally, Single Port reuses router component, therefore the configuration looks very similar to
+the Multi-tenancy.
+
+[source,xml]
+----
+
+<router-connector single-port-socket-binding="rest">
+    <single-port security-realm="SSLRealm1">
+        <hotrod name="hotrod" />
+        <rest name="rest" />
+    </single-port>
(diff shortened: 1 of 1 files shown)


galderz:
Documentation looks good but it's missing some examples. E.g. how does a Java client connect to the single port? Does it need to pass any special configuration parameters to the RemoteCacheManager, or does it just connect to the exposed port (having configured the server side)? What about for a Node.js/C/C++/C# client (cc @rigazilla)?

Also, any reason why this is just not enabled by default?

slaskawi:
@galderz This is server only implementation at the moment so there's no client that can use it (apart from HTTP clients that can use both HTTP/1.1 and HTTP/2; but that's nothing new, we had this type of functionality before). 

The next thing on my list is to create a multi-protocol Infinispan client. Such a client could use any of 3 protocols: HTTP/1.1, HTTP/2 or Hot Rod. But for this work I needed a server-side Single Port to be in. 

At this point, I would like to ask you opinion - would you prefer to design such a client in Infinispan repo or as a separate project? I would vote for the latter for the sake of simplicity. WDYT?

//cc @tristantarrant

tristantarrant:
I don't think it makes much sense to have a multiprotocol client (unless you can justify it), but definitely add the HTTPUpgrade part to the hot rod client

slaskawi:
@tristantarrant +1, that needs to be added. But you need to be aware of the limitation. If an upgrade (via HTTP/1.1) or negotiation (via TLS/ALPN) fails, the Hot Rod client will throw an exception and refuse to connect. Even though it may seems obvious but I think it's worth to say it out laud.

The multi-protocol client is already on my research path and I think it's worth exploring. I hope I will find some interesting use cases, like using unencrypted HTTP connection for handing one cache and encrypted Hot Rod for transferring sensitive data. It's totally new topic to me and my head is full of ideas... However, I believe that should be done in a separate repo (answering my own question).

Nevertheless guys, that's a bit off-topic discussion considering the scope of this PR. Is it good to go?

galderz:
Ok, if the only way to take advantage of this is using an HTTP client, could you show an example? With curl or something? I still find it vague to just show something that you can enable on the server but have no way of seeing that it "works".

gustavonalle:
+1, I was wondering what are the advantages of using a single port for all protocols over using a different port per protocol. 
Isn't it worse from the security point of view, i.e., once the single port is exposed, all protocols are exposed as well, so one cannot allow only Hot Rod, for example? Or is there any mechanism to filter that?

gustavonalle:
Answered my own question:
```
This approach is very convenient in Cloud Environments such 
as OpenShift or Kubernetes where the application (a Pod) is 
hidden behind the frontend proxy (a Route or an Ingress).
```

slaskawi:
> Ok, if the only way to take advantage of this is using an HTTP client, could you show an example? With curl or something? I still find it vague to just show something that you can enable on the server but have no way of seeing that it "works".

@galderz Perhaps it's worth to wait a while and add an example together with a Single Port Hot Rod Client implementation? I plan to work on it very shortly.

slaskawi:
Closing in favor of https://github.com/infinispan/infinispan/pull/6133

PR #3384, opened 2015-04-16
ISPN-5370 Make clear() non-transactional and lock free

pruivo:
* Forget to adapt the Cross-Site code

https://issues.jboss.org/browse/ISPN-5370

PR #3370, opened 2015-04-10
ISPN-5370 Make clear() non-transactional and lock free

pruivo:
* no concurrent operation
* no locks acquired
* no transaction

Also performed some cleanups in the touched classes

https://issues.jboss.org/browse/ISPN-5370

PR #6036, opened 2018-06-04
ISPN-9155 clear() should not send remove notifications

galderz:
https://issues.jboss.org/browse/ISPN-9155

Wait for CI to see what I've broken ;)

### DIFF
--- a/core/src/main/java/org/infinispan/Cache.java
+++ b/core/src/main/java/org/infinispan/Cache.java
@@ -4,6 +4,7 @@ import java.util.Collection;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
@@ -334,6 +335,10 @@ public interface Cache<K, V> extends BasicCache<K, V>, BatchingCache, FilteringL
     * concurrently.
     * <p/>
     * If the cache is transactional, it will not interact with the transaction.
+    *
+    * Starting with version 9.3, calling this method does not result if cache entry removed events being fired.
+    * If you rely on cache entry removed events
+    * , use {@link CacheStream#forEach(BiConsumer)} to individually remove each entry and get remove notifications.
     */
    @Override
    void clear();
--- a/documentation/src/main/asciidoc/user_guide/cache_api.adoc
+++ b/documentation/src/main/asciidoc/user_guide/cache_api.adoc
@@ -65,6 +65,19 @@ if (cachedPerson == null) {
 
 Please note that link:{javadocroot}/org/infinispan/Cache.html#putForExternalRead-K-V-[putForExternalRead] should never be used as a mechanism to update the cache with a new Person instance originating from application execution (i.e. from a transaction that modifies a Person's address). When updating cached values, please use the standard link:{jdkdocroot}/java/util/Map.html#put-K-V-[put] operation, otherwise the possibility of caching corrupt data is likely.
 
+==== Clearing the cache
+
+The fastest way to clear a cache is to call
+link:{javadocroot}/org/infinispan/Cache.html#clear--[clear] method.
+
+IMPORTANT: Starting with version 9.3, calling `clear` does no longer emit remove notifications for each removed key.
+If relying on removed events being fired upon calling `clear`, it is recommended to call `cache.keySet().stream().forEach(Cache::remove)`.
+Removing all entries in this way is optimized so that with distributed caches, removes happen in parallel.
+Further optimization can be achieved by calling `disableRehashAware` on the stream.
+However, when this option is enabled and concurrent writes are happening at the same time as cluster topology changes,
+a concurrent write might override the remove resulting in a non-empty cache.
+
+
 === The AdvancedCache interface
 In addition to the simple Cache interface, {brandname} offers an link:{javadocroot}/org/infinispan/AdvancedCache.html[AdvancedCache] interface, geared towards extension authors.  The AdvancedCache offers the ability to inject custom interceptors, access certain internal components and to apply flags to alter the default behavior of certain cache methods.  The following code snippet depicts how an AdvancedCache can be obtained:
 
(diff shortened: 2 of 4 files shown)


tristantarrant:
Do we want to add the complementary removeAll() that does keep these semantics ?

wburns:
We should mention this in the upgrading guide as well. https://github.com/infinispan/infinispan/blob/master/documentation/src/main/asciidoc/upgrading/upgrading.asciidoc

documentation/src/main/asciidoc/user_guide/cache_api.adoc
```diff
@@ -65,6 +65,19 @@ if (cachedPerson == null) {
 
 Please note that link:{javadocroot}/org/infinispan/Cache.html#putForExternalRead-K-V-[putForExternalRead] should never be used as a mechanism to update the cache with a new Person instance originating from application execution (i.e. from a transaction that modifies a Person's address). When updating cached values, please use the standard link:{jdkdocroot}/java/util/Map.html#put-K-V-[put] operation, otherwise the possibility of caching corrupt data is likely.
 
+==== Clearing the cache
+
+The fastest way to clear a cache is to call
+link:{javadocroot}/org/infinispan/Cache.html#clear--[clear] method.
```
danberindei: Should we recommend `clear()` here? The javadoc still says "This should never be invoked in production"

documentation/src/main/asciidoc/user_guide/cache_api.adoc
```diff
@@ -65,6 +65,19 @@ if (cachedPerson == null) {
...
+
+IMPORTANT: Starting with version 9.3, calling `clear` does no longer emit remove notifications for each removed key.
+If relying on removed events being fired upon calling `clear`, it is recommended to call `cache.keySet().stream().forEach(Cache::remove)`.
+Removing all entries in this way is optimized so that with distributed caches, removes happen in parallel.
+Further optimization can be achieved by calling `disableRehashAware` on the stream.
```
danberindei: I wouldn't say "optimized" here, since it's always going to be much slower than the old `clear()`

core/src/main/java/org/infinispan/Cache.java
```diff
@@ -334,6 +335,10 @@
     * concurrently.
     * <p/>
     * If the cache is transactional, it will not interact with the transaction.
+    *
+    * Starting with version 9.3, calling this method does not result if cache entry removed events being fired.
+    * If you rely on cache entry removed events
+    * , use {@link CacheStream#forEach(BiConsumer)} to individually remove each entry and get remove notifications.
```
danberindei: We should document this in the `CacheEntryRemoved` javadoc as well, and check that we don't implicitly rely on the listeners being invoked internally (e.g. `MigrationTask`, global state, L2, JCache/Spring adapters)

danberindei:
@galderz as I mentioned in JIRA, we already have `SKIP_LISTENER_NOTIFICATION`, so I don't see a really good reason for this.

At the very least we should make sure we never use clear+listeners internally, and add extra warnings where we add listeners internally and the user can call `clear()` directly without necessarily knowing about our internal listeners.

wburns:
My guess is that the issue is that `RemoteCache` has no way to ignore listeners on a clear invocation. Since clear is a lot slower when any cluster listener is installed (including remote listener).

danberindei:
Speaking of `RemoteCache`, how would the near cache deal with `cache.clear()` from an embedded node or from another client?

anistor:
My vote for this: -1. I've added my concerns in the jira comments.

gustavonalle:
Just an idea: why not keep the current method "as is" to avoid breaking clients, and provide another one named ```truncate```, ```purge``` or whatever, documenting on the javadoc that it is a "fast" clean?

danberindei:
`truncate()` is better because it doesn't affect existing callers, but I'm still not convinced that we need a new method that does the same thing as `clear()+SKIP_LISTENER_NOTIFICATION`. Remember, the `clear()` javadoc still recommends not using it in production:

     * <b>Note:</b> This should never be invoked in production unless you can guarantee no other invocations are ran
     * concurrently.

gustavonalle:
@danberindei usability/syntactic sugar? ```truncate()```  would be implemented internally as ```clear()+SKIP_LISTENER_NOTIFICATION```

galderz:
I'm not totally sure I understand why you all are so worried about changing the semantics of clear. We already changed the semantics before (in a minor release too) ([ISPN-5370](https://issues.jboss.org/browse/ISPN-5370)) and I don't recall such a palava ([here](https://github.com/infinispan/infinispan/pull/3370) and [here](https://github.com/infinispan/infinispan/pull/3384)).

Sure, this PR has more impact on the users since it'd affect remove notifications, but it's **`clear()`** we're talking about. It's already defined pretty liberally, even more so after ISPN-5370.

anistor:
It's about that ``` more impact``` that we're concerned about. It does not matter if we're doing this in a minor or major, it's bad anyway because it breaks stuff. Comparing this to a precedent like ISPN-5370 is not ideal because ISPN-5370 was a compromise needed in order to fix really badly broken things.

I think implementing the proposed ```clear()+SKIP_LISTENER_NOTIFICATION``` is more than enough good compromise and should make everybody happy. Btw, we should ensure the skip_listener_notification flag works with any write command.

danberindei:
@galderz I'm not that worried about the user listeners, at some point they'll put a breakpoint in their listeners and see they're not invoked during clear. I'm more worried that maybe we use listeners internally for some other features (near cache?), and the connection between those features and `clear()` is not obvious to the users. Maybe it's less of a reason to keep `clear()` as is and more of a reason to not use listeners internally :)

galderz:
> I think implementing the proposed clear()+SKIP_LISTENER_NOTIFICATION is more than enough good compromise and should make everybody happy. Btw, we should ensure the skip_listener_notification flag works with any write command.

`Flag.SKIP_LISTENER_NOTIFICATION` does not exist for remote caches.

galderz:
Nor makes sense to add such flag.

tristantarrant:
I'm postponing this for 9.4 as I'd like to get full agreement from everybody on the solution

PR #6037, opened 2018-06-04
ISPN-9171 Address outstanding Clustered Max Idle PR comments

wburns:
https://issues.jboss.org/browse/ISPN-9171


wburns:
Actually added a test and we are doing the timestamp update correctly for the get. I will update the documentation. Also in the test it fails with scattered cache, so there may be something there I have to plug.

wburns:
Should be all updated now.

core/src/main/java/org/infinispan/factories/ExpirationManagerFactory.java
```diff
@@ -13,13 +14,16 @@
...
-@DefaultFactoryFor(classes = ExpirationManager.class)
+@DefaultFactoryFor(classes = {ExpirationManager.class, InternalExpirationManager.class})
 public class ExpirationManagerFactory extends AbstractNamedCacheComponentFactory implements
          AutoInstantiableFactory {
 
    @Override
    @SuppressWarnings("unchecked")
    public <T> T construct(Class<T> componentType) {
+      if (componentType.equals(ExpirationManager.class)) {
+         return componentType.cast(componentRegistry.getOrCreateComponent(InternalExpirationManager.class));
```
danberindei: IMO any component that injects an `ExpirationManager` should inject an `InternalExpirationManager` instead.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -990,7 +988,8 @@ public Object visitRemoveExpiredCommand(InvocationContext ctx, RemoveExpiredComm
 
             return asyncValue(completableFuture).thenApply(ctx, command, (rCtx, rCommand, max) -> {
                if (max != null) {
-                  // Make sure to fail the command for other interceptors
+                  // Make sure to fail the command for other interceptors, such as CacheWriterInterceptor, so it
+                  // it won't remove it in the store
```
danberindei: 1 `it` should be enough :)

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
```diff
@@ -1037,4 +1036,10 @@ public Object visitRemoveExpiredCommand(InvocationContext ctx, RemoveExpiredComm
          }
       }
    }
+
+   public Object maxIdleRemoval(InvocationContext ctx, RemoveExpiredCommand command, DistributionInfo distributionInfo)
```
danberindei: Could be `protected`, and I'd prefer a more "active" name, say `invokeRemoveExpiredCommand`

Looking at `NonTxDistributionInterceptor.handleNonTxWriteCommand()` it also checks `ctx.isOriginLocal()`, so I wonder if we need some tests for clustered expiration in a non-tx cache that uses it (with `PrivateGlobalConfigurationBuilder.serverMode()`).

documentation/src/main/asciidoc/user_guide/eviction.adoc
```diff
@@ -190,16 +190,20 @@ Max idle in a cluster is quite a bit more complicated. Since each node can have
 
 This check is done when a given node finds an entry has expired via maxIdle. It then will ask all the other owners what their last access time was (if it hasn't expired). If all owners say that the entry has expired, then it will not be returned to the user and removed based on the operation.
 
+When an entry is found to have expired due to max idle in the expiration reaper, instead of updating the last access time to current time, it will update all nodes to the most recent access of all owners.
```
danberindei: I think the "updating the last access time to current time" bit is better suited in the paragraph above, and/or this paragraph should be moved after line 199

core/src/main/java/org/infinispan/expiration/impl/ExpirationManagerImpl.java
```diff
@@ -42,6 +42,8 @@
    @Inject protected CacheNotifier<K, V> cacheNotifier;
    @Inject protected TimeService timeService;
 
+   // Used to ensure only a single start invocation actually starts the thread
+   protected final AtomicBoolean started = new AtomicBoolean();
```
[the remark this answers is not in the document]
wburns: This is a problem with having the 2 components (ExpirationManager & InternalExpirationManager). If we don't have both this can be removed.

core/src/main/java/org/infinispan/factories/ExpirationManagerFactory.java
wburns: Yeah I kept going back and forth on this one. Originally I had it throw an exception to tell the user/component that they have to Inject the `InternalExpirationManager`. But maybe we should just have it not have any component factory for `ExpirationManager`.

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
wburns: :)

documentation/src/main/asciidoc/user_guide/eviction.adoc
wburns: Sure

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
wburns: Sure

Let me look at it closer.

documentation/src/main/asciidoc/user_guide/eviction.adoc
wburns: Moved to 199

core/src/main/java/org/infinispan/interceptors/distribution/BaseDistributionInterceptor.java
wburns: As far as I can tell, the `NonTxDistributionInterceptor.handleNonTxWriteCommand` is only invoked for a `RemoveExpiredCommand` with lifespan, in which case this isn't an issue.

Also this issue was because `TriangleDistributionInterceptor` has the backup node send the response to a different node then the message was sent from.

I will add a test to the server ones though.

wburns: Added a test to `MixedExpiryTest` for this.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/MixedExpiryTest.java
```diff
@@ -68,6 +72,33 @@ public void testMixedExpiryMaxIdle() {
       assertMetadataAndValue(cache0.getWithMetadata(key), "value3", 1000, -1);
    }
 
+   public void testMaxIdleRemovedOnAccess() throws InterruptedException, IOException {
+      RemoteCacheManager client0 = client(0);
+      RemoteCache<String, String> cache0 = client0.getCache();
```
danberindei: Please rename to `remoteCache0`, it's too close to `cache(0)`

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/MixedExpiryTest.java
```diff
@@ -68,6 +72,33 @@ public void testMixedExpiryMaxIdle() {
...
+      Object keyStorage = cache(0).getAdvancedCache().getValueDataConversion().toStorage(marshalledKey);
+
+      assertNull(cache0.put(key, "value1", -1, TimeUnit.SECONDS, 100, TimeUnit.MILLISECONDS));
+
+      for (Cache cache : caches()) {
+         DataContainer dataContainer = cache.getAdvancedCache().getDataContainer();
+         assertNotNull(dataContainer.peek(keyStorage));
+      }
+
+      Thread.sleep(150);
```
danberindei: `ControlledTimeService`?

wburns:
Updated

danberindei:
Integrated, thanks Will!

PR #6038, opened 2018-06-05
ISPN-9258 javax.transaction dependency is not provided

galderz:
https://issues.jboss.org/browse/ISPN-9258

Not really a provided dependency based on how the code is laid out right now. 

Let's integrate this and @pruivo can take over to really make it provided?


pruivo:
integrated! thanks @galderz.

PR #6040, opened 2018-06-05
ISPN-9259 Feature-packs should utilise modules/system/add-ons dir

ryanemerson:
https://issues.jboss.org/browse/ISPN-9259

@Sanne @fax4ever FYI after becoming aware of the following [guidelines](https://developer.jboss.org/wiki%20/LayeredDistributionsAndModulePathOrganization) we're moving the wildfly-modules to `modules/system/add-ons/ispn`. Let me know if you know of any issues etc.

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -165,6 +165,7 @@
                                 <experimental>true</experimental>
                                 <infinispanversion>${infinispan.base.version}</infinispanversion>
                                 <brandname>${infinispan.brand.name}</brandname>
+                                <moduleprefix>${infinispan.module.slot.prefix}</moduleprefix>
                                 <infinispanslot>${infinispan.module.slot.prefix}-${parsedVersion.majorVersion}.${parsedVersion.minorVersion}</infinispanslot>
                                 <defaults>${defaults.file}</defaults>
                                 <javadocroot>${javadoc.root}</javadocroot>
--- a/documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
+++ b/documentation/src/main/asciidoc/upgrading/upgrading.asciidoc
@@ -30,6 +30,10 @@ Max idle in a transactional clustered cache does not remove expired entries on a
 
 Iteration in a clustered cache will still show entries that are expired via maxIdle to ensure good performance, but could be removed at any point due to expiration reaper.
 
+=== Wildfly Modules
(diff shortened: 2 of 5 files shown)


alanfx:
@wfink you should probably look at this too, since it is going to affect patching of the server

fax4ever:
@ryanemerson thank you for concern, there are no problems on our side. It's all about the way wildfly-feature-pack maven plugin find dependent modules, and it will able to found them in both cases. I followed the topic on Zulip, I agree that it is better to avoid to install static modules on default (no package) or base layer.

Sanne:
@alanfx I'm not aware about "patching" capabilities of the feature packs? We use the feature packs to build the (custom) server.

Would love to know more about patching capabilities as we had questions about that on the Hibernate forums.

Sanne:
@ryanemerson I see no problems with the move, however I don't think the guidelines have been updated when we moved to feature packs.

As I asked also to the WildFly team, IMO which "layer" it is going to be written into should be something that the provisioning tool to control - e.g. I might want to take you pack and include it into another custom layer, give it a different name.

ryanemerson:
@Sanne I agree that it would be much nicer if we could control the destination layer as part of the provisioning.

alanfx:
@Sanne the product uses patching for micro releases. I just wanted to make Wolf aware so he can build patches if the modules are moved to the `modules/system/add-ons/ispn` dir.

Sanne:
@alanfx ah, so you just care about the path. Got it, thanks.

tristantarrant:
According to the guidelines, add-ons can be patched in the same ways:
`/modules/system/add-ons/abc/patches/patch-abc-1
/modules/system/add-ons/abc`

alanfx:
Thanks, @tristantarrant! @wfink, hope this works for you.

ryanemerson:
@tristantarrant This PR is good to go. The failures in [build #2](https://ci.infinispan.org/job/Infinispan/job/PR-6040/2/) have happened elsewhere and are not caused by these changes.

tristantarrant:
@ryanemerson can you add a note to the upgrading guide ?

ryanemerson:
@tristantarrant I noticed a slight problem with the dir layout whilst writing the docs, so lets wait for a full CI run before merging.

ryanemerson:
Failures not related

tristantarrant:
Merged, thanks

PR #6041, opened 2018-06-05
ISPN-9020 Remote query: disabling indexing for a type with @Indexed(false) still creates empty documents in index

anistor:
…ng to handle conditional indexing for @Indexed(false)

https://issues.jboss.org/browse/ISPN-9020

* add ProtobufValueWrapperInterceptor to intercept writing of ProtobufValueWrapper objects in the cache and discover the inner message type
* add ProtobufValueWrapperIndexingInterceptor to provide conditional indexing
* rename ProtostreamWrapper to ProtobufWrapper because it's really the encoding format not the encoding library that's important


anistor:
CI failures are not related. The directory failures are also on other branches.

gustavonalle:
Integrated, thanks

PR #6044, opened 2018-06-06

arkaev:
In case of CacheLoader usage (e.g. JCache's CacheLoader)

Problem:
1. Element loaded throw CacheLoader with current time
2. Then checked by expiration
3. Expiration failed
4. Returned null instead of loaded element

More complex case:
1. Element in cache
2. Element requested from JCache
3. No expiration happend when check for reloading in JCache
4. Second check in PersistenceUtil and expiration happend
5. Returned null

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/PersistenceUtil.java
+++ b/core/src/main/java/org/infinispan/persistence/PersistenceUtil.java
@@ -143,6 +143,8 @@ public class PersistenceUtil {
 
    public static <K, V> MarshalledEntry<K, V> loadAndCheckExpiration(PersistenceManager persistenceManager, Object key,
                                                         InvocationContext context, TimeService timeService) {
+      long expireCheckTime = timeService.wallClockTime();
+
       final MarshalledEntry<K, V> loaded = persistenceManager.loadFromAllStores(key, context.isOriginLocal());
       if (trace) {
          log.tracef("Loaded %s for key %s from persistence.", loaded, key);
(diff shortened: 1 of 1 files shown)


tristantarrant:
ok to test

wburns:
Technically doing either is fine from our perspective. Since the entry expired during the operation, it is undefined if it is returned or not.

I don't know if there is anything in JCache spec that says if it should be removed or not.

arkaev:
I have strange situation: CacheLoader never returns null, but cache return null in some expiration situations

wburns:
So is your loader computing this value every invocation and sets the expiration extremely low? Just trying to understand your use case? Are you trying not to store anything in the Cache itself?

arkaev:
No. In my case I have frequently situation, when cache return null value after some time in cache.

wburns:
That seems like a different issue than what this was created for. All this would do is make it so that if during the computation of an entry returned from the cache loader it expired. It sounds like you are having a different issue (concurrent remove, node failures, eviction). You said there was no expiration metadata brought along with the entry created from the loader, so it cannot expire.

arkaev:
I think it's at least one of problem. On high load we will receive null values on any expiration time (on end of life of element). After expiration cache should reload value but it returns null

wburns:
Do you have a tests case for this, sorry? I just fail to see how this fix can fix the issues you are talking about.

It sounds like you are saying your issue is that an entry expired and the cache loader wasn't queried to replace the value?

wburns:
So this was the case I mentioned earlier where the expiration is set extremely low. And that the value can expire between creation and checking the timestamp.

Actually thinking about it more, this check in `PersistenceUtil` doesn't make any sense though. Each `org.infinispan.persistence.spi.CacheLoader` implementation should be verifying that the entry isn't expired and not the caller.

So in that case I would say this check should be removed all together.

wburns:
I created #6068 for master to see it if passes alright.

wburns:
This PR is superseded by https://github.com/infinispan/infinispan/pull/6071

PR #6045, opened 2018-06-06
ISPN-9181 & ISPN-9184 Remove compat mode from Remote Listeners and Remote Iterator

gustavonalle:
https://issues.jboss.org/browse/ISPN-9181
https://issues.jboss.org/browse/ISPN-9184

This PR removes compat mode and hardcoded marshallers from remote listeners and remote iterator.

All conversions are now done on-demand, based on 1) The format of the data sent by clients 2) the configured Mediatype of the cache 3) and the filter/converter MediaType

Overview of changes:

* Removes the limitation of a server wide event marshaller
* Removes the limitation of a server wide iterator marshaller
* Removes the limitation of listeners/filters/converters params to be unmarshalled by GenericJBossMarshaller
* Allow filters and converters to specify the format to process data
* Created a new protobuf deployment unit where .proto files and extra marshallers can be added
* Introduces a media type "application/unknown" to flag that a cache has no media type configured


galderz:
Looks good to me :)

gustavonalle:
Thanks @galderz, there are 2 or 3 minor things to fix (loggers) and a couple of commented tests, will get those sorted out.

galderz:
Needs rebase

gustavonalle:
On it

gustavonalle:
Rebased, removed TODOs, commented out tests and added loggers. Only docs missing

server/hotrod/src/main/java/org/infinispan/server/hotrod/iteration/DefaultIterationManager.java
```diff
@@ -77,78 +75,73 @@ public void segmentCompleted(Set<Integer> segments) {
...
          CollectionFactory.makeConcurrentMap();
 
-   public DefaultIterationManager(EmbeddedCacheManager cacheManager) {
-      this.cacheManager = cacheManager;
-   }
-
    @Override
-   public String start(Cache cache, Optional<BitSet> segments, Optional<KeyValuePair<String, List<byte[]>>> namedFactory, int batch, boolean metadata) {
+   public String start(CacheDecodeContext cdc, Optional<BitSet> segments, Optional<KeyValuePair<String, List<byte[]>>> namedFactory, int batch, boolean metadata) {
       String iterationId = UUID.randomUUID().toString();
```
tristantarrant: Use https://github.com/infinispan/infinispan/blob/41b438342378e603caf34040d16d1cd8bbd6cc23/commons/src/main/java/org/infinispan/commons/util/Util.java#L918

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/deployments/ProtoStreamMarshallerExtensionProcessor.java
```diff
@@ -0,0 +1,152 @@
...
+import org.jboss.msc.service.StopContext;
+import org.jboss.msc.value.InjectedValue;
+import org.jboss.vfs.VirtualFile;
+
+/**
+ * Processor for protobuf deployments, containing protostream marshallers and proto files.
+ *
+ * @since 9.3
+ */
+public class ProtoStreamMarshallerExtensionProcessor implements DeploymentUnitProcessor {
```
tristantarrant: Nice!

commons/src/main/java/org/infinispan/commons/dataconversion/DefaultTranscoder.java
```diff
@@ -23,21 +27,33 @@
 
    private static final Set<MediaType> supportedTypes = new HashSet<>();
 
-   public static final DefaultTranscoder INSTANCE = new DefaultTranscoder();
+   private final GenericJBossMarshaller jbossMarshaller;
+   private JavaSerializationMarshaller javaSerializationMarshaller;
```
anistor: Can this be final?

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerWithFilterAndProtobufTest.java
```diff
@@ -48,24 +50,26 @@
...
       RemoteCache<String, String> metadataCache = client(0).getCache(ProtobufMetadataManagerConstants.PROTOBUF_METADATA_CACHE_NAME);
       metadataCache.put("sample_bank_account/bank.proto", Util.getResourceAsString("/sample_bank_account/bank.proto", getClass().getClassLoader()));
       assertFalse(metadataCache.containsKey(ProtobufMetadataManagerConstants.ERRORS_KEY_SUFFIX));
 
+      // The server already has a Protostream <=> Object transcoder, register extra marshallers in the ProtobufMetadataManager.
+      // In standalone servers, those marshallers can be deployed using a jar to have the same effect.
+      for (int i = 0; i < NUM_NODES; i++) {
+         server(i).addCacheEventFilterFactory("custom-filter-factory", new CustomCacheEventFilterFactory());
+         ProtobufMetadataManagerImpl metadataManager = ProtobufMetadataManagerImpl.getProtobufMetadataManager(server(i).getCacheManager());
+         MarshallerRegistration.getMarshallers().forEach(metadataManager::registerMarshaller);
```
anistor: I would prefer not to make ```ProtobufMetadataManagerImpl.getProtobufMetadataManager(EmbeddedCacheManager cacheManager)``` public.

An alternative is to use          MarshallerRegistration.registerMarshallers(ProtobufMetadataManagerImpl.getSerializationContextInternal(server(i).getCacheManager()));

That's also not going to require adding MarshallerRegistration.getMarshallers() anymore.  Or are we in any trouble because MarshallerRegistration.registerMarshallers also registers the proto files?

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/deployments/ProtoStreamMarshallerExtensionProcessor.java
anistor: Nice but dangerous like a hydrogen based Zeppelin :).  The unpredictable order of processing protos and marshallers will lead to chaos and cannot be completely controlled with dependencies declared via meta-inf manifest. Since the whole thing is based on deploying a jar (not a bad thing) I would go a bit further and give the user maximum programatic control. Rather than we do the auto-scanning of META-INF for proto files I would prefer to rely on ServiceLoader mechanism for the whole thing (not just the marshallers). Let's define an interface the user can implement that receives a SerializationContext and the user would be able to register marshallers and protos to it as he wishes. That way he can control the order and can handle errors, or at least fail with dignity, in a predictable way :)
```
interface ProtostreamSerializationContextInitializer {
 void init(SerializationContext ctx);
}
```
Ok, predictability and transparency solved, less mystery for the user.

Now the next question is why do we want a DeploymentUnitProcessor specifically for this. We can handle this quite easily by using the serviceloader mechanims to lookup ProtostreamSerializationContextInitializer implementations during  cache manager startup in the remote query module intializer handler.  How do these implementation classes coming from deployed jar become visible to remote query module intializer handler? We use the <modules> mechanism added in 9.1. That's one proposal to save us from adding one more DeploymentUnitProcessor that is probably not going to survive _The Second Coming_ of the server :)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/DataFormat.java
```diff
@@ -41,11 +41,15 @@ public DataFormat withoutValueType() {
    }
 
    public MediaType getKeyType() {
-      return keyType;
+      if(keyType != null) return keyType;
```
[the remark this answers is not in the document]
anistor: There are many such missing spaces in this PR. @gustavonalle Please scan for ```if(``` in your changes and fix them. Let's avoid WWIII. They said it is going to be a space war :)

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerWithFilterAndProtobufTest.java
gustavonalle: OK, I will not make it public, it was only a convenience  for tests....

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/deployments/ProtoStreamMarshallerExtensionProcessor.java
gustavonalle: > The unpredictable order of processing protos and marshallers 

The problem that you are describing will happen when you deploy one jar with .proto only, and another with marshaller only; in the same jar it will work
 or am I missing something?

> why do we want a DeploymentUnitProcessor specifically for this
> We use the <modules> mechanism added in 9.1. That's one proposal to save us from adding > one more DeploymentUnitProcessor

On master, you can deploy a marshaller and it becomes the global event marshaller. So each time you are using filters/converters and protobuf caches in the server, you'd need to deploy your whole full-fledge ProtoStream marshaller, protofile, create your own serCtx, and so on. 

This PR is doing something similar, but it assumes .proto files are on META-INF/ and it reads and registers for you, but the big difference is now there is not more global event marshaller and user doesn't need to have its own ProtoStream marshaller, but it can extend the *existing* serCtx in the server, by adding extra messagemarshallers and protos. So it's easier to use.

>  That's one proposal to save us from adding one more DeploymentUnitProcessor

The new server will need to support deployed jars, we have too many of those scenarios already, so I am not worried about it

Your proposal could work, it gives more power but it's more legwork for the user, and this PR is already a great improvement over what we have now, so could we create a JIRA and postpone for 9.4?

gustavonalle: > We use the <modules> mechanism added in 9.1

btw, this also requires the ```<module>``` mechanism. Which means, the jar with the marshallers and .proto must be visible during the server startup, because marshalling involves entities and those entities need to be visible to the ProtoStream transcoder during starting

anistor: Yes, such split jars will happen.

And also the order of registering marshallers may matter. Not for the most cases of BaseMarshaller but certainly for the generate ones.
 
Oh, speaking about the generated ones, what I don't like about proto files being regular files under meta-inf is that we cannot use the annotation based proto schema generator of protostream, which I would like people to use whenever possible. For generating the schema, and for generating the marshallers.

Also, protostream 5 will support a DSL with lambdas for defining marshallers, without implementing a class derived from BaseMarshaller. So looking up BaseMarshaller classes with ServiceLoader is not the whole story.

The more legwork for the user is implementing an interface with one method which is quite trivial and the code to be written there is what they already write for the client side. I'm not even suggesting they should copy-paste that code :) 

I like the fact there is no more global event marshaller, so I'm all in for most changes in this PR except this DeploymentUnitProcessor.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerWithFilterAndProtobufTest.java
anistor: No fallout from nucular tests please :)

server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/deployments/ProtoStreamMarshallerExtensionProcessor.java
anistor: There is also the question which of the many cache manager gets which protos and marshallers. My assumption is we have a good answer already: based on the contents of ```<modules>``` of each cache manager.   What gets listed under modules just needs to be a simple jar deployed under the deplouyments dir, not necessarily something that was specially processed by a DeploymentUnitProcessor.

gustavonalle: yes, that's how the PR works at the moment, I will simply remove the DUP and use service loader instead

gustavonalle:
Rebased, addressed reviews and added docs

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerWithFilterAndProtobufTest.java
```diff
@@ -48,24 +50,25 @@
...
+      defaultClusteredCacheConfig.encoding().value().mediaType(MediaType.APPLICATION_PROTOSTREAM_TYPE);
+      ConfigurationBuilder cfgBuilder = hotRodCacheConfiguration(defaultClusteredCacheConfig);
       createHotRodServers(NUM_NODES, cfgBuilder);
       waitForClusterToForm();
 
-      for (int i = 0; i < NUM_NODES; i++) {
-         server(i).addCacheEventFilterFactory("custom-filter-factory", new CustomCacheEventFilterFactory());
-
-         // WARNING! This is not the actual instance used at runtime. A new instance is created instead, so all state is lost unless it is re-created in the constructor!
-         ProtoStreamMarshaller marshaller = new CustomProtoStreamMarshaller();
```
anistor: CustomProtoStreamMarshaller class is no longer used here but it still exists in ClientListenerWithFilterAndProtobufTest. Please remove the dead code.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerWithFilterAndProtobufTest.java
```diff
@@ -48,24 +50,25 @@
 
    @Override
    protected void createCacheManagers() throws Throwable {
-      ConfigurationBuilder cfgBuilder = hotRodCacheConfiguration(getDefaultClusteredCacheConfig(CacheMode.DIST_SYNC, false));
+      ConfigurationBuilder defaultClusteredCacheConfig = getDefaultClusteredCacheConfig(CacheMode.DIST_SYNC, false);
```
anistor: I see this PR touched ClientListenerWithFilterAndProtobufTest and I'm a bit surprised ClientListenerWithFilterAndRawProtobufTest is left unchanged. Overlooked or that the actual intention?

persistence/remote/src/main/java/org/infinispan/persistence/remote/upgrade/RemovedFilter.java
```diff
@@ -18,6 +19,11 @@ public boolean accept(Object key, Object oldValue, Metadata oldMetadata, Object
...
    }
 
+   @Override
+   public MediaType format() {
+      return null;
+   }
+
    public static class Externalizer extends NoStateExternalizer<RemovedFilter> {
 
       @Override
```
anistor: No exceptions thrown by this method.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/impl/iteration/ProtobufRemoteIteratorTest.java
```diff
@@ -53,26 +61,25 @@ protected void createCacheManagers() throws Throwable {
       metadataCache.put("sample_bank_account/bank.proto", Util.getResourceAsString("/sample_bank_account/bank.proto", getClass().getClassLoader()));
       assertFalse(metadataCache.containsKey(ProtobufMetadataManagerConstants.ERRORS_KEY_SUFFIX));
 
-      ProtoStreamMarshaller marshaller = new CustomProtoStreamMarshaller();
-
-      servers.forEach(s -> s.setMarshaller(marshaller));
+      // Register extra marshallers in the ProtobufMetadataManager. In standalone servers this can be done with a deployment jar.
+      cacheManagers.forEach(cm -> {
```
anistor: Lines 65-72 can be replaced with
```
      for (EmbeddedCacheManager cm : cacheManagers) {
         MarshallerRegistration.registerMarshallers(ProtobufMetadataManagerImpl.getSerializationContextInternal(cm));
      }
```

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod.adoc
```diff
@@ -1062,10 +1066,38 @@ marshaller class implementation.
 3. Deploy the JAR file in the {brandname} Server.
 
 Note that the Marshaller could be deployed in either a separate jar, or in the
-same jar as the `CacheEventConverter` and/or `CacheEventFilter` instances. Also, currently
-deployment of a single `org.infinispan.commons.marshall.Marshaller` instance
-is supported. If multiple marshaller instances are deployed, warning messages
-will be displayed as reminder indicating which marshaller instance will be used.
+same jar as the `CacheEventConverter` and/or `CacheEventFilter` instances.
+
+===== Protostream deployment
```
anistor: ```Protobuf deployment``` would sound better. No need to stress on protostream, it's just an implementation detail. Nobody knows what protostream is :)

feature-pack/server/src/main/resources/modules/org/infinispan/server/endpoint/slot/module.xml
```diff
@@ -40,6 +40,7 @@
       <module name="org.jboss.marshalling" services="import" slot="@infinispan.module.slot@"/>
       <module name="org.jboss.resteasy.resteasy-jaxrs" services="import" />
       <module name="org.jboss.security.negotiation" />
+      <module name="org.jboss.vfs"/>
```
anistor: Is vfs still needed? I thought it was only used for examining the proto files inside the jar by the DUP, but that one no longer exists.

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/logging/Log.java
```diff
@@ -58,4 +58,7 @@
          " 'option indexed_by_default = false;' to your schema file to disable indexing of message types" +
          " that do not have indexing annotations.", id = 28012)
    void legacyIndexingIsDeprecated(String typeName, String fileName);
+
+   @Message(value = "Error initializing the serialization context", id = 28013)
```
anistor: The error message is not helpful enough for the average Joe. Let's say ```Error during execution of protostream serialization context initializer %s in module %```, where first arg is the class name and second arg is module name/id.

anistor: Hmmm, maybe module name is hard to obtain, and the class name is already in the stacktrace. Still, let's change the error message.

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerImpl.java
```diff
@@ -234,7 +239,7 @@ SerializationContext getSerializationContext() {
    /**
     * Obtains the ProtobufMetadataManagerImpl instance associated to a cache manager.
     */
-   private static ProtobufMetadataManagerImpl getProtobufMetadataManager(EmbeddedCacheManager cacheManager) {
+   public static ProtobufMetadataManagerImpl getProtobufMetadataManager(EmbeddedCacheManager cacheManager) {
```
anistor: This can be private again.

server/hotrod/src/main/java/org/infinispan/server/hotrod/iteration/IterationManager.java
```diff
@@ -4,17 +4,16 @@
...
+import org.infinispan.server.hotrod.CacheDecodeContext;
 import org.infinispan.util.KeyValuePair;
 
 /**
  * @author wburns
  * @since 9.0
  */
 public interface IterationManager {
-   String start(Cache cache, Optional<BitSet> segments,
+   String start(CacheDecodeContext cache, Optional<BitSet> segments,
```
anistor: Still name this arg ```cache``` after changing type to CacheDecodeContext ?

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/ClientListenerWithFilterAndProtobufTest.java
gustavonalle: No need to change that one since it uses "raw" filters/converters, but I will port over the comments

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod.adoc
gustavonalle: I think I meant "Protostream marshallers deployment"

feature-pack/server/src/main/resources/modules/org/infinispan/server/endpoint/slot/module.xml
gustavonalle: left over...

gustavonalle:
Updated

gustavonalle:
CI failed to run the build due to some connection error, retriggered it... 
In the meantime, @anistor, any other review?

anistor:
Thanks @gustavonalle ! LGTM. Let's wait for CI

gustavonalle:
Ci looks good

PR #6047, opened 2018-06-08
ISPN-9219 ISPN-9266 (Weak|Strong)CounterAPITest.testAdd random failures

pruivo:
https://issues.jboss.org/browse/ISPN-9219
https://issues.jboss.org/browse/ISPN-9266

at least locally, the failures are caused by retried functional commands. 

https://github.com/infinispan/infinispan/pull/5381 should fix those. In the meanwhile, lets make sure the counter's cache is started before start the testing.


gustavonalle:
There are failures in ```RemoveCounterTest```, I guess that's out of scope of this fix?

pruivo:
@gustavonalle I've triggered another build. it isn't related. this PR changes the tests in the hot rod client/server and the RemoveCounterTest runs in embedded mode. 
The error seems there is something in the cache while it is supposed to be empty... lets see if it fails again or not.

gustavonalle:
ok, @pruivo merging this then, thanks!

PR #6049, opened 2018-06-08
ISPN-9281 Handle ClassNotFoundException for io.netty.channel.epoll.Epoll

tristantarrant:
https://issues.jboss.org/browse/ISPN-9281

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/TransportHelper.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/TransportHelper.java
@@ -22,14 +22,25 @@ class TransportHelper {
    private static final boolean USE_NATIVE_EPOLL;
 
    static {
-      if (Epoll.isAvailable()) {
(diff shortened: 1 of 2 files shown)


gustavonalle:
LGTM

gustavonalle:
There are several failures in integration tests, are they related?

tristantarrant:
No, they seem to be something related to a failed server that is not relinquishing ports. @ryanemerson could these be related to the store avail stuff ?

rvansa:
@tristantarrant Why would you omit the jar from the classpath? I get that on some platforms you can't use the native part, but the classes should still be there.

rvansa:
> EAP wants all native bits packaged in separate RPMs

Eh, okay ... so to satisfy their needs EAP could use the artifact without native bits in the base deployment and add the bits from `:linux-x86_64` in this separate RPM. In the worst case the classes would be on the classpath twice but as long as noone forgots to update the version it should do the trick.

I am not turning down your solution, but the reason for the handling in code is non-obvious and ^ seems more elegant.

tristantarrant:
Unfortunately, creating such a build takes more time than making the epoll
dependency optional...

PR #6051, opened 2018-06-08
ISPN-9061 X-site replication with functional commands throws NullPointerException

pruivo:
https://issues.jboss.org/browse/ISPN-9061

### DIFF
--- a/core/src/test/java/org/infinispan/xsite/CacheOperationsTest.java
+++ b/core/src/test/java/org/infinispan/xsite/CacheOperationsTest.java
@@ -9,6 +9,8 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import javax.transaction.TransactionManager;
+
 import org.infinispan.Cache;
 import org.infinispan.commons.util.Util;
 import org.infinispan.configuration.cache.CacheMode;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/infinispan/xsite/CacheOperationsTest.java
```diff
@@ -228,6 +230,19 @@ private void testFunctional(String site) {
       for (Object key : keys) {
          assertEquals(null, backup.get(key));
       }
+
+      if (transactional) {
+         TransactionManager tm = cache(site, 0).getAdvancedCache().getTransactionManager();
+         tm.begin();
+         rw.eval(keys[0], "v4", MarshallableFunctions.setValueReturnPrevOrNull()).join();
```
rvansa: Why don't you execute the operation on all keys in the set? (the test is supposed to test several combinations).

pruivo: because it wouldn't reproduce the NPE.

The NPE is caused by an `evalMany()` with a read-only function in in-existent keys. This will create an `CacheEntry` with `null` value in the `InvocationContext`.

The `eval(keys[0])` is to make sure the transaction is committed in the backup site with the read-only `evalMany()`.

rvansa: Do we have to miss *all* keys from the set to hit the NPE?

pruivo: no. but if I don't miss all key, I'll have to create a new function :) without any benefit for the test

rvansa:
Okay, merged then. Thanks.

PR #6055, opened 2018-06-11
ISPN-9187 Lost segments logged when node leaves during rebalance

danberindei:
https://issues.jboss.org/browse/ISPN-9187

Keep joiners in the pending CH when updating members after a leave.


gustavonalle:
Merged, thanks

PR #6058, opened 2018-06-11
ISPN-9287 Integration test suite fails when JAVA_HOME is not set

danberindei:
https://issues.jboss.org/browse/ISPN-9287

* Only fail the build if all tools have a non-zero exit code
* Fix the infinispan-server.xml path in the rolling upgrades profile

### DIFF
--- a/server/integration/src/main/ant/infinispan-server.xml
+++ b/server/integration/src/main/ant/infinispan-server.xml
@@ -28,6 +28,7 @@
 
     <target name="start-server" depends="-check.skipped" unless="tests.skipped">
         <echo>Starting Infinispan server from ${server.dist}</echo>
+
         <exec dir="${server.dist}/bin" executable="chmod" osfamily="unix">
             <arg value="+x"/>
             <arg value="../bin"/>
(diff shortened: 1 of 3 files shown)


server/integration/src/main/ant/infinispan-server.xml
```diff
@@ -53,127 +54,125 @@
...
-            </filterchain>
-        </loadfile>
         <exec executable="taskkill" osfamily="windows">
-            <arg line="/F /T /PID ${cpid}"/>
+            <arg line="/F /T /PID ${cmd.pid}"/>
         </exec>
-        <delete file="cmd.pid"/>
     </target>
 
     <target name="kill-server" depends="-do.kinit, -do.ps, -do.lsof, -do.jps, -do.cmd"/>
```
gustavonalle: Suggestion: the server can be stopped on Linux with:
```
<HOME>/bin/ispn-cli.sh --connect command=:shutdown  
```

and on Windows:

```
<HOME>/bin/ispn-cli.bat --connect command=:shutdown
```

server/integration/testsuite/pom.xml
```diff
@@ -1245,10 +1245,10 @@
                                 <configuration>
                                     <skip>${skipTests}</skip>
                                     <target>
-                                        <ant antfile="${basedir}/src/test/resources/ant/infinispan-server.xml" target="kill-server">
+                                        <ant antfile="../src/main/ant/infinispan-server.xml" target="kill-server">
```
gustavonalle: I'd prefer to use variables rather than relative paths, what was the problem?

server/integration/src/main/ant/infinispan-server.xml
gustavonalle: You probably could get rid of all this logic coded in XML by simply calling the ispn-cli above, as it takes care of all the platform/JVM specific details  :)

danberindei: I'm not sure if it's still a problem nowadays but back when the scripts were first written we had problems with the server not stopping properly (https://issues.jboss.org/browse/WFLY-4433 I think). So even if we use the cli, we'd still have to keep `kill -9` as a backup.

server/integration/testsuite/pom.xml
danberindei: The problem was that `src/test/resources/ant/infinispan-server.xml` no longer exists, `infinispan-server.xml` is now in the parent module/directory.

I can add `${basedir}` in front if you prefer, I just copied the path that we were already using elsewhere.

server/integration/src/main/ant/infinispan-server.xml
gustavonalle: Provided that the ```kill -9``` would cause the test to fail rather than hide the problem, I'm fine with it.

PR #6059, opened 2018-06-12
ISPN-9188 Cancelling the last segment doesn't remove transfer task

danberindei:
https://issues.jboss.org/browse/ISPN-9188

* Cancel the task when there are no more unfinished segments
* Cancelling shouldn't affect the initial segments, only the unfinished
  segments


danberindei:
Some of the test failures may be related, intestigating

danberindei:
Fixed and rebased

PR #6060, opened 2018-06-12
ISPN-9289 Ignore TransactionSynchronizationRegistry

pruivo:
Ignore the registry injected by Wildfly server for Hot Rod Transactions.

https://issues.jboss.org/browse/ISPN-9289

### DIFF
--- a/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
+++ b/core/src/main/java/org/infinispan/transaction/impl/TransactionTable.java
@@ -243,6 +243,19 @@ public class TransactionTable implements org.infinispan.transaction.TransactionT
       }
    }
 
+   public void enlistClientTransaction(Transaction transaction, LocalTransaction localTransaction) {
+      if (!localTransaction.isEnlisted()) {
+         SynchronizationAdapter sync = new SynchronizationAdapter(localTransaction, this);
+         try {
+            transaction.registerSynchronization(sync);
+         } catch (Exception e) {
+            log.failedSynchronizationRegistration(e);
+            throw new CacheException(e);
+         }
+         ((SyncLocalTransaction) localTransaction).setEnlisted(true);
+      }
+   }
+
    public void failureCompletingTransaction(Transaction tx) {
       final LocalTransaction localTransaction = localTransactions.get(tx);
       if (localTransaction != null) {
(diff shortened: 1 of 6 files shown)


rvansa:
Hmm, I thought that we're not using TM on server when running HR transactions... or do we?
If it's so why don't we stick with the registry that's configured in? (I'd rather clear that setting from configuration if it does not make sense than have a special case in the code).

pruivo:
@rvansa we aren't using the TM for client transaction. But the `TransactionTable` resumes/suspends the transaction when the registry is configured and you will get the following exception:
```
ISPN004005: Error received from the server: javax.transaction.InvalidTransactionException: WFTXN0002: Transaction is not a supported instance: TransactionImpl{xid=Xid{formatId=1213355096, globalTransactionId=ACE72FF3C45B1C95ECDD0CABA4474FBD00000163EF327F9A0000000000000002,branchQualifier=ACE72FF3C45B1C95ECDD0CABA4474FBD00000163EF327F9A0000000000000002}, status=ACTIVE}
```

rvansa:
@pruivo neither the EmbeddedTransactionManager, right? So how does it come that if `transactionManagerLookup` returns `null` (?) and `transactionSynchronizationRegistryLookup` is set?

pruivo:
@rvansa neither the ETM. 
The `transactionManagerLookup` never returns `null`. It is running inside the wildfly server and you have access to naryana or  `ContextTransactionManager`. 
In addition, you can't start a transactional cache without a `TransactionManager`.

rvansa:
So the cache has a TM configured (via TMLookup) but it's not used? I find that confusing. Could we avoid that, e.g. by setting a NoTMLookup class for HR caches?
Can HR transactions use transactional cache store when persisting the transactions?

pruivo:
@rvansa it isn't worth the complexity. You can have another endpoints (memcached, rest) interacting with the cache and they use the TM to create an implicit transaction. Or an old client version without tx support. In that case, the `CacheImpl` already handles everything for me :)
 
Yes, the tx cache store can be used. They check the transaction from the `InvocationContext`. 
Note that you can't interact with the `TransactionManager` in the interceptors since they are async and the `Transaction` is attached to a `ThreadLocal`.

gustavonalle:
@rvansa That was fixed today: https://github.com/infinispan/infinispan/pull/6062

rvansa:
Ok, pushing this in, then.

pruivo:
@danberindei fix: https://github.com/infinispan/infinispan/pull/6074

PR #6061, opened 2018-06-12
WriteBehindFaultToleranceTest Failure Fix

ryanemerson:
https://issues.jboss.org/browse/ISPN-9286

### DIFF
--- a/core/src/test/java/org/infinispan/persistence/WriteBehindFaultToleranceTest.java
+++ b/core/src/test/java/org/infinispan/persistence/WriteBehindFaultToleranceTest.java
@@ -3,6 +3,7 @@ package org.infinispan.persistence;
 
 import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertFalse;
+import static org.testng.AssertJUnit.assertNotNull;
 import static org.testng.AssertJUnit.assertTrue;
 
 import java.util.Map;
@@ -16,6 +17,7 @@ import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.configuration.global.GlobalConfiguration;
 import org.infinispan.configuration.global.GlobalConfigurationBuilder;
 import org.infinispan.manager.DefaultCacheManager;
+import org.infinispan.marshall.core.MarshalledEntry;
 import org.infinispan.persistence.async.AdvancedAsyncCacheWriter;
 import org.infinispan.persistence.dummy.DummyInMemoryStore;
 import org.infinispan.persistence.dummy.DummyInMemoryStoreConfigurationBuilder;
@@ -51,7 +53,7 @@ public class WriteBehindFaultToleranceTest extends AbstractInfinispanTest {
       DummyInMemoryStore store = (DummyInMemoryStore) TestingUtil.extractField(AdvancedAsyncCacheWriter.class, asyncWriter, "actual");
       store.setAvailable(true);
       cache.put(1, 1);
-      TestingUtil.sleepThread(AVAILABILITY_INTERVAL * 5);
+      eventually(() -> store.load(1) != null);
       assertEquals(1, store.size());
 
       PersistenceManager pm = TestingUtil.extractComponent(cache, PersistenceManager.class);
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -258,11 +259,15 @@ protected void processIterableBatch(InvocationContext ctx, PutMapCommand cmd, Pe
...
-      Iterable<MarshalledEntry> iterable = () -> cmd.getMap().keySet().stream()
+      Iterator<MarshalledEntry> iterator = cmd.getMap().keySet().stream()
             .filter(filter)
             .map(key -> createMarshalledEntry(ctx, key))
             .iterator();
-      persistenceManager.writeBatchToAllNonTxStores(iterable, mode, cmd.getFlagsBitSet());
+
+      // Ignore empty batches
+      if (iterator.hasNext()) {
+         persistenceManager.writeBatchToAllNonTxStores(() -> iterator, mode, cmd.getFlagsBitSet());
```
danberindei: I don't think this is right. `Iterable.iterator()` is supposed to return a fresh `Iterator` every time, and some store implementations may rely on that.

core/src/test/java/org/infinispan/persistence/WriteBehindFaultToleranceTest.java
```diff
@@ -51,7 +53,11 @@ public void testBlockingOnStoreAvailabilityChange() {
       DummyInMemoryStore store = (DummyInMemoryStore) TestingUtil.extractField(AdvancedAsyncCacheWriter.class, asyncWriter, "actual");
       store.setAvailable(true);
       cache.put(1, 1);
-      TestingUtil.sleepThread(AVAILABILITY_INTERVAL * 5);
+      MarshalledEntry storedEntry = null;
+      for (int i = 0; i < 100 && (storedEntry = store.load(1)) == null; i++) {
+         TestingUtil.sleepThread(100);
```
danberindei: Isn't this like `eventually(() -> store.load(1) != null)`?

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
ryanemerson: Yes good point, I'll push the fix

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -262,7 +262,11 @@ protected void processIterableBatch(InvocationContext ctx, PutMapCommand cmd, Pe
             .filter(filter)
             .map(key -> createMarshalledEntry(ctx, key))
             .iterator();
-      persistenceManager.writeBatchToAllNonTxStores(iterable, mode, cmd.getFlagsBitSet());
+
+      // Ignore empty batches
+      if (iterable.iterator().hasNext()) {
```
wburns: Sorry, I can't see why this was causing issues yet. I thought we ensured that all stores can handle empty `Iterator`s ?

ryanemerson: We did, however introducing the `StoreUnavailableException` has caused another problem with empty iterables. In the [CacheWriterInterceptor](https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java#L253-L254) we call `processIterableBatch` twice, which results in two calls `PersistenceManager::writeBatchToAllNonTxStores`. It is possible for the store availability to change between the first and second method call, which results in the second call throwing a `StoreUnavailableException`. If the Iterable is non-empty, this is fine as it's not possible to write the modification to the store, however if the Iterable is empty this exception should not be thrown.

wburns: Gotcha, so it isn't that it is called twice, but rather because [PersistenceManagerImpl](https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java#L594) checks the availability without caring about if there are any entries to write. In that case I wonder if this is correct spot to fix this.

ryanemerson: Yes very true. I will update all of the the PersistenceManager methods so that if the iterable is empty, then we return before even attempting to acquire the readLock

wburns: I guess this means we can also remove the guideline that stores have to support empty Iterators now as well.

ryanemerson: I've added a revert commit to the PR

wburns:
Looks good to me, just waiting on CI to run again.

wburns:

PR #6064, opened 2018-06-12
ISPN-9292 PessimisticTxAtomicLockingTest and OptimisticTxAtomicLockingTest not running

rvansa:
https://issues.jboss.org/browse/ISPN-9292

### DIFF
--- a/core/src/test/java/org/infinispan/atomic/BaseAtomicMapLockingTest.java
+++ b/core/src/test/java/org/infinispan/atomic/BaseAtomicMapLockingTest.java
@@ -5,16 +5,16 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.concurrent.Callable;
 import java.util.concurrent.Future;
 import javax.transaction.Transaction;
 
+import org.infinispan.commands.remote.recovery.TxCompletionNotificationCommand;
 import org.infinispan.commands.tx.PrepareCommand;
+import org.infinispan.commands.tx.VersionedCommitCommand;
 import org.infinispan.commands.write.PutKeyValueCommand;
 import org.infinispan.configuration.cache.CacheMode;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.context.InvocationContext;
-import org.infinispan.distribution.DistributionTestHelper;
 import org.infinispan.distribution.MagicKey;
 import org.infinispan.interceptors.base.BaseCustomInterceptor;
 import org.infinispan.interceptors.impl.TxInterceptor;
@@ -23,6 +23,7 @@ import org.infinispan.transaction.LockingMode;
 import org.infinispan.util.ControlledRpcManager;
 import org.infinispan.util.concurrent.locks.LockManager;
 import org.testng.Assert;
+import org.testng.annotations.AfterMethod;
 import org.testng.annotations.Test;
 
 /**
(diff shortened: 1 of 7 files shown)


core/src/test/java/org/infinispan/atomic/BaseAtomicMapLockingTest.java
```diff
@@ -174,9 +167,12 @@ public Boolean call() throws Exception {
          } else {
             blockedPrepareResponses.receive();
          }
+
+         rpcManager.expectCommand(TxCompletionNotificationCommand.class).sendWithoutResponses();
          Assert.assertTrue(txOutcome.get());
       } finally {
-         rpcManagers[txExecutor].stopBlocking();
+         rpcManager.stopBlocking();
+         rpcManager.revertRpcManager(cache(txExecutor));
```
danberindei: Either one should be enough. 

Actually could you change `revertRpcManager()` to call `stopBlocking()` first, and then `stopBlocking()` to call `assertEquals(emptyList(), queuedRequests)` instead of logging an error message (which doesn't even get printed on the console)?

danberindei:
There's also an unused import that fails the build

core/src/test/java/org/infinispan/atomic/BaseAtomicMapLockingTest.java
```diff
@@ -188,24 +183,12 @@ protected void assertKeysLocked(int index, Object... keys) {
...
-         boolean onLockOwner = DistributionTestHelper.isFirstOwner(cache(0), key);
-         if ((toLockOwner && onLockOwner) || (!toLockOwner && !onLockOwner)) {
-            return true;
-         }
-      }
-      return false;
-   }
-
-   private void resetBeforeMethod() {
+   @AfterMethod
```
danberindei: Needs `alwaysRun=true`, otherwise TestNG won't run it after a failed test. You could also use `@BeforeMethod` instead.

core/src/test/java/org/infinispan/tx/exception/ReplicationTxExceptionTest.java
```diff
@@ -34,15 +35,18 @@ protected void createCacheManagers() throws Throwable {
...
+      });
       try {
          TransactionManager tm = cache(0).getAdvancedCache().getTransactionManager();
          tm.begin();
          cache(0).put("k0", "v");
          Exceptions.expectException(RollbackException.class, tm::commit);
       } finally {
-         controlledRpcManager.revertRpcManager(cache);
          future.get(30, TimeUnit.SECONDS);
+         controlledRpcManager.revertRpcManager();
```
danberindei: I think it would be better to put the `future.get()` inside the `try` block, so `revertRpcManager()` isn't skipped if it times out.

WDYT about moving the tx in the `fork()` and the `expectCommand()+future.get()` in the `try` body, seems all the other tests do it that way?

rvansa: OK, though `fork()` itself shouldn't ever fail...

danberindei: True, I was only thinking it would look better if starting the future and waiting for it were in the same block. I won't complain if you don't feel the same way :)

rvansa:
@danberindei Hopefully I got it this time.

danberindei:
Thanks Radim! I almost integrated it on Friday anyway, but I had to leave early :)

PR #6066, opened 2018-06-13
Conflict Resolution Fixes

ryanemerson:
https://issues.jboss.org/browse/ISPN-8852
https://issues.jboss.org/browse/ISPN-8976
https://issues.jboss.org/browse/ISPN-9142


core/src/main/java/org/infinispan/partitionhandling/impl/AvailablePartitionHandlingManager.java
```diff
@@ -37,13 +39,13 @@ public void setAvailabilityMode(AvailabilityMode availabilityMode) {/*no-op*/}
    public void checkWrite(Object key) {/*no-op*/}
 
    @Override
-   public void checkRead(Object key) {/*no-op*/}
+   public void checkRead(Object key, Set<Flag> flags) {/*no-op*/}
```
wburns: Thinking we may want to use `long` instead of `Set<Flag>` as is used in other places now.

wburns: Then you can check the flag being set via `FlagBitSets` etc.

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingManagerImpl.java
```diff
@@ -240,26 +244,38 @@ private boolean isTopologyStable(CacheTopology cacheTopology) {
...
       Collection<Address> owners = cacheTopology.getDistribution(key).writeOwners();
       List<Address> actualMembers = cacheTopology.getActualMembers();
-      if (!actualMembers.containsAll(owners) && !isOperationAllowed(isWrite)) {
+      boolean operationAllowed = isOperationAllowed(isWrite, flagBitSet);
+      if (!actualMembers.containsAll(owners) && !operationAllowed) {
          if (trace) log.tracef("Partition is in %s mode, PartitionHandling is set to to %s, access is not allowed for key %s", availabilityMode, partitionHandling, key);
          throw log.degradedModeKeyUnavailable(key);
       } else {
          if (trace) log.tracef("Key %s is available.", key);
+         System.err.println(String.format("Key %s is available.", key));
```
wburns: Forgot to remove this guy :)

ryanemerson: Good spot! Fixed

core/src/main/java/org/infinispan/partitionhandling/impl/PartitionHandlingManagerImpl.java
```diff
@@ -240,26 +244,37 @@ private boolean isTopologyStable(CacheTopology cacheTopology) {
...
-            (!isWrite && partitionHandling != PartitionHandling.DENY_READ_WRITES);
+   protected boolean isOperationAllowed(boolean isWrite, long flagBitSet) {
+      if (availabilityMode == AvailabilityMode.AVAILABLE)
+         return true;
+
+      switch (partitionHandling) {
+         case ALLOW_READ_WRITES:
+            return true;
+         case ALLOW_READS:
+            if (EnumUtil.enumSetOf(flagBitSet, Flag.class).contains(Flag.FORCE_WRITE_LOCK))
```
danberindei: `EnumUtil.containsAny(flagBitSet, FlagBitSets.FORCE_WRITE_LOCK)`

core/src/main/java/org/infinispan/util/logging/Log.java
```diff
@@ -1804,4 +1804,7 @@ CacheConfigurationException offHeapMemoryEvictionSizeNotLargeEnoughForAddresses(
 
    @Message(value = "Maximum startup attempts exceeded for store %s", id = 527)
    PersistenceException storeStartupAttemptsExceeded(String storeName, @Cause Throwable t);
+
+   @Message(value = "Cannot FORCE_WRITE_LOCK as this partition is DEGRADED", id = 528)
```
danberindei: I wonder if we have a similar check for regular locks, maybe we need a more generic message that can be used there as well?

core/src/test/java/org/infinispan/partitionhandling/BasePartitionHandlingTest.java
```diff
@@ -358,6 +359,17 @@ public void assertKeyNotAvailableForRead(Object key) {
          return caches;
       }
 
+      public void assertExceptionWithForceLock(Object key) {
+         for (Cache<Object, Object> c : cachesInThisPartition()) {
+            try {
+               c.getAdvancedCache().withFlags(Flag.FORCE_WRITE_LOCK).get(key);
+               fail();
+            } catch (AvailabilityException ae) {
+               //expected!
```
danberindei: `Exceptions.expectException(() -> ...)`

core/src/test/java/org/infinispan/partitionhandling/BasePartitionHandlingTest.java
```diff
@@ -142,6 +142,28 @@ public String toString() {
...
+         if (p instanceof Discovery) {
+            if (!(p instanceof TEST_PING)) throw new IllegalStateException("TEST_PING required for this test.");
+            ((TEST_PING) p).suspend();
+         }
+      }
+   }
+
+   protected void enableDiscoveryProtocol(JChannel c) {
+      for (Protocol p : c.getProtocolStack().getProtocols()) {
+         if (p instanceof Discovery) {
```
danberindei: It's looks odd that the `instanceof` checks are not symmetric.

core/src/test/java/org/infinispan/partitionhandling/IndependentClustersMergeTest.java
```diff
@@ -0,0 +1,66 @@
...
+      return new Object[] {
+            new IndependentClustersMergeTest().partitionHandling(PartitionHandling.ALLOW_READS),
+            new IndependentClustersMergeTest().partitionHandling(PartitionHandling.DENY_READ_WRITES)
+      };
+   }
+
+   public IndependentClustersMergeTest() {
+      this.numMembersInCluster = 2;
+      this.mergePolicy = (preferredEntry, otherEntries) -> {
+         CacheEntry entry = preferredEntry != null ? preferredEntry : (CacheEntry) otherEntries.get(0);
```
danberindei: Unnecessary cast. Could also avoid the unchecked warning by declaring `entry` and `mergePolicy` with generic parameters `<String, String>`.

core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
```diff
@@ -197,6 +181,11 @@ public void stop() {
       }
    }
 
+   @Override
+   public ClusterManagerStatus getManagerStatus() {
```
danberindei: I wouldn't call it `getManagerStatus()` because it makes me think of DCM, maybe just `getStatus`

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -959,7 +959,10 @@ public synchronized void setInitialTopologyId(int initialTopologyId) {
    public boolean resolveConflictsOnMerge() {
       // It doesn't make sense to resolve conflicts if we are not going to rebalance the cache as entries on "old" owners
       // will not be deleted when no rebalance occurs.
-      return resolveConflictsOnMerge && clusterTopologyManager.isRebalancingEnabled() && rebalancingEnabled;
+      return resolveConflictsOnMerge &&
+            clusterTopologyManager.getManagerStatus().isRunning() &&
```
danberindei: The CTMI status only changes when CTMI itself is stopping, but the caches are all stopped before this point. IIRC conflict resolution requires the cache to be running on the coordinator, so you'd need to check for that explicitly.

core/src/main/java/org/infinispan/util/logging/Log.java
ryanemerson: Not sure I follow. Where would/does this check exist?

danberindei: It doesn't exist now, but it would be in `PartitionHandlingInterceptor.visitLockControlCommand`. I'm not sure how unavailable owners for lock commands are handled now, but since there's no handling in `PHI` I have a feeling they block waiting for a new topology in `StateTransferInterceptor`.

This got me thinking about `entrySet`/`keySet` as well. They're not really "bulk read" operations, because they return a view over the cache, and only iteration or stream operations fetch all the keys. A long time could pass between the call to keySet/entrySet and their iteration, so I don't think they should perform availability checks. They also ignore the `FORCE_WRITE_LOCK` flag ;)

ryanemerson:
failures not related.

core/src/test/java/org/infinispan/partitionhandling/BasePartitionHandlingTest.java
```diff
@@ -142,6 +143,28 @@ public String toString() {
       }
    }
 
+   protected void disableDiscoveryProtocol(JChannel c) {
+      c.getProtocolStack().getProtocols().stream()
```
pruivo: you can use the `findProtocol()` method.

```
TEST_PING ping = c.getProtocolStack().findProtocol(TEST_PING.class);
ping.suspend();
```

PR #6068, opened 2018-06-13

wburns:
* Removed extra expiration check
* CacheLoader should handle check

https://issues.jboss.org/browse/ISPN-8988


wburns:
Testing out this change, as it shouldn't be required as the various CacheLoaders should handle expired entries themselves.

PR #6069, opened 2018-06-14
ISPN-8601 Support entity and collection sharing same region name

galderz:
https://issues.jboss.org/browse/ISPN-8601


galderz:
@rvansa done

PR #6071, opened 2018-06-14

wburns:
* Removed extra expiration check
* CacheLoader should handle check

https://issues.jboss.org/browse/ISPN-8988

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/PersistenceUtil.java
+++ b/core/src/main/java/org/infinispan/persistence/PersistenceUtil.java
@@ -147,13 +147,6 @@ public class PersistenceUtil {
       if (trace) {
          log.tracef("Loaded %s for key %s from persistence.", loaded, key);
       }
-      if (loaded == null) {
-         return null;
-      }
-      InternalMetadata metadata = loaded.getMetadata();
-      if (metadata != null && metadata.isExpired(timeService.wallClockTime())) {
-         return null;
-      }
       return loaded;
    }
 


arkaev:
Big thanks for your attention!
Sadly commit doesn't help... still returns null with PersistentUtil.loadAndStoreInDataContainer()
Occurs `expired.set(Boolean.TRUE)` so 
`      if (expired.get() == Boolean.TRUE) {
         return null; ...`

wburns:
That only affects if an entry is expired in the data container. This check has to stay the way it is, otherwise it would violate our current consistency guarantees.

Unfortunately, we don't currently support reading from the store if an entry is expired from the container. We assume that if it is expired in the container, that the store would also be expired.

You can workaround this by doing 2 things.

1. I would decrease the reaper interval, it will clean out expired entries more often (thus causing the store to be read when it isn't in the store).
2. If you call get a second time, it will be guaranteed to return a non null value. (this wouldn't work with max idle in a cluster though).

What you want is a store that is consulted regardless if the entry is expired. There were talks around something like this before. Basically it would return the expired entry and the store entry would be computed in the background and replaced.  I thought we had a JIRA for this, but we didn't. I created https://issues.jboss.org/browse/ISPN-9365. I would suggest you add any comments to that issue moving forward.

arkaev:
I understand that check can be usefull for cluster... (my problem in local cache)
But for now new versions of infinispan looks like not correctly implement JCache :(((

Workarounds looks not good too

We are on 8.2 version for now (without this problem) and I need some timeout to understand what we should do with problem... May be we should switch to another jcache implementation at least for local caches

tristantarrant:
We think this should work as you expect and we should fix it in 9.3.x. See the discussion at https://infinispan.zulipchat.com/#narrow/stream/118645-infinispan/subject/expiration.20and.20cache.20loader

arkaev:
Guys, you are awesome!

wburns:
Interesting, looking further this was an unintended consequence of https://github.com/infinispan/infinispan/commit/0e7925282273dea82bacb0acbaf63a30240c9d00. But it makes sense to keep it how it is and only query the persistence layer if there is an instance that only implements `CacheLoader`.

I have created https://issues.jboss.org/browse/ISPN-9370 which should get into 9.3.1.Alpha1 barring any unforeseen issues.

wburns:
Upon further investigation actually ISPN core itself seems okay with the current code base as the data container is always cleared by the time it gets into `PersistenceUtil` due to the `EntryWrappingInterceptor` expiring the entry.

I will see if the same happens when using JCache. Either way this code will be changed to handle this case.

wburns:
Actually it could happen in core if it expires between the `EntryWrappingInterceptor` and the `CacheLoaderInterceptor` so it is quite a tight window.

wburns:
The only way I can see this happening now is you have 2 concurrent threads

t1 -> `EntryWrappingInterceptor` finds no entry in data container
t2 -> `EntryWrappingInterceptor` finds no entry in data container
t2 -> Loads entry from loader and puts in data container with "short" expiration period
t1 -> Reads from data container in PersistenceUtil and finds entry expired and doesn't consult loader

The time window for this is extremely small, I am surprised your use case would run into this.

I was able to confirm via test that this is a possibility and the new changes will fix this.

arkaev:
Teo days ago I reproduce this bug on one thread, but with 3-4 same cache requests.

It looks like first time cache internally get expired element from previous load, return null and cache this null value for next expiration interval.

I can try to get more information if you need, but it's not too easy to handle problem in debug mode

wburns:
Even after the fix in this PR? Before this PR I can see a single thread causing an issue, but after this one, it should be just concurrent threads.

wburns:
The fix should be in https://github.com/infinispan/infinispan/pull/6146 if you wanted to try manually btw @arkaev

arkaev:
Thanks! I'll check later if it helps

wburns:
We have a build on Friday the 13th for `9.3.1.Final` which #6146 should get into if it gets merge to master. If you are able to confirm it doesn't help, then I might be able to fix it before the build.

PR #6072, opened 2018-06-15
ISPN-9016 Delete temporary SingleFileStore .dat files after tests

danberindei:
https://issues.jboss.org/browse/ISPN-9016 removed the gitignore for *.dat files, and this exposed a problem in ExpirationSingleFileStoreListenerFunctionalTest

I initially thought the stale files were from SingleFileStoreFunctionalTest, so I changed it to skip the cache instantiation in the parse tests as well.


wburns:

PR #6073, opened 2018-06-15
ISPN-8431 ScatteredSplitAndMergeTest random failures fix

rvansa:
https://issues.jboss.org/browse/ISPN-8431

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
@@ -67,6 +67,7 @@ import org.infinispan.commons.util.EnumUtil;
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.container.entries.InternalCacheValue;
 import org.infinispan.container.entries.RemoteMetadata;
+import org.infinispan.container.impl.EntryFactory;
 import org.infinispan.container.versioning.EntryVersion;
 import org.infinispan.container.versioning.InequalVersionComparisonResult;
 import org.infinispan.context.Flag;
@@ -107,6 +108,7 @@ public class PrefetchInterceptor<K, V> extends DDAsyncInterceptor {
    @Inject protected CommandsFactory commandsFactory;
    @Inject protected RpcManager rpcManager;
    @Inject protected Cache<K, V> cache;
+   @Inject protected EntryFactory entryFactory;
 
    protected int numSegments;
 
@@ -233,6 +235,10 @@ public class PrefetchInterceptor<K, V> extends DDAsyncInterceptor {
          if (maxValue == null) {
             return;
          }
+         // The put below could fail updating the context if the data container got the updated value while we were
+         // prefetching that. Also, we need to call RepeatableReadEntry.updatePreviousValue() to get return value
+         // from the main command correct.
+         entryFactory.wrapExternalEntry(ctx1, cmd.getKey(), maxValue.toInternalCacheEntry(cmd.getKey()), true, true);
          PutKeyValueCommand putKeyValueCommand =
                commandsFactory.buildPutKeyValueCommand(cmd.getKey(), maxValue.getValue(), segment, maxValue.getMetadata(),
                                                        STATE_TRANSFER_FLAGS);
@@ -338,6 +344,12 @@ public class PrefetchInterceptor<K, V> extends DDAsyncInterceptor {
          }
          if (map.isEmpty()) {
             return CompletableFutures.completedNull();
+         }
+         // The put below could fail updating the context if the data container got the updated value while we were
+         // prefetching that. Also, we need to call RepeatableReadEntry.updatePreviousValue() to get return value
+         // from the main command correct.
+         for (Map.Entry<Object, InternalCacheValue> entry : map.entrySet()) {
+            entryFactory.wrapExternalEntry(ctx, entry.getKey(), entry.getValue().toInternalCacheEntry(entry.getKey()), true, true);
          }
          PutMapCommand putMapCommand = commandsFactory.buildPutMapCommand(map, null, STATE_TRANSFER_FLAGS);
          putMapCommand.setTopologyId(topologyId);
--- a/core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
@@ -27,7 +27,6 @@ import org.infinispan.commands.AbstractTopologyAffectedCommand;
 import org.infinispan.commands.CommandsFactory;
 import org.infinispan.commands.DataCommand;
 import org.infinispan.commands.FlagAffectedCommand;
-import org.infinispan.commands.SegmentSpecificCommand;
 import org.infinispan.commands.TopologyAffectedCommand;
 import org.infinispan.commands.VisitableCommand;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
```diff
@@ -194,50 +196,44 @@ protected Object handleReadCommand(InvocationContext ctx, DataCommand command) t
...
-   }
-
-   private Object handleLocallyLookedUpEntry(InvocationContext ctx1, VisitableCommand command1, Object rv) {
-      GetCacheEntryCommand cmd = (GetCacheEntryCommand) command1;
-      if (trace) {
-         log.tracef("Locally prefetched entry %s", rv);
+   private InvocationStage lookupLocalAndRetrieveRemote(InvocationContext ctx, Object key, TopologyAffectedCommand cmd, int segment) {
+      // We need to lookup the dataContainer directly as GetCacheEntryCommand won't return entry with null value
+      InternalCacheEntry entry = dataContainer.get(key);
+            if (trace) {
```
danberindei: Indent looks wrong

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
```diff
@@ -194,50 +196,44 @@ protected Object handleReadCommand(InvocationContext ctx, DataCommand command) t
...
       } else {
-         future = retrieveRemoteValue(null, cmd.getKey(), segment, cmd);
+         future = retrieveRemoteValue(null, key, segment, cmd);
       }
       return asyncValue(future.thenAccept(maxValue -> {
          if (maxValue == null) {
             return;
          }
+         // The put below could fail updating the context if the data container got the updated value while we were
+         // prefetching that. Also, we need to call RepeatableReadEntry.updatePreviousValue() to get return value
```
danberindei: There's no call to `updatePreviousValue()` in `PrefetchInterceptor`, so I'm not sure how to read this

rvansa: `wrapExternalEntry` calls `updatePreviousValue()` for you. The thing is that we must add the prefetched value to context as the state-transfer-put would not do that if the version is not higher than the one in DC (which happens when we've received the entry in the meantime).

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
```diff
@@ -196,54 +196,44 @@ protected Object handleReadCommand(InvocationContext ctx, DataCommand command) t
...
          // The put below could fail updating the context if the data container got the updated value while we were
-         // prefetching that. Also, we need to call RepeatableReadEntry.updatePreviousValue() to get return value
+         // prefetching that (then the version would not be higher than the on in DC).
+         // We need to call RepeatableReadEntry.updatePreviousValue() (through wrapExternalEntry) to get return value
          // from the main command correct.
-         entryFactory.wrapExternalEntry(ctx1, cmd.getKey(), maxValue.toInternalCacheEntry(cmd.getKey()), true, true);
-         PutKeyValueCommand putKeyValueCommand =
-               commandsFactory.buildPutKeyValueCommand(cmd.getKey(), maxValue.getValue(), segment, maxValue.getMetadata(),
-                                                       STATE_TRANSFER_FLAGS);
+         entryFactory.wrapExternalEntry(ctx, key, maxValue.toInternalCacheEntry(key), true, true);
```
danberindei: Sorry Radim, but every time you add more info it raises more questions for me :(

* Who checks that the version in the PKVC is higher than the version in the DC?
* What happens if the version in DC increases after `PrefetchInterceptor`?
* This code runs before EWI. So why do we have to worry about the previous value specifically, do we expect the context to already have a previous value for this key that's wrong?

rvansa: 1) ScatteredDistributionInterceptor
2) The value from put-for-state-transfer is not written in SDI. It is not possible that the version in DC increases in other way that receiving the most recent value, because there is always only one node that can write it and assign a version number (this is synced using `CONFIRM_REVOKED SEGMENTS`).
3) EWI cannot read the value because there's no read owner at this point. `PrefetchInterceptor` retrieves the value from all nodes and picks the highest version, or uses `RemoteMetadata` to retrieve it from single node if the location of the highest version is already known.

danberindei:
I'm afraid PrefetchTest is failing:

https://ci.infinispan.org/job/Infinispan/job/PR-6073/13/#showFailuresLink

danberindei:
Yeah, it's pretty weird that it would be sent to other nodes with `CACHE_MODE_LOCAL`. To help debugging, I suggest you remove this line, so that `RpcManager` sees the missing topology id and throws an exception on the spot:

https://github.com/danberindei/infinispan/blob/55b67faf1c8e20a3ecce3204efbe50d27b1c5c4d/core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java#L343

rvansa:
It is safe to ignore the PutMapCommand in ControlledRpcManager, therefore I am removing the Changes Required label. Still trying to reproduce the `ScatteredCrashInSequenceTest` failure (800 test runs **without** trace logging was not enough :-/) but that should not block this PR imo. I'll freak out if CI will see that failure again...

rvansa:
@danberindei Do you have any further issue with this?

core/src/main/java/org/infinispan/interceptors/impl/PrefetchInterceptor.java
```diff
@@ -194,50 +199,45 @@ protected Object handleReadCommand(InvocationContext ctx, DataCommand command) t
...
+         // The put below could fail updating the context if the data container got the updated value while we were
+         // prefetching that (then the version would not be higher than the on in DC).
+         // We need to call RepeatableReadEntry.updatePreviousValue() (through wrapExternalEntry) to get return value
+         // from the main command correct.
+         entryFactory.wrapExternalEntry(ctx, key, maxValue.toInternalCacheEntry(key), true, true);
+         PutKeyValueCommand putKeyValueCommand = commandsFactory.buildPutKeyValueCommand(
+               key, maxValue.getValue(), segment, maxValue.getMetadata(), STATE_TRANSFER_FLAGS);
          putKeyValueCommand.setTopologyId(cmd.getTopologyId());
-         invokeNext(ctx1, putKeyValueCommand);
+         invokeNext(ctx, putKeyValueCommand);
```
rvansa: Oops, this is probably a bug - if the command blocks in this call we're going to return too early.

danberindei: I think we may have discussed this, but back then you thought it couldn't block :)

I think you can leave the "naked" `invokeNext` if you remove the `setTopologyId` line above, so that the command can't be sent remotely.

rvansa:
And now with more InvocationStages and less futures!

Actually I've fixed not returning value from `invokeNext` - we don't care about the value but should wait for the command to complete. And since I've backported the change from my [ISPN-7889](https://issues.jboss.org/browse/ISPN-7889) branch I use the invocation stage since there's some locking magic that happens only when you write `asyncValue(future).thenApply(ctx, command, ...)` instead of `asyncValue(future.thenApply(...))`.

danberindei:
Integrated, thanks Radim! Sorry for the wait, but I do like it better now :)

PR #6074, opened 2018-06-15
ISPN-9289 Ignore TransactionSynchronizationRegistry

pruivo:
* Fix test


rvansa:
Could you explain what was wrong?

pruivo:
@rvansa you can't use the word `default` in the caches names :)

danberindei:
@pruivo where does that restriction come from?

tristantarrant:
I believe it is a problem with the infinispan arquillian machinery which is matching by prefix

pruivo:
it is caused by https://github.com/infinispan/infinispan-arquillian-container/blob/master/infinispan-impl/src/main/java/org/infinispan/arquillian/utils/MBeanUtils.java#L118
It tries to find all the MBeans with cache's name `default` in its name. That includes the `default` cache name and the transactional caches I did create.

PR #6082, opened 2018-06-21
ISPN-9305 NullPointerException in cache.size() after cache.containsKey()

danberindei:
https://issues.jboss.org/browse/ISPN-9305
https://issues.jboss.org/browse/ISPN-9314

* Replace InvocationContext.getLookedUpEntries() with
  forEachLookedUpEntry(BiConsumer) and lookedUpEntriesCount(),
  giving access to the key even for entries wrapped as NullCacheEntry
* Exclude null entries from distributed streams
* Use a non-tx context for AtomicKeySetImpl deserialization


core/src/test/java/org/infinispan/stream/DistributedStreamIteratorRepeatableReadTxTest.java
```diff
@@ -134,4 +131,25 @@ public void testKeyFilterConverterWithExistingTransaction() throws NotSupportedE
          tm.rollback();
       }
    }
+
+   public void testStreamWithMissedKeyInTransaction() throws Exception {
+      AdvancedCache<Object, String> cache = advancedCache(0, CACHE_NAME);
+      TransactionManager tm = TestingUtil.extractComponent(cache, TransactionManager.class);
```
pruivo: what is wrong with `cache.getTransactionManager()`

danberindei: I didn't think about it, I just copied the code from the method above. Let me change it everywhere.

core/src/main/java/org/infinispan/context/impl/NonTxInvocationContext.java
```diff
@@ -56,6 +57,18 @@ public void putLookedUpEntry(Object key, CacheEntry e) {
                    Collections.emptyMap() : lookedUpEntries);
    }
 
+   @Override
+   public void forEachEntry(BiConsumer<Object, CacheEntry> consumer) {
```
pruivo: most of the places iterate over `!entry.isRemoved() && !entry.isNull()`. 
should we add a method that only iterates over valid entries?

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
```diff
@@ -630,6 +630,7 @@ protected Object getKey(CacheEntry<K, V> value) {
       private final TxInvocationContext<LocalTransaction> ctx;
       // We store all the not yet seen context entries here.  We rely on the fact that the cache entry reference is updated
       // if a change occurs in between iterations to see updates.
+      // TODO Dan: Only true with REPEATABLE_READ isolation level, not true with READ_COMMITTED
```
pruivo: what is the point here? with RC, that change happens in the context as well.

danberindei: Not always. With RC, a read will store the `InternalCacheEntry` in the context as is. If the same tx modifies the entry, a new `ReadCommittedEntry` is created and stored in the context instead -- the old ICE is not modified until commit (and even then, only if the entry is local).

core/src/main/java/org/infinispan/context/impl/NonTxInvocationContext.java
danberindei: Ok, makes sense

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
pruivo: well, RC doesn't work if the entry is remote. I'm planning to fix it by removing the entry from the context if it isn't modified. that way, it would force to read it from DataContainer/CacheLoader/Remote node. 

so, should we expose some kind of iterator in `EntryLookup` instead of using a snapshot?

danberindei: @pruivo I'm not sure what you mean, this queue is maintained by `TxInterceptor` and not by `EntryLookup` implementations.

danberindei:
@pruivo Updated, the latest build only failed because of https://github.com/infinispan/infinispan/pull/6088

core/src/main/java/org/infinispan/context/EntryLookup.java
```diff
@@ -24,10 +25,48 @@
...
+    *
+    * @deprecated Since 9.3, please use {@link #forEachEntry(BiConsumer)} or {@link #lookedUpEntriesCount()} instead.
     */
+   @Deprecated
    Map<Object, CacheEntry> getLookedUpEntries();
 
+   /**
+    * Execute an action for each value in the context.
+    *
+    * Entries that do not have a value (because the key was removed, or it doesn't exist in the cache).
```
pruivo: seems you didn't complete the sentence

core/src/main/java/org/infinispan/interceptors/impl/TxInterceptor.java
pruivo: we can discuss this later. I triggered another CI run and lets integrate this :)

core/src/main/java/org/infinispan/stream/impl/tx/TxClusterStreamManager.java
```diff
@@ -106,7 +106,7 @@ public void forgetOperation(Object id) {
          Supplier<Map.Entry<Address, IntSet>> segments, Set<K> keysToInclude, IntFunction<Set<K>> keysToExclude,
          boolean includeLoader, boolean entryStream, Iterable<IntermediateOperation> intermediateOperations) {
 
-      if (ctx.getLookedUpEntries().isEmpty()) {
+      if (ctx.lookedUpEntriesCount() != 0) {
```
wburns: Just happen to be peeking around and found this is a bug, this should really be `== 0` instead.

danberindei: Thanks for fixing it in https://github.com/infinispan/infinispan/pull/7016 @wburns !

PR #6089, opened 2018-06-21
ISPN-9286 WriteBehindFaultToleranceTest.testBlockingOnStoreAvailabili…

ryanemerson:
…tyChange fix

https://issues.jboss.org/browse/ISPN-9286

### DIFF
--- a/core/src/test/java/org/infinispan/persistence/WriteBehindFaultToleranceTest.java
+++ b/core/src/test/java/org/infinispan/persistence/WriteBehindFaultToleranceTest.java
@@ -7,6 +7,7 @@ import static org.testng.AssertJUnit.assertNotNull;
 import static org.testng.AssertJUnit.assertTrue;
 
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
@@ -22,6 +23,7 @@ import org.infinispan.persistence.async.AdvancedAsyncCacheWriter;
 import org.infinispan.persistence.dummy.DummyInMemoryStore;
 import org.infinispan.persistence.dummy.DummyInMemoryStoreConfigurationBuilder;
 import org.infinispan.persistence.manager.PersistenceManager;
+import org.infinispan.persistence.manager.PersistenceManagerImpl;
 import org.infinispan.persistence.spi.StoreUnavailableException;
 import org.infinispan.test.AbstractInfinispanTest;
 import org.infinispan.test.Exceptions;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/infinispan/persistence/WriteBehindFaultToleranceTest.java
```diff
@@ -60,11 +60,10 @@ public void testBlockingOnStoreAvailabilityChange() {
       store.setAvailable(false);
       assertFalse(store.isAvailable());
 
-      TestingUtil.sleepThread(AVAILABILITY_INTERVAL * 5);
       // PM & AsyncWriter should still be available as the async modification queue is not full
       assertTrue(asyncWriter.isAvailable());
       assertFalse(TestingUtil.extractField(asyncWriter, "delegateAvailable"));
-      assertTrue(pm.isAvailable());
+      eventually(pm::isAvailable);
```
danberindei: Doesn't the fact that we have to wait for the PM to become available contradict the comment above?

ryanemerson: Yes you're right. I was obviously in auto-pilot when I changed this line.

danberindei:
@ryanemerson There are 2 more instances of `TestingUtil.sleepThread(AVAILABILITY_INTERVAL * 5)`, and sleeping for 50ms really doesn't guarantee anything in CI :)

ryanemerson:
@danberindei Good point. I've updated the test to remove the remaining offenders.

core/src/test/java/org/infinispan/persistence/WriteBehindFaultToleranceTest.java
```diff
@@ -49,43 +51,42 @@
...
 
       // Add entries >= modification queue size so that store is no longer available
       cache.putAll(intMap(0, 10));
       assertEquals(1, store.size());
 
-      TestingUtil.sleepThread(AVAILABILITY_INTERVAL * 5);
       // PM and writer should not be available as the async modification queue is now oversubscribed and the delegate is still unavailable
       assertFalse(asyncWriter.isAvailable());
-      assertFalse(pm.isAvailable());
+      eventually(() -> !pm.isAvailable());
```
danberindei: Sorry, but I don't think this is right: IMO the new `eventually` calls should be exactly where the `sleepThread` calls were

ryanemerson: done

danberindei: @ryanemerson I should have been more clear: I was thinking about all the new `eventually` calls (except for `eventually(() -> 10 == store.size());`, that one's fine where it is) :)

ryanemerson: @danberindei I think that's what I have done, except one of the `Thread.sleep` calls was in fact redundant so I don't need to replace with an eventually call.

danberindei: Ok, when I looked it seemed like you only changed this instance. Merging...

danberindei:
Integrated, thanks Ryan!

PR #6090, opened 2018-06-21
PolarionJUnitXMLReporter based on JUnitReportReporter

diegolovison:


diegolovison:
The idea to open this PR is that I need your CI running.

PR #6094, opened 2018-06-26
ISPN-8431 Fix test name in @Test annotation

danberindei:
Forgot to check the CI results before merging #6064 and the build is now failing because the testName is incorrect.


Sanne:
thanks! merged

PR #6095, opened 2018-06-26
JTA API upgrade

Sanne:
Couldn't run the testsuite locally as the testsuite is broken... I won't comment on that, let's just say it's on you all to take this PR futher then :)

 - https://issues.jboss.org/browse/ISPN-9324


pruivo:
integrated! thanks @Sanne !

PR #6096, opened 2018-06-26
Deps fix

Sanne:
I did not create a JIRA or include a JIRA reference as I'm fixing an issue debated on team chat but I don't know if there is a related JIRA... please feel free to edit the commits and merge, I am unlikely to have time for that myself tomorrow.

Fundamentally the problem is that the BOM enforces Dependency Management, and thus overrides components which are defined by the WildFly feature pack - we really need to test with whatever library versions come with WildFly, at least when it comes to those modules which are meant to be used as additional modules for a vanilla WildFly server.

I suggest most modules should not inherit from the BOM, making the BOM import optional and making it easier for each module to opt-in (or not) to the global version enforcement.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -371,26 +371,6 @@
          </dependency>
          <dependency>
             <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-hibernate-cache-spi</artifactId>
-            <version>${project.version}</version>
-         </dependency>
-         <dependency>
-            <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-hibernate-cache-commons</artifactId>
-            <version>${project.version}</version>
-         </dependency>
-         <dependency>
-            <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-hibernate-cache-v51</artifactId>
-            <version>${project.version}</version>
-         </dependency>
-         <dependency>
-            <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-hibernate-cache-v53</artifactId>
-            <version>${project.version}</version>
-         </dependency>
-         <dependency>
-            <groupId>org.infinispan</groupId>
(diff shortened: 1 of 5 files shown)


danberindei:
@Sanne, @diegolovison created an issue yesterday: https://issues.jboss.org/browse/ISPN-9325

Is it still necessary to revert the parallel build fix, or is it more like the fix is not needed any more?

gustavonalle:
Parallel builds still fail with 
```
[ERROR] Failed to execute goal org.wildfly.build:wildfly-server-provisioning-maven-plugin:1.2.10.Final:build (server-provisioning-node2) on project infinispan-as-lucene-integration: Execution server-provisioning-node2 of goal org.wildfly.build:wildfly-server-provisioning-maven-plugin:1.2.10.Final:build failed: java.lang.RuntimeException: java.io.FileNotFoundException: /home/gfernandes/github/infinispan-review/query/target/classes (Is a directory) -> [Help 1]
```

ryanemerson:
Actually there might be a simpler way (although we may want Sanne's fixes anyway). I missed the fact that attach-license was the phase that was failing. This can be fixed by simply skipping this phase as we don't need it for the tests anyway. See https://github.com/infinispan/infinispan/pull/6097

danberindei:
I don't know enough about the provisioning plugin to comment on whether it's a good solution or not.

But I'm not convinced by the commit comment: applications can override all versions defined in the BOM, so I'm not sure if the change really helps others.

ryanemerson:
I just checked and https://github.com/infinispan/infinispan/pull/6097 also works with parallel builds

gustavonalle:
@ryanemerson could you also revert my previous commit on #6097 to see if both sequential build (on an empty repo) and parallel work? If it does, then I think we can close this issue and integrate #6097

danberindei:
I integrated Ryan's #6097 instead.

Sanne:
awesome, thanks all!

PR #6100, opened 2018-06-27
ISPN-9328 Release build doesn't commit version change before tagging

danberindei:
https://issues.jboss.org/browse/ISPN-9328

### DIFF
--- a/Jenkinsfile-release
+++ b/Jenkinsfile-release
@@ -64,7 +64,8 @@ pipeline {
 
         stage('Tag') {
             steps {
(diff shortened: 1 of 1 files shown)


danberindei:
@tristantarrant we also need to decide what to do with the `9.3.0.Final` tag. I suggest we delete the tag and then run the `versions:set`, `scm:checkin`, and `scm:tag` steps manually.

rhusar:
@tristantarrant @danberindei   This is missing from 9.3.x branch, the problem is will happen again...

PR #6101, opened 2018-06-28
ISPN-9329 Fix NPE during cache start due to circular component dependencies

pferraro:
https://issues.jboss.org/browse/ISPN-9329


danberindei:
@pferraro I was chatting with @wburns about this, I think the dependency cycle should be broken in `GroupManagerImpl`. Will volunteered to make the change, but feel free to update your PR instead ;)

pferraro:
@danberindei I'll just update this PR.

wburns:
I should have a PR here in a few minutes, so many distractions this morning :)

pferraro:
@wburns OK - will let you handle it. :)

PR #6102, opened 2018-06-28
ISPN-9327 Always use UTF-8 in memcached server and tests

danberindei:
https://issues.jboss.org/browse/ISPN-9327

The test is failing because both the test and the server assume the platform encoding is UTF-8. E.g. the server uses UTF-8 when reading the key, and the platform encoding when writing it back.

On this machine the platform encoding is US-ASCII, and that breaks the test and the server.

@tristantarrant do you think it's useful for the server to use the platform encoding instead? (We'd have to change the way we read keys as well.)

Could you also explain a bit what kind of buffer overflow we are testing? If it's the buffer replay in the server, I think it would be better to include this in every test by using `SingleByteFrameDecoderChannelInitializer` to force a replay on every byte, without a separate test.


tristantarrant:
The only sane configuration is the one where we use UTF-8 throughout and ignore platform encoding. I have bad memories of all those databases in weird encodings. As for the test, I have no recollection of the specifics.

PR #6103, opened 2018-06-28
ISPN-9329 NPE during cache start due to circular component dependencies

wburns:
* Don't wrap custom IDC
* Move circular dependency into GroupManagerImpl instead

https://issues.jboss.org/browse/ISPN-9329

### DIFF
--- a/core/src/main/java/org/infinispan/distribution/group/impl/GroupManagerImpl.java
+++ b/core/src/main/java/org/infinispan/distribution/group/impl/GroupManagerImpl.java
@@ -16,73 +16,41 @@ import org.infinispan.commons.util.Util;
 import org.infinispan.distribution.DistributionManager;
 import org.infinispan.distribution.group.Group;
 import org.infinispan.distribution.group.Grouper;
+import org.infinispan.factories.ComponentRegistry;
 import org.infinispan.factories.annotations.Inject;
+import org.infinispan.factories.annotations.Start;
 import org.infinispan.remoting.transport.Address;
 
 public class GroupManagerImpl implements GroupManager {
 
-    private interface GroupMetadata {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/infinispan/distribution/group/impl/GroupManagerImpl.java
```diff
@@ -62,7 +64,8 @@ else if (possibleMethods.size() == 1)
             throw new IllegalStateException(Util.formatString("Cannot define more that one @Group method for class hierarchy rooted at %s", clazz.getName()));
     }
 
-    @Inject private DistributionManager distributionManager;
+    @Inject private ComponentRegistry componentRegistry;
+    private DistributionManager distributionManager;
```
danberindei: Please move the comment about the dependency cycle here

wburns: Sure

wburns:
Updated

wburns:
Imports should be fixed now as well.

pferraro:
@wburns FYI - while you're touching this class, it looks like DefaultSegmentedDataContainer contains a unnecessary @Listener annotation and calls cache.removeListener(...) during @Stop, even though it has no corresponding listener registration nor does it handle any events.

wburns:
It is harmless. I actually already fixed it in a different upcoming PR https://github.com/wburns/infinispan/commit/9a5de439afcc58f08008222b3d43c8b6034489ab#diff-6d6afaa6691141286cacb1e0057e6dcf

pferraro:
I realize it's harmless - it was just a little confusing. :)

danberindei:
Integrated, thanks Will!

PR #6104, opened 2018-06-28
ISPN-9333 Delete destination file before rename

rigazilla:
On windows File.renameTo() fails if destination file exists. Quick fix provided.

https://issues.jboss.org/browse/ISPN-9333

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/util/Util.java
+++ b/commons/src/main/java/org/infinispan/commons/util/Util.java
@@ -20,6 +20,8 @@ import java.lang.reflect.Modifier;
 import java.net.Socket;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileLock;
+import java.nio.file.Files;
+import java.nio.file.StandardCopyOption;
 import java.text.NumberFormat;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
@@ -33,7 +35,6 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
-import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 
 import javax.naming.Context;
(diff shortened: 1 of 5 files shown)


pruivo:
@danberindei @rigazilla should we use `move()` instead? https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#move

rigazilla:
@pruivo yeah I picked up a very old solution (let me run a test with move).
@danberindei ok? It seems that you prefers a different approach, but I'm not sure I understand.

rigazilla:
btw the renameTempFile is called only twice in the code, so the solution proposed by @danberindei can be also easily implemented. Or maybe we can just rename the method to suggest that an overwrite is ongoing

pruivo:
-1 
I think the operation should be atomic (or closest to that). Also, the file is updated under the a lock. you shouldn't remove it before invoke the method. 
I prefer the` move()` approach.

danberindei:
@pruivo -1 to failing the move if the target exists? 

BTW, I'm not sure how a move could not be atomic: it either succeeds or it fails, it can't move half the file.

pruivo:
@danberindei yes. the target is always there (except in the first run). Every time a cache-configuration or a counter-configuration is defined, the file is updated.
Also, I don't like the idea of "the caller delete the file before rename it"

commons/src/main/java/org/infinispan/commons/util/Util.java
```diff
@@ -1108,8 +1110,11 @@ public static void renameTempFile(File tempFile, File lockFile, File dstFile, Bi
       FileLock lock = null;
       try (FileOutputStream lockFileOS = new FileOutputStream(lockFile)) {
          lock = lockFileOS.getChannel().lock();
-         if (!tempFile.renameTo(dstFile)) {
-            renameFailed.accept(tempFile, dstFile);
+         try {
+             Files.move(tempFile.toPath(), dstFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
+         }
+         catch (Exception ex) {
```
pruivo: the `catch` should be in the same line as` } `
also, it would be nice to log the exception as well :)

rigazilla: It seems that the renameFailed biconsumer is handling the error and already logs everything. Should we also log the exception?

pruivo: yes.

danberindei:
Integrated, thanks Vittorio!

PR #6109, opened 2018-06-30
fix code highlight of Hot Rod Transaction document

kazuhira-r:


pruivo:
integrated! Thank you @kazuhira-r !

PR #6110, opened 2018-06-30
add javadoc, Hot Rod Transaction properties

kazuhira-r:
I added the following two properties to the ConfigurationBuilder Javadoc of Hot Rod Client.

* infinispan.client.hotrod.transaction.transaction_manager_lookup
* infinispan.client.hotrod.transaction.transaction_mode

It is a property added by Hot Rod Transaction, so I think that it was better to have a description in Javadoc.


galderz:
@pruivo Can you check?

pruivo:
integrated! Thank you @kazuhira-r !

PR #6111, opened 2018-07-02
Docs: Document how to pass in a custom JChannel to the transport conf…

tristantarrant:


galderz:
That's an interesting JIRA number ;)

documentation/src/main/asciidoc/user_guide/configuration.adoc
```diff
@@ -252,6 +253,22 @@ GlobalConfiguration globalConfig = new GlobalConfigurationBuilder().transport()
...
+If you want to construct the JGroups link:http://www.jgroups.org/manual4/index.html#JChannel[JChannel] by yourself, you can do so.
+
+NOTE: The JChannel must not be already connected.
+
+[source,java]
+----
+GlobalConfigurationBuilder global = new GlobalConfigurationBuilder();
+JChannel jchannel = new JChannel();
+// Configure the jchannel to your needs.
+JGroupsTransport transport = new JGroupsTransport(jchannel);
```
pruivo: we have a JGroupsChannelLookup, but I guess it isn't easier to use when compared with your example :) https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/remoting/transport/jgroups/JGroupsChannelLookup.java

tristantarrant:
JGroupsChannelLookup is far too complex for most user's needs, as it is tailored for WildFly which needs to control the lifecycle of the channel too.

wburns:
Maybe just add a JIRA and we can integrate.

pruivo:
is this PR still valid? It needs rebase :)

ryanemerson:
@tristantarrant Can you rebase and add a jira number so that it makes 9.4.0.Final?

tristantarrant:
Closed in favor of #6314

PR #6120, opened 2018-07-04
[9.3.x] ISPN-9303 IndexingMetadata.getFieldMapping results in NPE if the type…

anistor:
… is not indexed

https://issues.jboss.org/browse/ISPN-9303


gustavonalle:
Why didn't you backport the test?

PR #6121, opened 2018-07-04
[9.2.x] ISPN-9303 IndexingMetadata.getFieldMapping results in NPE if the type is not indexed

anistor:
https://issues.jboss.org/browse/ISPN-9303


wburns:
Believe you can close this as we are only backporting to 9.3.x now. Correct, @tristantarrant ?

PR #6125, opened 2018-07-05
Expose a method for deriving the index name from cache name rather th…

anistor:
…an using a constant


wburns:
Whoops forgot to make sure it had a JIRA :D

anistor:
Thanks Will!

PR #6131, opened 2018-07-06
ISPN-8905 Segment-aware non-shared cache stores

wburns:
https://issues.jboss.org/browse/ISPN-8905


core/src/main/java/org/infinispan/persistence/manager/PersistenceManager.java
```diff
@@ -222,4 +255,28 @@ void prepareAllTxStores(Transaction transaction, BatchModification batchModifica
...
     */
    boolean isAvailable();
+
+   /**
+    * Notifies any underlying segmented stores that the segments provided are owned by this cache and to start/configure
+    * any underlying resources required to handle requests for entries on the given segments.
+    * <p>
+    * This only affects stores that are not shared as shared stores have to keep all segments running at all times
+    * @param segments segments this cache owns
+    * @return if a configured store couldn't add segments (non segmented and non shared)
```
ryanemerson: It seems strange to me that we return true if segments can't be added.

core/src/main/java/org/infinispan/util/rxjava/FlowableFromIntSetFunction.java
```diff
@@ -0,0 +1,424 @@
...
+import io.reactivex.exceptions.Exceptions;
+import io.reactivex.internal.functions.ObjectHelper;
+import io.reactivex.internal.fuseable.ConditionalSubscriber;
+import io.reactivex.internal.subscriptions.BasicQueueSubscription;
+import io.reactivex.internal.subscriptions.EmptySubscription;
+import io.reactivex.internal.subscriptions.SubscriptionHelper;
+import io.reactivex.internal.util.BackpressureHelper;
+
+/**
+ * TODO: add docs
```
ryanemerson: Don't forget :smiley:

core/src/main/java/org/infinispan/util/rxjava/FlowableFromIntSetFunction.java
```diff
@@ -0,0 +1,424 @@
...
+      void fastPath() {
+         PrimitiveIterator.OfInt it = this.it;
+         IntFunction<T> intFunction = this.intFunction;
+         ConditionalSubscriber<? super T> a = actual;
+         for (;;) {
+            if (cancelled) {
+               return;
+            }
+
+            T t;
```
ryanemerson: Nitpick, but all of the whitespace in `fastPath` and `slowPath` seems excessive

core/src/test/java/org/infinispan/persistence/CacheLoaderFunctionalTest.java
```diff
@@ -86,6 +86,7 @@ protected ConfigurationBuilder getConfiguration() {
       ConfigurationBuilder cfg = new ConfigurationBuilder();
       cfg.persistence()
          .addStore(DummyInMemoryStoreConfigurationBuilder.class)
+            .segmented(true)
```
ryanemerson: Do we not want this test to run with segmented true and false?

ryanemerson:
@wburns Great work! There are a couple of NPE in the CacheLoader tests though

wburns:
Yeah I am still fixing the CacheLoader tests - I already know the problem, just couldn't think of a good way to fix it on Friday yet :)

core/src/main/java/org/infinispan/container/impl/DefaultSegmentedDataContainer.java
```diff
@@ -42,23 +40,21 @@
  * @author wburns
  * @since 9.3
  */
-@Listener
```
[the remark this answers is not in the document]
wburns: It was an artifact from before and isn't needed. It used to register itself as a listener, but that was removed and forgot to get this part :)

core/src/main/java/org/infinispan/interceptors/impl/CacheLoaderInterceptor.java
```diff
@@ -581,7 +574,7 @@ public boolean remove(Object o) {
          Publisher<CacheEntry<K, V>> publisher = flowable
                .map(me -> (CacheEntry<K, V>) PersistenceUtil.convert(me, iceFactory));
          // This way we don't subscribe to the flowable until after the first iterator is fully exhausted
-         return new LazyConcatIterator<>(localIterator, () -> org.infinispan.util.Closeables.iterator(publisher, 64));
+         return new LazyConcatIterator<>(localIterator, () -> org.infinispan.util.Closeables.iterator(publisher, 128));
```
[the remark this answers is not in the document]
wburns: It helps a bit yeah, and the numbers are so small it doesn't matter too much for this case.

core/src/main/java/org/infinispan/persistence/support/AbstractSegmentedAdvancedLoadWriteStore.java
```diff
@@ -0,0 +1,40 @@
...
+ * Abstract segment loader writer that implements all the single key non segmented methods by invoking the segmented
+ * equivalent by passing in the segment returned from invoking {@link #getKeyMapper()}. These methods are also all
+ * declared final as to make sure the end user does not implement the incorrect method. All other methods must be
+ * implemented as normal.
+ * @author wburns
+ * @since 9.4
+ */
+public abstract class AbstractSegmentedAdvancedLoadWriteStore<K, V> implements SegmentedAdvancedLoadWriteStore<K, V> {
+   public ToIntFunction<Object> getKeyMapper() {
+      throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
wburns: Yeah I can just make this abstract

core/src/main/java/org/infinispan/persistence/support/AbstractSegmentedAdvancedLoadWriteStore.java
```diff
@@ -0,0 +1,40 @@
...
+   }
+
+   @Override
+   public final MarshalledEntry<K, V> load(Object key) {
+      return load(getKeyMapper().applyAsInt(key), key);
```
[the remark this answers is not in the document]
wburns: The idea was going to be that no one should touch this implementation with a method that doesn't have a segment. But unfortunately we aren't in a position to guarantee this yet - I will just make it abstract as mentioned in the other.

core/src/main/java/org/infinispan/persistence/spi/SegmentedAdvancedLoadWriteStore.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * An interface implementing both {@link AdvancedCacheWriter} and {@link AdvancedCacheLoader} as well as overrides
+ * of those methods that can be optimized when a segment is already known for the key or a subset of segments are to
+ * be used instead of the entire store.
+ * @author wburns
+ * @since 9.4
+ */
+@ThreadSafe
+public interface SegmentedAdvancedLoadWriteStore<K, V> extends AdvancedLoadWriteStore<K, V> {
+   // CacheLoader methods
```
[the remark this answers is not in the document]
wburns: Sure, yeah I haven't gone through and added all the docs yet. Need to do that for everything still including user guide.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManager.java
wburns: The problem is normally this method would say true if the segments were added (the issue is there could be multiple stores). I can change it instead to return false if not all stores could add the given segment.

core/src/main/java/org/infinispan/util/rxjava/FlowableFromIntSetFunction.java
wburns: Yep, I still have to go through and do all of them.

core/src/main/java/org/infinispan/util/rxjava/FlowableFromIntSetFunction.java
wburns: This is because this is taken as an example from the Iterator Flowable class in rxjava2. Let me add that to the docs of this class.

core/src/test/java/org/infinispan/persistence/CacheLoaderFunctionalTest.java
wburns: We should

core/src/main/java/org/infinispan/persistence/spi/SegmentedAdvancedLoadWriteStore.java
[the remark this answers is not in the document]
wburns: Actually returning null is an implementation detail. It could be that the store computes some value.

wburns:
Code should be pretty stable, assuming it passes all the tests this time. Just have to add in documentation.

wburns:
Not exactly sure why the RocksDBStoreReaderTest was failing on CI. I reworked the way the segmented file locations are done to be more clean imo, hoping this cleans it up.

wburns:
Still not sure why the CI box keeps failing on the RocksDBReaderTest. I can't reproduce it on my local box. Need to look a bit closer.

wburns:
Just realized the store migrator can't just say segmented, it has to actually tell the target how many segments to create. Need to tweak that.

documentation/src/main/asciidoc/user_guide/persistence.adoc
```diff
@@ -241,6 +243,34 @@ This strategy is suited for cache stores with low latency and cheap operation co
...
+==== Segmented Stores
+
+A store can optionally be segmented. That is that data is stored by the segment
+that its key maps to. Information pertaining to segments and the ownership of
+them can be found at <<key_ownership,Key Ownership>>.
+
+Segmenting a store has the benefit of increasing the read performance when all
+of the data that maps is used. This would be used for any stream operation
+performed upon the cache, size method as well as preloading and state
+transfer in a cache.
```
ryanemerson: I think the first sentence of this paragraph needs clarifying. I'm not sure what you're trying to say with "all of the data that maps is used".

documentation/src/main/asciidoc/user_guide/persistence.adoc
```diff
@@ -241,6 +243,34 @@ This strategy is suited for cache stores with low latency and cheap operation co
...
+
+If you reconfigure the `numSegments` setting, the cache store must be migrated
+between to the new one config. This is due to having a different amount of segments
+causes entries to be mapped to different segments. Failure to do so may cause a loss
+of data.
+
+Not all stores support being segmented, please see the appropriate section
+for the store you are interested in to see if it supports being segmented.
+If you are interested in converting or writing a new store to support being
+segmented, please the SPI section below that talks about more details
```
ryanemerson: "please see the SPI" ?

core/src/test/java/org/infinispan/distribution/rehash/RehashAfterPartitionMergeTest.java
```diff
@@ -12,7 +12,7 @@
 import org.testng.annotations.Test;
 
 @Test(groups = "functional", testName =  "distribution.rehash.RehashAfterPartitionMergeTest")
-@InCacheMode({ CacheMode.DIST_SYNC, CacheMode.SCATTERED_SYNC })
+@InCacheMode({ /*CacheMode.DIST_SYNC,*/ CacheMode.SCATTERED_SYNC })
```
ryanemerson: Why are we ignoring DIST_SYNC? Or is this a leftover

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -679,11 +769,70 @@ public int size() {
...
+      storesMutex.readLock().lock();
+      try {
+         checkStoreAvailability();
+         for (CacheLoader l : loaders) {
+            if (l instanceof SegmentedAdvancedLoadWriteStore)
+               return ((SegmentedAdvancedLoadWriteStore) l).size(segments);
+         }
+         for (CacheLoader l : loaders) {
+            if (l instanceof AdvancedCacheLoader)
+               return ((AdvancedCacheLoader) l).size();
```
ryanemerson: I'm not sure this fits with the description in the javadoc, "how many entries are in the store which map to the given segments" as we are returning the total number of entries for non-segmented stores. To accurately return the number of entries associated with the specified segments, we would obviously have to iterate over all the entries in a non-segmented store, which is probably overkill. Does it make more sense to make this method only return a value for segment aware stores?

core/src/main/java/org/infinispan/persistence/PersistenceUtil.java
```diff
@@ -122,119 +119,50 @@ public static InternalMetadata internalMetadata(InternalCacheValue icv) {
...
-               metadata;
-         //noinspection unchecked
-         return factory.create(loaded.getKey(), loaded.getValue(), actual, metadata.created(), metadata.lifespan(),
-                               metadata.lastUsed(), metadata.maxIdle());
-      } else {
-         //metadata is null!
-         //noinspection unchecked
-         return factory.create(loaded.getKey(), loaded.getValue(), (Metadata) null);
-      }
+      return org.infinispan.persistence.internal.PersistenceUtil.convert(loaded, factory);
```
ryanemerson: Any reason for not importing `PersistenceUtil`?

wburns:
Updated, thanks for the additional comments @ryanemerson !

wburns:
Squashed commits together.

danberindei:
@wburns thinking about the open files, maybe we should reduce the default number of segments to 128, or even 64?

With the default numSegments=256 and numOwners=2, a node in a 4-node cluster would run 128 RocksDB instances *for each cache*. I wonder if RocksDB has performance tests with 1000 instances running on the same machine ;)

gustavonalle:
Couple of things I've found on the topic:

Apparently RocksDB supports logic partitioning of a single database aka "Column Family" https://github.com/facebook/rocksdb/wiki/Column-Families

Also, it seems to be possible to reduce resource consumption by aggregating resources used for each database:
https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide#sharing-cache-and-thread-pool

tristantarrant:
We definitely need to consider the option of "collapsing" multiple segments in a single RocksDB file, otherwise users with 10s/100s of caches will suffer

tristantarrant:
Also, the RocksDB cache store currently uses two databases: one for the entries and another for the expiry data. Using cf's we could store everything in a single DB.

danberindei:
@gustavonalle column families sound like exactly what we need!

tristantarrant:
RocksDB has a plethora of tuning options that we can support by allowing passthrough of string-based properties. I've created https://issues.jboss.org/browse/ISPN-9371 to track this.

ryanemerson:
@wburns CI failed due to Checkstyle "unused imports" on PersistenceUtil.

core/src/main/java/org/infinispan/commands/read/EntrySetCommand.java
```diff
@@ -106,8 +105,7 @@ public String toString() {
             return Closeables.iterator(dataContainer.iterator());
          }
          Iterator<CacheEntry<K, V>> iterator = new DataContainerRemoveIterator<>(cache, dataContainer);
-         RemovableIterator<CacheEntry<K, V>> removableIterator = new RemovableIterator<>(iterator, e -> cache.remove(e.getKey(), e.getValue()));
-         return new IteratorMapper<>(removableIterator, e -> new EntryWrapper<>(cache, e));
+         return new IteratorMapper<>(iterator, e -> new EntryWrapper<>(cache, e));
```
danberindei: I sense an opportunity to eliminate some more lambdas :) There are 3 more instances of `e -> new EntryWrapper<>(cache, e)`, so it would be better to write a new iterator that handles both `Iterator.remove()` and `Entry.setValue()`.

Ideally though, I'd like us to stop wrapping `EntrySet` in the interceptor chain: replace `EntrySetCommand`/`KeySetCommand`/`SizeCommand` with a single `IterateCommand` that returns a read-only `Iterator` /`Spliterator`/`Publisher` (whatever is easier to implement), and only wrap it into an entries/keys/values collection once, in `CacheImpl` (removing `DistributedBulkInterceptor` would also be a plus).

Off-topic, since the main topic is lambdas now, `DataContainerRemoveIterator` is too specific, it could use any `Iterator`.

wburns:
@gustavonalle great find! In that case I can rewrite rocksDB for Beta1 to support that form of segmentation instead.

wburns:
Fixed checkstyle error

wburns:
I created https://issues.jboss.org/browse/ISPN-9375 to handle the RocksDB column family conversion.

ryanemerson:
Good for me locally, hitting the button

PR #6132, opened 2018-07-09
ISPN-9352 Add nanoseconds stats for read/write/remove averages

tristantarrant:
https://issues.jboss.org/browse/ISPN-9352


tristantarrant:
Also fixed the Simple Cache stats to now really report milliseconds instead of nanos...

tristantarrant:
@ryanemerson added another commit to use correct types in the netty transport stats (an artifact of the Scala code)

tristantarrant:
Ok. CI is better now

PR #6133, opened 2018-07-09
Single port client

slaskawi:
https://issues.jboss.org/browse/ISPN-8756

This PR combines https://github.com/infinispan/infinispan/pull/6035 and provides implementation for Single Port support in the Hotrod client. The implementation works with TLS/ALPN (but a user must provide an OpenSSL library for Netty, like BoringSSL - see the docs) and, so called, Plain Text Upgrade (using `HTTP 1.1 Upgrade` header).

Since the tests need to use both Hotrod Server and Client, I had to put them in the server testsuite.

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/ConfigurationProperties.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/ConfigurationProperties.java
@@ -82,6 +82,7 @@ public class ConfigurationProperties {
    public static final String STATISTICS = ICH + "statistics";
    // Transaction properties
    public static final String TRANSACTION_MANAGER_LOOKUP = ICH + "transaction.transaction_manager_lookup";
+   public static final String SINGLE_PORT = "infinispan.client.hotrod.single_port";
    public static final String TRANSACTION_MODE = ICH + "transaction.transaction_mode";
    public static final String TRANSACTION_TIMEOUT = ICH + "transaction.timeout";
    // Near cache properties
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
@@ -6,6 +6,7 @@ import static org.infinispan.client.hotrod.impl.Util.await;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -28,6 +29,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import org.infinispan.client.hotrod.CacheTopologyInfo;
 import org.infinispan.client.hotrod.FailoverRequestBalancingStrategy;
@@ -43,6 +45,9 @@ import org.infinispan.client.hotrod.impl.consistenthash.ConsistentHashFactory;
 import org.infinispan.client.hotrod.impl.operations.OperationsFactory;
 import org.infinispan.client.hotrod.impl.protocol.Codec;
 import org.infinispan.client.hotrod.impl.protocol.HotRodConstants;
+import org.infinispan.client.hotrod.impl.transport.netty.singleport.CustomProtocolUpgradeCodec;
+import org.infinispan.client.hotrod.impl.transport.netty.singleport.SinglePortAdapter;
+import org.infinispan.client.hotrod.impl.transport.netty.singleport.SslHandlerFactory;
 import org.infinispan.client.hotrod.logging.Log;
 import org.infinispan.client.hotrod.logging.LogFactory;
 import org.infinispan.commons.marshall.Marshaller;
@@ -66,6 +71,7 @@ import net.jcip.annotations.ThreadSafe;
 public class ChannelFactory {
 
    public static final String DEFAULT_CLUSTER_NAME = "___DEFAULT-CLUSTER___";
+   private static final Collection<Integer> SINGLE_PORT_PORTS = Arrays.asList(80, 8080, 8081, 443, 8443, 8444);
    private static final Log log = LogFactory.getLog(ChannelFactory.class, Log.class);
    private static final boolean trace = log.isTraceEnabled();
    private static final CompletableFuture<ClusterSwitchStatus> NOT_SWITCHED_FUTURE = completedFuture(ClusterSwitchStatus.NOT_SWITCHED);
@@ -178,8 +184,26 @@ public class ChannelFactory {
       if (maxConnections < 0) {
          maxConnections = Integer.MAX_VALUE;
       }
-      ChannelInitializer channelInitializer = new ChannelInitializer(bootstrap, address, operationsFactory, configuration, this);
-      bootstrap.handler(channelInitializer);
+
+      SslHandlerFactory sslHandlerFactoryForChannelInitializer = null;
+      SslHandlerFactory sslHandlerFactoryForSinglePortAdapter = null;
+
+      if (configuration.security().ssl().enabled()) {
+         if (isSinglePort()) {
+            sslHandlerFactoryForSinglePortAdapter = alloc -> SslHandlerHelper.createSslHandler(configuration, alloc, "HR");
+         } else {
+            sslHandlerFactoryForChannelInitializer = alloc -> SslHandlerHelper.createSslHandler(configuration, alloc);
+         }
+      }
+
+      ChannelInitializer channelInitializer = new ChannelInitializer(bootstrap, address, operationsFactory, configuration, this, sslHandlerFactoryForChannelInitializer);
+      if (isSinglePort()) {
+         CustomProtocolUpgradeCodec hotRodUpgradeCodec = new CustomProtocolUpgradeCodec("HR", channelInitializer);
+         SinglePortAdapter singlePortAdapter = new SinglePortAdapter(sslHandlerFactoryForSinglePortAdapter, hotRodUpgradeCodec);
+         bootstrap.handler(singlePortAdapter);
+      } else {
+         bootstrap.handler(channelInitializer);
+      }
       ChannelPool pool = new ChannelPool(bootstrap.config().group().next(), address, channelInitializer, configuration.connectionPool().exhaustedAction(),
             configuration.connectionPool().maxWait(), maxConnections, configuration.connectionPool().maxPendingRequests());
       channelInitializer.setChannelPool(pool);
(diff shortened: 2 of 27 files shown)


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelInitializer.java
```diff
@@ -86,8 +79,6 @@ protected void initChannel(Channel channel) throws Exception {
          channel.pipeline().addLast("idle-state-handler",
                new IdleStateHandler(0, 0, configuration.connectionPool().minEvictableIdleTime(), TimeUnit.MILLISECONDS));
       }
-      ChannelRecord channelRecord = new ChannelRecord(unresolvedAddress, channelPool);
```
slaskawi: @rvansa Could you please review this and `ActivationFuture` changes a couple of lines below?

I believe we had a race condition here. The `ChannelRecord` entry was set in the `ChannelInitializer#initChannel` method but was used in `ActivationFuture#operationComplete`. I couldn't find any documentation that says that `initChannel` method is guaranteed to finish before a listener attached to channel activation is fired. As the matter of fact, I've been observing this failing pretty often when I added TLS/ALPN support which waits until the protocol is negotiated and *then* initializes the pipeline. This resulted in a `NullPointerException` thrown from `ChannelRecord.of(channel)`.

So I decided to move the `ChannelRecord` creation to the `ActivationFuture`. I ran the client testsuite a couple of times but it's so flaky that it's hard to detect if I broke anything or not. Please tell me what do you think about this change.

slaskawi:
Two failures are related. I fix them right away but they do not block the review process.

```
org.infinispan.server.test.client.router.SinglePortIT.testHttp2SwitchThroughUpgradeHeader
org.infinispan.server.test.client.router.SinglePortIT.testHttp2SwitchThroughALPN
```

slaskawi:
Rebased and fixed tests.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/ConfigurationProperties.java
```diff
@@ -70,6 +70,7 @@
    public static final String BATCH_SIZE = "infinispan.client.hotrod.batch_size";
    public static final String TRANSACTION_MANAGER_LOOKUP = "infinispan.client.hotrod.transaction.transaction_manager_lookup";
    public static final String TRANSACTION_MODE = "infinispan.client.hotrod.transaction.transaction_mode";
+   public static final String SINGLE_PORT = "infinispan.client.hotrod.single_port";
```
galderz: Please update `ConfigurationTest` to verify this is read/written correctly.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
```diff
@@ -177,8 +180,20 @@ private ChannelPool newPool(SocketAddress address) {
       if (maxConnections < 0) {
          maxConnections = Integer.MAX_VALUE;
       }
-      ChannelInitializer channelInitializer = new ChannelInitializer(bootstrap, address, operationsFactory, configuration, this);
-      bootstrap.handler(channelInitializer);
+
+      SslHandlerFactory sslHandlerFactoryForChannelInitializer = configuration.security().ssl().enabled() && !configuration.isSinglePort()?
+            alloc -> SslHandlerHelper.createSslHandler(configuration, alloc) : null;
+      SslHandlerFactory sslHandlerFactoryForSinglePortAdapter = configuration.security().ssl().enabled() && configuration.isSinglePort() ?
```
galderz: This code can be improved, with a single check of `isSinglePort` and do all you need inside there. With the current code you're checking this twice, in this lines and then shortly below, making it more confusing to read.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelInitializer.java
galderz: > it's so flaky

What's flaky about it? We're not seeing such flakyness on CI.

slaskawi:
> What's flaky about it? We're not seeing such flakyness on CI.

@galderz This part seems to rely on Netty internal ordering of events. When developing Single Port Client I noticed that it is possible that a connection upgrade finishes and the `ActivationFuture` is still "in flight". So in my opinion we are using a bit risky construct there which may change even in minor Netty upgrades.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelInitializer.java
slaskawi: This part seems to rely on Netty internal ordering of events. When developing Single Port Client I noticed that it is possible that a connection upgrade finishes and the ActivationFuture is still "in flight". So in my opinion we are using a bit risky construct there which may change even in minor Netty upgrades.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/ConfigurationProperties.java
slaskawi: Done

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
slaskawi: Unfortunately those objects are tightly coupled and decomposing them as you proposed won't be easy. Here's why...

The `ChannelInitializer` requires an `SslHandlerFactory`, so it means it needs to be initialized *after* it. The `CustomProtocolUpgradeCodec` needs a Netty pipeline that needs to plugged in after a successful upgrade. So again, `CustomProtocolUpgradeCodec` comes after `ChannelInitializer` (since it creates the pipeline). The `SinglePortAdapter` needs am upgrade codec (long story short - this is due to Netty's HTTP/1.1 design), so it comes after `CustomProtocolUpgradeCodec` and an `SslHandlerFactory` (because it can operate with or without SSL). Finally, the `ChannelPool` requires `ChannelInitializer` and vice versa. So here's the creation order:

1. `SslHandlerFactory`
2. `ChannelInitializer`
3. `CustomProtocolUpgradeCodec`
4. `SinglePortAdapter`
5. `ChannelPool`

I can swap if/else statements but I need to preserve exactly this order. That means I will be either branching by `configuration.isSinglePort()` or `configuration.security().ssl().enabled()`. This way or another the amount of if/else statements will remain the same (I've just checked it in my IDE). 

If you have any other idea how to refactor it, please paste me a code snippet (but please check it in your IDE before that....this one looks easy only on paper :D).

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelInitializer.java
rvansa: @slaskawi `initChannel` is called from `io.netty.channel.ChannelInitializer` when the channel is *registered*, while `ActivationFuture.operationCompleted` is called when the promise returned by `Bootstrap.connect()` is completed. This should be done only after the registration completes (see `Bootstrap.doResolveAndConnect()`)

rvansa: Not that I would have any issues with the way you did it instead, as long as it's non-blocking and testsuite passes.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
rvansa: I think Galder meant something like
```java
SslHandlerFactory sslHandlerFactoryForChannelInitializer = null;
SslHandlerFactory sslHandlerFactoryForSinglePortAdapter = null;
if (configuration.security().ssl().enabled()) {
   if (configuration.isSinglePort()) {
      sslHandlerFactoryForSinglePortAdapter = ...
   } else {
      sslHandlerFactoryForChannelInitializer = ...
   }
}

tristantarrant:
Why do we need a config attribute to enable single port? Can't this be
automatically detected by talking to the server?

rvansa:
@tristantarrant I was thinking along the same lines but if the client starts talking to Hot Rod-only port with `GET / HTTP/1.1` it's not going to work.

On the other hand we could enhance our HotRod-only server so that it handles correctly both HTTP connections and Hot Rod protocol connections, and make this flag `true` by default - when a new client talks to old server, the clients will have to set this off.

danberindei:
@slaskawi could we have a test that starts the server in-vm, to make it easier to debug?

tristantarrant:
@rvansa the client should be doing the heavy lifting: 

1. send HTTP/1.1 request (HEAD /)
2. if response is not HTTP, send a Hot Rod PING otherwise go to 3.
3. try to send an HTTP upgrade to Hot Rod
4. if 3 did not succeed, attempt a HTTP/2 upgrade

rvansa:
@tristantarrant I would expect the server to close the connection if it the format does not match, but you're right that it just spits errors until the magic byte comes and lets you continue from there.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelInitializer.java
slaskawi: Ok, agreed. Thanks @rvansa for the input!

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/netty/ChannelFactory.java
slaskawi: Ok, implemented.

slaskawi:
@rvansa 
> I miss (even basic) tests in client-hotrod module, integration tests are far after that.
And I'd like to see some assertions verifying that after the protocol upgrade any upgrade-related handlers are gone from the stack.

Unfortunately it can not be done. It's a chick/egg problem. The `router` module already uses `hotrod-client` module. And the only way to tell if Single Port really works is to use it on the server. Sorry for bad info guys...

@tristantarrant 
> Why do we need a config attribute to enable single port? Can't this be
automatically detected by talking to the server?

For the TLS part - yes, because we know what protocol are we going to talk because of ALPN extension. For, so called clear text upgrade (aka `HTTP/1.1 Upgrade`), no, we can not know that. That's why we need a configuration switch. I could use some brute force - always send HTTP message, but that would probably generate some warning or error on the server and upset the majority of our users who do not care about single port.

@danberindei 
> @slaskawi could we have a test that starts the server in-vm, to make it easier to debug?

As I said before, it's a chick/egg problem. I can not figure an easy way to do it without introducing a circular dependency.

@tristantarrant 
> @rvansa the client should be doing the heavy lifting:
send HTTP/1.1 request (HEAD /)
if response is not HTTP, send a Hot Rod PING otherwise go to 3.
try to send an HTTP upgrade to Hot Rod
if 3 did not succeed, attempt a HTTP/2 upgrade

I think that's a good case for another iteration of this feature. Maybe we should create an enhancement JIRA?

slaskawi:
I rebased the PR and slightly changed the automatic behavior according to the Infinispan DEV email thread: http://lists.jboss.org/pipermail/infinispan-dev/2018-December/018218.html

@tristantarrant PTAL

PR #6135, opened 2018-07-09
ISPN-9353 CacheMgmtInterceptor GetAllCommand exceptions

danberindei:
https://issues.jboss.org/browse/ISPN-9353


danberindei:
Updated.

I should have mentioned this in the description, but I was wondering whether we really need `invokeNextAndFinally` here, since an exception is neither a hit nor a miss. Maybe a separate counter for errors would be better, WDYT @wburns @tristantarrant @pruivo ?

wburns:
To me partition handling blocking a request isn't an error, so maybe we may want to add two stats? One for general errors and one for partition handling blocking a request. Could be a sign for them to try to fix their network :)

wburns:

danberindei:
@wburns I've created https://issues.jboss.org/browse/ISPN-9379

IMO there's no difference between getting an `AvailabilityException` and a `TimeoutException`, the admin should investigate both, but maybe more information would be helpful here.

danberindei:
I cherry-picked it to 9.3.x as well

PR #6136, opened 2018-07-10
ISPN-9357 404 returned from REST for caches with object storage and non…

gustavonalle:
…-string keys

https://issues.jboss.org/browse/ISPN-9357

### DIFF
--- a/server/integration/testsuite/src/test/java/org/infinispan/server/test/dataconversion/Currency.java
+++ b/server/integration/testsuite/src/test/java/org/infinispan/server/test/dataconversion/Currency.java
@@ -1,0 +1,44 @@
+package org.infinispan.server.test.dataconversion;
+
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.io.Serializable;
+
+import org.infinispan.commons.marshall.SerializeWith;
+
+/**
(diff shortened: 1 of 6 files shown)


server/integration/testsuite/src/test/java/org/infinispan/server/test/dataconversion/Currency.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.io.Serializable;
+
+import org.infinispan.commons.marshall.SerializeWith;
+
+/**
+ * @since 9.4
+ */
+@SerializeWith(value = Currency.Externalizer.class)
```
anistor: SerializeWith does not exist. According to documentation :)))

gustavonalle: This is not actually needed, I could just use ```implements Serializable```, but adding this allowed me to test  deploying entities that depends on a internal infinispan module

anistor: I was only joking. We can use that. It even works.

gustavonalle:
Included in #6137

PR #6147, opened 2018-07-13
ISPN-9293 Convert Set<Integer> places to IntSet where appropriate

wburns:
https://issues.jboss.org/browse/ISPN-9293


danberindei:
Integrated, thanks Will!

PR #6150, opened 2018-07-13
Separating User and Internal Marshaller

ryanemerson:
https://issues.jboss.org/browse/ISPN-7409

This is still very much a preview. I still need to rework the SerializationConfiguration so that custom marshallers have a way of performing initialisation/configuration of the user marshaller; I was considering a similar approach to store configuration. 

My approach of separating the two marshallers is as follows:
- All user key/values/metada are marshalled via a MarshalledEntryImpl object.
    - The global marshaller externalizers the MarshallEntryImpl object, but the user marshaller generates the bytes of the key/value/metadata
   - We can then still use the MarshalledEntryImpl externalizer ID to reduce the size of the payload for RPCs
- User marshallers only need to be able to marshall their types, plus potential metadata implementations
- RPC commands are now passed a MarshalledEntryFactory so that user values can be wrapped and marshalled for internal communication. 

In another branch I have a protostream based marshaller impl that works for our internal metadata classes and data structures, but still needs updating to handle test POJOs etc.

### DIFF
--- a/commons/src/main/java/org/infinispan/commons/marshall/AbstractDelegatingMarshaller.java
+++ b/commons/src/main/java/org/infinispan/commons/marshall/AbstractDelegatingMarshaller.java
@@ -16,6 +16,7 @@ import org.infinispan.commons.io.ByteBuffer;
  * @author Galder Zamarreño
  * @since 5.0
  */
+@Deprecated
 public abstract class AbstractDelegatingMarshaller implements StreamingMarshaller {
 
    protected StreamingMarshaller marshaller;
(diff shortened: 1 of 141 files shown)


ryanemerson:
Here's the branch with the WIP of a protostrem marshaller: https://github.com/ryanemerson/infinispan/tree/protostream_marshaller

ryanemerson:
Currently `ReplicableCommand::writeTo` passes in the MarshalledEntryFactory as well as the ObjectOutput, but in hindsight I don't think this is the best abstraction. Instead we could have a `UserAwareObjectOutput` interface which extends `ObjectOutput` and provides all the write operations that are currently controlled by `MarshalledEntryUtil`. Underthehood we could extend `DelegatingObjectOutput` for the impl. Similarly, we could also wrap ObjectInput to provide similar functionality.

ryanemerson:
Closing for now as I have realised that I have not taken into account the creation of custom commands and how a user marshaller can be passed to their externalizers.

PR #6156, opened 2018-07-18
ISPN-9385 Add near cache properties to the Hot Rod client

tristantarrant:
https://issues.jboss.org/browse/ISPN-9385
Also adds a regex pattern to selectively enable near-caching for matching caches
https://issues.jboss.org/browse/ISPN-9398
Expose all configuration properties with setters
https://issues.jboss.org/browse/ISPN-9098
Add a test for expiration puts with near cache


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/NearCacheConfigurationBuilder.java
```diff
@@ -25,6 +28,11 @@ public NearCacheConfigurationBuilder mode(NearCacheMode mode) {
       return this;
    }
 
+   public NearCacheConfigurationBuilder cacheNamePattern(String pattern) {
```
wburns: I have to admit I am stuck on the fact that the `NearCacheConfiguration` and `NearCacheConfigurationBuilder` classes aren't symmetrical in regards to the type for `cacheNamePattern`. One has a String and the other a Pattern.

I wonder if a user may want to specify a `Pattern` instance with different flags. This might not be that useful though since most of the flags I use for this are related to multiline or dot all, which shouldn't be needed for this case.

We could however have a `cacheNameRegexString` and `cacheNamePattern` method on this class. And the former would just invoke the latter with the compiled `Pattern`. But this is just my thinking, not sure what you think.

tristantarrant:
Addressed all issues.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/ConfigurationProperties.java
```diff
@@ -149,76 +177,196 @@ public int getSoTimeout() {
...
+   public void setNearCacheMode(String nearCacheMode) {
+      props.setProperty(NEAR_CACHE_MODE, nearCacheMode);
+   }
+
    public int getNearCacheMaxEntries() {
       return props.getIntProperty(NEAR_CACHE_MAX_ENTRIES, -1);
    }
 
-   public Pattern getNearCacheNamePattern() {
-      return (Pattern) props.get(NEAR_CACHE_NAME_PATTERN);
```
wburns: Looks much better with this as well :)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/NearCacheConfigurationBuilder.java
```diff
@@ -33,13 +71,15 @@ public void validate() {
...
-      return new NearCacheConfiguration(mode, maxEntries == null ? -1 : maxEntries);
+      return new NearCacheConfiguration(mode, maxEntries == null ? -1 : maxEntries, cacheNamePattern);
    }
 
    @Override
    public Builder<?> read(NearCacheConfiguration template) {
       mode = template.mode();
       maxEntries = template.maxEntries();
+      Pattern pattern = template.cacheNamePattern();
+      cacheNamePattern = pattern;
```
wburns: Can be just `cacheNamePattern = template.cacheNamePattern();`

tristantarrant:
@wburns updated with another very minor test to verify that ISPN-9098 is fixed by ISPN-9393
I also overhauled the javadoc for the configuration properties.

galderz:
https://issues.jboss.org/browse/ISPN-7337
https://github.com/infinispan/infinispan/pull/6182

😅🙈

galderz:
Last build timed out, restarting it

tristantarrant:
Rebased

galderz:
I have doubts that we really needed ISPN-9385. A very simple workaround would have been to use different remote cache managers depending on whether you wanted near cache or not. So, for caches with near cache, you'd use remote cache manager with near cache enabled, and for caches with no near cache, remote cache manager with near cache disabled. The cost of remote cache managers has gone down now that we don't have those connection pools any more and we have a true async trasnport.

tristantarrant:
I don't think it is acceptable to require the use of multiple RCMs as that is much more legwork for the user. I see the use of multiple RCMs necessary when you need different transport settings, not different cache settings.
Ideally we should probably make the mechanism a bit more flexible and not limited to near caching.
`infinispan.client.hotrod.cache.a*.marshaller=a.b.c
infinispan.client.hotrod.cache.mycache.marshaller=c.d.e
infinispan.client.hotrod.cache.user*.near_cache.mode=INVALIDATED`

galderz:
> Ideally we should probably make the mechanism a bit more flexible and not limited to near caching.

Our configuration is complicated enough as is :(

galderz:
I really don't see the need to have a per cache marshaller either.

galderz:
I can see the reason for selectively enabling near cache, so that's why I'm mildly approving this, but I really don't want to see a per cache marshalling settings...etc.

PR #6157, opened 2018-07-18
ISPN-9387 Server-side TCP KeepAlive

tristantarrant:
https://issues.jboss.org/browse/ISPN-9387

For naming consistency, I've also made some backwards incompatible API change to the Router endpoint. I don't think this is a problem, but let me know if you think otherwise.


server/integration/endpoint/src/main/java/org/infinispan/server/endpoint/subsystem/RouterConnectorResource.java
```diff
@@ -55,13 +55,13 @@
...
-                   .setXmlName(ModelKeys.KEEP_ALIVE)
-                   .setRestartAllServices()
-                   .setDefaultValue(new ModelNode().set(false))
-                   .build();
+   static final SimpleAttributeDefinition TCP_KEEPALIVE =
+         new SimpleAttributeDefinitionBuilder(ModelKeys.TCP_KEEPALIVE, ModelType.BOOLEAN, true)
+               .setAllowExpression(true)
+               .setXmlName(ModelKeys.TCP_KEEPALIVE)
+               .setRestartAllServices()
+               .setDefaultValue(new ModelNode().set(true))
```
pruivo: isn't the default `false`?

pruivo:
integrated! 
@tristantarrant check my comment for CR1. I've run the tests locally and they are fine. 
thank you!

PR #6160, opened 2018-07-18
ISPN-9375 Change RocksDB segmentation to use column family instead of

wburns:
per instance

* Refactor to have column handler based on if segmented

https://issues.jboss.org/browse/ISPN-9375

### DIFF
--- a/persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java
+++ b/persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java
@@ -5,7 +5,9 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.PrimitiveIterator;
+import java.util.Properties;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Executor;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -58,6 +60,8 @@ import io.reactivex.schedulers.Schedulers;
 @ConfiguredBy(RocksDBStoreConfiguration.class)
 public class RocksDBStore<K,V> implements SegmentedAdvancedLoadWriteStore<K,V> {
     private static final Log log = LogFactory.getLog(RocksDBStore.class, Log.class);
+    static final String databasePropertyNameWithSuffix = "database.";
+    static final String columnFamilyPropertyNameWithSuffix = "data.";
 
     private RocksDBStoreConfiguration configuration;
     private BlockingQueue<ExpiryEntry> expiryEntryQueue;
@@ -69,6 +73,8 @@ public class RocksDBStore<K,V> implements SegmentedAdvancedLoadWriteStore<K,V> {
     private Semaphore semaphore;
     private WriteOptions dataWriteOptions;
     private RocksDBHandler handler;
+    private Properties databaseProperties;
+    private Properties columnFamilyProperties;
     private volatile boolean stopped = true;
 
     @Override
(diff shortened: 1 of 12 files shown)


persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java
```diff
@@ -634,4 +516,580 @@ public RocksEntryIterator(RocksIterator it, Predicate<? super K> filter, boolean
...
+        }
+
+        @Override
+        void addSegments(IntSet segments) {
+            // Do nothing
+        }
+
+        @Override
+        void removeSegments(IntSet segments) {
+            clear(segments);
```
ryanemerson: Shouldn't this be a no-op?

persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java
```diff
@@ -634,4 +516,580 @@ public RocksEntryIterator(RocksIterator it, Predicate<? super K> filter, boolean
...
+            this.handles = new AtomicReferenceArray<>(segmentCount);
+        }
+
+        byte[] byteArrayFromInt(int val) {
+            byte[] b = new byte[4];
+            b[3] = (byte) (val);
+            b[2] = (byte) (val >>> 8);
+            b[1] = (byte) (val >>> 16);
+            b[0] = (byte) (val >>> 24);
+            return b;
```
ryanemerson: It doesn't really matter, but could be:

```java
return new byte[] {
            (byte) (val >>> 24),
            (byte) (val >>> 16),
            (byte) (val >>> 8),
            (byte) val
};
```

persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java
wburns: Nope. This method would be invoked during state transfer when segments are removed from this node. Since the non segmented implementation can't easily remove the segments it has to clear out the data pertaining to those segments.

persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java
wburns: Sure, this is just how the `Bits` class normally does it :) But I think this is a little easier to read.

wburns:
I have to update user guide for rocks db to talk about segmentation further.

wburns:
Changed up documentation as well.

persistence/rocksdb/src/main/java/org/infinispan/persistence/rocksdb/RocksDBStore.java
```diff
@@ -91,15 +121,22 @@ private WriteOptions dataWriteOptions() {
         return dataWriteOptions;
     }
 
-    private Options dataDbOptions() {
-        Options options = new Options().setCreateIfMissing(true);
-        options.setCompressionType(CompressionType.getCompressionType(configuration.compressionType().toString()));
+    private DBOptions dataDbOptions() {
+        DBOptions options = new DBOptions()
```
tristantarrant: if you could use DBOptions.getDBOptionsFromProps(configuration.properties()) we should be able to take care of https://issues.jboss.org/browse/ISPN-9371

wburns: The only problem with that is now we have `DBOptions` and `ColumnFamilyOptions`. I would have to check if it ignores improper values and subsequently determine if we should set compression or not.

wburns: I think this needs to be thought through a little bit more in a different PR. Unfortunately now that we are using column families they allow for different properties than `DBOptions`. And both of those are slightly different than the `Options` class.

`Options` - used when a database is only configured with the default column family (aka no column families)
`DBOptions` - used when column families are used and includes config for things not covered by column families
`ColumnFamilyOptions` - used for each column family things such as compression etc.

wburns: Also to keep in mind you can't provide a non mapped property, otherwise RocksDB will return null options object back. So we need a way to namespace these. I was thinking maybe something like prefix names `database` and `data` for `DBOptions` and `ColumnFamilyOptions` respectively.

tristantarrant: Yes. And fail fast if it the null dboptions is returned

wburns:
Updated with the additional properties being able to be set.

wburns:
Reader test should be fixed now - were writing to same location :D

ryanemerson:
@wburns The latest changes LGTM, I'll wait for CI though

PR #6161, opened 2018-07-19
Add surefire report to archived artifacts

rvansa:


rvansa:
This is the command to get test durations from artifacts:
```sh
xmllint --xpath '/testsuite/testcase' TEST-infinispan-core.xml | sed 's:/>:\n:g' | awk -F '"' '{ print $4" "$6"."$2 }' | sort -n
```

danberindei:
The build failed with a timeout connecting to github, but it looks safe enough to me.

PR #6162, opened 2018-07-19
ISPN-9393 Support async methods with the Hot Rod near cache

tristantarrant:
https://issues.jboss.org/browse/ISPN-9393

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
@@ -200,15 +200,20 @@ public class RemoteCacheImpl<K, V> extends RemoteCacheSupport<K, V> {
    }
 
    @Override
-   public MetadataValue<V> getWithMetadata(K key) {
-      return getWithMetadata(key, dataFormat);
+   public CompletableFuture<MetadataValue<V>> getWithMetadataAsync(K key) {
+      return getWithMetadataAsync(key, dataFormat);
    }
 
-   private MetadataValue<V> getWithMetadata(K key, DataFormat dataFormat) {
+   @Override
+   public MetadataValue<V> getWithMetadata(K key) {
+      return await(getWithMetadataAsync(key));
+   }
+
+   private CompletableFuture<MetadataValue<V>> getWithMetadataAsync(K key, DataFormat dataFormat) {
       assertRemoteCacheManagerIsStarted();
       GetWithMetadataOperation<V> op = operationsFactory.newGetWithMetadataOperation(
             compatKeyIfNeeded(key), keyToBytes(key), dataFormat);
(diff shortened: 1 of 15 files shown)


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
```diff
@@ -199,16 +199,21 @@ public boolean replaceWithVersion(K key, V newValue, long version, long lifespan
...
+   }
+
    @Override
    public MetadataValue<V> getWithMetadata(K key) {
-      return getWithMetadata(key, dataFormat);
+      return await(getWithMetadataAsync(key));
    }
 
-   private MetadataValue<V> getWithMetadata(K key, DataFormat dataFormat) {
+   protected CompletableFuture<MetadataValue<V>> getWithMetadataAsync0(K key, DataFormat dataFormat) {
```
rvansa: Other methods don't have the `0` suffix, why here?

documentation/src/main/asciidoc/user_guide/server_protocols_hotrod.adoc
```diff
@@ -1163,6 +1163,9 @@ updated or removed server-side, the cached near cache entries are invalidated.
 If a key is requested after it's been invalidated, it'll have to be re-fetched
 from the server.
 
+WARNING: You should not use `maxIdle` expiration with near caches, as the
+behaviour is undefined.
```
rvansa: *Undefined* is convenient, but is there a reason to not say that accesses to the near cache will be ignored for the purposes of expiration?

tristantarrant: Yes, let me clear that. I also found a compilation failure.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/RemoteCacheImpl.java
tristantarrant: oops, a leftover

tristantarrant:
Updated

pruivo:
I guess `max-idle` can be supported if  `ClientCacheEntryExpired` is triggered and the near-cache service listens to it.

tristantarrant:
@pruivo yes, but in order for maxIdle to work properly, every time the near cache does a local read, we would have to propagate the access time on the server as well, which sort of defeats the near cache

rvansa:
@tristantarrant Not entirely, knowing maxIdle you could setup a timer on client to (deadline - 1 second) and issue an update command only at that time.
Not that I would think that it's worth the effort.

tristantarrant:
Thanks but no thanks :)

tristantarrant:
1 unrelated failure

rvansa:
In that goes, then.

PR #6163, opened 2018-07-19
ISPN-8213 Functional commands are not replayed in tx on non-read owner

rvansa:
https://issues.jboss.org/browse/ISPN-8213

If you think that the code in distribution interceptors for multi-key commands is messy, I agree 100%. Maybe too much DRY. I've also hit some limitation related to https://issues.jboss.org/browse/ISPN-7889.

### DIFF
--- a/core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
@@ -54,6 +54,7 @@ import org.infinispan.context.InvocationContext;
 import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.context.impl.LocalTxInvocationContext;
 import org.infinispan.context.impl.TxInvocationContext;
+import org.infinispan.distribution.DistributionInfo;
 import org.infinispan.distribution.LocalizedCacheTopology;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/infinispan/interceptors/distribution/TxDistributionInterceptor.java
```diff
@@ -631,10 +636,15 @@ protected void handleRemotelyRetrievedKeys(InvocationContext ctx, List<?> remote
...
    }
 
-   private static List<Mutation> getMutationsOnKey(TxInvocationContext ctx, Object key) {
+   private static List<Mutation> getMutationsOnKey(TxInvocationContext ctx, WriteCommand untilCommand, Object key) {
       List<Mutation> mutations = new ArrayList<>();
       // We don't use getAllModifications() because this goes remote and local mods should not affect it
       for (WriteCommand write : ctx.getCacheTransaction().getModifications()) {
+         if (write == untilCommand) {
+            // We've reached this command in the modifications list; this happens when we're replaying a prepared
+            // transaction - see EntryWrappingInterceptor.wrapEntriesForPrepareAndApply
```
danberindei: I'm not sure this is relevant, but is it really true? To me it seems like we always call `getMutationsOnKey(ctx, command, key)` on the originator when we don't have the entry in the context.

PR #6168, opened 2018-07-24
ISPN-9067 Fix test node names

danberindei:
https://issues.jboss.org/browse/ISPN-9067

I had this lying around for a while...

* Only set the node name once
* Fail the test if a node doesn't have a proper name set
* Don't start the cache manager in the configuration parsing tests


danberindei:
Rebased

tristantarrant:
Merged, thanks

PR #6172, opened 2018-07-25
ISPN-9399 Upgrade to protostream 4.2.1.Final

anistor:
https://issues.jboss.org/browse/ISPN-9399

### DIFF
--- a/remote-query/remote-query-client/src/test/java/org/infinispan/query/remote/client/BaseProtoStreamMarshallerTest.java
+++ b/remote-query/remote-query-client/src/test/java/org/infinispan/query/remote/client/BaseProtoStreamMarshallerTest.java
@@ -1,9 +1,9 @@
 package org.infinispan.query.remote.client;
 
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
-
-import java.time.Instant;
-import java.util.Date;
 
 import org.infinispan.protostream.ProtobufUtil;
 import org.infinispan.protostream.SerializationContext;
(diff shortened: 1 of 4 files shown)


remote-query/remote-query-client/src/test/java/org/infinispan/query/remote/client/BaseProtoStreamMarshallerTest.java
```diff
@@ -19,21 +19,39 @@
...
+      roundtrip('a');
+      roundtrip(0);
+      roundtrip(0L);
+      roundtrip(0.0D);
+      roundtrip(0.0F);
+      roundtrip((byte) 0);
+      roundtrip((short) 0);
+      roundtrip(true);
+//      roundtrip(new Date(0));
+//      roundtrip(Instant.now());
```
ryanemerson: If this is not supported, should update `BaseProtostreamMarshaller::isMarshallable`  to not include them?

anistor: ... or add support for them. I'm pondering if we actually need it or not.

ryanemerson: @anistor I don't think support for those should be in protostream, but maybe we should provide marshaller implementations for the client? @gustavonalle any thoughts?

ryanemerson: What I meant was that in the protostream lib I'm not sure we should handle Date, but for `BaseProtoStreamMarshaller` we could provide a `MessageMarshaller<Data>` implementation that is added to the serialization context. That way users of `ProtostreamMarshaller` can still marshall Date objects seamlessly. However, if the lack of support for marshalling Date has not been noticed yet, maybe we should just not support it.

anistor: Handling Date and Instant as primitives is for the sake of binary representation efficiency. If it gets handled as a message with a MessageMarshaller the payload gets bigger and we should also have a corresponding protobuf type for Date, which we do not have atm. 

Also, this handling of representing Date and Instant as a long timestamp is already used for Date/Instant message fields. The missing part is implementing the same for Date/Instant when it is a top level object.

gustavonalle: > people rarely use protobuf unless they use remote query

Protobuf is the best data format for mixing rest, hot rod and non-java clients, so I expect the usage to increase :)

> we can always plug entity marshallers for enums and messages types, but not for primitives

There are many JDK classes apart from Date/Instant that can (and will) be used as a cache value or key. I hope there will be a way to plug them in the serCtx?

anistor: No problem, we can handle them as messages. But we'll need a schema and a marshaller for them.

anistor:
I've restarted the CI build

anistor:
green build!

ryanemerson:
Thanks @anistor

PR #6175, opened 2018-07-26
ISPN-8023 Provide infinispan-hibernate-cache-v53 feature

fax4ever:
https://issues.jboss.org/browse/ISPN-8023

### DIFF
--- a/integrationtests/osgi/src/test/resources/maven.properties
+++ b/integrationtests/osgi/src/test/resources/maven.properties
@@ -10,3 +10,4 @@ version.mockito=${version.mockito}
 version.mockito_dep.objenesis=${version.mockito_dep.objenesis}
 version.mockito_dep.bytebuddy=${version.mockito_dep.bytebuddy}
 localRepository=${settings.localRepository}
+version.hibernate.osgi=${version.hibernate.osgi}
--- a/hibernate/cache-commons/pom.xml
+++ b/hibernate/cache-commons/pom.xml
@@ -9,7 +9,7 @@
    </parent>
 
    <artifactId>infinispan-hibernate-cache-commons</artifactId>
-   <packaging>jar</packaging>
+   <packaging>bundle</packaging>
    <name>Infinispan Hibernate Cache Commons</name>
    <description>Infinispan Hibernate Cache Commons module</description>
 
@@ -66,6 +66,24 @@
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration combine.self="override">
                <skipTests>true</skipTests>
+            </configuration>
+         </plugin>
+         <plugin>
+            <groupId>org.apache.felix</groupId>
+            <artifactId>maven-bundle-plugin</artifactId>
+            <configuration>
+               <instructions>
+                  <Export-Package>
+                     ${project.groupId}.hibernate.cache.commons.*,
+                     ${project.groupId}.hibernate.cache.commons.access.*,
+                     ${project.groupId}.hibernate.cache.commons.util.*;version=${project.version};-split-package:=error
+                  </Export-Package>
+                  <Include-Resource>
+                     {maven-resources},
+                     /META-INF/services=${project.basedir}/target/classes/META-INF/services,
+                     ${project.build.outputDirectory}/${project.artifactId}-component-metadata.dat
+                  </Include-Resource>
+               </instructions>
             </configuration>
          </plugin>
       </plugins>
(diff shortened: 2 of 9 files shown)


rvansa:
@fax4ever Thanks, but `org.infinispan.it.osgi.hibernate.cache.v53.HibernateCacheIntegrationTest.testActivation` seems failing with timeout.

fax4ever:
Hi @rvansa, I'm checking it. BTW could you be interested in such as contribution?

fax4ever:
Because we are removing this integration test from hibernate side.

rvansa:
I personally don't know much about OSGi, but the rest of the Infinispan supports OSGi and therefore I guess the 2LC should, too. And if we support something, a basic test is a must - therefore, we're grateful for such contribution.

fax4ever:
There was a ClassNotFoundException on my test, now using the `probeIsolationWorkaround` it should work. Moreover, I would like bundle hibernate cache modules, then remove the wraps protocol from the feature. I would do in a subsequent commit.

pom.xml
```diff
@@ -271,6 +271,7 @@
       <version.h2.driver-within-wildfly>1.4.193</version.h2.driver-within-wildfly>
       <version.h2.driver>1.4.180</version.h2.driver>
       <version.hibernate.osgi>${version.hibernate.core}</version.hibernate.osgi>
+      <version.hibernate.v53>5.3.3.Final</version.hibernate.v53>
```
galderz: We already have `version.hibernate.core` above pointing to 5.3.1.Final. Do you really need 5.3.3?

galderz:
> Moreover, I would like bundle hibernate cache modules, then remove the wraps protocol from the feature. I would do in a subsequent commit.

That sounds good. You can fill a follow-up JIRA and do it there.

pom.xml
fax4ever: I think that we need two versions of Hibernate. A 5.1.x version to test the feature `infinispan-hibernate-cache-v51` and a 5.3.x version to test the feature `infinispan-hibernate-cache-v53`.

fax4ever:
The current issue name is `Make Hibernate Cache produce OSGI bundles`. I think that this issue would not be complete without providing a real bundle for cache modules. Instead of creating another issue for this task, what do you think if I added it as subsequent commit? By the way, for me it is the same.... I'm just wondering to be consistent with issue name.

galderz:
Happy with a subsequent commit, but it'd have to be within the same release to be part of the same JIRA really.

pom.xml
galderz: My point is not about 5.1.x vs 5.3.x. It's about different 5.3.x versions...

Above in the pom it defines:

    <version.hibernate.core>5.3.1.Final</version.hibernate.core>

And you define:

    <version.hibernate.v53>5.3.3.Final</version.hibernate.v53>

To reiterate, why can't you use 5.3.1.Final?

galderz: This line should be removed...

integrationtests/osgi/src/test/resources/maven.properties
```diff
@@ -10,3 +10,4 @@ version.mockito=${version.mockito}
 version.mockito_dep.objenesis=${version.mockito_dep.objenesis}
 version.mockito_dep.bytebuddy=${version.mockito_dep.bytebuddy}
 localRepository=${settings.localRepository}
+version.hibernate.v53=${version.hibernate.v53}
```
galderz: And this should be:

    version.hibernate.v53=${version.hibernate.core}

pom.xml
galderz: Actually...

Leave this line as is, so leave `version.hibernate.v53` here, but then make cache-v53 use:

    <version.hibernate.core>${version.hibernate.v53}</version.hibernate.core>

galderz: By doing that, we fix the Hibernate 5.3.x version in a single place, `version.hibernate.v53`. Although the question begs while you need to add `version.hibernate.v53` to root pom.xml rather than define it in `cache-v53/pom.xml`, keeping all more encapsulated.

integrationtests/osgi/src/test/resources/maven.properties
fax4ever: I've removed `version.hibernate.v53` property in favor of `version.hibernate.osgi`

galderz:
@fax4ever Is this ready now after Hibernate 5.3.4 upgrade?

galderz:
Sorry for delay, let's give it another spin just in case.

fax4ever:
Hi @galderz. I think there is no rush, even better to test deeply :)

galderz:
Looks good, thx @fax4ever :)

fax4ever:
Thanks!

PR #6177, opened 2018-07-27
ISPN-8799 Limit the default number of shards in the AffinityIndexManager

gustavonalle:
https://issues.jboss.org/browse/ISPN-8799

### DIFF
--- a/query/src/main/java/org/infinispan/query/affinity/AffinityShardIdentifierProvider.java
+++ b/query/src/main/java/org/infinispan/query/affinity/AffinityShardIdentifierProvider.java
@@ -35,8 +35,10 @@ import org.infinispan.remoting.transport.LocalModeAddress;
 public class AffinityShardIdentifierProvider implements ShardIdentifierProvider {
 
    private static final Log log = LogFactory.getLog(AffinityShardIdentifierProvider.class, Log.class);
+   public static final int DEFAULT_NUMBER_SHARDS = 4;
 
    private static final String NUMBER_OF_SHARDS_PROP = "nbr_of_shards";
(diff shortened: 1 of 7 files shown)


query/src/main/java/org/infinispan/query/affinity/ShardAllocationManagerImpl.java
```diff
@@ -69,7 +69,7 @@ public String getShardFromKey(Object key) {
    @Override
    public void initialize(Integer numberOfShards, int numSegments) {
       this.numSegments = numSegments;
-      this.numShards = numberOfShards == null ? numSegments : numberOfShards;
+      this.numShards = numberOfShards;
```
anistor: numberOfShards is Integer, not int, so it can still be null in theory. Maybe it's worth keeping the old line here or change numberOfShards from Integer to int.

query/src/main/java/org/infinispan/query/affinity/AffinityShardIdentifierProvider.java
```diff
@@ -121,6 +122,9 @@ private DistributionManager getDistributionManager() {
 
    private static Integer getNumberOfShards(Properties properties) {
       String nShards = properties.getProperty(NUMBER_OF_SHARDS_PROP);
+      if (nShards == null) {
+         nShards = properties.getProperty(SHARDING_STRATEGY_PROP + "." + NUMBER_OF_SHARDS_PROP);
```
anistor: By passing the default number of shard as a 2nd arg to getProperty we can ensure getNumberOfShards always has a non-null value to return, so it can return int

gustavonalle:
Updated. Now using ```int``` everywhere.

anistor:

PR #6179, opened 2018-07-31
ISPN-9261 Hot Rod Lightweight Transaction (recovery)

pruivo:
* Server initial changes

This is a very "raw" PR but I need to get this in before Beta1. 
It is missing the javadoc, documentation and the client (I'll do a PR for it tomorrow)
All the missing bits should be ready for CR1/Final.

https://issues.jboss.org/browse/ISPN-9261

Don't close the JIRA since I'll attach the client PR as well.


gustavonalle:
I suppose you should increase the Hot Rod version to 2.9?

pruivo:
@gustavonalle I don't think.
2.8 was introduced in 9.4 Alpha1. Do I need to change in the same minor?

gustavonalle:
@pruivo http://infinispan.org/docs/stable/user_guide/user_guide.html#hot_rod_protocol_2_8

pruivo:
updated!

pruivo:
@gustavonalle yes. 2.8 introduce new parameters in the header, but the test HotRodClient wasn't updated. I'm still using 2.7 in my tests and 2.9 in the client-hotrod tests.

PR #6180, opened 2018-07-31
ISPN-9369 Non-tx query cache results should be cached regardless of tx outcome

galderz:
https://issues.jboss.org/browse/ISPN-9369

### DIFF
--- a/hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/ReadOnlyTest.java
+++ b/hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/ReadOnlyTest.java
@@ -9,15 +9,18 @@ package org.infinispan.test.hibernate.cache.commons.functional;
 import java.util.List;
 import java.util.Map;
 
+import org.hibernate.testing.TestForIssue;
 import org.infinispan.hibernate.cache.commons.InfinispanBaseRegion;
 import org.infinispan.hibernate.cache.commons.util.InfinispanMessageLogger;
 import org.hibernate.stat.Statistics;
 import org.infinispan.test.hibernate.cache.commons.functional.entities.Item;
+import org.infinispan.test.hibernate.cache.commons.util.InducedException;
 import org.infinispan.test.hibernate.cache.commons.util.TestRegionFactory;
 import org.infinispan.util.ControlledTimeService;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 
 /**
  * Parent tests for both transactional and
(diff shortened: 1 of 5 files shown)


rvansa:
This definitely deserves some discussion! I understand the read-only case mentioned in JIRA, but then we have to handle read-only case and not cache all results.
The problem is that you could add entities into the DB and then execute a query in the same transaction. The query result could be affected, and if you cached = shared the query result, you'd be exposing an uncommitted entity.

hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/ReadOnlyTest.java
```diff
@@ -91,6 +94,44 @@ public void testInsertClearCacheDeleteEntity() throws Exception {
 		});
 	}
 
+   @Test
+   @TestForIssue(jiraKey = "ISPN-9369")
+   public void testCacheQueryResultsOnRollback() throws Exception {
```
rvansa: I believe that the request in JIRA requests to expose the query results when the transaction is in-flight (as it's never committed), not just on rollback.

galderz:
> This definitely deserves some discussion! 

Sure, let's discuss. That's what the PR is for :)

> I understand the read-only case mentioned in JIRA, but then we have to handle read-only case and not cache all results.The problem is that you could add entities into the DB and then execute a query in the same transaction. The query result could be affected, and if you cached = shared the query result, you'd be exposing an uncommitted entity.

In transaction, uncommitted data will be presented to the user when doing queries. That's independent of the fix here which affects what happens after tx is committed/rollback. I've just added a commit with a test demonstrates that it works as expected. If you add an entity in a transaction and you roll that back, in the next transaction you won't see that entity magically appearing in the query.

rvansa:
You're dealing with another issue than the one described in JIRA, then. Shouldn't you rather create a separate one for this change?

In `testCacheQueryResultsOnRollback` you
1) obtain query result from DB
2) rollback the tx but put the query result into cache
3) read the query result from cache

In `testAvoidUncommittedResultsInQuery` you
1) obtain query result from DB
2) put the query result into cache (committed tx)
3) obtain uncommitted query result from DB
4) update timestamp on Item, invalidating the query result
5) put the uncommitted query result to cache
6) verify that now you get correct result from DB

Have you verified that 4) and 5) happen in this order? Otherwise there's a time window to read the uncommitted query result. And even if it is, I am not too comfortable with inserting something that's potentially uncommitted (not just stale) into the cache.

galderz:
> You're dealing with another issue than the one described in JIRA, then. Shouldn't you rather create a separate one for this change?

Why?

> In testCacheQueryResultsOnRollback you

In `testCacheQueryResultsOnRollback` I replicate the issue reported by the user.

> In testAvoidUncommittedResultsInQuery you

In `testAvoidUncommittedResultsInQuery` I replicate the potential issue you reported [here](https://github.com/infinispan/infinispan/pull/6180#discussion_r206782934). Feel free to improve the reproducer as you see fit.

rvansa:
> In testCacheQueryResultsOnRollback I replicate the issue reported by the user.

I disagree. The user does not commit/or rollback the transaction, user starts a read-only transaction and does not complete it. Since with your fix the query cache is updated only after transaction completes you're not helping this user.

> In testAvoidUncommittedResultsInQuery I replicate the potential issue

Yes, I didn't realize the first time that the fact that **uncommitted** result is stored in the cache is mitigated by updating the timestamp. But I feel this opens a hole for trouble: suppose that with a replicated query cache the timestamp update is delayed (and it can be as it's async) but the query cache update is not. Someone could read the uncommitted data then.

I am reluctant to work on the reproducer to prove such issues as I think that the fix itself is conceptually wrong. Rolled back transactions should not write anything to a shared cache.

rvansa:
Actually I think that the rolled back transactions should not update timestamps cache either, but that would be a different issue.

galderz:
> I disagree. The user does not commit/or rollback the transaction, user starts a read-only transaction and does not complete it. Since with your fix the query cache is updated only after transaction completes you're not helping this user.

You're right. I'm going to make this won't fix.

PR #6181, opened 2018-07-31
[9.1.x] ISPN-9405 Fix non-existing node xsite failover

galderz:
https://issues.jboss.org/browse/ISPN-9405


tristantarrant:
Why do we need this for 9.1 ?

PR #6184, opened 2018-08-02
 ISPN-9261 Hot Rod Lightweight Transaction (recovery) [client]

pruivo:
https://issues.jboss.org/browse/ISPN-9261

Depends on https://github.com/infinispan/infinispan/pull/6179

Still missing some javadoc/docs that will be added for CR1.


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/SyncModeTransactionTable.java
```diff
@@ -81,24 +84,26 @@ private SynchronizationAdapter createSynchronizationAdapter(Transaction transact
...
       public String toString() {
          return "SynchronizationAdapter{" +
-               "registeredCaches=" + registeredCaches.keySet() +
-               ", transaction=" + transaction +
-               ", xid=" + xid +
-               '}';
+                "registeredCaches=" + registeredCaches.keySet() +
+                ", transaction=" + transaction +
+                ", xid=" + xid +
+                '}';
```
danberindei: I see you've started aligning wrapped lines to the expression start in some cases, but not yet everywhere, e.g. with parameters... don't worry, you'll get there ;)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/SyncModeTransactionTable.java
```diff
@@ -134,15 +134,15 @@ public void beforeCompletion() {
...
+         TransactionContext<?, ?> ctx = registeredCaches.values().iterator().next();
+         //the server commits everything when the first request arrives.
          try {
             boolean commit = status == Status.STATUS_COMMITTED;
-            for (TransactionContext<?, ?> ctx : preparedCaches) {
-               ctx.complete(xid, commit);
-            }
+            ctx.complete(xid, commit);
          } finally {
+            ctx.forget(xid); //no recovery sync
```
danberindei: Since the commit is a single operation sent to a single server, I'm not sure we need a separate forget operation after a successful commit.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/TransactionContext.java
```diff
@@ -209,7 +214,8 @@ int prepareContext(Xid xid, boolean onePhaseCommit) {
    int complete(Xid xid, boolean commit) {
       try {
          if (trace) {
-            log.tracef("%s transaction xid=%s, remote-cache=%s", commit ? "Committing" : "Rolling-Back", xid, cacheName);
+            log.tracef("%s transaction xid=%s, remote-cache=%s", commit ? "Committing" : "Rolling-Back", xid,
```
danberindei: Off-topic, but I'd prefer "Complete(true|false) transaction..."

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
```diff
@@ -133,8 +139,13 @@ public void commit(Xid xid, boolean onePhaseCommit) throws XAException {
          if (trace) {
             log.tracef("XaResource.commit(%s, %s)", xid, onePhaseCommit);
          }
-         assertStartInvoked();
-         assertSameXid(xid, XAException.XAER_INVAL);
+         if (currentXid == null) {
+            //no transaction running. we are doing some recovery work
```
danberindei: "doing some recovery work" sounds too vague to me, like we could be doing more than just committing the tx

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
```diff
@@ -151,10 +162,16 @@ public void rollback(Xid xid) throws XAException {
          if (trace) {
             log.tracef("XaResource.rollback(%s)", xid);
          }
-         assertStartInvoked();
-         assertSameXid(xid, XAException.XAER_INVAL);
+         boolean ignoreNoTx = true;
+         if (currentXid == null) {
+            //no transaction running. we are doing some recovery work
```
danberindei: Same here

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
```diff
@@ -173,25 +190,50 @@ public void forget(Xid xid) {
...
+            throw new XAException(XAException.XAER_INVAL);
+         }
+         return it.next();
       }
 
       @Override
-      public boolean setTransactionTimeout(int i) {
-         //not supported yet
-         return false;
+      public boolean setTransactionTimeout(int timeout) { //timeout is in seconds
```
danberindei: Please move the comment to the next line, or even better rename the parameter to `timeoutSeconds`

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
```diff
@@ -208,121 +250,44 @@ private void assertSameXid(Xid otherXid, int xaErrorCode) throws XAException {
...
-               //unknown. we throw what we received
-               throw new XAException(unknownErrorCode);
-            }
-         } else if (hasCommit && hasRollback) {
-            //we have something committed and something rolled back
-            throw new XAException(XAException.XA_HEURMIX);
-         } else if (hasRollback) {
-            //rollbacks only
-            throw new XAException(XAException.XA_HEURRB);
+         int xa_code = getAnyTransactionContext().complete(currentXid, true);
```
danberindei: It might be cleaner if the operations were executed at the adapter level and not at the `TransactionContext` level. Collecting the modifications is done per cache, but the modifications list can be a parameter of `prepareContext()`.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
```diff
@@ -392,13 +359,15 @@ private void onePhaseCommit() throws XAException {
...
          return new TransactionContext<>(remoteCache.keyMarshaller(), remoteCache.valueMarshaller(),
-               remoteCache.getOperationsFactory(), remoteCache.getName());
+               remoteCache.getOperationsFactory(), remoteCache.getName(), remoteCache.isRecoveryEnabled());
       }
 
       private void cleanup() {
          registeredTransactions.remove(transaction);
-         registeredCaches.clear();
-         preparedCaches.clear();
+         //this instance can be used for recovery. we need at least one cache registered in order to access
```
danberindei: Do TMs really reuse XAResources like this? I would have thought it's more likely to create a new XAResource and call `recover()` on it directly.

Anyway, I think we should move the tx operations to a `GlobalOperationsFactory` that doesn't require a cache.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/tx/RecoveryTest.java
```diff
@@ -0,0 +1,247 @@
...
+      //start and end together is fine!
+      xaResource.recover(XAResource.TMSTARTRSCAN | XAResource.TMENDRSCAN);
+
+      //no iteration in progress
+      assertInvalidException(() -> xaResource.recover(XAResource.TMNOFLAGS));
+   }
+
+   public void testRecoveryIteration() throws Exception {
+      XAResource xaResource_0 = xaResource(0);
+      XAResource xaResource_1 = xaResource(1);
```
danberindei: Underscore looks a bit odd

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/tx/RecoveryTest.java
```diff
@@ -0,0 +1,247 @@
...
+      Xid[] actual = xaResource.recover(XAResource.TMSTARTRSCAN);
+      assertTrue(actual.length != 0);
+      if (actual.length == 1) {
+         //it returned only the local transaction
+         assertEquals(local, actual[0]);
+         actual = xaResource.recover(XAResource.TMENDRSCAN);
+         assertEquals(1, actual.length); //other client transaction
+         assertEquals(remote, actual[0]);
+      } else {
+         //the server replied quick enough
```
danberindei: Hmmm, maybe I need to re-read that part of the code... I didn't realize recovery iteration could stop because of a timeout

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/tx/RecoveryTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+   private XAResource xaResource(int index) throws Exception {
+      RemoteTransactionManager tm = remoteTM(index);
+      tm.begin();
+      RemoteCache<String, String> cache = remoteCache(index);
+      cache.put("_k_", "_v_");
+      TransactionImpl tx = (TransactionImpl) tm.getTransaction();
+      XAResource xaResource = tx.getEnlistedResources().iterator().next();
+      tm.commit();
+      xaResource.forget(tx.getXid());
```
danberindei: I really think `recover()` should work without a dummy transaction, e.g. immediately after a client restart.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
pruivo: create a new XaResource from where? There is no interfaces in the specifications to create XaResources, but they can be reused in other transactions (that's why we have Xid as parameters everywhere).

I don't see a `GlobalOperationFactory` but I really need one for transactions and counters...

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
pruivo: I had the operation in adapter before and I end up duplicate the code between SyncAdapter and XaAdapter. So I moved them to the context

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
pruivo: recovery only commits or rollbacks in-doubt and prepared transactions. there is nothing else to do.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/SyncModeTransactionTable.java
pruivo: in case if the client times-out and the retry kicks in, it finds the transaction in server with committed state.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/SyncModeTransactionTable.java
pruivo: every time I update intellij it decides to change my format config. I gave up...

server/hotrod/src/main/java/org/infinispan/server/hotrod/LifecycleCallbacks.java
```diff
@@ -80,12 +106,9 @@ private void registerServerTransactionTable(ComponentRegistry componentRegistry,
...
-      Cache<CacheXid, TxState> globalTxTable = cacheManager.getCache(GLOBAL_TX_TABLE_CACHE_NAME);
-      ServerTransactionTable transactionTable = new ServerTransactionTable(globalTxTable,
-            ByteString.fromString(cacheName), cacheManager.getAddress());
-      componentRegistry.registerComponent(transactionTable, ServerTransactionTable.class);
-      ServerTransactionOriginatorChecker checker = new ServerTransactionOriginatorChecker();
-      componentRegistry.registerComponent(checker, TransactionOriginatorChecker.class);
+      createGlobalTxTable(cacheManager);
+      componentRegistry.registerComponent(new PerCacheTxTable(cacheManager.getAddress()), PerCacheTxTable.class);
+      componentRegistry.registerComponent(new ServerTransactionOriginatorChecker(), TransactionOriginatorChecker.class);
       componentRegistry.rewire();
```
danberindei: Could other core components depend on the components we've just added? Otherwise `rewire()` doesn't do anything.

server/hotrod/src/main/java/org/infinispan/server/hotrod/TransactionRequestProcessor.java
```diff
@@ -38,41 +36,70 @@
...
-      executor.execute(() -> prepareTransactionInternal(header, cache, xid, onePhaseCommit, writes));
+      executor.execute(() -> prepareTransactionInternal(header, cache, xid, onePhaseCommit, writes, recoverable, timeout));
+   }
+
+   void forgetTransaction(HotRodHeader header, Subject subject, XidImpl xid) {
+      //TODO authentication?
+      GlobalTxTable txTable = server.getCacheManager().getGlobalComponentRegistry().getComponent(GlobalTxTable.class);
+      executor.execute(() -> {
+         try {
+            txTable.forgetTransaction(xid);
```
danberindei: I really wish our internal tx stuff was async... OTOH the first prototype of project loom is already here, maybe we won't need to make it async after all :)

http://mail.openjdk.java.net/pipermail/loom-dev/2018-July/000061.html

server/hotrod/src/main/java/org/infinispan/server/hotrod/TransactionRequestProcessor.java
```diff
@@ -82,7 +109,7 @@ private void prepareTransactionInternal(HotRodHeader header, AdvancedCache<byte[
             writeResponse(header, createTransactionResponse(header, XAResource.XA_RDONLY));
             return;
          }
-         PrepareTransactionDecodeContext txContext = new PrepareTransactionDecodeContext(cache, xid);
+         PrepareCoordinator txContext = new PrepareCoordinator(cache, xid, recoverable, timeout); //TODO timeout
```
danberindei: Please rename the variable as well

server/hotrod/src/main/java/org/infinispan/server/hotrod/TransactionRequestProcessor.java
```diff
@@ -113,60 +141,27 @@ private void prepareTransactionInternal(HotRodHeader header, AdvancedCache<byte[
...
+                         txContext.rollback() :
+                         txContext.prepare(onePhaseCommit);
             writeResponse(header, createTransactionResponse(header, xaCode));
          } catch (Exception e) {
             writeResponse(header, createTransactionResponse(header, txContext.rollback()));
          } finally {
             EmbeddedTransactionManager.dissociateTransaction();
          }
       } catch (Throwable t) {
+         log.error("meh...", t);
```
danberindei: :D

server/hotrod/src/main/java/org/infinispan/server/hotrod/TransactionRequestProcessor.java
```diff
@@ -195,39 +190,50 @@ private void validateConfiguration(AdvancedCache<byte[], byte[]> cache) {
...
+         //in any case, we send a ignore reply and the client is free to retry (or rollback)
+         writeNotExecuted(header);
+         return true;
+      }
+      //originator is dead...
+
+      //First phase state machine
+      //success ACTIVE -> PREPARING -> PREPARED
+      //failed ACTIVE -> MARK_ROLLBACK -> ROLLED_BACK or ACTIVE -> PREPARING -> ROLLED_BACK
+      //1PC success ACTIVE -> PREPARING -> MARK_COMMIT -> COMMITTED
```
danberindei: Need some more comments on these, I'm sure what MARK_COMMIT is for, and I'm starting to doubt that I know what PREPARING means as well :)

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/PrepareCoordinator.java
```diff
@@ -0,0 +1,325 @@
...
+import org.infinispan.transaction.impl.LocalTransaction;
+import org.infinispan.transaction.impl.TransactionTable;
+import org.infinispan.transaction.tm.EmbeddedTransaction;
+import org.infinispan.transaction.tm.EmbeddedTransactionManager;
+import org.infinispan.transaction.xa.GlobalTransaction;
+import org.infinispan.transaction.xa.TransactionFactory;
+import org.infinispan.util.ByteString;
+
+/**
+ * A base decode context to handle prepare, commit and rollback requests from a client.
```
danberindei: No longer a decode context, I guess

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/tx/RecoveryTest.java
pruivo: the recovery() interface supports batching. so, I'm not waiting for the server reply when it is first invoked and I'm returning whatever info the client has locally.
when the end-scan flag is used, it is where I wait for the server reply.

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/tx/RecoveryTest.java
pruivo: you still need an XaResource and there is no XaResource factories.

server/hotrod/src/main/java/org/infinispan/server/hotrod/LifecycleCallbacks.java
pruivo: at least the TransactionTable in core needs the `TransactionOriginatorChecker`.

server/hotrod/src/main/java/org/infinispan/server/hotrod/TransactionRequestProcessor.java
pruivo: PREPARING means the transaction has invoked tx.runPrepare()
MARK_COMMIT / ROLLBACK means that we know the client's TM decision and we are performing the commit / rollback.

pruivo: I'm going to add javadoc to the Status enum

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/tx/APITxTest.java
```diff
@@ -61,7 +62,11 @@
 
             new APITxTest<String, String>().keyValueGenerator(STRING_GENERATOR).transactionMode(NON_DURABLE_XA),
             new APITxTest<byte[], byte[]>().keyValueGenerator(BYTE_ARRAY_GENERATOR).transactionMode(NON_DURABLE_XA),
-            new APITxTest<Object[], Object[]>().keyValueGenerator(GENERIC_ARRAY_GENERATOR).transactionMode(NON_DURABLE_XA)
+            new APITxTest<Object[], Object[]>().keyValueGenerator(GENERIC_ARRAY_GENERATOR).transactionMode(NON_DURABLE_XA),
+
+            new APITxTest<String, String>().keyValueGenerator(STRING_GENERATOR).transactionMode(FULL_XA),
+            new APITxTest<byte[], byte[]>().keyValueGenerator(BYTE_ARRAY_GENERATOR).transactionMode(FULL_XA),
+            new APITxTest<Object[], Object[]>().keyValueGenerator(GENERIC_ARRAY_GENERATOR).transactionMode(FULL_XA)
```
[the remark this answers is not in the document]
pruivo: I don't think so. but I feel safe if we run them :)

pruivo:
updated!

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/table/Status.java
```diff
@@ -0,0 +1,84 @@
...
+    * The client transaction was replayed successful (versions read are validated) and the transactions is preparing.
+    */
+   PREPARING(4),
+   /**
+    * The transaction is successful prepared and it waits for the client {@link TransactionManager} decision to commit
+    * or rollback
+    */
+   PREPARED(5),
+   /**
+    * The client {@link TransactionManager} decided to commit. It can't rollback anymore.
```
danberindei: What if the prepare fails?

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/table/TxState.java
```diff
@@ -0,0 +1,163 @@
...
+@Immutable
+public class TxState {
+
+   public static final AdvancedExternalizer<TxState> EXTERNALIZER = new Externalizer();
+   private final GlobalTransaction globalTransaction;
+   private final Status status;
+   private final List<WriteCommand> modifications;
+   private final boolean recoverable;
+   private final long timeout; //ms
+   private final long lastAccessTimeNs; //ns
```
danberindei: No need for the comment here, `ns` is already in the name.

But I'm not sure using nanos and `nanoTime()` is the right approach in a class that is sent remotely:

     * <p>The values returned by this method become meaningful only when
     * the difference between two such values, obtained within the same
     * instance of a Java virtual machine, is computed.

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/table/TxState.java
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * A transaction state stored globally in all the cluster members.
+ *
+ * @author Pedro Ruivo
+ * @since 9.4
+ */
+@Immutable
+public class TxState {
+
+   public static final AdvancedExternalizer<TxState> EXTERNALIZER = new Externalizer();
```
danberindei: I'd rather put the empty line between the static and instance fields.

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/table/TxState.java
pruivo: it isn't sent remotely. the class is updated via functional API, so it uses the local TimeService and not the remote value.

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/table/GlobalTxTable.java
```diff
@@ -0,0 +1,385 @@
...
+      if (isRemote(rpcManager, state.getOriginator())) {
+         //remotely originated transaction
+         //this is a weird state. the originator may crashed or it may be in another partition and communication with the client
+         //in any case, we can rollback the tx
+         //if we are in the minority partition, updating the global tx table will fail and we do nothing
+         //if we are in the majority partition, the originator can't commit/rollback since it would fail to update the global tx table
+         rollbackOldTransaction(cacheXid, state, () -> rollbackRemote(cr, cacheXid, state));
+      } else {
+         //local transaction prepared.
+         PerCacheTxTable txTable = cr.getComponent(PerCacheTxTable.class);
```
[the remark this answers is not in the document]
pruivo: the PerCacheTxTable doesn't make any sense now. I'll remove it in the future (it will be merged with the GlobalTxTable)

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/table/GlobalTxTable.java
```diff
@@ -0,0 +1,385 @@
...
+ */
+@Scope(value = Scopes.GLOBAL)
+public class GlobalTxTable implements Runnable, Lifecycle {
+
+   //TODO think about the possibility of using JGroups RAFT instead of replicated cache?
+
+   private static final Log log = LogFactory.getLog(GlobalTxTable.class, Log.class);
+   private static final boolean trace = log.isTraceEnabled();
+
+   private final Cache<CacheXid, TxState> storage;
```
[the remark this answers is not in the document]
pruivo: yes and no. it would require some mapping inside with the cache name and the modifications made. Also, the status would have to be in the mappings (I guess).
In any case, I can investigate further before final

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/table/Status.java
pruivo: if the prepare fails, it changes from PREPARING to MARK_ROLLBACK and replies negatively to the transaction manager. The transaction manager has to decide to rollback the transaction.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
pruivo: ok, I'll change for CR1

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/table/TxState.java
danberindei: The externalizer does read+write the timestamp... maybe the externalizer isn't needed at all?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
pruivo: The ResourceAdapter seems to be an interface used by EE connector. And the Hot Rod client isn't one. Should it be? I'm not sure...
I was aware of XAResourceRecovery but it is related only to naryana. And, it looks like naryana likes to serialize and persist the xa-resources when possible. But I'm not sure if we can serialize ours...
The recovery process is really vague. Everything refers to XaResource.recover() but nothing says how to create them :(

server/hotrod/src/main/java/org/infinispan/server/hotrod/LifecycleCallbacks.java
pruivo: it used in hibernate cache to replace some components. I wasn't aware of that!
a flag in GlobalTransaction will eliminate the use of ServerTransactionOriginatorChecker. 
I actually would like to replace the GlobalTransaction to an interface and have different impls for embedded and remote...

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/SyncModeTransactionTable.java
pruivo: ok. I can change for CR1. this is Synchronization mode anyway...

server/hotrod/src/main/java/org/infinispan/server/hotrod/TransactionRequestProcessor.java
rvansa: The first LTS JDK with loom will be 14 (if everything goes well), that's 2 years...

pruivo:
updated!

server/hotrod/src/main/java/org/infinispan/server/hotrod/tx/table/TxState.java
pruivo: I going to need to fix it later. The externalizer is used mainly by the state transfer and in that case, it is value is "garbage". In any case, it ignores the access time if the originator is still in the view :)

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transaction/XaModeTransactionTable.java
danberindei: I found some more detail here, again Narayana-specific:

http://narayana.io/docs/product/index.html#d0e1032

Yeah, they like to serialize XAResources :) We could probably do it, we'd only need to name the cache managers and register them somewhere (e.g. JNDI) to allow deserialization.

I *think* we need to be a connector to implement crash recovery properly without TM-specific stuff, but I don't even want to investigate how much work that would mean right now :)

danberindei:
Integrated, thanks Pedro!

PR #6191, opened 2018-08-09
Test query on bytes ProtoBuffer type field

fax4ever:
I'm trying to query filtering on Protocol Butters type property.
Test:  `findMovieByRating_protoBytesTypeProperty`.

I'm trying to query filtering on starting with 'v' property.
Test: `findMovieByViews_startsV`.

I'm definitely trying to do something wrong. thank you

### DIFF
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/query/RemoteNonIndexedQueryStringTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/query/RemoteNonIndexedQueryStringTest.java
@@ -138,6 +138,14 @@ public class RemoteNonIndexedQueryStringTest extends RemoteQueryStringTest {
       super.testFullTextRegexp2();
    }
 
+   @Override public void findMovieByViews_startsV() {
+      // Not applicable to non-indexed caches
+   }
+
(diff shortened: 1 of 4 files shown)


galderz:
Assigning to @anistor

tristantarrant:
Retriggering CI

fax4ever:
Hi @tristantarrant! Just rebased it. Thank you

anistor:
But why do we need this test @fax4ever ? I remember having a chat about it a while ago. Querying on byte[] is not supported so why test it? I remember you were trying to query on byte[] because OGM uses a byte[] of size 1 to represent in place of byte type (because protobuf lacks a native byte type) but that's not the recommended way of mapping unsupported numerics to proto. The way to do it is to use int32.

fax4ever:
Hi @anistor! I agree with you, it would have no sense querying on ProtoBuf bytes type. The PR contains also another test, it seems that something goes wrong when the field name starts with a 'v'. But I'm definitely trying to do something wrong. thank you

fax4ever:
Thank you @anistor. Thus I will close the issue. :)

fax4ever:
Thank you @anistor. :P

PR #6196, opened 2018-08-09
ISPN-9421 Ensure worker configuration is applied to transport

tristantarrant:
https://issues.jboss.org/browse/ISPN-9421


diegolovison:
run performance tests please
cs

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
```diff
@@ -273,7 +273,7 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor e) {
    protected ExecutorService getExecutor(String threadPrefix) {
       if (this.executor == null || this.executor.isShutdown()) {
          DefaultThreadFactory factory = new DefaultThreadFactory(threadPrefix + "-ServerHandler");
-         int workerThreads = Integer.getInteger(WORKER_THREADS_SYS_PROP, configuration.workerThreads());
```
rvansa: Remove `WORKER_THREADS_SYS_PROP` if it's not used anymore.

server/hotrod/src/main/java/org/infinispan/server/hotrod/configuration/HotRodServerConfiguration.java
```diff
@@ -1,69 +1,83 @@
...
+      topologyAwaitInitialTransfer = attributes.attribute(TOPOLOGY_AWAIT_INITIAL_TRANSFER);
+      topologyStateTransfer = attributes.attribute(TOPOLOGY_STATE_TRANSFER);
    }
 
    public String proxyHost() {
-      return proxyHost;
+      return proxyHost.get();
+   }
+
+   public String actualHost() {
```
rvansa: `actualHost` isn't the best name IMO... `publicHost`? `exposedHost`?

tristantarrant: publicHost is good

server/core/src/main/java/org/infinispan/server/core/AbstractProtocolServer.java
```diff
@@ -184,4 +184,8 @@ public boolean isTransportEnabled() {
    public NettyTransport getTransport() {
       return transport;
    }
+
+   public int getWorkerThreads() {
+      return Integer.getInteger("infinispan.server." + protocolName.toLowerCase() + ".workerThreads", configuration.workerThreads());
```
rvansa: can't we keep this abstract?

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
tristantarrant: ok

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
```diff
@@ -273,7 +273,7 @@ public void rejectedExecution(Runnable r, ThreadPoolExecutor e) {
...
+         int workerThreads = getWorkerThreads();
```
rvansa: Hey, this should not use the same number as netty threads. This should still be able to grow to hundreds. In case that you have a persistence layer, you'll need those.

server/rest/src/main/java/org/infinispan/rest/configuration/RestServerConfiguration.java
```diff
@@ -1,38 +1,55 @@
...
-      this.extendedHeaders = extendedHeaders;
-      this.contextPath = contextPath;
-      this.maxContentLength = maxContentLength;
-      this.corsRules = corsRules;
-      this.compressionLevel = compressionLevel;
+   public static final AttributeDefinition<ExtendedHeaders> EXTENDED_HEADERS = AttributeDefinition.builder("extended-header", ExtendedHeaders.ON_DEMAND).immutable().build();
+   public static final AttributeDefinition<String> CONTEXT_PATH = AttributeDefinition.builder("context-path", "rest").immutable().build();
+   public static final AttributeDefinition<Integer> MAX_CONTENT_LENGTH = AttributeDefinition.builder("max-content-length", 10 * 1024 * 1024).immutable().build();
+   public static final AttributeDefinition<List<CorsConfig>> CORS_RULES = AttributeDefinition.builder("cors-rules", null, (Class<List<CorsConfig>>) (Class<?>) List.class)
+         .initializer(new AttributeInitializer<List<CorsConfig>>() {
```
rvansa: can't you use lambda here instead?

server/core/src/main/java/org/infinispan/server/core/AbstractProtocolServer.java
tristantarrant: Sure

diegolovison:
run performance tests please
cs

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
rvansa: OK, so what about configuring the Netty event loop threads as 'ioThreads' and using `workerThreads` for the blocking ops?

Looking into REST it seems that the operation is invoked directly on the netty thread, which is a problem. (how is it possible that we haven't seen scalability issues in perfack?) @slaskawi can you confirm?

tristantarrant: ioThreads is much better than slaveThreads. I'll use that. PerfAck doesn't use persistence, so maybe the impact is not as bad ?

rvansa: Hot Rod moves the execution to different thread pool with persistence because the `putAsync` it executes actually can be blocking (if this node is a primary owner). RPC is non-blocking internally so it does not become an issue.
However REST seems to run everything using synchronous `put` (unless special headers are included) and that means that the Netty thread will be blocked until we do the roundtrip to other owners.

diegolovison:
run performance tests please
cs

tristantarrant:
Refreshed the PR. 
* The netty slave/worker threads are now named IO Threads, they can be configured explicitly (through the ioThreads attribute on the endpoint) and use our CGroups-aware CPU count by default (ProcessorInfo.getAvalableProcessors())
* The server protocols configuration has been reworked to use Attribute/AttributeSet
* Server defaults reference the actual internal defaults

tristantarrant:
There is an additional commit which adds a subjectAltName to the generated certificates which makes the testsuite pass with stricter JDKs (>= 8u181) which require the hostname to be reflected in the certificate CN or subjectAltName

server/core/src/main/java/org/infinispan/server/core/configuration/ProtocolServerConfigurationBuilder.java
```diff
@@ -102,58 +109,55 @@ public SslConfigurationBuilder ssl() {
...
+         throw log.illegalSendBufferSize(attributes.attribute(SEND_BUF_SIZE).get());
       }
-      if (sendBufSize < 0) {
-         throw log.illegalSendBufferSize(sendBufSize);
+      if (attributes.attribute(RECV_BUF_SIZE).get() < 0) {
+         throw log.illegalReceiveBufferSize(attributes.attribute(RECV_BUF_SIZE).get());
       }
-      if (recvBufSize < 0) {
-         throw log.illegalReceiveBufferSize(recvBufSize);
+      if (attributes.attribute(IO_THREADS).get() < 0) {
```
rvansa: It seems to me that `0` should be illegal value, too, since it does not retrigger default settings (based on CPU count). Same for worker threads.

tristantarrant: 0 is perfectly legal for netty slaves https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/MultithreadEventLoopGroup.java#L52

tristantarrant: As for documenting properties, my opinion is that we should but only for those properties which are not exposed in the configuration, because there we support property expansion anyway.
I was not happy about keeping that workerThreads property either, but did so because of backwards compatibility. Maybe we should list all such properties as deprecated so that we can remove them  in Infinispan 10?

rvansa: If it's not documented, there's no place to deprecate. We could remove that in 10 without warning.

rvansa:
Integrated, thanks!

PR #6197, opened 2018-08-10
ISPN-9408 Support for query results in different formats in Hot Rod

gustavonalle:
https://issues.jboss.org/browse/ISPN-9408

Adds support for JSON queries over the Hot Rod protocol


galderz:
Looks good, integrating.

galderz:
I'll wait for CI...

bom/pom.xml
```diff
@@ -104,6 +104,7 @@
       <version.commons.logging>1.2</version.commons.logging>
       <version.commons.math>2.2</version.commons.math>
       <version.gnu.getopt>1.0.13</version.gnu.getopt>
+      <version.gson>2.8.2</version.gson>
```
anistor: So I guess I'll need to update protostream to use the same gson version @gustavonalle

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/query/RemoteQuery.java
```diff
@@ -72,9 +81,20 @@ private void executeQuery() {
...
-         QueryResponse response = await(op.execute());
-         totalResults = (int) response.getTotalResults();
-         results = unwrapResults(response.getProjectionSize(), response.getResults());
+         Object response = await(op.execute());
+         if (response instanceof QueryResponse) {
+            QueryResponse resp = (QueryResponse) response;
+            totalResults = (int) resp.getTotalResults();
+            results = unwrapResults(resp.getProjectionSize(), resp.getResults());
+         }
+         if (response instanceof JsonObject) {
```
anistor: Why no else here?

bom/pom.xml
gustavonalle: Gson is shaded inside protostream @anistor, so I decided to use an explicit dependency so make it easier to upgrade

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/QuerySerializers.java
```diff
@@ -0,0 +1,24 @@
...
+import org.infinispan.commons.dataconversion.MediaType;
+
+public final class QuerySerializers {
+
+   private final ConcurrentHashMap<String, QuerySerializer<?>> serializers = new ConcurrentHashMap<>(2);
+
+   public QuerySerializers() {
+   }
+
+   public void addSerializer(MediaType mediaType, QuerySerializer<?> qQuerySerializer) {
```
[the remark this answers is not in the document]
gustavonalle: ops

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/query/RemoteQuery.java
gustavonalle: Because only two kinds of query response are possible. Actually it'd be better to extract an interface, let me fix this

bom/pom.xml
anistor: right. forgot about the shading

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/query/RemoteQuery.java
anistor: Sure, but the code would have allowed an unrecognised 3rd alternative to go unnoticed.

gustavonalle: indeed, I forgot to revisit this code :)

gustavonalle:
@anistor updated.

anistor:
@gustavonalle Looks good but I do not see any reason to keep those as separate commits. So I'll squash it and merge after the new CI round completes.

gustavonalle:
Feel free, I was going to squash them after CI completion

gustavonalle:
CI seems ok

PR #6198, opened 2018-08-13
ISPN-7035 Spring 5

karesti:
https://issues.jboss.org/browse/ISPN-7035

Concerning Spring support (read all [here](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Versions))

- 5.1.x is the latest and recommended line, supported until the end of 2019 (to be superseded by the longer-lived 5.2.x in mid 2019).

- 5.0.x has been superseded by 5.1.x already but keeps being supported for Boot 2.0's lifetime still (~ March 2019).

- 4.3.x is the last feature branch of the 4th generation. It comes with extended maintenance until 2020, so you may migrate your applications to 5.x at your own pace.


karesti:
@galderz I'm trying but it's too hard due to the configuration. I realised locally that test pass because it's using the old xsd (not the splitted ones) and I'm thinking that this split won't work if somebody is using in spring an embedded AND remote cache managers at the same time.

The problem @galderz is the commons module where we define today the xml configuration.

karesti:
@galderz well, test pass with the xsd split with another modification I've done locally, but if somebody is using remote and embedded cache managers I'm not sure it will.

tristantarrant:
This work should be postponed to Infinispan 10.

karesti:
@tristantarrant ok !

spring/spring5/spring5-remote/src/test/java/org/infinispan/spring/remote/session/InfinispanRemoteSessionRepositoryTest.java
```diff
@@ -37,9 +39,9 @@ public void afterMethod() {
 
    @AfterClass
    public void afterClass() {
+      embeddedCacheManager.stop();
```
galderz: Why change the order? You normally stop in opposite order to creation. So, if you create embedded cache manager first and then the server, you normally stop server first, then the cache manager.

karesti: probably changed by mistake. I will put it the way it was. good catch

karesti:
@galderz I changed the order and removed the drop table

galderz:
Thx @karesti, integrating...

PR #6200, opened 2018-08-14
ISPN-9409 Fix HotRodMergeTest

gustavonalle:
https://issues.jboss.org/browse/ISPN-9409

PR #6202, opened 2018-08-14
ISPN-9185 Remove compat mode from the Hot Rod protocol

gustavonalle:
https://issues.jboss.org/browse/ISPN-9185

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/PingOperation.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/PingOperation.java
@@ -14,10 +14,12 @@ import org.infinispan.client.hotrod.impl.transport.netty.ChannelOperation;
 import org.infinispan.client.hotrod.impl.transport.netty.HeaderDecoder;
 import org.infinispan.client.hotrod.logging.Log;
 import org.infinispan.client.hotrod.logging.LogFactory;
+import org.infinispan.commons.dataconversion.MediaType;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.DecoderException;
 import net.jcip.annotations.Immutable;
 
 /**
@@ -27,7 +29,7 @@ import net.jcip.annotations.Immutable;
  * @since 4.1
  */
 @Immutable
-public class PingOperation extends HotRodOperation<PingOperation.PingResult> implements ChannelOperation {
+public class PingOperation extends HotRodOperation<PingOperation.PingResponse> implements ChannelOperation {
    private static final Log log = LogFactory.getLog(PingOperation.class);
    private static final boolean trace = log.isTraceEnabled();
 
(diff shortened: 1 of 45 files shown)


gustavonalle:
Test failure is not related (fixed by #6200)

gustavonalle:
rebased

gustavonalle:
Only failures are on ```org.infinispan.counter.RemoveCounterTest``` that does not seem related, @pruivo  are you aware of those?

pruivo:
@gustavonalle it is new to me. maybe some issue with the VM :)

server/core/src/main/java/org/infinispan/server/core/dataconversion/JBossMarshallingTranscoder.java
```diff
@@ -85,6 +86,8 @@ public Object transcode(Object content, MediaType contentType, MediaType destina
       }
       if (destinationType.equals(APPLICATION_UNKNOWN)) {
          try {
+            //TODO: Remove wrapping of byte[] into WrappedByteArray from the Hot Rod Multimap operations.
```
anistor: Is there a jira for this todo?

gustavonalle: Not yet, I'll create it

server/hotrod/src/main/java/org/infinispan/server/hotrod/Encoder2x.java
```diff
@@ -475,14 +495,35 @@ private void writeHeaderNoTopology(ByteBuf buffer, long messageId, HotRodOperati
    }
 
    private void writeStatus(HotRodHeader header, ByteBuf buf, HotRodServer server, boolean compatibilityEnabled, OperationStatus status) {
-      if (server == null || HotRodVersion.HOTROD_24.isOlder(header.version))
+      if (server == null || HotRodVersion.HOTROD_24.isOlder(header.version) || HotRodVersion.HOTROD_29.isAtLeast(header.version))
          buf.writeByte(status.getCode());
       else {
-         OperationStatus st = OperationStatus.withCompatibility(status, compatibilityEnabled);
+         OperationStatus st = OperationStatus.withLegacyStorageHint(status, compatibilityEnabled);
```
anistor: ```withLegacyStorageHint``` is a strange name. Why do we avoid 'compatibility' from naming? I'd call this ```withLegacyCompatibility``` so we'll know from a glance ... :)

gustavonalle: The idea is to wipe compat mode, as long as variable names still refers to it, it'll not be truly dead!  :)

gustavonalle: I can put it back if you think it's essential, but regardless, I think storageHint reflects more precisely the fact that how data is stored in the cache piggybacks on the response of every operation

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/PingOperation.java
```diff
@@ -74,30 +77,64 @@ public void acceptResponse(ByteBuf buf, short status, HeaderDecoder decoder) {
...
+         return status;
+      }
 
       public boolean isSuccess() {
-         return this == SUCCESS || this == SUCCESS_WITH_COMPAT;
+         return HotRodConstants.isSuccess(status);
+      }
+
+      public boolean isObjectStorage() {
+         boolean storageHinted = status == NO_ERROR_STATUS_OBJ_STORAGE
```
galderz: I'm not sure I like these checks staying here. They should be moved to Codec28 so that these checks are only done if using HR protocol 2.8 (or whichever is the last HR protocol version that sent them). If you leave them here, you're going to carry checking even with protocol 2.9 which is wasteful since the server does not send them.

server/hotrod/src/main/java/org/infinispan/server/hotrod/Encoder2x.java
anistor: So there will be no mention of it, not even in a migration guide? That would be so sad:(

gustavonalle: It will be in the migration guide for sure

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/PingOperation.java
gustavonalle: ok, let me refactor it

server/core/src/main/java/org/infinispan/server/core/dataconversion/JBossMarshallingTranscoder.java
gustavonalle: https://issues.jboss.org/browse/ISPN-9458

gustavonalle:
@galderz updated

gustavonalle:
The only failures are on ```org.infinispan.counter.RemoveCounterTest``` but those also happen on master

PR #6206, opened 2018-08-17
ISPN-9434 Reading JSON content from untyped caches return escaped quotes

gustavonalle:
https://issues.jboss.org/browse/ISPN-9434


galderz:
Thanks @gustavonalle

PR #6210, opened 2018-08-20
ISPN-9440 Hibernate second level cache - Add native query test

diegolovison:
ISPN-9440 Hibernate second level cache - Add native query test

### DIFF
--- a/hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/NativeQueryTest.java
+++ b/hibernate/cache-commons/src/test/java/org/infinispan/test/hibernate/cache/commons/functional/NativeQueryTest.java
@@ -1,0 +1,81 @@
+package org.infinispan.test.hibernate.cache.commons.functional;
+
+import java.util.List;
+
+import org.hibernate.Cache;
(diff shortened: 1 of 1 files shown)


galderz:
Why the need for such a test? What is the difference in Hibernate code that leads to think that the second level cache would deal with such native query differently? I don't want to add more tests unless they are testing something not already covered from a second-level cache perspective. IOW, we're not trying to take all Hibernate features and run them as tests with second-level cache and see whether something fails... It just wouldn't scale.

diegolovison:
If this test was covered somehow. I fine to close this PR

galderz:
Queries are being tested, but I don't know whether native queries are somehow different from a second-level cache perspective. That's what I was hoping you'd look into ;)

diegolovison:
Yep. I did it :)

diegolovison:
createSQLQuery will invalidate all regions. If you would like to avoid that you can set `addSynchronizedEntityClass`.
Invalidation was covered in other tests.
I am closing this PR.

galderz:
Thanks @diegolovison :)

PR #6215, opened 2018-08-22
ISPN-9426 Distributed script execution returning null results in Null…

gustavonalle:
…PointerException

https://issues.jboss.org/browse/ISPN-9426

### DIFF
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExecTypedTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/ExecTypedTest.java
@@ -2,6 +2,7 @@ package org.infinispan.client.hotrod;
 
 import static org.infinispan.client.hotrod.test.HotRodClientTestingUtil.withClientListener;
 import static org.infinispan.client.hotrod.test.HotRodClientTestingUtil.withScript;
+import static org.infinispan.commons.dataconversion.MediaType.APPLICATION_JSON;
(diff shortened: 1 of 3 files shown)


gustavonalle:
The script output is dictated by the media type set in the header when executing it. So differentiating between ```null``` and ```""``` seems more correct when the script output is a textual format. It'd allow to differentiate between:
* non existent values  ```null```, e.g. ```cache.get('inexistent-key')```
* empty strings  ```""```, e.g. ```cache.get('key-with-empty-value')```
* null string ```"null"```, e.g.  ```cache.get('key-with-string-value')```

gustavonalle:
Also, in JSON terms,  both ```["",""]```  and ```[null, null]``` are valid and mean different things

gustavonalle:
For java clients running scripts without any particular result format specified (the default), we currently erroneously map ```null``` to ```""```, this should be fixed in another JIRA

galderz:
> Also, in JSON terms, both ["",""] and [null, null] are valid and mean different things

I'm fine with both, but looking at the code, aren't you returning `["null", "null"]` ?

galderz:
https://github.com/infinispan/infinispan/pull/6215/files#diff-d030c6ace7871c24c274b7ae3728e904R72

gustavonalle:
It is returning ```[null,null]``` (without quotes):  https://github.com/infinispan/infinispan/pull/6215/files#diff-c94e51670c5b2358c05a2ac4c4406e4fR115

galderz:
@gustavonalle Worked fine, thanks!

PR #6218, opened 2018-08-23
ISPN-9453: Docs update for FORCE_WRITE_LOCK non-tx caches

oraNod:
Adding note to clarify about FORCE_WRITE_LOCK flag for non-tx caches

### DIFF
--- a/documentation/src/main/asciidoc/user_guide/clustering.adoc
+++ b/documentation/src/main/asciidoc/user_guide/clustering.adoc
@@ -779,9 +779,7 @@ try {
 The correct way to implement this is to use
 `cache.getAdvancedCache().withFlags(Flag.FORCE_WRITE_LOCK).get(k)`.
 
-In caches with optimistic locking writes can return a stale previous value as well, and
-the only way protect against it is to enable write-skew checks and to catch
-`WriteSkewException`.
+In caches with optimistic locking, writes can also return stale previous values. Write skew checks can avoid stale previous values. For more information, see <<tx:write-skew, Write Skews>>.
 
 === Partition handling
 
--- a/documentation/src/main/asciidoc/user_guide/transactions.adoc
+++ b/documentation/src/main/asciidoc/user_guide/transactions.adoc
@@ -107,7 +107,7 @@ builder.versioning()
 
 * `isolation` - configures the isolation level. Check section <<tx:isolation-levels>> for more details.
 Default is `REPEATABLE_READ`.
-* `write-skew` - enables the write skew check. Check section <<tx:write-skew>> for more details. Default is `false`.
+* `write-skew` - enables write skew checks. Check section <<tx:write-skew>> for more details. Default is `true`. Defaults to `false` automatically for the `READ_COMMITTED` isolation level. This attribute is deprecated.
 * `locking` - configures whether the cache uses optimistic or pessimistic locking. Check section <<tx:locking>> for more details.
 Default is `OPTIMISTIC`.
 * `auto-commit` - if enable, the user does not need to start a transaction manually for a single operation. The transaction is automatically started and committed.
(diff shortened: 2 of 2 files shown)


oraNod:
@rvansa can you review this doc update, please?

documentation/src/main/asciidoc/user_guide/transactions.adoc
```diff
@@ -240,10 +240,14 @@ Pessimistic transactions are more costly by their nature: each write operation p
...
 To detect and rollback one of the transaction, `write-skew` should be enabled.
 
-NOTE: The write skew check is only performed for `REPEATABLE_READ` isolation.
+[NOTE]
+====
+* The write skew check is only performed for `REPEATABLE_READ` isolation.
 
-NOTE: Pessimistic transaction does not perform any write skew check.
-It can be avoided by locking the key at read time. Look how at the example below.
+* Pessimistic transactions do not need to perform write skew checks if they lock keys at read time with `Flag.FORCE_WRITE_LOCK`.
```
rvansa: This wording is confusing: in pessimistic mode there's no WSC functionality, neither when you don't lock the keys explicitly - you're just left with stale reads.

wburns:
Actually thinking about this more, I wonder if we should enforce this in the code. We currently don't validate any of the flags passed in by the user in the `DecoratedCache`. We could verify that the underlying cache is configured pessimistic tx and if not throw some type of exception if `FORCE_WRITE_LOCK` is passed. @tristantarrant WDYT?

oraNod:
@rvansa I gave this a more thorough read and tried to re-organize the info a bit. I think part of the confusion was that info wasn't clearly separated. it makes more sense to me now. hopefully the changes look alright to you. cheers.

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -779,9 +779,7 @@ try {
 The correct way to implement this is to use
 `cache.getAdvancedCache().withFlags(Flag.FORCE_WRITE_LOCK).get(k)`.
 
-In caches with optimistic locking writes can return a stale previous value as well, and
-the only way protect against it is to enable write-skew checks and to catch
-`WriteSkewException`.
+In caches with optimistic locking, writes can also return stale previous values. The only way to avoid stale previous values is to enable write-skew checks to catch `WriteSkewException`.
```
rvansa: This sounds like you must enable write skew checks explicitly, but that's no longer true (since version 8.x or so). Write skew checks are on by default; you can disable them as a performance optimization.

documentation/src/main/asciidoc/user_guide/transactions.adoc
```diff
@@ -235,34 +235,15 @@ That is because the optimistic transactions rollback if data has changed between
...
 
-=== Write Skew [[tx:write-skew]]
+=== Write Skews [[tx:write-skew]]
 
-The write skew anomaly occurs when 2 transactions read and update the same key and both of them can commit successfully without having seen the update performed by the other.
-To detect and rollback one of the transaction, `write-skew` should be enabled.
+Write skews occur when two transactions independently and simultaneously read and write to the same key. The result of a write skew is that both transactions successfully commit updates to the same key but with different values.
 
-NOTE: The write skew check is only performed for `REPEATABLE_READ` isolation.
+Enabling `write-skew` in the {brandname} configuration automatically detects and rolls back one of the transactions. However, write skew checks apply to `REPEATABLE_READ` isolation only.
```
rvansa: again, enabled by default

oraNod: @rvansa from the user guide: "write-skew - enables the write skew check. Check section Write Skew for more details. Default is false."

http://infinispan.org/docs/stable/user_guide/user_guide.html#tx:configuration

looking at the schema docs, the default is true. so it looks like docs need to be updated here. schema docs also say write skew is deprecated. is that the case?

rvansa:
OK, please fix conflicts and I'll merge.

documentation/src/main/asciidoc/user_guide/transactions.adoc
```diff
@@ -276,7 +262,27 @@ Or
...
 ----
 
-NOTE: `SIMPLE` versioning is an implementation of the proposed `EntryVersion` interface, backed by a long that is incremented each time the entry is updated.
+==== Forcing write locks on keys in pessimitic transactions
+
+Write-skews checks do not occur with pessimistic transactions. To avoid write-skews with pessimistic transactions, lock keys at read-time with `Flag.FORCE_WRITE_LOCK`. For example,
+
+[source, java]
+----
+if (!cache.getAdvancedCache().lock(key)) {
```
wburns: I wonder if we should mention this has to be in an ongoing transaction?

oraNod: @wburns the "FORCE_WRITE_LOCK" flag can be used with ongoing transactions? so the transaction cannot be complete, aborted, or some other state?

documentation/src/main/asciidoc/user_guide/transactions.adoc
```diff
@@ -276,7 +262,34 @@ Or
...
+* In non-transactional caches, `Flag.FORCE_WRITE_LOCK` does not work. The `get()` call reads the key value but does not acquire locks remotely.
+
+* You should use `Flag.FORCE_WRITE_LOCK` with transactions in which the entity is updated later in the same transaction.
+====
+
+Compare the following code snippets for an example of `Flag.FORCE_WRITE_LOCK`:
+
+[source, java]
+----
+//Abort the transaction if the key is not locked.
```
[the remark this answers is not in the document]
oraNod: thanks, Radim! I think those code examples look a lot better now.

rvansa:
Integrated, thanks!

oraNod:
@rvansa @wburns my first upstream contribution. woot! thanks, guys.

PR #6219, opened 2018-08-27
fixed default value of isolation level of document

kazuhira-r:
I think the default isolation level after Infinispan 9.0 is REPEATABLE READ, but the document is READ COMMITTED.

[Enable write-skew for optimistic + repeatable-read transactions](https://issues.jboss.org/browse/ISPN-7613)

https://github.com/infinispan/infinispan/blob/9.4.0.CR1/core/src/main/java/org/infinispan/configuration/cache/LockingConfiguration.java#L19


rvansa:
Thanks, integrated.

Btw., the full process of filing a PR involves creating a [tracking JIRA](https://issues.jboss.org/) and the commit header should start with the `ISPN-xxxx` identifier and upper case letter.

PR #6222, opened 2018-08-29
ISPN-9459 ISPN-7940 Deprecate compat mode and remove it from the Memcached server

gustavonalle:
https://issues.jboss.org/browse/ISPN-9459
https://issues.jboss.org/browse/ISPN-7940

Summary of changes:

* The Memcached server does not need compat mode anymore (or storing keys as pojos) to share a cache with other endpoints
* The Memcached server no longer stores keys as ```java.lang.String```, but as UTF-8
* The Memcached server has an extra optional configuration ```client-encoding``` to accept and return values in a format more convenient to clients.
* Compatibility Mode is deprecated, some usages removed.


gustavonalle:
Added a new commit with ISPN-7940

tristantarrant:
Merged, thanks !

PR #6223, opened 2018-08-30
ISPN-9464 Notify listeners without holding the shared topology lock

danberindei:
https://issues.jboss.org/browse/ISPN-9464


wburns:

PR #6226, opened 2018-08-30
ISPN-9475 AbstractInfinispanTest mistakenly detects some test with params as duplicates

anistor:
* the culprit is AbstractInfinispanTest
* after fixing this we find many tests were previously ignored by mistake and need params to be declared 
https://issues.jboss.org/browse/ISPN-9475

### DIFF
--- a/core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
+++ b/core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
@@ -29,6 +29,7 @@ import java.util.concurrent.locks.LockSupport;
 import java.util.function.BooleanSupplier;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
+
 import javax.transaction.TransactionManager;
 
 import org.infinispan.commons.api.BasicCache;
@@ -50,7 +51,6 @@ import org.jgroups.stack.Protocol;
 import org.testng.IMethodInstance;
 import org.testng.IMethodInterceptor;
 import org.testng.ITestContext;
-import org.testng.TestNGException;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeClass;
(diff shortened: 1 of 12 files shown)


core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
```diff
@@ -96,7 +96,7 @@
             if (instance instanceof AbstractInfinispanTest) {
                String instanceName = ((AbstractInfinispanTest) instance).getTestName();
                Object otherInstance = instancesByName.putIfAbsent(instanceName, instance);
-               if (otherInstance != null) {
+               if (otherInstance != null && instance.getClass() != otherInstance.getClass()) {
```
danberindei: The interceptor only processes the methods of one class (i.e. TestNG XML test) at once, so this condition will never be true. Better to keep the condition but log something to the console instead of throwing the exception and skipping all the methods.

anistor:
retest this please

core/src/test/java/org/infinispan/test/AbstractInfinispanTest.java
anistor: Done!

anistor:
merged in master and 9.3.x.

gustavonalle:
Apparently this PR caused the following failures:
```
    org.infinispan.scripting.ScriptingDataStoresTest.testScriptWithoutParam[OFF_HEAP]
    org.infinispan.scripting.ScriptingDataStoresTest.testScriptWithParam[OBJECT]
    org.infinispan.scripting.ScriptingDataStoresTest.testScriptWithParam[OFF_HEAP]
    org.infinispan.scripting.ScriptingDataStoresTest.testScriptWithParam[BINARY]
    org.infinispan.scripting.ScriptingDataStoresTest.testScriptWithoutParam[OBJECT]
    org.infinispan.scripting.ScriptingDataStoresTest.testScriptWithoutParam[BINARY]
```

anistor:
That's an issue that went under the radar before applying this fix, see ISPN-9474 @gustavonalle . Maybe you or @tristantarrant know how to fix ScriptingDataStoresTest.

But it's a bit strange you say this fails in 9.3.x. for me it only failed on 9.4.x, but maybe I'm wrong

gustavonalle:
ok @anistor, missed the fact this test was not being executed before, I'll take a look

PR #6227, opened 2018-08-31
ISPN-9465 Cannot join cache after killing the only member 9.3.x

danberindei:
https://issues.jboss.org/browse/ISPN-9465

When the last cache member leaves, the coordinator's ClusterCacheStatus
stays in state RUNNING, and the next join doesn't reinitialize the cache
topology.


wburns:
Integrated into 9.3.x, thanks @danberindei !

PR #6228, opened 2018-08-31
[WIP] Downstream backport

pruivo:
testing some changes


tristantarrant:
We don't want the RHQ plugin.

tristantarrant:
bom/pom.xml must not be removed. It is essentially our public "contract" of artifacts we publish.

tristantarrant:
The doclets have been moved upstream to javadoc/doclets. No need to bring them back in tools

pruivo:
> bom/pom.xml must not be removed. It is essentially our public "contract" of artifacts we publish.

as far as I understand, it is replaced by `build-configuration/bom/pom.xml` and it is deployed in maven with the artifact-id `infinispan-bom`. Do we still need a `bom/pom.xml`? what will be its content?

> The doclets have been moved upstream to javadoc/doclets. No need to bring them back in tools

Now it makes sense.

distribution/pom.xml
```diff
@@ -389,8 +391,8 @@
             <configuration>
                <outputDirectory>${project.build.directory}/distribution</outputDirectory>
                <workDirectory>${project.build.directory}/assembly/work</workDirectory>
-               <!-- Do not attach ZIP files to the Maven repository -->
-               <attach>false</attach>
+               <!-- (Do not attach ZIP files to the Maven repository) -->
+               <attach>true</attach><!-- we need that actually -->
```
tristantarrant: Please let us make this controllable via a property. For community we do NOT want to deploy distribution zips to Maven.

karesti:
needs rebase

distribution/pom.xml
pruivo: I guess this can be removed from community too:

https://github.com/infinispan/infinispan/pull/6228/files#diff-9cf918424f312edfa0b7dd14b008eac5R174

tristantarrant: Indeed. Remember to rebase as well

pruivo:
updated

server/integration/build/assembly-fat-server.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tristantarrant: I want all distro assemblies (servers, sources, etc) to be relegated to the distribution module and not scattered everywhere. Why is this different from the server zip generated in distribution ?

server/integration/build/pom.xml
```diff
@@ -99,49 +100,134 @@
...
                             <finalName>${server.output.dir.prefix}-${infinispan.brand.version}</finalName>
                             <appendAssemblyId>false</appendAssemblyId>
                             <outputDirectory>${project.build.directory}</outputDirectory>
-                            <workDirectory>${project.build.directory}/assembly/work</workDirectory>
+                            <workDirectory>${project.build.directory}/assembly/work-dir/thin</workDirectory>
                             <tarLongFileMode>gnu</tarLongFileMode>
                         </configuration>
                     </execution>
+
+                    <!-- bin vs. fat? -->
```
tristantarrant: Please remove these commented sections

server/integration/build/assembly-fat-server.xml
pruivo: this is no longer in use and I'll remove it. But, the only difference I can see, is the copy of the licenses.

tristantarrant:
We don't need server/integration/pom.xml to generate a source zip just with the server sources, since the distribution module already generates a zip with all sources.

PR #6232, opened 2018-09-03
ISPN-9127 Component registry

danberindei:
https://issues.jboss.org/browse/ISPN-9127

I extracted the main functionality of the component registries in a `BasicComponentRegistry` interface, implemented by `BasicComponentRegistryImpl`.

* It has no cache-specific logic or references to specific components.
* All the dependencies are wired before being injected in another component. `ComponentRef<T>` fields or injection method parameters can be used as lazy dependencies to break dependency cycles. Lazy dependencies are only created and wired when the `wired()` method is called.
* Start/stop priorities only have a meaning within a component, if a component has multiple start/stop methods. A component's start methods are always invoked after the start methods of its non-lazy dependencies.
* `registerComponent()` doesn't overwrite existing components.
* `replaceComponent()` replaces existing components, but it is intended only for tests.
* `getComponent()` and `registerComponent()` return a `ComponentRef<T>`, and the component is always at least wired (all the dependencies have been injected). Use `running()` to get the actual instance, or `wired()` iff `running()` would create a cycle.

I also changed the behaviour of `[Global]ComponentRegistry` a bit:
* `getComponent()` always tries to create and start the component (skips the start if the registry is only INSTANTIATED)
* `registerComponent()` doesn't overwrite existing components


tristantarrant:
Needs rebase

tristantarrant:
And CI says it's green !

gustavonalle:
Let's wait for CI

gustavonalle:
CI timed out on some query tests:

A few of those were thrown:
```
Exception in thread "remote-thread-MassIndexingTest-NodeB-p11241-t5" org.infinispan.IllegalLifecycleStateException: Cannot wire or start components while the registry is not running
	at org.infinispan.factories.impl.BasicComponentRegistryImpl.prepareWrapperChange(BasicComponentRegistryImpl.java:610)
```
and the one that crashed:
```
Test org.infinispan.query.affinity.ShardDistributionTest.testAllocation has been running for more than 300 seconds. Interrupting the test thread and dumping thread stacks of the test suite process and its children.
Dumping thread stacks of process 17174 to /home/infinispan/workspace/Infinispan_PR-6232-CX6DTOWKPIX5PKDSWX2BR3HEF5425SISGE6LZF6JTVH5U7U5JLAQ/query/threaddump-org_infinispan_query_affinity_ShardDistributionTest_testAllocation-2018-10-04-17174.log
```

gustavonalle:
needs rebase again

gustavonalle:
I am trying to reproduce the query issue here

danberindei:
Updated! I thought I'd fix the `IllegalLifecycleStateException` in the log as well, but it's taking longer than I thought so I'm going to make it a separate PR.

gustavonalle:
Merged, thanks @danberindei !

PR #6235, opened 2018-09-05
ISPN-9498 Near Cache filter converter which filters out values and cr…

tristantarrant:
https://issues.jboss.org/browse/ISPN-9484

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
@@ -16,7 +16,7 @@ import org.infinispan.client.hotrod.event.ClientCacheEntryModifiedEvent;
 import org.infinispan.client.hotrod.event.ClientCacheEntryRemovedEvent;
 import org.infinispan.client.hotrod.event.ClientCacheFailoverEvent;
 import org.infinispan.client.hotrod.event.impl.ClientListenerNotifier;
-import org.infinispan.client.hotrod.impl.MetadataValueImpl;
+import org.infinispan.client.hotrod.exceptions.HotRodClientException;
 import org.infinispan.client.hotrod.logging.Log;
 import org.infinispan.client.hotrod.logging.LogFactory;
 import org.infinispan.commons.io.UnsignedNumeric;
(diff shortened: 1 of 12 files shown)


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
```diff
@@ -49,17 +50,27 @@ public void start(RemoteCache<K, V> remote) {
...
+         listener = new SmartInvalidatedNearCacheListener<>(this, remote.getRemoteCacheManager().getMarshaller());
+         try {
+            remote.addClientListener(listener);
+         } catch (HotRodClientException e) {
+            // It failed, use the old listener instead
+            listener = new InvalidatedNearCacheListener<>(this);
+            remote.addClientListener(listener);
+         }
+      } else {
+         listener = new EagerNearCacheListener(this, remote.getRemoteCacheManager().getMarshaller());
```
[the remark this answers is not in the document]
tristantarrant: good !

tristantarrant:
@danberindei addressed your comments

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
```diff
@@ -49,17 +49,22 @@ public void start(RemoteCache<K, V> remote) {
...
       }
    }
 
-   private Object createListener(RemoteCache<K, V> remote) {
-      return config.mode().invalidated()
-            ? new InvalidatedNearCacheListener(this)
-            : new EagerNearCacheListener(this, remote.getRemoteCacheManager().getMarshaller());
+   private void createAndRegisterListener(RemoteCache<K, V> remote) {
+      // Attempt to register the smart invalidation listener
+      listener = new KeyOnlyInvalidatedNearCacheListener<>(this, remote.getRemoteCacheManager().getMarshaller());
```
wburns: I am not a big fan of this, the `ClientCacheEntryCreatedEvent`, `ClientCacheEntryModifiedEvent`, `ClientCacheEntryRemovedEvent` & `ClientCacheEntryExpiredEvent` just allow access to the key and possibly version. Thus these should always not contain the value for an event. This would then make all listeners faster as well, not just near cache.

Assuming this is in place then we just need to remove the `handleCreatedEvent` method that was there before.

I will look at this a little bit to see if there is an easy way to do this.

tristantarrant: I was fooled by https://github.com/infinispan/infinispan/blob/1560df5ee5e36d7ac4eee784fd1663af39768d00/server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java#L383 passing the value, which is then ignored.
However for the server-side event, we do want to avoid sending values around.

tristantarrant: Actually, wouldn't a flag on the listener, akin to IGNORE_RETURN_VALUES, be useful in the more general use-case ?

wburns: I added

```   
public void testNoConverterFactoryCustomEvents() {
      NoConverterFactoryListener l = new NoConverterFactoryListener<>(remoteCacheManager.getCache());
      withClientListener(l, remote -> {
         l.expectNoEvents();
         remote.put(1, "one");
         l.expectCreatedEvent(new CustomEvent(1, "one", 0));
      });
   }
```
to the class `ClientCustomEventsTest.java` and when I debug I see that it creates a `CreatedEventImpl` which is the non custom form. This actually causes no event to be seen at all by the client then.

wburns: I should have some changes for this pretty soon.

wburns: Having some issues with the whole key and value data conversions now though - have to keep plugging.

wburns:
https://github.com/infinispan/infinispan/pull/6241 supersedes this.

PR #6236, opened 2018-09-05
ISPN-9485 RoundRobinBalancingStrategy always starts from server 0

danberindei:
https://issues.jboss.org/browse/ISPN-9485

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/RoundRobinBalancingStrategy.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/RoundRobinBalancingStrategy.java
@@ -4,6 +4,7 @@ import java.net.SocketAddress;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Set;
+import java.util.concurrent.ThreadLocalRandom;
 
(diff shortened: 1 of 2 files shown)


client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/transport/tcp/RoundRobinBalancingStrategy.java
```diff
@@ -28,7 +29,7 @@ public void setServers(Collection<SocketAddress> servers) {
       this.servers = servers.toArray(new SocketAddress[servers.size()]);
       // keep the old index if possible so that we don't produce more requests for the first server
       if (index >= this.servers.length) {
-         index = 0;
+         index = ThreadLocalRandom.current().nextInt(this.servers.length);
```
tristantarrant: Shouldn't `index` be initialized to MAX_INT in order to trigger this on 1st call ?

danberindei: Oops, I meant to remove the `if`

danberindei:
Updated to remove the if and to fix the test

wburns:

PR #6240, opened 2018-09-06
ISPN-7409 Marshallers separated into Persistence and Internal marshal…

ryanemerson:
…ler.

PersistenceMarshaller created based upon ProtoStream. Internal
persistence classes are marshalled using the messages defined in
'persistence.proto' and their corresponding MessageMarshaller
implementation.

User objects are marshalled via the marshaller specified in
SerializationConfiguration and the resulting bytes are wrapped in a
UserObject message. The default marshaller is the JBossMarshaller
with just the configured external externalizers loaded.

StreamingMarshaller interface has been deprecated for external use and
has been replaced with the simplified StreamAwareMarshaller interface.

@wburns @danberindei @anistor Based upon our discussions on Zulip I have created the `StreamAwareMarshaller`. I couldn't think of a better name, but accept that is pretty terrible, so I'm very much open to suggestions. Currently all the core tests pass on my local machine, but haven't tested further than that. If you could take a look before I get back from PTO in October that would be much appreciated :smile:.

@anistor Do you know of a better way of wrapping/handling user objects with protostream than using the `UserObject` message?

Open questions:
- How do you think we should expose the serialization context of the `PersistenceMarshaller` to external modules. Is it sufficient to simply add a method such as `PersistenceManager::getPersistenceMarshaller`?
- Do any of the current modules require any non commons/core classes to be persisted?

N.B. I have `SerializationConfiguration::advancedExternalizers` is still in place so that modules can still utilise custom externalizers for internal communication.


core/src/main/java/org/infinispan/functional/impl/MetaParams.java
```diff
@@ -59,12 +59,17 @@
...
 @Experimental
-final class MetaParams implements Iterable<MetaParam<?>> {
+public final class MetaParams implements Iterable<MetaParam<?>> {
 
    private static final MetaParam<?>[] EMPTY_ARRAY = {};
    private MetaParam<?>[] metas;
    private int length;
 
+   public MetaParams() {
+      this.metas = new MetaParam[]{};
```
anistor: EMPTY_ARRAY was not good enough :)

commons/pom.xml
```diff
@@ -14,6 +14,11 @@
    <description>Infinispan Commons</description>
 
    <dependencies>
+      <dependency>
```
anistor: This has the potential to upset many.

commons/src/main/java/org/infinispan/commons/marshall/StreamAwareMarshaller.java
```diff
@@ -0,0 +1,37 @@
...
+    */
+   void writeObject(Object o, OutputStream out) throws IOException;
+
+   /**
+    * Unmarshall an object from the {@link InputStream}
+    *
+    * @param in the {@link InputStream} to unmarshall an object from
+    * @return the unmarshalled object instance
+    * @throws IOException if unmarshalling cannot complete due to some I/O error
+    * @throws ClassNotFoundException if the class of the object trying to unmarshall is unknown
```
anistor: 'unknown' and 'not found' are very different.

commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
```diff
@@ -125,4 +126,30 @@ public WrappedByteArray readObject(ObjectInput input) throws IOException {
...
+
+      @Override
+      public String getTypeName() {
+         return "persistence.WrappedByteArray";
+      }
+
+      @Override
+      public WrappedByteArray readFrom(ProtoStreamReader reader) throws IOException {
+         byte[] bytes = reader.readBytes("wrappedBytes");
+         int hashcode = reader.readInt("hashcode");
```
anistor: why marshall haschode? cant we recompute it? or let it be recomputed lazily

core/src/main/java/org/infinispan/atomic/impl/AtomicKeySetImpl.java
```diff
@@ -554,4 +556,54 @@ public Externalizable readObject(ObjectInput input) throws IOException, ClassNot
...
+      }
+
+      @Override
+      public void writeTo(ProtoStreamWriter writer, AtomicKeySetImpl atomicKeySet) throws IOException {
+         writer.writeString("cachename", atomicKeySet.cacheName.toString());
+         writer.writeObject("group", new UserObject(atomicKeySet.group), UserObject.class);
+      }
+
+      @Override
+      public Class<? extends AtomicKeySetImpl> getJavaClass() {
```
anistor: getJavaClass() should declare the exact class. no need for wildcards. not just here, everywhere.

core/src/main/java/org/infinispan/container/versioning/SimpleClusteredVersion.java
```diff
@@ -108,4 +109,27 @@ public Integer getId() {
          return Collections.<Class<? extends SimpleClusteredVersion>>singleton(SimpleClusteredVersion.class);
       }
    }
+
+   public static class Marshaller implements MessageMarshaller<SimpleClusteredVersion> {
+      @Override
+      public SimpleClusteredVersion readFrom(ProtoStreamReader reader) throws IOException {
+         return new SimpleClusteredVersion(reader.readInt("topology"), reader.readLong("version"));
```
anistor: I'd rather not read all of them on the same line

core/src/main/resources/persistence.proto
```diff
@@ -0,0 +1,71 @@
...
+    required bytes wrappedObject = 1;
+}
+
+message WrappedByteArray {
+    required bytes wrappedBytes = 1;
+    optional int32 hashcode = 2;
+}
+
+message KeyValuePair {
+    optional bytes key = 1;
```
anistor: are key and value really optional ?

core/src/main/resources/persistence.proto
```diff
@@ -0,0 +1,71 @@
+package persistence;
+
+message UserObject {
+    required bytes wrappedObject = 1;
+}
+
+message WrappedByteArray {
+    required bytes wrappedBytes = 1;
+    optional int32 hashcode = 2;
```
anistor: don't think we want to persist hashcode

persistence/jdbc/src/test/java/org/infinispan/persistence/jdbc/stringbased/JdbcStringBasedStoreAltMapperTest.java
```diff
@@ -43,7 +43,7 @@
    protected TableManager tableManager;
    protected static final Person MIRCEA = new Person("Mircea", "Markus", 28);
```
anistor: They never grow up :)

commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
wburns: This was added a while back to optimize the case with very large key or values.

remote-query/remote-query-client/pom.xml
```diff
@@ -19,11 +19,6 @@
          <artifactId>infinispan-commons</artifactId>
       </dependency>
 
-      <dependency>
-         <groupId>org.infinispan.protostream</groupId>
```
anistor: yes, we get this dependency anyway transitively from commons. but why remove it? it is also useful to document a fact

remote-query/remote-query-client/src/test/java/org/infinispan/query/remote/client/BaseProtoStreamMarshallerTest.java
```diff
@@ -5,6 +5,7 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import org.infinispan.commons.marshall.protostream.BaseProtoStreamMarshaller;
```
anistor: the test is in hr client and the implementation in commons :)

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/CompatibilityProtoStreamMarshaller.java
```diff
@@ -3,7 +3,7 @@
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.protostream.SerializationContext;
-import org.infinispan.query.remote.client.BaseProtoStreamMarshaller;
```
anistor: woud it be possible not to move BaseProtoStreamMarshaller to commons ?

commons/src/main/java/org/infinispan/commons/marshall/WrappedByteArray.java
ryanemerson: I guess we should probably remove it from this marshaller then as I imagine it will not make a significant difference when loading from persistence.

core/src/main/java/org/infinispan/atomic/impl/AtomicKeySetImpl.java
ryanemerson: Good point. This was just pulled in by intellij's "implement methods" as `BaseMarshaller::getJavaClass` is `Class<? extends T> getJavaClass();`.

I can update though.

core/src/main/resources/persistence.proto
ryanemerson: There are many cases in the testsuite where a null key/value is used, so I went with this. But thinking about it the fields should be required here and a an empty `byte[]` passed in the aforementioned cases.

remote-query/remote-query-client/src/test/java/org/infinispan/query/remote/client/BaseProtoStreamMarshallerTest.java
ryanemerson: Good spot!

core/src/main/resources/persistence.proto
anistor: well, if these are truly nullable, then let's keep them optional without empty byte[] tricks

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/CompatibilityProtoStreamMarshaller.java
ryanemerson: Yeah good point. I will move this and avoid adding the dependency to `infinispan-commons` as well. Originally I though that there would be more complex changes to `BaseProtoStreamMarshaller` when I thought implementing `StreamingMarshaller` was required.

commons/src/main/java/org/infinispan/commons/marshall/StreamingMarshaller.java
```diff
@@ -124,4 +128,92 @@
 
    void start();
 
+   static StreamingMarshaller from(StreamAwareMarshaller marshaller) {
+      return new StreamingMarshaller() {
+         @Override
+         public ObjectOutput startObjectOutput(OutputStream os, boolean isReentrant, int estimatedSize) throws IOException {
+            throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
ryanemerson: @wburns In this case the alternative to throwing an exception is to return null, I'm not sure that is any better as it will result in NPE from existing code.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -146,6 +147,8 @@ public void start() {
       enabled = configuration.persistence().usingStores();
       if (!enabled)
          return;
+      // Needed to pass ThreeNodeGlobalStateRestartTest but do we want this?
```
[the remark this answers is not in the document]
ryanemerson: This was a leftover from debugging, removed.

commons/src/main/java/org/infinispan/commons/marshall/StreamingMarshaller.java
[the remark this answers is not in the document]
wburns: Weird, I must have been looking at a different diff or something. Since this requires invoking the new `from` method you can disregard this comment.

core/src/main/java/org/infinispan/persistence/marshallers/MapMarshaller.java
```diff
@@ -0,0 +1,89 @@
...
+   @Override
+   public void writeTo(ProtoStreamWriter writer, Map map) throws IOException {
+      if (map.isEmpty()) {
+         writer.writeString("keyType", "");
+      } else {
+         try {
+            List<byte[]> keys = new ArrayList<>(map.size());
+            List<byte[]> values = new ArrayList<>(map.size());
+            for (Map.Entry entry : (Set<Map.Entry>) map.entrySet()) {
+               keys.add(marshaller.objectToByteBuffer(entry.getKey()));
```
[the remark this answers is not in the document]
ryanemerson: I have updated the `persistence.Map` message and MapMarshaller.java to make things a bit nicer, however it's not currently possible to serialize a stream of values to a repeatbale field with protostream. @anistor is aware of this and will allow for this in a future protostream release.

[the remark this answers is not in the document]
anistor: Please create a jira for IPROTO.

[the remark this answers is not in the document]
ryanemerson: https://issues.jboss.org/browse/IPROTO-74

core/src/main/java/org/infinispan/marshall/core/UserExternalizerObjectTable.java
```diff
@@ -0,0 +1,57 @@
...
+      return writer != null ? writer::accept : null;
+   }
+
+   @Override
+   public Object readObject(Unmarshaller unmarshaller) throws IOException, ClassNotFoundException {
+      AdvancedExternalizer<Object> ext = findExternalizerIn(unmarshaller);
+      return ext.readObject(unmarshaller);
+   }
+
+   BiConsumer<ObjectOutput, Object> findWriter(Object object) {
```
wburns: I say lets just return the `AdvancedExternalizer` instead. Then we only need 1 lambda since `getObjectWriter` can change to `return ext != null ? (m, obj) -> writeExternalClean(obj, ext, m);` or something like that.

core/src/main/java/org/infinispan/persistence/keymappers/MarshallingTwoWayKey2StringMapper.java
```diff
@@ -10,5 +11,15 @@
  * @since 5.2
  */
 public interface MarshallingTwoWayKey2StringMapper extends TwoWayKey2StringMapper {
-   void setMarshaller(StreamingMarshaller marshaller);
+
+   /**
+    * @deprecated since 10.0, use {@link #setMarshaller(Marshaller)} instead
+    */
+   default void setMarshaller(StreamingMarshaller marshaller) {
```
wburns: We should probably add the @Deprecated annotation as well,just to be safe.

core/src/main/java/org/infinispan/persistence/marshallers/EntryVersionMarshaller.java
```diff
@@ -0,0 +1,34 @@
...
+public class EntryVersionMarshaller implements MessageMarshaller<EntryVersion> {
+   @Override
+   public EntryVersion readFrom(ProtoStreamReader reader) throws IOException {
+      EntryVersion version = reader.readObject("numeric", NumericVersion.class);
+      return version != null ? version : reader.readObject("clustered", SimpleClusteredVersion.class);
+   }
+
+   @Override
+   public void writeTo(ProtoStreamWriter writer, EntryVersion version) throws IOException {
+      if (version instanceof NumericVersion)
```
wburns: Question: This seems overly complex. Is there no way to just write Object where you don't have to give it the type? For example what if this had 10 different implementations, would we have to do an instance of check and define a proto field for each? This seems very cumbersome.

core/src/main/java/org/infinispan/persistence/marshallers/MapMarshaller.java
```diff
@@ -0,0 +1,66 @@
...
+      if (!map.isEmpty()) {
+         List<KeyValuePair> kvp = ((Set<Map.Entry>) map.entrySet()).stream()
+               .map(e -> new KeyValuePair<>(e.getKey(), e.getValue()))
+               .collect(Collectors.toList());
+
+         writer.writeCollection("entry", kvp, KeyValuePair.class);
+      }
+
+      if (!(map instanceof HashMap)) {
+         writer.writeString("implementation", map.getClass().getName());
```
wburns: Shouldn't this come first to mirror the read?

core/src/main/java/org/infinispan/persistence/marshallers/MapMarshaller.java
```diff
@@ -0,0 +1,66 @@
...
+            throw new IllegalStateException(e);
+         }
+      };
+   }
+
+   @Override
+   public void writeTo(ProtoStreamWriter writer, Map map) throws IOException {
+      if (!map.isEmpty()) {
+         List<KeyValuePair> kvp = ((Set<Map.Entry>) map.entrySet()).stream()
+               .map(e -> new KeyValuePair<>(e.getKey(), e.getValue()))
```
wburns: Man this is looking even worse than before.... 👎 Wish there was a nice way to do this.

core/src/main/java/org/infinispan/persistence/marshallers/PersistenceMarshaller.java
```diff
@@ -0,0 +1,184 @@
...
+   }
+
+   @Override
+   public boolean isMarshallable(Object o) {
+      return !isBlacklisted(o) && (isInternalClass(o) || isUserMarshallable(o));
+   }
+
+   private boolean isBlacklisted(Object o) {
+      Class clazz = o.getClass();
+      if (clazz.isArray())
```
wburns: We check arrays, should we check Collections as well?

core/src/main/java/org/infinispan/persistence/marshallers/PersistenceMarshaller.java
```diff
@@ -0,0 +1,184 @@
...
+
+   @Override
+   protected ByteBuffer objectToBuffer(Object o, int estimatedSize) throws IOException {
+      try {
+         byte[] bytes = ProtobufUtil.toWrappedByteArray(getSerializationContext(), wrap(o));
+         return new ByteBufferImpl(bytes, 0, bytes.length);
+      } catch (java.io.NotSerializableException nse) {
+         // TODO do we still want this? I think it assumes too much about the configured user marshaller as it may not
+         // even support the Serializable interface. Replace with generic MarshallingException?
+         if (log.isDebugEnabled()) log.debug("Object is not serializable", nse);
```
wburns: At least log the object?

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -472,7 +471,7 @@ void storeEntry(InvocationContext ctx, Object key, FlagAffectedCommand command)
 
    MarshalledEntry marshalledEntry(InvocationContext ctx, Object key) {
       InternalCacheValue sv = entryFactory.getValueFromCtx(key, ctx);
-      return sv != null ? new MarshalledEntryImpl(key, sv.getValue(), internalMetadata(sv), marshaller) : null;
+      return sv != null ? marshalledEntryFactory.newMarshalledEntry(key, sv.getValue(), internalMetadata(sv)): null;
```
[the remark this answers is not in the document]
ryanemerson: It will involve changing a fair bit of test code, but I think it's worth it.

core/src/main/java/org/infinispan/persistence/marshallers/PersistenceMarshaller.java
ryanemerson: The details are slightly hazy at this point as it was a while since I encountered the issue, but this check is required for how lambda's are sometimes passed through to the marshaller. I don't think it's necessary todo the same for collections as I don't know of a scenario where a collection of `SerializedLambda` etc will be passed.

core/src/main/java/org/infinispan/persistence/marshallers/EntryVersionMarshaller.java
ryanemerson: I don't think so @wburns as protobuf doesn't support inheritance. @anistor Any thoughts?

core/src/main/java/org/infinispan/persistence/marshallers/MapMarshaller.java
ryanemerson: @anistor When are you aiming to release 4.3.0? I'm mainly thinking of https://issues.jboss.org/browse/IPROTO-74

ryanemerson:
@anistor @wburns The last run didn't have any related failures, if you could take another look it would be much appreciated.

@gustavonalle You will probably be interested in this PR as well.  I have moved the `Protostream*Transcoders` to core and I have also modified `remote-query-server` so that it reuses the PersistenceMarshaller's SerializationContext. Could take a look at the encoding/transcoding side of the PR to make sure I haven't done anything stupid :smile:

gustavonalle:
Will do @ryanemerson , thanks for the heads-up

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerImpl.java
```diff
@@ -55,32 +56,40 @@
...
+      persistenceMarshaller.setProtobufConfiguration(cfgBuilder.build());
+
       internalCacheRegistry.registerInternalCache(PROTOBUF_METADATA_CACHE_NAME,
             getProtobufMetadataCacheConfig().build(),
             EnumSet.of(Flag.USER, Flag.PROTECTED, Flag.PERSISTENT));
    }
 
+   @Start
+   public void start() {
+      this.serCtx = persistenceMarshaller.getSerializationContext();
```
gustavonalle: The ProtobufMetadataManager is used to manage user provided protobuf files (backed by the internal __protobuf_metadata cache), I don't think it should reuse the same serCtx as the persistence marshaller.
Also, it seems weird having the persistence marshaller in the remote query server :)

core/src/main/java/org/infinispan/dataconversion/PersistenceMarshallerEncoder.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.infinispan.commons.dataconversion.EncoderIds;
+import org.infinispan.commons.dataconversion.MarshallerEncoder;
+import org.infinispan.commons.dataconversion.MediaType;
+import org.infinispan.marshall.persistence.PersistenceMarshaller;
+
+/**
+ * @since 10.0
+ */
+public class PersistenceMarshallerEncoder extends MarshallerEncoder {
```
gustavonalle: Why is this needed? Generally speaking, persistence and encoding/transcoding live in different layers: while transcoding is between the user and the cache API (it does not happen throughout the interceptors), persistence is between the cache and the stores.

core/src/main/java/org/infinispan/dataconversion/ProtostreamJsonTranscoder.java
```diff
@@ -1,4 +1,4 @@
-package org.infinispan.query.remote.impl.dataconversion;
+package org.infinispan.dataconversion;
```
gustavonalle: Why do the transcoders need to move to core? Nothing against it, but It seems they are not used in the persistence layer.

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
```diff
@@ -33,6 +35,8 @@
 import org.infinispan.marshall.core.ClassToExternalizerMap.IdToExternalizerMap;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
+import org.jboss.marshalling.ObjectTable;
```
galderz: This is not right the right way. GlobalMarshaller is independent from underlying marshalling library. It should not be depending on JBoss Marshalling itself. Any JBoss Marshalling code should be limited to ExternalJBossMarshaller.

core/src/main/java/org/infinispan/persistence/spi/MarshalledEntry.java
```diff
@@ -0,0 +1,47 @@
...
+import org.infinispan.metadata.InternalMetadata;
+
+/**
+ * Defines an externally persisted entry. External stores that keep the data in serialised form should return an
+ * MarshalledEntry that contains the data in binary form (ByteBuffer) and unmarshall it lazily when
+ * getKey/Value/Metadata are invoked. This approach avoids unnecessary object (de)serialization e.g
+ * when the entries are fetched from the external store for the sole purpose of being sent over the wire to
+ * another requestor node.
+ *
+ * @author Mircea Markus
```
galderz: Mircea still working for us?

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
ryanemerson: Ok, I can change this. FYI the GlobalMarshaller as it currently is implemented will be refactored heavily with https://issues.jboss.org/browse/ISPN-9622, so ^ code should be temporary.

gustavonalle:
> Protostream does not support generation of classes from .proto

AFAICT, you can generate the marshallers and the protobuf file given an annotated pojo, using the  ```ProtoSchemaBuilder```

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerImpl.java
ryanemerson: The reason that the serialization contexts is shared was to allow `ProtostreamSerializationContextInitializer` implementations to register protofiles with both the query and persistence context simultaneously. 

Originally I had created a implementation of the SerializationContext interface to wraps the two contexts and ensures that any registering/unregistering of marshallers happens on both contexts. However, the problem is that the protostream implementation casts to `SerializationContextImpl` in a few places, so this approach did not work. 

I agree that it's strange that the persistence marshaller is in the remote query server :see_no_evil:. So I guess the question is how best to abstract the registration of protobuf files with multiple contexts? I could create `InfinispanSerializationContext` which delegates to the two `SerializationContext` instances and expose this via `ProtostreamSerializationContextInitializer`? We would have to deprecate the previous `init` method though.

ryanemerson:
@gustavonalle Correct about the generation, but we decided that this was not suitable as it does not support final or private variables and requires a no-args constructor. This would be problematic when we refactor the GlobalMarshaller to be protostream based as most of the internal classes have fields that are private/package-private, are final and/or don't currently have no-args constructors. Furthermore, relying on annotated Pojos for our definitive schema IMO is harder to maintain then having a schema file with compatible Marshaller implementations, as the specification becomes distributed across core.

remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerImpl.java
gustavonalle: My understanding was that ```ProtostreamSerializationContextInitializer``` is a user facing construct to allow registering extra user marshallers, and the persistence serCtx is a private context that works only with Infinispan internal classes and treats user data as opaque byte[]. This private content is known at "boot" time, and no extra proto/classes are needed at runtime (thus no registration). 

Is there a use case of exposing the persistence ctx to the user, so that it can add, remove and change marshallers/protofiles?

ryanemerson: > Is there a use case of exposing the persistence ctx to the user, so that it can add, remove and change marshallers/protofiles?

Good point and I don't think there is. I was getting ahead of myself/confused by the idea of having a protobuf based user marshaller, but that's for embedded mode anyway.

core/src/main/java/org/infinispan/dataconversion/PersistenceMarshallerEncoder.java
wburns: Who would use this? I don't see this in use other than registering it in the encoder registry.

core/src/main/java/org/infinispan/marshall/core/GlobalMarshaller.java
galderz: Fair enough. Just make sure that when you refactor, 3rd party marshalling code is separated away.

galderz:
> @galderz If the marshaller implementation does not match up with the protobuf schema then an exception will be thrown.

@ryanemerson ^ Is that already in place?

galderz:
> Protostream does not support generation of classes from .proto files like the google libraries, hence why we implement our own marshallers. 

Yeah... but is unnecessary duplication :\

galderz:
> The code generated by the google libraries contains lots unnecessary metadata apparently, so is not really suitable for Infinispan, but @anistor is the expert to talk about that.

https://github.com/square/wire#why-wire

galderz:
Duplication we can (kinda) live with it, but eventually should go away. What is more important here is not let marshaller changes happen inadvertently.

ryanemerson:
> > @galderz If the marshaller implementation does not match up with the protobuf schema then an exception will be thrown.
> 
> @ryanemerson ^ Is that already in place?

Yes, but the exception is thrown by protostream at runtime. So it shouldn't get past our testsuite ...

ryanemerson:
> > The code generated by the google libraries contains lots unnecessary metadata apparently, so is not really suitable for Infinispan, but @anistor is the expert to talk about that.
> 
> https://github.com/square/wire#why-wire

@galderz That looks very interesting. Being able to generate our code purely from `.proto` files is more appealing to me if we can do it in a performant way. @anistor have you looked into this implementation before? Any thoughts?

galderz:
> Yes, but the exception is thrown by protostream at runtime. So it shouldn't get past our testsuite ...

What's the cost of doing that? Is it done once on startup?

core/src/main/java/org/infinispan/marshall/persistence/impl/EntryVersionMarshaller.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.infinispan.container.versioning.EntryVersion;
+import org.infinispan.container.versioning.NumericVersion;
+import org.infinispan.container.versioning.SimpleClusteredVersion;
+import org.infinispan.protostream.MessageMarshaller;
+
+final class EntryVersionMarshaller implements MessageMarshaller<EntryVersion> {
+   @Override
+   public EntryVersion readFrom(ProtoStreamReader reader) throws IOException {
+      EntryVersion version = reader.readObject("numeric", NumericVersion.class);
```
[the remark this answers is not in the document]
ryanemerson: adding to Wills question ... if ^ is a performance concern, can we get around it by setting a default value of null on the `numeric` field in the `*.proto` definition @anistor?

ryanemerson:
@galderz It will only be thrown when an attempt is made to read/write the message defined in the `*.proto` file

anistor:
> @galderz That looks very interesting. Being able to generate our code purely from `.proto` files is more appealing to me if we can do it in a performant way. @anistor have you looked into this implementation before? Any thoughts?

I'm not very much into the idea of generating source code.

ryanemerson:
> I'm not very much into the idea of generating source code.

Care to elaborate why? :smile: 

Personally I prefer the idea of generating code from a single `*.proto` file, instead of utilising Pojos, because it means that we can have all of the message definitions in a single place[1] which aids maintenance and ensuring compatibility. Having annotated POJOs distributed across core would be much harder to maintain IMO. 

[1] Actually two, because we would have a `.proto` file for the global marshaller and the persistence marshaller to make sure the two are independent.

anistor:
> I'm not very much into the idea of generating source code.

I'm leaning towards a java-centric approach and generate the *.proto from the (annotated) java rather than doing it the other way around. Or just check that the annotated java conforms to an existing manually provided schema, if such a schema exists.

ryanemerson:
@gustavonalle I fixed the issues with the transcoders/encoders:
- Encoder impl removed
- Moved the `Protostream*Transcoder`s to `server-core` so that they can be utilised in the `persistence/remote` tests

galderz:
> I'm not very much into the idea of generating source code.

I like it more and more actually. It allows us to automate and provide consistency. Things like [JavaPoet](https://github.com/square/javapoet) make it easier than ever before to generate readable java code.

commons/src/main/java/org/infinispan/commons/dataconversion/EncoderIds.java
```diff
@@ -15,4 +15,5 @@
    short MEMCACHED_COMPAT = 8;
    short JAVA_COMPAT = 9;
    short UTF8_COMPAT = 10;
+   short PERSISTENCE_MARSHALLER = 11;
```
gustavonalle: This can be removed

ryanemerson: Fixed :+1:

ryanemerson:
run performance tests please
library
cs

core/src/test/java/org/infinispan/util/ControlledConsistentHashFactory.java
```diff
@@ -145,4 +152,51 @@ public void setOwnerIndexes(int[][] segmentOwners) {
          super.setOwnerIndexes(segmentOwners);
       }
    }
+
+   public static class Externalizer implements AdvancedExternalizer<ControlledConsistentHashFactory> {
```
danberindei: Shouldn't this be in the next PR?

ryanemerson:
@danberindei I'll issue a separate PR for ISPN-9693 and then rebase on top of that once it's in.

core/src/main/java/org/infinispan/util/KeyValuePair.java
```diff
@@ -84,4 +86,48 @@ public int hashCode() {
...
+         this.marshaller = marshaller;
+      }
+
+      @Override
+      public KeyValuePair readFrom(ProtoStreamReader reader) throws IOException {
+         byte[] keyBytes = reader.readBytes("key");
+         byte[] valueBytes = reader.readBytes("value");
+         try {
+            Object key = keyBytes == null ? null : marshaller.objectFromByteBuffer(keyBytes);
+            Object value = valueBytes == null ? null : marshaller.objectFromByteBuffer(valueBytes);
```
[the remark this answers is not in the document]
ryanemerson: I'm not against that in theory, although I will need to figure out how many uses there actually is. The only issue I can think of is if users have used `KeyValuePair` as part of custom stores, but I suppose it's not in the SPI so they can use the user marshaller to serialize.

core/src/test/java/org/infinispan/util/ControlledConsistentHashFactory.java
ryanemerson: No we must add this Externalizer here, as the user marshaller needs to know how to marshall the `List<Address>` elements. Previously, this was not an issue as the GlobalMarshaller would utilise the configured `JGroupsAddress.Externalizer`. As the user marshaller is designed to only load the externalizers configured as part of the SerializationConfiguration, it's necessary to add a externalizer in the `NumOwnersNodeCrashInSequenceTest#createCacheManagers`. Looking at this again though, it's overkill to create an Externalizer for `ControlledConsistentHashFactory`, when it's sufficient to simply add the `JGroupsAddress.Externalizer`.

ryanemerson:
Closing in favour of https://github.com/infinispan/infinispan/pull/6906

PR #6241, opened 2018-09-06
ISPN-9484 Near cache invalidation should ignore created events and

wburns:
should not contain values

* All listeners without filter or converter don't use value
* Near cache ignores create remote events

https://issues.jboss.org/browse/ISPN-9484

### DIFF
--- a/core/src/main/java/org/infinispan/notifications/cachelistener/filter/KeyValueFilterConverterAsCacheEventFilterConverter.java
+++ b/core/src/main/java/org/infinispan/notifications/cachelistener/filter/KeyValueFilterConverterAsCacheEventFilterConverter.java
@@ -1,0 +1,71 @@
+package org.infinispan.notifications.cachelistener.filter;
+
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.util.Collections;
+import java.util.Set;
+
+import org.infinispan.commons.marshall.AdvancedExternalizer;
+import org.infinispan.commons.marshall.Ids;
+import org.infinispan.factories.ComponentRegistry;
+import org.infinispan.factories.annotations.Inject;
+import org.infinispan.filter.KeyValueFilterConverter;
+import org.infinispan.metadata.Metadata;
+
+/**
+ * {@link CacheEventFilterConverter} that uses an underlying {@link KeyValueFilterConverter} to do the conversion and
+ * filtering. The new value and metadata are used as arguments to the underlying filter converter as it doesn't take
+ * both new and old.
+ * @author wburns
+ * @since 9.4
+ */
+public class KeyValueFilterConverterAsCacheEventFilterConverter<K, V, C> implements CacheEventFilterConverter<K, V, C> {
+   private final KeyValueFilterConverter<K, V, C> keyValueFilterConverter;
+
+   public KeyValueFilterConverterAsCacheEventFilterConverter(KeyValueFilterConverter<K, V, C> keyValueFilterConverter) {
+      this.keyValueFilterConverter = keyValueFilterConverter;
+   }
+
+   @Override
+   public C filterAndConvert(K key, V oldValue, Metadata oldMetadata, V newValue, Metadata newMetadata, EventType eventType) {
+      return keyValueFilterConverter.convert(key, newValue, newMetadata);
+   }
+
+   @Override
+   public C convert(K key, V oldValue, Metadata oldMetadata, V newValue, Metadata newMetadata, EventType eventType) {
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/infinispan/notifications/cachelistener/filter/KeyValueFilterConverterAsCacheEventFilterConverter.java
```diff
@@ -0,0 +1,60 @@
...
+   public boolean accept(K key, V oldValue, Metadata oldMetadata, V newValue, Metadata newMetadata, EventType eventType) {
+      return keyValueFilterConverter.accept(key, newValue, newMetadata);
+   }
+
+   @Inject
+   protected void injectDependencies(ComponentRegistry cr) {
+      cr.wireDependencies(keyValueFilterConverter);
+   }
+
+   public static class KVFCACEFCExternalizer implements Externalizer<KeyValueFilterConverterAsCacheEventFilterConverter> {
```
danberindei: Could be named `Externalizer`, and it should extend `AdvancedExternalizer` to avoid writing the class name to the stream.

tristantarrant:
I'd also add a performance note in the docs about skipping values for custom  converters  in server

wburns:
@tristantarrant not sure what you mean here. We have http://infinispan.org/docs/dev/user_guide/user_guide.html#filter_and_custom_events already. And these changes only affect regular events with no filter/converter applied.

core/src/main/java/org/infinispan/notifications/cachelistener/filter/KeyValueFilterConverterAsCacheEventFilterConverter.java
wburns: Sure I was adding one, but then thought about it more that this is a very minor used operation (scales only with number of listeners - which should be small). But I can add in an ID for it since it seems we want to :)

wburns: And I can name it `Externalizer` only when it extends `AdvancedExternalizer` (unless I added the entire package name). Just FYI.

wburns:
Updated to add in AdvancedExternalizer.

tristantarrant:
The docs suffer from the same issue shared by a lot of our documentation:
they lack a practical example and suggestions/tips on ensuring good
performance

wburns:
This is one of the few that actually has an example of optionally ignoring the value payload for a configurable key. And with this PR people shouldn't need to create a custom event for performance any longer, it is just for what it should be, custom events.

wburns:
Looking

wburns:
I haven't figured out yet why, but somehow the last put is creating a CREATE event in addition to the second to last one. The locking should prevent this from occurring. I have to look further.

wburns:
Bleh, didn't pay attention to the put arguments. The problem was that it was inserting a value with 3 millisecond lifespan, which can timeout and expire in some cases, thus causing 2 create events to occur.

However, we don't send an expiration event though, which we would need in this case. Let me look at that.

wburns:
Currently we don't generate an expired event when a write operation occurs. This is mostly because remove expired would cause a looping problem. I would have to think about how to handle that. In this case we may not be able to ignore Created events yet.

wburns:
So the solutions I see atm are:

1. generate additional remove expired on write of expired entry
2. return expired entries on write
3. punt and just listen to create events for near cache

The solution for 1 is the best for consistency, but loses on performance. It also will require possibly quite a bit more changes and testing to verify it works and doesn't break stuff.

The solution for 2 changes consistency guarantee on write operations with expired entries which some tests rely on atm.

I personally am not a big fan of any of these fixes. @tristantarrant WDYT?

danberindei:
IMO solution 1 wouldn't have worse performance, because it would only generate the expired event if the entry was expired. So yeah, if the application uses a lifespan smaller than 2 * the purge interval there could be some extra expiration events, but that application wouldn't be using the cache very efficiently anyway.

wburns:
Well I meant it would have worse performance when it occurs. But that should not occur often. Also it is more about access pattern of the app itself. I am finding that someone may use `putIfAbsent` all the time instead of `get` and `put` which would run into this issue quite easily.

To be honest I really want solution 1, I am worried about timing for this type of change though, since we only have just over 2 days until the Final build. WDYT @tristantarrant ?

wburns:
I created https://issues.jboss.org/browse/ISPN-9511 for the test failure. This PR should be able to be integrated now, by itself.

tristantarrant:
Merged. Thanks

PR #6242, opened 2018-09-10
ISPN-9496 Some xsite tests hang during teardown

danberindei:
https://issues.jboss.org/browse/ISPN-9496

Enable FD only in xsite tests


pruivo:
it seems to work.
Integrated! thanks @danberindei

PR #6244, opened 2018-09-10
ISPN-9332 REPL local iteration optimization cannot be used when store

wburns:
has write behind

* Also fixed async store segment filtering
* Also fixed clear losing later updates

https://issues.jboss.org/browse/ISPN-9332


wburns:
Whooo, green again. Never thought I would see the day.

tristantarrant:
Merged, thanks

PR #6245, opened 2018-09-11
ISPN-9500 ConcurrentSmallIntSet.clear() does not always set size to 0

danberindei:
https://issues.jboss.org/browse/ISPN-9500

Fix the handling of negative values in clear() and several other
methods:
* toIntArray()
* forEach(IntConsumer)
* removeIf(IntPredicate)


wburns:
Good catch, not sure why I didn't think of this when working on it. I even used unsigned bit shift in a few other places. May have been one of those things where you implement something one way and then mess it up when changing it to another way. Anyways, nice! 👍

wburns:
Also this should be cherry-picked to 9.3.x, which should be conflict free. Putting as a reminder.

danberindei:
Thanks for remembering 9.3.x :)

tristantarrant:
CI is green

wburns:

wburns:
Also cherry-picked to 9.3.x

PR #6247, opened 2018-09-11
ISPN-9487 Upgrade to Wildfly 14.0.1

tristantarrant:
https://issues.jboss.org/browse/ISPN-9487

This PR incorporates @ryanemerson 's and also upgrades the server configuration files to use the jboss:domain:8.0 schema

### DIFF
--- a/feature-pack/wf-modules/src/main/resources/modules/org/infinispan/hibernate-search/directory-provider/for-hibernatesearch-5.10/module.xml
+++ b/feature-pack/wf-modules/src/main/resources/modules/org/infinispan/hibernate-search/directory-provider/for-hibernatesearch-5.10/module.xml
@@ -23,8 +23,8 @@ and obviously needs to avoid dependencies to different versions of Hibernate Sea
     </resources>
 
     <dependencies>
-        <module name="org.apache.lucene" slot="5.5" />
-        <module name="org.hibernate.search.engine" slot="@hibernate.search.module.slot@" />
+        <module name="org.apache.lucene"/>
+        <module name="org.hibernate.search.engine"/>
         <module name="javax.api"/>
         <module name="org.infinispan.commons" slot="@infinispan.module.slot@" export="true"/>
         <module name="org.infinispan.core" slot="@infinispan.module.slot@" export="true" services="export" />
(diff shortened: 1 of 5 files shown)


tristantarrant:
1 unrelated failure.

feature-pack/wf-modules/src/main/resources/modules/org/infinispan/hibernate-search/directory-provider/for-hibernatesearch-5.10/module.xml
```diff
@@ -23,8 +23,8 @@ and obviously needs to avoid dependencies to different versions of Hibernate Sea
     </resources>
 
     <dependencies>
-        <module name="org.apache.lucene" slot="5.5" />
-        <module name="org.hibernate.search.engine" slot="@hibernate.search.module.slot@" />
+        <module name="org.apache.lucene"/>
```
gustavonalle: I am surprised this was removed and none of the tests failed...Does Wildfly 14 include lucene 5.5?

pom.xml
```diff
@@ -234,7 +234,7 @@
       <version.jgroups>4.0.13.Final</version.jgroups>
       <version.jta>1.0.1.Final</version.jta>
       <version.lucene>5.5.5</version.lucene>
-      <version.lucene.module.slot>5.5</version.lucene.module.slot>
+      <version.lucene.module.slot>5.5.5</version.lucene.module.slot>
```
gustavonalle: This was changed from 5.5.5 to 5.5 in https://issues.jboss.org/browse/ISPN-9410 because of linkage error. Was it causing issues with Widlfly 14?

feature-pack/wf-modules/src/main/resources/modules/org/infinispan/hibernate-search/directory-provider/for-hibernatesearch-5.10/module.xml
gustavonalle: Also, this module is to be linked against Wildfly internal hibernate search engine (as of WF 13, it's 5.10), and should not use the main lucene and hsearch slots, but specific versions hard-coded as search 5.10

tristantarrant: Yes: WildFly has made 5.10 the default (main slot) and included Lucene 5.5. There is also a 5.10 slot, but instead of being an alias, it contains a copy of the jars, thus causing LinkageErrors. That is why we have to use the main slot here.

PR #6249, opened 2018-09-11
ISPN-9505 Off-heap data container puts all the entries in a single

wburns:
bucket

* Make sure buckets are address count not scaled to CPU

https://issues.jboss.org/browse/ISPN-9505

### DIFF
--- a/core/src/main/java/org/infinispan/container/offheap/ContiguousOffsetCalculator.java
+++ b/core/src/main/java/org/infinispan/container/offheap/ContiguousOffsetCalculator.java
@@ -1,5 +1,8 @@
 package org.infinispan.container.offheap;
 
+import org.infinispan.commons.hash.Hash;
+import org.infinispan.commons.hash.MurmurHash3;
+
 /**
  * OffsetCalculator that provides an offset where it is calculated by dividing the positive int values into contiguous
  * blocks. As an example if the int hash spectrum was 8 and the numBlocks provided was 2, then the blocks would contain
@@ -15,12 +18,16 @@ class ContiguousOffsetCalculator implements OffsetCalculator {
 
    private final int offsetDivisor;
 
+   // We use murmur hash as simple spread doesn't work well with contiguous blocks (similar hash codes end up in the
+   // same block otherwise)
+   private final static Hash hash = MurmurHash3.getInstance();
+
    /**
     * Creates a new contiguous offset calculator using the provided offset segment the entire positive int spectrum
     * up to {@link Integer#MAX_VALUE}.
     * @param numBlocks how many blocks there will be - note this <b>MUST</b> be a power of two
     */
-   public ContiguousOffsetCalculator(int numBlocks) {
+   ContiguousOffsetCalculator(int numBlocks) {
       // We need a positive offset count
       // And it has to be a power of 2
       if (numBlocks <= 0 && (numBlocks & (numBlocks - 1)) == 0) {
@@ -36,6 +43,7 @@ class ContiguousOffsetCalculator implements OffsetCalculator {
    }
 
    private static int spread(int h) {
-      return (h ^ (h >>> 16)) & HASH_BITS;
+      // Spread using murmur hash then ensure it is positive before finding block to use
+      return hash.hash(h) & HASH_BITS;
    }
 }
--- a/core/src/main/java/org/infinispan/factories/DataContainerFactory.java
+++ b/core/src/main/java/org/infinispan/factories/DataContainerFactory.java
@@ -58,8 +58,8 @@ public class DataContainerFactory extends AbstractNamedCacheComponentFactory imp
                if (clusteringConfiguration.cacheMode().needsStateTransfer()) {
                   int segments = clusteringConfiguration.hash().numSegments();
                   Supplier mapSupplier = () -> {
-                     OffHeapEntryFactory entryFactory = componentRegistry.getComponent(OffHeapEntryFactory.class);
-                     OffHeapMemoryAllocator memoryAllocator = componentRegistry.getComponent(OffHeapMemoryAllocator.class);
+                     OffHeapEntryFactory entryFactory = componentRegistry.getOrCreateComponent(OffHeapEntryFactory.class);
+                     OffHeapMemoryAllocator memoryAllocator = componentRegistry.getOrCreateComponent(OffHeapMemoryAllocator.class);
                      // TODO: find better way to handle size here or is it okay? internally it will round to next power of 2
                      OffHeapConcurrentMap offHeapMap = new OffHeapConcurrentMap(addressCount / segments, memoryAllocator, entryFactory, null);
                      offHeapMap.start();
--- a/core/src/main/java/org/infinispan/container/offheap/OffHeapConcurrentMap.java
+++ b/core/src/main/java/org/infinispan/container/offheap/OffHeapConcurrentMap.java
@@ -134,7 +134,11 @@ public class OffHeapConcurrentMap implements ConcurrentMap<WrappedBytes, Interna
       lockCount = Util.findNextHighestPowerOfTwo(ProcessorInfo.availableProcessors() << 1);
       memoryAddressCount = getActualAddressCount(desiredSize, lockCount);
       // Unfortunately desired size directly correlates to lock size
-      locks = new StripedLock(lockCount, new ContiguousOffsetCalculator(lockCount));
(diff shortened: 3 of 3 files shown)


core/src/main/java/org/infinispan/container/offheap/ContiguousOffsetCalculator.java
```diff
@@ -36,6 +43,7 @@ public int calculateOffsetUsingHashCode(int hashCode) {
    }
 
    private static int spread(int h) {
-      return (h ^ (h >>> 16)) & HASH_BITS;
+      // Spread using murmur hash then ensure it is positive before finding block to use
+      return hash.hash(h) & HASH_BITS;
```
wburns: Thinking about this some more, I am not sure if this how we want this to finally work. Unfortunately this means we are using the same spread as the distribution, which in DIST and REPL means that we will have a lot of entries that will probably map to close spread values. This might increase hit ratio internally.

danberindei: I wouldn't say it increases the hit ratio, because that sounds like a good thing, increasing the number of conflicts sounds a bit more accurate :) But I agree, if we use the same MH3 spread + division algorithm as `HashFunctionPartitioner` we're likely to end up with some empty buckets.

OTOH I think the modulo offset calculator should be just fine :)

core/src/main/java/org/infinispan/factories/DataContainerFactory.java
```diff
@@ -58,8 +58,8 @@
                if (clusteringConfiguration.cacheMode().needsStateTransfer()) {
                   int segments = clusteringConfiguration.hash().numSegments();
                   Supplier mapSupplier = () -> {
-                     OffHeapEntryFactory entryFactory = componentRegistry.getComponent(OffHeapEntryFactory.class);
-                     OffHeapMemoryAllocator memoryAllocator = componentRegistry.getComponent(OffHeapMemoryAllocator.class);
+                     OffHeapEntryFactory entryFactory = componentRegistry.getOrCreateComponent(OffHeapEntryFactory.class);
+                     OffHeapMemoryAllocator memoryAllocator = componentRegistry.getOrCreateComponent(OffHeapMemoryAllocator.class);
```
danberindei: I must say I don't like `getOrCreateComponent` here, IMO the logic in the supplier can be implemented in `DefaultSegmentedDataContainer` directly.

core/src/main/java/org/infinispan/container/offheap/ContiguousOffsetCalculator.java
danberindei: There's an interesting discussion of hash functions here: https://news.ycombinator.com/item?id=17328756

core/src/main/java/org/infinispan/factories/DataContainerFactory.java
wburns: Hrmm let me look again.

core/src/main/java/org/infinispan/container/offheap/ContiguousOffsetCalculator.java
wburns: Yeah I meant conflict (as in same address bucket).

I can revert this to use modulus offset and 9.4.1 can explore changing this. It should just require undoing the `entryStream` and references to modulus offset.

wburns: I can then test specifically the different offset calculators (get and put should be same perf, but iteration _should_ be faster for contiguous).

core/src/main/java/org/infinispan/factories/DataContainerFactory.java
wburns: I am not sure how you mean here. The `DefaultSegmentedDataContainer` is used for both off heap and on heap containers. This saves a ton of code duplication and fixed issues with expiration for off heap.

Can you share some code for how you recommend doing this? I defer to your registry wizardry :)

wburns:
@danberindei I added a commit to revert to modulus for now. I can fixup later, just let me know.

core/src/main/java/org/infinispan/factories/DataContainerFactory.java
danberindei: No wizardry, just check the configuration in `DefaultSegmentedDataContainer` and create the appropriate maps...

wburns: Don't forget that `DefaultSegmentedDataContainer` is used for the bounded off heap, bounded segmented off heap and bounded segmented on heap variants as well.

The class `DefaultSegmentedDataContainer` currently is a very simple lightweight reusable class that deals with just segmented implementation parts, I am not a fan of complicating it with all the config logic as well.

I would say if you want we can create a separate JIRA to look further into this.

danberindei: Either `DefaultSegmentedDataContainer` is very simple, or using it saved a ton of code duplication, you can't have both :)

OTOH I strongly believe component factories should be very simple and that `ComponentRegistry.get[OrCreate]Component()` should be used sparingly. With ISPN-9127 I'm making the components start only after their dependencies, and calling `get[OrCreate]Component()` hides the dependency.

core/src/main/java/org/infinispan/container/offheap/ContiguousOffsetCalculator.java
wburns: So the modulus implementation of `entryStream` currently prioritizes contiguous memory access to ensure CPU hit performance at the cost of lock acquisition and possible memory allocation.

Note https://github.com/infinispan/infinispan/pull/6249/commits/36bd8127e91c39568df2af61b20bae652f6e500c#diff-8234af2c969ad123972c741f990542bdR677 is invoked for every address pointer, thus having to acquire the `readLock` address pointer number of times (and having to create a Stream for every non 0 address pointer reference).

The contiguous version also reads in contiguous memory for address pointer, but since the locks are kept in contiguous order, we don't have to acquire a lock more than once and only one stream (in best case)
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/container/offheap/OffHeapConcurrentMap.java#L681

In the case that address pointers are higher, which is the more common case it does batching of 256 address pointers at a time. So this reduces the lock acquisition and `Stream` allocation to be 1/256th of what it is with modulus.

https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/container/offheap/OffHeapConcurrentMap.java#L709

Obviously the 256 can be tweaked very easily in the file, hate magic numbers myself.

danberindei: Do you really need a lock to compute the address/bucket index?

If not, you could compute the bucket index outside the lock, and compute the lock index as `bucket index & (lock count - 1)` (assuming lock count is a power of 2).
Then for iteration you could have the same strategy of locking a range of buckets and creating a stream for all their entries.

wburns: So the lock is required to ensure that the off heap memory is still available. If we don't acquire the lock we could crash the JVM if we had a concurrent stop when trying to access memory via the `memoryLookup` object.

Lock count is always a power of 2, and the same with address count. I had thought about locking a range of buckets with modulus, I tried it originally, but it gets ugly really fast as it isn't very conducive to the Stream API. I just sorta gave up. But I would be open to exploring it more as well, just maybe for 9.4.1 :)

core/src/main/java/org/infinispan/factories/DataContainerFactory.java
wburns: Okay :D

I live on the other side of the fence that the Components should be simple and config lives separately, properly separating concerns. I agree that Components should be Injected when possible though, unfortunately the class has different required components based on the config, heh.

danberindei: I don't want to over-complicate the components either, but I'd argue that having to call a supplier that can do anything, like instantiate another component, makes the component more complicated than injecting that component directly.

core/src/main/java/org/infinispan/container/offheap/ContiguousOffsetCalculator.java
danberindei: Definitely +1 to explore locking a contiguous range of buckets.

danberindei:
Integrated, thanks Will!

PR #6252, opened 2018-09-13
ISPN-9512 *TxPartitionAndMerge*Test tests hang during teardown

danberindei:
https://issues.jboss.org/browse/ISPN-9512

Restore the number of threads and queue in transport thread pool.


tristantarrant:
It looks like this solves the issue. There are two failures in core, InitialClusterSizeTest and RetryMechanismTest, but the JVM is no longer killed, so LGTM

tristantarrant:
It fixes the hangs as advertised.

PR #6253, opened 2018-09-13
ISPN-2575 Key Transformer registration is required on all nodes of the cluster, in case of custom keys

anistor:
https://issues.jboss.org/browse/ISPN-2575

Previously, this was done via SearchManagerImplementor.registerKeyTransformer. A very very bad idea, since the SearchManager needs to be up and running to be able to call this method,  but at the same time since it is running somebody (on a different cluster node maybe) might have already added/might be adding right now data to cache that uses a key type that I have not yet been able to register a transformer for it -> Kabooom!

 SearchManagerImplementor.registerKeyTransformer is not going away, it will continue to be supported besides registration via indexing config, but with a big WARNING in javadocs.


anistor:
This needs a new test in server to really make sure it works. I'm adding it right now, so for now it's marked 'preview', please do not integrate it yet.

anistor:
This PR has been replaced by another, ignore it for now

tristantarrant:
Closing this. Please reopen when it's ready.

PR #6256, opened 2018-09-14
ISPN-9514 Entry replaced with same expiration can expire immediately

wburns:
* Make removal more strict in checking if expired
* However provide 100 ms buffer

https://issues.jboss.org/browse/ISPN-9514

### DIFF
--- a/core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
+++ b/core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
@@ -10,7 +10,9 @@ import java.util.Objects;
 import org.infinispan.commands.CommandInvocationId;
 import org.infinispan.commands.Visitor;
 import org.infinispan.commons.io.UnsignedNumeric;
+import org.infinispan.commons.time.TimeService;
 import org.infinispan.commons.util.EnumUtil;
+import org.infinispan.container.entries.ExpiryHelper;
 import org.infinispan.container.entries.MVCCEntry;
 import org.infinispan.container.versioning.IncrementableEntryVersion;
 import org.infinispan.context.InvocationContext;
@@ -30,10 +32,15 @@ import org.infinispan.util.logging.LogFactory;
 public class RemoveExpiredCommand extends RemoveCommand {
    public static final int COMMAND_ID = 58;
    private static final Log log = LogFactory.getLog(RemoveExpiredCommand.class);
+   private static final boolean trace = log.isTraceEnabled();
+
+   // The amount in milliseconds of a buffer we allow the system clock to be off, but still allow expiration removal
+   private static final int CLOCK_BUFFER = 100;
 
    private boolean maxIdle;
    private Long lifespan;
    private IncrementableEntryVersion nonExistentVersion;
+   private TimeService timeService;
 
    public RemoveExpiredCommand() {
       // The value matcher will always be the same, so we don't need to serialize it like we do for the other commands
@@ -41,13 +48,15 @@ public class RemoveExpiredCommand extends RemoveCommand {
    }
 
    public RemoveExpiredCommand(Object key, Object value, Long lifespan, boolean maxIdle, CacheNotifier notifier, int segment,
-                               CommandInvocationId commandInvocationId, IncrementableEntryVersion nonExistentVersion) {
+                               CommandInvocationId commandInvocationId, IncrementableEntryVersion nonExistentVersion,
+         TimeService timeService) {
       //valueEquivalence can be null because this command never compares values.
       super(key, value, notifier, segment, EnumUtil.EMPTY_BIT_SET, commandInvocationId);
       this.lifespan = lifespan;
       this.maxIdle = maxIdle;
       this.valueMatcher = ValueMatcher.MATCH_EXPECTED_OR_NULL;
       this.nonExistentVersion = nonExistentVersion;
+      this.timeService = timeService;
    }
 
    @Override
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
```diff
@@ -85,15 +91,27 @@ public Object perform(InvocationContext ctx) throws Throwable {
...
             // Lastly if there is metadata we have to verify it equals our lifespan and the value match.
-            // TODO: add a threshold to verify this wasn't just created with the same value/lifespan just before expiring
             if (valueMatcher.matches(prevValue, value, null)) {
-               e.setExpired(true);
-               return performRemove(e, prevValue, ctx);
+               // Only remove the entry if it is still expired.
+               // Due to difference in system clocks between nodes - we also accept the expiration within 100 ms
+               // of now. This along with the fact that entries are not normally access immediately when they
+               // expire should give a good enough buffer range to not create a false positive.
+               if (ExpiryHelper.isExpiredMortal(lifespan, e.getCreated(), timeService.wallClockTime() + 100)) {
```
tristantarrant: Can you make this into a constant. Hardcoding is fine, but I like the visibility at the top of the class

wburns:
Updated and fixed a really old typo

tristantarrant:
Merged, thanks

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
```diff
@@ -85,15 +93,27 @@ public Object perform(InvocationContext ctx) throws Throwable {
...
             // If the entries lifespan is not positive that means it can't expire so don't even try to remove it
             // Lastly if there is metadata we have to verify it equals our lifespan and the value match.
-            // TODO: add a threshold to verify this wasn't just created with the same value/lifespan just before expiring
             if (valueMatcher.matches(prevValue, value, null)) {
-               e.setExpired(true);
-               return performRemove(e, prevValue, ctx);
+               // Only remove the entry if it is still expired.
+               // Due to difference in system clocks between nodes - we also accept the expiration within 100 ms
+               // of now. This along with the fact that entries are not normally access immediately when they
+               // expire should give a good enough buffer range to not create a false positive.
```
danberindei: @wburns I'm not sure why we need this, didn't we already check that the entry is expired on all the owners before invoking `performExpire()`?

core/src/test/java/org/infinispan/expiration/impl/ClusterExpirationFunctionalTest.java
```diff
@@ -91,13 +91,14 @@ public void testLifespanExpiredOnBackupOwner() throws Exception {
...
       MagicKey key = createKey(primaryOwner, backupOwner);
-      primaryOwner.put(key, key.toString(), 10, TimeUnit.MINUTES);
+      primaryOwner.put(key, key.toString(), 10, TimeUnit.MILLISECONDS);
 
       assertEquals(key.toString(), primaryOwner.get(key));
       assertEquals(key.toString(), backupOwner.get(key));
 
       // Now we expire on cache0, it should still exist on cache1
-      timeService.advance(TimeUnit.MINUTES.toMillis(10) + 1);
+      // Note this has to be within the buffer in RemoveExpiredCommand (100 ms the time of this commit)
```
danberindei: I don't understand this comment either, doesn't the entry have the same timestamp on both nodes?

core/src/test/java/org/infinispan/expiration/impl/ClusterExpirationFunctionalTest.java
```diff
@@ -313,4 +314,29 @@ public void testMaxIdleAccessSuspectedExpiredEntryRefreshesProperly() {
 
       assertEquals(value, cache0.get(key));
    }
+
+   public void testWriteExpiredEntry() throws InterruptedException {
+      String key = "key";
+      String value = "value";
+
+      for (int i = 0; i < 100; ++i) {
```
danberindei: Why 100 and not just 1?

wburns: It requires replacing an expired entry - and doesn't always happen since the remove expired is async.

core/src/test/java/org/infinispan/expiration/impl/ClusterExpirationFunctionalTest.java
wburns: It has the same lifespan, but the created time can be off due to each node will insert at a slightly different time. For example:

n1: might insert at 1000
n2: might insert at 1013

due to network latency. This buffer is in case if we had a read from n1 at say 1004 and it tried to remove the entry from n2 at 1011, it would fail. This buffer is just to make it a little more lax.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: We do, but this is usually done outside of a lock such as a get. In this case we have to acquire the lock during remove expired and double check it is still expired (in case if a concurrent write happened).

danberindei: Right, but do we even acquire the lock if the entry was not expired on all the owners during the get?

core/src/test/java/org/infinispan/expiration/impl/ClusterExpirationFunctionalTest.java
danberindei: I agree in general, but this test uses `ControlledTimeService`, so the time on n2 should be the same as on n1

core/src/test/java/org/infinispan/expiration/impl/ClusterExpirationFunctionalTest.java
danberindei: Couldn't we use `ControlTimeService` to make it happen deterministically?

wburns: We have to also coordinate 2 rpcs to interleave at exact points.

core/src/test/java/org/infinispan/expiration/impl/ClusterExpirationFunctionalTest.java
wburns: There are two time services in this test and it only increments one of those.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: hrmm, there is no lock acquisition at all during a get. We don't fire the remove expired command during a get if the entry wasn't expired. Or maybe you are trying to ask something else?

PR #6259, opened 2018-09-17
ISPN-9320 Hot Rod protocol client / server handshake

tristantarrant:
https://issues.jboss.org/browse/ISPN-9320

* It bumps the Hot Rod protocol to 3.0 (we cannot do 2.10 because the code internally does major*10+minor. We could separate the version name from the version number, but I'd rather not introduce confusion)
* The PING operation has been "enhanced" to include the maximum version supported by the server as well as a list of supported requests with their opcode (so that we make ops and protocol version orthogonal in the future).
* The clients default to an "automatic" version.


gustavonalle:
Is this being target for final? It seems we should postpone to 10.0.0 as this is a change in the protocol with potential side effects for clients

tristantarrant:
Yes, this is for 10

tristantarrant:
And we should probably bump the protocol to 3.0

documentation/src/main/asciidoc/hotrod_protocol/hotrod_protocol.asciidoc
```diff
@@ -2,24 +2,25 @@
...
+* link:#hot_rod_protocol_2_6[Hot Rod Protocol 2.6 (Infinispan 9.0)]
+* link:#hot_rod_protocol_2_7[Hot Rod Protocol 2.7 (Infinispan 9.2)]
+* link:#hot_rod_protocol_2_8[Hot Rod Protocol 2.8 (Infinispan 9.3)]
+* link:#hot_rod_protocol_2_9[Hot Rod Protocol 2.9 (Infinispan 9.4)]
+* link:#hot_rod_protocol_3_0[Hot Rod Protocol 3.0 (Infinispan 10.0)]
 
 ==== Hot Rod Protocol 1.0
 
-.{brandname} versions
+.Infinispan versions
```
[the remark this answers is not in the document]
oraNod: We can use conditional statements to exclude the versions that we don't want in the product docs. Like this:

```ifndef::productized[]
String to exclude from the product docs.
endif::productized[]```

I added a `:productized:` attribute in documentation/src/main/asciidoc/topics/attributes/product-attributes.adoc so we can include/exclude text as needed.

**Comment by tristantarrant** [comment]
This is now ready. I've greatly simplified the original proposal by folding the changes into the PING operation. The bug I uncovered wrt topology changes closing the channel before processing the whole operation will be resolved in https://issues.jboss.org/browse/ISPN-9827

**Comment by tristantarrant** [comment]
Rebased. Build is now green.

File: client/hotrod-client/src/main/java/org/infinispan/client/hotrod/impl/operations/PingResponse.java [hunk 2]
```diff
@@ -0,0 +1,128 @@
...
+                     break;
+                  }
+                  ++decoderState;
+               case 1:
+                  serverVersion = ProtocolVersion.getBestVersion(buf.readUnsignedByte());
+                  decoder.checkpoint();
+                  ++decoderState;
+               case 2:
+                  serverOpsCount = ByteBufUtil.readVInt(buf);
+                  serverOps = new TreeSet<>();
```
wburns: Is there a reason this is a `TreeSet`?

tristantarrant: I wanted it to maintain order

wburns:

PR #6260, opened 2018-09-18
ISPN-9511 Expired event is not fired when modifying an expired entry

wburns:
https://issues.jboss.org/browse/ISPN-9511

### DIFF
--- a/core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
+++ b/core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
@@ -11,7 +11,6 @@ import org.infinispan.commands.CommandInvocationId;
 import org.infinispan.commands.Visitor;
 import org.infinispan.commons.io.UnsignedNumeric;
 import org.infinispan.commons.time.TimeService;
-import org.infinispan.commons.util.EnumUtil;
 import org.infinispan.container.entries.ExpiryHelper;
 import org.infinispan.container.entries.MVCCEntry;
 import org.infinispan.container.versioning.IncrementableEntryVersion;
@@ -48,10 +47,10 @@ public class RemoveExpiredCommand extends RemoveCommand {
    }
 
    public RemoveExpiredCommand(Object key, Object value, Long lifespan, boolean maxIdle, CacheNotifier notifier, int segment,
-                               CommandInvocationId commandInvocationId, IncrementableEntryVersion nonExistentVersion,
+                               long flagBitSet, CommandInvocationId commandInvocationId, IncrementableEntryVersion nonExistentVersion,
          TimeService timeService) {
       //valueEquivalence can be null because this command never compares values.
-      super(key, value, notifier, segment, EnumUtil.EMPTY_BIT_SET, commandInvocationId);
+      super(key, value, notifier, segment, flagBitSet, commandInvocationId);
       this.lifespan = lifespan;
       this.maxIdle = maxIdle;
       this.valueMatcher = ValueMatcher.MATCH_EXPECTED_OR_NULL;
@@ -160,6 +159,8 @@ public class RemoveExpiredCommand extends RemoveCommand {
          output.writeBoolean(false);
       }
       output.writeBoolean(maxIdle);
+      output.writeLong(FlagBitSets.copyWithoutRemotableFlags(getFlagsBitSet()));
+
    }
 
    @Override
@@ -175,6 +176,7 @@ public class RemoveExpiredCommand extends RemoveCommand {
          lifespan = null;
       }
       maxIdle = input.readBoolean();
+      setFlagsBitSet(input.readLong());
    }
 
    @Override
@@ -189,12 +191,6 @@ public class RemoveExpiredCommand extends RemoveCommand {
    @Override
    public int hashCode() {
       return Objects.hash(super.hashCode(), lifespan, maxIdle);
-   }
-
-   @Override
-   public long getFlagsBitSet() {
-      // Override the flags
-      return FlagBitSets.SKIP_CACHE_LOAD;
    }
 
    /**
--- a/core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
+++ b/core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
@@ -2,6 +2,7 @@ package org.infinispan.container.impl;
 
 import static org.infinispan.commons.util.Util.toStr;
 
+import org.infinispan.commons.time.TimeService;
 import org.infinispan.configuration.cache.Configuration;
 import org.infinispan.configuration.cache.Configurations;
 import org.infinispan.container.entries.CacheEntry;
@@ -13,11 +14,12 @@ import org.infinispan.container.entries.RepeatableReadEntry;
 import org.infinispan.container.entries.VersionedRepeatableReadEntry;
 import org.infinispan.container.versioning.VersionGenerator;
 import org.infinispan.context.InvocationContext;
+import org.infinispan.distribution.DistributionManager;
+import org.infinispan.expiration.impl.InternalExpirationManager;
 import org.infinispan.factories.annotations.Inject;
 import org.infinispan.factories.annotations.Start;
 import org.infinispan.metadata.EmbeddedMetadata;
 import org.infinispan.metadata.Metadata;
-import org.infinispan.commons.time.TimeService;
 import org.infinispan.util.concurrent.IsolationLevel;
 import org.infinispan.util.logging.Log;
 import org.infinispan.util.logging.LogFactory;
@@ -37,6 +39,8 @@ public class EntryFactoryImpl implements EntryFactory {
    @Inject private Configuration configuration;
    @Inject private TimeService timeService;
    @Inject private VersionGenerator versionGenerator;
+   @Inject private DistributionManager distributionManager;
+   @Inject private InternalExpirationManager expirationManager;
 
    private boolean isL1Enabled;
    private boolean useRepeatableRead;
@@ -94,7 +98,7 @@ public class EntryFactoryImpl implements EntryFactory {
             log.tracef("Updated context entry %s -> %s", contextEntry, mvccEntry);
       } else {
          // Not in the context yet.
-         CacheEntry cacheEntry = getFromContainer(key, segment, isOwner, true);
+         CacheEntry cacheEntry = getFromContainerForWrite(key, segment, isOwner);
          if (cacheEntry == null) {
             return;
          }
@@ -125,7 +129,7 @@ public class EntryFactoryImpl implements EntryFactory {
             log.tracef("Updated context entry %s -> %s", contextEntry, mvccEntry);
       } else {
          // Not in the context yet.
-         CacheEntry cacheEntry = innerGetFromContainer(key, segment, true, true);
+         CacheEntry cacheEntry = innerGetFromContainerForWrite(key, segment, true, false);
          if (cacheEntry == null) {
             cacheEntry = NullCacheEntry.getInstance();
          }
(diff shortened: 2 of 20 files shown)


wburns:
Upon thinking about this further, the fix for ISPN-9530 is not correct.

The following line https://github.com/infinispan/infinispan/blob/master/server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java#L206 should really use regular `registerFilterListener`. I need to look into fixing it when a converter is passed, why it didn't work before. My guess is that it applied the encoding is done twice. I will have to look further.

wburns:
I will need to rebase this after #6263 is integrated. However it should hopefully pass CI now.

wburns:
Tests passing, rebased to latest master and added some more docs

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
```diff
@@ -193,8 +196,8 @@ public int hashCode() {
 
    @Override
    public long getFlagsBitSet() {
-      // Override the flags
-      return FlagBitSets.SKIP_CACHE_LOAD;
+      // Override the flags to always include skip cache load
+      return flags | FlagBitSets.SKIP_CACHE_LOAD;
```
danberindei: Maybe it would be nicer to add the `SKIP_CACHE_LOAD` flag in `CacheImpl`

core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
```diff
@@ -197,17 +201,25 @@ private CacheEntry getFromContext(InvocationContext ctx, Object key) {
...
+         final InternalCacheEntry ice = innerGetFromContainer(key, segment, false, isPrimaryOwner(isOwner, segment));
          if (trace)
             log.tracef("Retrieved from container %s", ice);
          if (ice == null) {
             return NullCacheEntry.getInstance();
          }
          return ice;
       } else if (isL1Enabled) {
-         final InternalCacheEntry ice = innerGetFromContainer(key, segment, writeOperation, false);
+         final InternalCacheEntry ice = innerGetFromContainer(key, segment, false, isPrimaryOwner(isOwner, segment));
```
danberindei: Could it really be a primary owner if it wasn't an owner?

core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
```diff
@@ -228,22 +240,19 @@ private CacheEntry getFromContainerForRead(Object key, int segment, boolean isOw
...
+   private InternalCacheEntry innerGetFromContainer(Object key, int segment, boolean returnExpired, boolean isPrimaryOwner) {
+      InternalCacheEntry ice = container.peek(segment, key);
+      if (ice != null && !returnExpired) {
+         long currentTime = timeService.wallClockTime();
+         if (ice.isExpired(currentTime)) {
+            // This means it is a write operation that isn't expiration and we are the owner, thus we should
+            // actually expire the entry from memory
+            if (isPrimaryOwner) {
+               // This method is always called from a write operation
+               expirationManager.entryExpiredInMemory(ice, currentTime, true).join();
```
danberindei: I'm not sure about this, it feels like we're spreading the expiration logic in too many places. And wrapping entries shouldn't block...

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
```diff
@@ -126,25 +127,28 @@ private void addAndWaitIfFull(CompletableFuture future, List<CompletableFuture>
       }
    }
 
-   CompletableFuture<Void> handleLifespanExpireEntry(K key, V value, long lifespan) {
+   // Skip locking is available in case if the entry was found expired due to a write, which would already
+   // have the lock and remove lifespan expired is fired async
```
danberindei: I'm a bit confused... If the `RemoveExpiredCommand` is fired async, why would it need to skip locking? Also, does `available` mean `true` here?

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
```diff
@@ -188,12 +194,18 @@ private void addAndWaitIfFull(CompletableFuture future, List<CompletableFuture>
          expiredMortal = ExpiryHelper.isExpiredMortal(lifespan, entry.getCreated(), currentTime);
       }
       if (expiredMortal) {
-         handleLifespanExpireEntry(entry.getKey(), value, lifespan);
+         // If the entry as found expired in memory during a write, that means we already have the lock for the
+         // given entry. Thus we must skip locking and wait for the remove expired to complete before continuing
+         // forward with the operation to let the event be raised properly.
```
danberindei: Do we guarantee that expiration listeners are notified synchronously with the read/write that triggered the expiration, or does "the event be raised properly" mean something else?

core/src/main/java/org/infinispan/interceptors/locking/NonTransactionalLockingInterceptor.java
```diff
@@ -40,6 +41,11 @@ protected Object visitDataWriteCommand(InvocationContext ctx, DataWriteCommand c
       return visitNonTxDataWriteCommand(ctx, command);
    }
 
+   @Override
+   public Object visitRemoveExpiredCommand(InvocationContext ctx, RemoveExpiredCommand command) throws Throwable {
+      return super.visitRemoveExpiredCommand(ctx, command);
```
danberindei: Is this really needed?

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
wburns: It needs to skip locking, because it has to remove the expired entry and generate the event before the current write can occur. The current write has the lock, so the async remove expired can't acquire locks. It waits for the remove to complete causing them to be serialized and guarantees event ordering so the expiration event comes before the write event.

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
wburns: yes they are done in a serial manner, expiration event has to come before the write event.

core/src/main/java/org/infinispan/interceptors/locking/NonTransactionalLockingInterceptor.java
wburns: No :) Was just there to add a debug break point and forgot to remove.

core/src/main/java/org/infinispan/commands/write/RemoveExpiredCommand.java
wburns: Hrmm, okay. Let me try to reorg it.

core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
wburns: No, but it seemed a little cleaner than adding an && here and hard coding a value in `isPrimaryOwner`. But either way.

core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
wburns: Where would you recommend the logic should go? I don't have a good spot since we don't want this in the `DataContainer`. Unless you think we should move the other logic out of the `DataContainer` into here?

Wrapping entries already blocks for max idle :)

danberindei: Why not in the `DataContainer`? It already deals with expiration...

Although moving the expiration logic somewhere where it could be made non-blocking does sound better.

wburns: I responded quickly to run out, but the more I am thinking I would rather probably have this in `EntryFactoryImpl` and make both methods non blocking with interceptor stack. Although I haven't looked at details if this is possible.

core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
wburns: I was going to change it yeah :)

wburns: Gotcha, I wasn't look at which branch it was on 👍

core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
wburns: Now that I got a chance to look at this, I am not sure how to make this non blocking without changing the `EntryFactory` interface. What do you think @danberindei ? You have had a lot more experience changing these things.

danberindei: Yeah, making it non-blocking is a lot more involved. It's in an impl package so changing the interface is not a big deal, but I wouldn't push for it in 9.4.

Not sure if moving everything expiration-related to `EntryFactoryImpl` right now is feasible either, that's why my first thought was to keep everything in the data container.

wburns: Yeah to add it to the `DataContainer` would require changing that interface, which is a lot more difficult to do. So I would rather move it into `EntryFactoryImpl` at some point and eventually try to push that to the `EntryWrappingInterceptor`, but I don't see a nice easy way to do that. I would say what is here is probably what will have to be for 9.4.0.

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
danberindei: Ok, but if we block the current write until the `RemoveExpiredCommand` finishes running, I wouldn't really call it async, even if it does run on another thread.

core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
```diff
@@ -197,17 +201,25 @@ private CacheEntry getFromContext(InvocationContext ctx, Object key) {
       return cacheEntry;
    }
 
-   private CacheEntry getFromContainer(Object key, int segment, boolean isOwner, boolean writeOperation) {
+   private boolean isPrimaryOwner(boolean isOwner, int segment) {
+      if (distributionManager != null) {
+         return distributionManager.getCacheTopology().getSegmentDistribution(segment).isPrimary();
+      } else {
+         return isOwner;
```
danberindei: Actually `distributionManager == null` only in `LOCAL` caches, so you can return `true` here and remove the parameter.

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
wburns: okay

core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
danberindei: Ok for now :)

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
wburns: technically it is async from this method :)

core/src/main/java/org/infinispan/container/impl/EntryFactoryImpl.java
wburns: I agree.

core/src/main/java/org/infinispan/expiration/impl/ClusterExpirationManager.java
danberindei: Sure, but that doesn't explain why it needs to skip locking :)

wburns: Sure, it just was supposed to say required instead of available. The problem is that the write already has the lock as it mentioned. I can add another sentence stating that it could cause a deadlock.

I think the confusion is because the blocking is done by the caller. I will also add a note mentioning that when `skipLocking` is true it is up to the caller to properly ensure that the lock is held for the entire duration.

wburns: I reworded and changed variable names so it should be more clear now.

danberindei: Thanks, it's a little more detail than I wanted now, but still much clearer :)

danberindei:
Integrated, thanks Will!
The build failure was caused by ISPN-9588.

PR #6265, opened 2018-09-20
ISPN-7863 Ickle lexer wrongly discards letter v as whitespace ruining…

anistor:
… parsing of identifiers containing v

* the \v in the WS definition should stand for vertical tab but this escape char
  is not recognized and is treated like letter v instead

* just undo the change in IckleLexer.g to see the bug happening

https://issues.jboss.org/browse/ISPN-7863


tristantarrant:
Merged, thanks

PR #6268, opened 2018-09-20
ISPN-9535 Feature toggles

tristantarrant:
https://issues.jboss.org/browse/ISPN-9535

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.java
@@ -29,6 +29,7 @@ import org.infinispan.client.hotrod.logging.LogFactory;
 import org.infinispan.commons.configuration.Builder;
 import org.infinispan.commons.marshall.Marshaller;
 import org.infinispan.commons.marshall.jboss.GenericJBossMarshaller;
+import org.infinispan.commons.util.Features;
 import org.infinispan.commons.util.StringPropertyReplacer;
 import org.infinispan.commons.util.TypedProperties;
 import org.infinispan.commons.util.Util;
@@ -77,8 +78,8 @@ public class ConfigurationBuilder implements ConfigurationChildBuilder, Builder<
    private int batchSize = ConfigurationProperties.DEFAULT_BATCH_SIZE;
    private final TransactionConfigurationBuilder transaction;
    private final StatisticsConfigurationBuilder statistics;
-
    private final List<ClusterConfigurationBuilder> clusters = new ArrayList<>();
+   private Features features;
 
    public ConfigurationBuilder() {
       this.classLoader = new WeakReference<>(Thread.currentThread().getContextClassLoader());
@@ -435,11 +436,12 @@ public class ConfigurationBuilder implements ConfigurationChildBuilder, Builder<
 
       return new Configuration(asyncExecutorFactory.create(), balancingStrategyFactory, classLoader == null ? null : classLoader.get(), clientIntelligence, connectionPool.create(), connectionTimeout,
             consistentHashImpl, forceReturnValues, keySizeEstimate, marshaller, marshallerClass, protocolVersion, servers, socketTimeout, security.create(), tcpNoDelay, tcpKeepAlive,
-            valueSizeEstimate, maxRetries, nearCache.create(), serverClusterConfigs, whiteListRegExs, batchSize, transaction.create(), statistics.create());
+            valueSizeEstimate, maxRetries, nearCache.create(), serverClusterConfigs, whiteListRegExs, batchSize, transaction.create(), statistics.create(), features);
    }
 
    @Override
    public Configuration build() {
+      features = new Features(classLoader.get());
       return build(true);
    }
 
--- a/commons/src/main/java/org/infinispan/commons/util/Features.java
+++ b/commons/src/main/java/org/infinispan/commons/util/Features.java
@@ -1,0 +1,57 @@
+package org.infinispan.commons.util;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.util.Enumeration;
+import java.util.Properties;
+
+import org.infinispan.commons.CacheConfigurationException;
+import org.infinispan.commons.logging.Log;
+import org.infinispan.commons.logging.LogFactory;
+
+/**
+ * Features allow conditional enabling/disabling of Infinispan's functionality.
+ * They are represented as named properties of the form 'org.infinispan.feature.*' and can either be set via one or
+ * more 'META-INF/infinispan-features.properties' file on the classpath or by setting a system property,
+ * e.g. -Dorg.infinispan.feature.A=true
+ */
+public class Features {
+   private static final Log log = LogFactory.getLog(Features.class);
+   public static final String FEATURE_PREFIX = "org.infinispan.feature.";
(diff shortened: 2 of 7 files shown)


commons/src/main/java/org/infinispan/commons/util/Features.java
```diff
@@ -0,0 +1,40 @@
...
+         // Ignore
+      }
+   }
+
+   public Features() {
+      this(Features.class.getClassLoader());
+   }
+
+   public boolean isAvailable(String featureName) {
+      return Boolean.getBoolean(featureName) || Boolean.parseBoolean(features.getProperty(featureName, "true"));
```
anistor: So this allows an env prop to enable a feature (in spite of whatever is found in the meta-inf files) but it does not allow for disabling a feature. Is this asymmetry intended? I find it quite confusing. I'd imagine it would be less confusing if env props would act as an override for both true and false values.

tristantarrant: While the sysprop should indeed "override" a feature both ways, disabling an otherwise enabled feature thrrough a property is not something I see as particularly useful.

tristantarrant:
I have amended the behaviour so that sysprops always override with both meanings (true/false). I also made the feature name prefix a constant so that code just has to use Features.isAvailable("xyz") to mean "org.infinispan.feature.xyz"

commons/src/main/java/org/infinispan/commons/util/Features.java
```diff
@@ -0,0 +1,53 @@
...
+      }
+   }
+
+   public Features() {
+      this(Features.class.getClassLoader());
+   }
+
+   public boolean isAvailable(String featureName) {
+      String name = FEATURE_PREFIX + featureName;
+      String sysprop = SecurityActions.getProperty(name);
```
anistor: Oh yes, I was just preparing to comment on the SecurityManager :)

anistor:
This ```Features``` class has no intrinsic state of its own. All 'state' is purely static, coming from env vars or classloader resources. So I wonder if this class needs to have any instances to be carried around in our config objects or we could just have a static method ```Features.isEnabled(String f)```

anistor:
Oh, actually the classloader ref is its internal state, so this making this a static global singleton is not doable. Pls ignore my earlier comment :)

commons/src/main/java/org/infinispan/commons/util/Features.java
```diff
@@ -0,0 +1,53 @@
...
+import org.infinispan.commons.logging.LogFactory;
+
+/**
+ * Features allow conditional enabling/disabling of Infinispan's functionality.
+ * They are represented as named properties of the form 'org.infinispan.feature.*' and can either be set via one or
+ * more 'META-INF/infinispan-features.properties' file on the classpath or by setting a system property,
+ * e.g. -Dorg.infinispan.feature.A=true
+ */
+public class Features {
+   private static final Log log = LogFactory.getLog(Features.class);
```
wburns: Seems unused.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.java
```diff
@@ -398,6 +399,7 @@ public ConfigurationBuilder withProperties(Properties properties) {
 
    @Override
    public void validate() {
+
```
wburns: Unneeded

commons/src/main/java/org/infinispan/commons/util/Features.java
wburns: Unused

commons/src/main/java/org/infinispan/commons/util/Features.java
```diff
@@ -0,0 +1,53 @@
...
+      try {
+         Enumeration<URL> featureFiles = classLoader.getResources(FEATURES_FILE);
+         while (featureFiles.hasMoreElements()) {
+            URL url = featureFiles.nextElement();
+            try (InputStream is = url.openStream()) {
+               features.load(is);
+            } catch (IOException e) {
+            }
+         }
+      } catch (IOException e) {
```
wburns: I wonder should we put this try/catch inside the while loop, so we can try to read other resources?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/configuration/ConfigurationBuilder.java
tristantarrant: Fixed.

commons/src/main/java/org/infinispan/commons/util/Features.java
tristantarrant: I am now using it by listing loaded features.

commons/src/main/java/org/infinispan/commons/util/Features.java
```diff
@@ -0,0 +1,57 @@
...
+   public Features(ClassLoader classLoader) {
+      features = new Properties();
+      try {
+         Enumeration<URL> featureFiles = classLoader.getResources(FEATURES_FILE);
+         while (featureFiles.hasMoreElements()) {
+            URL url = featureFiles.nextElement();
+            try (InputStream is = url.openStream()) {
+               features.load(is);
+            } catch (IOException e) {
+               log.debugf(e, "Error while loading %s", url);
```
anistor: Why not fail fast and loud ?

commons/src/main/java/org/infinispan/commons/util/Features.java
```diff
@@ -0,0 +1,57 @@
...
+            }
+         }
+         if (log.isDebugEnabled()) {
+            features.entrySet().forEach(e -> log.debugf("Feature %s=%s", e.getKey(), e.getValue()));
+         }
+      } catch (IOException e) {
+         log.debugf(e, "Error while attempting to obtain `%s` resources from the classpath", FEATURES_FILE);
```
anistor: I'd let it fail.

commons/src/main/java/org/infinispan/commons/util/Features.java
tristantarrant: In case this fails, nothing bad really happens. It's just that all features will be available.

tristantarrant:
Fail fast at your service

anistor:
LGTM

commons/src/main/java/org/infinispan/commons/util/Features.java
```diff
@@ -0,0 +1,60 @@
...
+      features = new Properties();
+      try {
+         Enumeration<URL> featureFiles = classLoader.getResources(FEATURES_FILE);
+         while (featureFiles.hasMoreElements()) {
+            URL url = featureFiles.nextElement();
+            try (InputStream is = url.openStream()) {
+               features.load(is);
+            } catch (IOException e) {
+               log.debugf(e, "Error while loading %s", url);
+               throw e;
```
wburns: Should just `throw new CacheConfigurationException(e)` ? Otherwise the outer catch block will print additional messages. Unless you want those when this happens as well?

anistor: But this will never happen. It cannot happen. Never. :)

wburns:
I see you took out the inner catch block. User wouldn't know what file it errored on, but this shouldn't happen anyways. I would just like it in master anyways, so if you have no objections @anistor, I would like to integrate.

anistor:
yes @wburns , I'm ok with those little snags. as I said, it's the kind of error that never happen ;)

wburns:

PR #6269, opened 2018-09-21
ISPN-9528 suggested edits for clustered maxIdle

oraNod:


documentation/src/main/asciidoc/user_guide/eviction.adoc
```diff
@@ -145,12 +145,14 @@ times can be configured by adding the
...
 element to your `<*-cache />`  configuration sections.
 
-When an entry expires it will reside in the data container or cache store
-until it is accessed again by a user request.  There is also an optional
-expiration reaper that can run at a given configurable interval of
-milliseconds which will check for expired entries and remove them. Expiration reaper
-is always enabled when a cache store is present. It is highly recommended to always enable
-the reaper when using max idle in a clustered cache.
+When an entry expires it resides in the data container or cache store
+until it is accessed again by a user request. A reaper is also available to check for expired entries and remove them. The reaper runs at a configurable interval of milliseconds and is always enabled when a cache store is present.
```
wburns: The more I wonder, maybe this should be stated as "cannot be disabled" instead of "is always enabled". The reaper is enabled by default as well. Not sure if we should mention that here as well.

documentation/src/main/asciidoc/user_guide/eviction.adoc
```diff
@@ -145,12 +145,14 @@ times can be configured by adding the
...
+//dnaro: 'always enabled when a cache store is present' - Need to clarify this a bit maybe. By default the expiration reaper is enabled when a cache store...
```
wburns: Not sure if we want to specify this or not, but the reason the reaper can't be disabled when a store is present, is because we don't remove entries that are expired from the store on access (only in memory - which removes both).

documentation/src/main/asciidoc/user_guide/eviction.adoc
```diff
@@ -163,45 +165,57 @@ and if this limit is exceeded, some candidates are found to be removed according
...
+Specifies how long entries can remain idle before they expire. An entry in the cache is idle when no operation is performed with the key. The default value is `-1`, which is unlimited time.
 
 ===  Expiration details
 
 .  _Expiration_ is a top-level construct, represented in the configuration as well as in the cache API.
-.  While eviction is _local to each cache instance_ , expiration is _cluster-wide_ .  Expiration lifespans and maxIdle values are replicated along with the cache entry.
-.  Max idle requires additional network messages to work in a cluster.  As such using maxIdle in a clustered cache will cause some operations to be slower at times.
-.  Expiration lifespan and maxIdle are also persisted in CacheStores, so this information survives eviction/passivation.
-
+.  While eviction is _local to each cache instance_ , expiration is _cluster-wide_ .  Expiration lifespans and `maxIdle` values are replicated along with the cache entry.
```
wburns: We should probably be consistent with `lifespan` and `maxIdle` both.

documentation/src/main/asciidoc/user_guide/eviction.adoc
```diff
@@ -163,45 +165,57 @@ and if this limit is exceeded, some candidates are found to be removed according
...
+[[max_idle_clustered]]
+===== Clustered Max Idle
 
-This check is done when a given node finds an entry has expired via maxIdle. It then will ask all the other owners what their last access time was (if it hasn't expired). If all owners say that the entry has expired, then it will not be returned to the user and removed based on the operation.
+In clustered environments, nodes in the cluster can have different access times for the same entry. Entries do not expire from the cache until they reach the maxium idle time for all owners across the cluster.
 
-WARN: Clustered caches should always enable reaper when using max idle expiration due to some limitations listed in the section below.
+When a node detects that an entry has reached the maximum idle time and is expired, the node gets the last time that the entry was accessed from the other owners in the cluster. If the other owners indicate that the entry is expired, that entry is not returned to the user and removed from the cache.
+//dnaro: Returned to the user? Maybe returned to the requester?
```
wburns: Sure, whichever is more consistent with rest of docs.

documentation/src/main/asciidoc/user_guide/eviction.adoc
```diff
@@ -163,45 +165,57 @@ and if this limit is exceeded, some candidates are found to be removed according
...
-. When an entry is found to have expired due to max idle in the expiration reaper, instead of updating the last access time to current time, it will update all nodes to the most recent access of all owners.
-. Iteration in a clustered cache *will* return entries that this node thinks are expired via max idle. This is to ensure performance as no remote invocations are done. The entry is not refreshed and may be removed by the reaper or get access.
-. A clustered transactional cache does *NOT* remove entries expired via max idle on a get. These entries are only removed via reaper. This is prevented as doing so can cause deadlocking.
+* If one or more owner in the cluster detects that an entry is not expired, then a `Cache.get` operation returns the entry. The last access time for that entry is also updated to the current time.
+* When the expiration reaper finds entries that are expired with the maximum idle time, the reaper updates all nodes to the most recent access time instead of updating the last access time to the current time.
```
wburns: Think I left part of this out. It should be something more like:
"When the expiration reaper finds a possibly expired entry, but another node says it is not expired, all nodes will update their last access time to the most recent non expired access time."

Or something like that.

documentation/src/main/asciidoc/user_guide/eviction.adoc
```diff
@@ -163,45 +165,57 @@ and if this limit is exceeded, some candidates are found to be removed according
...
+//dnaro: Why would it update the last access time to the current time? Wouldn't that make the entry not expired?
+//dnaro: The reaper updates the expired entries across all nodes so that the access time for the entries is the most recent time?
+* Clustered transactional caches do *not* remove entries that are expired with the maximum idle time on get operations. These expired entries are removed with the expiration reaper thread only, otherwise deadlocking can occur.
+* Iteration across a clustered cache returns entries that might be expired with the maximum idle time. This behavior ensures performance because no remote invocations done during the iteration. However this does not refresh any expired entries, which are removed by the expiration reaper or when accessed directly (`Cache.get`).
```
wburns: probably should be "because no remote invocations are performed during the iteration"

documentation/src/main/asciidoc/user_guide/eviction.adoc
```diff
@@ -163,45 +165,57 @@ and if this limit is exceeded, some candidates are found to be removed according
...
 
-WARN: Care should be used when utilizing max idle expiration and exception based eviction since expired entries that aren't removed still will count towards container sizing
-
-INFO: Entries transferred via state transfer will be created on the new node with a last access time equal to the current time, essentially refreshing any of these entries
+[IMPORTANT]
+====
+* Clustered caches should always use the expiration reaper with the `maxIdle` configuration.
+//dnaro: Feel like we should expand this to be more explicit with the reasons why.
```
wburns: The biggest reasons are the 2 just above and 1 just below here

1. They aren't removed via get when using transactions 
2. Iteration still includes them until they are removed
3. Eviction normally removes old entries, but exception based doesn't do this - taking up space for possibly a long time (this is true of lifespan expiration as well though)

oraNod: yeah, maybe we should leave it as-is to avoid unnecessary repetition

tristantarrant:
Rebased, squashed and pushed.

PR #6270, opened 2018-09-24
ISPN-9536 Distributed stream iteration timeout during join

danberindei:
https://issues.jboss.org/browse/ISPN-9536

* Don't consider local segments to be lost while starting
* Delete StreamSegmentResponseCommand


core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
```diff
@@ -284,7 +285,8 @@ public void close() {
             if (localRun) {
                localValue = op.performOperation();
                // TODO: we can do this more efficiently - since we drop all results locally
-               if (dm.getReadConsistentHash().equals(ch)) {
+               LocalizedCacheTopology newCacheTopology = dm.getCacheTopology();
```
wburns: Just to confirm this is just to trace the cache topology here right? There is no functional difference?

core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
```diff
@@ -294,13 +296,13 @@ public void close() {
                   }
                   remoteResults.onCompletion(null, ourSegments, localValue);
                } else {
+                  IntSet ourSegments = IntSets.mutableFrom(ch.getPrimarySegmentsForOwner(localAddress));
```
wburns: Maybe we should have this a little closer to what it was before, but I like the new debug message. This way we don't copy the IntSet for the case when `segmentsToProcess` is null, which can happen during a rehash very easily.

```java
IntSet ourSegments;
if (segmentsToProcess != null) {
   ourSegments = IntSets.mutableFrom(ch.getPrimarySegmentsForOwner(localAddress));
   ourSegments.retainAll(segmentsToProcess);
} else {
   ourSegments = IntSets.from(ch.getPrimarySegmentsForOwner(localAddress));
}
getLog().debugf("Local cache topology changed from %d to %d, suspecting local segments %s",
   cacheTopology.getTopologyId(), newCacheTopology.getTopologyId(), ourSegments);
remoteResults.onSegmentsLost(ourSegments);
```

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -321,7 +321,7 @@ private void handleResponseError(CompletionStage<ValidResponse> rpcFuture, Objec
             log.tracef("UnRegistered change listener for %s", requestId);
          }
       }
-      if (cache.getStatus() != ComponentStatus.RUNNING) {
+      if (cache.getStatus() != ComponentStatus.RUNNING && cache.getStatus() != ComponentStatus.INITIALIZING) {
```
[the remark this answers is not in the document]
danberindei: Yes, this is the fix

core/src/main/java/org/infinispan/stream/impl/StreamResponseCommand.java
```diff
@@ -33,11 +36,16 @@ public StreamResponseCommand(ByteString cacheName) {
    }
 
    public StreamResponseCommand(ByteString cacheName, Address origin, Object id, boolean complete, R response) {
+      this(cacheName, origin, id, complete, IntSets.immutableEmptySet(), response);
+   }
+   public StreamResponseCommand(ByteString cacheName, Address origin, Object id, boolean complete,
```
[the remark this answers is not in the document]
danberindei: Yeah, initially I just wanted to add a `toString()` implementation in the subclass, but then I thought an empty set should be just 1 byte so there isn't much to optimize.

core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
danberindei: Yes, initially I thought the segments were marked as lost because of a topology change, so I wanted to log the topology ids.

core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
danberindei: I was thinking that this shouldn't happen too often because the topology doesn't change that often, but I'll look into using `LocalizedCacheTopology.getLocalReadSegments()` as well.

core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
```diff
@@ -283,24 +284,23 @@ public void close() {
...
                // TODO: we can do this more efficiently - since we drop all results locally
-               if (dm.getReadConsistentHash().equals(ch)) {
-                  IntSet ourSegments;
-                  if (segmentsToProcess != null) {
-                     ourSegments =  IntSets.mutableFrom(ch.getPrimarySegmentsForOwner(localAddress));
-                     ourSegments.retainAll(segmentsToProcess);
-                  } else {
-                     ourSegments = IntSets.from(ch.getPrimarySegmentsForOwner(localAddress));
-                  }
+               LocalizedCacheTopology newCacheTopology = dm.getCacheTopology();
```
danberindei: @wburns I'm not sure about the comment here, does it mean we could do better than drop all the local segments, or does it mean that we could replace the CH equality check with something more efficient?

danberindei: Sorry, in the side-by-side view I was writing my comment just below the comment on line 286 (295 with my changes) :)

I was thinking that would work if the local node didn't lose any segments in the new CH, so we could check directly if `newCacheTopology.getLocalReadSegments()` is a superset of `cacheTopology.getLocalReadSegments()` or not instead of checking all the segments of all the nodes.

I'm not sure I understood your comment, did you have something like this in mind, or were you thinking about when `segmentsToFilter` is set and we don't even need all the segments that were local in the initial CH?

wburns: This is what I was thinking, but unfortunately using `newCacheTopology.getLocalReadSegments` and `cacheTopology.getLocalReadSegments` is not sufficient. It is entirely possible for the cache topology to change more than once, and I have tests where a segment can move from owned to not owned and back to owned again :( In this case it is possible for an entry to be missed. This is why the LocalStreamManagerImpl registers a listener and tracks the segments across all the topology changes not just the first and most recent.

wburns:

core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
```diff
@@ -283,24 +284,23 @@ public void close() {
             boolean localRun = ch.getMembers().contains(localAddress);
             if (localRun) {
                localValue = op.performOperation();
+
+               IntSet ourSegments;
+               if (segmentsToProcess != null) {
+                  ourSegments =  IntSets.mutableFrom(cacheTopology.getLocalReadSegments());
```
wburns: Actually this is broken, I didn't realize that this was changed. Unfortunately we _have_ to use the primary owned segments here. Otherwise it retries too many segments - duplicating data. Unfortunately I don't see an equivalent method on the `LocalizedCacheTopology` to get the primary owned segments.

wburns: It looks like I can just change it back to `ch.getSegmentsForOwner(localAddress)`

wburns: I created https://issues.jboss.org/browse/ISPN-9542 to fix this and look at the remaining issues.

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
wburns: Actually now that I think about this more, is INITIALIZING an okay state? Could the rehash still be in progress?

core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
danberindei: Oops, I was too happy to save an allocation to notice that I had changed the results :(

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
danberindei: Yes, a rebalance could be in progress, but a rebalance can also be in progress in the `RUNNING` state so TBH I'm not sure what this is guarding against.

wburns: I found in testing that when a node is shutting down it can clear the data container yet still return a response (this led to commands that have a subset of data). The more I think about it, the real fix is probably to ensure that a node can't send a command response before clearing the data container on shut down.

danberindei: This should take care of not sending successful responses after clearing the data container: https://github.com/infinispan/infinispan/blob/a72f53962a68f7e38fc8a472144c878be35f3d46/core/src/main/java/org/infinispan/remoting/inboundhandler/BaseBlockingRunnable.java#L118

True, it checks the `PerCacheInboundInvocationHandler` status, not the cache status, but the data container clear should be the last thing to happen during shutdown. Maybe in your failed test the data was in a store, and `PersistenceManagerImpl.stop()` is running before `PerCacheInboundInvocationHandler.stop()`?

wburns: The test is in memory only, I can try running the test without these lines. However there is another bug somewhere I am trying to track down first. I will definitely try this out though.

danberindei: Ok, should I open a PR with just the fix for `getLocalReadSegments()`?

wburns: I created https://issues.jboss.org/browse/ISPN-9542 already which has the fix. I also added an optimization for REPL so it doesn't retry due to rehash anymore. :)

There is still a bug in DIST I can't figure out yet, been trying to isolate it.

wburns: So it fails with these lines commented out every time. The issue isn't that the cache is stopped, but rather that it is running, accepts the command and then stops while in the middle of processing it.

wburns: And the reason it doesn't use the `BaseBlockingRunnable` check is because it sends its own response manually. https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java#L406

PR #6276, opened 2018-09-26
ISPN-9543 Unwrap caches before obtaining stats

tristantarrant:
https://issues.jboss.org/browse/ISPN-9543


wburns:

PR #6277, opened 2018-09-26
ISPN-9542 DistributedStreamRehashStressTest fails many of the tests

wburns:
* Use copy for listener
* Add optimization for REPL to not retry segments

https://issues.jboss.org/browse/ISPN-9542

### DIFF
--- a/core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
+++ b/core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
@@ -287,10 +287,10 @@ public abstract class AbstractCacheStream<Original, T, S extends BaseStream<T, S
 
                IntSet ourSegments;
                if (segmentsToProcess != null) {
-                  ourSegments =  IntSets.mutableFrom(cacheTopology.getLocalReadSegments());
+                  ourSegments =  IntSets.mutableFrom(ch.getPrimarySegmentsForOwner(localAddress));
                   ourSegments.retainAll(segmentsToProcess);
                } else {
-                  ourSegments = IntSets.from(cacheTopology.getLocalReadSegments());
+                  ourSegments = IntSets.from(ch.getPrimarySegmentsForOwner(localAddress));
                }
                // TODO: we can do this more efficiently - since we drop all results locally
                LocalizedCacheTopology newCacheTopology = dm.getCacheTopology();
--- a/core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
+++ b/core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
@@ -19,6 +19,7 @@ import org.infinispan.commons.util.CloseableIterator;
 import org.infinispan.commons.util.CollectionFactory;
 import org.infinispan.commons.util.IntSet;
 import org.infinispan.commons.util.IntSets;
+import org.infinispan.configuration.cache.ClusteringConfiguration;
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.context.Flag;
 import org.infinispan.distribution.DistributionManager;
@@ -59,6 +60,7 @@ public class LocalStreamManagerImpl<Original, K, V> implements LocalStreamManage
    @Inject private CommandsFactory factory;
    @Inject private IteratorHandler iteratorHandler;
    private boolean hasLoader;
+   private boolean isReplicated;
    private int maxSegment;
 
    private Address localAddress;
@@ -72,7 +74,8 @@ public class LocalStreamManagerImpl<Original, K, V> implements LocalStreamManage
       private final IntSet segmentsLost;
 
       SegmentListener(IntSet segments, SegmentAwareOperation op) {
-         this.segments = segments;
+         // Need to make a copy since segments may be modified by user after
+         this.segments = IntSets.mutableCopyFrom(segments);
          this.op = op;
          this.segmentsLost = IntSets.concurrentSet(maxSegment);
       }
@@ -129,9 +132,14 @@ public class LocalStreamManagerImpl<Original, K, V> implements LocalStreamManage
    public void start() {
       cacheName = ByteString.fromString(cache.getName());
       hasLoader = cache.getCacheConfiguration().persistence().usingStores();
-      this.maxSegment = cache.getCacheConfiguration().clustering().hash().numSegments();
+      ClusteringConfiguration clusteringConfiguration = cache.getCacheConfiguration().clustering();
+      this.maxSegment = clusteringConfiguration.hash().numSegments();
       localAddress = rpc.getAddress();
-      cache.addListener(this);
+      isReplicated = clusteringConfiguration.cacheMode().isReplicated();
+      // Replicated never removes segments so we don't have to worry about rehash event
+      if (isReplicated) {
+         cache.addListener(this);
+      }
    }
 
    /**
@@ -214,6 +222,17 @@ public class LocalStreamManagerImpl<Original, K, V> implements LocalStreamManage
    private Stream<Original> getRehashStream(CacheSet<Original> cacheEntrySet, Object requestId,
          SegmentListener listener, boolean parallelStream, boolean entryStream, IntSet segments,
          Set<K> keysToInclude, Set<K> keysToExclude) {
+      if (listener != null) {
+         handleSuspectSegmentsBeforeStream(requestId, listener, segments);
+      }
+
+      return getStream(cacheEntrySet, parallelStream, entryStream, segments, keysToInclude, keysToExclude);
+   }
+
+   // We should first handle any segments we already lost before processing. Remove from {@code segments} all the
+   // segments for which this node is no longer an owner add them to {@code listener.segmentsLost}. Note this means
+   // that we allow backup owners to respond even though it was thought to be a primary owner.
+   private void handleSuspectSegmentsBeforeStream(Object requestId, SegmentListener listener, IntSet segments) {
       LocalizedCacheTopology topology = dm.getCacheTopology();
       if (trace) {
          log.tracef("Topology for supplier is %s for id %s", topology, requestId);
@@ -227,7 +246,7 @@ public class LocalStreamManagerImpl<Original, K, V> implements LocalStreamManage
             int segment = iterator.nextInt();
             // If the segment is not owned by both CHs we can't use it during rehash
             if (!pendingCH.isSegmentLocalToNode(localAddress, segment)
-                    || !readCH.isSegmentLocalToNode(localAddress, segment)) {
+                  || !readCH.isSegmentLocalToNode(localAddress, segment)) {
                iterator.remove();
                lostSegments.set(segment);
             }
@@ -251,8 +270,6 @@ public class LocalStreamManagerImpl<Original, K, V> implements LocalStreamManage
             log.tracef("Hash %s for id %s", readCH, requestId);
          }
       }
-
-      return getStream(cacheEntrySet, parallelStream, entryStream, segments, keysToInclude, keysToExclude);
    }
 
    private void handleResponseError(CompletionStage<ValidResponse> rpcFuture, Object requestId,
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -214,41 +222,44 @@ public void dataRehashed(DataRehashedEvent<K, V> event) {
...
-         while (iterator.hasNext()) {
-            int segment = iterator.nextInt();
-            // If the segment is not owned by both CHs we can't use it during rehash
-            if (!pendingCH.isSegmentLocalToNode(localAddress, segment)
-                    || !readCH.isSegmentLocalToNode(localAddress, segment)) {
-               iterator.remove();
-               lostSegments.set(segment);
-            }
+      // Don't even bother checking topology if listener is null - this means REPL
+      if (listener != null) {
```
danberindei: This method is getting kind of big, the listener part could be a separate method.

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -298,41 +309,51 @@ private void handleResponseError(CompletionStage<ValidResponse> rpcFuture, Objec
...
+               log.tracef("UnRegistered change listener for %s", requestId);
+            }
          }
       }
       if (cache.getStatus() != ComponentStatus.RUNNING && cache.getStatus() != ComponentStatus.INITIALIZING) {
          if (trace) {
             log.tracef("Cache status is no longer running, all segments are now suspect for %s", requestId);
          }
-         listener.segmentsLost.addAll(segments);
+         lostSegments = segments;
```
danberindei: IMO we shouldn't send a `StreamResponseCommand` at all if the cache is shutting down.

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -417,8 +450,8 @@ public IteratorResponse startIterator(Object requestId, Address origin, IntSet s
       }
       CacheSet<Original> originalSet = toOriginalSet(includeLoader, entryStream);
 
-      // Have to make a copy of segments as we will remove from it below
-      SegmentListener listener = new SegmentListener(IntSets.mutableCopyFrom(segments), i -> true);
+      // Unfortunately we have to ALWAYS use a listener - due to storing state between requests in the listener
```
danberindei: IMO this isn't very helpful for readers, because it answers one question that I didn't think of asking (why do we use a listener here) and asks more questions for which I now need to interrupt my flow and find the answers somewhere else:

1. Why wouldn't we use a listener?
2. What kind of data do we store in the listener?
3. Can multiple iteration requests share a listener?

core/src/test/java/org/infinispan/stream/stress/DistributedStreamRehashStressTest.java
```diff
@@ -78,15 +76,16 @@ protected EmbeddedCacheManager addClusterEnabledCacheManager(TransportFlags flag
 
    public void testStressNodesLeavingWhileMultipleCollectors() throws Throwable {
       testStressNodesLeavingWhilePerformingCallable((cache, masterValues, iteration) -> {
+         SerializablePredicate<Map.Entry<Integer, Integer>> predicate = e -> (e.getKey() & 1) == 1;
          Map<Integer, Integer> results = cache.entrySet().stream().filter(
-                 (Serializable & Predicate<Map.Entry<Integer, Integer>>)
-                         e -> (e.getKey() & 1) == 1).collect(
+               predicate).collect(
```
danberindei: Formatting is a bit weird now

core/src/main/java/org/infinispan/stream/impl/AbstractCacheStream.java
```diff
@@ -287,10 +287,10 @@ public void close() {
 
                IntSet ourSegments;
                if (segmentsToProcess != null) {
-                  ourSegments =  IntSets.mutableFrom(cacheTopology.getLocalReadSegments());
+                  ourSegments =  IntSets.mutableFrom(ch.getPrimarySegmentsForOwner(localAddress));
```
danberindei: Sorry about this, thanks for finding the problem so quickly!

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -129,9 +132,14 @@ public void inject(Cache<K, V> cache) {
    public void start() {
       cacheName = ByteString.fromString(cache.getName());
       hasLoader = cache.getCacheConfiguration().persistence().usingStores();
-      this.maxSegment = cache.getCacheConfiguration().clustering().hash().numSegments();
+      ClusteringConfiguration clusteringConfiguration = cache.getCacheConfiguration().clustering();
+      this.maxSegment = clusteringConfiguration.hash().numSegments();
       localAddress = rpc.getAddress();
-      cache.addListener(this);
+      // Replicated never removes segments so we don't have to worry about rehash event
+      requiresListener = !clusteringConfiguration.cacheMode().isReplicated();
```
danberindei: I think using a `isReplicated` field might actually be clearer, since in some cases we still use a listener.

wburns: Sure, sounds good.

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
wburns: Hrmm, I will look into this. As long as we don't send a response to the command, this should be okay as I would expect to get a `SuspectException` at https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/stream/impl/ClusterStreamManagerImpl.java#L210 Or maybe it is `CacheNotFoundException`? Either way I can look into it if it all fails.

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
wburns: Sadly this another one of those shut down things. I can try to explain this one a bit better in comments. It is because of the lazily marshalled approach it takes when sending a response.

core/src/test/java/org/infinispan/stream/stress/DistributedStreamRehashStressTest.java
wburns: Oh whoops, not sure how that happened :)

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
wburns: Okay

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
danberindei: `SingleResponseCollector.validOnly()` throws a `SuspectException`. `CacheNotFoundException` is a server thing :)

wburns: Alright :) Actually give me a couple more hours to look at this again, I want to verify something.

wburns: After sleeping on it I realized we can't just remove the response. This will cause a timeout for the caller, since the response is completed based on either it sending the `StreamResponseCommand` or the `StreamRequestCommand` returning a `SuspectException`. If we just return normally it won't get a `SuspectException` so we have to send this command for now. We could revisit changing this, but it requires more changes than I would like to introduce this late in the cycle.

wburns: Have to look a little further, doesn't seem to be the case.

wburns: Yeah so it seems to be a timing issue if I take it out. Sometimes it gets the `SuspectException` and sometimes not. I added in some trace messages to see when it happens in a test:
```
09:55:19,454 FATAL (remote-thread-DistributedStreamRehashStressTest-NodeAO-p322-t11) [LocalStreamManagerImpl] Cache status is no longer running, all segments are now suspect for DistributedStreamRehashStressTest-NodeA-43398#371
09:55:29,080 FATAL (remote-thread-DistributedStreamRehashStressTest-NodeAT-p362-t10) [LocalStreamManagerImpl] Cache status is no longer running, all segments are now suspect for DistributedStreamRehashStressTest-NodeA-43398#422
09:55:39,083 FATAL (ForkThread-1,DistributedStreamRehashStressTest) [DistributedCacheStream] Timeout encountered with id: DistributedStreamRehashStressTest-NodeA-43398#422
```

So as you can see the first time was okay, more than likely because it got the `SuspectException` back before it could return the empty response.

wburns: This needs me to change it to send a response instead of a `StreamResponseCommand` which is a bit much for this PR imo.

wburns:
Updated and should be ready again now.

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -59,6 +60,9 @@
    @Inject private CommandsFactory factory;
    @Inject private IteratorHandler iteratorHandler;
    private boolean hasLoader;
+   // This variable is true when dealing with a replicated cache. In this case we don't need to register a segment
+   // listener, since replicated caches never generates data rehash events
+   private boolean isReplicated;
```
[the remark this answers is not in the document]
wburns: Yes, sorry, I worded it incorrectly. I meant that entries aren't removed due to a data rehash event.

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -214,6 +224,22 @@ public void dataRehashed(DataRehashedEvent<K, V> event) {
...
+      return getStream(cacheEntrySet, parallelStream, entryStream, segments, keysToInclude, keysToExclude);
+   }
+
+   /**
+    * We should first handle any segments we already lost since they were requested - this check should happen before
+    * performing the desired operation. The segments may be modified to remove suspect segments and those same
+    * segments will be added to the listeners suspect segments.
+    * @param requestId the request identifier for the operation
+    * @param listener the listener that handles carrying lost segments
+    * @param segments segments that we want to operate upon
```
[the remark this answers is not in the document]
wburns: Ah yes this was because originally I wanted to add {@link} inside the comments, but ended up moving it. I can reduce it to standard // comments.

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -214,6 +224,22 @@ public void dataRehashed(DataRehashedEvent<K, V> event) {
...
+         handleSuspectSegmentsBeforeStream(requestId, listener, segments);
+      }
+
+      return getStream(cacheEntrySet, parallelStream, entryStream, segments, keysToInclude, keysToExclude);
+   }
+
+   /**
+    * We should first handle any segments we already lost since they were requested - this check should happen before
+    * performing the desired operation. The segments may be modified to remove suspect segments and those same
+    * segments will be added to the listeners suspect segments.
```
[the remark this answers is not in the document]
wburns: Sure, but I have to change the wording as it isn't exactly true. We only suspect a segment if it is not owned anymore. So we send the request to primary owners, but allow backup owners to respond (if they changed during that period).

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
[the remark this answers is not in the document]
wburns: I will just remove it anyways.

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
```diff
@@ -417,8 +462,25 @@ public IteratorResponse startIterator(Object requestId, Address origin, IntSet s
...
-      // Have to make a copy of segments as we will remove from it below
-      SegmentListener listener = new SegmentListener(IntSets.mutableCopyFrom(segments), i -> true);
+      // Normally we wouldn't use a listener a here when using replicated cache. Unfortunately due to the way that
+      // lazily send results for iterator, we need to retain this listener. This listener's sole purpose is for
+      // communicating lost segments in a response.
+      // If you look at IteratorResponses$IteratorResponsesExternalizer#writeObject method it will send the
+      // completed segments and suspected segments when the iterator is completely exhausted of elements. When
+      // this exhaustion occurs it also closes the iterator. We hook into the iterator closing via the Runnable just
+      // below and if the cache is not in a running state we suspect all of the segments which will then be
+      // written to the response
```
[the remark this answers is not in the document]
wburns: okay, let me work some magic that you will probably want changed ;)

wburns:
Updated comments, I just sided on the err of removing.

core/src/main/java/org/infinispan/stream/impl/LocalStreamManagerImpl.java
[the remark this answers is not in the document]
danberindei: Nope, that's exactly what I wanted :+1:

PR #6122, opened 2018-07-04
HHH-12531 Handle legacy names in configuration

rvansa:
https://hibernate.atlassian.net//browse/HHH-12531

PR #6282, opened 2018-09-27
HHH-12531 Handle legacy names in configuration

tristantarrant:
https://hibernate.atlassian.net//browse/HHH-12531

Replaces https://github.com/infinispan/infinispan/pull/6122


tristantarrant:
Also aligns Hibernate dependency with WildFly 14

PR #6283, opened 2018-09-27
ISPN-9552 Remove duplicated brand info

pruivo:
Branding info is already in build-configuration/pom.xml

https://issues.jboss.org/browse/ISPN-9552


tristantarrant:
Merged, thanks

PR #6286, opened 2018-09-27
ISPN-9559 Server <modules> configuration with a single entry can break classloading

anistor:
See message on commit

https://issues.jboss.org/browse/ISPN-9559

### DIFF
--- a/query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
+++ b/query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
@@ -6,7 +6,6 @@ import static org.infinispan.query.impl.IndexPropertyInspector.getMetadataCacheN
 import static org.infinispan.query.impl.IndexPropertyInspector.hasInfinispanDirectory;
 
 import java.util.Collection;
-import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Properties;
@@ -94,18 +93,14 @@ import org.infinispan.util.logging.LogFactory;
 import org.kohsuke.MetaInfServices;
 
 /**
- * Lifecycle of the Query module: initializes the Hibernate Search engine and shuts it down
- * at cache stop.
+ * Lifecycle of the Query module: initializes the Hibernate Search engine and shuts it down at cache stop.
  *
- * @author Sanne Grinovero <sanne@hibernate.org> (C) 2011 Red Hat Inc.
+ * @author Sanne Grinovero &lt;sanne@hibernate.org&gt; (C) 2011 Red Hat Inc.
  */
 @MetaInfServices(org.infinispan.lifecycle.ModuleLifecycle.class)
-@SuppressWarnings("unused")
 public class LifecycleManager implements ModuleLifecycle {
 
    private static final Log log = LogFactory.getLog(LifecycleManager.class, Log.class);
-
-   private static final Object REMOVED_REGISTRY_COMPONENT = new Object();
 
    private MBeanServer mbeanServer;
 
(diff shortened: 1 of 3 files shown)


query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
```diff
@@ -311,25 +305,23 @@ private boolean verifyChainContainsQueryInterceptor(ComponentRegistry cr) {
...
-         programmaticSearchMappingProviders.add(new DefaultSearchMappingProvider());
-         Collection<LuceneAnalysisDefinitionProvider> analyzerDefProviders = ServiceFinder.load(LuceneAnalysisDefinitionProvider.class, aggregatedClassLoader);
-         // Set up the search factory for Hibernate Search first.
-         SearchConfiguration config = new SearchableCacheConfiguration(indexingConfiguration.indexedEntities(),
-               indexingConfiguration.properties(), programmaticSearchMappingProviders, analyzerDefProviders, cr, aggregatedClassLoader);
-
-         searchFactory = new SearchIntegratorBuilder().configuration(config).buildSearchIntegrator();
-         cr.registerComponent(searchFactory, SearchIntegrator.class);
+   private SearchIntegrator createSearchIntegrator(IndexingConfiguration indexingConfiguration, ComponentRegistry cr, ClassLoader aggregatedClassLoader) {
+      // prevent multiple init
```
danberindei: This shouldn't be necessary either, `cacheStarting()` only runs once for a cache.

I think this pattern first appeared when query added the interceptors to the current chain *and* to the interceptors configuration, meaning the interceptor would already be registered after a restart. But even in that case it would be simpler if we always registered the interceptors the same way, without modifying the configuration.

query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
```diff
@@ -121,11 +117,12 @@ public void cacheStarting(ComponentRegistry cr, Configuration cfg, String cacheN
          boolean isIndexed = cfg.indexing().index().isEnabled();
          AdvancedCache<?, ?> cache = cr.getComponent(Cache.class).getAdvancedCache();
 
+         ClassLoader aggregatedClassLoader = makeAggregatedClassLoader(cr.getGlobalComponentRegistry().getGlobalConfiguration().classLoader());
```
danberindei: I've been looking at the `makeAggregatedClassLoader()` implementation, and I really think we should stop adding the TCCL (we already use the TCCL as the default configuration classloader) and the system classloader (all the others delegate to it anyway).

query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
anistor: Nope. That pattern appeared because it registers the SearchIntegrator on starting but has no way to unregister it on stopping. But it does ```close``` the SearchIntegrator. After restart it needs a way to detect that SearchIntegrator it find in CR is garbage and throw it away, so a dummy Object is used to replace an existing component, as marker for a stopped SearchIntegrator.  This dirty trick is no longer needed because SearchIntegrator.isStopped was introduced at some later point in time.

The ```// prevent multiple init``` comment is there also because this method previously called ```getSearchFactory``` used to be called multiple times in older code. Now it's only one invocation and I have renamed it. I'll remove the comment.

anistor: Actually, the whole thing boils down to the fact ```getSearchFactory``` is no longer invoked multiple times, so no danger of double init appears. All that code will be gone, no just the comment. I'm so happy we remove that crap.

query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
anistor: Does that hurt or it is just something extra that we can remove?

query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
anistor: actually, I think I'll keep the ```throw new IllegalStateException("SearchIntegrator already initialized!");``` for a while. I don't trust anything and anybody these days :)

anistor:
Rebased an updated.

anistor:
Rebased and updated again.

anistor:
Rebased and updated. Also removed all unrelated commits that I pushed on top of this to piggyback on CI while testing 2 other unrelated query issues :)

server/integration/infinispan/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerConfigurationBuilder.java
```diff
@@ -290,15 +290,16 @@ private ClassLoader makeGlobalClassLoader(ModuleLoader moduleLoader, ModuleIdent
...
+        ClassLoader infinispanSubsystemClassloader = CacheContainerConfiguration.class.getClassLoader();
+        if (classLoaders.isEmpty()) {
+            // use our default CL if nothing was specified by user
+            return infinispanSubsystemClassloader;
         }
+        if (cacheContainerModule == null) {
+            // do not use the infinispan subsystem classloader if the user specifically requested a different one
+            classLoaders.add(infinispanSubsystemClassloader);
+        }
+        return new AggregatedClassLoader(classLoaders);
```
gustavonalle: When running the server, ```AggregatedClassLoader``` will always be created, since ```cacheContainerModule``` will always be ```org.infinispan.extension:ispn-9.4```. Should we return instead ```moduleLoader.loadModule(cacheContainerModule).getClassLoader()``` when there is no ```additionalModules```?

anistor: if ```classLoaders``` has just one element we could return that instead of making an AggregatedClassLoader

anistor: done

PR #6287, opened 2018-09-27
ISPN-9556 The netty server module depends on the log4j and sun.jdk mo…

tristantarrant:
https://issues.jboss.org/browse/ISPN-9556

### DIFF
--- a/feature-pack/client/src/main/resources/modules/io/netty/slot/module.xml
+++ b/feature-pack/client/src/main/resources/modules/io/netty/slot/module.xml
@@ -17,5 +17,7 @@
     <dependencies>
         <module name="javax.api"/>
         <module name="org.javassist" optional="true"/>
+        <module name="org.apache.log4j"/>
+        <module name="sun.jdk"/>
     </dependencies>
 </module>


feature-pack/client/src/main/resources/modules/io/netty/slot/module.xml
```diff
@@ -17,5 +17,7 @@
     <dependencies>
         <module name="javax.api"/>
         <module name="org.javassist" optional="true"/>
+        <module name="org.apache.log4j"/>
```
danberindei: Isn't this module log4j 1.x? I think Netty uses log4j 2.x.

tristantarrant: WildFly comes with neither. It has a JBoss Logging shim that pretends to be log4j [12]

danberindei: Ah ok

danberindei:
Integrated, thanks Tristan

PR #6290, opened 2018-09-28
ISPN-9558 Export Data Grid stats to Prometheus

vblagoje:
Preview of Prometheus integration. *Do not* integrate yet. Need a review from @slaskawi @ryanemerson and @tristantarrant

### DIFF
--- a/server/integration/feature-pack/src/main/resources/content/bin/prometheus.sh
+++ b/server/integration/feature-pack/src/main/resources/content/bin/prometheus.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+if [ -z "$JMX_CONF" ]
+then
+   if [ -z "$1" ]
+   then
+       echo "Neither JMX_CONF nor input parameters are defined. Exiting"
+       exit 1
+   else
+       JMX_CONF=$1
+   fi
+fi
+
+PROMETHEUS_JAR=$(ls $JBOSS_HOME/modules/system/add-ons/@infinispan.module.slot.prefix@/io/prometheus/jmx/@infinispan.module.slot@/*.jar | sed -e "s| |:|g")
+JBOSS_LOG_MANAGER_LIB=$(ls $JBOSS_HOME/modules/system/layers/base/org/jboss/logmanager/main/*.jar | sed -e "s| |:|g")
+WILDFLY_COMMON_LIB=$(ls $JBOSS_HOME/modules/system/layers/base/org/wildfly/common/main/*.jar | sed -e "s| |:|g")
+export JAVA_OPTS="$JAVA_OPTS -Djava.util.logging.manager=org.jboss.logmanager.LogManager -Xbootclasspath/p:$JBOSS_LOG_MANAGER_LIB -Xbootclasspath/p:$WILDFLY_COMMON_LIB -javaagent:$PROMETHEUS_JAR=$JMX_CONF"
--- a/server/integration/feature-pack/src/main/resources/content/bin/standalone.sh
+++ b/server/integration/feature-pack/src/main/resources/content/bin/standalone.sh
@@ -117,7 +117,7 @@ if [ -r "$RUN_CONF" ]; then
 fi
 
 if [ "x$JMX_CONF" = "x" ]; then
-    JMX_CONF="$DIRNAME/agent.conf"
+    JMX_CONF="9779:$DIRNAME/prometheus_config.yaml"
 fi
 
 # Set debug settings if not already set
@@ -333,7 +333,7 @@ if [ "$SECMGR" = "true" ]; then
 fi
 
 if [ "x$USE_JMX_COLLECTORS" != "x" ]; then
-    . $DIRNAME/jolokia.sh $JMX_CONF
+    . $DIRNAME/prometheus.sh $JMX_CONF
 fi
 
 
--- a/build-configuration/bom/pom.xml
+++ b/build-configuration/bom/pom.xml
@@ -183,6 +183,7 @@
         <versionx.io.dropwizard.metrics.metrics-json>3.1.5</versionx.io.dropwizard.metrics.metrics-json>
         <versionx.io.dropwizard.metrics.metrics-jvm>3.1.0</versionx.io.dropwizard.metrics.metrics-jvm>
         <versionx.io.fabric8.agent-bond-agent>1.0.2</versionx.io.fabric8.agent-bond-agent>
+        <versionx.io.prometheus.jmx.jmx_prometheus_javaagent>0.3.1</versionx.io.prometheus.jmx.jmx_prometheus_javaagent>
         <versionx.io.netty.netty>${version.netty}</versionx.io.netty.netty>
         <versionx.io.netty.netty-all>${version.netty}</versionx.io.netty.netty-all>
         <versionx.io.netty.netty-bom>${version.netty}</versionx.io.netty.netty-bom>
@@ -1529,6 +1530,11 @@
                 <groupId>io.fabric8</groupId>
                 <artifactId>agent-bond-agent</artifactId>
                 <version>${versionx.io.fabric8.agent-bond-agent}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.prometheus.jmx</groupId>
+                <artifactId>jmx_prometheus_javaagent</artifactId>
+                <version>${versionx.io.prometheus.jmx.jmx_prometheus_javaagent}</version>
             </dependency>
             <dependency>
                 <groupId>io.netty</groupId>
--- a/server/integration/feature-pack/pom.xml
+++ b/server/integration/feature-pack/pom.xml
@@ -126,6 +126,11 @@
         </dependency>
 
         <dependency>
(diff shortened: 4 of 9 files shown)


server/integration/feature-pack/src/main/resources/content/bin/standalone.sh
```diff
@@ -333,7 +333,7 @@ if [ "$SECMGR" = "true" ]; then
 fi
 
 if [ "x$USE_JMX_COLLECTORS" != "x" ]; then
-    . $DIRNAME/jolokia.sh $JMX_CONF
+    . $DIRNAME/prometheus.sh $JMX_CONF
```
tristantarrant: You forgot to `git add prometheus.sh` somewhere

vblagoje: Indeed. Thanks @tristantarrant

vblagoje:
@tristantarrant @ryanemerson this should be approximately what we want. We'll turn on prometheus, turn of jolokia. We can remove jolokia later. We should integrate this one post-release?

You can try it out by running "./standalone.sh --jmx 9779:prometheus_config.yaml" from the bin directory

ryanemerson:
@vblagoje If this PR is being merged post-release, then I would say Jolokia should be removed in a separate commit as part of this PR.

vblagoje:
> I'd merge it now. Leave jolokia in there, and remove it post release

Ok @tristantarrant , but that makes Prometheus agent of choice when running --jmx switch. Let me test everything a bit more if this is going to a release. Do we have anyone on Windows machine to test the bat scripts?

tristantarrant:
Vittorio has a Windows installation

vblagoje:
@rigazilla would you please load up this PR tomorrow and boot your server with both:
./bin/standalone.bat 
and
./bin/standalone.bat --jmx 9779:bin/prometheus_config.yaml
Both should work. In the second instance hit http://localhost:9779/metrics and see if you get a response. Thank you

vblagoje:
@ryanemerson and @tristantarrant so we have the current standalone (that is agent.conf) script assuming that agent bond is being used. We can still boot server without jmx switch but wehn jmx switch is used it has to be in the form of:
./standalone.sh --jmx 9779:prometheus_config.yaml 

After we integrate this PR and post-release. I'll remove Jolokia references, agent bond jar, agent.conf and simplify things quite a bit.

rigazilla:
@vblagoje running bin/standalone.bat I get this exception:
https://gist.github.com/rigazilla/afaeed09214786092d1b6c3877725e1d

vblagoje:
@rigazilla one more time please, refresh the branch

rigazilla:
@vblagoje 
./bin/standalone.bat
looks OK


./bin/standalone.bat --jmx 9779:bin/prometheus_config.yaml fails as below:
https://gist.github.com/rigazilla/bb263454b5a10c2809f0c5eedb990990

vblagoje:
@rigazilla could you please fiddle with prometheus.bat script a bit to make it work? It is not picking up the libraries it should for the server command line

vblagoje:
@tristantarrant let's not integrate this one yet. We'll integrate it in first post 9.4.0.Final release

vblagoje:
Rebased, and ready for integration. Please check the build yourself @tristantarrant and @ryanemerson 
The following use cases should be tested:
./standalone.sh (normal boot)
./standalone.sh --jmx (default Prometheus boot)
./standalone.sh --jmx 9779:prometheus_config.yaml (default Prometheus boot)

@rigazilla already checked all three Windows boot version and has helped me tremendously to get this one out of the door

vblagoje:
@tristantarrant @ryanemerson can you give some love to this PR today? Please do try to run in it all above three configurations. See ^^

ryanemerson:
@vblagoje I ran a build earlier today and it was still resulting in a OOM error even after @tristantarrant fix was integrated. I've scheduled another build, let's see what happens.

vblagoje:
Ok @ryanemerson can you run it locally to make sure everything is ok as well? I know you are very diligent and I look forward to your reviews as they give me a piece of mind 👍

server/integration/feature-pack/src/main/resources/content/bin/prometheus.sh
```diff
@@ -0,0 +1,17 @@
...
+       echo "Neither JMX_CONF nor input parameters are defined. Exiting"
+       exit 1
+   else
+       JMX_CONF=$1
+   fi
+fi
+
+PROMETHEUS_JAR=$(ls $JBOSS_HOME/modules/system/add-ons/ispn/io/prometheus/jmx/ispn-9.4/*.jar | sed -e "s| |:|g")
+JBOSS_LOG_MANAGER_LIB=$(ls $JBOSS_HOME/modules/system/layers/base/org/jboss/logmanager/main/*.jar | sed -e "s| |:|g")
+WILDFLY_COMMON_LIB=$(ls $JBOSS_HOME/modules/system/layers/base/org/wildfly/common/main/*.jar | sed -e "s| |:|g")
```
ryanemerson: You cannot hardcode `ispn-9.4` anywhere as the code will break in 10.0. You need to use `@infinispan.module.slot.prefix@`

See https://github.com/infinispan/infinispan/commit/0756eea115e6d42a2d8a6e7eae85d110578995ec for an example

ryanemerson: Using `main` for logmanager and wildlfy-common should be fine as they are provided as part of the WFLY distribution.

server/integration/feature-pack/src/main/resources/content/bin/prometheus.sh
```diff
@@ -0,0 +1,17 @@
...
+   if [ -z "$1" ]
+   then
+       echo "Neither JMX_CONF nor input parameters are defined. Exiting"
+       exit 1
+   else
+       JMX_CONF=$1
+   fi
+fi
+
+PROMETHEUS_JAR=$(ls $JBOSS_HOME/modules/system/add-ons/ispn/io/prometheus/jmx/ispn-9.4/*.jar | sed -e "s| |:|g")
```
ryanemerson: You cannot use `*.jar` in these commands as it breaks when utilising the "thin" server. The thin server does not store the jar files in the slot directory, instead it pulls them from the local maven repository.

An alternative solution is todo something like:
```
PROMETHEUS_JAR=jmx_prometheus_javaagent-@prometheus_version@.jar
```
with the value of `@prometheus_version@` set in `server/integration/feature-pack/build.xml`.

To test future fixes, make sure you try running the thin server at: `server/integration/build/target/server-distribution-thin/infinispan-server-9.4.1-SNAPSHOT`

vblagoje: @ryanemerson ok, IIRC correctly Tristan and I talked about being able to run the server in the fat mode with Prometheus...but sure, let's make it work in the thin as well.

server/integration/feature-pack/src/main/resources/content/bin/prometheus.sh
vblagoje: I did change ispn-9.4 to @infinispan.module.slot.prefix@ but then it resurfaced again :-(

server/integration/feature-pack/src/main/resources/content/bin/prometheus.sh
vblagoje: But @ryanemerson I won't be able to pull up two other jars (logmanager, common) in thin server  either. I need these three jars. But thin server will work in non jmx mode which is ok. @tristantarrant said we run thin server for tests only and that fat version matters. Thoughts?

ryanemerson: Well we distribute the thin server via maven, so I would say it should work.

Another solution, which is probably the way to go, is to not use placeholders. Instead process the `module.xml` of the three modules in order to retrieve the jar name. It is specified in the following format:

`<artifact name="org.wildfly.common:wildfly-common:1.4.0.Final"/>```

ryanemerson: Sorry, just realised that still doesn't solve the problem of getting the absolute path to the jar

vblagoje: Yeah, I think Tristan thought this through as well and concluded that we can only run JMX in fat server mode...unless there is some way we can reference these jars

ryanemerson: Yeah fair enough, fat server is sufficient for OS which is the primary target anyway.

vblagoje: Ok, then the last commit fix should be sufficient. Please check @ryanemerson

vblagoje:
@ryanemerson rebased and squashed. Let's see what CI says and once again please try it out yourself before integrating.

ryanemerson:
@vblagoje Sorry to ask for another change, but I noticed that the port is hardcoded ... we really should make this configurable

PR #6294, opened 2018-10-02
ISPN-9568 Doc attributes and clean up

oraNod:
- Updating several instances of hard-coded "Infinispan" that should use the {brandname} variable. Also replaced the variable with "Infinispan" in some sections where it only makes sense to talk about upstream versions.
- Including attributes for product Javadoc and Config URLs to match Infinispan variables.
- Resolving the build warning for the asciimath syntax that RH doc tooling does not support.
- Fixing broken links and listed items that do not terminate properly and nest subsequent sections.

This is the Jira ticket: https://issues.jboss.org/browse/ISPN-9568


documentation/src/main/asciidoc/getting_started/example_with_groovy.adoc
```diff
@@ -49,7 +49,13 @@ Groovy Version: 1.6.3 JVM: 1.6.0_14
 If you get a similar result as shown, everything went well.
 
 ==== Installing {brandname}
-Now you should add the {brandname} libraries to the Groovy Platform so you will able to access the API from the Groovy console. Add the infinispan-core.jar and its dependencies to the $USER_HOME/.groovy/lib directory, the jar is located in $INFINISPAN_HOME/modules/core and the dependencies at $INIFINISPAN_HOME/modules/core/lib.
+Now you should add the {brandname} libraries to the Groovy Platform so you will able to access the API from the Groovy console. Add the infinispan-core.jar and its dependencies to the $USER_HOME/.groovy/lib directory, the jar is located in
+ifndef::productized[]
+$INFINISPAN_HOME/modules/core and the dependencies at $INFINISPAN_HOME/modules/core/lib.
```
ryanemerson: I'm not familiar with the syntax, but is it not possible to have an else clause?

documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -39,6 +39,9 @@ nodes; reads are local.
 * *Distributed*: changes are replicated to a fixed number of nodes, reads request the
 value from at least one of the owner nodes.
 
+//dnaro: filtering this table from product docs. RH doc tooling does not
+//support asciimath or text coloring. Including this table throws a warning.
+ifndef::productized[]
```
ryanemerson: What's the plan for including this in the product in the future instead of just ignoring? It's a useful table that would be good to include.

oraNod: Hey Ryan, agreed that the table is useful. It contains an equation for calculating the cache size or something like that, which requires asciimath to render correctly. RH docs tooling doesn't support asciimath though so the equation just renders as asciidoc markup, which doesn't make any sense and negates the benefit for users.
I'm trying to find out if we can get support for asciimath, but probably not. Otherwise I think maybe we can include the equation as an image and re-format the table. This is definitely on my to-do list and I don't want to just ignore it in the docs.

ryanemerson: Sounds good. An image seems like a good backup if need be.

documentation/src/main/asciidoc/getting_started/example_with_groovy.adoc
oraNod: I think the short answer is no. https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/_includes/conditional-preprocessor-directives.adoc
I'm definitely on for more elegant approaches if you have suggestions.

PR #6295, opened 2018-10-02
[experiment] Fix dependencies divergences

pruivo:
Fix divergences detected by `mvn project-info-reports:dependency-convergence`

There are current 4 divergences:

`hibernate-core` infinispan-hibernate-cache-v51 is fecthing old version
`infinispan-server-build` old version is fetched for rolling upgrades
`infinispan-core` for some reason, infinispan-arquillian-impl is fecthing 7.0.0.Alpha4
`mockito-core` spring-boot-it needs an old version

I'm just testing if everything still works


tristantarrant:
+370 −2,569 is nice :)

pruivo:
@tristantarrant yes it is nice. but I'm not sure how reliable `dependency:tree` is. It seems not going very deep and it definitely can read the dependencies from zip imports. 

There are a bunch of deps removed...

PR #6296, opened 2018-10-02
ISPN-9569 Indexing is not working for clustered caches if it is trans…

anistor:
…actional

https://issues.jboss.org/browse/ISPN-9569

__UPDATE__ the operation principle of this fix is: removing ```ProtobufValueWrapperInterceptor``` (I gave up fixing it, it's waaaay too complex to properly intercept all commands and extract the value to be intercepted, just look at the complex duo: QueryInterceptor+TxQueryInterceptor), introduced ```ProtobufValueWrapperSearchWorkCreator``` in ```QueryIntercetptor```, which is a custom ```SearchWorkCreator``` that delegates to the existing one after it intercepts the ```ProtobufValueWrapper``` value.  Thus all ```ProtobufValueWrapper``` values going through this ```SearchWorkCreator``` will have their ```setMessageDescriptor``` called before the values reaches the hibernate search stack. Problem solved! (all fingers crossed)


anistor:
This is WIP. There are still some commands that use default handling in ProtobufValueWrapperInterceptor and ProtobufMetadataManagerInterceptor and should be specifically handled. Another update will come soon.

gustavonalle:
@anistor shouldn't this go to 9.4.0.Final?

anistor:
It should, but so far I'm not happy with the fix. It fixes the test that was added for this but it still does not cover all cases. So I guess the easiest fix for the 9.4.0 final scope is to just remove the offending interceptor. But I'm still looking at this ...

anistor:
Updated. Should be fine now, but I might still remove some of the changes here because are merely cosmetic and not needed for the fix.

anistor:
Updated. Removed the extra stuff.

anistor:
rebased.

anistor:
Green!

anistor:
Rebased again for 9.4.1

PR #6297, opened 2018-10-03
[PREVIEW] ISPN-9541 Module initialization is not thread-safe 

danberindei:
https://issues.jboss.org/browse/ISPN-9541

Do not integrate: Waiting to see the results of the server integration testsuite.


danberindei:
This pull request does too little, we need a more comprehensive solution that compiles a list of components provided by each module before starting any component.

PR #6302, opened 2018-10-03
ISPN-9575 Log a WARN when compat mode is used

gustavonalle:
https://issues.jboss.org/browse/ISPN-9575


danberindei:
Integrated, thanks Gustavo!

PR #6303, opened 2018-10-03
ISPN-9581 Backport downstream fixes to upstream

pruivo:
I think in upstream the `infinispan.brand.version` is always the same as `project.version`. If it is true, I can change the property

I'll squash the commits in the end and open a JIRA

https://issues.jboss.org/browse/ISPN-9581


pruivo:
Fixed.
Squashed commits and added JIRA#

tristantarrant:
Merged, thanks

PR #6304, opened 2018-10-03
ISPN-9576 CustomInterceptorInjectionTest uses deprecated API

anistor:
https://issues.jboss.org/browse/ISPN-9576


anistor:
I''ve rebased and removed a change that was definitely causing a failure. Let's see CI now...

anistor:
Rebased.

core/src/main/java/org/infinispan/configuration/cache/IndexingConfiguration.java
```diff
@@ -20,12 +19,7 @@
    public static final AttributeDefinition<Boolean> AUTO_CONFIG = AttributeDefinition.builder("autoConfig", false).immutable().build();
    public static final AttributeDefinition<Set<Class<?>>> INDEXED_ENTITIES = AttributeDefinition.builder("indexed-entities", null, (Class<Set<Class<?>>>) (Class<?>) Set.class)
          .copier(CollectionAttributeCopier.INSTANCE)
-         .initializer(new AttributeInitializer<Set<Class<?>>>() {
-            @Override
-            public Set<Class<?>> initialize() {
-               return new HashSet<>(5);
-            }
-         }).immutable().build();
+         .initializer(() -> new HashSet<>(5)).immutable().build();
```
danberindei: I think it's better to use the default constructor now, because it doesn't allocate an array at all

core/src/main/java/org/infinispan/interceptors/base/CommandInterceptor.java
```diff
@@ -94,12 +92,6 @@ public final boolean hasNext() {
       return myIndex < interceptors.size();
    }
 
-   /**
-    * Does nothing since 9.0.
-    */
-   public final void setNext(CommandInterceptor ignored) {
```
danberindei: It's public API, I don't think we should delete it in 9.4. Same with `getLockAcquisitionTimeout()` and `getCacheWithFlags()`, there could be user subclasses using them in the wild.

core/src/test/java/org/infinispan/configuration/CustomInterceptorInjectionTest.java
```diff
@@ -16,14 +25,29 @@
...
+      SomeInterceptor someInterceptor = (SomeInterceptor) interceptor;
+      assertSame(cache.getAdvancedCache().getLockManager(), someInterceptor.lm);
+      assertSame(cache.getAdvancedCache().getDataContainer(), someInterceptor.dc);
    }
 
-   public void testInjectionWorks() {
-      final Cache<Object,Object> cache1 = cacheManager.getCache();
-      assert cache1.getAdvancedCache().getInterceptorChain().get(0).getClass().equals(SomeInterceptor.class);
-      assert SomeInterceptor.lm != null;
+   static class SomeInterceptor extends BaseCustomAsyncInterceptor {
```
danberindei: I think we should keep the `CommandInterceptor`-based test as well, for now.

query/src/test/java/org/infinispan/query/blackbox/ClusteredCacheWithLongIndexNameTest.java
```diff
@@ -64,32 +64,55 @@ public CacheMode getCacheMode() {
...
 
    public void testAdditionOfNewNode() {
-      prepareData();
-
-      SearchManager sm = Search.getSearchManager(cache3);
-      QueryBuilder qb = sm.buildQueryBuilderForClass(VeryLongIndexNamedClass.class).get();
-      Query q = qb.keyword().wildcard().onField("name").matching("value*").createQuery();
-      CacheQuery<?> cq = sm.getQuery(q, VeryLongIndexNamedClass.class);
+      for (int i = 0; i < 100; i++) {
+         cache0.put("key" + i, new ClassWithLongIndexName("value" + i));
```
danberindei: Reading the JIRA, one entry should be enough -- although TBH this was a HibernateSearch bug, I don't think we need to keep a reproducer in our test suite.

danberindei:
You can't keep adding commits and keep the Trivial label as well, @anistor :)

core/src/main/java/org/infinispan/interceptors/base/CommandInterceptor.java
anistor: Ok, let's not upset 'wildlife' then :)

core/src/main/java/org/infinispan/configuration/cache/IndexingConfiguration.java
anistor: indeed

query/src/test/java/org/infinispan/query/blackbox/ClusteredCacheWithLongIndexNameTest.java
anistor: It was too cute to delete it...

core/src/test/java/org/infinispan/configuration/CustomInterceptorInjectionTest.java
anistor: done.

PR #6305, opened 2018-10-03
ISPN-9516 docs for segmentation

oraNod:
Jira: https://issues.jboss.org/browse/ISPN-9516

### DIFF
--- a/documentation/src/main/asciidoc/user_guide/clustering.adoc
+++ b/documentation/src/main/asciidoc/user_guide/clustering.adoc
@@ -269,58 +269,59 @@ While the primary is waiting for the confirmation message from the backup, reads
 backup will see the new value, but reads from the primary will see the old one.
 
 [[key_ownership]]
-==== Key ownership
-Distributed caches split entries into a fixed number of segments, and assign each segment
-to a list of owner nodes.
-Replicated caches do the same, except every node is an owner.
+==== Key Ownership
+Distributed caches split entries into a fixed number of segments and assign
+each segment to a list of owner nodes. Replicated caches do the same, with the
+exception that every node is an owner.
 
-The first node in the owners list is called the *primary owner*, and the others are
-called *backup owners*.
-The segment ownership table is broadcast to every node when the cache topology changes
-(i.e. a node joins or leaves the cluster).
-This way, a node can compute the location of a key itself, without resorting to multicast
-requests or maintaining per-key metadata.
+The first node in the list of owners is the *primary owner*. The other nodes in
+the list are *backup owners*. When the cache topology changes, because a node
+joins or leaves the cluster, the segment ownership table is broadcast to every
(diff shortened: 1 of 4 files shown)


documentation/src/main/asciidoc/user_guide/clustering.adoc
```diff
@@ -269,58 +269,59 @@ While the primary is waiting for the confirmation message from the backup, reads
...
 DefaultConsistentHashFactory::
-  It achieves a more even distribution than `SyncConsistentHashFactory`, but it has one
-  disadvantage: the mapping of segments to nodes depends on the order in which caches
-  joined the cluster, so a key's owners are not guaranteed to be the same in all the
-  caches running in a cluster.
-  Used to be the default from version 5.2 to version 8.1 (with server hinting disabled).
+Achieves a more even distribution than `SyncConsistentHashFactory`, but with one
+disadvantage. The mapping of segments to nodes depends on the order in which
+caches join the cluster. As a result the owners of a key cannot be guaranteed
+to be the same in all the caches running in a cluster.
```
wburns: This might need to be worded differently, it is quite confusing imo. I am guessing this is talking about the fact that a key may map to a different owner in a different cache, depending on the order of the cache topology.

documentation/src/main/asciidoc/user_guide/persistence.adoc
```diff
@@ -267,33 +348,42 @@ The following is an example configuration for the write-behind strategy:
...
+Loss of data can occur if the number of segments in a cache store are not
+changed gracefully. For this reason, if you change the `numSegments` setting in the store configuration, you must migrate the existing store to use the new configuration.
 
-Unfortunately depending on the store implementation, there may be some performance
-loss for write operations. This is especially true for batch writes which can
-occur with transactions or write behind stores, which is mentioned in the previous
-section. Therefore, before enabling segmented stores, this overhead should be
-possibly measured to ensure it is acceptable in turn to gain bulk read performance.
+The recommended method to migrate the cache store configuration is to perform a
+rolling upgrade. The store migrator does not currently support migrating
```
wburns: I wonder if we should be more explicit, it kinda sounds like it doesn't support it at all currently.

> The store migrator currently supports migrating from a non segmented store to a segmented store, however it does not yet support migrating from a segmented store.

Something like that?

documentation/src/main/asciidoc/user_guide/clustering.adoc
oraNod: I believe there is an error in the original version: "the mapping of segments to nodes depends on the order in which caches joined the cluster". Shouldn't this be "the order in which nodes joined the cluster"?

wburns: What you have sounds much better now :)

wburns:
Integrated into master, thanks @oraNod !

PR #6306, opened 2018-10-03
ISPN-9508 NullPointerException from putAll on a cache with a store

wburns:
during rehash

* Don't attempt to insert non owned entries during batch

https://issues.jboss.org/browse/ISPN-9508

### DIFF
--- a/core/src/main/java/org/infinispan/container/impl/InternalEntryFactoryImpl.java
+++ b/core/src/main/java/org/infinispan/container/impl/InternalEntryFactoryImpl.java
@@ -185,11 +185,11 @@ public class InternalEntryFactoryImpl implements InternalEntryFactory {
    }
 
    @Override
-   public InternalCacheValue getValueFromCtxOrCreateNew(Object key, InvocationContext ctx) {
+   public <K, V> InternalCacheValue<V> getValueFromCtx(K key, InvocationContext ctx) {
       CacheEntry entry = ctx.lookupEntry(key);
       if (entry instanceof InternalCacheEntry) {
          return ((InternalCacheEntry) entry).toInternalCacheValue();
-      } else {
+      } else if (entry != null) {
          if (ctx.isInTxScope()) {
             EntryVersionsMap updatedVersions =
                   ((TxInvocationContext) ctx).getCacheTransaction().getUpdatedEntryVersions();
@@ -213,6 +213,8 @@ public class InternalEntryFactoryImpl implements InternalEntryFactory {
             }
          }
          return create(entry).toInternalCacheValue();
+      } else {
+         return null;
       }
    }
 
--- a/core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
@@ -36,19 +36,19 @@ import org.infinispan.commands.write.PutMapCommand;
 import org.infinispan.commands.write.RemoveCommand;
 import org.infinispan.commands.write.ReplaceCommand;
 import org.infinispan.commands.write.WriteCommand;
-import org.infinispan.distribution.ch.KeyPartitioner;
-import org.infinispan.functional.Param;
-import org.infinispan.functional.Param.PersistenceMode;
 import org.infinispan.commons.marshall.StreamingMarshaller;
 import org.infinispan.configuration.cache.PersistenceConfiguration;
-import org.infinispan.container.impl.InternalEntryFactory;
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.container.entries.InternalCacheValue;
+import org.infinispan.container.impl.InternalEntryFactory;
 import org.infinispan.context.InvocationContext;
 import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.context.impl.TxInvocationContext;
+import org.infinispan.distribution.ch.KeyPartitioner;
 import org.infinispan.factories.annotations.Inject;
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/infinispan/container/impl/InternalEntryFactoryImpl.java
```diff
@@ -185,11 +185,11 @@ public CacheEntry copy(CacheEntry cacheEntry) {
    }
 
    @Override
-   public InternalCacheValue getValueFromCtxOrCreateNew(Object key, InvocationContext ctx) {
+   public <K, V> InternalCacheValue<V> getValueFromCtxOrCreateNew(K key, InvocationContext ctx) {
```
danberindei: -1 for the generic arguments, the callers ignore them anyway.
Also -1 for `CreateNew`, to me that implies that the result will never be `null`.

And finally, I'm not sure why we need this method, because all the callers immediately use it to create a `MarshalledEntry`. @ryanemerson do you remember why this was needed?

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
```diff
@@ -461,14 +463,16 @@ public int getNumberOfPersistedEntries() {
...
+         persistenceManager.writeToAllNonTxStores(entry, SegmentSpecificCommand.extractSegment(command, key, keyPartitioner),
+               skipSharedStores(ctx, key, command) ? PRIVATE : BOTH, command.getFlagsBitSet());
+         if (trace) getLog().tracef("Stored entry %s under key %s", entry.getValue(), key);
+      }
    }
 
    MarshalledEntry createMarshalledEntry(InvocationContext ctx, Object key) {
       InternalCacheValue sv = entryFactory.getValueFromCtxOrCreateNew(key, ctx);
-      return new MarshalledEntryImpl(key, sv.getValue(), internalMetadata(sv), marshaller);
+      return sv != null ? new MarshalledEntryImpl(key, sv.getValue(), internalMetadata(sv), marshaller) : null;
```
danberindei: Again `create` in the method name suggests a non-null return value, I suggest `marshallEntry`

core/src/test/java/org/infinispan/commands/PutMapCommandStressTest.java
```diff
@@ -59,6 +60,11 @@ protected void createCacheManagers() throws Throwable {
       if (biasAcquisition != null) {
          builderUsed.clustering().biasAcquisition(biasAcquisition);
       }
+      // UNCOMMENT the following to test put command under topology change with a store
+//      builderUsed.persistence()
+//            .addStore(DummyInMemoryStoreConfigurationBuilder.class)
+//               .shared(true)
+//               .storeName(PutMapCommandStressTest.class.toString());
```
danberindei: We don't have a job for stress tests in CI anyway, but we should have a parameter tho enable the store.

core/src/main/java/org/infinispan/container/impl/InternalEntryFactoryImpl.java
ryanemerson: So originally this method was just in `CacheWriterInterceptor` as `getStoredValue` (seemingly a much better name), but the same logic was required in `TxBatchUpdater` which is why I added this method to `InternalEntryFactory`. 

I think this method is required so that the correct version is retrieved for the metadata. This metadata is then persisted as part of the `MarshalledEntry`.

core/src/main/java/org/infinispan/interceptors/impl/CacheWriterInterceptor.java
karesti: +1 with @danberindei the method should never return null

wburns: Sure I can change the name.

core/src/main/java/org/infinispan/container/impl/InternalEntryFactoryImpl.java
wburns: There are generics on all the other methods, I was making it consistent with them.

Also I agree re: `CreateNew` I just didn't have time last night to remove it.

core/src/test/java/org/infinispan/commands/PutMapCommandStressTest.java
wburns: Yeah, time constraints :)

wburns:
And btw the issue with rehash is more that it seems it has too much data on all the nodes, which I find to be a better issue that having less :) But I only found that in the little time I had yesterday to run the test a few times.

core/src/main/java/org/infinispan/container/impl/InternalEntryFactoryImpl.java
wburns: Oh and also re: generics they are already on the interface, so there is no reason not to put them here really.

https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/container/impl/InternalEntryFactory.java#L168

danberindei:
Ok, this is good enough for 9.4. Let's just wait for CI again after the renames :)

core/src/main/java/org/infinispan/container/impl/InternalEntryFactoryImpl.java
danberindei: Oh well, at least only the methods are parametrized, not the interface itself :))

danberindei:
Integrated, thanks Will!

PR #6310, opened 2018-10-04
ISPN-9578 Move Hot Rod wire docs to contributing guide

tristantarrant:
https://issues.jboss.org/browse/ISPN-9578


gustavonalle:
Are you sure "contributing guide" is the best place? To me contrib guide is how can you prepare yourself to be a contributor (git setup, code conventions, etc). Maybe the HR protocol should go to server guide?

tristantarrant:
No, the server guide is an admin guide.
This could be part of a document which describes low level integration points (together with Cache Stores)

oraNod:
I also feel like this should be in a separate reference guide. IMO it might be a good idea to include this as part of the wider book re-structure effort for the 10 release.

PR #6311, opened 2018-10-04
Avoid parsing media type when an id is provided

gustavonalle:


gustavonalle:
This won't work, will send another PR as part of https://issues.jboss.org/browse/ISPN-9580

PR #6316, opened 2018-10-05
Bump console version to 9.4.0.Final

vblagoje:
Tried and tested locally to pull the 9.4.0.Final console. Also tested that the console actually loads and works as well. This one is gtg.


tristantarrant:
Merged

PR #6318, opened 2018-10-05
ISPN-9589 Client Listener notifications can deadlock

wburns:
https://issues.jboss.org/browse/ISPN-9589
https://issues.jboss.org/browse/ISPN-7087

### DIFF
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
@@ -1,27 +1,20 @@
 package org.infinispan.client.hotrod.near;
 
-import java.nio.ByteBuffer;
-
 import org.infinispan.client.hotrod.MetadataValue;
 import org.infinispan.client.hotrod.RemoteCache;
-import org.infinispan.client.hotrod.annotation.ClientCacheEntryCreated;
 import org.infinispan.client.hotrod.annotation.ClientCacheEntryExpired;
 import org.infinispan.client.hotrod.annotation.ClientCacheEntryModified;
 import org.infinispan.client.hotrod.annotation.ClientCacheEntryRemoved;
 import org.infinispan.client.hotrod.annotation.ClientCacheFailover;
 import org.infinispan.client.hotrod.annotation.ClientListener;
 import org.infinispan.client.hotrod.configuration.NearCacheConfiguration;
-import org.infinispan.client.hotrod.event.ClientCacheEntryCustomEvent;
 import org.infinispan.client.hotrod.event.ClientCacheEntryExpiredEvent;
 import org.infinispan.client.hotrod.event.ClientCacheEntryModifiedEvent;
 import org.infinispan.client.hotrod.event.ClientCacheEntryRemovedEvent;
 import org.infinispan.client.hotrod.event.ClientCacheFailoverEvent;
 import org.infinispan.client.hotrod.event.impl.ClientListenerNotifier;
-import org.infinispan.client.hotrod.impl.MetadataValueImpl;
 import org.infinispan.client.hotrod.logging.Log;
 import org.infinispan.client.hotrod.logging.LogFactory;
-import org.infinispan.commons.io.UnsignedNumeric;
-import org.infinispan.commons.marshall.Marshaller;
 import org.infinispan.commons.util.Util;
 
 /**
@@ -50,17 +43,11 @@ public class NearCacheService<K, V> implements NearCache<K, V> {
          // Create near cache
          cache = createNearCache(config);
          // Add a listener that updates the near cache
-         listener = createListener(remote);
+         listener = new InvalidatedNearCacheListener<>(this);
          remote.addClientListener(listener);
          // Get the listener ID for faster listener connected lookups
          listenerId = listenerNotifier.findListenerId(listener);
       }
-   }
-
-   private Object createListener(RemoteCache<K, V> remote) {
-      return config.mode().invalidated()
-            ? new InvalidatedNearCacheListener(this)
-            : new EagerNearCacheListener(this, remote.getRemoteCacheManager().getMarshaller());
    }
 
    public void stop(RemoteCache<K, V> remote) {
@@ -81,7 +68,7 @@ public class NearCacheService<K, V> implements NearCache<K, V> {
 
    public static <K, V> NearCacheService<K, V> create(
          NearCacheConfiguration config, ClientListenerNotifier listenerNotifier) {
-      return new NearCacheService(config, listenerNotifier);
+      return new NearCacheService<>(config, listenerNotifier);
    }
 
    @Override
@@ -189,77 +176,4 @@ public class NearCacheService<K, V> implements NearCache<K, V> {
          cache.remove(key);
       }
    }
-
-   /**
-    * An near cache listener that eagerly populates the near cache as cache
-    * entries are created/modified in the server. It uses a converter in order
-    * to ship value and version information as well as the key. To avoid sharing
-    * classes between client and server, it uses raw data in the converter.
-    * This listener does not apply any filtering, so all keys are considered.
-    */
-   @Deprecated
-   @ClientListener(converterFactoryName = "___eager-key-value-version-converter", useRawData = true)
-   private static class EagerNearCacheListener<K, V> {
-      private static final Log log = LogFactory.getLog(EagerNearCacheListener.class);
-      private final NearCache<K, V> cache;
-      private final Marshaller marshaller;
-
-      private EagerNearCacheListener(NearCache<K, V> cache, Marshaller marshaller) {
-         this.cache = cache;
-         this.marshaller = marshaller;
-      }
-
-      @ClientCacheEntryCreated
-      @ClientCacheEntryModified
-      @SuppressWarnings("unused")
-      public void handleCreatedModifiedEvent(ClientCacheEntryCustomEvent<byte[]> e) {
-         ByteBuffer in = ByteBuffer.wrap(e.getEventData());
-         byte[] keyBytes = extractElement(in);
-         byte[] valueBytes = extractElement(in);
-         K key = unmarshallObject(keyBytes, "key");
-         V value = unmarshallObject(valueBytes, "value");
-         long version = in.getLong();
-         if (key != null && value != null) {
-            MetadataValueImpl<V> entry = new MetadataValueImpl<>(0, 0, 0, 0, version, value);
-            cache.put(key, entry);
-         }
-      }
-
-      @SuppressWarnings("unchecked")
-      private <T> T unmarshallObject(byte[] bytes, String element) {
-         try {
-            return (T) marshaller.objectFromByteBuffer(bytes);
-         } catch (Exception e) {
-            log.unableToUnmarshallBytesError(element, Util.toStr(bytes), e);
-            return null;
-         }
-      }
-
-      @ClientCacheEntryRemoved
-      @ClientCacheEntryExpired
-      @SuppressWarnings("unused")
-      public void handleRemovedEvent(ClientCacheEntryCustomEvent<byte[]> e) {
-         ByteBuffer in = ByteBuffer.wrap(e.getEventData());
-         byte[] keyBytes = extractElement(in);
-         K key = unmarshallObject(keyBytes, "key");
-         if (key != null) {
-            cache.remove(key);
-         }
-      }
-
-      @ClientCacheFailover
-      @SuppressWarnings("unused")
-      public void handleFailover(ClientCacheFailoverEvent e) {
-         if (trace) log.trace("Clear near cache after fail-over of server");
-         cache.clear();
-      }
-
-      private static byte[] extractElement(ByteBuffer in) {
-         int length = UnsignedNumeric.readUnsignedInt(in);
-         byte[] element = new byte[length];
-         in.get(element);
-         return element;
-      }
-   }
-
 }
--- a/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
+++ b/client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
@@ -5,12 +5,14 @@ import java.nio.ByteBuffer;
 import org.infinispan.client.hotrod.MetadataValue;
 import org.infinispan.client.hotrod.RemoteCache;
 import org.infinispan.client.hotrod.annotation.ClientCacheEntryCreated;
+import org.infinispan.client.hotrod.annotation.ClientCacheEntryExpired;
 import org.infinispan.client.hotrod.annotation.ClientCacheEntryModified;
 import org.infinispan.client.hotrod.annotation.ClientCacheEntryRemoved;
 import org.infinispan.client.hotrod.annotation.ClientCacheFailover;
 import org.infinispan.client.hotrod.annotation.ClientListener;
 import org.infinispan.client.hotrod.configuration.NearCacheConfiguration;
 import org.infinispan.client.hotrod.event.ClientCacheEntryCustomEvent;
+import org.infinispan.client.hotrod.event.ClientCacheEntryExpiredEvent;
 import org.infinispan.client.hotrod.event.ClientCacheEntryModifiedEvent;
 import org.infinispan.client.hotrod.event.ClientCacheEntryRemovedEvent;
 import org.infinispan.client.hotrod.event.ClientCacheFailoverEvent;
@@ -170,6 +172,12 @@ public class NearCacheService<K, V> implements NearCache<K, V> {
          invalidate(event.getKey());
       }
 
+      @ClientCacheEntryExpired
+      @SuppressWarnings("unused")
+      public void handleExpiredEvent(ClientCacheEntryExpiredEvent<K> event) {
+         invalidate(event.getKey());
+      }
+
       @ClientCacheFailover
       @SuppressWarnings("unused")
       public void handleFailover(ClientCacheFailoverEvent e) {
@@ -228,6 +236,7 @@ public class NearCacheService<K, V> implements NearCache<K, V> {
       }
 
       @ClientCacheEntryRemoved
+      @ClientCacheEntryExpired
       @SuppressWarnings("unused")
       public void handleRemovedEvent(ClientCacheEntryCustomEvent<byte[]> e) {
          ByteBuffer in = ByteBuffer.wrap(e.getEventData());
--- a/server/hotrod/src/main/java/org/infinispan/server/hotrod/BaseDecoder.java
+++ b/server/hotrod/src/main/java/org/infinispan/server/hotrod/BaseDecoder.java
@@ -16,6 +16,7 @@ import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.server.core.ServerConstants;
 import org.infinispan.server.hotrod.logging.Log;
 
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.ByteToMessageDecoder;
 import io.netty.handler.codec.DecoderException;
@@ -62,11 +63,15 @@ abstract class BaseDecoder extends ByteToMessageDecoder {
    @Override
    public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
       super.channelWritabilityChanged(ctx);
+      Channel channel = ctx.channel();
+      boolean writeable = channel.isWritable();
       if (trace) {
-         log.tracef("Channel %s writability changed", ctx.channel());
+         log.tracef("Channel %s writability changed to %b", channel, writeable);
       }
-      server.getClientListenerRegistry().findAndWriteEvents(ctx.channel());
-      server.getClientCounterNotificationManager().channelActive(ctx.channel());
+      if (writeable) {
+         server.getClientListenerRegistry().findAndWriteEvents(channel);
+         server.getClientCounterNotificationManager().channelActive(channel);
+      }
    }
 
    @Override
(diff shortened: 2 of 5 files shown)


wburns:
You can verify test by running `ClientEventStressTest` that failed everytime before on master for me due to TimeoutException.

wburns:
I am not sold on using Caffeine for this, but it was a quick fix for the issue of dynamically adding or removing listeners. I didn't have that much time to vet it more, but this should fix stuff :D

wburns:
This is not quite ready yet. It can still deadlock an event thread in the event submission. 

This invocation https://github.com/infinispan/infinispan/blob/master/server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java#L393 can still block forever  if this is invoked on the event loop thread. We may have to timeout and drain then attempt to readd. I can try some different options later.

If anyone has any suggestions they would be appreciated as well.

wburns:
Updated and should work properly now. There are still performance issues, especially if a swamp of async writes occur. But this will have to be fixed in a different PR.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
```diff
@@ -170,6 +172,12 @@ public void handleRemovedEvent(ClientCacheEntryRemovedEvent<K> event) {
          invalidate(event.getKey());
       }
 
+      @ClientCacheEntryExpired
```
galderz: Seems like you're fixing [ISPN-7087](https://issues.jboss.org/browse/ISPN-7087) indirectly here?

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
```diff
@@ -117,13 +126,20 @@ public HotRodServer() {
...
+   }
+
    private Address clusterAddress;
    private ServerAddress address;
    private Cache<Address, ServerAddress> addressCache;
    private Map<String, AdvancedCache> knownCaches = CollectionFactory.makeConcurrentMap(4, 0.9f, 16);
    private Map<String, Configuration> knownCacheConfigurations = CollectionFactory.makeConcurrentMap(4, 0.9f, 16);
    private Map<String, ComponentRegistry> knownCacheRegistries = CollectionFactory.makeConcurrentMap(4, 0.9f, 16);
-   private final Map<String, CacheInfo> cacheInfo = new ConcurrentHashMap<>();
+   // This needs to be an expiration map so we can update knowledge about if sync listeners are registered
```
galderz: Why do you need expiration exactly?

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
```diff
@@ -228,6 +236,7 @@ public void handleCreatedModifiedEvent(ClientCacheEntryCustomEvent<byte[]> e) {
       }
 
       @ClientCacheEntryRemoved
+      @ClientCacheEntryExpired
```
danberindei: 2 methods annotated with `@ClientCacheEntryExpired`?

server/hotrod/src/main/java/org/infinispan/server/hotrod/BaseDecoder.java
```diff
@@ -62,11 +63,15 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
    public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
       super.channelWritabilityChanged(ctx);
+      Channel channel = ctx.channel();
+      boolean writeable = channel.isWritable();
       if (trace) {
-         log.tracef("Channel %s writability changed", ctx.channel());
+         log.tracef("Channel %s writability changed to %b", channel, writeable);
+      }
+      if (writeable) {
+         server.getClientListenerRegistry().findAndWriteEvents(channel);
```
danberindei: `writeEventsIfPossible()` already checks `channel.isWritable()` before writing.

server/hotrod/src/main/java/org/infinispan/server/hotrod/BaseDecoder.java
```diff
@@ -62,11 +63,15 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
+         server.getClientCounterNotificationManager().channelActive(channel);
```
danberindei: Maybe we should check that the channel is still writable after sending the events?

server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
```diff
@@ -388,16 +390,38 @@ void sendEvent(byte[] key, byte[] value, long dataVersion, CacheEntryEvent event
...
-         boolean waitingForFlush = !ch.isWritable();
-         try {
-            eventQueue.put(remoteEvent);
-         } catch (InterruptedException e) {
-            throw new CacheException(e);
-         }
+         EventLoop loop = ch.eventLoop();
+         // We shouldn't be in the event loop, but just in case we can't get stuck blocking on putting into the queue
+         // so we offer and try to catch up on events if possible
+         if (loop.inEventLoop()) {
```
danberindei: IMO we could just throw an exception instead of complicating the code

server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
```diff
@@ -388,16 +390,38 @@ void sendEvent(byte[] key, byte[] value, long dataVersion, CacheEntryEvent event
...
+                  throw new CacheException(e);
+               }
+            }
+            if (ch.isWritable()) {
+               // Now that the entry is finally in eventQueue, submit the events to be written at a later point
+               loop.submit(writeEventsIfPossible);
+            }
+         } else {
+            try {
+               eventQueue.put(remoteEvent);
```
danberindei: I think we should use `eventQueue.offer(remoteEvent, remoteTimeout, MILLISECONDS)` here, so we eventually time out and report an exception.

server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
```diff
@@ -388,16 +390,38 @@ void sendEvent(byte[] key, byte[] value, long dataVersion, CacheEntryEvent event
...
+            } catch (InterruptedException e) {
+               throw new CacheException(e);
+            }
 
-         if (!waitingForFlush) {
-            // Make sure we write any event in main event loop
-            ch.eventLoop().submit(writeEventsIfPossible);
+            if (ch.isWritable()) {
+               // Make sure we write any event in main event loop
+               loop.submit(writeEventsIfPossible);
```
danberindei: This is the same in both branches. The comment is different, but this one is better ;)

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
```diff
@@ -117,13 +126,20 @@ public HotRodServer() {
...
+   private final Map<String, CacheInfo> cacheInfo = expirationMap(10, TimeUnit.SECONDS);
```
danberindei: I don't like this approach, it means we could have event loop threads processing write requests for 10s after the listener is added. Since we don't really care about all cache listeners, only about client listeners, and only on the originator, we could intercept client listener add requests here and update the map synchronously.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
```diff
@@ -432,12 +460,14 @@ public CacheInfo getCacheInfo(AdvancedCache<byte[], byte[]> cache, HotRodHeader
                .noFlags().withFlags(LOCAL_NON_BLOCKING_GET);
          if (cache.getStatus() != ComponentStatus.RUNNING) {
             // stay on the safe side
-            return new CacheInfo(localNonBlocking, true, true);
+            return new CacheInfo(localNonBlocking, true, true, true);
```
danberindei: I wonder which parameter the comment above refers to...

danberindei:
Shouldn't we change `ClientEventStressTest` to use only 3  (or even 2) io threads by default?

I'd prefer a non-stress test that blocks the client and generates just enough events for the server-side queue to fill up, but I'm not sure how easy or hard that would be to write.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
wburns: Yeah let me make a separate commit for it.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
wburns: One is for eager and the other is for invalidated near cache.

server/hotrod/src/main/java/org/infinispan/server/hotrod/BaseDecoder.java
wburns: Yes, but `findAndWriteEvents` doesn't and allocates additional objects and uses event loop resources. Is a minor optimization.

server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
wburns: Unfortunately we can't do that really. The problem is someone could add a sync listener after the `CacheRequestProcessor` has already checked and not found one which could cause it possibly to fire in the event loop.

server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
wburns: Sure.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
wburns: :) I was assuming it was all of them - since the optimizations only fire if the values are false.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
wburns: Hrmm, we care about all sync listeners, doesn't matter where they come from. Otherwise we would be blocking the io event loop for user listeners, cluster listeners or others as well. The two types mentioned are added outside of this class. I will see if I can do something else though.

wburns:
@danberindei It should be pretty easy to write a test that exhibits this issue without my changes since we have a few other issues with listeners imo. All you need is a client listener that takes even a millisecond and a lot of writes and it will fill up the buffer and eventually hit the event loop and deadlock.

wburns:
Also this is just a bandaid until 10.0 when I am planning on adding support for listeners working with async interceptor stack in which case this issue goes away.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
wburns: So the reason for expiration is due to the fact that this `CacheInfo` is cached. Before this PR this information was static to the Cache configuration. However now that we require checking if a sync listener is installed, which can change over time, we need a way to recheck this every so often.

There may be a way to plug into listener installation to see if one is sync, but I have my doubts about that working.

And as I mentioned in my below comment, when I refactor listeners for 10.0 this is no longer needed and can be reverted to a regular `ConcurrentHashMap`

wburns:
Actually working on this more, unfortunately making a simple test is not that simple. As you need to use multiple threads to overwhelm the io thread pools. It would essentially become a stress tests of itself. I personally feel like we need to make sure we have the stress tests running on CI more often and actually pay attention to the outcomes :)

wburns:
Broke out additional commit and also updated rework comments that I could.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
wburns: I am not sure of an easy way to do this without changing a bunch of stuff. Given that this is temporary anyways, I would rather just touch as little as possible for now.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
danberindei: I didn't realize `EagerNearCacheListener` is still there, but it's private and never used, so I'd rather delete it.
There is a reference to it in the code, but that condition in `NearCacheService.createListener()` is never false.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
danberindei: I didn't realize EagerNearCacheListener is still there, but it's private and never used, so I'd rather delete it.
There is a reference to it in the code, but that condition in NearCacheService.createListener() is never false.

server/hotrod/src/main/java/org/infinispan/server/hotrod/BaseDecoder.java
danberindei: I meant it's much more likely that the channel is no longer active after we wrote some events than right after we got the notification that the channel is active. Unless `super.channelWritabilityChanged(ctx)` can also cause some writes to the channel?

server/hotrod/src/main/java/org/infinispan/server/hotrod/BaseDecoder.java
danberindei: Oops, I've seen the method name at least 10 times and only now I see that the method is also called when the channel becomes non-writable :)

But I think `findAndWriteEvents()` also needs some work, because it does the event loop thing even if there is no client sender associated with this channel, and the way we look up the client sender in the map is pretty atrocious.

server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
danberindei: TBH I'm still a little confused why we would ever keep the event loop busy while we are running any cache operation. I mean that prevents any other thread from reading from the channel, so the server is going to spend a lot of time twiddling its thumbs unless there are lots of clients per server. I understand that we wouldn't want to incur a context switch to process short requests, but couldn't we allow another thread to start reading from/writing to the channel while we are doing some other work?

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
danberindei: I was thinking this is the server, so we shouldn't have any sync or async listeners except for the client listeners, but I guess it's fine if you add a comment that it's all going away in 10 so I can hold you to it :)

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
danberindei: I don't think this check is needed, we can't get a reference to a cache before it's started and if it's stopping/stopped we'll throw an exception immediately.

danberindei:
> I personally feel like we need to make sure we have the stress tests running on CI more often and actually pay attention to the outcomes :)

I feel the same way, but I'm not creating the job in https://ci.infinispan.org/job/InfinispanAlternateBuilds/ either :D

server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
wburns: That is just the way @rvansa wrote it when he changed it over to share the connection.

I originally wrote it the way you said, except gets and ping were exceptions allowing reads on the io thread.

https://github.com/infinispan/infinispan/blob/12839273381601249956986f1987d7ad8758e844/server/hotrod/src/main/java/org/infinispan/server/hotrod/transport/HotRodChannelInitializer.java#L47

When things were more using the async stack was when Radim felt it would be better off allowing all operations that could be async to do this. This PR/bug is a side effect of missing a case :)

I am not sure if what we have is better or worse atm than before (assuming no bugs with blocking). But maybe write operations are too big even when CPU only bound?

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
wburns: A client listener is registered as a sync cluster listener on other nodes, which is the big case this misses. Technically this shouldn't be an issue as you mentioned as it should be the primary owner, but sometimes it may not :)

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
wburns: I have no idea personally, just how it was added by @rvansa. He would know more about why this check is present. But I would be up for removing, however I wouldn't want to do it in this PR.

server/hotrod/src/main/java/org/infinispan/server/hotrod/BaseDecoder.java
wburns: Not sure if this is intermingled with the other comment chain now? But yes I agree that the channel may not be writeable most likely after exhausting all 100+ elements from the event queue. But I don't understand what change you are proposing here though. If the channel is no longer writeable after writing the events, we just wait until we get the callback again telling us the channel is writeable to continue writing more.

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
wburns: I will add a sub JIRA under https://issues.jboss.org/browse/ISPN-8070 for it.

server/hotrod/src/main/java/org/infinispan/server/hotrod/BaseDecoder.java
wburns: Hrmm, it only invokes `writeEventsIfPossible` if the channel that is now writeable is associated with the given event sender in `findAndWriteEvents`. And you are referring to how we have to loop through all values I am guessing :(

I didn't write this, but I don't see why we couldn't have a `ConcurrentMap<Channel, ConcurrentHashSet<BaseClientEventSender>>` or something similar.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
danberindei: Why not? :smiley:

server/hotrod/src/main/java/org/infinispan/server/hotrod/BaseDecoder.java
danberindei: Ok, so I finally started reading a bit about the Netty thread model and I saw my expectations were completely wrong. I didn't realize a channel was actually bound to a single thread, I thought the event loop was only bound to a thread pool and every event arrived on another thread in the pool, something like the [5.0 thread model](https://netty.io/wiki/new-and-noteworthy-in-5.0.html#even-more-flexible-thread-model), only with regular thread pools instead of `ForkJoinPool`.

server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
danberindei: We should trigger a perf ack job with and without passing requests to another thread pool again, but I'd argue that even the serialization of read results could be heavy enough to justify doing it on another thread.

server/hotrod/src/main/java/org/infinispan/server/hotrod/HotRodServer.java
danberindei: I was thinking that even on the primary owner, the client listener is triggered only after the backups ack the write, on a remote thread. But your approach is definitely safer, the primary could temporarily be the only owner, and we don't want to make too many assumptions about how distribution works here in the server.

danberindei:
> As you need to use multiple threads to overwhelm the io thread pools.

If Netty really uses a single thread for each channel, then we can add a `ChannelInboundHandler` that waits on a latch.

danberindei:
Could this failure be related to your changes?

https://ci.infinispan.org/job/Infinispan/job/PR-6318/9/testReport/junit/org.jsr107.tck.event/CacheListenerTest/testCacheClearListener_true_/

wburns:
Okay looking closer, it is a randomly failing test, remote cache notifies listeners asynchronously. The remote JCache implementation doesn't ignore listeners for a clear method, where as embedded does. So the failure is a different issue. My fixes may have made the event be generated a little quicker, so it may exhibit the issue more frequently.

wburns:
It looks like this was discussed in https://issues.jboss.org/browse/ISPN-9155 and https://github.com/infinispan/infinispan/pull/6036 but nothing came out of it

client/hotrod-client/src/main/java/org/infinispan/client/hotrod/near/NearCacheService.java
danberindei: @wburns I would still like to remove `EagerNearCacheListener` in this PR, I don't think we'll get any eager near cache-related PRs any time soon.

wburns: Sure

wburns:
Updated to remove eager near cache listener

danberindei:
Integrated, thanks Will!

PR #6319, opened 2018-10-08
ISPN-9591 Adds spring boot as a local dependency

karesti:
https://issues.jboss.org/browse/ISPN-9591

### DIFF
--- a/integrationtests/spring-boot-it/pom.xml
+++ b/integrationtests/spring-boot-it/pom.xml
@@ -19,6 +19,24 @@
         <overridden.version.mockito>1.10.19</overridden.version.mockito>
     </properties>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-bom</artifactId>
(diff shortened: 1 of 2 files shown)


karesti:
test failures are unrelated

pruivo:
@jsenko will this change work with PME?

integrationtests/spring-boot-it/pom.xml
```diff
@@ -19,6 +19,24 @@
...
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-bom</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-dependencies</artifactId>
+                <version>1.5.16.RELEASE</version>
```
karesti: we can keep the version and the property, but not in the bom, but in the build-configuration pom.xml, because it's the parent pom of infinispan

karesti: @jsenko

PR #6320, opened 2018-10-08
ISPN-9593 OpenShift doc attributes

oraNod:
Need to include attributes for the OpenShift docs with this Jira: https://issues.jboss.org/browse/ISPN-9593


oraNod:
Closing this PR. I think these attributes are specific to Data Grid and probably better to manage in the downstream repo only.

PR #6322, opened 2018-10-09
ISPN-8003 ProtobufMetadataManagerInterceptor fails the InterceptorDefinesAllReadWritesCheck

anistor:
https://issues.jboss.org/browse/ISPN-8003


anistor:
Rebased.

query/src/main/java/org/infinispan/query/backend/QueryKnownClasses.java
```diff
@@ -151,40 +154,55 @@ void stop() {
...
       Transaction tx = transactionHelper.suspendTxIfExists();
       try {
-         runCommand(() -> knownClassesCache.put(new KeyValuePair<>(cacheName, clazz), value));
+         // put and retry if SuspectException was thrown
+         while (true) {
+            try {
+               knownClassesCache.put(new KeyValuePair<>(cacheName, clazz), value);
+               break;
+            } catch (CacheException e) {
+               if (SuspectException.isSuspectExceptionInChain(e)) {
```
danberindei: This shouldn't be needed, when we get a `SuspectException` we retry internally (in `StateTransferInterceptor`)

remote-query/remote-query-server/src/test/java/org/infinispan/query/remote/impl/ProtobufMetadataManagerInterceptorTest.java
```diff
@@ -268,6 +272,38 @@ public void testStatusAfterRemove() {
...
+         FunctionalMap.ReadWriteMap<String, String> rwMap = ReadWriteMapImpl.create(FunctionalMapImpl.create(cache0.getAdvancedCache()));
+
+         rwMap.eval("test.proto", "val", (value, view) -> {
+            view.set(value);
+            return "ret";
+         }).join();
+
+      } catch (CompletionException e) {
+         assertTrue(e.getCause() instanceof CacheException);
+         assertTrue(e.getCause().getMessage().contains("ISPN028014"));
```
danberindei: You could use `Exceptions.expectException(CacheException.class, rwMap.eval(...))`, although there's no overload with a `messageRegex` parameter yet it should be easy to add.

But I also wonder about the error code, I found https://github.com/infinispan/infinispan/blob/81b60c6f22dbfe035a4c004a678606d60544cd00/counter/src/main/java/org/infinispan/counter/logging/Log.java#L65

anistor: That logging id belongs to query. Counter module came 6 months later and used the same id range, not sure how that happened, and worse, some of those ids are used in commons too.

danberindei:
Integrated, thanks Adrian!

Please have a chat with @pruivo about the overlapping log id ranges :)

PR #6323, opened 2018-10-09
Add dependency on infinispan-feature-pack-wf-modules in wildfly-modul…

danberindei:
…es integration tests


tristantarrant:
Merged

PR #6324, opened 2018-10-09
ISPN-8616 DistAsyncFuncTest.testMergeFromNonOwner random failures

danberindei:
https://issues.jboss.org/browse/ISPN-8816
https://issues.jboss.org/browse/ISPN-8298

* Wait for replication in locked stream tests
* Delay command execution to trigger random failures more often

Should fix ISPN-8298 as well, although the `ReplListener` approach still has the risk of asynchronous commands leaking from one test method to the next. I hope to have more time to switch to `ControlledRpcManager` in the future.

### DIFF
--- a/core/src/test/java/org/infinispan/distribution/DistAsyncFuncTest.java
+++ b/core/src/test/java/org/infinispan/distribution/DistAsyncFuncTest.java
@@ -32,8 +32,8 @@ public class DistAsyncFuncTest extends DistSyncFuncTest {
 
    ReplListener r1, r2, r3, r4;
    ReplListener[] r;
-   Map<Cache<?, ?>, ReplListener> listenerLookup;
-   ConcurrentMap<Cache<?, ?>, List<InvalidateL1Command>> expectedL1Invalidations = new ConcurrentHashMap<>();
+   Map<Address, ReplListener> listenerLookup;
+   ConcurrentMap<Address, List<InvalidateL1Command>> expectedL1Invalidations = new ConcurrentHashMap<>();
 
    @Override
    public Object[] factory() {
@@ -57,7 +57,7 @@ public class DistAsyncFuncTest extends DistSyncFuncTest {
       r4 = new ReplListener(c4, true, true);
       r = new ReplListener[]{r1, r2, r3, r4};
       listenerLookup = new HashMap<>();
-      for (ReplListener rl : r) listenerLookup.put(rl.getCache(), rl);
+      for (ReplListener rl : r) listenerLookup.put(rl.getCache().getCacheManager().getAddress(), rl);
 
       for (Cache c : caches) {
          TestingUtil.wrapComponent(c, RpcManager.class, original -> new AbstractDelegatingRpcManager(original) {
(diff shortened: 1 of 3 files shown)


tristantarrant:
46 failures ...

core/src/test/java/org/infinispan/distribution/DistAsyncFuncTest.java
```diff
@@ -105,21 +100,21 @@ protected void asyncWait(Object key, Predicate<VisitableCommand> command) {
          for (ReplListener rl : r) rl.waitForRpc();
       } else {
          for (Cache<?, ?> c : getOwners(key)) {
-            listenerLookup.get(c).expect(command);
-            listenerLookup.get(c).waitForRpc();
+//            listenerLookup.get(address(c)).expect(command);
+//            listenerLookup.get(address(c)).waitForRpc();
```
danberindei: Oops, I was testing that the test really fails without the wait, and forgot to uncomment these afterwards.

danberindei:
Should be ok now @tristantarrant

core/src/test/java/org/infinispan/distribution/DistAsyncFuncTest.java
wburns: Yeah I actually saw that, but forgot to submit my review comments for this :)

core/src/test/java/org/infinispan/test/ReplListener.java
```diff
@@ -259,7 +259,11 @@ private boolean isPrimaryOwner(VisitableCommand cmd) {
       @Override
       protected Object handleDefault(InvocationContext ctx, VisitableCommand cmd) throws Throwable {
          try {
-            // first pass up chain
+            if (!ctx.isOriginLocal() || (watchLocal && isPrimaryOwner(cmd))) {
+               infof("Delaying command %s", cmd);
```
wburns: I am guessing this is not needed anymore? It was just to help reproduce the issue correct?

wburns: Or if we do keep it, which it looks like we wanted, this should be debug

danberindei: We do need the delay, otherwise the test is kind of pointless -- although it would be much nicer to use `ControlledRpcManager` instead and unblock the requests immediately instead of sleeping.

I logged it at INFO leve because the other logs in `ReplListenerInterceptor` already use `infof()`, but I'll change it to `debugf()` everywhere.

danberindei: It was actually logging at debug level, only the method name stayed `infof()` :)

wburns:

danberindei:
@wburns I missed the comment here, sorry.

The real fix was the first bullet point: "Wait for replication in locked stream tests." Unfortunately I didn't add it correctly, so I had to open a follow-up PR: https://github.com/infinispan/infinispan/pull/6336

Replacing the Cache with Address was just a cosmetic thing, because I found the stream operations in `createCacheManagers()` really obscured the logic there. I also changed a logic a bit so that `InvalidateL1Command` requests broadcast to the entire cluster are properly counted, but I don't think it made a difference.

PR #6326, opened 2018-10-09
ISPN-8799 Limit the default number of shards in the AffinityIndexManager

diegolovison:
https://issues.jboss.org/browse/ISPN-8799


gustavonalle:
Integrated, thanks @diegolovison !

PR #6329, opened 2018-10-10
ISPN-9588 Workaround for no JGroups view after coordinator leaving

danberindei:
https://issues.jboss.org/browse/ISPN-9588

This is just a workaround, please don't resolve the issue.
Since I've been exercising the long tests watcher I've made some small improvements in that area as well.

### DIFF
--- a/commons-test/src/main/java/org/infinispan/commons/test/TestSuiteProgress.java
+++ b/commons-test/src/main/java/org/infinispan/commons/test/TestSuiteProgress.java
@@ -96,7 +96,7 @@ public class TestSuiteProgress {
          actualColor = color;
          actualReset = RESET;
       }
-      out.printf("%s[OK: %5s, KO: %5s, SKIP: %5s] %s%s%n", actualColor, succeeded.get(), failed.get(), skipped.get(), message, actualReset);
+      out.printf("%s[OK:%5s, KO:%3s, SKIP:%3s] %s%s%n", actualColor, succeeded.get(), failed.get(), skipped.get(), message, actualReset);
       if (t != null) {
          t.printStackTrace(out);
       }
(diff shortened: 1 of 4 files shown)


commons-test/src/main/java/org/infinispan/commons/test/TestSuiteProgress.java
```diff
@@ -96,7 +96,7 @@ synchronized void progress(String color, CharSequence message, Throwable t) {
          actualColor = color;
          actualReset = RESET;
       }
-      out.printf("%s[OK: %5s, KO: %5s, SKIP: %5s] %s%s%n", actualColor, succeeded.get(), failed.get(), skipped.get(), message, actualReset);
+      out.printf("%s[OK:%5s, KO:%3s, SKIP:%3s] %s%s%n", actualColor, succeeded.get(), failed.get(), skipped.get(), message, actualReset);
```
tristantarrant: I guess if we have more than 999 failures, who cares about alignment :)

alanfx: I've got 999 problems, but alignment ain't one!

danberindei: I hope it doesn't mess with anyone's scripts, but I have one that adds an escape sequence to go up one line after a test progress line, so the more progress messages fit into a terminal line, the less scrolling.

danberindei:
Unfortunately [Build #3](https://ci.infinispan.org/job/Infinispan/job/PR-6329/3) still has some hanged tests.

danberindei:
The next version of JGroups will have a fix, closing.

PR #6330, opened 2018-10-11
ISPN-9601 JMX cluster switching

tristantarrant:
https://issues.jboss.org/browse/ISPN-9601
https://issues.jboss.org/browse/ISPN-9602

Introduces thread name prefix/suffix for the client async pool.

Let's see if https://issues.jboss.org/browse/ISPN-9064 is also resolved

### DIFF
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/HitsAwareCacheManagersTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/HitsAwareCacheManagersTest.java
@@ -17,8 +17,10 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.infinispan.Cache;
 import org.infinispan.client.hotrod.test.HotRodClientTestingUtil;
 import org.infinispan.commands.VisitableCommand;
+import org.infinispan.commands.read.AbstractDataCommand;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.context.InvocationContext;
+import org.infinispan.context.impl.FlagBitSets;
 import org.infinispan.interceptors.BaseAsyncInterceptor;
 import org.infinispan.lifecycle.ComponentStatus;
 import org.infinispan.manager.CacheContainer;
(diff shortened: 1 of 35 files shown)


tristantarrant:
The *fix* for ISPN-9064 is to use 127.0.0.1 instead of localhost in the client configuration.
When switching sites, the pools with the failed servers are closed and new pools for the added servers are created. Unfortunately, in the test, sometimes the subsequent PUT operation would still be executed with a channel from the old pool: the request would be processed by the server, but the client would ignore the response because the channel was invalidated in flight. This causes the client retry logic to invoke the operation a second time. @rvansa while the test should be now fixed, I believe the underlying issue should warrant its own Jira.

danberindei:
@tristantarrant how exactly does 127.0.0.1 help? Otherwise LGTM.

tristantarrant:
the client configuration points to localhost:port for the backup site, but when switching, it performs a ping and receives the topology for the backup cluster, which includes 127.0.0.1:port. This causes the "localhost:port" pools to be invalidated and "127.0.0.1:port" pools to be created instead. For some reason the PUT after the switch is being created with a "localhost:port" connection, which gets invalidated soon after sending to the server and then sent again via a connection from the "127.0.0.1:port". The server actually executes 2 PUTs.

danberindei:
Could you put this explanation in a comment in the test @tristantarrant? I'm definitely not going to remember it :)

tristantarrant:
Added it

tristantarrant:
1 unrelated failure

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/HitsAwareCacheManagersTest.java
```diff
@@ -188,26 +190,33 @@ protected void destroy() {
...
+      private final AtomicInteger localSiteInvocationCount = new AtomicInteger(0);
+      private final AtomicInteger backupSiteInvocationCount = new AtomicInteger(0);
 
       @Override
-      public Object visitCommand(InvocationContext ctx, VisitableCommand command) throws Throwable {
+      public Object visitCommand(InvocationContext ctx, VisitableCommand command) {
          if (ctx.isOriginLocal()) {
-            int count = invocationCount.incrementAndGet();
-            log.infof("Hit %d for %s", count, command);
+            if ((command instanceof AbstractDataCommand) && ((AbstractDataCommand)command).getFlags().contains(Flag.SKIP_XSITE_BACKUP)) {
```
danberindei:
```suggestion
            if ((command instanceof AbstractDataCommand) && ((AbstractDataCommand)command).hasAnyFlag(FlagBitSets.SKIP_XSITE_BACKUP)) {
```
(needs an import as well)

tristantarrant:
Applied your suggestion @danberindei

danberindei:
Integrated, thanks Tristan!

PR #6334, opened 2018-10-11
ISPN-9604 Support to run converters in the user requested format

gustavonalle:
https://issues.jboss.org/browse/ISPN-9604

### DIFF
--- a/core/src/main/java/org/infinispan/notifications/cachelistener/filter/CacheEventConverter.java
+++ b/core/src/main/java/org/infinispan/notifications/cachelistener/filter/CacheEventConverter.java
@@ -1,6 +1,7 @@
 package org.infinispan.notifications.cachelistener.filter;
 
 import org.infinispan.commons.dataconversion.MediaType;
+import org.infinispan.commons.util.Experimental;
 import org.infinispan.metadata.Metadata;
 
 /**
@@ -25,6 +26,16 @@ public interface CacheEventConverter<K, V, C> {
 
    default MediaType format() {
       return MediaType.APPLICATION_OBJECT;
+   }
+
+   /**
+    * @return if true, {@link #convert(Object, Object, Metadata, Object, Metadata, EventType)} will be presented with data
+    * in the request format rather than the format specified in {@link #format()}. The request format is defined as the MediaType
+    * that a cache was previously decorated with {@link org.infinispan.AdvancedCache#withMediaType(String, String)}.
+    */
+   @Experimental
+   default boolean useRequestFormat() {
+      return false;
    }
 
 }
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/JsonKeyValueRawEventsTest.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/JsonKeyValueRawEventsTest.java
@@ -1,0 +1,103 @@
+package org.infinispan.client.hotrod.event;
+
+import static org.infinispan.commons.dataconversion.MediaType.APPLICATION_JBOSS_MARSHALLING_TYPE;
+import static org.infinispan.commons.dataconversion.MediaType.APPLICATION_JSON;
+import static org.testng.Assert.assertEquals;
+import static org.testng.AssertJUnit.assertNotNull;
+
+import java.nio.ByteBuffer;
+import java.util.Collections;
+import java.util.Queue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+
+import org.infinispan.client.hotrod.DataFormat;
+import org.infinispan.client.hotrod.RemoteCache;
+import org.infinispan.client.hotrod.annotation.ClientCacheEntryCreated;
+import org.infinispan.client.hotrod.annotation.ClientCacheEntryModified;
+import org.infinispan.client.hotrod.annotation.ClientListener;
+import org.infinispan.client.hotrod.test.SingleHotRodServerTest;
+import org.infinispan.commons.configuration.ClassWhiteList;
+import org.infinispan.commons.io.UnsignedNumeric;
+import org.infinispan.commons.marshall.UTF8StringMarshaller;
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.manager.EmbeddedCacheManager;
+import org.infinispan.test.data.Person;
+import org.infinispan.test.fwk.TestCacheManagerFactory;
+import org.infinispan.util.KeyValuePair;
+import org.testng.annotations.Test;
+
+/**
+ * Test for receiving events in different formats when using a converter.
+ *
+ * @since 9.4
+ */
+@Test(groups = {"functional",}, testName = "client.hotrod.event.JsonKeyValueRawEventsTest")
+public class JsonKeyValueRawEventsTest extends SingleHotRodServerTest {
+
+   @Override
+   protected EmbeddedCacheManager createCacheManager() throws Exception {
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/infinispan/notifications/cachelistener/filter/CacheEventConverter.java
```diff
@@ -27,4 +27,13 @@ default MediaType format() {
       return MediaType.APPLICATION_OBJECT;
    }
 
+   /**
+    * @return if true, {@link #convert(Object, Object, Metadata, Object, Metadata, EventType)} will be presented with data
+    * in the request format rather than the format specified in {@link #format()}. The request format is defined as the MediaType
+    * that a cache was previously decorated with {@link org.infinispan.AdvancedCache#withMediaType(String, String)}.
+    */
+   default boolean useRequestFormat() {
```
danberindei: The converter and listener are already strongly coupled, e.g. you can't use `KeyValueVersionConverter` without knowing the format of the `byte[]` it produces. IMO it would be clearer if the listeners with converters ignored the request format completely and relied only on the converter's input/output format -- although I'm not sure how that would work for NearCache/Spark.

gustavonalle: > it would be clearer if the listeners with converters ignored the request format completely and relied only on the converter's input/output format

It would mean the converter itself would need to get hold of the transcoders and convert the data to the request format. This seems a bit complicated, as converters can be deployed by the user, and getting hold of component registry, obtaining transcoders, injecting, etc should not be exposed to the user IMO.

gustavonalle: So far the alternative solutions I though were:

* Assume converters produce custom formats that will not be transcoded to the request format, and let the clients do it all. This will bloat clients with lots of deps to support all formats, things  that the server can do it already :-1:  :-1:
* Create a special KeyValueConverter that has access to the request format and can perform the conversions by calling the appropriate transcoders. Downside: handling of this special converter everywhere  :-1: 
* This PR, "flag" to allow the converter to process data in the request format already :+1: 
* Have a way to send Keys and Values in an event, so I don't even need a converter to glue a Key and Value in a single byte[]. :+1: :+1: :+1:  Most clean solution, but I have a feeling it's not for a micro version, as we'd need to provide a new Converter interface

gustavonalle: > it should be easy to agree on the "inner" format as well without involving the request format at all.
Could you give an example?

danberindei: > > it should be easy to agree on the "inner" format as well without involving the request format at all.
> Could you give an example?

The test you added would have to deploy another converter that knows about JSON.
The eager cache listener would have to always receive keys and values in the JBoss Marshalling format (or use a different converter).

IMO `KeyValueVersionConverter` was meant to be a generic converter, but only as long as the client uses JBoss Marshalling. We shouldn't force it into something else.

gustavonalle: > The test you added would have to deploy another converter that knows about JSON.

One converter per each different K/V pair? And each converter will have to handle internally the conversion rather than delegate to the transcoders? Not sure about this...

> The eager cache listener would have to always receive keys and values in the JBoss Marshalling format (or use a different converter).

The client decides what format it wants, that's the basis of transcoding. Sometimes it deals in protobuf only, some other times in UTF-8, sometimes keys and values are in different formats, and even no transcoding at all. 

> IMO KeyValueVersionConverter was meant to be a generic converter, but only as long as the client uses JBoss Marshalling. We shouldn't force it into something else.

You're being too kind with ```KeyValueVersionConverter```, it is actually a hack, because there is no way to receive keys and values in an event :)

danberindei: > One converter per each different K/V pair? 

Yes, one converter for each mime type combination.

> And each converter will have to handle internally the conversion rather than delegate to the transcoders? Not sure about this...

The converter can still request a mime type via its `format()` method, so I don't see why it would have to handle the conversion internally.

> The client decides what format it wants, that's the basis of transcoding.

I guess that's the source of our disagreement. I see the converter as a part of the application that just happens to be deployed on the server instead of the client. So I don't think of transcoding as "the client can request anything, and any converter deployed on the server must be able to deliver something that the client understands", but as "the client can request anything, and one of the ways to request a certain format is to use a deployed converter with that exact format".

gustavonalle: > Yes, one converter for each mime type combination.

:disappointed:

> The converter can still request a mime type via its format() method, so I don't see why it would have to handle the conversion internally.

That's a good point, I actually tried to create a parameterised converter that receives the media type for keys and values, but hit a wall with  ```useRaw=true```, as it is supposed to have precedence over format(), and also affect the way parameters are sent, this is something that still need to change, probably as part of the remote listeners new API.Even if it worked, there's the issue described below...
 
> "the client can request anything, and one of the ways to request a certain format is to use a deployed converter with that exact format"

It's sure one way of doing it, but that would impose that the converter produces something in the exact format the client expects. There are actually plenty of tests where a converter deployed in the server produces a random POJO, and the client receives a jboss serialized byte[].

Actually, I'd rather not touch any of this stuff in a micro version. This PR was a 'conservative' way of supporting the Spark use case, if it helps I can mark the method as 'dangerous/experimental/advanced' otherwise we can just postpone the work to the next version.

danberindei: > [...] I actually tried to create a parameterised converter that receives the media type for keys and values, but hit a wall with `useRaw=true`, as it is supposed to have precedence over format(), and also affect the way parameters are sent [...]

IMO if `useRawData` and `format()` overlap, it should be impossible to use both, but of course there's no way to not use `format()`...

> It's sure one way of doing it, but that would impose that the converter produces something in the exact format the client expects. There are actually plenty of tests where a converter deployed in the server produces a random POJO, and the client receives a jboss serialized byte[].

Sure, there's a bit of magic to translate from jboss-marshalled objects to and from POJOs, but because the same magic happens on the client, the client listener ends up seeing exactly the POJO that the converter generated. So I don't think that invalidates my point.

> Actually, I'd rather not touch any of this stuff in a micro version. This PR was a 'conservative' way of supporting the Spark use case, if it helps I can mark the method as 'dangerous/experimental/advanced' otherwise we can just postpone the work to the next version.

`@Experimental` works for me, I know it's the least intrusive fix, but it seems like a hack on top of a hack, so I really don't want to commit to it as an API. And if you can, please throw an exception when the converter uses it together with a custom `format()` and/or `useRawData=true`.

gustavonalle: @danberindei Update, marking it as experimental

danberindei: Nitpicking: near cache already doesn't need ISPN-9635, because eager mode is no longer supported ;)

danberindei:
Integrated, thanks Gustavo!

core/src/main/java/org/infinispan/notifications/cachelistener/filter/CacheEventConverter.java
gustavonalle: Indeed, I saw that the converter was deprecated since years :)
Still, JCache needs it

PR #6336, opened 2018-10-12
ISPN-8616 Fix the replication wait in locked stream tests

danberindei:
https://issues.jboss.org/browse/ISPN-8616


wburns:

PR #6343, opened 2018-10-22
ISPN-9626 RocksDB documentation is not compiling because of properties

diegolovison:
ISPN-9626 RocksDB documentation is not compiling because of properties


tristantarrant:
Merged, thanks

PR #6349, opened 2018-10-23
ISPN-9613 Fix infinispan-bom

pruivo:
**Work in Progress**

**Changes**

* `./build-configuration/bom/pom.xml`: kept only the `org.infinispan*` dependencies that may be used by an user. **please review**
The dependencies in doubt are with a comment `TODO should be here?`
~~Should the `infinispan-spark` / `infinispan-hadoop` be here as well?~~ [NO!]

* ~~`./build-configuration/pom.xml`: all remaining `org.infinispan*` modules/tests. Can be imported by other external components.~~ [Moved to `./pom.xml`]

* `./pom.xml`: dump all the dependencies and transitive dependencies. IMO, the transitive dependencies can be removed. Suggestions?

**TODO**

* ~~Remove commentend xml~~
* ~~Cleanup properties~~


pruivo:
> I don't think we should have any test jars or bundles in the bom.

All the `test-jar` are already commented. The only test in bom is `infinispan-common-test`. I'm not sure if it should be here...

Removed `infinispan-hibernate-cache-v51` and `infinispan-persistence-cli` bundles! 
I guess `infinispan-marshaller-kryo-bundle` and `infinispan-marshaller-protostuff-bundle` should stay in bom, right?



> In the build-configuration parent you define `infinispan-cachestore-jdbc` test-jar, amongst others, int he dependencyManagement. This should just be in infinispan-parent IMO

yes I left them there in case it needs to be used by some external project (spark, hadoop connectors, etc...) I can move them `./pom.xml` if you think it is a better place although I would like to hear from others :)



> You also define `infinispan-embedded`, will that not cause issues? If using maven I thought our position was that we should discourage the use of the uber jar.

it is there because it is used by the `distribution` module to publish the uber jar. Except if I did some mistake, they aren't upload to maven.

build-configuration/bom/pom.xml
```diff
@@ -77,6844 +77,1188 @@
...
-        <versionx.org.infinispan.infinispan-javadoc-remote>${version.infinispan}</versionx.org.infinispan.infinispan-javadoc-remote>
-        <versionx.org.infinispan.infinispan-jcache>${version.infinispan}</versionx.org.infinispan.infinispan-jcache>
-        <versionx.org.infinispan.infinispan-jcache-commons>${version.infinispan}</versionx.org.infinispan.infinispan-jcache-commons>
-        <versionx.org.infinispan.infinispan-jcache>${version.infinispan}</versionx.org.infinispan.infinispan-jcache>
-        <versionx.org.infinispan.infinispan-jcache-remote>${version.infinispan}</versionx.org.infinispan.infinispan-jcache-remote>
-        <versionx.org.infinispan.infinispan-js-client>${version.infinispan}</versionx.org.infinispan.infinispan-js-client>
-        <versionx.org.infinispan.infinispan-license>${version.infinispan}</versionx.org.infinispan.infinispan-license>
-        <versionx.org.infinispan.infinispan-lucene-directory>${version.infinispan}</versionx.org.infinispan.infinispan-lucene-directory>
+        <versionx.org.infinispan.elasticsearch-indexmanager-it>9.4.1-SNAPSHOT</versionx.org.infinispan.elasticsearch-indexmanager-it>
```
tristantarrant: Do we really need all these individual versions ? They bloat the pom and contribute nothing

pruivo:
updated with "squashed" arquillian and wildfly version.

ryanemerson:
> it is there because it is used by the `distribution` module to publish the uber jar. Except if I did some mistake, they aren't upload to maven.

The problem isn't whether it's uploaded to maven, the problem is that it will be exposed as part of the infinispan-bom, hence exposed to the user when we want to discourage it. Maybe I'm splitting hairs though and it's not a big deal.

> `nfinispan-marshaller-kryo-bundle` and `infinispan-marshaller-protostuff-bundle` should stay in bom

:+1: 



> All the `test-jar` are already commented. The only test in bom is `infinispan-common-test`. I'm not sure if it should be here...

Aren't the test-jars defined in the `infinispan-build-configuration-parent` `<dependencyManagement>` section inherited by `infinispan-bom`? For example `infinispan-server-core` test-jar is in the parent.

I assume we're going to remove the comments before merging?

pruivo:
@ryanemerson I don't think maven will import the parent dependencies if `infinispan-bom` is imported.
I can move them to `./pom.xml`. At least, we have everything in the same place

When I get the green light, I'll remove all the comment deps and squash the commits into a single one.

pruivo:
@gustavonalle To be honest, I'm not sure. If I put the direct deps in bom, you will get the same issue as before. I guess the issue was with spring-boot (or spark connector?)  that depends on netty 3.x while our Hot Rod client depends on netty 4.x.

I guess if we want the bom usable, we would have to split in embedded, remote, jcache, etc.... 

I'm open to suggestions!!!

gustavonalle:
@pruivo on master, ```<artifactId>cache-api</artifactId>``` is in the dependencyMagmt section, while in this PR it's not. So it will break the tutorials again for a different reason than spark and spring boot.

gustavonalle:
@pruivo Answering your question in the first comment:

> Should the infinispan-spark / infinispan-hadoop be here as well?

Never! As far as https://github.com/infinispan/infinispan/ is concerned, infinispan-spark and infinispan-hadoop don't exist  :)

pruivo:
Updated! 
* Moved cache-api to bom
* Fixed the jsr107 test failures

pruivo:
squashed!

tristantarrant:
Can we please get rid of all the infinispan.brand.version* properties ?

tristantarrant:
If I force the version of the console to 9.4.0.Final, the build completes, but the distribution artifacts use the infinispan.brand.version stuff:
`infinispan/distribution/target/distribution:`
`infinispan-9.4.1-SNAPSHOT-all.zip`
`infinispan-9.4.1-SNAPSHOT-minimal.zip`
`infinispan-9.4.1-SNAPSHOT-remote.zip`
`infinispan-9.4.1-SNAPSHOT-src.zip`
`infinispan-server-9.4.1-SNAPSHOT.zip`
`infinispan-wildfly-modules-9.4.1-SNAPSHOT.zip`

tristantarrant:
in essence, drop the infinispan.brand.version major/minor/micro stuff, just have a single property and default it to project.version.

pruivo:
the `infinispan-management-console` it is my fault :P I replaced `versionx.org.infinispan*` by `version.infinispan`

pruivo:
@tristantarrant should be fixed!

pruivo:
I can squash the commits after the review :+1:

tristantarrant:
Looks good !

karesti:
@pruivo failures seem unrelated. I push the button ? 💣

pruivo:
@karesti go for it :) fire in the hole!

PR #6350, opened 2018-10-23
ISPN-3791 Silence "Received invalid rebalance confirmation from NodeX"

danberindei:
https://issues.jboss.org/browse/ISPN-3791

The warnings are completely harmless, change them to trace logs.


danberindei:
I added another commit to silence "ISPN000210: Failed to request state of cache X from node Y" exceptions as well. Hope it stays green :)

PR #6352, opened 2018-10-24
Javadoc ClientEventsOOMTest so we remember quickly what it tests :)

anistor:


wburns:

PR #6353, opened 2018-10-24
ISPN-9641 Using expiration with off heap doesn't subtract size properly

wburns:
with expiration

* Make sure to include expiration when not using version or custom

https://issues.jboss.org/browse/ISPN-9641

### DIFF
--- a/core/src/test/java/org/infinispan/container/offheap/OffHeapBoundedMemoryTest.java
+++ b/core/src/test/java/org/infinispan/container/offheap/OffHeapBoundedMemoryTest.java
@@ -3,6 +3,8 @@ package org.infinispan.container.offheap;
 import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertNull;
 
+import java.util.concurrent.TimeUnit;
+
 import org.infinispan.AdvancedCache;
 import org.infinispan.Cache;
 import org.infinispan.commons.CacheConfigurationException;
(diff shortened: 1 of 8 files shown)


core/src/test/java/org/infinispan/container/offheap/OffHeapBoundedMemoryTest.java
```diff
@@ -96,6 +98,59 @@ public void testAllocatedAmountEqualWithVersion() {
       assertEquals(allocator.getAllocatedAmount(), container.currentSize);
    }
 
+   public void testAllocatedAmountEqualWithExpiration() {
+      ConfigurationBuilder builder = new ConfigurationBuilder();
+      builder.memory()
+            .size(MemoryUnit.MEGABYTES.toBytes(20))
```
danberindei: Why not 2MB?

wburns: That is just what the others are using, tbh the value doesn't matter as it is never hit :)

core/src/test/java/org/infinispan/expiration/impl/ExpirationStoreListenerFunctionalTest.java
```diff
@@ -14,13 +14,21 @@
...
 
 @Test(groups = "functional", testName = "expiration.impl.ExpirationStoreListenerFunctionalTest")
 public class ExpirationStoreListenerFunctionalTest extends ExpirationStoreFunctionalTest {
    protected ExpiredCacheListener listener = new ExpiredCacheListener();
    protected ExpirationManager manager;
 
+   @Factory
+   public Object[] factory() {
+      return new Object[]{
+            new ExpirationStoreListenerFunctionalTest(),
```
danberindei: Missing `@Override` and maybe a comment to explain why we don't need super's parameters.

core/src/test/java/org/infinispan/container/offheap/OffHeapBoundedMemoryTest.java
danberindei: Ah right, I was thinking that off-heap preallocates the memory. Since only the `MemoryAddressHash` table is preallocated and that doesn't change with the memory size, I guess it doesn't matter.

core/src/test/java/org/infinispan/expiration/impl/ExpirationStoreListenerFunctionalTest.java
wburns: Sure, let me fix it.

wburns:
Updated.

danberindei:
Integrated, thanks Will!

PR #6354, opened 2018-10-24
[9.3.x] ISPN-9641 Using expiration with off heap doesn't subtract size properly

wburns:
with expiration

* Make sure to include expiration when not using version or custom

https://issues.jboss.org/browse/ISPN-9641


danberindei:
Integrated, thanks Will!

PR #6355, opened 2018-10-25
ISPN-8320 Javadoc comments contain malformed HTML

anistor:
* running Javadoc results in many warnings
* some author's email addresses contain < > which should be encoded as html entities
* some chars like ```&``` need to be encoded as html entities
* self closing ```<p/>``` tags are not allowed in javadocs

Many files were changed but since most of it was done with replace in all java files this should not cause much pain to validate.
Automatically replacing ```* <p/>``` with ```* <p>``` in all java files solved 99% of the issues. A handful of ```<p/>```embedded in text were solved manually. Email addresses were all solved with replace.

This still solves only some of the malformed javadoc problems, many more to go, so please keep the jira open.

### DIFF
--- a/cdi/common/src/main/java/org/infinispan/cdi/common/util/AbstractImmutableBean.java
+++ b/cdi/common/src/main/java/org/infinispan/cdi/common/util/AbstractImmutableBean.java
@@ -21,7 +21,7 @@ import org.infinispan.commons.logging.LogFactory;
  * collections are defensively copied on instantiation. It uses the defaults
  * from the specification for properties if not specified.
  * </p>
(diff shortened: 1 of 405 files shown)


tristantarrant:
Actually, <p> (unlike <br>) needs to be closed, so in theory we should be making sure it is paired with a corresponding </p>. The easiest fix is to replace all instances of <p/> with <p></p>

anistor:
Yup, doing that quickly now.

anistor:
closing this for now, replaced by https://github.com/infinispan/infinispan/pull/6358

PR #6357, opened 2018-10-25
ISPN-9638 Make Single port server a feature

wburns:
* Prevent startup when single port is configured but not enabled

https://issues.jboss.org/browse/ISPN-9638


tristantarrant:
A bit overengineered (it is code we will want to remove at some point). I think we'd better have a generic "feature x is disabled" log in commons' Log and just do if (!featureInstance.isAvailable(x)) throw log.featureDisabled(x)). I now see that PersistenceManager also has a specific log. Can you fix that too please ?

wburns:
Sure.

wburns:
Updated to use same log method.

tristantarrant:
Merged, thanks

PR #6359, opened 2018-10-26
QueryResponse has a proper externalizer, no need to add it to the mar…

anistor:
…shalling whitelist

* the marshaller was added by ISPN-9502


gustavonalle:
Merged, thanks @anistor !

PR #6362, opened 2018-10-30
[9.3] ISPN-8844 Avoid package overlap between commons and core

tristantarrant:
https://issues.jboss.org/browse/ISPN-8844


gustavonalle:
Merged, thanks @tristantarrant

PR #6365, opened 2018-10-30
ISPN-9373 Upgrade testng to 6.14.3

diegolovison:
https://issues.jboss.org/browse/ISPN-9373 Bump testng

### DIFF
--- a/jcache/embedded/src/test/java/org/infinispan/jcache/LimitExpiryFactoryTest.java
+++ b/jcache/embedded/src/test/java/org/infinispan/jcache/LimitExpiryFactoryTest.java
@@ -1,6 +1,6 @@
-package org.infinispan.jcache.embedded;
+package org.infinispan.jcache;
 
-import static org.junit.Assert.assertEquals;
+import static org.testng.AssertJUnit.assertEquals;
 
 import java.util.concurrent.TimeUnit;
 
@@ -8,7 +8,8 @@ import javax.cache.expiry.Duration;
 import javax.cache.expiry.EternalExpiryPolicy;
 import javax.cache.expiry.ExpiryPolicy;
 
-import org.junit.Test;
+import org.infinispan.jcache.embedded.LimitExpiryFactory;
+import org.testng.annotations.Test;
 
 public class LimitExpiryFactoryTest {
 
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/CustomMarshallerEventIT.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/CustomMarshallerEventIT.java
@@ -1,7 +1,7 @@
 package org.infinispan.client.hotrod.event;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertNotNull;
 
 import java.io.Serializable;
 import java.util.concurrent.ArrayBlockingQueue;
--- a/core/src/test/java/org/infinispan/configuration/JsonSerializationTest.java
+++ b/core/src/test/java/org/infinispan/configuration/JsonSerializationTest.java
@@ -38,7 +38,7 @@ import org.infinispan.transaction.LockingMode;
 import org.infinispan.transaction.TransactionMode;
 import org.infinispan.transaction.TransactionProtocol;
 import org.infinispan.util.concurrent.IsolationLevel;
-import org.junit.Assert;
+import org.testng.Assert;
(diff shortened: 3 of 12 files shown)


pruivo:
needs rebase :)

diegolovison:
Please don't integrate yet. I would like to compare the Jenkins Test Result with the commit d0bef987f5

diegolovison:
I am able to reproduce "some" errors on my local. I will work on that.

pruivo:
why adding `junit` dependency? we use `testng` in most of the places :)

diegolovison:
@pruivo with the new testng version, the "junit dependency" goes away. I would like to keep the code as close as possible to reduce the noise.

For example: `FunctionalMapTest` it is failing only when you run: `mvn verify -Dtest=FunctionalMapTest`. If you try to debug it works. if you run one method only, it works.

I can remove the `junit` dependency when the test start working again.

diegolovison:
Please don't integrate yet.

integrationtests/security-it/pom.xml
```diff
@@ -83,13 +83,7 @@
...
             <artifactId>maven-failsafe-plugin</artifactId>
             <configuration>
-               <!-- Force use of JUnit, since TestNG+Arquillian break in wonderful ways -->
-               <testNGArtifactName>none:none</testNGArtifactName>
-               <forkCount>1</forkCount>
-               <reuseForks>false</reuseForks>
-               <properties>
-                  <listener>${junitListener}</listener>
-               </properties>
+               <forkMode>once</forkMode>
```
danberindei: `forkMode` is [deprecated](http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#forkMode), and are you sure we don't need to force the use of JUnit any more?

spring/spring4/spring4-remote/src/test/java/org/infinispan/spring/provider/SpringRemoteCacheManagerFactoryBeanContextTest.java
```diff
@@ -1,6 +1,6 @@
 package org.infinispan.spring.provider;
 
-import static org.junit.Assert.assertEquals;
+import static org.testng.Assert.assertEquals;
```
danberindei: We should use `AssertJUnit`, which is part of TestNG but keeps the same parameters as the JUnit `Assert`

danberindei:
> @pruivo with the new testng version, the "junit dependency" goes away. 

It took me a while to parse this sentence, with surefire selecting the provider based on what's on the classpath I never would have thought that TestNG has a dependency on junit :)

Anyway, we should remove the dependency, but we should be using `org.testng.AssertJUnit` instead of `org.testng.Assert`, because we still have dreams of switching to JUnit everywhere in the future.

tristantarrant:
Can this be updated ?

diegolovison:
There are some changes like
```
org.junit.Assert.assertEquals
```
to
```
org.testng.AssertJUnit.assertEquals
```

Testng had junit as a dependency, now with the version `6.14.3` the Junit dependency didn't exists.
Should I add the Junit dependency in the projects?

danberindei:
I say no, we should use `AssertJUnit`

tristantarrant:
AssertJUnit everywhere please.

diegolovison:
Cool. My PR is ready for review

jcache/embedded/src/test/java/org/infinispan/jcache/LimitExpiryFactoryTest.java
```diff
@@ -1,14 +1,15 @@
-package org.infinispan.jcache.embedded;
+package org.infinispan.jcache;
```
diegolovison: Weird change

diegolovison:
Ready for review

danberindei:
Not sure if you've noticed @diegolovison @tristantarrant, but the all the test methods are now displayed in IntelliJ as `TestClassName (number)`.

`FunctionalListenersTest` is also failing since the upgrade, probably because the newer TestNG is more strict about running configuration methods from another group.

diegolovison:
@danberindei 

Adding the following methods in `FunctionalListenersTest`, the NPE from `mvn verify -Dtest=FunctionalListenersTest` went away

```java
   @Override
   @BeforeClass
   public void createBeforeClass() throws Throwable {
      super.createBeforeClass();
   }

   @Override
   @BeforeMethod
   public void createBeforeMethod() throws Throwable {
      super.createBeforeMethod();
   }
```

But, it doesn't fix the IntelliJ issue

diegolovison:
About IntelliJ see: https://youtrack.jetbrains.com/issue/IDEA-208083

PR #6374, opened 2018-11-05
ISPN-9677 Check tx running when updating query cache

galderz:
https://issues.jboss.org/browse/ISPN-9677


galderz:
@rvansa Mind reviewing this? It's breaking Hibernate simple tutorials

rvansa:
After reading the javadoc for `getCacheTransactionSynchronization()` three times I noticed the caveat:

> This **may** return null when the Session is not currently part of a transaction

I thought that not in tx implies `sync == null`. Thanks for fixing this, Galder!

PR #6375, opened 2018-11-05
ISPN-9511 Expired event is not raised when modifying an expired entry

gustavolira:
[https://issues.jboss.org/browse/ISPN-9511](https://issues.jboss.org/browse/ISPN-9511)


wburns:
Test looks good @gustavolira, thanks!

If you could tweak the commit message to be in the format, then I will integrate it.
```
ISPN-XXXX <JIRA Desc>
 * <notes>
 * <additional notes>

danberindei:
@wburns actually there should be an empty line between the `ISPN-XXXX` line and the body ;)

wburns:
Integrated into master, thanks @gustavolira !

PR #6379, opened 2018-11-07
ISPN-7863 Create a test that really fails

anistor:
The fix is already in but the existing test does not demonstrate the failure, so here's another take on it.

https://issues.jboss.org/browse/ISPN-7863

Please backport this to 9.4.x also.


anistor:
Merged in master and 9.4.x

PR #6385, opened 2018-11-09
ISPN-9697 Licenses

tristantarrant:
https://issues.jboss.org/browse/ISPN-9697


tristantarrant:
Some heads-up:
@ryanemerson @oraNod @jsenko this changes the packaging of the wildfly/eap zip: it brands the name, it has a root directory which corresponds to the zip name, it moves all modules under the modules subdirectory, it includes the licenses and a README. I have amended the docs to describe this.
All jars now include a META-INF/licenses.xml file in the "blessed" format. Uberjars contain an aggregated version with all licenses.
The server licenses are contained in a docs/licenses/server-licenses.xml which goes alongside the other license files packaged by WildFly)

tristantarrant:
Did you build the server with -Pdistribution ? No ? That's it then :smile:

ryanemerson:
I did `mvn clean install -DskipTests -Pdistribution` on the root dir

tristantarrant:
Fixed and made some further clean ups to the licenses.xsl mapping

PR #6386, opened 2018-11-09
ISPN-6728 Add ability to override BooleanQuery.maxClauseCount

anistor:
https://issues.jboss.org/browse/ISPN-6728

### DIFF
--- a/query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
+++ b/query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
@@ -16,6 +16,7 @@ import java.util.concurrent.ConcurrentMap;
 import javax.management.MBeanServer;
 import javax.management.ObjectName;
 
+import org.apache.lucene.search.BooleanQuery;
 import org.hibernate.search.analyzer.definition.LuceneAnalysisDefinitionProvider;
 import org.hibernate.search.cfg.spi.SearchConfiguration;
 import org.hibernate.search.engine.service.classloading.spi.ClassLoaderService;
@@ -107,6 +108,13 @@ public class LifecycleManager implements ModuleLifecycle {
    private static final Log log = LogFactory.getLog(LifecycleManager.class, Log.class);
 
    /**
+    * Optional integer system property that sets value of {@link BooleanQuery#setMaxClauseCount}.
+    */
+   public static final String MAX_BOOLEAN_CLAUSES_SYS_PROP = "infinispan.query.lucene.max-boolean-clauses";
+
+   private static boolean maxBooleanClausesWasSet = false;
+
+   /**
     * Caching the looked-up MBeanServer for the lifetime of the cache manager is safe.
     */
    private MBeanServer mbeanServer;
@@ -123,6 +131,8 @@ public class LifecycleManager implements ModuleLifecycle {
          SearchIntegrator searchFactory = null;
          boolean isIndexed = cfg.indexing().index().isEnabled();
          if (isIndexed) {
+            setBooleanQueryMaxClauseCount();
+
             cr.registerComponent(new ShardAllocationManagerImpl(), ShardAllocatorManager.class);
             searchFactory = createSearchIntegrator(cfg.indexing(), cr, aggregatedClassLoader);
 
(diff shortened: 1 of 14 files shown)


karesti:
java.lang.AssertionError: Expected test to throw (an instance of org.infinispan.client.hotrod.exceptions.HotRodClientException and exception with message a string containing "org.apache.lucene.search.BooleanQuery$TooManyClauses: maxClauseCount is set to 1025")

query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
```diff
@@ -475,4 +485,31 @@ public void cacheManagerStarting(GlobalComponentRegistry gcr, GlobalConfiguratio
    public void cacheManagerStopped(GlobalComponentRegistry gcr) {
       mbeanServer = null;
    }
+
+   /**
+    * Sets {@link BooleanQuery#setMaxClauseCount} according to the value of {@link #MAX_BOOLEAN_CLAUSES_SYS_PROP} system
+    * property. This is executed only once, when first indexed cache is started.
+    */
+   private void setBooleanQueryMaxClauseCount() {
+      if (!maxBooleanClausesWasSet) {
```
karesti: I would do 
```java 
if(maxBooleanClausesWasSet) return;
```
And then the bloc

query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
```diff
@@ -475,4 +485,31 @@ public void cacheManagerStarting(GlobalComponentRegistry gcr, GlobalConfiguratio
...
+         maxBooleanClausesWasSet = true;
+         String maxClauseCountProp = SecurityActions.getSystemProperty(MAX_BOOLEAN_CLAUSES_SYS_PROP);
+         if (maxClauseCountProp != null) {
+            int maxClauseCount;
+            try {
+               maxClauseCount = Integer.parseInt(maxClauseCountProp);
+            } catch (NumberFormatException e) {
+               log.failedToParseSystemProperty(MAX_BOOLEAN_CLAUSES_SYS_PROP, e);
+               throw e;
```
karesti: Could be prefer to launch an exception containing a configuration exception with the property name on it? Instead a numberformat.

query/src/main/java/org/infinispan/query/impl/SecurityActions.java
```diff
@@ -29,6 +29,10 @@ private SecurityActions() {
             AccessController.doPrivileged(action) : Security.doPrivileged(action);
    }
 
+   static String getSystemProperty(String propertyName) {
```
karesti: Won't be better to return Optional<String> ? and then deal with it in the code. Just suggesting

query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
anistor: I do not like returning from the middle of the method. Unless the method is long and boring, then I do return abruptly.

query/src/main/java/org/infinispan/query/impl/SecurityActions.java
anistor: I hate Optional. Sorry :)

query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
karesti: is not the middle of the method, is the beginning! as there are additional if statements inside, imho flattens it helps to understand, but ok, it's simple enough like this

query/src/main/java/org/infinispan/query/impl/SecurityActions.java
karesti: there such good tings to do with Optional, specially since java 9 even moreee! but ok 😄

query/src/main/java/org/infinispan/query/impl/LifecycleManager.java
anistor: This is not a cache manager / cache configuration related error, it's a command line system property, so throwing a CacheConfigurationException would not be appropriate here. I would like to also avoid inventing a new exception class. Or do we already have one for such situations?

PR #6390, opened 2018-11-12
ISPN-9647 adding default analyzers

oraNod:
https://issues.jboss.org/browse/ISPN-9647

@gustavonalle and @anistor review, please. ISPN docs don't really cover analysis and I'm not sure what level of detail we need to go into here. maybe it will be sufficient to mention the default analyzers for now and wait to add more detail on using them when we address the larger issues with querying and indexing in docs. thanks!


tristantarrant:
Can you please merge the commits ?

PR #6393, opened 2018-11-12
ISPN-9702 ClassCastException when using converters in remote events

gustavonalle:
https://issues.jboss.org/browse/ISPN-9702

### DIFF
--- a/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/CustomEventLogListener.java
+++ b/client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/CustomEventLogListener.java
@@ -238,6 +238,30 @@ public abstract class CustomEventLogListener<K, E> implements RemoteCacheSupplie
       public DynamicCustomEventWithStateLogListener(RemoteCache<K, ?> r) { super(r); }
    }
 
+   @ClientListener(converterFactoryName = "simple-converter-factory")
+   public static class SimpleListener<K> extends CustomEventLogListener<K, String> {
+      public SimpleListener(RemoteCache<K, ?> r) {
+         super(r);
+      }
+   }
+
+   @NamedFactory(name = "simple-converter-factory")
+   static class SimpleConverterFactory<K> implements CacheEventConverterFactory {
+      @Override
+      @SuppressWarnings("unchecked")
+      public CacheEventConverter<String, String, CustomEvent> getConverter(Object[] params) {
+         return new SimpleConverter();
+      }
+
+      static class SimpleConverter<K> implements CacheEventConverter<K, String, String>, Serializable, ExternalPojo {
+         @Override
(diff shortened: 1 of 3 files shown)


client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/CustomEventLogListener.java
```diff
@@ -238,6 +238,30 @@ private void expectTimeOrdered(CustomEvent before, CustomEvent after) {
...
+      @Override
+      @SuppressWarnings("unchecked")
+      public CacheEventConverter<String, String, CustomEvent> getConverter(Object[] params) {
+         return new SimpleConverter();
+      }
+
+      static class SimpleConverter<K> implements CacheEventConverter<K, String, String>, Serializable, ExternalPojo {
+         @Override
+         public String convert(K key, String oldValue, Metadata oldMetadata, String newValue, Metadata newMetadata, EventType eventType) {
+            if (newValue != null) return newValue;
```
karesti: I prefer ```return newValue==null? oldValue:newValue;```

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/CustomEventLogListener.java
```diff
@@ -238,6 +238,30 @@ private void expectTimeOrdered(CustomEvent before, CustomEvent after) {
       public DynamicCustomEventWithStateLogListener(RemoteCache<K, ?> r) { super(r); }
    }
 
+   @ClientListener(converterFactoryName = "simple-converter-factory")
```
karesti: And wondering if these names could be placed in a constant somewhere

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/CustomEventLogListener.java
gustavonalle: This JIRA has 3 PRs, so I since this is only a test, let's keep it "as is"  :)

client/hotrod-client/src/test/java/org/infinispan/client/hotrod/event/CustomEventLogListener.java
gustavonalle: I also noticed some formatting is due in those listener tests, but this is not related to the JIRA, I will send another PR later (master only)

karesti: Ok! if jenkins is ok I will merge the 3 of them unless you need an additional review

gustavonalle: Thanks @karesti !

PR #6397, opened 2018-11-13
ISPN-9708 Expose the executor services through JMX

tristantarrant:
https://issues.jboss.org/browse/ISPN-9708

This adds MBeans for all global executors (notifications, timeout, async, etc) as well as for the worker executors associated with the protocol servers. Since only the Hot Rod server uses the worker executor, I am setting the maximum pool size for REST and Memcached to 1 (as 0 is not a valid argument).

### DIFF
--- a/core/src/main/java/org/infinispan/executors/ManageableExecutorService.java
+++ b/core/src/main/java/org/infinispan/executors/ManageableExecutorService.java
@@ -1,0 +1,129 @@
+package org.infinispan.executors;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import org.infinispan.factories.scopes.Scope;
+import org.infinispan.factories.scopes.Scopes;
+import org.infinispan.jmx.annotations.DataType;
+import org.infinispan.jmx.annotations.DisplayType;
+import org.infinispan.jmx.annotations.MBean;
+import org.infinispan.jmx.annotations.ManagedAttribute;
+
+/**
+ * @author Tristan Tarrant &lt;tristan@infinispan.org&gt;
+ * @since 10.0
+ **/
+@MBean
+@Scope(Scopes.GLOBAL)
+public abstract class ManageableExecutorService<T extends ExecutorService> {
+
+   protected T executor;
+
+   @ManagedAttribute(
+         description = "Returns the number of threads in this executor.",
+         displayName = "Number of executor threads",
+         dataType = DataType.TRAIT,
+         displayType = DisplayType.SUMMARY
+   )
+   int getPoolSize() {
+      if (executor instanceof ThreadPoolExecutor) {
+         return ((ThreadPoolExecutor) executor).getPoolSize();
+      } else {
+         return -1;
+      }
+   }
+
+   @ManagedAttribute(
+         description = "Returns the number of active executor threads.",
+         displayName = "Number of active executor threads",
+         dataType = DataType.TRAIT,
+         displayType = DisplayType.SUMMARY
+   )
+   int getActiveCount() {
+      if (executor instanceof ThreadPoolExecutor) {
+         return ((ThreadPoolExecutor) executor).getActiveCount();
+      } else {
+         return -1;
+      }
+   }
+
+   @ManagedAttribute(
+         description = "Returns the maximum number of executor threads.",
+         displayName = "Maximum number of executor threads",
+         dataType = DataType.TRAIT,
+         displayType = DisplayType.SUMMARY,
+         writable = true
+   )
+   int getMaximumPoolSize() {
+      if (executor instanceof ThreadPoolExecutor) {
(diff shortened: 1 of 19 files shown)


pruivo:
isn't the `BlockingTaskAwareExecutorService` missing?

tristantarrant:
@pruivo Added !

commons/src/main/java/org/infinispan/commons/jmx/JmxUtil.java
```diff
@@ -78,7 +78,7 @@ public static void registerMBean(Object mbean, ObjectName objectName, MBeanServe
     * @throws Exception If unregistration could not be completed.
     */
    public static void unregisterMBean(ObjectName objectName, MBeanServer mBeanServer) throws Exception {
-      if (mBeanServer.isRegistered(objectName)) {
+      if (objectName != null && mBeanServer.isRegistered(objectName)) {
```
danberindei: IMO the caller should do this check, in fact most callers already do.

core/src/main/java/org/infinispan/executors/ManageableExecutorService.java
```diff
@@ -0,0 +1,87 @@
...
+      if (executor instanceof ThreadPoolExecutor) {
+         return ((ThreadPoolExecutor) executor).getMaximumPoolSize();
+      } else {
+         return -1;
+      }
+   }
+
+   void setMaximumPoolSize(int maximumPoolSize) {
+      if (executor instanceof ThreadPoolExecutor) {
+         ((ThreadPoolExecutor) executor).setMaximumPoolSize(maximumPoolSize);
```
danberindei: For executors that have a queue, we set the core size to be the same as the maximum size, so setting the maximum size should also change the core size.

Maybe we also need attributes for the queue size and the maximum queue size.

server/core/src/main/java/org/infinispan/server/core/AbstractProtocolServer.java
```diff
@@ -88,48 +97,75 @@ protected void startTransport() {
...
-   protected void registerTransportMBean() {
+   protected ThreadPoolExecutor getExecutor() {
+      if (this.executor == null || this.executor.isShutdown()) {
+         DefaultThreadFactory factory = new DefaultThreadFactory(getQualifiedName() + "-ServerHandler");
+         int workerThreads = getWorkerThreads();
+         this.executor = new ThreadPoolExecutor(
+               workerThreads,
+               workerThreads,
+               0L, TimeUnit.MILLISECONDS,
+               new LinkedBlockingQueue<>(),
```
danberindei: Not sure about having an unlimited queue here.

server/memcached/src/main/java/org/infinispan/server/memcached/MemcachedServer.java
```diff
@@ -76,6 +76,8 @@ public void stop() {
 
    @Override
    public int getWorkerThreads() {
-      return Integer.getInteger("infinispan.server.memcached.workerThreads", configuration.workerThreads());
+      //return Integer.getInteger("infinispan.server.memcached.workerThreads", configuration.workerThreads());
+      return 1; // Unused for now, so just return the smallest possible valid value
```
danberindei: IMO the end-of-line comment should be a regular comment, and we don't need the commented-out line. In fact, now that thread information is exposed separately, maybe we should deprecate `getWorkerThreads()`?

server/core/src/main/java/org/infinispan/server/core/AbstractProtocolServer.java
tristantarrant: That's how it has always been. The intent of this PR is not to change that behaviour atm.

commons/src/main/java/org/infinispan/commons/jmx/JmxUtil.java
tristantarrant: Ok, will remove it here and add it where it's missing in the callers.

core/src/main/java/org/infinispan/executors/ManageableExecutorService.java
tristantarrant: Ok for the core size.
As for the queue, I'll add getQueueSize() and getQueueRemainingCapacity() which map to the underlying equivalent methods in the queue impl.

tristantarrant:
@danberindei updated

core/src/main/java/org/infinispan/executors/ManageableExecutorService.java
danberindei: Unfortunately it's still not correct, the core size is 0 for executors that don't have a queue.

So we only need to update the core size if a queue is present, and we should probably add a comment about how bad ThreadPoolExecutor is at creating new threads when a queue is present. Maybe even a link to https://issues.jboss.org/browse/ISPN-4585, although it needs a more exotic example now that prepare commands don't block threads (e.g. state transfer, which still blocks a lot and has commands depending on each other).

tristantarrant:
Green !

danberindei:
Integrated, thanks Tristan!

The latest build failed in CI, but that was only because of a hung server.

PR #6411, opened 2018-11-15
ISPN-9720 JDK11: An illegal reflective access operation has occurred

TomasHofman:
https://issues.jboss.org/browse/ISPN-9720
https://issues.jboss.org/browse/WFLY-11355

Upstream PR: https://github.com/infinispan/infinispan/pull/6426

This is to avoid illegal access warnings on JDK 10+. Three classes were changed, which used reflection to access private fields in JDK types:
* IntSummaryStatisticsExternalizer
* LongSummaryStatisticsExternalizer
* DoubleSummaryStatisticsExternalizer (see bellow for details on this one)

Since JDK 10, relevant JDK types ([Int|Long|Double]SummaryStatistics) have gotten new parametrized constructors, so reflective access to set their inner state is no longer necessary.

The `DoubleSummaryStatistics` case is however somewhat problematic - except for `sum`, `count`, `min` and `max` fields (that are accessible by public API), the externalizer used to also access and serialize `sumCompensation` and `simpleSum` fields, which are technically implementation specific fields (i.e. do not have to exist in all JDKs) and are not accessible by the API. The purpose of those fields is to improve precision of the summation. *This change omits those fields during (de)serialization, which may lead to lower precision of deserialized instance compared to the original instance.*

### DIFF
--- a/core/src/main/java/org/infinispan/marshall/exts/DoubleSummaryStatisticsExternalizer.java
+++ b/core/src/main/java/org/infinispan/marshall/exts/DoubleSummaryStatisticsExternalizer.java
@@ -3,6 +3,7 @@ package org.infinispan.marshall.exts;
 import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.util.DoubleSummaryStatistics;
 import java.util.Set;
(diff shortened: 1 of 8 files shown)


wburns:
ok to test

TomasHofman:
@wburns Thanks, if you'd prefer multi version let me know and I can rework. I guess that would mean what you have in commons module - second source dir. Although I'm not sure if java10 classes can be tested in this multi version setup.

core/src/main/java/org/infinispan/marshall/exts/DoubleSummaryStatisticsExternalizer.java
```diff
@@ -28,18 +33,35 @@
...
+         countField = SecurityActions.getField(DoubleSummaryStatistics.class, "count");
+         sumField = SecurityActions.getField(DoubleSummaryStatistics.class, "sum");
+         sumCompensationField = SecurityActions.getField(DoubleSummaryStatistics.class, "sumCompensation");
+         simpleSumField = SecurityActions.getField(DoubleSummaryStatistics.class, "simpleSum");
+         minField = SecurityActions.getField(DoubleSummaryStatistics.class, "min");
+         maxField = SecurityActions.getField(DoubleSummaryStatistics.class, "max");
+
+         // We can only properly serialize if all of the fields are non null
+         canSerialize = countField != null && sumField != null && sumCompensationField != null && simpleSumField != null
+                 && minField != null && maxField != null;
```
danberindei: Do you really need the `sumCompensation` and `simpleSum` fields? If we're ignoring them on JDK10+, we could just ignore them everywhere.

core/src/main/java/org/infinispan/marshall/exts/DoubleSummaryStatisticsExternalizer.java
```diff
@@ -56,33 +78,51 @@ private void verifySerialization() {
...
+      final DoubleSummaryStatistics summaryStatistics;
+      if (constructor != null) {
+         // JDK 10+, pass values to constructor
+
+         final long count = input.readLong();
+         final double sum = input.readDouble();
+         final double sumCompensation = input.readDouble(); // ignored, see writeObject()
+         final double simpleSum = input.readDouble(); // ignored, see writeObject()
+         final double min = input.readDouble();
+         final double max = input.readDouble();
```
danberindei: I'd move the stream reads before the `if`, so it's clearer that the format is the same regardless of JDK version.

core/src/main/java/org/infinispan/marshall/exts/DoubleSummaryStatisticsExternalizer.java
TomasHofman: Yeah, the method handles are not needed anymore.

core/src/main/java/org/infinispan/marshall/exts/DoubleSummaryStatisticsExternalizer.java
TomasHofman: Good point.

TomasHofman:
I updated the PR. I will create an upstream PR as well now.

TomasHofman:
Updated to force retest.

wburns:
Integrated into 9.4.x, thanks @TomasHofman !

PR #6416, opened 2018-11-15
[9.3.x] ISPN-9725 Blackbox query tests are failing due to improper teardown that does not take store and index into account

anistor:
https://issues.jboss.org/browse/ISPN-9725


tristantarrant:
Merged

PR #6420, opened 2018-11-16
ISPN-9732 Local iteration optimization with write behind is valid for

wburns:
non shared stores

* Only prevent local optimization for shared stores with write behind

https://issues.jboss.org/browse/ISPN-9732


danberindei:
I ran the core tests locally and they passed :)

PR #6431, opened 2018-11-19
[9.4.x] ISPN-9708 Fix remote threadpool attributes in JMX

danberindei:
https://issues.jboss.org/browse/ISPN-9708

Fix remote command executor attributes


danberindei:
I pushed it myself, thanks Will!

PR #6433, opened 2018-11-19
ISPN-9742 Analyzers not working for inner objects

gustavonalle:
https://issues.jboss.org/browse/ISPN-9742

Reproducer only


remote-query/remote-query-server/src/main/java/org/infinispan/query/remote/impl/indexing/ProtobufValueWrapperAnalyzerDiscriminator.java
```diff
@@ -17,11 +18,14 @@
 
    @Override
    public String getAnalyzerDefinitionName(Object value, Object entity, String field) {
+      if (field == null) return null;
```
anistor: `field` is a Lucene field name, so it should be always non-null. Did we get a NPE here?

gustavonalle: I can't remember if there was a NPE or if it was a change to improve safety

PR #6440, opened 2018-11-20
ISPN-9699 0.0.f Global Capacity Factor

karesti:
https://issues.jboss.org/browse/ISPN-9699

depends on https://github.com/infinispan/infinispan/pull/6546

Backport to 9.4.x: https://github.com/infinispan/infinispan/pull/6633

### DIFF
--- a/core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
@@ -34,6 +34,7 @@ public class GlobalConfigurationBuilder implements GlobalConfigurationChildBuild
    private final Map<Class<?>, Builder<?>> modules;
    private final SiteConfigurationBuilder site;
    private Optional<String> defaultCacheName;
+   private boolean zeroCapacityNode;
    private Features features;
 
    public GlobalConfigurationBuilder() {
@@ -57,6 +58,7 @@ public class GlobalConfigurationBuilder implements GlobalConfigurationChildBuild
       this.asyncThreadPool = new ThreadPoolConfigurationBuilder(this);
       this.modules = new LinkedHashMap();
       this.defaultCacheName = Optional.empty();
+      this.zeroCapacityNode = false;
    }
 
    /**
@@ -165,6 +167,11 @@ public class GlobalConfigurationBuilder implements GlobalConfigurationChildBuild
       return (T)modules.get(moduleClass);
    }
 
+   public GlobalConfigurationBuilder zeroCapacityNode(boolean zeroCapacityNode) {
+      this.zeroCapacityNode = zeroCapacityNode;
+      return this;
+   }
+
    public GlobalConfigurationBuilder clearModules() {
       modules.clear();
       return this;
@@ -259,7 +266,8 @@ public class GlobalConfigurationBuilder implements GlobalConfigurationChildBuild
             site.create(),
             defaultCacheName,
             cl,
-            features);
+            features,
+            zeroCapacityNode);
    }
 
    public GlobalConfigurationBuilder read(GlobalConfiguration template) {
--- a/core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
+++ b/core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
@@ -167,6 +167,16 @@ public class GlobalConfigurationBuilder implements GlobalConfigurationChildBuild
       return (T)modules.get(moduleClass);
    }
 
+   /**
+    * Set the zero capacity node to true to configure a global capacity factor 0.0f for every distributed cache.
+    * The node will join the cluster but won't keep data on it.
+    * However, this flag does not affect replicated caches.
+    * Replicated caches will continue to keep copies of the data in this node.
+    * Use only distributed caches to make the best use of this feature.
+    *
+    * @param zeroCapacityNode value, true or false
+    * @return GlobalConfigurationBuilder instance
+    */
    public GlobalConfigurationBuilder zeroCapacityNode(boolean zeroCapacityNode) {
       this.zeroCapacityNode = zeroCapacityNode;
       return this;
--- a/core/src/main/java/org/infinispan/configuration/parsing/Parser.java
+++ b/core/src/main/java/org/infinispan/configuration/parsing/Parser.java
@@ -628,11 +628,15 @@ public class Parser implements ConfigurationParser {
                break;
             }
             case STATISTICS: {
-               builder.globalJmxStatistics().enabled(Boolean.valueOf(value));
+               builder.globalJmxStatistics().enabled(Boolean.parseBoolean(value));
                break;
             }
             case SHUTDOWN_HOOK: {
                builder.shutdown().hookBehavior(ShutdownHookBehavior.valueOf(value));
+               break;
+            }
+            case ZERO_CAPACITY_NODE: {
+               builder.zeroCapacityNode(Boolean.parseBoolean(value));
                break;
             }
             default: {
--- a/core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
+++ b/core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
@@ -94,6 +94,8 @@ public class StateTransferManagerImpl implements StateTransferManager {
          persistentStateChecksum = Optional.empty();
       }
 
+      float capacityFactor = globalConfiguration.isZeroCapacityNode() ? 0.0f : configuration.clustering().hash().capacityFactor();
+
       CacheJoinInfo joinInfo = new CacheJoinInfo(pickConsistentHashFactory(globalConfiguration, configuration),
             configuration.clustering().hash().hash(),
             configuration.clustering().hash().numSegments(),
@@ -101,7 +103,7 @@ public class StateTransferManagerImpl implements StateTransferManager {
             configuration.clustering().stateTransfer().timeout(),
             configuration.transaction().transactionProtocol().isTotalOrder(),
             configuration.clustering().cacheMode(),
-            configuration.clustering().hash().capacityFactor(),
+            capacityFactor,
             localTopologyManager.getPersistentUUID(),
             persistentStateChecksum);
 
--- a/core/src/main/resources/schema/infinispan-config-10.0.xsd
+++ b/core/src/main/resources/schema/infinispan-config-10.0.xsd
@@ -315,6 +315,11 @@
         <xs:documentation>Indicates the default cache for this cache container</xs:documentation>
       </xs:annotation>
     </xs:attribute>
+    <xs:attribute name="zero-capacity-node" type="xs:boolean" default="false">
+      <xs:annotation>
+        <xs:documentation>If 'true' then no data is stored in this node. Defaults to 'false'.</xs:documentation>
+      </xs:annotation>
+    </xs:attribute>
     <xs:attribute name="start">
       <xs:annotation><xs:documentation>Unused XML attribute</xs:documentation></xs:annotation>
     </xs:attribute>
--- a/core/src/test/java/org/infinispan/distribution/ZeroCapacityNodeTest.java
+++ b/core/src/test/java/org/infinispan/distribution/ZeroCapacityNodeTest.java
@@ -1,0 +1,59 @@
+package org.infinispan.distribution;
+
+import static org.testng.AssertJUnit.assertEquals;
+
+import java.util.Map;
+
+import org.infinispan.configuration.cache.CacheMode;
+import org.infinispan.configuration.cache.ConfigurationBuilder;
+import org.infinispan.configuration.global.GlobalConfigurationBuilder;
+import org.infinispan.distribution.ch.ConsistentHash;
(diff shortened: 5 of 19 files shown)


karesti:
I'm wondering about the security (Secured Actions etc)

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
```diff
@@ -122,6 +125,20 @@ public void rebalance(CacheTopology cacheTopology) {
       }
    }
 
+   private Optional<Float> globalCapacityFactor() {
+      String globalCapacityFactorProperty = System.getProperty(GLOBAL_CAPACITY_FACTOR);
```
danberindei: As Tristan said, this should be an on-off switch, and it would be better to make it a global setting protected by a feature flag.

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
```diff
@@ -94,14 +94,16 @@ public void start() throws Exception {
          persistentStateChecksum = Optional.empty();
       }
 
+      float capacityFactor = globalConfiguration.isZeroCapacityNode()? 0.0f : configuration.clustering().hash().capacityFactor();
```
tristantarrant: Formatting (missing a space before the ?

core/src/main/resources/schema/infinispan-config-10.0.xsd
```diff
@@ -266,6 +266,11 @@
         <xs:documentation>Indicates the default cache for this cache container</xs:documentation>
       </xs:annotation>
     </xs:attribute>
+    <xs:attribute name="zero-capacity-node" type="xs:boolean" default="false">
+      <xs:annotation>
+        <xs:documentation>If 'false' then no data is stored in this node. Defaults to 'false'.</xs:documentation>
```
tristantarrant: Maybe add a note mentioning that it is equivalent to setting capacity factor 0 on all clustered caches.

lock/src/main/java/org/infinispan/lock/impl/ClusteredLockModuleLifecycle.java
```diff
@@ -66,7 +66,10 @@ public void cacheManagerStarted(GlobalComponentRegistry gcr) {
 
    private static Configuration createClusteredLockCacheConfiguration() {
       ConfigurationBuilder builder = new ConfigurationBuilder();
-      builder.clustering().cacheMode(CacheMode.REPL_SYNC)
+      // ISPN-9699: We want the highest consistency, but we can't apply capacity factor 0 to replicated caches
+      // This is a requirement for the Vert.x Cluster Manager, were we want to have all the caches - public and internal
+      // With capacity factor 0, so vert.x nodes are like light members not containing any data
+      builder.clustering().cacheMode(CacheMode.DIST_SYNC).hash().numOwners(Integer.MAX_VALUE)
```
tristantarrant: Maybe we should introduce the ability to configure CF=0 for members of repl caches which would semantically behave like this but still report as REPL

karesti: This was discussed on the meeting, and we decided to use a DIST cache on locks with max num of owners because for replicated caches apparently is not that simple ... @danberindei @wburns @pruivo any thoughts ?

core/src/main/resources/schema/infinispan-config-10.0.xsd
karesti: It's not exactly true, because internal caches won't have capacity factor 0, and with this modification all caches are capacity factor 0

pruivo: do you mean `If true`?

core/src/test/java/org/infinispan/distribution/ZeroCapacityNodeTest.java
```diff
@@ -0,0 +1,57 @@
...
+import static org.testng.AssertJUnit.assertEquals;
+
+/**
+ * Test the capacity factor for lite instance
+ *
+ * @author Katia Aresti
+ * @since 9.4
+ */
+@Test(groups = "functional", testName = "distribution.ch.ZeroCapacityNodeTest")
+public class ZeroCapacityNodeTest extends MultipleCacheManagersTest {
```
pruivo: why extending `MultipleCacheManagersTest`?

core/src/test/java/org/infinispan/distribution/ZeroCapacityNodeTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+   public static final int NUM_SEGMENTS = 60;
+
+   @Override
+   protected void createCacheManagers() throws Throwable {
+      // Do nothing here, create the cache managers in the test
+   }
+
+   public void testCapacityFactorForLiteInstance() {
```
pruivo: I was expecting more tests to be honest. 
* Does both cache mode behaves correctly with zero and non-zero nodes? 
* Will counter and multimap work correctly? 
* what happens if you start a cache in a non-zero node? Will all the zero nodes start the cache as well?

core/src/test/java/org/infinispan/distribution/ZeroCapacityNodeTest.java
```diff
@@ -0,0 +1,57 @@
...
+      cb.clustering().cacheMode(CacheMode.DIST_SYNC);
+      cb.clustering().hash().numSegments(NUM_SEGMENTS);
+
+      cb.clustering().hash().capacityFactor(0.5f);
+      addClusterEnabledCacheManager(gc, cb);
+      waitForClusterToForm();
+      assertCapacityFactors(0.0f);
+   }
+
+   private void assertCapacityFactors(float... expectedCapacityFactors) {
```
pruivo: why an array?

lock/src/main/java/org/infinispan/lock/impl/ClusteredLockModuleLifecycle.java
pruivo: @tristantarrant the replicated mode uses a simplified version of `ConsistentHashs` and totally ignores the CF and `num-owners`. I guess it can be done but we had to drop the `ReplicatedConsistentHash` and do some tunning probably (or replace it with the `DefaultConsistentHash` with `num-owners == MAX_INTEGER`.

core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
```diff
@@ -157,6 +161,11 @@ public ShutdownConfigurationBuilder shutdown() {
       return modules;
    }
 
+   public GlobalConfigurationBuilder zeroCapacityNode(Boolean zeroCapacityNode) {
```
wburns: Should be boolean, otherwise NPE will be thrown.

core/src/main/java/org/infinispan/configuration/parsing/ConfigurationBuilderHolder.java
```diff
@@ -50,11 +50,7 @@ public ConfigurationBuilder getCurrentConfigurationBuilder() {
    }
 
    public ConfigurationBuilder getDefaultConfigurationBuilder() {
-      if (globalConfigurationBuilder.defaultCacheName().isPresent()) {
-         return namedConfigurationBuilders.get(globalConfigurationBuilder.defaultCacheName().get());
-      } else {
-         return null;
-      }
+       return globalConfigurationBuilder.defaultCacheName().map(namedConfigurationBuilders::get).orElse(null);
```
wburns: Technically this is another lambda and out of scope of this PR imo.

core/src/main/java/org/infinispan/configuration/parsing/Parser.java
```diff
@@ -553,6 +553,10 @@ private void parseContainer(XMLExtendedStreamReader reader, ConfigurationBuilder
                builder.shutdown().hookBehavior(ShutdownHookBehavior.valueOf(value));
                break;
             }
+            case ZERO_CAPACITY_NODE: {
+               builder.zeroCapacityNode(Boolean.valueOf(value));
```
wburns:
```suggestion
               builder.zeroCapacityNode(Boolean.parseBoolean(value));
```

core/src/test/java/org/infinispan/distribution/ZeroCapacityNodeTest.java
karesti: refactored, the other was a test I took inspiration from but not working one

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
```diff
@@ -94,14 +94,16 @@ public void start() throws Exception {
          persistentStateChecksum = Optional.empty();
       }
 
+      float capacityFactor = globalConfiguration.isZeroCapacityNode() ? 0.0f : configuration.clustering().hash().capacityFactor();
```
wburns: I really wish this was set in the `HashConfigurationBuilder#create` method instead, so we don't have to worry about all the references. But we don't have access to the `GlobalConfiguration` there as far as I know. :(

core/src/main/java/org/infinispan/configuration/parsing/ConfigurationBuilderHolder.java
karesti: ah yes, haha I saw it and I thought in the refactoring but can remove it without any problem.
I forgot I did it here

core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
karesti: yes

core/src/main/java/org/infinispan/configuration/parsing/Parser.java
karesti: I was using valueOf as for the statistics, but parseBoolean is the same

karesti:
There is a test failure that does not seem related. @pruivo do you think that we really need additional tests for counters and locks?

karesti:
@pruivo I'm going to add a test on locks

karesti:
@pruivo added some tests on counters and detected a problem with locks. The zero capacity node is counted in the cluster as a non owner, so if we say we need Max Integer values owners, cluster won't start. I think this must be set up depending on the configuration. By default I say 2, and I have this issue ISPN-9625 to make locks xml configurable (such as how should they behave in when there is a split brain)

core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
```diff
@@ -15,6 +15,8 @@
 
 public class GlobalConfigurationBuilder implements GlobalConfigurationChildBuilder {
 
+   public static final String ZERO_CAPACITY_NODE_FEATURE = "zero-capacity-node";
```
wburns: I am not liking this here. Especially if this feature is removed earlier. I would prefer if this variable was defined on a class that isn't marked as public.

core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
```diff
@@ -157,6 +161,11 @@ public ShutdownConfigurationBuilder shutdown() {
       return modules;
    }
 
+   public GlobalConfigurationBuilder zeroCapacityNode(boolean zeroCapacityNode) {
```
wburns: Seems to be missing some documentation. Also I don't remember the discussions, but is this going to be fully supported mode? I wonder if we should add an @Experimental tag to it?

karesti: what do you mean by fully supported?

core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
karesti: I moved the constant to be private and if I need to access from somewhere else I will find another place

core/src/test/java/org/infinispan/distribution/ZeroCapacityNodeTest.java
karesti: @pruivo I need to test that in a cluster I have a node with zero capacity and it works

core/src/main/java/org/infinispan/statetransfer/StateTransferManagerImpl.java
karesti: Can I do something here now then ?

core/src/main/java/org/infinispan/configuration/global/GlobalConfigurationBuilder.java
wburns: I am saying as in it doesn't have @Experimental tag on it and is shown to users and properly documented everything. For example this method has no javadoc on it, doesn't tell me what this means at all.

karesti: @wburns I added some extra javadoc in a commit

karesti: @wburns some java doc added in another commit

wburns:

PR #6442, opened 2018-11-20
ISPN-9748 Default transport for the test suite should be same

diegolovison:
Default transport for the test suite should be the same except if it was activated by another profile

PR #6443, opened 2018-11-20
[9.4] ISPN-9748 Default transport for the test suite should be the same

diegolovison:
Default transport for the test suite should be the same except if it was activated by another profile

### DIFF
--- a/core/src/test/java/org/infinispan/test/fwk/JGroupsConfigBuilder.java
+++ b/core/src/test/java/org/infinispan/test/fwk/JGroupsConfigBuilder.java
@@ -74,7 +74,7 @@ public class JGroupsConfigBuilder {
    };
 
    static {
-      JGROUPS_STACK = LegacyKeySupportSystemProperties.getProperty("infinispan.test.jgroups.protocol", "protocol.stack", "tcp");
+      JGROUPS_STACK = LegacyKeySupportSystemProperties.getProperty("infinispan.test.jgroups.protocol", "protocol.stack", "udp");
       System.out.println("Transport protocol stack used = " + JGROUPS_STACK);
    }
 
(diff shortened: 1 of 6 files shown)


core/src/test/java/org/infinispan/test/fwk/JGroupsConfigBuilder.java
```diff
@@ -74,7 +74,7 @@ protected Integer initialValue() {
    };
 
    static {
-      JGROUPS_STACK = LegacyKeySupportSystemProperties.getProperty("infinispan.test.jgroups.protocol", "protocol.stack", "tcp");
+      JGROUPS_STACK = LegacyKeySupportSystemProperties.getProperty("infinispan.test.jgroups.protocol", "protocol.stack", "udp");
```
danberindei: I'm a bit concerned about changing the default to `udp` because we've had problems with multicasting on Macs before, but I guess it's already the default in the server so it must be working now.

Also, I'm thinking about changing the name to `infinispan.test.jgroups.stack`, using `infinispan.test.jgroups.protocol` as fallback in library tests and `default.transport.stack` as fallback in server tests. `protocol` always seemed a bit weird here, but I do want to keep `infinispan.test` as part of the property name :)

diegolovison: > multicasting on Macs

Multicasting on Macs is still an issue for Embedded **and** Server mode. If you are using UDP without `-Djava.net.preferIPv4Stack=true` the cluster won't form.
The Jgroups issue is https://issues.jboss.org/browse/JGRP-2305 and it was fixed in 4.0.16. ISPN is using 4.0.15.Final and Jgroups 4.0.16 was not released yet. I prefer cut a release for Jgroups, update the Jgroups version for ISPN and then merge that PR. What do you think?

> Properties

Do you mean `getProperty("infinispan.test.jgroups.stack", "default.transport.stack", "udp")` ?

diegolovison:
I am closing this until having a final solution for https://github.com/infinispan/infinispan/pull/6442

diegolovison:
See #6827

PR #6447, opened 2018-11-20
ISPN-9749 Rest PUT fails to replace value when storing pojos without equality

gustavonalle:
https://issues.jboss.org/browse/ISPN-9749

### DIFF
--- a/server/rest/src/main/java/org/infinispan/rest/resources/CacheResource.java
+++ b/server/rest/src/main/java/org/infinispan/rest/resources/CacheResource.java
@@ -159,7 +159,6 @@ public class CacheResource implements ResourceHandler {
             CacheEntry<Object, Object> entry = restCacheManager.getInternalEntry(cacheName, key, true, keyContentType, contentType);
             if (entry instanceof InternalCacheEntry) {
                InternalCacheEntry ice = (InternalCacheEntry) entry;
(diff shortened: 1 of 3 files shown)


server/rest/src/main/java/org/infinispan/rest/resources/CacheResource.java
```diff
@@ -174,7 +173,7 @@ private RestResponse putValueToCache(RestRequest request) {
             boolean useAsync = request.getPerformAsyncHeader();
             Long ttl = request.getTimeToLiveSecondsHeader();
             Long idle = request.getMaxIdleTimeSecondsHeader();
-            return putInCache(responseBuilder, useAsync, cache, key, data, ttl, idle, oldData);
+            return putInCache(responseBuilder, useAsync, cache, key, data, ttl, idle);
```
rvansa: Isn't this breaking the case when the request includes an `ETag`? I would rather keep that working rather than support ill-defined objects.

gustavonalle: What scenario specifically? We have ETag using PUT tests on ```BaseRestOperationsTest```, but I can extend them

karesti:
@rvansa I'm going to merge, and if there is a need for an additional issue let's open one

PR #6449, opened 2018-11-20
ISPN-9754 Cache size method should consult Shared cache stores when not

wburns:
async

* Directly ask loader for size when a shared store

https://issues.jboss.org/browse/ISPN-9754

### DIFF
--- a/core/src/main/java/org/infinispan/context/Flag.java
+++ b/core/src/main/java/org/infinispan/context/Flag.java
@@ -280,6 +280,12 @@ public enum Flag {
     */
    REMOTE_ITERATION,
 
+   /**
+    * Flag that can be used to skip any size optimizations - forcing iteration of entries to count. User shouldn't
+    * normally need to use this flag. This is helpful if there are concerns that can cause just a simple size invocation
+    * from being consistent (eg. on-going transaction with modifications).
+    */
+   SKIP_SIZE_OPTIMIZATION
    ;
 
    /**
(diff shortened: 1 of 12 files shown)


wburns:
To be honest a large chunk of this PR is just noise around making the AccessMode a predicate to make some methods a bit more flexible.

The real meat is contained in:
1. Actual interceptor changes - https://github.com/infinispan/infinispan/pull/6449/files#diff-d70f5361c5448faaa6e2d954a02b42bbR284
2. Size method via Predicate - https://github.com/infinispan/infinispan/pull/6449/files#diff-d93c51f709416559fc199f2b1687e2f7R774
3. Explicit test - https://github.com/infinispan/infinispan/pull/6449/files#diff-75800b15db75d9791a8459795a87b74fR106

wburns:
Actually this is broken when running in a tx. We could only read the underlying entries in the tx and manually add/subtract. This would still prevent size being an N operation with shared store. This needs to be redone.

ryanemerson:
@wburns NullPointerException on HotRodUpgradeWithStoreTest

wburns:
Thanks! Actually I forgot to add a check for the flag skip cache loader as
well that needs to be added.

wburns:
@ryanemerson weird... the NPE is from something that would have happened before even. Somehow the loaders collection and config collection are out of sync despite holding the read lock for the stores mutex.

wburns:
Aha, wow okay we have a pretty big bug with disabling a store...
https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java#L1186

If you notice it always removes the loader from the status. It should only remove it if the loader matched the type.

wburns:
It seems this bug is in 9.3, 9.4 and 10.0. Crept in with the store availability changes.

wburns:
I created https://issues.jboss.org/browse/ISPN-9773 to fix the NPE.

wburns:
Rebased on the other PR to fix the disable store NPE.

PR #6455, opened 2018-11-21
[9.4.x] ISPN-9742 Analyzers not working for inner objects

gustavonalle:
https://issues.jboss.org/browse/ISPN-9742


karesti:
@gustavonalle this LGTM, do you need something else ? it is already in master, right ?

PR #6459, opened 2018-11-21
ISPN-9756 Improve ETag generation in the REST server

gustavonalle:
https://issues.jboss.org/browse/ISPN-9756

This is just a preview to see the effects on the performance. I'm afraid the VersionGenerators available are not sufficient to be used as Etags, since they are  based on transient attributes such as memory only atomic integers, view Ids, etc. Once the server is restarted, it is possible that a change in an entry will not result in different version for the entry.

Ideally we should use something that uses time as part of the calculation and needs no coordination cluster wide, such as https://blog.twitter.com/engineering/en_us/a/2010/announcing-snowflake.html

### DIFF
--- a/server/rest/src/main/java/org/infinispan/rest/resources/CacheResource.java
+++ b/server/rest/src/main/java/org/infinispan/rest/resources/CacheResource.java
@@ -22,6 +22,7 @@ import org.infinispan.commons.dataconversion.MediaType;
 import org.infinispan.commons.hash.MurmurHash3;
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.container.entries.InternalCacheEntry;
+import org.infinispan.container.versioning.LongEntryVersion;
 import org.infinispan.encoding.DataConversion;
 import org.infinispan.metadata.Metadata;
 import org.infinispan.rest.CacheControl;
@@ -121,7 +122,7 @@ public class CacheResource implements ResourceHandler {
 
          if (entry instanceof InternalCacheEntry) {
             InternalCacheEntry<Object, Object> ice = (InternalCacheEntry<Object, Object>) entry;
-            String etag = calcETAG(ice.getValue());
+            String etag = getETag(ice.getMetadata());
             String clientEtag = request.getEtagIfNoneMatchHeader();
             if (clientEtag == null || clientEtag.equals(etag)) {
                responseBuilder.status(HttpResponseStatus.OK.code());
@@ -162,7 +163,7 @@ public class CacheResource implements ResourceHandler {
                oldData = entry.getValue();
                String etagNoneMatch = request.getEtagIfNoneMatchHeader();
                if (etagNoneMatch != null) {
-                  String etag = calcETAG(ice.getValue());
+                  String etag = getETag(ice.getMetadata());
                   if (etagNoneMatch.equals(etag)) {
                      //client's and our ETAG match. Nothing to do, an entry is cached on the client side...
                      responseBuilder.status(HttpResponseStatus.NOT_MODIFIED.code());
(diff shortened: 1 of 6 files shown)


gustavonalle:
run performance tests please
cs-rest

gustavonalle:
run performance tests please
cs-rest

server/rest/src/main/java/org/infinispan/rest/resources/CacheResource.java
```diff
@@ -287,14 +287,19 @@ private void writeValue(Object value, MediaType requested, MediaType configuredM
       if (returnBody) responseBuilder.entity(value);
    }
 
-   private <V> String calcETAG(V value) {
-      return String.valueOf(hashFunc.hash(value));
+   private String calcETAG(Metadata metadata) {
+      LongEntryVersion entryVersion = (LongEntryVersion) metadata.version();
+      if (entryVersion != null) {
```
rvansa: Under what circumstances the entry does not have a version?

gustavonalle: When the entry is not inserted via Hot Rod or Rest

rvansa: Hmm, I see that it is not enforced, it seems that server is adding the version info. I am thinking about adding VersionInterceptor in server all the time, or enforcing that somewhere else... otherwise the conditional ops won't work correctly.

gustavonalle: ```adding VersionInterceptor in server all the time,```

Would that cover the case when someone executes a script or a task  in the server getting hold of a cache directly and writing an entry?

danberindei: `VersionInterceptor` doesn't really add versions to new entries @rvansa. And there's still the possibility of embedded nodes running in the same cluster with server nodes... I'd rather force embedded to use versions all the time as well, if the perf hit isn't too big.

diegolovison:
run performance tests please
cs-rest

diegolovison:
> I don't think the perfack job uses any ETags.
ETags will be present in the response. Do have any scenario in mind?

I saw different performance tests scenarios in that case

1. Measure the drawback when ETag is in the Response
`curl -H "Content-Type: text/plain" -i http://localhost:8080/rest/default/1`
> HTTP/1.1 200 OK
> ETag: "a4s654564a5ds4a56ds4ew65"

2. Measure `curl -H "Content-Type: text/plain" -H 'If-None-Match: "a4s654564a5ds4a56ds4ew65"'
 -i http://localhost:8080/rest/default/1`
> HTTP/1.1 304 Not Modified
> ETag: "f88dd058fe004909615a64f01be66a7"

3. Measure 
3.1. `curl -H "Content-Type: text/plain" -i -X PUT --data 'bar' http://localhost:8080/rest/default/1`
> HTTP/1.1 200 OK
> ETag: "qw8e73sd156asd6f48qw9e5ads45"

3.2. `curl -H "Content-Type: text/plain" -H 'If-None-Match: "a4s654564a5ds4a56ds4ew65"'
 -i http://localhost:8080/rest/default/1`
> HTTP/1.1 200 OK
> ETag: "03cb37ca667706c68c0aad4cb04c3a211"

4. Will we support `If-Match`?

If I understood correctly, at this time we are looking for the item number 1.

gustavonalle:
@diegolovison The REST server currently keeps calculating the murmur hash to put in the Etag response every read. 
This PR uses a numeric version when inserting data and stores it in the metadata, and during reads it simply returns it. So the current perf job should be enough to measure the diff even if it does not use etag headers in the request

server/rest/src/main/java/org/infinispan/rest/resources/CacheResource.java
gustavonalle: @danberindei Both generators should be deleted :)

gustavonalle:
Updated, not using the version generators anymore, but simply storing the current ETag in the metadata.

gustavonalle:
run performance tests please
cs-rest

server/rest/src/main/java/org/infinispan/rest/resources/CacheResource.java
rvansa: @danberindei Scattered cache relies on topology ids in the entry version (though it does not use the version generator).

gustavonalle:
So the conclusion is: REST reads would be about 5% faster when storing the etag in the metadata instead of calculating it on the fly, but since the entry can also be written from other endpoints and embedded (scripts, tasks, etc), a consensus upon entry versions should exist otherwise it's pointless: https://issues.jboss.org/browse/ISPN-9770

Closing for now.

PR #6467, opened 2018-11-26
ISPN-9768 Do not register MBeans if they are disabled

tristantarrant:
- avoids initializing classes

https://issues.jboss.org/browse/ISPN-9768


tristantarrant:
Failures unrelated

PR #6469, opened 2018-11-26
ISPN-5997 JMX operation ClusterCacheStats.resetStatistics() not working

vblagoje:
*Preview only* For master branch but could potentially be backported - I am not that familiar with the process for the core these days.

### DIFF
--- a/core/src/main/java/org/infinispan/marshall/core/ExternallyMarshallable.java
+++ b/core/src/main/java/org/infinispan/marshall/core/ExternallyMarshallable.java
@@ -42,6 +42,7 @@ public final class ExternallyMarshallable {
       whiteListClasses.add("org.infinispan.scripting.impl.DataType"); // prod
       whiteListClasses.add("org.infinispan.scripting.impl.DistributedScript");
       whiteListClasses.add("org.infinispan.stats.impl.ClusterCacheStatsImpl$DistributedCacheStatsCallable"); // prod
+      whiteListClasses.add("org.infinispan.stats.impl.ClusterCacheStatsImpl$ResetStatsDistributedCacheStatsCallable"); // prod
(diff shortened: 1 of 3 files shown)


karesti:
@vblagoje still preview?

tristantarrant:
This should really use the ClusteredExecutor as the DES is on its way out. /cc @wburns can provide a hint on how to do it

vblagoje:
> This should really use the ClusteredExecutor as the DES is on its way out. /cc @wburns can provide a hint on how to do it

@tristantarrant but we have other DES'es around being used as well. Perhaps we can integrate it and then upgrade DESes to ClusteredExecutor in a separate issue?

vblagoje:
@tristantarrant @ryanemerson @wburns I looked into ClusterExecutor and although it is designed well and much better than the old executor the new one does not allow (if I am not mistaken) access to caches on remote nodes like old one [does](https://github.com/infinispan/infinispan/pull/6469/commits/c32d187710823c10d6baeb2c89366c94aae19642#diff-cb130c1ca09d594e1dd93484448ddf54R461). @wburns is there a way to access remote cache on the nodes where Runnable is executed?

vblagoje:
@wburns how would I get access to remote caches from Runnable in ClusteredExecutor?

tristantarrant:
https://github.com/infinispan/infinispan/blob/356aed336d074424687d79fbbfa40515e0714869/server/integration/event-logger/src/main/java/org/infinispan/server/eventlogger/ServerEventLogger.java#L101

wburns:
I will try to tackle https://issues.jboss.org/browse/ISPN-9784 which should help with this :)

tristantarrant:
Closing this as it needs a different implementation

PR #6474, opened 2018-11-27
ISPN-9775 - ISPN build for IBM Java ends with error in Ant execution task

andyuk1986:


server/hotrod/pom.xml
```diff
@@ -153,6 +153,10 @@
                               <arg value="${project.build.testOutputDirectory}/password_client_truststore.p12"/>
                               <arg value="-deststorepass"/>
                               <arg value="secret"/>
+                              <arg value="-srcstoretype"/>
+                              <arg value="pkcs12"/>
+                              <arg value="-deststoretype"/>
+                              <arg value="pkcs12"/>
```
wburns: Do we not want "JCEKS" as the other one in client hotrod?

andyuk1986: @wburns I don't know. According to the last commits in server/hotrod all keystores were changed to pkcs12. I have just added the required part for IBM JDK to be able to build them.

wburns: Okay, it seems to be fine with our CI, but I haven't confirmed with IBM, so I trust you guys on that front :)

wburns:
Integrated into master, thanks @andyuk1986

PR #6476, opened 2018-11-28
ISPN-9778 XidImpl implementations can optimize some byte[] allocations

wburns:
* Remove unneeded additional branch id
* Reuse byte[] in constructor for XidImpl

https://issues.jboss.org/browse/ISPN-9778


wburns:
I can't remove the branch id in all cases, as we have to make sure to keep it from an external source.

PR #6479, opened 2018-11-29
ISPN-9777 Default AdvancedCacheLoader#publishEntries can hang with more

wburns:
than 2048 entries

* Use single thread pool temporarily

https://issues.jboss.org/browse/ISPN-9777


ryanemerson:
@wburns We should probably backport this to 9.4.x

wburns:
@ryanemerson I was debating about that, since it has an easy enough workaround for the user to change their loader. But I can create one real quick.

PR #6487, opened 2018-11-30
ISPN-9727 Ickle query ignores an IN clause if there is only one value

gustavonalle:
* The boolean OR query around the single value has this side effect; this can be fixed
  by transforming an IN predicate with a single value into a simple equals right after parsing

https://issues.jboss.org/browse/ISPN-9727

### DIFF
--- a/object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/parser/ExpressionBuilder.java
+++ b/object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/parser/ExpressionBuilder.java
@@ -90,7 +90,12 @@ final class ExpressionBuilder<TypeMetadata> {
          ComparisonExpr booleanExpr = new ComparisonExpr(valueExpr, new ConstantValueExpr(typedValue), ComparisonExpr.Type.EQUAL);
          children.add(booleanExpr);
       }
-      push(new OrExpr(children));
+      if (children.size() == 1) {
+         // simplify INs with just one clause by removing the wrapper boolean expression
+         push(children.get(0));
+      } else {
+         push(new OrExpr(children));
+      }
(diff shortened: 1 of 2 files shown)


wburns:
@gustavonalle The change looks good to me, however I have a question - why didn't https://github.com/infinispan/infinispan/blob/master/query/src/test/java/org/infinispan/query/dsl/embedded/QueryDslConditionsTest.java#L1070 fail before? My guess is that that it only failed if the IN clause had 1 variable and there was another qualifier?

object-filter/src/main/java/org/infinispan/objectfilter/impl/syntax/parser/ExpressionBuilder.java
```diff
@@ -90,7 +90,12 @@ public void addIn(PropertyPath<?> propertyPath, List<Object> values) {
          ComparisonExpr booleanExpr = new ComparisonExpr(valueExpr, new ConstantValueExpr(typedValue), ComparisonExpr.Type.EQUAL);
          children.add(booleanExpr);
       }
-      push(new OrExpr(children));
+      if (children.size() == 1) {
+         // simplify INs with just one clause by removing the wrapper boolean expression
+         push(children.get(0));
+      } else {
+         push(new OrExpr(children));
```
danberindei: @gustavonalle This looks like an optimization, not a fix... is there a problem with `OrExpr` when it has a single child, and shouldn't we fix that as well?

gustavonalle: @danberindei I can certify this change fixes the reproducer, I'll let @anistor comment on other possible cases

PR #6490, opened 2018-11-30
ISPN-9732 Local iteration optimization with write behind

diegolovison:
ISPN-9732 Local iteration optimization with write behind is valid for non shared stores

### DIFF
--- a/core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
+++ b/core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
@@ -133,14 +133,14 @@ public class DistributedCacheStream<Original, R> extends AbstractCacheStream<Ori
     * Stream from an IntStream, DoubleStream etc.
     * @param other other instance of {@link AbstractCacheStream} to copy details from
     */
-   protected DistributedCacheStream(AbstractCacheStream other) {
+   public DistributedCacheStream(AbstractCacheStream other) {
       super(other);
 
       Configuration configuration = registry.getComponent(Configuration.class);
       writeBehindShared = hasWriteBehindSharedStore(configuration.persistence());
    }
 
-   boolean hasWriteBehindSharedStore(PersistenceConfiguration persistenceConfiguration) {
+   public boolean hasWriteBehindSharedStore(PersistenceConfiguration persistenceConfiguration) {
       for (StoreConfiguration storeConfiguration : persistenceConfiguration.stores()) {
          if (storeConfiguration.shared() && storeConfiguration.async().enabled()) {
             return true;
--- a/core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
+++ b/core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
@@ -1,18 +1,26 @@
 package org.infinispan.stream;
 
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.spy;
+
 import java.io.Serializable;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.infinispan.Cache;
 import org.infinispan.CacheStream;
 import org.infinispan.configuration.cache.CacheMode;
 import org.infinispan.configuration.cache.ConfigurationBuilder;
 import org.infinispan.configuration.cache.HashConfigurationBuilder;
 import org.infinispan.container.entries.CacheEntry;
+import org.infinispan.container.entries.ImmortalCacheEntry;
 import org.infinispan.distribution.ch.ConsistentHash;
+import org.infinispan.distribution.ch.KeyPartitioner;
 import org.infinispan.distribution.ch.impl.DefaultConsistentHash;
 import org.infinispan.distribution.ch.impl.ScatteredConsistentHash;
 import org.infinispan.filter.Converter;
@@ -20,7 +28,9 @@ import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.marshall.core.ExternalPojo;
 import org.infinispan.metadata.Metadata;
 import org.infinispan.remoting.transport.Address;
+import org.infinispan.stream.impl.ClusterStreamManager;
 import org.infinispan.test.MultipleCacheManagersTest;
+import org.infinispan.test.TestingUtil;
 import org.infinispan.test.fwk.TestCacheManagerFactory;
 import org.infinispan.transaction.TransactionMode;
 import org.infinispan.util.BaseControlledConsistentHashFactory;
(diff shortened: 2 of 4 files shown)


core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
```diff
@@ -143,4 +153,24 @@ public String convert(Object key, String value, Metadata metadata) {
          }
       }
    }
+
+   protected ClusterStreamManager replaceWithSpy(Cache<?,?> cache) {
+      ClusterStreamManager component = TestingUtil.extractComponent(cache, ClusterStreamManager.class);
+      ClusterStreamManager clusterStreamManager = spy(component);
+      TestingUtil.replaceComponent(cache, ClusterStreamManager.class, clusterStreamManager, false);
+      reset(clusterStreamManager);
```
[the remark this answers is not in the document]
diegolovison: According to my tests, `TestingUtil.replaceComponent` is "interacting" the variable. Without the `reset` the `verifyZeroInteractions(clusterStreamManager)` will fail.

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
```diff
@@ -0,0 +1,70 @@
...
+      this(false, CacheMode.DIST_SYNC);
+   }
+
+   public DistributedWriteBehindStreamIteratorTest(boolean tx, CacheMode mode) {
+      super(tx, mode);
+   }
+
+   @Override
+   protected void enhanceConfiguration(ConfigurationBuilder builder) {
+      builder.persistence().addStore(DummyInMemoryStoreConfigurationBuilder.class).shared(true).async().enable();
```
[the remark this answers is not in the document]
diegolovison: I will add the name but should we add a validation into the builder to fail when `shared=true` and `name is empty`?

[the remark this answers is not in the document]
wburns: That is probably a good idea for the DummyInMemoryStore, yeah. Maybe just log a JIRA for it.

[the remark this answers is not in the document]
diegolovison: https://issues.jboss.org/browse/ISPN-9803

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
```diff
@@ -0,0 +1,70 @@
...
+   public void testStayLocalIfAllSegmentsPresentLocallyWithWriteBehindStoreWithoutRehash() throws Exception {
+      testStayLocalIfAllSegmentsPresentLocallyWithWriteBehindStore(false);
+   }
+
+   protected void testStayLocalIfAllSegmentsPresentLocallyWithWriteBehindStore(boolean rehashAware) {
+      Cache<Object, String> cache0 = cache(0, CACHE_NAME);
+
+      ClusterStreamManager clusterStreamManager = replaceWithSpy(cache0);
+
+      IntStream.rangeClosed(0, 499).boxed().forEach(i -> cache0.put(i, i.toString()));
```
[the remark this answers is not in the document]
diegolovison: Before answering your question, I would like to know how I know what should be the value? I was using https://github.com/infinispan/infinispan/blob/master/core/src/test/java/org/infinispan/stream/DistributedStreamIteratorTest.java#L398 as an example

core/src/test/java/org/infinispan/stream/BaseSetupStreamIteratorTest.java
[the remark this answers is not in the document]
wburns: Hrmm, okay.

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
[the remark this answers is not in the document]
wburns: Just something of enough size, when I added that I was a bit overzealous is all :)

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
```diff
@@ -133,14 +133,14 @@ public DistributedCacheStream(Address localAddress, boolean parallel, Distributi
     * Stream from an IntStream, DoubleStream etc.
     * @param other other instance of {@link AbstractCacheStream} to copy details from
     */
-   protected DistributedCacheStream(AbstractCacheStream other) {
+   public DistributedCacheStream(AbstractCacheStream other) {
```
wburns: Hrmm it seems these changes snuck in?

diegolovison: I would like to reuse the `DistributedCacheStream` implementation. See https://github.com/infinispan/infinispan/pull/6490/files#diff-b3f3c6e34c53f47dfb4deaa44a2d3c50R69

wburns: I don't think that assertion is really needed as the setup part of the test already guarantees that. (ie. this isn't related to this test) If there was a problem with that it would be a bug with the configuration class.

diegolovison: If the configuration is
```java
   @Override
   protected void enhanceConfiguration(ConfigurationBuilder builder) {
      builder.persistence().addStore(DummyInMemoryStoreConfigurationBuilder.class).async().enable();
   }
```
the test will pass.
I would like make sure that we are using `storeConfiguration.shared() && storeConfiguration.async().enabled()`

diegolovison: And now I remember the main reason behind the change in the interface.
I cannot mock `DistributedCacheStream` because of 
```java
   public DistributedCacheStream(AbstractCacheStream other) {
      super(other);

      Configuration configuration = registry.getComponent(Configuration.class);
      writeBehindShared = hasWriteBehindSharedStore(configuration.persistence());
   }
```

wburns: > If the configuration is
> 
> ```java
>    @Override
>    protected void enhanceConfiguration(ConfigurationBuilder builder) {
>       builder.persistence().addStore(DummyInMemoryStoreConfigurationBuilder.class).async().enable();
>    }
> ```
> 
> the test will pass.
> I would like make sure that we are using `storeConfiguration.shared() && storeConfiguration.async().enabled()`

This is a problem then. All entries shouldn't be returned in that case, there must be a different setup issue. I can look a bit closer.

wburns: > And now I remember the main reason behind the change in the interface.
> I cannot mock `DistributedCacheStream` because of
> 
> ```java
>    public DistributedCacheStream(AbstractCacheStream other) {
>       super(other);
> 
>       Configuration configuration = registry.getComponent(Configuration.class);
>       writeBehindShared = hasWriteBehindSharedStore(configuration.persistence());
>    }
> ```

I don't understand why you need to mock this class.

diegolovison: > I don't understand why you need to mock this class.

The test is to test a write-behind shared store, the only way that I found that we are really testing a write-behind share store is mocking `DistributedCacheStream` and doing an assert against the method `hasWriteBehindSharedStore`

> This is a problem then. All entries shouldn't be returned in that case, there must be a different setup issue. I can look a bit closer.

https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java#L659 will return `true`

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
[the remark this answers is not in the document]
danberindei: I say if you can get away with using 1 entry, use 1 entry @diegolovison :)

core/src/main/java/org/infinispan/stream/impl/DistributedCacheStream.java
danberindei: @diegolovison If you don't trust that the test is set up with a write-behind store you can use `cache.getConfiguration().persistence().usingAsyncStore()` in the test.

wburns:
Succeeded by https://github.com/infinispan/infinispan/pull/6522

PR #6477, opened 2018-11-29
[9.3.x] ISPN-9701 TransactionTable does not shutdown gracefully

danberindei:
https://issues.jboss.org/browse/ISPN-9701

PR #6494, opened 2018-12-03
ISPN-9701 Fix InitiatorCrashPessimisticTest

danberindei:
Forward port of the test fix in #6477


core/src/test/java/org/infinispan/lock/singlelock/AbstractInitiatorCrashTest.java
```diff
@@ -30,6 +30,7 @@ public void testInitiatorCrashesBeforeReleasingLock() throws Exception {
       releaseLocksLatch.await(30, TimeUnit.SECONDS);
 
       assert checkTxCount(0, 0, 1);
```
karesti: could we start considering refactoring these assert statements whenever we touch a test that have them?

danberindei: Sorry, I missed your comment @karesti. We already have a policy of replacing the `assert` keyword with `AssertJUnit` when we touch the test, but honestly I didn't even think about the asserts at the time: the test was failing and I wanted to get the fix in ASAP.

PR #6503, opened 2018-12-04
ISPN-9763 Persistence element added to server configuration

ryanemerson:
https://issues.jboss.org/browse/ISPN-9763

PR #6507, opened 2018-12-05
ISPN-9798 Fix potential deadlock.

xiaodong-xie:
https://issues.jboss.org/browse/ISPN-9798

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
+++ b/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
@@ -27,6 +27,7 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -172,7 +173,9 @@ public void start() {
 
             // Now schedule the availability check
             long interval = configuration.persistence().availabilityInterval();
-            availabilityFuture = persistenceExecutor.scheduleAtFixedRate(this::pollStoreAvailability, interval, interval, TimeUnit.MILLISECONDS);
+            if (interval > 0) {
+               availabilityFuture = persistenceExecutor.scheduleAtFixedRate(this::pollStoreAvailability, interval, interval, TimeUnit.MILLISECONDS);
```
xiaodong-xie: If Wildfly could expose `availabilityFuture` as a parameter, we could use `-1` to disable it if needed.

galderz:
@ryanemerson Can you have a look?

ryanemerson:
@xiaodong-xie Thank you for your contribution! I have created another PR https://github.com/infinispan/infinispan/pull/6510 that approaches the problem in a slightly different way. My solution removes the use of the `writeLock` in favour of the `readLock`. Please try it out if you can.

The reason I created my PR, was that with your solution I am concerned that it's possible for the availability check to never succeed. This is fine if the PersistenceManager is available and the persistence operations are executing as expected. However, if the PersistenceManager has been marked as unavailable, it's possible that on a cache with many ongoing operations the PersistenceManager will never become available again.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
ryanemerson: For the infinispan server I have created a PR to expose the availability parameters https://github.com/infinispan/infinispan/pull/6503. However, the infinispan subsystem provided as part of wildfly 14 is out of our control.

What we recommend is to utilise our `wildfly-modules` zip to utilise the latest version of Infinispan  on Wildfly. This will allow you to use our subsystem extensions which have more configuration options. Please read our [user guide](http://infinispan.org/docs/stable/user_guide/user_guide.html#infinispan_modules_for_wildfly_eap) for more details.

xiaodong-xie: Cool, thanks a lot for the info. 👍 😄

xiaodong-xie:
Thanks for the info and fix, @ryanemerson. 👍 

#6510 is much better, closing this one. 😄

PR #6510, opened 2018-12-06
ISPN-9798 Availability deadlock fixes

ryanemerson:
https://issues.jboss.org/browse/ISPN-9798
https://issues.jboss.org/browse/ISPN-9800

As `checkStoreAvailability()` always utilises the volatile `unavailableException` to determine whether the PersistenceManager is available, we don't need to use the writeLock as the `storeStatuses` map is never updated, only the `StoreStatus` itself. ~~However, we do need to introduce a lock to prevent scheduled calls to `pollStoreAvailability` overlapping.~~

### DIFF
--- a/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
+++ b/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
@@ -27,7 +27,9 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
@@ -132,6 +134,7 @@ public class PersistenceManagerImpl implements PersistenceManager {
    private final List<TransactionalCacheWriter> txWriters = new ArrayList<>();
    private final Semaphore publisherSemaphore = new Semaphore(Integer.MAX_VALUE);
    private final ReadWriteLock storesMutex = new ReentrantReadWriteLock();
+   private final Lock availabilityMutex = new ReentrantLock();
    @GuardedBy("storesMutex")
    private final Map<Object, StoreStatus> storeStatuses = new HashMap<>();
    private AdvancedPurgeListener<Object, Object> advancedListener;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -185,25 +189,32 @@ public void start() {
...
-         boolean availabilityChanged = false;
-         boolean failureDetected = false;
-         for (StoreStatus status : storeStatuses.values()) {
-            if (status.availabilityChanged())
-               availabilityChanged = true;
-            if (availabilityChanged && !status.availability && !failureDetected) {
-               failureDetected = true;
-               unavailableException = new StoreUnavailableException(String.format("Store %s is unavailable", status.store));
-               cacheNotifier.notifyPersistenceAvailabilityChanged(false);
+         // If the lock is not available, then it means that an availability check is in progress due to scheduled calls overlapping
```
wburns: The method can't be invoked multiple times from the scheduler per the javadoc, it just gets delayed. The only way this could be concurrently called is if someone disabled a store at the same time.

wburns:
I am fine with the changes, however I am concerned why this even happened in this case. This appears to be that the ReentrantReadWriteLock is acting more like a fair lock, in that a second read lock attempt is blocked when even though another thread has acquired just the read lock. It seems that when a waiting writer is present it is blocking it.

wburns:
I wrote up this quick test case and it seems even worse than I thought for non fair locks

https://gist.github.com/wburns/3d6a036cd08a787bee93f844ae9dac49

wburns:
This means any ReentrantReadWriteLock.readLock has to be treated as a mutually exclusive object for dead lock purposes in all code sections, which is a bit scary 👎

xiaodong-xie:
@wburns, IMHO it's a bit by design, in order to avoid writer starvation, yes? 

```java
    /**
     * Nonfair version of Sync
     */
    static final class NonfairSync extends Sync {
        private static final long serialVersionUID = -8159625535654395037L;
        final boolean writerShouldBlock() {
            return false; // writers can always barge
        }
        final boolean readerShouldBlock() {
            /* As a heuristic to avoid indefinite writer starvation,
             * block if the thread that momentarily appears to be head
             * of queue, if one exists, is a waiting writer.  This is
             * only a probabilistic effect since a new reader will not
             * block if there is a waiting writer behind other enabled
             * readers that have not yet drained from the queue.
             */
            return apparentlyFirstQueuedIsExclusive();
        }
    }

```

So, if there is a writer at the head of queue, all remaining reader would be blocked.

wburns:
Oh I agree that is how the code is, however it doesn't state that in the javadoc. Sadly it only says unspecified, and the code assumed it worked in the opposite way of fair.

I am more pointing out the fact that this means we have to essentially treat any and all ReentrantReadWriteLoc readLock instances as essentially write locks in terms of deadlock possibilities.

xiaodong-xie:
Yeah, I agree. 👍

wburns:
And because of that, there is still a very small chance of deadlock when a user removed a store and https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java#L331 is invoked.

wburns:
I am not sure I like that as much. I have been thinking about a similar issue recently (https://issues.jboss.org/browse/ISPN-9722) and tbh the main issue is this line https://github.com/infinispan/infinispan/blob/master/core/src/main/java/org/infinispan/persistence/internal/PersistenceUtil.java#L119 We can't be acquiring such wide reaching locks while inside the DataContainer lock. I have changes for that issue that instead of using the DataContainer lock it acquires the InfinispanLock for the key instead, preventing this issue.

Though I think the `pollStoreAvailability` should probably dropped to a `readLock` like Ryan mentioned since the volatile variable `unavailableException` handles all subsequent checks and it will improve concurrency quite a bit, since we won't be acquiring a write lock at all really.

Also to note I assume this cache is not clustered? This deadlock shouldn't be possible in a clustered cache.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -1235,14 +1238,14 @@ private StoreConfiguration getStoreConfig(Object store) {
...
-      volatile boolean availability = true;
+      boolean availability = true;
 
       StoreStatus(Object store, StoreConfiguration config) {
          this.store = store;
          this.config = config;
       }
 
-      boolean availabilityChanged() {
+      synchronized boolean availabilityChanged() {
```
wburns: The locking here is a bit questionable, but the subsequent read is on the same thread so it is fine.

xiaodong-xie:
No, we do not use clustered cache. The deadlock I mentioned in https://issues.jboss.org/browse/ISPN-9798 was from a local cache.

wburns:
Integrated into master, thanks @xiaodong-xie and @ryanemerson 

We need this on 9.4.x and 9.3.x as well I assume @ryanemerson ?

xiaodong-xie:
Thanks, @ryanemerson and @wburns 👍 😄

danberindei:
@wburns the `ReadWriteLock` javadoc is much clearer:

```quote
Writer preference is common, as writes are expected to be short and infrequent. Reader preference is less common as it can lead to lengthy delays for a write if the readers are frequent and long-lived as expected. Fair, or "in-order" implementations are also possible.
```

I take this to mean that unfair mode is going to be really unfair to reads. Not sure why they went for "the order of entry to the read and write lock is unspecified" in the `ReentrantReadWriteLock` javadoc instead of saying it explicitly, people will rely on the writer preference choice anyway.

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
danberindei: I also don't like the fact that the name of the method is worded like it's a simple getter, but it can block for a long time (like it happened with the `RocksDB` store: https://issues.jboss.org/browse/ISPN-9762)

I would have preferred the lists of loaders/writers to hold a wrapper that contained the loader/writer reference + status + configuration, but it would only have a `compareAndSet` method, with the actually availability check being triggered directly in the `pollStoreAvailability` method. I think I've already mentioned that I like to avoid map lookups in other discussions :)

danberindei:
@ryanemerson I'm guessing this also fixes https://issues.jboss.org/browse/ISPN-9762?

PR #6522, opened 2018-12-07
[9.4] ISPN-9732 Local iteration optimization with write behind is valid for non shared stores

wburns:
Added more tests for the use case.

https://issues.jboss.org/browse/ISPN-9732


core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
```diff
@@ -0,0 +1,134 @@
...
+      testStayLocalIfAllSegmentsPresentLocallyWithWriteBehindStore(false);
+   }
+
+   protected void testStayLocalIfAllSegmentsPresentLocallyWithWriteBehindStore(boolean rehashAware) {
+      Cache<Object, String> cache0 = cache(0, CACHE_NAME);
+
+      ClusterStreamManager clusterStreamManager = replaceWithSpy(cache0);
+
+      int amountToInsert = 200;
+      IntStream.range(0, amountToInsert).boxed().forEach(i -> cache0.put(i, i.toString()));
```
danberindei: Why so many keys and segments @wburns?

wburns: Well the 3 segments in the original didn't work, because the originator had to be a backup owner of a given node for it to work. I can reduce them a bit more if you wanted, abut this didn't seem like too many to make sure the test always passed.

danberindei:
@wburns What's the relationship to https://github.com/infinispan/infinispan/pull/6490?

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
danberindei: Surely all the segments in the parent have a backup owner as well?

wburns:
I closed the other the same time you asked :) I missed doing that originally.

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
wburns: I can try messing with it, just seemed easier to just increase the segments at the time :)

wburns:
Updated to use less segments and entries.

wburns:
@danberindei this PR should be ready now.

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
```diff
@@ -0,0 +1,129 @@
...
+      if (cacheMode.isReplicated()) {
+         Map<Object, String> entries = mapFromIterator(stream.iterator());
+         // We should not see keys from other segments, but there may be segments without any keys
+         assertEquals(caches(CACHE_NAME).size(), entries.size());
+         invocationCount = cacheManagers.size() - 1;
+      } else {
+         // Distributed cache
+         // Cache1 owns 1 (primary) and 2 (backup)
+         // When it is a write behind shared store it will have to go remote otherwise will stay local
+         Map<Object, String> entries = mapFromIterator(stream.filterKeySegments(IntSets.immutableSet(2)).iterator());
```
danberindei: Question: is the optimization performed only when all the requested segments are local, or is it also performed for backup-owned segments when the application requests all the segments? If it's the former, then why? If it's the latter, then I think it would be best to stream all segments in dist caches as well.

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
```diff
@@ -0,0 +1,129 @@
...
+   public void testStayLocalIfAllSegmentsPresentLocallyWithWriteBehindStoreWithReHash() throws Exception {
+      testStayLocalIfAllSegmentsPresentLocallyWithWriteBehindStore(true);
+   }
+
+   @Test
+   public void testStayLocalIfAllSegmentsPresentLocallyWithWriteBehindStoreWithoutRehash() throws Exception {
+      testStayLocalIfAllSegmentsPresentLocallyWithWriteBehindStore(false);
+   }
+
+   protected void testStayLocalIfAllSegmentsPresentLocallyWithWriteBehindStore(boolean rehashAware) {
```
danberindei: I don't like the test name, because the test asserts that a stream request **is** sent to other nodes in some configurations. I'd call it something like `testBackupSegmentsOptimizationWithWriteBehindStore` and get `rehashAware` from a `DataProvider`.

ryanemerson:
@wburns Are you going to issue a PR for master as well?

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+      // Remember that segment ownership is as {0, 1}, {1, 2}, {2, 1}
+
+      CacheStream<Map.Entry<Object, String>> stream = cache1.entrySet().stream();
+      if (!rehashAware) stream = stream.disableRehashAware();
+
+      int invocationCount;
+      if (cacheMode.isReplicated()) {
+         Map<Object, String> entries = mapFromIterator(stream.iterator());
+         // We should not see keys from other segments, but there may be segments without any keys
```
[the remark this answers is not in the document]
wburns: Removed

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+@Test(groups = {"functional", "smoke"}, testName = "iteration.DistributedWriteBehindStreamIteratorTest")
+public class DistributedWriteBehindStreamIteratorTest extends BaseSetupStreamIteratorTest {
+
+   private boolean asyncStore;
+   private boolean sharedStore;
+
+   public DistributedWriteBehindStreamIteratorTest() {
+      // Cache Mode is provided in factory methods
+      super(false, null);
```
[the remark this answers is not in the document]
wburns: Sure

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
wburns: If the local node owns all requested segments it will stay local for them - unless shared write behind mode. Are you asking about a subset? In this case it only does primary on the local node. The new publisher however does backups even when it doesn't own all of them.

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
wburns: Okay

wburns:
Updated again @danberindei :)

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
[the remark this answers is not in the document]
danberindei: Not sure what you mean. You implement `factory()` yourself, so you can construct the test instance any way you want:

https://github.com/infinispan/infinispan/blob/7453ebbb8e1a7452acaa419504bbd3a18e8abe3c/core/src/test/java/org/infinispan/commands/GetAllCommandTest.java#L37

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
danberindei: Yes, I was asking about dist caches, where not all requested segments are local. Glad to hear that the new publisher optimizes backups owned by the originator in that case as well.

core/src/test/java/org/infinispan/stream/DistributedWriteBehindStreamIteratorTest.java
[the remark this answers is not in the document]
wburns: The problem is the factory method requires a default constructor in the first place. It first invokes the default constructor then the `factory` method to find the instances. So I figured it was less code than requiring another constructor is all.

wburns:
My comment doesn't show properly on the change above so pasting here for more visibility:

"The problem is the factory method requires a default constructor in the first place. It first invokes the default constructor then the factory method to find the instances. So I figured it was less code than requiring another constructor is all."

wburns:
Master version is https://github.com/infinispan/infinispan/pull/6741

PR #6523, opened 2018-12-07
Make MarshalledEntry and MarshalledEntryFactory part of the persistence SPI

ryanemerson:
https://issues.jboss.org/browse/ISPN-9693


ryanemerson:
@danberindei @wburns This is just from my marshalling PR to reduce the diff size https://github.com/infinispan/infinispan/pull/6240

core/src/main/java/org/infinispan/marshall/core/MarshalledEntry.java
```diff
@@ -12,36 +9,8 @@
  *
  * @author Mircea Markus
  * @since 6.0
+ * @deprecated use {@link org.infinispan.persistence.spi.MarshalledEntry}
```
wburns:
```suggestion
 * @deprecated since 10.0, use {@link org.infinispan.persistence.spi.MarshalledEntry}
```

core/src/main/java/org/infinispan/marshall/core/MarshalledEntryFactory.java
```diff
@@ -1,19 +1,11 @@
...
 
-import org.infinispan.commons.io.ByteBuffer;
-import org.infinispan.metadata.InternalMetadata;
-
 /**
  * Factory for {@link MarshalledEntry}.
  *
- * @author Mircea Markus
  * @since 6.0
+ * @deprecated use {@link org.infinispan.persistence.spi.MarshalledEntryFactory} instead
```
wburns:
```suggestion
 * @deprecated since 10.0, use {@link org.infinispan.persistence.spi.MarshalledEntryFactory} instead
```

ryanemerson:
@wburns @danberindei I was thinking about this over the weekend ... Maybe now is a good time to also fix the fact that we expose `InternalMetadata` via `MarshalledEntry`. Our store implementations rely on a lot of the methods that are only exposed in `InternalMetadata`, and I'm sure they would be useful for other store impls. So I think this is mainly a naming issue.

Proposal:

Create `org.infinispan.persistence.spi.EntryMetadata` for the SPI:
```java
interface `EntryMetadata extends InternalMetadata {}`
```

Update `org.infinispan.persistence.spi.MarshalledEntry` so that `InternalMetadata getMetadata()` is removed and `EntryMetadata getEntryMetadata` is added.

For backwards compatibility, we must then add `InternalMetadata getMetadata()` to the deprecated `org.infinispan.marshall.core.MarshalledEntry`

wburns:
Sounds good to me, however I don't think the new interface should extend `InternalMetadata`. It should be the inverse and we hide `InternalMetadata` if we need to have some methods that are for internal use only. It might be that we don't need to hide anything, but I would rather have it available if we need to.

```java
interface EntryMetadata {
 ... needed methods ...
}
```

```java
interface InternalMetadata extends EntryMetadata {
 ... methods just for us ...
}
```

This would mirror how we handle `DataContainer` and `InternalDataContainer`.

wburns:
However, I am fine handling that in another PR so we can integrate this one right now :)

ryanemerson:
@wburns Yeah that sounds a better approach and I agree lets put it in another PR :smile:

danberindei:
@ryanemerson I think the reason for the name is that the timestamps are stored in the `InternalCacheEntry` implementations, not in the `Metadata` implementations. We allow custom `Metadata` implementations, and we don't require them to store the timestamps, so the timestamps are part of the persistence SPI but not part of the custom metadata SPI. 

Personally I'd like to keep all the timestamps and the lifespan/maxIdle parameters out of `Metadata` **and** out of `InternalMetadata`, so they wouldn't be included in `getMetadataBytes()` and instead each store could use its own custom storage. I'm thinking especially of JDBC stores, but for other stores as well it may be nice to have the expiration time at a fixed offset in the entry. Not sure how to do it in a backwards compatible way, though :)

danberindei:
@ryanemerson @wburns I just realized that `Metadata` doesn't have `isExpired()`, so it should be safe to keep everything as it is now only adding adding `ByteBuffer getUserMetadataBytes()` to `MarshalledEntry` for stores that can/need to store the timestamps separately and moving the `InternalMetadata` methods to `MarshalledEntry` (and keeping the old methods as deprecated). That way `Metadata` would only hold data that needs to be updated atomically on a `put()`, and we could get rid of the `Builder` :)

Somewhat related, we should really move most of the methods in `CacheEntry` to a new interface `ContextCacheEntry`, they really don't make sense when returned from `getCacheEntry()`.

ryanemerson:
@danberindei Just to make sure I follow before doing anything. You're proposing that the `InternalMetadata` methods should be directly part of the new `persistence.spi.MarshalledEntry` interface like so:

```java
   ByteBuffer getKeyBytes();

   ByteBuffer getValueBytes();

   K getKey();

   V getValue();

   // Start of new methods
   Metadata getUserMetadata();

   ByteBuffer getUserMetadataBytes();

   long created();

   long lastUsed();

   boolean isExpired(long now);

   long expiryTime();
```

PR #6525, opened 2018-12-10
ISPN-4075 State transfer should preserve the timestamps of entries 

danberindei:
https://issues.jboss.org/browse/ISPN-4075

The fix is a bit crude because I didn't want to add the timestamps in regular PutKeyValueCommands, but it does the job.

### DIFF
--- a/core/src/main/java/org/infinispan/container/impl/InternalDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/impl/InternalDataContainer.java
@@ -47,12 +47,19 @@ public interface InternalDataContainer<K, V> extends DataContainer<K, V> {
    /**
     * Same as {@link DataContainer#put(Object, Object, Metadata)} except that the segment of the key can provided to
     * write/lookup entries without calculating the segment for the given key.
+    *
+    * <p>Note: The timestamps ignored if the entry already exists in the data container.</p>
+    *
     * @param segment segment for the key
     * @param k key under which to store entry
     * @param v value to store
     * @param metadata metadata of the entry
+    * @param createdTimestamp creation timestamp, or {@code -1} to use the current time
+    * @param lastUseTimestamp last use timestamp, or {@code -1} to use the current time
+    *
+    * @since 10.0
     */
-   void put(int segment, K k, V v, Metadata metadata);
+   void put(int segment, K k, V v, Metadata metadata, long createdTimestamp, long lastUseTimestamp);
 
    /**
     * Same as {@link DataContainer#containsKey(Object)}  except that the segment of the key can provided to
--- a/core/src/main/java/org/infinispan/container/offheap/SegmentedBoundedOffHeapDataContainer.java
+++ b/core/src/main/java/org/infinispan/container/offheap/SegmentedBoundedOffHeapDataContainer.java
@@ -97,8 +97,8 @@ public class SegmentedBoundedOffHeapDataContainer extends AbstractDelegatingInte
    }
 
    @Override
-   public void put(int segment, WrappedBytes wrappedBytes, WrappedBytes wrappedBytes2, Metadata metadata) {
-      super.put(segment, wrappedBytes, wrappedBytes2, metadata);
+   public void put(int segment, WrappedBytes key, WrappedBytes value, Metadata metadata, long createdTimestamp, long lastUseTimestamp) {
+      super.put(segment, key, value, metadata, createdTimestamp, lastUseTimestamp);
       // The following is called outside of the write lock specifically - since we may not have to evict and even
       // if we did it would quite possibly need a different lock
       ensureSize();
--- a/core/src/main/java/org/infinispan/interceptors/impl/CallInterceptor.java
+++ b/core/src/main/java/org/infinispan/interceptors/impl/CallInterceptor.java
@@ -168,7 +168,16 @@ public class CallInterceptor extends BaseAsyncInterceptor implements Visitor {
          throw new IllegalStateException("Not wrapped");
       }
 
-      Object newValue = command.getValue();
+      Object newValue;
+      if (!command.hasAllFlags(FlagBitSets.PUT_FOR_STATE_TRANSFER | FlagBitSets.CACHE_MODE_LOCAL)) {
+         newValue = command.getValue();
+      } else {
+         // [Scattered]StateConsumerImpl guarantees the value is actually an entry
+         InternalCacheEntry ice = (InternalCacheEntry) command.getValue();
+         e.setCreated(ice.getCreated());
+         e.setLastUsed(ice.getLastUsed());
+         newValue = ice.getValue();
+      }
       Object prevValue = e.getValue();
       if (!valueMatcher.matches(prevValue, null, newValue)) {
          command.fail();
--- a/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
+++ b/core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
@@ -641,7 +641,8 @@ public class StateConsumerImpl implements StateConsumer {
                ctx = icf.createSingleKeyNonTxInvocationContext();
             }
 
-            PutKeyValueCommand put = commandsFactory.buildPutKeyValueCommand(e.getKey(), e.getValue(), segmentId,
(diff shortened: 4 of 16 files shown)


core/src/main/java/org/infinispan/commands/write/PutForStateTransferCommand.java
```diff
@@ -0,0 +1,52 @@
+package org.infinispan.commands.write;
+
+import org.infinispan.commands.CommandInvocationId;
+import org.infinispan.metadata.Metadata;
+
+public class PutForStateTransferCommand extends PutKeyValueCommand {
```
wburns: I am thinking we should have this class should implement `LocalCommand` so we can easily see which commands aren't to be replicated.

danberindei:
I realized I had forgotten about scattered caches, and then I noticed they already use a different approach: a `PutMapCommand(PUT_FOR_STATE_TRANSFER)` with `InternalCacheValue` instead of regular values. So I changed my code to use `InternalCacheEntry` values instead of adding a new command type (ICE to avoid allocating an ICV instance in `ICE.toInternalCacheValue()`).

core/src/main/java/org/infinispan/container/impl/InternalDataContainer.java
```diff
@@ -47,12 +47,19 @@
...
     * @param k key under which to store entry
     * @param v value to store
     * @param metadata metadata of the entry
+    * @param createTimestamp creation timestamp, or 0 to use the current time
+    * @param lastUseTimestamp last use timestamp, or 0 to use the current time
+    *
+    * @since 10.0
     */
-   void put(int segment, K k, V v, Metadata metadata);
+   void put(int segment, K k, V v, Metadata metadata, long createTimestamp, long lastUseTimestamp);
```
wburns: I am glad we added this internal interface :)

core/src/main/java/org/infinispan/container/offheap/SegmentedBoundedOffHeapDataContainer.java
```diff
@@ -97,8 +97,8 @@ public void put(WrappedBytes key, WrappedBytes value, Metadata metadata) {
    }
 
    @Override
-   public void put(int segment, WrappedBytes wrappedBytes, WrappedBytes wrappedBytes2, Metadata metadata) {
-      super.put(segment, wrappedBytes, wrappedBytes2, metadata);
+   public void put(int segment, WrappedBytes wrappedBytes, WrappedBytes wrappedBytes2, Metadata metadata, long createTimestamp, long lastUseTimestamp) {
```
wburns:
```suggestion
   public void put(int segment, WrappedBytes key, WrappedBytes value, Metadata metadata, long createTimestamp, long lastUseTimestamp) {
```
Oops. Didn't realize these were messed up before.

core/src/main/java/org/infinispan/container/offheap/SegmentedBoundedOffHeapDataContainer.java
```diff
@@ -97,8 +97,8 @@ public void put(WrappedBytes key, WrappedBytes value, Metadata metadata) {
...
+      super.put(segment, wrappedBytes, wrappedBytes2, metadata, createTimestamp, lastUseTimestamp);
```
wburns:
```suggestion
      super.put(segment, key, value, metadata, createTimestamp, lastUseTimestamp);
```

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
```diff
@@ -641,7 +641,8 @@ private void doApplyState(Address sender, int segmentId, Collection<InternalCach
                ctx = icf.createSingleKeyNonTxInvocationContext();
             }
 
-            PutKeyValueCommand put = commandsFactory.buildPutKeyValueCommand(e.getKey(), e.getValue(), segmentId,
+            // CallInterceptor knows PUT_FOR_STATE_TRANSFER values are actually InternalCacheEntries
+            PutKeyValueCommand put = commandsFactory.buildPutKeyValueCommand(e.getKey(), e, segmentId,
```
wburns: The only thing I wonder is, we could be duplicating the metadata we send across right? Maybe we should pass null for metadata?

core/src/test/java/org/infinispan/statetransfer/StateTransferTimestampsTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.infinispan.test.MultipleCacheManagersTest;
+import org.infinispan.test.TestingUtil;
+import org.infinispan.util.ControlledTimeService;
+import org.testng.annotations.Test;
+
+/**
+ * Tests concurrent startup of replicated and distributed caches
+ *
+ * @author Dan Berindei
+ * @since 7.2
```
wburns: Wrong version :)

core/src/test/java/org/infinispan/statetransfer/StateTransferTimestampsTest.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+@Test(testName = "statetransfer.StateTransferTimestampsTest", groups = "functional")
+public class StateTransferTimestampsTest extends MultipleCacheManagersTest {
+   public static final String CACHE_NAME = "cache";
+   private ControlledTimeService timeService;
+
+   @Override
+   public Object[] factory() {
+      return new Object[]{
+         new StateTransferTimestampsTest().cacheMode(CacheMode.REPL_SYNC),
```
wburns: No DIST_SYNC ?

core/src/test/java/org/infinispan/statetransfer/StateTransferTimestampsTest.java
```diff
@@ -0,0 +1,80 @@
...
+      };
+   }
+
+   @Override
+   protected void createCacheManagers() throws Throwable {
+      createCluster(new ConfigurationBuilder(), 2);
+
+      timeService = new ControlledTimeService();
+      ConfigurationBuilder replConfig = new ConfigurationBuilder();
+      replConfig.clustering().cacheMode(CacheMode.REPL_SYNC).hash().numSegments(4);
```
wburns: Seems we aren't testing Scattered then?

core/src/main/java/org/infinispan/interceptors/impl/CallInterceptor.java
```diff
@@ -73,6 +73,7 @@
 import org.infinispan.container.entries.CacheEntry;
 import org.infinispan.container.entries.ExpiryHelper;
 import org.infinispan.container.entries.InternalCacheEntry;
+import org.infinispan.container.entries.InternalCacheValue;
```
wburns: Pesky checkstyle.

core/src/main/java/org/infinispan/statetransfer/StateConsumerImpl.java
danberindei: These put commands are never sent over the network.

core/src/test/java/org/infinispan/statetransfer/StateTransferTimestampsTest.java
danberindei: Oops!

core/src/test/java/org/infinispan/statetransfer/StateTransferTimestampsTest.java
danberindei: The state transfer process is the same for all clustered caches except scattered.

core/src/test/java/org/infinispan/statetransfer/StateTransferTimestampsTest.java
danberindei: I thought I was just lucky that it passed the first time I ran the test :)

danberindei:
Ok, fixing scattered mode took longer than I thought, but it works now.

wburns:
Changes look good to me, however CI seems to have puked https://ci.infinispan.org/job/Infinispan/job/PR-6525/9/consoleFull#-897946742ea5a4324-14c7-4883-b210-e33383209aa7
I haven't looked recently if this is related or not.

wburns:
👍

wburns:

PR #6527, opened 2018-12-10
ISPN-9801 ClusterTopologyManagerImpl hangs when restarting a FORK node

danberindei:
https://issues.jboss.org/browse/ISPN-9801

* ClusterTopologyManagerImpl should process new views while waiting
  for member hearbeats for a previous view
* Member availability check should ignore nodes that didn't join
  any cache
* Test doesn't reproduce the problem described in the issue,
  but it's still good to have
* Small changes in other tests

### DIFF
--- a/core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
+++ b/core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
@@ -379,14 +379,20 @@ public class ClusterCacheStatus implements AvailabilityStrategyContext {
       }
    }
 
-   public synchronized void doHandleClusterView() {
+   public synchronized void doHandleClusterView(int viewId) {
       // TODO Clean up ClusterCacheStatus instances once they no longer have any members
       if (currentTopology == null)
          return;
 
       List<Address> newClusterMembers = transport.getMembers();
+      int newViewId = transport.getViewId();
+      if (newViewId != viewId) {
+         log.debugf("Cache %s skipping members update for view %d, newer view received: %d",
+                    cacheName, viewId, newViewId);
+         return;
+      }
+      if (trace) log.tracef("Cache %s updating members for view %d: %s", cacheName, viewId, newClusterMembers);
       boolean cacheMembersModified = retainMembers(newClusterMembers);
-      if (trace) log.tracef("Cache %s doHandleClusterView newClusterMembers=%s", cacheName, newClusterMembers);
       availabilityStrategy.onClusterViewChange(this, newClusterMembers);
 
       if (cacheMembersModified) {
--- a/core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
+++ b/core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
@@ -17,6 +17,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
@@ -26,6 +27,7 @@ import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import net.jcip.annotations.GuardedBy;
 import org.infinispan.commands.ReplicableCommand;
 import org.infinispan.commons.CacheException;
 import org.infinispan.commons.util.CollectionFactory;
@@ -65,6 +67,7 @@ import org.infinispan.remoting.rpc.ResponseFilter;
 import org.infinispan.remoting.rpc.ResponseMode;
 import org.infinispan.remoting.transport.Address;
 import org.infinispan.remoting.transport.Transport;
+import org.infinispan.remoting.transport.impl.VoidResponseCollector;
 import org.infinispan.remoting.transport.jgroups.SuspectException;
 import org.infinispan.statetransfer.RebalanceType;
 import org.infinispan.util.concurrent.CompletableFutures;
@@ -74,8 +77,6 @@ import org.infinispan.util.logging.LogFactory;
 import org.infinispan.util.logging.events.EventLogCategory;
 import org.infinispan.util.logging.events.EventLogManager;
 import org.infinispan.util.logging.events.EventLogger;
-
-import net.jcip.annotations.GuardedBy;
 
 /**
  * The {@code ClusterTopologyManager} implementation.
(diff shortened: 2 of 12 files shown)


core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
```diff
@@ -379,14 +379,20 @@ private void updateMembers() {
...
 
-   public synchronized void doHandleClusterView() {
+   public synchronized void doHandleClusterView(int viewId) {
       // TODO Clean up ClusterCacheStatus instances once they no longer have any members
       if (currentTopology == null)
          return;
 
       List<Address> newClusterMembers = transport.getMembers();
+      int newViewId = transport.getViewId();
+      if (newViewId != viewId) {
```
wburns: Is it possible to get a viewId from coordinator before our transport has updated its own?

core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
```diff
@@ -500,29 +501,53 @@ private void recoverClusterStatus(int newViewId, final List<Address> clusterMemb
...
          if (clusterManagerStatus.isRunning()) {
-            log.errorUpdatingMembersList(e);
+            log.errorUpdatingMembersList(viewId, e);
          }
       }
    }
 
-   private void confirmMembersAvailable() throws Exception {
-      transport.invokeRemotely(null, HeartBeatCommand.INSTANCE, ResponseMode.SYNCHRONOUS, getGlobalTimeout(), null, DeliverOrder.NONE, false);
+   public void doUpdateCacheMembers(int viewId, Throwable throwable) throws Throwable {
```
wburns: Method seems a bit superfluous - Do you think if you consolidated it would look a bit nicer? I personally am not a big fan of using exceptions for flow control when we will catch it in the same class and ignore it.

core/src/test/java/org/infinispan/statetransfer/ConcurrentStartChanelLookupTest.java
```diff
@@ -103,7 +103,7 @@ public void testConcurrentStart(int eagerManager, int lazyManager) throws Except
 
    private EmbeddedCacheManager createCacheManager(String name1, JChannel ch1) {
       GlobalConfigurationBuilder gcb1 = new GlobalConfigurationBuilder();
-      gcb1.transport().nodeName(ch1.getName()).distributedSyncTimeout(10, SECONDS);
+      gcb1.transport().defaultTransport().nodeName(ch1.getName()).distributedSyncTimeout(10, SECONDS);
```
wburns: Is this a different bug?

core/src/test/java/org/infinispan/test/transport/DelayedViewJGroupsTransport.java
```diff
@@ -19,11 +24,21 @@ public void receiveClusterView(View newView) {
...
+            }
          } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
          }
       }
       super.receiveClusterView(newView);
    }
+
+   public void assertUnblocked() {
+      assert unblocked;
```
wburns: Maybe return boolean and do assertTrue in the calling code?

core/src/test/java/org/infinispan/test/transport/DelayedViewJGroupsTransport.java
```diff
@@ -19,11 +24,21 @@ public void receiveClusterView(View newView) {
       // check if this is an event of node going down, and if so wait for a signal to apply new view
       if (waitLatch != null && getMembers().size() > newView.getMembers().size()) {
          try {
-            waitLatch.await(10, TimeUnit.SECONDS);
+            log.debugf("Delaying view %s", newView);
+            unblocked = waitLatch.await(10, TimeUnit.SECONDS);
+            if (unblocked) {
+               log.debugf("Unblocking view %s", newView);
+            } else {
+               log.errorf("Timed out waiting for view to be unblocked: %s", newView);
```
wburns: Should this throw an exception instead?

core/src/main/java/org/infinispan/topology/ClusterCacheStatus.java
danberindei: Yes, the topology commands are OOB so they could be delivered before the view message. But this is not about receiving the view later, it's about receiving a newer view while confirming all the members are available. I'll add a comment.

core/src/main/java/org/infinispan/topology/ClusterTopologyManagerImpl.java
danberindei: I was trying to keep the code that ignores exceptions in the same method, but you're right, it's not very pretty.

core/src/test/java/org/infinispan/statetransfer/ConcurrentStartChanelLookupTest.java
danberindei: No, I just changed `CustomChannelLookup.registerChannel` to not set the transport in the global config because it's not in the name... I guess I could have renamed it to `configureTransportWithChannel()` instead :)

core/src/test/java/org/infinispan/test/transport/DelayedViewJGroupsTransport.java
danberindei: It's not a very strong preference, but I like specific methods because then the error is right there in the stack trace and you don't have to open the source code to see what failed.

core/src/test/java/org/infinispan/test/transport/DelayedViewJGroupsTransport.java
danberindei: JGroups may or may not log the exception, but the test is not going to see it anyway.

danberindei:
I guess this is one of the good things about code reviews, we get to learn new parts of the code base :smile: 

The failure log shows 2 HeartBeatCommands per view, when there should be only 1:

```
10:34:09,792 DEBUG (Transport-1,Test:[ViewHandling]) [ClusterTopologyManagerImpl] Recovering cluster status for view 310:34:09,793 DEBUG (Transport-1,Test:[ViewHandling]) [ClusterTopologyManagerImpl] Got 1 status responses. members are [B#1]
10:34:09,794 DEBUG (Transport-1,Test:[ViewHandling]) [MockTransport] Intercepted command org.infinispan.topology.HeartBeatCommand@65351f9c
10:34:09,794 DEBUG (Transport-1,Test:[ViewHandling]) [MockTransport] Intercepted command org.infinispan.topology.HeartBeatCommand@65351f9c
10:34:09,796 DEBUG (Transport-2,Test:[ViewHandling]) [ClusterTopologyManagerImpl] Skipping cache members update for view 3 after receiving view 4
10:34:09,796 DEBUG (Transport-2,Test:[ViewHandling]) [ClusterTopologyManagerImpl] Skipping cache members update for view 3 after receiving view 4
```

I wasn't able to reproduce the failure locally, but I have enhanced the test to check that there is only one HBC for view 3.

danberindei:
@wburns I finally managed to reproduce and fix the failure in ClusterTopologyManagerImplTest, if that's the one you were referring to.

wburns:

PR #6528, opened 2018-12-10
[9.4.x] ISPN-9809 Added version.component.console

pruivo:
* To avoid patch conflicts with downstream every time console gets updated!

https://issues.jboss.org/browse/ISPN-9809


ryanemerson:
@pruivo Are you going to create this for master as well?

pruivo:
@ryanemerson yes :)

PR #6534, opened 2018-12-12
ISPN-9819 Listeners correction in spring session

karesti:
https://issues.jboss.org/browse/ISPN-9819


server/hotrod/src/main/java/org/infinispan/server/hotrod/KeyValueRetrieverConverter.java
```diff
@@ -0,0 +1,86 @@
...
+   private int putBytes(byte[] bytes, int offset, byte[] out) {
+      int localOffset = offset;
+      for (byte b : bytes) {
+         out[localOffset] = b;
+         localOffset += 1;
+      }
+      return localOffset - offset;
+   }
+
+   private int putLong(long l, int offset, byte[] out) {
```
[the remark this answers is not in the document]
wburns: I am aware that the existing `KeyValueVersionConverter` did this, we can update that one as well.

server/hotrod/src/main/java/org/infinispan/server/hotrod/KeyValueRetrieverConverter.java
```diff
@@ -0,0 +1,86 @@
...
+
+      byte[] out = new byte[capacity];
+      int offset = UnsignedNumeric.writeUnsignedInt(out, 0, key.length);
+      offset += putBytes(key, offset, out);
+      if (newValue != null) {
+         offset += UnsignedNumeric.writeUnsignedInt(out, offset, newValue.length);
+         offset += putBytes(newValue, offset, out);
+         putLong(((NumericVersion) newMetadata.version()).getVersion(), offset, out);
+      }
+      if (oldValue != null) {
```
wburns: Hrmm, I am debating if we want to only just return a single value instead of both. I know this only affects modification events, which you aren't currently using.

server/hotrod/src/main/java/org/infinispan/server/hotrod/KeyValueRetrieverConverter.java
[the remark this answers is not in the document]
karesti: I didn't know how to handle this duplication, ok to move the code to that class and refactor both classes

server/hotrod/src/main/java/org/infinispan/server/hotrod/KeyValueRetrieverConverter.java
```diff
@@ -0,0 +1,86 @@
...
+         offset += UnsignedNumeric.writeUnsignedInt(out, offset, oldValue.length);
+         offset += putBytes(oldValue, offset, out);
+         putLong(((NumericVersion) newMetadata.version()).getVersion(), offset, out);
+      }
+
+      return out;
+   }
+
+   private int putBytes(byte[] bytes, int offset, byte[] out) {
```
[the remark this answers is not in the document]
karesti: Great!

karesti:
@wburns Ok for the tests you mentioned

server/hotrod/src/main/java/org/infinispan/server/hotrod/KeyValueRetrieverConverter.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Set;
+
+import org.infinispan.commons.io.UnsignedNumeric;
+import org.infinispan.commons.marshall.AbstractExternalizer;
+import org.infinispan.container.versioning.NumericVersion;
+import org.infinispan.metadata.Metadata;
+import org.infinispan.notifications.cachelistener.filter.CacheEventConverter;
+import org.infinispan.notifications.cachelistener.filter.EventType;
+
+class KeyValueRetrieverConverter implements CacheEventConverter<byte[], byte[], byte[]> {
```
gustavonalle: I think we could reuse the existing converter rather than creating a new one that is very similar. 
The idea is to add a converter parameter to indicate if previous values should be returned, the default is ```false``` (the current behaviour). The Spring session would set it to ```true```.

Here's an example of a parameterised converter: https://github.com/infinispan/infinispan/blob/313b19301055c6267c6f2ea9065a7ab1b68099fe/server/integration/testsuite/src/test/java/org/infinispan/server/test/client/hotrod/CustomPojoEventConverterFactory.java

In the client side, you could set the parameter with:
```java
remoteCache.addClientListener(eventListener, null, new Object[]{Boolean.TRUE});
```

wburns: 👍

karesti: nice, doing this

karesti:
I will amend all the commits in the end

karesti:
@wburns I've been working on the tests, and it's too much effort today to spend time trying to from spring that the listeners are clustered. Same applies to checking a value inside the httpsession. I added a test to check that the events are raised and that contain a non null http session inside. Which is the original problem that was fixed from spring session and now from our side as well.

karesti:
@wburns concerning integration tests, they are missing. But I would add those tests in the spring-boot starter, where we do provide nice support for spring-boot, instead of infinispan repository. I have already an issue for that. Spring session integration tests are missing there, and others too

karesti:
@gustavonalle @wburns failed tests unrelated

server/hotrod/src/main/java/org/infinispan/server/hotrod/KeyValueRetrieverConverter.java
karesti: this means we leave it as it is ?

server/hotrod/src/main/java/org/infinispan/server/hotrod/KeyValueRetrieverConverter.java
karesti: as we discussed, this parameter stuff is a bit broken, so I did something very simple to grab the old value as well

server/hotrod/src/main/java/org/infinispan/server/hotrod/KeyValueRetrieverConverter.java
karesti: so this class has been removed, we use the same class for both.

danberindei:
@karesti I just noticed @wburns's comment, and he's right that we should check a session attribute in the test to see that it's a full session and not a dummy one.

Please also rename `callEviction` to `processExpiration`, and configure the cache managers with a `ControlledTimeService` so that the test doesn't have to sleep for 1 second.

karesti:
@danberindei I added a test to check that we have a session that it's not null. If the underlying session object exists - not null - it is not a dummy session but the session. This is how spring session is working now

karesti:
@danberindei as I said before, these kind of tests are good to test with spring boot and all the spring boot mechanisms. The starter is the good place to put these tests, instead of the spring integration tests in infinispan code base (older tests, before the starter existed)

danberindei:
@karesti I didn't mean to add a new test, but `testEventBridge` already checks that the number of emitted events is correct, so I was thinking it could be amended to set an attribute in the session then check that it's there in the event as well.

I prefer putting tests as close to the code they're testing as possible. If we only had tests in the starter we could make a breaking change in core, then we'd modify the spring integration to compile and push to master, and we'd only see a failure in the starter test suite much later.

server/hotrod/src/main/java/org/infinispan/server/hotrod/ClientListenerRegistry.java
```diff
@@ -232,10 +232,17 @@ private void addCacheListener(AdvancedCache<byte[], byte[]> cache, Object client
...
 
    private CacheEventConverterFactory findConverterFactory(String name, ConcurrentMap<String, CacheEventConverterFactory> factories) {
-      if (name.equals("___eager-key-value-version-converter"))
-         return KeyValueVersionConverterFactory.SINGLETON;
-      else
-         return findFactory(name, factories, "converter");
+      CacheEventConverterFactory cacheEventConverterFactory;
+      switch (name) {
+         case "___eager-key-value-version-converter" :
+            cacheEventConverterFactory = KeyValueVersionConverterFactory.SINGLETON;
```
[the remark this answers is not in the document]
karesti: Is this code in this PR ?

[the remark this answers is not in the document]
danberindei: No, it's old code... I have no idea why we use 2 ways of hard-coding a converter factory, but I'd like to get rid of one of them. And you already modified this method, so it's best to do the unification now.

spring/spring5/spring5-common/src/test/java/org/infinispan/spring/common/session/util/EventsWaiter.java
```diff
@@ -29,4 +35,30 @@ public static void assertNumberOfEvents(Supplier<Stream<ApplicationEvent>> event
...
+                                           String attName,
+                                           String attValue,
+                                           int timeout,
+                                           TimeUnit timeoutUnit) {
+
+      long stopTime = System.currentTimeMillis() + timeoutUnit.toMillis(timeout);
+
+      List<Session> sessions = null;
+
+      while (System.currentTimeMillis() < stopTime) {
```
danberindei: You need to modify the loop condition or add a break, otherwise the test will always block for 2 seconds here

server/hotrod/src/main/java/org/infinispan/server/hotrod/KeyValueVersionConverter.java
```diff
@@ -32,16 +40,17 @@ private KeyValueVersionConverter() {
          offset += putBytes(newValue, offset, out);
          putLong(((NumericVersion) newMetadata.version()).getVersion(), offset, out);
       }
+      if (returnOldValue && oldValue != null) {
+         offset += UnsignedNumeric.writeUnsignedInt(out, offset, oldValue.length);
```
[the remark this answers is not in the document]
karesti: ok, I did this

server/hotrod/src/main/java/org/infinispan/server/hotrod/KeyValueRetrieverConverter.java
danberindei: Ok, as discussed elsewhere we shouldn't modify the format of the existing converter so we'll include either the new value or the old value, never both.

spring/spring5/spring5-common/src/test/java/org/infinispan/spring/common/session/util/EventsWaiter.java
karesti: I did something else

spring/spring5/spring5-remote/src/main/java/org/infinispan/spring/remote/session/RemoteApplicationPublishedBridge.java
```diff
@@ -43,25 +48,47 @@ public void unregisterListener() {
...
+   }
+
+   private byte[] readElement(ByteBuffer buffer) {
+      byte[] element = null;
+      try {
+         int length = UnsignedNumeric.readUnsignedInt(buffer);
+         element = new byte[length];
+         buffer.get(element);
+      } catch (Exception ex) {
+
```
danberindei: I missed this... we should at least log the error if something went wrong

danberindei: Actually logging is probably best left to `ClientListenerInvocation`... here we should return `null` if `buffer.remaining() == 0`, but otherwise we should assume the buffer contains a full element.

karesti:
Test failure is unrelated.
Backport PR is here https://github.com/infinispan/infinispan/pull/6535

spring/spring5/spring5-remote/src/main/java/org/infinispan/spring/remote/session/RemoteApplicationPublishedBridge.java
danberindei: I pushed a commit to check the size to your branch @karesti 
Test suite looks good locally but I'll wait for CI to confirm

PR #6542, opened 2018-12-13
ISPN-9846 Ensure Backwards Compatibility with Persistence SPI changes

ryanemerson:
~~https://issues.jboss.org/browse/ISPN-9693~~
https://issues.jboss.org/browse/ISPN-9846

@wburns Please excuse the commit history for now. I've kept that in place for now so I can undo things if needed. I'll squash before merging obviously. So far I have only refactored the JdbcStringBasedStore to properly utilise the new Publisher based `writeBatch` method as I wanted your feedback on that Impl before wasting my time elsewhere.

The important changes are:
- `persistence.spi.MarshalledEntry` is now `persistence.spi.MarshallableEntry`, the same naming change applies to the respective factory as well. This is necessary as the `InitializationContext` previously exposed `marshall.core.MarshalledEntry` and the naming change provides us an easy way to differentiate the context methods as we can't have two `getMarshalledEntryFactory` methods due to type erasure.

- Due to type erasure it was necessary to introduce several new methods to the writer/loader methods. Deprecated/Replacement method list below:
        - CacheLoader `MarshalledEntry load(Object)` -> `MarshallableEntry get(Object)`
        - AdvancedCacheLoader `Publisher<MarshalledEntry<K, V>> publishEntries(..)` -> `Publisher<MarshallableEntry<K, V>> entryPublisher(...)`
        - CacheWriter `writeBatch(Iterable<MarshalledEntry>) -> `writeBatch(Publisher<MarshallableEntry>)`
        - InitializationContext `getMarshalledEntryFactory` -> `getMarshallableEntryFactory`
- As well as ^ changes to the interfaces I have also provided the required overidden methods where possible e.g `CacheWriter::write()`


core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
```diff
@@ -37,13 +50,27 @@
...
+   /**
+    * Persist all provided entries to the store in a single batch update. If this is not supported by the
+    * underlying store, then entries are written to the store individually via {@link #write(MarshallableEntry)}.
+    *
+    * @param publisher a {@link Publisher} of {@link MarshallableEntry} instances
+    * @throws NullPointerException if the publisher is null.
+    */
+   default void writeBatch(Publisher<MarshallableEntry<? extends K, ? extends V>> publisher) {
+      // TODO transform MarshallableEntry to MarshalledEntry and call writeBatch(Iterable)
+      Flowable.fromPublisher(publisher).blockingForEach(this::write);
```
ryanemerson: @wburns I wanted this to process the publisher, wrap the MarshallableEntries as a MarshalledEntry if required and then return a `blockingIterable` to pass to `writeBatch`, but I couldn't get generics to play nice. I tried many variations of below without success, any ideas?

```java
writeBatch(
            Flowable.fromPublisher(publisher)
                  .map(e -> MarshalledEntry.wrap(e))
                  .blockingIterable()
      );
```

ryanemerson:
The plan is to stop exposing `InternalMetadata` via the persistence SPI and add `created()` methods direct to `MarshallableEntry` as discussed [here](https://github.com/infinispan/infinispan/pull/6523). I will do this in another PR as it will change a lot of things, but by having the `MarshallableEntry` interface it makes the difference between the old and new API more explicit. So I will change MarshallableEntry to be like so:

```java
interface MarshallableEntry<K, V> {
   ByteBuffer getKeyBytes();

   ByteBuffer getValueBytes();

   K getKey();

   V getValue();

   Metadata metadata();

   ByteBuffer metadataBytes();

   long created();

   long lastUsed();

   boolean isExpired(long now);

   long expiryTime();
}
```

With `MarshalledEntry` then providing the old method for backwards compatiblity:
```
ByteBuffer getMetadataBytes();
InternalMetadata getMetadata();
```

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
wburns: So I am not liking the `void` return type. I am thinking maybe we want this to return a `CompletionStage<Void>`. This will play much nicer in the long run once we have a non blocking store SPI as well.

Looking re: default method.

wburns: The problem is the ? extends, you normally pass in a concrete type. I am always disappointed with nested wildcard generics.

What you have there is fine, you will just have to trick the generics by adding in an erased cast, sadly.

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stringbased/JdbcStringBasedStore.java
```diff
@@ -282,39 +283,41 @@ private void executeLegacyUpdate(Connection connection, MarshalledEntry entry, S
...
-               if (batchSize == configuration.maxBatchSize()) {
-                  batchSize = 0;
-                  upsertBatch.executeBatch();
-                  upsertBatch.clearBatch();
-               }
-            }
-
-            if (batchSize != 0)
+            })
+            .buffer(configuration.maxBatchSize())
```
wburns: So this is good. I have a possible idea of how to make it better, but not sure if it is doable with the rxjava2 API.

However I would suggest removing the first `doOnNext` and just move it into the second one and do a for loop over the list. This gives us one less lambda and should allow the JVM to optimize a bit easier.

wburns: We could technically do a more fancy thing using scan and takeLast to avoid adding to a list. But it doesn't look that clean to me, so I would say lets just stick with the buffer we have here and change the `doOnNext`

persistence/jpa/src/main/java/org/infinispan/persistence/jpa/JpaStore.java
```diff
@@ -423,14 +423,34 @@ public void write(MarshallableEntry entry) {
...
-   public void writeBatch(Iterable<MarshalledEntry<? extends K, ? extends V>> marshalledEntries) {
-      EntityManager em = emf.createEntityManager();
-      try {
-         EntityTransaction txn = em.getTransaction();
-         long txnBegin = timeService.time();
-         try {
-            txn.begin();
-            for (MarshallableEntry entry : marshalledEntries) {
+   public void writeBatch(Publisher<MarshallableEntry<? extends K, ? extends V>> publisher) {
+      long txnBegin = timeService.time();
```
wburns: Technically this should be initialized in the first using block. A Publisher is supposed to be able to be subscribed to multiple times. This would mean if it was retained and reused later it would have the wrong timestamp.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -201,14 +207,18 @@ public void write(MarshallableEntry entry) {
...
-            Flowable.fromPublisher((Publisher<MarshallableEntry>) publisher)
-                  .map(me -> new Store(me.getKey(), me))
-                  .cast(Modification.class)
-                  .toList()
-                  .blockingGet()
-      );
+   public CompletionStage<Void> writeBatch(Publisher publisher) {
+      CompletableFuture<Void> future = new CompletableFuture<>();
+      Flowable.fromPublisher((Publisher<MarshallableEntry>) publisher)
+            .subscribeOn(scheduler)
```
wburns: I would say we don't need the scheduler configured here. The only reason the scheduler should be used in the store itself is to add additional parallelism that you can't obtain by methods on the provided Publisher. For example we could do parallelism to do each batch insertion, but I don't want to do that yet.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
```diff
@@ -201,14 +207,18 @@ public void write(MarshallableEntry entry) {
...
    @Override
-   public void writeBatch(Publisher publisher) {
-      putAll(
-            Flowable.fromPublisher((Publisher<MarshallableEntry>) publisher)
-                  .map(me -> new Store(me.getKey(), me))
-                  .cast(Modification.class)
-                  .toList()
-                  .blockingGet()
-      );
+   public CompletionStage<Void> writeBatch(Publisher publisher) {
```
wburns: No types for the argument?

core/src/main/java/org/infinispan/persistence/manager/PersistenceManagerImpl.java
```diff
@@ -704,10 +707,22 @@ public void writeBatchToAllNonTxStores(Iterable<MarshallableEntry> entries,
       try {
          checkStoreAvailability();
          //noinspection unchecked
-         nonTxWriters.stream()
+         CompletableFuture batchFutures = CompletableFuture.allOf(
```
wburns: This should be alright for now, I will change this if I can ever get to https://issues.jboss.org/browse/ISPN-9722

core/src/main/java/org/infinispan/persistence/support/ComposedSegmentedLoadWriteStore.java
```diff
@@ -191,15 +193,21 @@ public void deleteBatch(Iterable<Object> keys) {
...
-   public void writeBatch(Iterable<MarshalledEntry<? extends K, ? extends V>> marshalledEntries) {
-      Flowable.fromIterable(marshalledEntries)
-            // Separate out batches by segment
+   public CompletionStage<Void> writeBatch(Publisher<MarshallableEntry<? extends K, ? extends V>> publisher) {
+      CompletableFuture<Void> future = new CompletableFuture<>();
+
+      Flowable.fromPublisher(publisher)
             .groupBy(me -> keyPartitioner.getSegment(me.getKey()))
-            .blockingForEach(groupedFlowable ->
+            .serialize()
```
wburns: We don't need this.

core/src/main/java/org/infinispan/persistence/support/ComposedSegmentedLoadWriteStore.java
```diff
@@ -191,15 +193,21 @@ public void deleteBatch(Iterable<Object> keys) {
...
+            .doOnNext(groupedFlowable ->
```
wburns: You should use `flatMap` here instead. To be honest we could parallelize this as I mentioned above, but lets just keep it simple for now.

core/src/main/java/org/infinispan/persistence/support/ComposedSegmentedLoadWriteStore.java
```diff
@@ -191,15 +193,21 @@ public void deleteBatch(Iterable<Object> keys) {
...
                   groupedFlowable
+                        .subscribeOn(scheduler)
                         .buffer(configuration.maxBatchSize())
-                        .blockingForEach(batch -> stores.get(groupedFlowable.getKey()).writeBatch(batch))
-            );
+                        .doOnNext(batch -> stores.get(groupedFlowable.getKey()).writeBatch(Flowable.fromIterable(batch)))
+                        .subscribe(Void -> {}, future::completeExceptionally, () -> future.complete(null))
+                  )
+            .subscribe(Void -> {}, future::completeExceptionally, () -> future.complete(null));
```
wburns: This is more like what I wanted to see :). You can use the static method for the empty consumer instead though.

core/src/main/java/org/infinispan/persistence/async/AsyncCacheWriter.java
ryanemerson: Unfortunately the `AsyncCacheWriter` has never had generics:

https://issues.jboss.org/browse/ISPN-7939

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
```diff
@@ -37,13 +53,38 @@
...
+    * @throws NullPointerException if the publisher is null.
+    */
+   @SuppressWarnings("unchecked")
+   default CompletionStage<Void> writeBatch(Publisher<MarshallableEntry<? extends K, ? extends V>> publisher) {
+      CompletableFuture<Void> future = new CompletableFuture<>();
+      try {
+         writeBatch(
+               Flowable.fromPublisher((Publisher) publisher)
+                     .map(e -> MarshalledEntry.wrap((MarshallableEntry) e))
+                     .blockingIterable()
```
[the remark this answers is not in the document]
ryanemerson: I think for this method we need to call `blockingIterable` as we need to pass an Iterable instance to the legacy `writeBatch(Iterable)`

persistence/jpa/src/main/java/org/infinispan/persistence/jpa/JpaStore.java
ryanemerson: If we do that I'm not sure how we would be able to pass the start time to the error handler for `stats.addBatchWriteTxFailed`

ryanemerson:
@wburns Updated to address your comments. None of the store `writeBatch` publishers utilise a scheduler anymore.

ryanemerson:
I've added a `testWriteAndDeleteBatchIterable` to ensure that the default impl of the deprecated `writeBatch` method functions as expected and delegates down to the appropriate `write` method if no implementation is found.

core/src/main/java/org/infinispan/persistence/spi/SegmentedAdvancedLoadWriteStore.java
```diff
@@ -23,6 +25,28 @@
...
+    * MarshalledValue<K, V> value = load(key);
+    * }
+    * </pre>
+    * @param segment the segment that the key maps to
+    * @param key the key of the entry to fetch
+    * @return the entry, or null if the entry does not exist
+    * @throws PersistenceException in case of an error, e.g. communicating with the external storage
+    */
+   default MarshallableEntry<K, V> get(int segment, Object key) {
+      return load(segment, key);
```
wburns: We should just invoke the get method as the javadoc says :)

core/src/test/java/org/infinispan/persistence/DummyInitializationContext.java
```diff
@@ -69,12 +70,17 @@ public ByteBufferFactory getByteBufferFactory() {
...
    }
 
    @Override
    public ExecutorService getExecutor() {
       return executorService;
    }
+
+   @Override
+   public MarshalledEntryFactory getMarshalledEntryFactory() {
+      throw new UnsupportedOperationException("User InitializationContext::getMarshallableEntryFactory instead");
```
wburns:
```suggestion
      throw new UnsupportedOperationException("Use InitializationContext::getMarshallableEntryFactory instead");
```

persistence/jdbc/src/main/java/org/infinispan/persistence/jdbc/stringbased/JdbcStringBasedStore.java
```diff
@@ -282,39 +286,44 @@ private void executeLegacyUpdate(Connection connection, MarshalledEntry entry, S
...
-               String keyStr = key2Str(entry.getKey());
-               prepareUpsertStatement(entry, keyStr, getSegment(entry), upsertBatch);
-               upsertBatch.addBatch();
-               batchSize++;
-
-               if (batchSize == configuration.maxBatchSize()) {
-                  batchSize = 0;
-                  upsertBatch.executeBatch();
-                  upsertBatch.clearBatch();
+         Flowable.fromPublisher(publisher).blockingForEach(this::write);
```
wburns: Missed making this one non blocking.

persistence/jpa/src/main/java/org/infinispan/persistence/jpa/JpaStore.java
wburns: You can pass it by changing the inner publisher's returned value to a single value and mapping that to the time.
`
.takeLast(1).map(ignore -> startTime)
`
I am sure there is a better way to do this, but can't think of it atm. Maybe encapsulating the whole Publisher in a flatMap operation or something.

persistence/remote/src/main/java/org/infinispan/persistence/remote/RemoteStore.java
```diff
@@ -85,6 +89,7 @@
    private static final String LIFESPAN = "lifespan";
    private static final String MAXIDLE = "maxidle";
    protected InitializationContext ctx;
+   private Scheduler scheduler;
```
wburns: Lets remove this guy.

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
danberindei: This seems to work:

```java
   default <K1 extends K, V1 extends V> void writeBatch(Iterable<MarshalledEntry<K1, V1>> entries) {
      entries.forEach(this::write);
   }
```

persistence/jpa/src/main/java/org/infinispan/persistence/jpa/JpaStore.java
```diff
@@ -441,24 +465,19 @@ public void writeBatch(Iterable<MarshalledEntry<? extends K, ? extends V>> marsh
...
-         }
-      } finally {
-         em.close();
-      }
+            })
+            .doOnComplete(() -> {
+               stats.addBatchWriteTxCommitted(timeService.time() - txnBegin);
+               txn.commit();
+            })
+            .doOnError(e -> stats.addBatchWriteTxFailed(timeService.time() - txnBegin))
```
ryanemerson: @wburns I ended up not using a single and just moving the stats collection to within the batch flowable method.

core/src/main/java/org/infinispan/persistence/spi/CacheWriter.java
danberindei: After discussing this with Will I suggest renaming `writeBatch` to `writeBulk` and removing this from the javadoc:

```
 * Persist all provided entries to the store in a single batch update. If this is not supported by the
 * underlying store, then entries are written to the store individually via {@link #write(MarshallableEntry)}.
```

core/src/main/java/org/infinispan/marshall/core/MarshalledEntryFactory.java
```diff
@@ -1,11 +1,21 @@
...
  *
  * @since 6.0
- * @deprecated since 10.0, use {@link org.infinispan.persistence.spi.MarshalledEntryFactory} instead
+ * @deprecated since 10.0, use {@link MarshallableEntryFactory} instead
  */
 @Deprecated
-public interface MarshalledEntryFactory<K,V> extends org.infinispan.persistence.spi.MarshalledEntryFactory<K,V> {
+public interface MarshalledEntryFactory<K,V> extends MarshallableEntryFactory<K,V> {
+
+   MarshalledEntry<K,V> newMarshalledEntry(ByteBuffer key, ByteBuffer valueBytes, ByteBuffer metadataBytes);
```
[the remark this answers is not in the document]
ryanemerson: Yeah good point. I think I did this as a convenience for `InitializationContextImpl` so I could pass one implementation, as a single param, for both `MarshalledEntryFactory` and `MarshallableEntryFactory`. But that's definitely not a good idea in hindsight :laughing:

core/src/main/java/org/infinispan/persistence/spi/SegmentedAdvancedLoadWriteStore.java
```diff
@@ -158,8 +231,10 @@ default boolean delete(int segment, Object key) {
     * @param fetchMetadata whether the metadata should be included in the marshalled entry
     * @return a publisher that will provide the entries from the store that map to the given segments
     */
-   Publisher<MarshalledEntry<K, V>> publishEntries(IntSet segments, Predicate<? super K> filter, boolean fetchValue,
-         boolean fetchMetadata);
+   default Publisher<MarshallableEntry<K, V>> entryPublisher(IntSet segments, Predicate<? super K> filter, boolean fetchValue,
```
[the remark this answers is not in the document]
ryanemerson: I don't think it is, because the 9.4.x store will not compile due to the required import changes.

core/src/main/java/org/infinispan/persistence/spi/CacheLoader.java
```diff
@@ -29,8 +30,24 @@
...
+
+   /**
+    * Fetches an entry from the storage. If a {@link MarshallableEntry} needs to be created here, {@link
+    * InitializationContext#getMarshallableEntryFactory()} ()} and {@link
+    * InitializationContext#getByteBufferFactory()} should be used.
+    *
+    * @return the entry, or null if the entry does not exist
+    * @throws PersistenceException in case of an error, e.g. communicating with the external storage
+    */
+   default MarshallableEntry<K, V> get(Object key) {
```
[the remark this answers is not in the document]
danberindei: I was thinking stores could still return `MarshalledEntry`, because the language allows covariant return types. But it requires recompilation, otherwise there's no bridge method returning a `MarshallableEntry` in the store implementation...

I'm still not sold on `get()` though, it's weird to have an interface `CacheLoader` where the without a proper `load` method. I quickly gave up on generating the bytecode directly to avoid the compiler's silly restriction on methods that only differ in their return type, so now I'm trying to come up with something that includes `load` in the name, like `loadEntry(key)` or `load(key, includeExpired)`. WDYT @wburns @tristantarrant?

core/src/main/java/org/infinispan/persistence/spi/SegmentedAdvancedLoadWriteStore.java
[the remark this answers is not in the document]
danberindei: The source of the 9.4.x store wouldn't have to change, it could still return a `Publisher<MarshalledEntry<K, V>>`. But it would have to be recompiled in order to generate the bridge method (even though the erased signature is the same, I think), and we should have binary compatibility as much as possible.

core/src/main/java/org/infinispan/persistence/spi/SegmentedAdvancedLoadWriteStore.java
```diff
@@ -23,6 +25,28 @@
...
+    * Fetches an entry from the storage given a segment to optimize this lookup based on. If a
+    * {@link MarshallableEntry} needs to be created here, {@link InitializationContext#getMarshallableEntryFactory()} and
+    * {@link InitializationContext#getByteBufferFactory()} should be used.
+    * <p>
+    * The provided segment may be used for performance purposes, however it it is acceptable to ignore this argument.
+    * <p>
+    * This method may be invoked invoked irrespective if the store is {@link StoreConfiguration#segmented()}.
+    * @implSpec Default implementation just invokes the {@link CacheLoader#get(Object)} method.
+    * <pre> {@code
+    * MarshalledValue<K, V> value = load(key);
```
danberindei: Reference to load

core/src/main/java/org/infinispan/persistence/spi/CacheLoader.java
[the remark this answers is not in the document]
ryanemerson: > I was thinking stores could still return `MarshalledEntry`

So that's a problem in itself, as it means we can't properly deprecate `MarshalledEntry` as there will be no new method for the user to migrate to. So at what point can we remove it without breaking compatibility?

ryanemerson:
@danberindei With regards to removing `? extends K` that will also mean changes to the `PersistenceManager` interface as it part of the `Predicate<? extends K> predicate` param on several methods. In addition to our new methods not being consistent with the old, but I appreciate change has to happen at sometime.

ryanemerson:
I've created [ISPN-9846](https://issues.jboss.org/browse/ISPN-9846) to describe what this PR is trying to achieve. There is a lot of context here that should be captured and shouldn't be grouped with [ISPN-9693](https://issues.jboss.org/browse/ISPN-9693)

danberindei:
@ryanemerson I keep thinking moving `MarshallableEntry` to the SPI package generates a lot more churn than initially expected, with not a lot of benefits, so maybe it's not worth making the change?

However, depending on the `o.i.marshall.core` package is also unappealing, and we do know now what it takes to change the SPI, so I do have an alternative proposal :)

How about we add transcoding support to the stores? If a store could require a particular format, then `SingleFileStore` could implement `AdvancedLoadWriteStore<Object, WrappedByteArray>`, because it would request keys in the data container storage format and it would only require marshalling to another (configurable per store?) format for values and metadata. Then we wouldn't need a separate `MarshalledEntry`/`MarshallableEntry`, we could use `CacheEntry<Object, WrappedByteArray>` (although I would like to move the context-specific stuff like `isCreated()` to another interface).

ryanemerson:
@danberindei These changes are intended to ensure that we have a clear separation of the persistence SPI and internal classes/interfaces. With [ISPN-9845](https://issues.jboss.org/browse/ISPN-9845) following up on this work to remove `InternalMetadata` that is wrongly exposed at present, which will ultimately make my marshalling worksimpler. I think these goals are still desirable (and most of the work already done). 

I like your idea of adding transcoding support to stores, but I don't think it solves the problem that this and associated work is trying to solve. If you implements SFS as ` AdvancedLoadWriteStore<Object, WrappedByteArray>`, you're still calling the SPI using internal classes like `MarshalledEntry` right?

I appreciate their is a lot of churn in these changes, but if we don't bite the bullet at some point, then we will forever be stuck with an SPI exposing internal classes.

danberindei:
> If you implements SFS as AdvancedLoadWriteStore<Object, WrappedByteArray>, you're still calling the SPI using internal classes like MarshalledEntry right?

@ryanemerson `MarshalledEntry`/`MarshallableEntry` is just an interface, so we could use `CacheEntry<Object, WrappedByteArray>` anywhere we are now using `MarshalledEntry`. And once we remove `getInternalMetadata()` and `get*Bytes()` on one side, and `is*` on the other, they look very much alike.

There is one hitch: I just noticed that the `org.infinispan.container.entries` contains all the entry implementations and as a consequence `CacheEntry` is not in the public javadoc. The good part is that means we need a new interface for the API anyway, and I believe we can use the same interface both for the API and for the SPI.

ryanemerson:
@danberindei So we need another interface that we pass through to the Writer and Loader interfaces, which means that we will *still* need to create all of the backward compatibility default methods that I have in this PR, albeit for an interface that is not called `MarshallableEntry`. So there will be a lot of churn regardless, which was your original objection.

> If a store could require a particular format, then `SingleFileStore` could implement `AdvancedLoadWriteStore<Object, WrappedByteArray>`, because it would request keys in the data container storage format and it would only require marshalling to another (configurable per store?) format for values and metadata

What if the storage format is simply Object and we're talking about any store that is not SFS? We will still need to marshall the key and we have made our code much more complicated. Also a  limitation of the transcoding approach with stores is that we lose the ability to write to streams, as the current encoder interface only provides `to/fromStorage` methods. We currently write to streams in the JDBC store and I know that @wburns would be very unhappy if we removed this capability as it would prevent future performance improvements. 

Finally, one thing that is nice about `MarshalledEntry` and `MarshallableEntry` having the `get*Bytes` methods, is that we only have to marshall keys/values/metatadata once, even if there are multiple stores. Moving towards an approach more centred towards `CacheEntry` means that we will lose this.

ryanemerson:
@danberindei Also, if we pass `CacheEntry<Object, WrappedByteArray>` are we not preventing store's from storing metadata and values independently without having to unmarshall the passed bytes?

ryanemerson:
> request keys in the data container storage format

If we went with this approach, we would have to change the `BinaryEncoder` to utilise the user marshaller instead of the GlobalMarshaller so that internal changes do not affect the keys storage format, but this change should probably happen anyway.

danberindei:
> @danberindei So we need another interface that we pass through to the Writer and Loader interfaces, which means that we will _still_ need to create all of the backward compatibility default methods that I have in this PR, albeit for an interface that is not called `MarshallableEntry`. So there will be a lot of churn regardless, which was your original objection.

Sure, the only way to avoid the churn is to keep using the existing interface in `o.i.marshall.core`. And I think that would work, as `MarshalledEntry` is an internal class in (package) name only, and only used internally for interacting with the persistence SPI. IMO we could just include it in the javadoc, without any churn.

I just started thinking about what changes I would like to see in the persistence SPI in the future, and the idea that each store should be able to use its own encoding kind of stuck out, so I derailed the discussion towards that. Sorry :)

> What if the storage format is simply Object and we're talking about any store that is not SFS? We will still need to marshall the key and we have made our code much more complicated.

I picked SFS as an example because it doesn't need to marshall the keys, but of course most stores would implement `CacheEntry<WrappedBytesArray, WrappedBytesArray>`. The persistence manager would be more complex, but the stores should be just like now except they would use `getKey()` and `getValue()` instead of `getKeyBytes()` and `getMetadataBytes()`.

> Also a limitation of the transcoding approach with stores is that we lose the ability to write to streams, as the current encoder interface only provides `to/fromStorage` methods. We currently write to streams in the JDBC store and I know that @wburns would be very unhappy if we removed this capability as it would prevent future performance improvements.

Hmmm, this is what the `MarshalledEntryImpl` externalizer does now:

         output.writeObject(me.getKeyBytes());
         output.writeObject(me.getValueBytes());
         output.writeObject(me.getMetadataBytes());

> Finally, one thing that is nice about `MarshalledEntry` and `MarshallableEntry` having the `get*Bytes` methods, is that we only have to marshall keys/values/metatadata once, even if there are multiple stores. Moving towards an approach more centred towards `CacheEntry` means that we will lose this.

Sure, caching is not possible when each store has a different encoding, but if they use the same encoding, the persistence manager could still cache the marshalled keys/values in a separate object.

ryanemerson:
Squashed

danberindei:
Integrated, thanks Ryan, and please excuse my ranting :)

TuomasKiviaho:
> How about we add transcoding support to the stores?

I'm curious to know if this idea went any further. Is there a ticket that I could follow? 

I'd personally benefit from this with the JPA store especially when I'm using it in read-oriented manner. Store could enhance the entities or replace the entities from/to a more appropriate format (and vice versa) already inside the entity manager transaction. I can of course override the factory, but there wouldn't be an easy way (afaik) to determine if it's a read or write operation that I'm dealing with or which context is being currently used since it's a global factory. The API encoder would not be any good in this case because the work is done outside of the transaction and that approach would not have an impact how the entities eventually end up being stored in-memory.

ryanemerson:
@TuomasKiviaho There aren't any plans to add this any time soon, certainly not Infinispan 10, due to other priorities. However, it could be an interesting feature to add in the future. There isn't a JIRA ticket to track this feature at the minute. It would be great if you could create one [here](https://issues.jboss.org/projects/ISPN/issues) with a detailed explanation about your use-case and requirements so we have a better indication of how users would like to use/benefit from this feature.

PR #6543, opened 2018-12-14
ISPN-9828 Server defaults missing for UFC_NB and MFC_NB

danberindei:
https://issues.jboss.org/browse/ISPN-9828


danberindei:
run performance tests please
cs

diegolovison:
run performance tests please
cs

diegolovison:
run performance tests please
cs

danberindei:
Performance looks good. Non-tx reads (both repl and dist) are still about 5% slower than the 7.2 baseline, but no messages are sent in those scenarios so it's probably from the server itself.

PR #6548, opened 2018-12-14
ISPN-9831 IdleStateHandlerProvider's log should be static

wburns:
https://issues.jboss.org/browse/ISPN-9831


danberindei:
Integrated, thanks Will!

PR #6564, opened 2018-12-17
ISPN-7245 ISPN-9476 ISPN-9540 Remove deprecated classes

anistor:
https://issues.jboss.org/browse/ISPN-7245
https://issues.jboss.org/browse/ISPN-9476
https://issues.jboss.org/browse/ISPN-9540


core/src/main/java/org/infinispan/lifecycle/AbstractModuleLifecycle.java
```diff
@@ -1,59 +0,0 @@
...
-import org.infinispan.factories.GlobalComponentRegistry;
-
-
-/**
- * A default, no-op implementation of the {@link org.infinispan.lifecycle.ModuleLifecycle} interface, designed for easy
- * extension.
- *
- * @author Manik Surtani
- * @version 4.0
- * @deprecated since 9.1. Just implement {@link ModuleLifecycle} instead.
```
danberindei: Not sure about this, it wasn't deprecated in 9.0

anistor: So in 10.0 we are only allowed to remove stuff deprecated in 9.0 but not 9.1+ ? So this and other cobwebs will have to be carried around until 11 ?

tristantarrant:
No please, let's kill it now

danberindei:
@anistor @tristantarrant what's the reason to delete it now? Does deleting it help us in any way?

anistor:
when would be the best moment to delete them @danberindei ?

danberindei:
I would say either in the 2nd major version in which they've been deprecated, or when maintaining them is too big of a burden (or impossible).

anistor:
Sounds like a good general common sense principle :) But what is the value in keeping these particular classes? BTW, some where deprecated in 7, 8 and I suspect they can be maintained forever since they are trivial. Let's stuff the dead cat and keep it around :))

danberindei:
Maybe I'm just wishful thinking, but I hope knowing that once we add an API we'll have to maintain it for at least 2 major versions will motivate us to be more careful about what we expose in the API. Think of it as a committment device @anistor :)

anistor:
That's perfectly right for public APIs but ModuleLifecycle's javadoc reads: ```"ModuleLifecycle is an internal API hook for  ..."```. The _internal_ part makes me think we do not need to be that extra cautious with this one, especially since that release was made more that one year ago.

danberindei:
Is it really an "internal API hook" though (whatever that means)? I thought modules are needed to define a custom commands, and defining custom commands is part of the API proper.

anistor:
We can ask the original author, but I think 'internal API' means SPI. Module lifecycle and custom commands are only used by a handful of modules within the infinispan org. AFAIK module lifecycle is not even mentioned in the user guide.

tristantarrant:
I've cut the conversation short by merging this. Executive decision.

PR #6566, opened 2018-12-17
ISPN-9771 Support for configuration in JSON format

gustavonalle:
https://issues.jboss.org/browse/ISPN-9771

Adds support Configuration => JSON conversion and back

There is a couple of minor TODOS but I don't anticipate any other change (unless reviewed otherwise) and they can go on follow up JIRAs:

- [x] Remove some redundant elements (this is also a problem of the existing XML conversion)
- ~~[x] Tweak JSON array creation for a couple of cases~~  Gave-up, as it introduces ambiguity

I created a docker image with this branch to test the conversion, to run it:

```
docker run -p 8080:8080 -e "APP_USER=user" -e "APP_PASS=pass" gustavonalle/infinispan-server:10.0.0-SNAPSHOT
```

To retrieve a cache configuration by name in JSON format:

```
curl -u user:pass "http://localhost:8080/rest/v2/configurations/indexed" | python -m json.tool
curl -u user:pass "http://localhost:8080/rest/v2/configurations/persistent-jdbc-string-keyed" | python -m json.tool
```

To retrieve the same configurations but in XML:
```
curl -u user:pass -H "Accept: application/xml" "http://localhost:8080/rest/v2/configurations/persistent-jdbc-string-keyed"
curl -u user:pass -H "Accept: application/xml"  "http://localhost:8080/rest/v2/configurations/indexed"
```

Finally to convert an existing XML configuration to JSON, save it to a file and do a POST:
```
curl -u user:pass --data-binary @infinispan.xml "http://localhost:8080/rest/v2/configurations?action=toJSON" 
```


commons/src/main/java/org/infinispan/commons/configuration/Json.java
```diff
@@ -0,0 +1,3247 @@
...
+    * @param property The property name.
+    * @param value The value of the property.
+    * @return this
+    */
+   public Json set(String property, Json value) {
+      throw new UnsupportedOperationException();
+   }
+
+   /**
+
```
tristantarrant: Do we need this `main()`?

gustavonalle: This class is from mJSON library, I avoided changing it to facilitate upgrades in the future.

gustavonalle:
> It looks good to me. I'd like to see some tests though (or maybe I missed them entirely): the only one I can find is JsonSerializationTest.java but I'd really like to see some cache creation over rest.

Cache creation over REST is not done yet, that's the next step. Since the PR was getting large, I decided to stop it :)

Tests are mainly done on the XMLs already available throughout the test suite, by doing XML -> Configuration -> JSON -> Configuration transformations.

Inside ```JsonSerializationTest``` it tests the https://github.com/infinispan/infinispan/blob/master/core/src/test/resources/configs/unified/10.0.xml that contains dozens of cache configs. Plus  ```AbstractConfigurationSerializerTest``` has been enhanced, so every module outside ```core``` that has XML parsing/serializing test is also testing JSON.

tristantarrant:
Merged. Thanks.

PR #6572, opened 2018-12-18
[9.4.x] ISPN-4075 State transfer should preserve the timestamps of entries

danberindei:
Backport of https://github.com/infinispan/infinispan/pull/6525
https://issues.jboss.org/browse/ISPN-4075

* Use PutKeyValueCommand with InternalCacheEntry as value to pass the
  created and last use timestamps
* Scattered caches also use PutMapCommand with InternalCacheValue values
* Always copy the timestamps when copying entries
* Ignore -1 timestamp values and use current time instead


danberindei:
`CoordinatorStopTest` failure is from ISPN-9588

wburns:
Integrated into 9.4.x, thanks @danberindei !

PR #6573, opened 2018-12-19
ISPN-9840 RemoteCacheManager shouldn't attempt to start a second time

wburns:
https://issues.jboss.org/browse/ISPN-9840


danberindei:
Integrated, thanks Will!

PR #6575, opened 2018-12-20
ISPN-9843 RocksDBStore should not write entire MarshalledEntry as sto…

ryanemerson:
…re value

https://issues.jboss.org/browse/ISPN-9843


ryanemerson:
This will break store compatibility, but that is already going to happen with the pending marshalling changes. Issuing as separate PR for simplicity.

danberindei:
The key is usually much smaller than the value, I was much more interested in getting rid of the separate expiration DB :)

After reading a bit more about RocksDB, however, I'd be happy with just https://issues.jboss.org/browse/ISPN-8242

tristantarrant:
Yes, using a single db with column families reduces open file descriptors.

danberindei:
Integrated, thanks Ryan!

PR #6587, opened 2019-01-10
ISPN-9861 Marshallers for remote querying should be public

wburns:
Marshallers for remote querying should be public

https://issues.jboss.org/browse/ISPN-9861


anistor:
-1 from me. I've commented on the jira

wburns:
Closing as a different solution can be done instead.

PR #6597, opened 2019-01-16
[9.4.x] ISPN-9873 Remove listeners when lock is removed

karesti:
Backport https://issues.jboss.org/browse/ISPN-9873


danberindei:
Integrated, thanks Katia!

PR #6601, opened 2019-01-17
[9.4] ISPN-9824 Move doclet plugin to an external project

tristantarrant:
https://issues.jboss.org/browse/ISPN-9824


danberindei:
Kinda disappointing that we had to do this, but it's in now :)

PR #6602, opened 2019-01-17
ISPN-9871 ProtobufMetadataManagementInterceptor ClassCastException

danberindei:
https://issues.jboss.org/browse/ISPN-9871

StateConsumerImpl uses PutKeyValueCommands with InternalCacheEntry
values in order to preserve timestamps, so read the value from the
context


anistor:
QueryInterceptor could be affected too. A quick glance indicates it accesses the value from context rather than the command, so maybe not an issue.

danberindei:
@anistor yeah, I checked `QueryInterceptor` and it doesn't seem affected. Plus I'm sure query has some tests with state transfer ;)

anistor:
Some tests indeed. Not sure those can cover the infinite oddities that can result from apparently minor changes in our interceptor stack. Which makes me think we need to hide that from users for good, and the commands too.

anistor:

